PR #24, opened 2015-08-07
Added Visual Studio 2015 as a target.

reghrafa:


davelamb:
We've fixed this on our side. The vsix found in WinObjC/bin (not in the SDK zip) has been updated to work with 2015. Give it a try, and let us know if you have any more problems. Thanks!

Issue #54, opened 2015-08-10
Define a code style to follow for all sources


vmagaziy:
Right now it seems there are no conventions for code style. Different sources are formatted differently, it order to have a coherent look for all sources it' proposed to choose of the options:

1) [LLVM] (http://llvm.org/docs/CodingStandards.html)
2) [Google] (http://google.github.io/styleguide/objcguide.xml)
3) [Chromium] (https://www.chromium.org/developers/coding-style)
4) [Mozilla] (https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Coding_Style
5) [WebKit] (https://www.webkit.org/coding/coding-style.html)

and follow them by using git pre-commit hooks driven by [clang-format] (http://clang.llvm.org/docs/ClangFormat.html). The styles from the list are supported out of the box.

I have an experience with setting this infrastructure up, so I am ready to proceed with re-formatting and adding the mentioned hooks. What I need is the decision which code style is to use.

jmsaunders-ms:
Partial +1

I'm all for using clang-format and have been pushing a bit to have that done. I'd specifically not use the LLVM style though since VariableNames look strange to me. Further, we make (imo) good use of exceptions in some of our tools, which goes against some of the style guides.

I'd prefer I think to base it on the Google style guide with some modifications. This is likely coming but I wouldn't expect it this week. It's a bit prickly because none of them are specifically Objective C style guides and that's the majority of  our code.

More on this later.

vmagaziy:
@jmsaunders-ms thanks for your answer and support of this idea, can you please have a look at [PL](https://github.com/Microsoft/WinObjC/pull/70) with initial changes related to this issue?

vmagaziy:
@s5msft, please close this issue or mark it as addressed by your team, otherwise someone else may decide to make the same mistake I did with my **PL**.

nicholasgerard-msft:
We've now published coding guidelines and standards as part of our wiki. Please take a look at [How to Contribute](https://github.com/Microsoft/WinObjC/wiki/How-to-Contribute) and [ClangFormat](https://github.com/Microsoft/WinObjC/wiki/ClangFormat) for more information. Thanks!

Issue #59, opened 2015-08-11
Define a roadmap for test harness


vmagaziy:
In order to have a test harness for checking that implemented frameworks and native ones have the same behaviour it has to create unit tests and run them on a Mac. For sure, there is a huge difference in visual representation of UI, so it won't be possible to have a complete coverage for UIKit, but will be possible to create such tests for all underlying frameworks like Foundation. The idea is to create a set of tests and run them on a Mac with native frameworks and on Windows with current implementation in order to make sure there is no difference. Can you please confirm whether this approach makes sense and whether you are ready go give it a go ahead?

s5msft:
@vmagaziy A test framework, based on GHUnit, will be along shortly.

> Can you please confirm whether this approach makes sense

Absolutely makes sense. Let us get our test framework in first :)

jaredhms:
@vmagaziy - apologies for the late response.  We're now using gtest for all of our unit tests, and we do have work planned to look at producing mac builds of our gtests, but we haven't had time to do so yet.  Any help would certainly be appreciated.

It's also worth mentioning that  we'll be rolling out some functional UAP test support in the ~near future, although those tests won't be good mac candidates.

PR #70, opened 2015-08-13
Clang format and configuration for code style

vmagaziy:
This **PL** contains initial preparations discussed in Issue #54 (_Define a code style to follow for all sources_):
* Initial version of Python pre-commit hook, so it requires having Python installed
* Scripts for installation the hook
* **clang-format** binaries v3.6.2 for Windows and OS X
* Configuration for **clang-format** based on [Google Code Style] (http://google.github.io/styleguide/objcguide.xml)

This setup has been tested on OS X and Windows only and in order to have it work properly on Windows it has to checkout sources with LF endings (another word it has to have `git config --global core.autocrlf false` executed before the checkout).

Please note, PowerShell environment is supposed to used for running Git, so all suggestions are provided assuming such commands as `cat` and `patch` are supported, please see example below:

```
C:\Users\USER\Documents\GitHub\WinObjC [clang_format_and_configuration_for_code_style +0 ~1 -0]> git diff
diff --git a/Frameworks/Foundation/NSArray.mm b/Frameworks/Foundation/NSArray.mm
index 061b13c..d3fb9aa 100644
--- a/Frameworks/Foundation/NSArray.mm
+++ b/Frameworks/Foundation/NSArray.mm
@@ -33,7 +33,7 @@ __declspec(dllimport) extern "C" int CFNSBlockCompare(id obj1, id obj2, void *bl
 @class NSXMLPropertyList, NSPropertyListReader, NSArrayConcrete, NSMutableArrayConcrete, NSPropertyListWriter_Binary;

 @implementation NSArray : NSObject
-    +(NSArray*) arrayWithObjects:(NSObject*)first , ...{
+    +(             NSArray                *) arrayWithObjects:(           NSObject * )first , ...{
         va_list pReader; va_start(pReader, first);
         NSArray* ret = [self new];
         if ( first == nil ) return [ret autorelease];
C:\Users\USER\Documents\GitHub\WinObjC [clang_format_and_configuration_for_code_style +0 ~1 -0]> git commit -a
ERROR: INCORRECT CODE STYLE
If you believe it is correct, use '-n' or '--no-verify' option to ignore this error.
Otherwise execute 'cat c:\users\user\appdata\local\temp\com.microsoft.WinObjC.codestyle.jeovzb.patch | patch -p0' to fix found issues.
Do NOT forget to add changes to the index if needed or use '-a' option, otherwise they will be lost.
--- Frameworks/Foundation/NSArray.mm    (before formatting)
+++ Frameworks/Foundation/NSArray.mm    (after formatting)
@@ -30,13 +30,15 @@

 __declspec(dllimport) extern "C" int CFNSBlockCompare(id obj1, id obj2, void *block);

-@class NSXMLPropertyList, NSPropertyListReader, NSArrayConcrete, NSMutableArrayConcrete, NSPropertyListWriter_Binary;
+@class NSXMLPropertyList, NSPropertyListReader, NSArrayConcrete,
+    NSMutableArrayConcrete, NSPropertyListWriter_Binary;

 @implementation NSArray : NSObject
-    +(             NSArray                *) arrayWithObjects:(           NSObject * )first , ...{
-        va_list pReader; va_start(pReader, first);
-        NSArray* ret = [self new];
-        if ( first == nil ) return [ret autorelease];
++ (NSArray *)arrayWithObjects:(NSObject *)first, ... {
+  va_list pReader;
+  va_start(pReader, first);
+  NSArray *ret = [self new];
+  if (first == nil) return [ret autorelease];
         CFArrayAppendValue((CFMutableArrayRef) ret, (const void *) first);

         id curVal = va_arg(pReader, id);

```


asimonov-msft:
@vmagaziy Checking out sources with LF line endings under Windows may be an issue. This has caused us serious problems in the past, so now we use `core.autocrlf true`. In particular, git would occasionally become confused about the mixed line endings in modified files, and would consider all contents of the file to be changed.

vmagaziy:
I don't have any experience of using Git under Windows, but I believe it
may only be the case when the user editor does not appreciate original line
endings.

vmagaziy:
@asimonov-msft , if resetting the `autocrlf` flag is not an option, I will try to rework the solution to appreciate CRLF line endings for Windows.

vmagaziy:
@asimonov-msft, please review this **PL** with fixes.

s5msft:
@vmagaziy we love this suggestion, and we have a tool to accomplish (essentially) this same task coming in the very near future

vmagaziy:
@s5msft, thanks but this just kills all my initiative to contribute to this project.  If you've got something happening under the hood, please make it public, so it becomes visible to the community and we don't have to reinvent the wheel.  I would be also great if you define a set of tasks and priorities which can be taken by the community to avoid duplications like this one.

jmsaunders-ms:
Hi @vmagaziy, I've been working on this a little bit. The reason why we don't want to accept your commit is that we have a lot of reservations about having a dependency on Python. I'm also not a huge fan of pre-commit hooks in general but emitting a patch is IMO not a great solution.

I've ported the important bits of the script to C#, which will instead function on the working version of the diffs (rather than staged). The plan is to show a diff which the user can accept or deny.

This work started after your pull request. We're happy with the broad strokes of what you want to do and I apologize for not making this explicit earlier.

vmagaziy:
@jmsaunders-ms, Thanks for your message and sharing the ideas. My initial thought was that Python could be used for both Windows and Mac and other platforms, but if the hook/verification tool is a native application it automatically means it's for Windows only. For instance, I wanted to contribute to implementation of Foundation and other non UI frameworks as well as create unit tests by running them on OS X (see #59 for details) and in that case it will be really helpful, as well as if someone running Linux may decide to tweak some implementation. But I 100% agree that having an additional dependency on Python is a bad thing -- on OS X, Python is pre-bundled which is not the case on Windows. AFAIK, it's now possible to run C# code on different platforms, so can you please provide binaries of your tool for different platforms once it's ready?

vmagaziy:
@jmsaunders-ms, having this tool executed by an intention is an error prone approach in my opinion, since authors of PLs may forget to use it while committing/pushing changes. For sure, if you're planning to create a script for submitting changes like in `git cl upload` used in *Chromium*, then this tool can be integrated and it becomes impossible to forget to execute it.

jmsaunders-ms:
Sorry - this is taking longer than I expected and I wanted to give some updates.

Clang-format has had some issues with our source base and we've had trouble with a couple things:
1. The preceding whitespace (and comments) on method definitions causes clang-format to severely break methods following empty methods. This is fixable with a quick sed prior to running the tool, but that is not always an effect-free process (multiline strings).
2. Laying out c-style struct initializers does not work great for us (especially with the C++11 option on, sadly). We would like to preserve some formatting of them (matrices, setting up some function pointer tables, etc). None of the options (I have seen) support behaviour that satisfies our needs.
3. The regular bike-shedding on choosing a coding style. :)

Right now I'm strongly considering running a pass over the code base once, cleaning up stuff I disagree with and pushing and then spending idle cycles trying to make clang-format work better for our purposes. Until then a commit hook to reject pull requests that don't match the style is untenable.

PR #105, opened 2015-08-19
Implement the new OS version APIs from OS X 10.10 and iOS 8.0.

jakepetroules:
I don't feel like setting up a Windows 10 instance to test this right now, but the winOsVersion() function was tested separately and works properly on Windows Desktop and WinRT (returns 6.3.9600 on my Windows 8.1 box).


jakepetroules:
Ping. I fixed the build issue...

PR #197, opened 2015-09-23
Fixed NSObject valueForKey: pointer return values

triplef:
Calling `-[NSObject valueForKey:]` with a key referencing a method returning a pointer was always returning NULL due to an incorrect pointer reference in `tryGetAccessor()`.


DHowett-MSFT:
This ought to be fixed as of afebcca

triplef:
Confirmed.

PR #200, opened 2015-09-24
Added missing NSValue unwrapping for KVC -setValue:forKey:

triplef:
In order for KVC to work with setter-methods for primitive types, NSValue objects passed to `-[NSObject setValue:forKey:]` have to be unwrapped.

This fixes code like the following, where previously the NSValue object was passed to the setter resulting in undefined behavior:
```objc
- (void)setFloat:(float)value
{
	_float = value;
}

- (void)test
{
    [self setValue:[NSValue valueWithFloat:1.23] forKey:@"float"];
}
```

### DIFF
--- a/Frameworks/limbo/NSObject_NSFoundation.mm
+++ b/Frameworks/limbo/NSObject_NSFoundation.mm
@@ -136,6 +136,85 @@
         return ret;
     }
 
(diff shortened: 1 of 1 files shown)


DHowett-MSFT:
This ought to be fixed as of afebcca

triplef:
Confirmed.

PR #201, opened 2015-09-25
Implemented -[NSConditionLock tryLock]

triplef:


rajsesh-msft:
Thank you for your contribution!

technotrance:
The platform is in ntsc and it is run under a 32bit system and the uploads require their own format play platform for eg a mp3 requires a mp3 programme software to run

PR #227, opened 2015-11-02
KVC: fixed setting primitive type accessor using non-matching NSValue type

triplef:
When setting a primitive type accessor via KVC, the code in `trySetViaAccessor()` didn’t consider that the passed-in NSValue instance might be of a different type than the setter’s argument type.

This caused e.g. the following to not work correctly:
```obj-c
// assume `object` with property:
@property (assign) float floatProperty;

[object setValue:@(1.23) forKey:@"floatProperty"]; // note @(1.23) is of type double

=> object.floatProperty != 1.23
```

### DIFF
--- a/Frameworks/Foundation/NSValueTransformers.h
+++ b/Frameworks/Foundation/NSValueTransformers.h
@@ -76,6 +76,73 @@ id valueFromDataWithType(void* data, const char* objcType) {
     }
     return [NSValue valueWithBytes:data objCType:objcType];
 }
+
+bool dataWithTypeFromValue(void* data, const char* objcType, id value) {
+	switch (objcType[0]) {
+	case '@':
+		woc::ValueTransformer<id>::store(value, data);
+		break;
+	case '#':
+		woc::ValueTransformer<Class>::store(value, data);
+		break;
+	case 'c':
+		woc::ValueTransformer<char>::store(value, data);
+		break;
+	case 'i':
+		woc::ValueTransformer<int>::store(value, data);
+		break;
+	case 's':
+		woc::ValueTransformer<short>::store(value, data);
+		break;
+	case 'l':
+		woc::ValueTransformer<long>::store(value, data);
+		break;
+	case 'q':
+		woc::ValueTransformer<long long>::store(value, data);
+		break;
+	case 'C':
+		woc::ValueTransformer<unsigned char>::store(value, data);
+		break;
+	case 'I':
+		woc::ValueTransformer<unsigned int>::store(value, data);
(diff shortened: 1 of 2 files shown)


DHowett-MSFT:
Hi triplef! Thanks for catching this. Would you be amenable to unifying these value-transformed stores with the ones in trySetViaIvar? The fewer places we can do ObjC type parsing, the better! There might be something in ValueTransformers.h we can use for that.
&nbsp;
Thanks!

triplef:
Yes good point! I’ve unified the code in `trySetViaAccessor()` and `trySetViaIvar()` by implementing a new function `dataWithTypeFromValue()` in NSValueTransformers.h.

This also fixes an issue in `trySetViaIvar()` where setting an object-type value was not retained on the ivar.

DHowett-MSFT:
Thanks!

Thinking a bit more about this, I'm interested in sussing out the behaviour of setting an `id` ivar.
Unless we can interrogate the property (which might not exist), we can't exactly know whether to retain new/release old... I'm somewhat worried prescribing that we should always retain/release the value in light of that.

Do you know how it works on iOS / OS X? If they always retain new/release old, we can and absolutely should mimic them. :)

triplef:
Good point. I couldn’t find any documented behavior regarding retain/release of directly accessed instance variables.

Using the following code as a quick test, it seems that both iOS and OS X definitely retain the object, but don’t seem to release it, which is what WinObjC did as well before my change. Not sure why it would behave that way, but I can definitely remove the change regarding retain/release from this PR and keep the original behavior (retain only). Let me know.

```obj-c
...
{
	NSObject	*_property;
}
... 

id object = [[NSObject alloc] init];
NSLog(@"retain count: %u", [object retainCount]);
	
[self setValue:object forKey:@"property"];
NSLog(@"retain count: %u", [object retainCount]);
	
[self setValue:nil forKey:@"property"];
NSLog(@"retain count: %u", [object retainCount]);
	
[self setValue:object forKey:@"property"];
NSLog(@"retain count: %u", [object retainCount]);
```

Result:
```
retain count: 1
retain count: 2
retain count: 2
retain count: 3
```

DHowett-MSFT:
That's crazy behaviour! However, it does appear to be valid. The docs say &quot;In a reference-counted environment, if the instance variable is accessed directly, value is retained.&quot;. Huh. Looks like we retain but never release the old value, as you observed.
If you don't mind making the change!

triplef:
Cool, I’ve made that change: new ivar value is retained, old ivar value is _not_ released.

PR #285, opened 2015-12-10
Fix link error with UINavigationItem.

ehren:
The relevant symbols needed to be exported.


Ramu-msft:
Thanks for your contribution

PR #287, opened 2015-12-17
Add timeZone property to NSDateComponents with related handling in NSCalendar dateFromComponents:

ehren:
Commit also adds unit test and fixes incorrect retain (should be copy) semantics with NSCalendar timeZone and locale properties.

### DIFF
--- a/Frameworks/Foundation/NSCalendar.mm
+++ b/Frameworks/Foundation/NSCalendar.mm
@@ -159,7 +159,7 @@
  @Status Interoperable
 */
 - (void)setTimeZone:(NSTimeZone*)timeZone {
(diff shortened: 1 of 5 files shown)


ehren:
Hmm, my "squash" seems to have removed your review comments. That should do it though (NSDateComponents header converted to properties and unnecessary copying of tz property avoided). Also added another unit test to ensure the NSDateComponents timeZone takes precedence over the NSCalendar timeZone.

ehren:
Latest updated/squashed commit removes the spurious ": NSObject" from the NSDateComponents.mm @implementation.

rajsesh-msft:
Thanks for your contributions!

PR #295, opened 2016-01-05
Fix mistake in quaternion multiplication

jminer:
I was referencing GLKMath (along with a couple of web sites) for some quaternion functions I was writing, and I noticed that `GLKQuaternionMultiply` was wrong. See https://en.wikipedia.org/wiki/Quaternion#Hamilton_product for the formula. There are two additions that should be multiplications.


Ramu-msft:
Thanks for your contribution.

PR #301, opened 2016-01-08
Implement NSObject cancelPreviousPerformRequestsWithTarget* class methods

ehren:
Commit also adds unit test and removes spurious cancelPreviousPerform* method declarations from NSRunLoop.h


DHowett-MSFT:
Hi @ehren! Thanks for your contribution.

We've identified a couple issues with this pull request.
* NSObject will leak a NSDelayedPerform on cancellation.
* The test code is very sensitive to timing. We have seen similar tests fail on machines under load. Would you be willing to rewrite that test in terms of std::condition_variable, NSCondtionLock, or some other completion signalling mechanism? I realize that reliably proving a negative ("method didn't get called") with a time constraint might be somewhat tricky!

Thanks!

PR #303, opened 2016-01-11
Submit additional tests for NSString* dataUsingEncoding

cloudmonkeypeng:
add additional tests for Issue #300

PR #315, opened 2016-01-18
Submit additional tests about NSString* dataUsingEncoding

cloudmonkeypeng:
Add tests for #303 

NSString* dataUsingEncoding method always get random tail string in the data.bytes.
So submit this test.


rajsesh-msft:
This is working by design, so this is not really a valid test.  dataUsingEncoding by design does not include the null termination character, which is needed if you are using %s to print the string.

PR #363, opened 2016-02-23
Fixed pre-defined NSValueTransformer constants.

triplef:
These match the reference implementation.

### DIFF
--- a/Frameworks/UnifiedFoundation/Foundation/NSValueTransformer.mm
+++ b/Frameworks/UnifiedFoundation/Foundation/NSValueTransformer.mm
@@ -17,11 +17,11 @@
 #import <StubReturn.h>
 #import <Foundation/NSValueTransformer.h>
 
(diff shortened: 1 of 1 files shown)


rajsesh-msft:
Is there a reason you are using the actual values for the constants? The change doesn't hurt, but want to make sure you are not relying on the actual values instead of the constants defined in the headers.

cloudmonkeypeng:
@rajsesh-msft    I have faced the same problem on the constants string  NSCalendarIdentifierGregorian / NSCalendarIndentifierChinese.  I'm not relying on the actual values. But in my opinion, when I want to output the constants string, it will be better if the actual values is same as the iOS.

rajsesh-msft:
@cloudmonkeypeng Thanks! if you have a pull  request, will be happy to take it. :)

triplef:
@rajsesh-msft thanks for the quick merge! We are using the actual values in XML documents for binding UI elements in our custom UI framework.

PR #364, opened 2016-02-24
make the const string NSCalendarIdentifier same as iOS

cloudmonkeypeng:
Change the constants values as the iOS.
ClangFormat by the Visual Studio plugin.


rajsesh-msft:
The change broke the build due to dependencies being reordered, I have backed this out, could you please resubmit?

PR #375, opened 2016-02-29
It's completely ok to give nil as errorCode pointer...

mgoemmel:
...so we should be prepared for this.

Thanks
Markus


mgoemmel:
WTF, sorry. Will resubmit right after closing this one. Sorry again.

PR #379, opened 2016-03-01
It's completely ok to give nil as errorCode pointer (2nd try)...

mgoemmel:
...so we should be prepared for this.


rajsesh-msft:
nitty comment to use {} with the if statements. 

if (outError) {
   *outError=nil;
}

mgoemmel:
> nitty comment to use {} with the if statements. 

Fixed.

rajsesh-msft:
Thanks again! One more step needed. You have to sign the CLA as mentioned here.  You have to do it only once for all MS open source projects.  

https://github.com/Microsoft/WinObjC/wiki/How-to-Contribute

rajsesh-msft:
Thanks for your contribution.

PR #423, opened 2016-04-06
Magnetometer APIs of CoreMotion framework, corresponding additions to CoreMotion WOCCatalog sample and timestamp property implementation for sensor readings.

afaruqui:


Frameworks/CoreMotion/CMMotionManager.mm
```diff
@@ -249,56 +321,140 @@ - (void)stopGyroUpdates {
...
-                                                 z:reading.angularVelocityZ * c_degreeToRadians];
+    _gyroData = [[CMGyroData alloc] initWithValues:reading.angularVelocityX * c_degreeToRadian 
+                                                 y:reading.angularVelocityY * c_degreeToRadian
+                                                 z:reading.angularVelocityZ * c_degreeToRadian
+                                              time:[self toSecondsSinceBoot:reading.timestamp]];
     return _gyroData;
  }
 
 
  -(void)setGyroUpdateInterval:(NSTimeInterval)updateInterval {
```
rajsesh-msft: The logic in this function could potentially be pathologically broken.  Ex: if mimimumReportInterval is 1010 ms, and updateInterval is 1sec, we will set _gyroUpdateInterval to 1sec and lateron try to update reportInterval to 1000ms, which is lower than 1010 ms.  In general, i think we should maintain the state in higher fidelity (in msec) and do the capping in the accessors.  Granted, this is not new code with this change.  Can we track that issue outside this PR?  Same issue in other methods that do similar math.

samples/WOCCatalog/WOCCatalog/CoreMotionViewController.mm
```diff
@@ -254,4 +267,100 @@ -(void)gyroStartButtonPressed:(UIButton*)button {
...
+    magnetoVal = [[UILabel alloc] initWithFrame:CGRectMake(0, magnetoHeight + 100, 350, 40)];
+    [magnetoVal setBackgroundColor:[UIColor whiteColor]];
+    [magnetoVal setText:@"X: 0.000        Y: 0.000        Z: 0.000"];
+    [magnetoVal setTextAlignment:NSTextAlignmentRight];
+    [scrollView addSubview:magnetoVal];
+}
+
+
+-(void)magnetoStopUpdates {
+    dispatch_queue_t backgroundQueue = dispatch_queue_create("stopMagneto", NULL);
```
rajsesh-msft: Why do we stop updates from background queue but start updates from main queue?  Also, if you really need to use background queue here, please use: dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_BACKGROUND, 0) instead of creating a new dispatch queue.

Frameworks/CoreMotion/CMMotionManager.mm
```diff
@@ -249,56 +321,140 @@ - (void)stopGyroUpdates {
...
+                [self.magnetometerQueue cancelAllOperations];
+                self.magnetometerQueue = nil;
+            }
+        }
+    }                         
+ }
+ 
+
+ // getter function for on-demand reading from sensor
+ -(CMMagnetometerData*)magnetometerData {
```
rajsesh-msft: @Interoperable?

Frameworks/CoreMotion/CMMotionManager.mm
```diff
@@ -249,56 +321,140 @@ - (void)stopGyroUpdates {
...
+     
+    WDSMagnetometerReading* reading = [self.magnetometer getCurrentReading];
```
rajsesh-msft: If we have code (like in the sample) where the calls are motionManager.magnetometerData.magneticField.x, motionManager.magnetometerData.magneticField.y,motionManager.magnetometerData.magneticField.z, we could end up with three different updates.  Is there value in caching this and the last update time and only call windows if the last update time is longer than the minimum update interval?

rajsesh-msft:
:-1:

Frameworks/CoreMotion/CMMotionManager.mm
```diff
@@ -249,56 +321,140 @@ - (void)stopGyroUpdates {
...
 - (void)stopMagnetometerUpdates {
-    UNIMPLEMENTED();
+    
+    @synchronized(self) {
+
+        if (self.magnetometerActive) {
+            self.magnetometerActive = false;
+        
+            // the reportInterval is set to 0, the default value.
+            self.magnetometer.reportInterval = 0;
```
bbowman: do we need to do this? I'm a little confused on how this will work since 0 obviously < min interval. I understand its a default but I know in other apis, there was some weirdness about not letting you set it back to that.

bbowman:
:+1:

Frameworks/CoreMotion/CMMotionManager.mm
jofre-ms: I believe caching already occurs internally for this property getter.

---
In reply to: [59075810](https://github.com/Microsoft/WinObjC/pull/423#discussion_r59075810) [](ancestors = 59075810)

jofre-ms:
:+1:

afaruqui:
Thanks for the feedback. Made changes and updated the PR.

Frameworks/CoreMotion/CMMotionManager.mm
jaredhms: yes, setting it to 0 is how you turn the sensor off.

rajsesh-msft:
:+1:

PR #445, opened 2016-04-15
ImageIO Destination and Properties

yweijiang:
-Added support for CGImageDestination
-Added support for Image Properties in both CGImageSource and CGImageDestination
-Added unit tests for new APIs


Frameworks/ImageIO/CGImageSource.mm
```diff
@@ -489,21 +1370,109 @@ size_t CGImageSourceGetCount(CGImageSourceRef isrc) {
...
+ @Notes A limited subset of source container properties are supported currently  
 */
 CFDictionaryRef CGImageSourceCopyProperties(CGImageSourceRef isrc, CFDictionaryRef options) {
-    UNIMPLEMENTED();
-    return StubReturn();
+    RETURN_NULL_IF(!isrc);
+    NSData* imageData = ((ImageSource*)isrc).data;
+    RETURN_NULL_IF(!imageData);
+
+    NSMutableDictionary *properties = [[NSMutableDictionary alloc] init];
```
jaredhms: nit: * on wrong side; did you clang-format your changes?

Frameworks/ImageIO/CGImageSource.mm
```diff
@@ -489,21 +1370,109 @@ size_t CGImageSourceGetCount(CGImageSourceRef isrc) {
...
+        PROPVARIANT propertyValue;
+        PropVariantInit(&propertyValue);
+        NSMutableDictionary *gifProperties = [[NSMutableDictionary alloc] init];
+
+        if (SUCCEEDED(imageMetadataReader->GetMetadataByName(L"/appext/Data", &propertyValue)) &&
+            propertyValue.vt == (VT_VECTOR | VT_UI1)) {
+            if (propertyValue.caub.pElems[0] == 3 && propertyValue.caub.pElems[1] == 1) {
+                NSNumber* loopCount = [NSNumber numberWithInt:propertyValue.caub.pElems[2] + (propertyValue.caub.pElems[3]<<8)];
+                [gifProperties setObject:loopCount forKey:(id)kCGImagePropertyGIFLoopCount];
+            }
```
jaredhms: could use comments; potentially pull 'magic' values out as constants too?

Frameworks/ImageIO/CGImageSource.mm
```diff
@@ -489,21 +1370,109 @@ size_t CGImageSourceGetCount(CGImageSourceRef isrc) {
...
-    UNIMPLEMENTED();
-    return StubReturn();
+    RETURN_NULL_IF(!isrc);
+    NSData* imageData = ((ImageSource*)isrc).data;
+    RETURN_NULL_IF(!imageData);
+
+    NSMutableDictionary *properties = [[NSMutableDictionary alloc] init]; 
+    [properties setObject:[NSNumber numberWithInt:[imageData length]] forKey:(id)kCGImagePropertyFileSize];
+
+    // Look for container-level metadata if the file is a GIF.
```
jaredhms: else? is it a caveat that we're only handling gifs here, or is that expected?

Frameworks/ImageIO/CGImageSource.mm
```diff
@@ -112,6 +115,884 @@ - (CFStringRef)getImageType {
 }
 @end
 
+CFDictionaryRef readJPEGProperties(IWICMetadataQueryReader* imageMetadataReader) {
+    PROPVARIANT propertyValue;
+    PropVariantInit(&propertyValue);
+    NSMutableDictionary *properties = [[NSMutableDictionary alloc] init];
```
jaredhms: nit

Frameworks/ImageIO/CGImageSource.mm
```diff
@@ -112,6 +115,884 @@ - (CFStringRef)getImageType {
...
+        propertyValue.vt == VT_LPSTR) {
+        [exifProperties setObject:[NSString stringWithUTF8String:propertyValue.pszVal] forKey:(id)kCGImagePropertyExifLensModel];
+    }
+
+    PropVariantClear(&propertyValue);
+    if (SUCCEEDED(imageMetadataReader->GetMetadataByName(L"/app1/ifd/exif/{ushort=37500}", &propertyValue)) && 
+        propertyValue.vt == VT_BLOB) {
+        [exifProperties setObject:[NSData dataWithBytesNoCopy:propertyValue.blob.pBlobData 
+                                                       length:propertyValue.blob.cbSize 
+                                                 freeWhenDone:YES]
```
jaredhms: Are you sure PropVariantClear doesn't free memory in the blob field, and that the memory was allocated with malloc?  It seems like we do need to make a copy here.

Frameworks/ImageIO/CGImageSource.mm
yweijiang: From the MSDN Documentation on the metadata query reader: 

If an image format does not support container-level metadata, this will return WINCODEC_ERR_UNSUPPORTEDOPERATION. The only Windows provided image format that supports container-level metadata is GIF. Instead, use IWICBitmapFrameDecode::GetMetadataQueryReader.

Only GIF can have metadata for the whole entire image source, which is what this API looks for. Individual frame properties are handled by the CGImageSourceCopyPropertiesAtIndex API.

Frameworks/ImageIO/CGImageSource.mm
yweijiang: Added a comment about what the GIF header contains for the looping properties, and made constants for the static bytes in front.

Frameworks/ImageIO/CGImageSource.mm
yweijiang: I did but it didn't detect anything. I may have set it up incorrectly so it wasn't diffing against the correct master and thought there were no changes.

Frameworks/ImageIO/CGImageSource.mm
jaredhms: thanks; can you please add a bit more info about that to this comment?

---
In reply to: [60331063](https://github.com/Microsoft/WinObjC/pull/445#discussion_r60331063) [](ancestors = 60331063)

Frameworks/ImageIO/CGImageSource.mm
yweijiang: You are correct, this actually does free the blob field. I changed it so it does make a copy, also rewrote the way the Exif version was being parsed to match what Apple does (make an array with say, {2, 2, 1} to represent version 2.2.1 instead of the raw characters "0221" which is what the blob bytes would contain).

Frameworks/ImageIO/CGImageSource.mm
```diff
@@ -112,6 +115,890 @@ - (CFStringRef)getImageType {
...
+    PropVariantClear(&propertyValue);
+    if (SUCCEEDED(imageMetadataReader->GetMetadataByName(L"/sRGB/RenderingIntent", &propertyValue)) && propertyValue.vt == VT_UI1) {
+        [pngProperties setObject:[NSNumber numberWithInt:propertyValue.bVal] forKey:(id)kCGImagePropertyPNGsRGBIntent];
+    }
+
+    PropVariantClear(&propertyValue);
+    if (SUCCEEDED(imageMetadataReader->GetMetadataByName(L"/chrominance/TableEntry", &propertyValue)) && 
+        propertyValue.vt == (VT_VECTOR | VT_UI1)) {
+        // [pngProperties setObject:[NSArray arrayWithObjects:(id)propertyValue.caub.pElems count:propertyValue.caub.cElems] 
+        //                   forKey:(id)kCGImagePropertyPNGChromaticities];
```
jaredhms: why is this commented out?

Frameworks/ImageIO/CGImageSource.mm
```diff
@@ -112,6 +115,890 @@ - (CFStringRef)getImageType {
...
+        [properties setObject:[NSNumber numberWithInt:propertyValue.uiVal] forKey:(id)kCGImagePropertyDepth];
+    }
+
+    PropVariantClear(&propertyValue);
+    if (SUCCEEDED(imageMetadataReader->GetMetadataByName(L"/app1/ifd/{ushort=274}", &propertyValue)) && propertyValue.vt == VT_UI2) {
+        [properties setObject:[NSNumber numberWithInt:propertyValue.uiVal] forKey:(id)kCGImagePropertyOrientation];
+    }
+
+    // JPEG Properties - Format-specific
+    NSMutableDictionary* jfifProperties = [[NSMutableDictionary alloc] init];
```
jaredhms: It looks like we're leaking these nested maps, since the retain count will be 1 here, and 2 when it's added to jpegProperties, but we never release our reference.  Can we enable ARC on these files?  If so, can you also #import "AssertARCEnabled.h" in each file so it's clear that ARC is enabled?

Frameworks/ImageIO/CGImageSource.mm
```diff
@@ -112,6 +115,890 @@ - (CFStringRef)getImageType {
...
+        [exifProperties setObject:exifVersionArray forKey:(id)kCGImagePropertyExifVersion];
+    }
+
+    PropVariantClear(&propertyValue);
+    if (SUCCEEDED(imageMetadataReader->GetMetadataByName(L"/ifd/exif/{ushort=41987}", &propertyValue)) && propertyValue.vt == VT_UI2) {
+        [exifProperties setObject:[NSNumber numberWithInt:propertyValue.uiVal] forKey:(id)kCGImagePropertyExifWhiteBalance];
+    }
+
+    // Add the Exif dictionary to the properties if there are any Exif properties
+    if ([[exifProperties allKeys] count] != 0) {
```
jaredhms: does this have to copy all of the keys into an NSArray just to get the count?  Can you just call [exifProperties count] directly?

jaredhms: same question applies in several spots...

---
In reply to: [60454530](https://github.com/Microsoft/WinObjC/pull/445#discussion_r60454530) [](ancestors = 60454530)

Frameworks/ImageIO/CGImageProperties.mm
```diff
@@ -17,298 +18,295 @@
...
-const CFStringRef kCGImagePropertyPNGChromaticities = static_cast<CFStringRef>(@"kCGImagePropertyPNGChromaticities");
-const CFStringRef kCGImagePropertyPNGAuthor = static_cast<CFStringRef>(@"kCGImagePropertyPNGAuthor");
-const CFStringRef kCGImagePropertyPNGCopyright = static_cast<CFStringRef>(@"kCGImagePropertyPNGCopyright");
-const CFStringRef kCGImagePropertyPNGCreationTime = static_cast<CFStringRef>(@"kCGImagePropertyPNGCreationTime");
-const CFStringRef kCGImagePropertyPNGDescription = static_cast<CFStringRef>(@"kCGImagePropertyPNGDescription");
-const CFStringRef kCGImagePropertyPNGModificationTime = static_cast<CFStringRef>(@"kCGImagePropertyPNGModificationTime");
-const CFStringRef kCGImagePropertyPNGSoftware = static_cast<CFStringRef>(@"kCGImagePropertyPNGSoftware");
-const CFStringRef kCGImagePropertyPNGTitle = static_cast<CFStringRef>(@"kCGImagePropertyPNGTitle");
-const CFStringRef kCGImagePropertyTIFFCompression = static_cast<CFStringRef>(@"kCGImagePropertyTIFFCompression");
+const CFStringRef kCGImagePropertyTIFFDictionary = static_cast<CFStringRef>(@"TIFF");
```
jaredhms: thanks for fixing all of these!

Frameworks/ImageIO/CGImageDestination.mm
```diff
@@ -15,11 +16,867 @@
...
+        MULTI_QI imageQueryInterface = {0};
+        static const GUID IID_IWICImagingFactory = {0xec5ec8a9,0xc395,0x4314,0x9c,0x77,0x54,0xd7,0xa9,0x35,0xff,0x70};
+        imageQueryInterface.pIID = &IID_IWICImagingFactory;
+        RETURN_NULL_IF_FAILED(
+            CoCreateInstanceFromApp(CLSID_WICImagingFactory, nullptr, CLSCTX_INPROC_SERVER, nullptr, 1, &imageQueryInterface));
+        
+        _idFactory = (IWICImagingFactory*)imageQueryInterface.pItf;
+        RETURN_NULL_IF_FAILED(_idFactory->CreateStream(&_idStream));
+
+        if (CFStringCompare(type, kUTTypeJPEG, NULL) == kCFCompareEqualTo) {
```
jaredhms: super nit: nullptr for all of the NULLs?

Frameworks/ImageIO/CGImageDestination.mm
```diff
@@ -15,11 +16,867 @@
...
+            }
+        } else if (data) {
+            _outData = data;
+
+            IStream* dataStream;
+            CreateStreamOnHGlobal(NULL, true, &dataStream);
+            RETURN_NULL_IF_FAILED(_idStream->InitializeFromIStream(dataStream));
+        } else {
+            // We are not currently handling data consumer as destination because data consumer is not implemented in WinObjC
+            UNIMPLEMENTED_WITH_MSG("Destination as Data Consumer is not handled right now");
```
jaredhms: ....or they passed in an invalid url path?  maybe we should log that error separately and return earlier?

Frameworks/ImageIO/CGImageDestination.mm
```diff
@@ -15,11 +16,867 @@
...
+            RETURN_NULL_IF_FAILED(_idStream->InitializeFromFilename(wideUrl.c_str(), GENERIC_WRITE));
+
+            if (![(NSURL*)url checkResourceIsReachableAndReturnError:NULL]) {
+                return NULL;
+            }
+        } else if (data) {
+            _outData = data;
+
+            IStream* dataStream;
+            CreateStreamOnHGlobal(NULL, true, &dataStream);
```
jaredhms: check for errors?

Frameworks/ImageIO/CGImageDestination.mm
```diff
@@ -15,11 +16,867 @@
...
+            std::wstring wideUrl((wchar_t*)[urlAsData bytes], [urlAsData length]/sizeof(wchar_t));
+            RETURN_NULL_IF_FAILED(_idStream->InitializeFromFilename(wideUrl.c_str(), GENERIC_WRITE));
+
+            if (![(NSURL*)url checkResourceIsReachableAndReturnError:NULL]) {
+                return NULL;
+            }
+        } else if (data) {
+            _outData = data;
+
+            IStream* dataStream;
```
jaredhms: shouldn't this be a ComPtr<IStream> so it's freed after transferring ownership to _idStream?

Frameworks/ImageIO/CGImageDestination.mm
```diff
@@ -15,11 +16,867 @@
...
+        [pathNSString release];
+    }
+}
+
+void setVariantFromDictionary(CFDictionaryRef dictionary,
+                              CFStringRef key,
+                              VARTYPE propertyType,
+                              LPCWSTR path,
+                              IWICMetadataQueryWriter* propertyWriter) {
+    if (dictionary && CFDictionaryContainsKey(dictionary, key)) {
```
jaredhms: will we ever not have a dictionary?

Frameworks/ImageIO/CGImageDestination.mm
```diff
@@ -15,11 +16,867 @@
...
+void setHighLowPartsSigned(LARGE_INTEGER* valueLarge, double valueDouble) {
+    // Check to see if the value has a decimal component. If not, just divide by 1.
+    if (valueDouble == (int)valueDouble) {
+        (*valueLarge).LowPart = (int)valueDouble;
+        (*valueLarge).HighPart = 1;
+    } else {
+        (*valueLarge).LowPart = (int)(valueDouble * 100);
+        (*valueLarge).HighPart = 100;
+    }
+}
```
jaredhms: nit we could condense these into a function template;

template <typename TOut>
void setHighLowParts(TOut& out, double value) {
    ...
}

Frameworks/ImageIO/CGImageDestination.mm
```diff
@@ -15,11 +16,867 @@
...
+            propertyToWrite.bVal = (unsigned char)[(id)CFDictionaryGetValue(dictionary, key) unsignedCharValue];
+        } else if (propertyType == VT_UI2) {
+            propertyToWrite.uiVal = (unsigned short)[(id)CFDictionaryGetValue(dictionary, key) unsignedShortValue];
+        } else if (propertyType == VT_UI4) {
+            propertyToWrite.ulVal = (unsigned long)[(id)CFDictionaryGetValue(dictionary, key) unsignedLongValue];      
+        } else if (propertyType == VT_UI8) {
+            double doubleOut = [(id)CFDictionaryGetValue(dictionary, key) doubleValue];
+            setHighLowPartsUnsigned(&propertyToWrite.uhVal, doubleOut);
+        } else if (propertyType == VT_I8) {
+            double doubleOut = [(id)CFDictionaryGetValue(dictionary, key) doubleValue];
```
jaredhms: nit: doubleValue?

Frameworks/ImageIO/CGImageDestination.mm
```diff
@@ -15,11 +16,867 @@
...
+            propertyToWrite.pszVal = (char*)[(NSString*)CFDictionaryGetValue(dictionary, key) UTF8String];
+        } else if (propertyType == VT_LPWSTR) {
+            NSString* wideStringBuffer = (NSString*)CFDictionaryGetValue(dictionary, key);
+            propertyToWrite.pwszVal = (wchar_t*)[wideStringBuffer cStringUsingEncoding:NSUnicodeStringEncoding];
+        } else if (propertyType == VT_BLOB) {
+            NSData* blobData = (NSData*)CFDictionaryGetValue(dictionary, key);
+            propertyToWrite.blob.cbSize = [blobData length];
+            propertyToWrite.blob.pBlobData = (unsigned char*)[blobData bytes];
+        } else {
+            NSTraceInfo(TAG, @"Encountered unknown value type when attempting to write property\n");
```
jaredhms: should we trace the property we're trying to write?

Frameworks/ImageIO/CGImageDestination.mm
```diff
@@ -15,11 +16,867 @@
...
+                             propertyWriter);
+}
+
+void writeJPEGProperties(IWICMetadataQueryWriter* propertyWriter, CFDictionaryRef properties, int imageWidth, int imageHeight) {
+    PROPVARIANT propertyToWrite;
+
+    // Manually writing constant values to properties when they don't come out of a dictionary
+    PropVariantInit(&propertyToWrite);
+    propertyToWrite.vt = VT_UI2;
+    propertyToWrite.uiVal = 8; // We are always using 8 bits per channel per pixel right now
```
jaredhms: where else are we hard-coding this value?

Frameworks/ImageIO/CGImageDestination.mm
```diff
@@ -28,75 +885,358 @@ CGImageDestinationRef CGImageDestinationCreateWithDataConsumer(CGDataConsumerRef
...
-    UNIMPLEMENTED();
-    return StubReturn();
-}
+    if (!idst) {
+        return false;
+    }
+
+    ImageDestination* imageDestination = (ImageDestination*)idst;
+
+    if (imageDestination.count != imageDestination.maxCount) {
```
jaredhms: should it be called 'expectedCount'? and is iOS this strict?

Frameworks/ImageIO/CGImageDestination.mm
```diff
@@ -28,75 +885,358 @@ CGImageDestinationRef CGImageDestinationCreateWithDataConsumer(CGDataConsumerRef
...
+        // Get stream stats in order to determine number of bytes that were written
+        STATSTG streamStats;
+        status = imageStream->Stat(&streamStats, STATFLAG_NONAME);
+        if (!SUCCEEDED(status)) {
+            NSTraceInfo(TAG, @"Fetch stream stats failed with status=%x\n", status);
+            return false;
+        }
+
+        // Copy stream into the mutable data
+        [dataNSPointer increaseLengthBy:streamStats.cbSize.QuadPart];
```
jaredhms: just want to confirm that this is only the size of the metadata that we written; we're not allocating more than we need to here, right?

Frameworks/ImageIO/CGImageDestination.mm
```diff
@@ -28,75 +885,358 @@ CGImageDestinationRef CGImageDestinationCreateWithDataConsumer(CGDataConsumerRef
...
+    }
+
+    ImageDestination* imageDestination = (ImageDestination*)idst;
+    if (imageDestination.count >= imageDestination.maxCount) {
+        NSTraceInfo(TAG, @"Max number of images in destination exceeded\n");
+        return;
+    }
+
+    ComPtr<IWICBitmapFrameEncode> imageBitmapFrame;
+    IPropertyBag2* propertyBag = NULL;
```
jaredhms: ComPtr, else we leak

Frameworks/ImageIO/CGImageDestination.mm
```diff
@@ -28,75 +885,358 @@ CGImageDestinationRef CGImageDestinationCreateWithDataConsumer(CGDataConsumerRef
...
+    }
+
+    // JPEG lossy compression property
+    // Note that while the scale goes from 0.0 to 1.0 on both Windows and iOS, WIC treats 0.0 as a much worse
+    // quality than iOS. In iOS, 0.0 looks relatively similar, while in Windows the image is heavily warped.
+    if (properties && CFDictionaryContainsKey(properties, kCGImageDestinationLossyCompressionQuality) &&
+            imageDestination.type == typeJPEG) {
+        PROPBAG2 option = { 0 };
+        std::wstring quality(L"ImageQuality");
+        option.pstrName = (wchar_t*)quality.c_str();
```
jaredhms: just assign directly to L"ImageQuality"?

Frameworks/ImageIO/CGImageDestination.mm
```diff
@@ -28,75 +885,358 @@ CGImageDestinationRef CGImageDestinationCreateWithDataConsumer(CGDataConsumerRef
...
+        VARIANT varValue;    
+        VariantInit(&varValue);
+        varValue.vt = VT_R4;
+        varValue.bVal = [(id)CFDictionaryGetValue(properties, kCGImageDestinationLossyCompressionQuality) doubleValue];
```
jaredhms: if you're setting VT_R4, shouldn't you be assigning fltVal?

Frameworks/ImageIO/CGImageDestination.mm
```diff
@@ -28,75 +885,358 @@ CGImageDestinationRef CGImageDestinationCreateWithDataConsumer(CGDataConsumerRef
...
+    if (!SUCCEEDED(status)) {
+        NSTraceInfo(TAG, @"Set Frame Size failed with status=%x\n", status);
+        return;
+    }
+
+    // TIFF on iOS will only set a resolution if both DPIWidth and DPIHeight are specified, with no default values
+    // This gets written as part of writing the frame metadata, so no else part here.
+    if (imageDestination.type != typeTIFF) {
+        // Using resolution of 72 dpi as standard for now, this seems to only affect metadata and not the image itself
+        status = imageBitmapFrame->SetResolution(72.0, 72.0);
```
jaredhms: why hard-code this?

Frameworks/ImageIO/CGImageDestination.mm
```diff
@@ -28,75 +885,358 @@ CGImageDestinationRef CGImageDestinationCreateWithDataConsumer(CGDataConsumerRef
...
+    }
+
+    status = imageBitmapFrame->SetPixelFormat(&formatGUID);
+    if (!SUCCEEDED(status)) {
+        NSTraceInfo(TAG, @"Set Pixel Format failed with status=%x\n", status);
+        return;
+    }
+
+    CGDataProviderRef provider = CGImageGetDataProvider(image);
+    NSData* imageByteData = (id)CGDataProviderCopyData(provider);
```
jaredhms: do we need to do an extra copy here?  won't imageFactory->CreateBitmapFromMemory also make a copy?

Frameworks/ImageIO/CGImageDestination.mm
```diff
@@ -28,75 +885,358 @@ CGImageDestinationRef CGImageDestinationCreateWithDataConsumer(CGDataConsumerRef
...
+                                                  imageHeight * imageWidth * 4,
+                                                  (unsigned char*)[imageByteData bytes],
+                                                  &inputImage);
+    [imageByteData release];
+    if (!SUCCEEDED(status)) {
+        NSTraceInfo(TAG, @"CreateBitmapFromMemory failed with status=%x\n", status);
+        return;
+    }
+
+    CGDataProviderRelease(provider);
```
jaredhms: leak if we return early above

Frameworks/ImageIO/CGImageSource.mm
yweijiang: We weren't handling Chrominance and I'm not sure where the path should actually be. I went ahead and removed Chrominance from source and destination.

Frameworks/ImageIO/CGImageSource.mm
yweijiang: I didn't have to, fixed this everywhere.

Frameworks/ImageIO/CGImageDestination.mm
yweijiang: I replaced all the NULLs with nullptr except for CGStringCompare, which doesn't compile with nullptr or nil. I'm not sure why, but I left a comment there saying that's why it's being left as NULL.

Frameworks/ImageIO/CGImageDestination.mm
yweijiang: Separated the check for url and whether the url is valid so we log if the url is invalid.

Frameworks/ImageIO/CGImageDestination.mm
yweijiang: If you mean using an HRESULT or something, CreateStreamOnHGlobal does not have an HRESULT and I don't believe it can actually fail. If not, could you elaborate on what I should check for?

Frameworks/ImageIO/CGImageDestination.mm
yweijiang: Fixed.

Frameworks/ImageIO/CGImageDestination.mm
yweijiang: Yes, there are places where I call this for a general image property without checking if properties are passed in to AddImage. This performs the role of both checking if there is a properties dictionary as well as checking if the specified key is in it.

Frameworks/ImageIO/CGImageDestination.mm
yweijiang: Changed the name.

Frameworks/ImageIO/CGImageDestination.mm
yweijiang: Fixed.

Frameworks/ImageIO/CGImageDestination.mm
yweijiang: Added a constant for this in 2 places.

Frameworks/ImageIO/CGImageDestination.mm
yweijiang: Yes, I changed the name to expectedCount. iOS does indeed return an error if you specified 3 frames on creating the destination and only added 2.

Frameworks/ImageIO/CGImageDestination.mm
yweijiang: Added a comment about how this works exactly as per our discussion.

Frameworks/ImageIO/CGImageDestination.mm
yweijiang: Fixed.

Frameworks/ImageIO/CGImageDestination.mm
jaredhms: Are you sure?  MSDN says it returns an HRESULT: https://msdn.microsoft.com/en-us/library/windows/desktop/aa378980(v=vs.85).aspx

Frameworks/ImageIO/CGImageDestination.mm
yweijiang: As per our discussion, I used that originally but C++ 11 didn't like it. It looks like (wchar_t*)L"ImageQuality" works though, so I changed it to that.

Frameworks/ImageIO/CGImageDestination.mm
yweijiang: Yes, this was a bug I actually missed because I was only testing quality = 0.0, which an unitialized bVal would produce anyways. Thanks for catching it.

Frameworks/ImageIO/CGImageDestination.mm
yweijiang: I went ahead and removed this, which leads to JPEGs giving X/Y resolutions as 0. Any image I produce on iOS has 72, and I can't seem to actually modify that number, but as we discussed letting WIC write 0 automatically probably makes more sense than hardcoding values.

Frameworks/ImageIO/CGImageDestination.mm
yweijiang: As far as I can tell, everyone online says that the way to get the bytes out of a CGImageRef object is to copy it to an array using CGDataProviderCopyData. I then have to call the imageFactory->CreateBitmapFromMemory to create a bitmap that WIC can use. I might be wrong but I'm not sure if this can be avoided since there doesn't seem to be a direct way to access the bytes in a CGImageRef.

Frameworks/ImageIO/CGImageDestination.mm
yweijiang: Fixed.

Frameworks/ImageIO/CGImageSource.mm
yweijiang: Can you clarify what you think the way would be to go about fixing this? As we discussed, we can't enable ARC on these files without breaking them right now, and I see the problem but I'm not sure what the best fix is.

Frameworks/ImageIO/CGImageDestination.mm
yweijiang: Went ahead and put this in a RETURN_NULL_IF_FAILED.

Frameworks/ImageIO/CGImageDestination.mm
yweijiang: Done.

Frameworks/ImageIO/CGImageDestination.mm
```diff
@@ -15,11 +16,847 @@
...
+                                 VT_UI4,
+                                 L"/gAMA/ImageGamma",
+                                 propertyWriter);
+
+        setVariantFromDictionary(pngDictionary,
+                                 kCGImagePropertyPNGsRGBIntent,
+                                 VT_UI1,
+                                 L"/sRGB/RenderingIntent",
+                                 propertyWriter);
+    }
```
jaredhms: Would it be worth at least keeping the 'unimplemented' tracing for kCGImagePropertyPNGChromaticities -- if (CFDictionaryContainsKey(pngDictionary, kCGImagePropertyPNGChromaticities))  ?

Frameworks/ImageIO/CGImageDestination.mm
```diff
@@ -28,75 +865,349 @@ CGImageDestinationRef CGImageDestinationCreateWithDataConsumer(CGDataConsumerRef
...
+        return;
+    }
+
+    // JPEG lossy compression property
+    // Note that while the scale goes from 0.0 to 1.0 on both Windows and iOS, WIC treats 0.0 as a much worse
+    // quality than iOS. In iOS, 0.0 looks relatively similar, while in Windows the image is heavily warped.
+    if (properties && CFDictionaryContainsKey(properties, kCGImageDestinationLossyCompressionQuality) &&
+            imageDestination.type == typeJPEG) {
+        PROPBAG2 option = { 0 };
+        option.pstrName = (wchar_t*)L"ImageQuality";
```
jaredhms: can you change this to const_cast<wchar_t*> so it's clear why you're casting here?

Frameworks/ImageIO/CGImageDestination.mm
```diff
@@ -28,75 +865,349 @@ CGImageDestinationRef CGImageDestinationCreateWithDataConsumer(CGDataConsumerRef
...
+        VariantInit(&varValue);
+        varValue.vt = VT_R4;
+        varValue.fltVal = [(id)CFDictionaryGetValue(properties, kCGImageDestinationLossyCompressionQuality) doubleValue];
+        if (varValue.fltVal >= 0.0 && varValue.fltVal <= 1.0) {
+            status = propertyBag->Write(1, &option, &varValue);
+            if (!SUCCEEDED(status)) {
+                NSTraceInfo(TAG, @"Property Bag Write failed with status=%x\n", status);
+                return;
+            }
+        }
```
jaredhms: else we just drop the value? should we log that it was out of range, or clamp instead?

Frameworks/ImageIO/CGImageDestination.mm
jaredhms: looks like you're right - thanks


---
In reply to: [60518362](https://github.com/Microsoft/WinObjC/pull/445#discussion_r60518362) [](ancestors = 60518362)

Frameworks/ImageIO/CGImageDestination.mm
jaredhms: thanks for the due diligence here

---
In reply to: [60517955](https://github.com/Microsoft/WinObjC/pull/445#discussion_r60517955) [](ancestors = 60517955)

Frameworks/ImageIO/CGImageSource.mm
jaredhms: Can you elaborate on why we can't use ARC in this file?  In the meantime, you can use StrongId<NSMutableDictionary> jfifProperties (IIRC from SmartTypes.h) in all places, which will handle calling release when the objects go out of scope.  Otherwise, you'd have to add 'release' calls everywhere you transfer ownership of the sub-property collections to the parent dictionary.

---
In reply to: [60630877](https://github.com/Microsoft/WinObjC/pull/445#discussion_r60630877) [](ancestors = 60630877)

yweijiang: Everything in the CGImageSourceInternal/CGImageDestinationInternal data structures seem to get deallocated after init with ARC enabled. Right now I just changed it to manually release the dictionaries after adding them to the parents.

Frameworks/ImageIO/CGImageDestination.mm
yweijiang: I'm not sure there's any reason to because there are other properties not being implemented and I don't think it's reasonable to have unimplemented tracing for every single property we don't use.

Frameworks/ImageIO/CGImageDestination.mm
yweijiang: Fixed.

Frameworks/ImageIO/CGImageDestination.mm
yweijiang: Apple seems to just drop the value, but I added a trace statement in an else here.

Frameworks/ImageIO/CGImageDestination.mm
```diff
@@ -15,11 +16,847 @@
...
+
+        if (propertyType == VT_UI1) {
+            propertyToWrite.bVal = (unsigned char)[(id)CFDictionaryGetValue(dictionary, key) unsignedCharValue];
+        } else if (propertyType == VT_UI2) {
+            propertyToWrite.uiVal = (unsigned short)[(id)CFDictionaryGetValue(dictionary, key) unsignedShortValue];
+        } else if (propertyType == VT_UI4) {
+            propertyToWrite.ulVal = (unsigned long)[(id)CFDictionaryGetValue(dictionary, key) unsignedLongValue];      
+        } else if (propertyType == VT_UI8) {
+            double dictionaryValue = [(id)CFDictionaryGetValue(dictionary, key) doubleValue];
+            setHighLowParts((ULARGE_INTEGER*)&propertyToWrite.uhVal, dictionaryValue);
```
jaredhms: nit: it doesn't seem like the cast here is necessary (nor below where you're casting to LARGE_INTEGER*).  If the cast is necessary, maybe  setHighLowParts<ULARGE_INTEGER>(&propertyToWrite.uhVal, dictionaryValue), etc. would be cleaner?

yweijiang: It's not necessary, and fixed.

Issue #302, opened 2016-01-08
Popover support


ehren:
The WOC sample app hints that full popover support is not ready.

A workable popover implementation would be highly useful for porting our primarily iPad targeted app to various Windows tablets. 

Just a comment: I see the conversion of iPad apps -> win tablet apps as more valuable than the conversion of e.g. primarily iPhone targeted apps to Windows apps (given some of the great Windows tablets now on offer).

ehren:
I'm going to look at https://github.com/werner77/WEPopover (and possibly also https://github.com/alvises/FPPopover) at some point - possibly by next week. If WEPopover converts reasonably well it might make the most sense to just use it internally for the popover implementation.

ehren:
basic implementation now in repo

PR #450, opened 2016-04-19
WOCOperationMode should change UIUserInterfaceIdiom. Support UIModalPresentationFormSheet on tablets (non-fullscreen presentation)

ehren:
Also adds examples to WOCCatalog app including resizing modal via
preferredContentSize and toggling tablet mode at runtime.

-----

First, sorry for going AWOL on the last patch!

A couple of things:
This patch doesn't address blocking touches on the parent's view (doesn't even set userInteractionEnabled = FALSE). I'm thinking that should be handled by a dummy full screen view underneath the modal - the same view could be used for the gray overlay applied to the parent's view when using a popover.

I require iPad style form sheet modals and popovers in my app. Are there any unreleased changes regarding popover support pending? I'm thinking popover support is not that bad e.g. not all that different from form sheet (I believe on iOS form sheet modals and popovers are presented in separate UIWindows which apparently doesn't need to be the case here). So the positioning logic for popovers would be similar (though ideally not embedded into UIViewController.mm also!). Popovers would also require the same "if not iPad use full screen" logic.

### DIFF
--- a/Frameworks/UIKit/UIViewController.mm
+++ b/Frameworks/UIKit/UIViewController.mm
@@ -344,14 +344,7 @@ UIInterfaceOrientation supportedOrientationForOrientation(UIViewController* cont
     appFrame = [[UIScreen mainScreen] applicationFrame];
 
     if (priv->_presentationStyle == UIModalPresentationFormSheet) {
-        CGRect screenFrame;
-        screenFrame = [[UIScreen mainScreen] applicationFrame];
-
-        appFrame.size.width = 540;
-        appFrame.size.height = 575;
-
-        appFrame.origin.x = screenFrame.origin.x + screenFrame.size.width / 2.0f - appFrame.size.width / 2.0f;
-        appFrame.origin.y = screenFrame.origin.y + screenFrame.size.height / 2.0f - appFrame.size.height / 2.0f;
+        appFrame = [self _modalPresentationFormSheetFrame];
     }
 
     CGRect rect;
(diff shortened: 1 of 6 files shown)


Ramu-msft:

Ramu-msft:

Frameworks/UIKit/UIViewController.mm
```diff
@@ -384,6 +377,35 @@ - (CGRect)_orientationRect:(UIInterfaceOrientation)orientation {
...
 }
 
+- (CGRect)_modalPresentationFormSheetFrame {
+    if (!GetCACompositor()->isTablet()) {
+        // fullscreen on non-tablets
+        return [[UIScreen mainScreen] applicationFrame];
+    }
+
+    CGRect frame;
+    if (!CGSizeEqualToSize(priv->_preferredContentSize, CGSizeZero)) {
```
davelamb: Should use accessors for these.

Frameworks/UIKit/UIViewController.mm
```diff
@@ -342,14 +342,7 @@ - (CGRect)_orientationRect:(UIInterfaceOrientation)orientation {
     appFrame = [[UIScreen mainScreen] applicationFrame];
 
     if (priv->_presentationStyle == UIModalPresentationFormSheet) {
```
davelamb: Not part of the change, but again should be using accessors for accessing properties.

bdrlamb:
:+1:

ehren:
I wasn't exhaustive with the accessors switch in the last commit - e.g. didn't touch priv->_view and some instances of private access to _parentViewController even in non init contexts seemed better to leave private. Let me know if that went overboard though (suppose there's a bit of overhead in the objective-c message dispatch added here).

I'm going to put further comments related to popovers in issue #302.

ehren:
No rush on these changes but perhaps I misunderstood the comment i.e. I should instead change UIViewControllerPriv to a class with public accessors?

In any case, I'm not sure if the public properties of e.g. UIViewController are invoked internally on iOS - you're definitely going to have a bad time with any non-trivial overriding of e.g. -(UIViewController*)parentViewController; but that's probably the case on iOS too.

davelamb:
Anything private can be accessed however you want (a struct is fine), I'm just worried about public-facing properties, since the accessors can be overridden. If we poke at the backing ivars outside the accessors, strange behaviors can pop up. This is just a general rule we haven't been too good about following, so I figured while you're in there... :)

Thanks!

---
In reply to: [212905371](https://github.com/Microsoft/WinObjC/pull/450#issuecomment-212905371) [](ancestors = 212905371)

davelamb:
:+1:

PR #462, opened 2016-04-28
Popovers

ehren:
UIPopoverController, UIPopoverPresentationController, and delegate classes implemented via internal use of WYPopoverController.

Commit also adds example use to WOCCatalog app and fixes presentViewController completion block called too soon (should happen after viewDidAppear:).

------------

Hello,

I've updated some of the bullet points below.

Notes

1) The commit uses https://github.com/nicolaschengdev/WYPopoverController
a) from trying out a few different popover implementations (WEPopover, FPPopover), think this is the best bet on Winobjc (without actually writing a UIPopoverController replacement from scratch)

b) ARC is enabled in WYPopoverController files for all builds (this is untested on ARM)

c) WYPopoverController itself has issues -
      ~~  - There are crashes requiring some drawing code to be disabled ~~. Fixed by 8cdbcab935eef41197620a9ccbd1509f38e8557b 
    
d) Most of the drawing code (e.g. popover arrow) doesn't actually work.

e) There is an apparently more up to date fork of WYPopoverController at https://github.com/sammcewan/WYPopoverController/ however this one doesn't work for some reason.

f) WYPopoverController is MIT licensed. I do need to update my public fork with the files contained in this pull 

g) I added a WYPopoverControllerPrivate.h header that adds underscores to the objective-c types via macros.

h) I included the WYStoryboardPopoverSegue class. It should be reasonable to add support for popover segues but this has not been done.

i) I had to end up fixing a number of bugs in WYPopoverController. One gotcha: this project had the habbit of calling methods on nil objects that return structs - which is now fixed in committed copy. I confirmed in the debugger this was causing problems (e.g. when accessing the height of a frame of a nil UIView). See e.g. https://llvm.org/bugs/show_bug.cgi?format=multiple&id=4863 ("If the method returns a struct, as defined by the Mac OS X ABI Function Call Guide to be returned in registers, then a message sent to nil returns 0.0 for every field in the data structure. Other struct data types will not be filled with zeros.")

2) Previous comments to this pull mentioned issues with "preferredContentSize". There is still an issue with the incorrect dimensions calculated when the popover's sourceView fills the entire window. However, the typical cases (presenting a popover from a particular subview e.g. a button or from a bar button item) work and the positioning logic seems fairly solid. I've modified the sample app to highlight these working cases. However, noticed today (May 10) that there is still an issue with the preferredContentSize height with release mode and optimizations enabled. I force pushed a change disabling optimization in WYPopoverController to address.


Ramu-msft:
@bdrlamb is added to the review. #Closed

Ramu-msft:

Ramu-msft:

ehren:
Note: still looking into this but a slight change in my app now triggers a stack overflow in WYPopoverController (in the method swizzling on load -- noticed the comments in #463). It appears the #ifdefing doesn't affect the ability to e.g. set the preferredContentSize - and I can even set it after load and see the animated size change.

So it might make more sense just to add more #ifdefs.

Also the original idea was to avoid hacking up WYPopoverController too badly - but if that's out then it might make sense to remove the added

```
    if (![self _popoverManagesPresentation]) {
        [self viewWillAppear:isAnimated];
    }
```
checks from UIViewController.mm (and remove the calling of these methods from WYPopoverController).

If I go the more #ifdef route, I'll submit here as a separate commit.

ehren:
Update:

I had to force push to remove the now merged modal changes from this pull. However, the original  	"UIPopoverController, UIPopoverPresentationController, and delegate classes implemented via internal use of WYPopoverController." commit is unchanged.

I have added an additional commit disabling the method swizzling.

However, I haven't been able to further investigate issue 1 d) listed above (inconsistent behaviour setting preferredContentSize). One note, this was developed against an earlier release (I believe "0.1 Preview April 14, 2016"). There were no issues with the preferredContentSize there. However these became apparent as I was testing with the latest changes at that time (0.1 Preview - April 21, 2016) - which required disabling optimizations to WYPopoverController in release builds.

Now I see with the WOCCatalog app that there is a similar issue with preferredContentSize in debug builds (against a Debug x86 build of the bridge) -- the width of the preferredContentSize is not being respected. At this point, I'm wondering if this issue 1 d) is due to miscompilation.

ehren:
I'll avoid any more push -f changes to the commit as of May 10 (will modify this comment if this changes).

build/UIKit/lib/UIKitLib.vcxproj
```diff
@@ -110,6 +110,20 @@
...
     <ClangCompile Include="..\..\..\Frameworks\UIKit\UIViewController.mm" />
     <ClangCompile Include="..\..\..\Frameworks\UIKit\UIWindow.mm" />
+    <ClangCompile Include="..\..\..\Frameworks\UIKit\WYPopoverController.m">
+      <ObjectiveCARC Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">true</ObjectiveCARC>
+      <ObjectiveCARC Condition="'$(Configuration)|$(Platform)'=='Debug|ARM'">true</ObjectiveCARC>
+      <ObjectiveCARC Condition="'$(Configuration)|$(Platform)'=='Release|ARM'">true</ObjectiveCARC>
+      <ObjectiveCARC Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">true</ObjectiveCARC>
+      <OptimizationLevel Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">Disabled</OptimizationLevel>
+      <OptimizationLevel Condition="'$(Configuration)|$(Platform)'=='Release|ARM'">Disabled</OptimizationLevel>
+    </ClangCompile>
```
oliversa-msft: Consolidate ddown to <ObjectiveCARC>true</ObjectiveCARC>

build/UIKit/lib/UIKitLib.vcxproj
```diff
@@ -110,6 +110,20 @@
...
+    <ClangCompile Include="..\..\..\Frameworks\UIKit\WYStoryboardPopoverSegue.m">
+      <ObjectiveCARC Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">true</ObjectiveCARC>
+      <ObjectiveCARC Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">true</ObjectiveCARC>
+      <ObjectiveCARC Condition="'$(Configuration)|$(Platform)'=='Debug|ARM'">true</ObjectiveCARC>
+      <ObjectiveCARC Condition="'$(Configuration)|$(Platform)'=='Release|ARM'">true</ObjectiveCARC>
+    </ClangCompile>
```
oliversa-msft: Ditto

build/UIKit/lib/UIKitLib.vcxproj
```diff
@@ -110,6 +110,20 @@
     <ClangCompile Include="..\..\..\Frameworks\UIKit\UIView.mm" />
     <ClangCompile Include="..\..\..\Frameworks\UIKit\UIViewController.mm" />
     <ClangCompile Include="..\..\..\Frameworks\UIKit\UIWindow.mm" />
+    <ClangCompile Include="..\..\..\Frameworks\UIKit\WYPopoverController.m">
+      <ObjectiveCARC Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">true</ObjectiveCARC>
+      <ObjectiveCARC Condition="'$(Configuration)|$(Platform)'=='Debug|ARM'">true</ObjectiveCARC>
+      <ObjectiveCARC Condition="'$(Configuration)|$(Platform)'=='Release|ARM'">true</ObjectiveCARC>
+      <ObjectiveCARC Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">true</ObjectiveCARC>
+      <OptimizationLevel Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">Disabled</OptimizationLevel>
+      <OptimizationLevel Condition="'$(Configuration)|$(Platform)'=='Release|ARM'">Disabled</OptimizationLevel>
```
oliversa-msft: Disabled optimization to address preferredContentSize issue? This is the part that I don't really like - not understanding the root cause. I'll let someone else be the tie breaker.

build/UIKit/lib/UIKitLib.vcxproj
```diff
@@ -110,6 +110,20 @@
     <ClangCompile Include="..\..\..\Frameworks\UIKit\UIView.mm" />
     <ClangCompile Include="..\..\..\Frameworks\UIKit\UIViewController.mm" />
     <ClangCompile Include="..\..\..\Frameworks\UIKit\UIWindow.mm" />
+    <ClangCompile Include="..\..\..\Frameworks\UIKit\WYPopoverController.m">
+      <ObjectiveCARC Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">true</ObjectiveCARC>
+      <ObjectiveCARC Condition="'$(Configuration)|$(Platform)'=='Debug|ARM'">true</ObjectiveCARC>
+      <ObjectiveCARC Condition="'$(Configuration)|$(Platform)'=='Release|ARM'">true</ObjectiveCARC>
+      <ObjectiveCARC Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">true</ObjectiveCARC>
```
oliversa-msft: Collapse down to <ObjectiveCARC>true</ObjectiveCARC>

build/UIKit/lib/UIKitLib.vcxproj
```diff
@@ -110,6 +110,20 @@
...
+      <OptimizationLevel Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">Disabled</OptimizationLevel>
+      <OptimizationLevel Condition="'$(Configuration)|$(Platform)'=='Release|ARM'">Disabled</OptimizationLevel>
+    </ClangCompile>
+    <ClangCompile Include="..\..\..\Frameworks\UIKit\WYStoryboardPopoverSegue.m">
+      <ObjectiveCARC Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">true</ObjectiveCARC>
+      <ObjectiveCARC Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">true</ObjectiveCARC>
+      <ObjectiveCARC Condition="'$(Configuration)|$(Platform)'=='Debug|ARM'">true</ObjectiveCARC>
+      <ObjectiveCARC Condition="'$(Configuration)|$(Platform)'=='Release|ARM'">true</ObjectiveCARC>
```
oliversa-msft: Ditto

Frameworks/include/UIPopoverControllerInternal.h
```diff
@@ -0,0 +1,29 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#ifndef _UIPOPOVERCONTROLLERINTERNAL_H_
+#define _UIPOPOVERCONTROLLERINTERNAL_H_
+
+@interface UIPopoverController () {
+@public
```
oliversa-msft: Normally we don't need this - is it necessary here?

Frameworks/include/UIPopoverPresentationControllerInternal.h
```diff
@@ -0,0 +1,32 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#ifndef _UIPOPOVERPRESENTATIONCONTROLLERINTERNAL_H_
+#define _UIPOPOVERPRESENTATIONCONTROLLERINTERNAL_H_
+
+@interface UIPopoverPresentationController () {
+@public
```
oliversa-msft: Same question as before - is this implicit already or necessary here

Frameworks/include/UIPopoverControllerInternal.h
```diff
@@ -0,0 +1,29 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#ifndef _UIPOPOVERCONTROLLERINTERNAL_H_
```
msft-Jeyaram: #pragma once should do it

Frameworks/UIKit/WYPopoverController.h
```diff
@@ -0,0 +1,259 @@
...
+ in all copies or substantial portions of the Software.
+ 
+ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, 
+ INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A 
+ PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT 
+ HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION 
+ OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE 
+ SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
+
```
msft-Jeyaram: need #pragma once

Frameworks/UIKit/WYPopoverPrivate.h
```diff
@@ -0,0 +1,41 @@
...
+ 
+ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, 
+ INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A 
+ PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT 
+ HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION 
+ OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE 
+ SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
+
+#ifndef WYPOPOVERPRIVATE_H
```
msft-Jeyaram: can be replaced by #pragma once

Frameworks/UIKit/UIPopoverController.mm
```diff
@@ -14,81 +14,269 @@
...
+
+@interface _WYPopoverDelegateInternal : NSObject <WYPopoverControllerDelegate>
+
+@property (nonatomic, assign) UIPopoverController* controller;
+@property (nonatomic, assign) id<UIPopoverControllerDelegate> delegate;
+
+@end
+
+@implementation _WYPopoverDelegateInternal {
+}
```
oliversa-msft: Remove

build/UIKit/lib/UIKitLib.vcxproj
ehren: @oliversa-msft It is to address the preferredContentSize issue. I'd have to do a clean framework build to confirm but I'm pretty sure you'll see an undefined (== fullscreen) height if you pref optimization back on.

In general WYPopoverController is pretty gross in that it swizzles all calls to preferredContentSize in order to modify it's internal state. A _from scratch_ popover implementation could instead handle this with some extra logic in UIViewController.mm

ehren: Also in debug builds most/all of the problems in WYPopoverController were from calling a method returning a struct on a nil object. In fact, if I add back the unconditional messaging to WYPopoverController, the issue should show up in debug builds as well. Totally speculative but I wonder if clang is removing the nil checks under some bad assumptions about the ABI (more likely some other issue though).

DHowett-MSFT: That's pretty interesting. I know the codegen on the reference platform changed to memset structs to 0 for deterministic behaviour, and I know that we don't exactly support that.

Frameworks/UIKit/UIPopoverController.mm
```diff
@@ -14,81 +14,269 @@
...
+*/
+- (void)setPassthroughViews:(NSArray*)passthroughViews {
+    return [_popoverControllerInternal setPassthroughViews:passthroughViews];
+}
+
+/**
+ @Status Interoperable
+*/
+- (UIPopoverArrowDirection)popoverArrowDirection {
     UNIMPLEMENTED();
```
jaredhms: remove if implemented

ehren: This isn't implemented - another WYPopoverController tweak would be necessary.

Frameworks/include/UIPopoverPresentationControllerInternal.h
ehren: Even the #pragma once isn't strictly necessary since we're #import ing these - but added them to all headers.

Frameworks/UIKit/UIPopoverController.mm
jaredhms: ah...then please update the annotation to @Status Stub

---
In reply to: [62900503](https://github.com/Microsoft/WinObjC/pull/462#discussion_r62900503) [](ancestors = 62900503)

ehren: squashed that change into the fixup commit

build/Headers/Headers.Shared/Internal.Shared.vcxitems
```diff
@@ -16,4 +16,9 @@
   <ItemGroup>
     <ProjectCapability Include="SourceItemsFromImports" />
   </ItemGroup>
+  <ItemGroup>
+    <ClInclude Include="$(MSBuildThisFileDirectory)..\..\..\Frameworks\UIKit\WYPopoverController.h" />
```
davelamb: These should be caught by the <Text ..> above, right?

Frameworks/UIKit/UIPopoverController.mm
```diff
@@ -14,81 +14,268 @@
...
+*/
+- (UIColor*)backgroundColor {
+    return _backgroundColor;
+}
+
+/**
+ @Status Interoperable
+*/
+- (void)setBackgroundColor:(UIColor*)color {
+    [_backgroundColor release];
```
davelamb: Just do the assignment, the operator overload on StrongId will release it at the right time (as it stands now, you might get an early release if you assign the same color).

build/Headers/Headers.Shared/Internal.Shared.vcxitems
ehren: my original commit to this pull actually undid someone's earlier change making _backgroundColor a StrongId. I've put that code back in the latest commit and also made better use of StrongId in UIPopoverController.mm.

Frameworks/UIKit/UIPopoverController.mm
ehren: my original commit to this pull actually undid someone's earlier change making _backgroundColor a StrongId. I've put that code back in the latest commit and also made better use of StrongId in UIPopoverController.mm.

build/Headers/Headers.Shared/Internal.Shared.vcxitems
ehren: I had the headers in the wrong directory. Now that I've moved them to Frameworks/include it was possible to remove those lines from the vcxitems.

include/UIKit/UIPopoverPresentationControllerDelegate.h
```diff
@@ -43,5 +43,5 @@
 - (void)popoverPresentationControllerDidDismissPopover:(UIPopoverPresentationController*)popoverPresentationController;
 - (void)popoverPresentationController:(UIPopoverPresentationController*)popoverPresentationController
           willRepositionPopoverToRect:(CGRect*)rect
-                               inView:(UIView*)view;
+                               inView:(UIView**)view;
```
davelamb: rect and view should be inout

include/UIKit/UIPopoverControllerDelegate.h
```diff
@@ -24,9 +24,11 @@
 #import <CoreGraphics/CGGeometry.h>
 
 @protocol UIPopoverControllerDelegate <NSObject>
+@optional
+
 - (void)popoverController:(UIPopoverController*)popoverController
     willRepositionPopoverToRect:(inout CGRect*)rect
-                         inView:(inout UIView* _Nonnull*)view;
```
davelamb: Is it not _Nonnull?

ehren: I think I'll have to make a few simple changes to the WYPopoverController headers to properly support these attributes throughout. I will change all of these in the next fixup commit.

include/UIKit/UIPopoverPresentationControllerDelegate.h
ehren: I've squashed  all changes adding the attributes into the last commit.

brianker:
:shipit:

Frameworks/UIKit/UIViewController.mm
```diff
@@ -1139,14 +1148,22 @@ - (void)dismissViewControllerAnimated:(BOOL)animated completion:(void (^)(void))
     [curController retain];
     [curController autorelease];
 
-    if (curController->priv->_modalViewController) {
+    bool popoverPresented = [curController _popoverManagesPresentation];
```
oliversa-msft: BOOL

Frameworks/UIKit/UIViewController.mm
```diff
@@ -1233,105 +1253,119 @@ - (void)_addToTop:(NSNumber*)animatedValue {
 
     priv->_isRootView = true;
 
+    bool displayPopover = false;
```
oliversa-msft: BOOL displayPopover  = NO;

oliversa-msft:
:shipit:

bdrlamb:
:shipit:

Frameworks/UIKit/UIViewController.mm
ehren: I've made the two s/bool/BOOL/ changes and squashed all review changes into a single review comments commit.

Frameworks/UIKit/UIViewController.mm
```diff
@@ -1233,105 +1253,119 @@ - (void)_addToTop:(NSNumber*)animatedValue {
...
+            displayPopover = ![self _hidesParent];
 
-        if (animated) {
-            g_presentingAnimated = TRUE;
-            [[UIApplication sharedApplication] beginIgnoringInteractionEvents];
-            [self _notifyViewWillAppear:TRUE];
-        } else if ([self _hidesParent]) {
-            [[[self parentViewController] view] setHidden:TRUE];
-        }
+            if (displayPopover) {
```
jaredhms: comments would be helpful here

Frameworks/UIKit/UIPopoverController.mm
```diff
@@ -14,81 +14,263 @@
...
-    return [[_backgroundColor copy] autorelease];
+    return [_popoverBackgroundViewClass autorelease];
 }
 
 /**
  @Status Stub
 */
-- (void)setBackgroundColor:(UIColor*)color {
+- (void)setPopoverBackgroundViewClass:(Class)popoverBackgroundViewClass {
     UNIMPLEMENTED();
```
jaredhms: is this implemented now?  does the annotation need to be changed to Interoperable?  same question in a few spots...

jaredhms: if interoperable, the UNIMPLEMENTED macros will also need to be removed.  my comments today won't hold up the PR though - we're running it through our CI builds right now and should be able to accept this today.  we might want to go back and fix these annotations up afterwards though.

---
In reply to: [65732473](https://github.com/Microsoft/WinObjC/pull/462#discussion_r65732473) [](ancestors = 65732473)

ehren: This is not implemented. I was concerned about memory leaks from the added popoverBackgroundViewClass retain property being autosynthesized (note the existing handling of the stub backgroundColor).

PR #464, opened 2016-04-28
UINavigationBar should retain _curItem and _newItem UINavigationItem instances (fix crash on UINavigationBar dealloc)

ehren:


Ramu-msft:
@bdrlamb is added to the review. #Closed

yiyang-msft:
:+1:

DHowett-MSFT:
>     _newItem = nil;

since you switched to idretain, we can remove this for cur/new

---
Refers to: Frameworks/UIKit/UINavigationBar.mm:577 in 0902644. [](commit_id = 0902644a6f40cdab45d87df39701e4db06726a60, deletion_comment = False)

DHowett-MSFT:
:+1:

ehren:
should the others like _leftButton go too?

msft-Jeyaram:
Yep, all of the variables wrapped in idretain do not need to be set to nil in dealloc.

ehren:
added removal of the rest of them (decided not to add unsetting of the delegate though)

PR #472, opened 2016-05-03
Implementation of NSOrderedSet.

### LINKED ISSUE #471: NSOrderedSet is missing.

RG-J:
Implementation of NSOrderedSet and unit tests.

Fix #471

### DIFF
--- a/Frameworks/Foundation/NSOrderedSet.mm
+++ b/Frameworks/Foundation/NSOrderedSet.mm
@@ -16,44 +16,41 @@
 
 #include "Starboard.h"
 #import <StubReturn.h>
+#include <vector>
 
-#include "Foundation/NSOrderedSet.h"
+#import <Foundation/NSOrderedSet.h>
+#import <Foundation/NSException.h>
 
-@implementation NSOrderedSet
+@implementation NSOrderedSet {
+    std::vector<StrongId<id>> setContainer;
+}
+
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 + (instancetype)orderedSet {
-    UNIMPLEMENTED();
-    return StubReturn();
+    return [[[NSOrderedSet alloc] init] autorelease];
 }
 
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 + (instancetype)orderedSetWithArray:(NSArray*)array {
-    UNIMPLEMENTED();
-    return StubReturn();
+    return [NSOrderedSet orderedSetWithArray:array range:NSMakeRange(0, [array count]) copyItems:NO];
 }
 
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 + (instancetype)orderedSetWithArray:(NSArray*)array range:(NSRange)range copyItems:(BOOL)flag {
-    UNIMPLEMENTED();
-    return StubReturn();
+    return [[[NSOrderedSet alloc] initWithArray:array range:range copyItems:flag] autorelease];
 }
 
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 + (instancetype)orderedSetWithObject:(id)object {
-    UNIMPLEMENTED();
-    return StubReturn();
+    return [[[NSOrderedSet alloc] initWithObject:object] autorelease];
 }
 
 /**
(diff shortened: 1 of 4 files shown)


rajsesh-msft:

rajsesh-msft:

rajsesh-msft:

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -66,84 +63,112 @@ + (instancetype)orderedSetWithObjects:(id)firstObj {
...
+    }
+
+    setContainer.push_back(object);
+}
+
+- (NSUInteger)count {
+    return setContainer.size();
+}
+
+// - (NSArray*) array {
```
DHowett-MSFT: What's happened here?

include/Foundation/NSOrderedSet.h
```diff
@@ -1,111 +1,108 @@
...
+//******************************************************************************
+#pragma once
+
+#import <Foundation/FoundationExport.h>
+#import <Foundation/NSFastEnumeration.h>
+#import <Foundation/NSObject.h>
+
+#import <Foundation/NSArray.h>
+#import <Foundation/NSKeyValueObserving.h>
+@class NSArray;
```
rajsesh-msft: Can you merge with master?  we want all the generics back :).

DHowett-MSFT:
> - (instancetype)initWithObjects:(id)firstObj {

it looks like our API tool consumed the ",..." here. Would you mind implementing this? It could be straightforward with ConvertVAListToVector (VAListHelper.h).

---
Refers to: Frameworks/Foundation/NSOrderedSet.mm:178 in ccaae2f. [](commit_id = ccaae2fac4db823343f5d195d685afe3c9b9e2c3, deletion_comment = False)

include/Foundation/NSOrderedSet.h
```diff
@@ -1,111 +1,108 @@
...
+@property (readonly, strong) NSArray* array STUB_PROPERTY;
+@property (readonly, strong) NSSet* set STUB_PROPERTY;
+@property (readonly, copy) NSString* description STUB_PROPERTY;
+- (BOOL)containsObject:(id)object;
+- (void)enumerateObjectsAtIndexes:(NSIndexSet*)indexSet
+                          options:(NSEnumerationOptions)opts
+                       usingBlock:(void (^)(id, NSUInteger, BOOL*))block STUB_METHOD;
+- (void)enumerateObjectsUsingBlock:(void (^)(id, NSUInteger, BOOL*))block STUB_METHOD;
+- (void)enumerateObjectsWithOptions:(NSEnumerationOptions)opts usingBlock:(void (^)(id, NSUInteger, BOOL*))block STUB_METHOD;
+- (id)objectAtIndex:(NSUInteger)index;
```
rajsesh-msft: Same comment as above, all the id stuff should now be ObjectType.

DHowett-MSFT:
> - (instancetype)initWithObjects:(id)firstObj {

that is to say: ConvertVAListToVector<T> returns a std::vector<T>!

---
In reply to: [216629907](https://github.com/Microsoft/WinObjC/pull/472#issuecomment-216629907) [](ancestors = 216629907)

---
Refers to: Frameworks/Foundation/NSOrderedSet.mm:178 in ccaae2f. [](commit_id = ccaae2fac4db823343f5d195d685afe3c9b9e2c3, deletion_comment = False)

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -254,30 +281,50 @@ - (void)enumerateObjectsWithOptions:(NSEnumerationOptions)opts usingBlock:(void
...
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 - (id)objectAtIndex:(NSUInteger)index {
-    UNIMPLEMENTED();
-    return StubReturn();
+    if (index >= [self count]) {
+        [NSException raise:NSRangeException format:@"index is greater than the OrsdereSet size %d", [self count]];
```
DHowett-MSFT: nit: this got typo'd

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -254,30 +281,50 @@ - (void)enumerateObjectsWithOptions:(NSEnumerationOptions)opts usingBlock:(void
...
+    }
+    return *std::next(setContainer.begin(), index);
```
DHowett-MSFT: what're the advantages to doing this versus setContainer[index]? operator[] for a vector should be slightly lighter-weight than advancing the begin iterator

include/Foundation/NSOrderedSet.h
RG-J: Yep, sorry just reverting that change.

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -254,30 +281,50 @@ - (void)enumerateObjectsWithOptions:(NSEnumerationOptions)opts usingBlock:(void
...
+        [NSException raise:NSInvalidArgumentException format:@"indexs should not be nil."];
+    }
+
+    NSMutableArray* ret = [NSMutableArray array];
+    [indexes enumerateIndexesUsingBlock:^(NSUInteger idx, BOOL* stop) {
+        [ret addObject:[self objectAtIndex:idx]];
+    }];
+
+    return [ret copy];
 }
```
DHowett-MSFT: This should be autoreleased

Frameworks/Foundation/NSOrderedSet.mm
RG-J: did you mean the +// - (NSArray*) array { implementation?

DHowett-MSFT: indeed :)

---
In reply to: [61936592](https://github.com/Microsoft/WinObjC/pull/472#discussion_r61936592) [](ancestors = 61936592)

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -16,44 +16,41 @@
...
 #include "Starboard.h"
 #import <StubReturn.h>
+#include <vector>
 
-#include "Foundation/NSOrderedSet.h"
+#import <Foundation/NSOrderedSet.h>
+#import <Foundation/NSException.h>
+
+@implementation NSOrderedSet {
+    std::vector<StrongId<id>> setContainer;
```
rajsesh-msft: Probably not the most efficient for set operations - many operations will now be O(N) (ex: contains) instead of O(1).

DHowett-MSFT: I came here to say that; the ideal case for an ordered set might be:

set<StrongId> set
unordered_map<id, set::iterator>

with a std::hash specialization for id that uses -[NSObject hash]

---
In reply to: [61937213](https://github.com/Microsoft/WinObjC/pull/472#discussion_r61937213) [](ancestors = 61937213)

tests/unittests/Foundation/NSOrderedSetTests.mm
```diff
@@ -0,0 +1,85 @@
...
+    id array[3] = { @2, @1, @1 };
+    NSOrderedSet* set = [NSOrderedSet orderedSetWithObjects:array count:3];
+
+    NSOrderedSet* orderedSet = [NSOrderedSet orderedSetWithOrderedSet:set range:NSMakeRange(0, [set count]) copyItems:NO];
+    ASSERT_EQ(2, [orderedSet count]);
+    ASSERT_OBJCEQ(@2, [orderedSet firstObject]);
+    ASSERT_OBJCEQ(@1, [orderedSet lastObject]);
+}
+
+TEST(NSOrderedSet, orderedSetWithSet) {
```
rajsesh-msft: thanks for writing the unit tests.  Can we add one or two that test that the set is indeed ordered? you are checking first and last, can we also add some that test the entire range?

Frameworks/Foundation/NSOrderedSet.mm
bbowman: Yes. I believe that we should be using both an array and a map here. The array is used for ordered operations while the map is used for set operations.

---
In reply to: [61937416](https://github.com/Microsoft/WinObjC/pull/472#discussion_r61937416) [](ancestors = 61937416,61937213)

bbowman: Note that "normal" ordered maps etc won't work for the ordered portion as it is not an ordering defined by hash value, or item comparison. IIRC, NSOrderedSet provides the ability to insert an item at a desired index.

---
In reply to: [61937816](https://github.com/Microsoft/WinObjC/pull/472#discussion_r61937816) [](ancestors = 61937816,61937416,61937213)

RG-J: I can switch the backing storage to a dictionary.
but then only the value gets retained in a dictionary. thus the contains would still be O(N).
or I can use an map have the key wrapped  in StrongId, but still then it would be (obj,1), so objectAtIndex would not be O(1)

DHowett-MSFT: That's why the combination vector+map works so well; for ordered and indexed operations, the vector comes into play. The map handles uniqueness and object-based lookup.

RG-J: usage of set would lead to the order being different. when it removes duplicates, the last encountered duplicate ends up coming at the start of the iterator. so the order is not maintained.

DHowett-MSFT: apologies; i meant `vector<StrongId>` and `map<id, vector<StrongId>::iterator>`

tests/unittests/Foundation/NSOrderedSetTests.mm
RG-J: sure.

Frameworks/Foundation/NSOrderedSet.mm
RG-J: yep indeed. operator[] would be might lighter.

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -66,84 +63,112 @@ + (instancetype)orderedSetWithObjects:(id)firstObj {
...
 
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 - (instancetype)initWithArray:(NSArray*)array range:(NSRange)range copyItems:(BOOL)flag {
-    UNIMPLEMENTED();
-    return StubReturn();
+    if (range.length > [array count]) {
```
jofre-ms: This check only works as intended when range.location == 0. Change to NSMaxRange(range)

RG-J: ops,  yep this is the offset from the location.

Frameworks/Foundation/NSOrderedSet.mm
RG-J: okay, if I use the map<id,vector<StrongId<id>>::Iterator>
the key is the object and the value is an iterator to the object?
Wouldn't it be simpler to just do this, map<id,NSUInteger> and during the insert call.

//Check if it's contained in the map
vector.push_back(obj);
map[obj] = (vector.size() -1);

this way we are keeping track of the order/index of the object in the map via value.

rajsesh-msft: It is more complicated than that. If you are doing NSMutableOrderedSet (which we should also do), they could share the implementation the iterator becomes important as items can be erased. On that note, you should also consider doing the Mutable version.  Yet another design change would be to do class clustering on this so that the backing store is in an implementation class and not directly on NSOrderedSet.  @bbowman is writing this up with issue #471 as a design. 

---
In reply to: [61966453](https://github.com/Microsoft/WinObjC/pull/472#discussion_r61966453) [](ancestors = 61966453)

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -156,75 +181,77 @@ - (instancetype)initWithObjects:(id)firstObj {
...
 
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 - (BOOL)containsObject:(id)object {
-    UNIMPLEMENTED();
-    return StubReturn();
+    return std::find(setContainer.begin(), setContainer.end(), object) != setContainer.end();
```
bviglietta: This is just going to compare pointers, isn't it? Not Objective-C object equality?

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -156,75 +181,77 @@ - (instancetype)initWithObjects:(id)firstObj {
...
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 - (instancetype)initWithOrderedSet:(NSOrderedSet*)orderedSet range:(NSRange)range copyItems:(BOOL)flag {
-    UNIMPLEMENTED();
-    return StubReturn();
+    auto values = orderedSet->setContainer;
+
+    NSMutableArray* arrayObjs = [NSMutableArray array];
```
bviglietta: It shouldn't be necessary to duplicate the set into an array.

Maybe initWithSet, initWithOrderedSet, and initWithArray should all delegate to a common initializer that takes NSEnumerator.

rajsesh-msft:
Dupe of #472.

PR #482, opened 2016-05-05
Implementation of NSOrderedSet.

### LINKED ISSUE #471: NSOrderedSet is missing.

RG-J:
Implementation of NSOrderedSet and unit tests.
Moved backing storage to internal header to support NSMutableOrderedSet.
Added unit tests to NSMutableOrderedSet.

Fix #471


rajsesh-msft:

rajsesh-msft:

rajsesh-msft:
@RG-J will review this today.  For future reference, please fork WinOBJC repo and push your changes there.  That way when you make updates, the original pull request will be updated instead of a new one being created.  This helps keep all the discussions in one place.

This discussion on [How to update pull request](http://stackoverflow.com/questions/7947322/preferred-github-workflow-for-updating-a-pull-request-after-code-review
) should help.

rajsesh-msft:
@benvi is added to the review. #Closed

rajsesh-msft:

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -14,311 +14,403 @@
...
 
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 - (BOOL)containsObject:(id)object {
-    UNIMPLEMENTED();
-    return StubReturn();
+    return mapContainer.find(object) != mapContainer.end();
```
bviglietta: This is just going to compare pointers, isn't it? Not Objective-C object equality?

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -14,311 +14,403 @@
...
+ @Status Interoperable
 */
 - (instancetype)initWithSet:(NSSet*)set copyItems:(BOOL)flag {
-    UNIMPLEMENTED();
-    return StubReturn();
+    if (set == nil) {
+        return [self init];
+    }
+
+    NSMutableArray* arrayObjs = [NSMutableArray array];
```
bviglietta: It shouldn't be necessary to duplicate the set into an array.

Maybe initWithSet, initWithOrderedSet, and initWithArray should all delegate to a common initializer that takes NSEnumerator.

Frameworks/Foundation/NSMutableOrderedSet.mm
```diff
@@ -74,7 +74,8 @@ - (void)addObjectsFromArray:(NSArray*)array {
  @Notes
 */
 - (void)insertObject:(id)object atIndex:(NSUInteger)idx {
-    UNIMPLEMENTED();
+    std::vector<StrongId<id>>::iterator newItem = setContainer.insert(setContainer.begin() + idx, object);
+    mapContainer[object] = newItem;
```
bviglietta: This will invalidate the iterators in mapContainer.

bviglietta:
>  @Status Stub

Interoperable

---
Refers to: Frameworks/Foundation/NSMutableOrderedSet.mm:41 in d365932. [](commit_id = d3659325425a354010ddbc9b4424d62b1cdd2ede, deletion_comment = False)

bviglietta:
> - (void)insertObject:(ObjectType)object atIndex:(NSUInteger)idx STUB_METHOD;

Delete if the method is implemented.

---
Refers to: include/Foundation/NSMutableOrderedSet.h:37 in d365932. [](commit_id = d3659325425a354010ddbc9b4424d62b1cdd2ede, deletion_comment = False)

Frameworks/include/NSOrderedSetInternal.h
```diff
@@ -0,0 +1,28 @@
...
+
+#import <Starboard.h>
+#import <Foundation/NSOrderedSet.h>
+#import <map>
+#import <vector>
+
+@interface NSOrderedSet () {
+@protected
+    std::vector<StrongId<id>> setContainer;
+    std::map<id, std::vector<StrongId<id>>::iterator> mapContainer;
```
ms-jihua: CC: private member vars usually have an underscore eg: _setContainer, _mapContainer

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -14,311 +14,403 @@
...
 
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
-- (instancetype)initWithObjects:(id)firstObj {
-    UNIMPLEMENTED();
-    return StubReturn();
+- (NSArray*)array {
```
bviglietta: You can simplify this some:

return [NSArray arrayWithObjects:setContainer.data() count:setContainer.size()];

And then you don't have to copy the array at the end.

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -14,311 +14,403 @@
...
+
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 + (instancetype)orderedSet {
-    UNIMPLEMENTED();
-    return StubReturn();
+    return [[[NSOrderedSet alloc] init] autorelease];
```
ms-jihua: the + convenience constructors should call [self alloc] instead of [NSOrderedSet alloc], otherwise this code won't correctly alloc for NSMutableOrderedSet or other subclasses

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -14,311 +14,403 @@
...
 
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 + (instancetype)orderedSetWithArray:(NSArray*)array {
-    UNIMPLEMENTED();
-    return StubReturn();
+    return [NSOrderedSet orderedSetWithArray:array range:NSMakeRange(0, [array count]) copyItems:NO];
```
DHowett-MSFT: Your factory methods should use `self` instead of `NSOrderedSet`. Otherwise, `[NSMutableOrderedSet orderedSet]` and friends will return *immutable* ones.

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -14,311 +14,403 @@
...
 
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
-- (instancetype)initWithObjects:(id _Nonnull const[])objects count:(NSUInteger)cnt {
-    UNIMPLEMENTED();
-    return StubReturn();
+- (NSSet*)set {
```
bviglietta: Similarly, you can use NSSet setWithObjects:count: here.

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -14,311 +14,403 @@
...
++ (instancetype)orderedSetWithObjects:(id)firstObj, ... {
+    va_list argList;
+    va_start(argList, firstObj);
+    std::vector<id> flatArgs = ConvertVAListToVector((id)firstObj, argList);
+    va_end(argList);
+
+    if (flatArgs.size() == 0) {
+        // return empty set if args are nil.
+        return [[NSOrderedSet new] autorelease];
+    }
```
DHowett-MSFT: this could just be return [self orderedSet]

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -526,30 +661,24 @@ - (NSString*)descriptionWithLocale:(id)locale indent:(NSUInteger)level {
...
 
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 - (NSUInteger)countByEnumeratingWithState:(NSFastEnumerationState*)state objects:(id _Nonnull[])stackbuf count:(NSUInteger)len {
-    UNIMPLEMENTED();
-    return StubReturn();
+    return [[self array] countByEnumeratingWithState:state objects:stackbuf count:len];
```
bviglietta: Is it legal to call this on a different NSArray object every time we're invoked? Can the state parameter be migrated from one object to another? I don't know the answer.

Frameworks/Foundation/NSMutableOrderedSet.mm
```diff
@@ -14,8 +14,9 @@
 //
 //******************************************************************************
 
-#import <StubReturn.h>
 #import <Foundation/NSMutableOrderedSet.h>
+#import <NSOrderedSetInternal.h>
+#import <StubReturn.h>
 
 @implementation NSMutableOrderedSet
```
bviglietta: You must override copyWithZone, even if it's just to stub it out.

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -427,48 +526,84 @@ - (void)removeObserver:(NSObject*)observer forKeyPath:(NSString*)keyPath context
...
+ @Status Interoperable
 */
 - (BOOL)isSubsetOfOrderedSet:(NSOrderedSet*)other {
-    UNIMPLEMENTED();
-    return StubReturn();
+    if (other == nil) {
+        return NO;
+    }
+
+    for (id item in other) {
```
bviglietta: The logic is backwards here: you want to check if self is a subset of other.

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -427,48 +526,84 @@ - (void)removeObserver:(NSObject*)observer forKeyPath:(NSString*)keyPath context
...
+ @Status Interoperable
 */
 - (BOOL)isSubsetOfSet:(NSSet*)set {
-    UNIMPLEMENTED();
-    return StubReturn();
+    if (set == nil) {
+        return NO;
+    }
+
+    for (id item in set) {
```
bviglietta: Backwards logic here too.

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -360,29 +450,38 @@ - (NSIndexSet*)indexesOfObjectsWithOptions:(NSEnumerationOptions)opts passingTes
...
+
+    if (NSMaxRange(range) > [self count]) {
+        [NSException raise:NSInvalidArgumentException format:@"The given range length is greater than the array length: %u.", [self count]];
+    }
+
+    NSUInteger end = NSMaxRange(range);
+    int index = 0;
+    for (NSUInteger i = range.location; i < end; ++i) {
+        objects[index] = [self objectAtIndex:i];
+        const char* s = [objects[index] UTF8String];
```
bviglietta: Looks like some debugging code slipped through.

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -14,311 +14,403 @@
...
- @Notes
+ @Status Interoperable
 */
 - (id)objectAtIndex:(NSUInteger)index {
-    UNIMPLEMENTED();
-    return StubReturn();
+    if (index >= [self count]) {
+        [NSException raise:NSRangeException format:@"index is greater than the orderedSet size %u", [self count]];
+    }
+    return setContainer[index];
```
bviglietta: Should retain-autorelease this object.

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -14,311 +14,403 @@
...
+        return [self init];
+    }
+
+    auto values = orderedSet->setContainer;
+
+    NSMutableArray* arrayObjs = [NSMutableArray array];
+
+    for (id item : values) {
+        [arrayObjs addObject:item];
+    }
```
ms-jihua: can just call [orderedSet array]?

ms-jihua: or, better yet, take bviglietta's suggestion below to delegate to a common NSEnumerator initializer, I'm not fond of making this intermediate array either

---
In reply to: [62376601](https://github.com/Microsoft/WinObjC/pull/482#discussion_r62376601) [](ancestors = 62376601)

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -14,311 +14,403 @@
...
+ @Status Interoperable
 */
 - (NSArray*)objectsAtIndexes:(NSIndexSet*)indexes {
-    UNIMPLEMENTED();
-    return StubReturn();
+    if (indexes == nil) {
+        [NSException raise:NSInvalidArgumentException format:@"indexes can not be nil."];
+    }
+
+    NSMutableArray* ret = [NSMutableArray array];
```
ms-jihua: __block?

Frameworks/include/NSOrderedSetInternal.h
bbowman: ( CC stands for coding convention btw. We use that abbreviation internally but figured you may not know)

---
In reply to: [62362657](https://github.com/Microsoft/WinObjC/pull/482#discussion_r62362657) [](ancestors = 62362657)

bbowman: Is this really what we want? seems like it may be more hassle than its worth. As ben mentioned elsewhere, inserts etc will invalidate the iterators in the map which will be expensive to rebuild. I assume the reasoning here was to be more performant in a few operations that otherwise would require linear traversal but I'm not so sure that is needed? It appears that OrderedSet is O(n) for "array like" operations so I don't think we need to be super aggressive at optimizing at this point. Thoughts?

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -14,311 +14,403 @@
...
- @Notes
+ @Status Interoperable
 */
 - (NSUInteger)indexOfObject:(id)object {
-    UNIMPLEMENTED();
-    return StubReturn();
+    if (![self containsObject:object]) {
+        return NSNotFound;
+    }
+    return mapContainer[object] - setContainer.begin();
```
bbowman: std::distance may be a better choice as pointer math can be hard to get right / hard to understand to the casual observer.

Frameworks/include/NSOrderedSetInternal.h
bbowman: Sorry for commenting super late here but what was the rational for using stl containers vs building on top of NSArray + NSSet? It seems like the general call patterns here may match closer with the NS counterparts and we can worry about optimizing / using a good backing store like stl there instead of re-implementing that "adapter" code. Just thinking that may be a bit easier in light of ben's concerns on iterators, the equals comparisons, etc...

Frameworks/Foundation/NSOrderedSet.mm
RG-J: __block? could you elaborate

ms-jihua: never mind, it's actually not necessary here. see [https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/Blocks/Articles/bxVariables.html#//apple_ref/doc/uid/TP40007502-CH6-SW6](https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/Blocks/Articles/bxVariables.html#//apple_ref/doc/uid/TP40007502-CH6-SW6) if you're curious though

---
In reply to: [62386471](https://github.com/Microsoft/WinObjC/pull/482#discussion_r62386471) [](ancestors = 62386471)

Frameworks/include/NSOrderedSetInternal.h
RG-J: since most of my initial implementation is based around set/array. Switching to NSMutableArray&NSMutableSet would not be much of a change and it would resolve the issue revolving around enumerator.

RG-J: Follow up, ping

RG-J: could you address the comment; I've made regarding switching from vector
and map to mutable array and set.

R

bviglietta: A combination of an array (for indexing) and a set (for membership checks) seems fine to me. It's not optimal, because you have to keep two data structures with duplicated data, and make sure they stay in sync, but I don't object too much.

NSOrderedSet seems like an odd data structure to me. To get sub-linear membership checks, where the objects are not necessarily comparable, we're probably talking about a hash table. A hash table doesn't usually provide an ordering, but it could, if you modified it so that the entries in the hash table are all joined in a big linked list. That would give you fast membership checks and an ordered enumeration, but slow lookup of an element by index. I don't really know if that's what Apple had in mind.

---
In reply to: [62575598](https://github.com/Microsoft/WinObjC/pull/482#discussion_r62575598) [](ancestors = 62575598)

bbowman: It is an odd data structure but yes I think a combination of NSArray + NSSet is the way to go. Keeping them in sync is a bit of a challenge but I don't believe Apple makes any thread safe promises about ordered set so it shouldn't be a huge burden. I see what ben is saying about an ordering on a hash table but that isn't what we want here. Hash table orderings are usually used internally for faster operations which implies an ordering according to some well defined scheme. NSOrderedSet however allows for programmatic index manipulations (insertAtIndex etc) that act like an array. A hash table with a linked list would need to be a single linked list which removes the benefit of the hash table unless I'm misunderstanding Ben's idea. For simplicity I would go with NSArray and NSSet as you have suggested.

---
In reply to: [62720222](https://github.com/Microsoft/WinObjC/pull/482#discussion_r62720222) [](ancestors = 62720222,62575598)

rajsesh-msft:
@RG-J How is this coming along?

---
In reply to: [217496322](https://github.com/Microsoft/WinObjC/pull/482#issuecomment-217496322) [](ancestors = 217496322)

RG-J:
@rajsesh-msft I've pushed in my changes reflecting the discussion above.

Frameworks/Foundation/NSMutableOrderedSet.mm
```diff
@@ -277,4 +285,11 @@ - (void)unionSet:(NSSet*)other {
     UNIMPLEMENTED();
 }
 
+/**
+ @Status Interoperable
+*/
+- (id)copyWithZone:(NSZone*)zone {
+    return [[NSMutableOrderedSet alloc] initWithOrderedSet:self];
```
bviglietta: copyWithZone should probably return an immutable object.

Frameworks/include/NSOrderedSetInternal.h
```diff
@@ -0,0 +1,33 @@
...
+// THE SOFTWARE.
+//
+//******************************************************************************
+#pragma once
+
+#import <Foundation/NSOrderedSet.h>
+#import <Starboard.h>
+
+@interface NSOrderedSet () {
+@package
```
bviglietta: Why @package?

build/Foundation/dll/Foundation.def
```diff
@@ -6,6 +6,7 @@ LIBRARY Foundation
 
         ; Foundation ObjC classes
         _OBJC_CLASS_NSArray DATA
+        _OBJC_CLASS__NSProxyOrderedSetArray DATA
```
bviglietta: Does this need to be exported?

bviglietta:
>  @Status Stub

Interoperable

---
Refers to: Frameworks/Foundation/NSMutableOrderedSet.mm:42 in 6294a35. [](commit_id = 6294a3590c4a56db43a834cbd12cc86e1711f414, deletion_comment = False)

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -14,217 +14,297 @@
...
 */
 - (instancetype)initWithSet:(NSSet*)set copyItems:(BOOL)flag {
-    UNIMPLEMENTED();
-    return StubReturn();
+    if (set == nil) {
+        return [self init];
+    }
+
+    if (self = [super init]) {
+        _setContainer = [[NSMutableSet alloc] initWithSet:set copyItems:flag];
```
bviglietta: This retains the set twice, once because of alloc and once in StrongId's assignment operator. You should use _setContainer.attach(...) instead.

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -14,217 +14,297 @@
...
+*/
+- (NSUInteger)count {
+    return [_arrayContainer count];
+}
+
+/**
+ @Status Interoperable
+*/
+- (NSArray*)array {
+    return [[[_NSProxyOrderedSetArray alloc] initWithOrderedSet:self] autorelease];
```
DHowett-MSFT: Why can't we just return an arrayContainer copy here?

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -14,217 +14,297 @@
...
+}
+
+/**
+ @Status Interoperable
+*/
+- (NSSet*)set {
+    return [[_setContainer copy] autorelease];
```
bviglietta: This doesn't satisfy the contract for this method. Subsequent modifications to the NSOrderedSet are supposed to be reflected in the returned NSSet.

Frameworks/Foundation/NSSet.mm
```diff
@@ -279,15 +279,32 @@ - (instancetype)initWithArray:(NSArray*)array {
...
-        assert(i < count);
-        objects[i++] = curObj;
+    if(flag) {
+        for (id curObj in set) {
+            assert(i < count);
+            id obj = [curObj copy];
+            objects[i++] = obj;
+            [obj release];
+        }
+    } else {
```
DHowett-MSFT: The logic here could be folded a little tighter, removing the duplicate assert and add

Frameworks/include/NSOrderedSetInternal.h
DHowett-MSFT: Package is themost technically-correct, if this were a public header.

bviglietta: Oh, I see. Because the field is accessed in _NSProxyOrderedSetArray.

---
In reply to: [63977056](https://github.com/Microsoft/WinObjC/pull/482#discussion_r63977056) [](ancestors = 63977056)

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -14,217 +14,297 @@
...
+ @Status Interoperable
 */
 - (instancetype)initWithOrderedSet:(NSOrderedSet*)orderedSet range:(NSRange)range copyItems:(BOOL)flag {
-    UNIMPLEMENTED();
-    return StubReturn();
+    if (orderedSet == nil) {
+        return [self init];
+    }
+
+    return [self initWithArray:orderedSet->_arrayContainer range:range copyItems:flag];
```
DHowett-MSFT: Since you're accessing the ivars here, you could just set set/arrayContainer to mutable copies of the other ones :)

Frameworks/Foundation/NSOrderedSet.mm
DHowett-MSFT: Oh, I see why the proxy array approach was taken. 
Interesting: we need to effectively strip the mutability without losing the passthrough.
Do you think we can just return the internal mutable container and hope nobody decides to manipulate it?

Frameworks/Foundation/NSOrderedSet.mm
DHowett-MSFT: Sorry, see below

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -14,217 +14,297 @@
...
+#import <Starboard.h>
 #import <StubReturn.h>
+#import <VAListHelper.h>
 
-#include "Foundation/NSOrderedSet.h"
+// Needed to make sure that the returned array behaves like a one way proxy to an array
+@implementation _NSProxyOrderedSetArray
+- (instancetype)initWithOrderedSet:(NSOrderedSet*)orderedSet {
+    if (self = [super init]) {
+        _array = orderedSet->_arrayContainer;
```
bviglietta: You should retain this array, and release it in dealloc.

Frameworks/Foundation/NSSet.mm
```diff
@@ -279,15 +279,32 @@ - (instancetype)initWithArray:(NSArray*)array {
...
 
-    for (id curObj in set) {
-        assert(i < count);
-        objects[i++] = curObj;
+    if(flag) {
+        for (id curObj in set) {
+            assert(i < count);
+            id obj = [curObj copy];
+            objects[i++] = obj;
+            [obj release];
```
bviglietta: You can't release the object yet, or it might be deallocated before you insert it into the set.

build/Foundation/dll/Foundation.def
RG-J: yes, when I tried to compile the foundation dll, it started to complain about this class.

Frameworks/Foundation/NSOrderedSet.mm
RG-J: for the array, it's working correctly with the pass through with the mutability being striped.
I was going to do the same for the set, but if you are fine with it, I can just return the internal container?

Frameworks/Foundation/NSOrderedSet.mm
RG-J: I decided to use the initWithArray because user can specify a range and a flag. which might be a subset of the nsorderedset :)

Frameworks/Foundation/NSOrderedSet.mm
RG-J: The proxy array is a reflection of the NSOrdredSet (the internal array), so if the internal array is no longer alive; our retain will cause the _arrayContainer to out live the actual NSOrderedSet

Frameworks/Foundation/NSOrderedSet.mm
bviglietta: Probably better to leave the method as a stub.

If we do push this code, we must at least mark the method @Caveat with a note explaining how it differs from the reference implementation.

---
In reply to: [64046573](https://github.com/Microsoft/WinObjC/pull/482#discussion_r64046573) [](ancestors = 64046573)

Frameworks/Foundation/NSOrderedSet.mm
bviglietta: What you describe seems OK to me.

The problem I want to avoid is with code like this:

NSOrderedSet* oset = [NSOrderedSet new];
NSArray* array = [oset array];
[oset release];
[array count];  // Crash accessing deallocated object

---
In reply to: [64047960](https://github.com/Microsoft/WinObjC/pull/482#discussion_r64047960) [](ancestors = 64047960)

Frameworks/Foundation/NSOrderedSet.mm
DHowett: Ah, range! Good call.

build/Foundation/dll/Foundation.def
bviglietta: What was the error? I can't think of a reason why it would be necessary to export this class.

---
In reply to: [64045867](https://github.com/Microsoft/WinObjC/pull/482#discussion_r64045867) [](ancestors = 64045867)

Frameworks/Foundation/NSOrderedSet.mm
bbowman: I agree with Dustin. It seems like a lot of work to make this proxy object just for the mutability weirdness. Can we validate that the returned object on the reference platform isn't actually a mutable one?

---
In reply to: [63978319](https://github.com/Microsoft/WinObjC/pull/482#discussion_r63978319) [](ancestors = 63978319)

build/Foundation/dll/Foundation.def
bbowman: I agree with Ben. In general I've learned that the compiler / linker teams have lots of smart people that have been working for years on them. Doing things to simply get warnings / errors to go away is rarely the right call as it usually will reveal a bigger underlying issue in the code. Lets figure out what the error was and see if we can understand why it thinks it needs to be exported.

---
In reply to: [64072253](https://github.com/Microsoft/WinObjC/pull/482#discussion_r64072253) [](ancestors = 64072253,64045867)

Frameworks/Foundation/NSOrderedSet.mm
bbowman: please prefer the of StrongId

---
In reply to: [63978985](https://github.com/Microsoft/WinObjC/pull/482#discussion_r63978985) [](ancestors = 63978985)

bbowman: Prefer using StrongId where possible.

---
In reply to: [63978985](https://github.com/Microsoft/WinObjC/pull/482#discussion_r63978985) [](ancestors = 63978985)

Frameworks/Foundation/NSOrderedSet.mm
bviglietta: The reference platform actually does return an immutable container.

---
In reply to: [64078518](https://github.com/Microsoft/WinObjC/pull/482#discussion_r64078518) [](ancestors = 64078518,63978319)

RG-J:
updated, please take a look.

R

Frameworks/Foundation/NSMutableOrderedSet.mm
bbowman: I disagree here. copyWithZone should return an instance of NSMutableOrderedSet (since that is the class being implemented here) but it should be a new, copied instance ... 

---
In reply to: [63976700](https://github.com/Microsoft/WinObjC/pull/482#discussion_r63976700) [](ancestors = 63976700)

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -14,217 +14,298 @@
...
+    }
+
+    return self;
+}
+
+#pragma mark NSArray
```
bbowman: ?

tests/unittests/Foundation/NSMutableOrderedSetTests.mm
```diff
@@ -0,0 +1,378 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#import <Foundation/Foundation.h>
+#import <TestFramework.h>
+
+void assertOrderedSetContent(NSMutableOrderedSet* set, NSObject* first, ...) {
```
bbowman: Not a big deal but I would have preferred an NSArray parameter here for clarity.

tests/unittests/Foundation/NSOrderedSetTests.mm
```diff
@@ -0,0 +1,324 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#import <Foundation/Foundation.h>
+#import <TestFramework.h>
+
+void assertOrderedSetContent(NSOrderedSet* set, NSObject* first, ...) {
```
bbowman: this function needs to be static or better yet shared between files. As is it breaks the one definition rule as C style functions are extern by default meaning both Translations units will see the definitions and the linker will just choose one for you. That is "fine" as long as they stay the exact same but that begs the question why they just aren't sharing the implementation to begin with.

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -14,217 +14,298 @@
...
-    UNIMPLEMENTED();
-    return StubReturn();
+    if (array == nil) {
+        return [self init];
+    }
+
+    if (NSMaxRange(range) > [array count]) {
+        [NSException raise:NSInvalidArgumentException
+                    format:@"The given range length is greater than the array length: %u.", [array count]];
+    }
```
bbowman: We have some macros for this that we'd prefer that you use. THROW_NS_IF_FALSE and friends are useful for these sorts of arg checks. Note that E_INVALIDARG maps to NSInvalidArgumentException.

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -234,98 +315,104 @@ - (BOOL)containsObject:(id)object {
 - (void)enumerateObjectsAtIndexes:(NSIndexSet*)indexSet
                           options:(NSEnumerationOptions)opts
                        usingBlock:(void (^)(id, NSUInteger, BOOL*))block {
+    // Uncomment this when enumerateObjectsAtIndexes is implemented.
+    // if (block == nil) {
+    //  [NSException raise:NSInvalidArgumentException format:@"block can not be nil."];
+    //}
+    //[_arrayContainer enumerateObjectsAtIndexes:indexSet options:opts usingBlock:block];
```
bbowman: Please remove commented out code. In general commented out code just clutters the code base and makes things harder to read. Because commented out code by definition never compiles, it often is quickest to get stale and lacks the appropriate context for why the code isn't working, what needs done to enable it, etc.

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -234,98 +315,104 @@ - (BOOL)containsObject:(id)object {
...
+    return [_arrayContainer indexOfObjectPassingTest:predicate];
 }
 
 /**
  @Status Stub
  @Notes
 */
 - (NSUInteger)indexOfObjectWithOptions:(NSEnumerationOptions)opts passingTest:(BOOL (^)(id, NSUInteger, BOOL*))predicate {
+    // TODO: uncomment when NSArray has implemented indexOfObjectWithOptions
+    // return [_arrayContainer indexOfObjectWithOptions:opts passingTest:predicate];
```
bbowman: please remove commented out code.

bbowman: (and everywhere else in this file)

---
In reply to: [64478371](https://github.com/Microsoft/WinObjC/pull/482#discussion_r64478371) [](ancestors = 64478371)

bbowman:
:clock1:

Frameworks/Foundation/NSOrderedSet.mm
RG-J: not sure, what are you referring to?

bbowman: Sorry was unclear. Meant to only comment on one line. The "#pragma mark NSArray" doesn't make sense to me here.

tests/unittests/Foundation/NSOrderedSetTests.mm
RG-J: I have renamed the functions: assertOrderedSetContent and assertMutableOrderedSetContent

@ "begs the question why they just aren't sharing the implementation to begin with"
We shouldn't, because we don't want the tests to depend on another test, in case someone decides to add different logic in the future

Frameworks/Foundation/NSMutableOrderedSet.mm
DHowett-MSFT: No: as per API contract, -copy should always return an immutable container.

```
[mut copy] => immutable
[immut copy] => immutable
[mut mutableCopy] => mutable
[immut mutableCopy] => mutable
```

tests/unittests/Foundation/NSOrderedSetTests.mm
bbowman: The test wouldn't depend on another test. Both would depend on a common helper function.

In terms of code hygiene and maintainability, we want to factor out duplicated code where possible. Imagine not that someone wants to add different logic in the future (as this is always easy by adding a new function and calling it) but instead imagine someone changing the logic to an existing function. With two functions of duplicated logic, this makes two places the author needs to change and increases the opportunity for bugs / mismatches to arise. 

Renaming the functions is not sufficient as the extern / one definition rule issue still exists if someone else comes along and creates their own function called this. We should be very precise about how we expose functions where possible. Either a function is private to that TU and should be marked as static or it is (possibly) shared across files and it is not. Here you are claiming they are private but only changed the name. 

Please change this to have a common helper .h file (along with an implementation .mm or .cpp) that each test includes for this function.

Frameworks/Foundation/NSMutableOrderedSet.mm
bbowman: Oops. I'm a dummy and forgot about the mutableCopy guys. Carry on. 

---
In reply to: [64486017](https://github.com/Microsoft/WinObjC/pull/482#discussion_r64486017) [](ancestors = 64486017)

tests/unittests/Foundation/NSOrderedSetTests.mm
RG-J: where should I add this helper file under?

Frameworks/Foundation/NSOrderedSet.mm
RG-J: I added it for code organization, given the classes don't really use them anyway for grouping, i removed it.

RG-J:
comments are addressed, please take a look.

R

Frameworks/Foundation/NSOrderedSet.mm
```diff
@@ -14,217 +14,289 @@
...
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
-- (instancetype)initWithObjects:(id _Nonnull const[])objects count:(NSUInteger)cnt {
-    UNIMPLEMENTED();
-    return StubReturn();
+- (instancetype)initWithObjects:(id _Nonnull const[])objects count:(NSUInteger)count {
+    THROW_NS_IF_FALSE(E_INVALIDARG, objects != nil);
```
bviglietta: objects can be NULL if count is 0.

(Also prefer NULL or nullptr for non-ObjC pointers.)

bbowman: Yes. i believe the _Nonnull there is saying that items in the objects[] are _Nonnull not that the pointer itself is _Nonnull. But we'd want to double check that.

bviglietta:
:shipit:

RG-J:
NULL and count 0 addressed. please take a look.

DHowett-MSFT:
:shipit:

tests/unittests/Foundation/TestUtils.h
```diff
@@ -0,0 +1,20 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+#pragma once
+
+#import <Foundation/NSOrderedSet.h>
+
+void assertOrderedSetContent(NSOrderedSet* set, NSObject* first, ...);
```
bbowman: This file seems specific to OrderedSet tests. seems like it should be named something more specific as TestUtils sounds very generic and applicable to all/most tests.

rajsesh-msft:
@RG-J your changes have conflicts with the latest master.  Could you please rebase and resubmit the pull request?  Thanks for working with us on addressing the feedback.

RG-J:
@rajsesh-msft  I have re-based  and resubmitted the pull request.

rajsesh-msft:
Merged, thanks for your contributions @RG-J.

alakoring:
Is this ongoing? My app relies on NSMutableOrderedSet…

Warning		'unionOrderedSet:' is deprecated: method not yet implemented
Warning		'removeObject:' is deprecated: method not yet implemented
Warning		'minusOrderedSet:' is deprecated: method not yet implemented

RG-J:
I can implement them.

- RJ

rajsesh-msft:
@RG-J thanks.

RG-J:
Have the changes ready with the tests.
I'm attempting to push it and dealing with some lfs issues

PR #484, opened 2016-05-09
Added implementation for -setValuesForKeysWithDictionary:

triplef:
Should match reference platform documentation:

> The default implementation invokes setValue:forKey: for each key-value pair, substituting nil for NSNull values in keyedValues.

### DIFF
--- a/Frameworks/Foundation/NSKeyValueCoding.mm
+++ b/Frameworks/Foundation/NSKeyValueCoding.mm
@@ -429,6 +429,19 @@ bool KVCSetViaIvar(NSObject* self, struct objc_ivar* ivar, id value) {
 }
 
 /**
+ @Status Interoperable
+*/
+- (void)setValuesForKeysWithDictionary:(NSDictionary*)keyedValues {
+    for (NSString *key in keyedValues) {
+        id value = [keyedValues objectForKey:key];
+        if (value == [NSNull null]) {
+            value = nil;
+        }
+        [self setValue:value forKey:key];
+    }
+}
+
+/**
  @Status Stub
  @Notes
 */
@@ -471,14 +484,6 @@ bool KVCSetViaIvar(NSObject* self, struct objc_ivar* ivar, id value) {
 - (NSMutableOrderedSet*)mutableOrderedSetValueForKeyPath:(NSString*)keyPath {
     UNIMPLEMENTED();
     return StubReturn();
-}
-
-/**
- @Status Stub
- @Notes
-*/
-- (void)setValuesForKeysWithDictionary:(NSDictionary*)keyedValues {
-    UNIMPLEMENTED();
 }
 
 /**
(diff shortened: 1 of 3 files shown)


DHowett-MSFT:
:shipit:

jofre-ms:
:+1:

davelamb:
:+1:

rajsesh-msft:
Could we add at least one unit test to verify this change please?

rajsesh-msft:
:clock1:

MSFTFox:
:shipit:

Frameworks/Foundation/NSKeyValueCoding.mm
```diff
@@ -429,6 +429,18 @@ - (void)setValue:(id)value forUndefinedKey:(NSString*)key {
 }
 
 /**
+ @Status Interoperable
+*/
+- (void)setValuesForKeysWithDictionary:(NSDictionary*)keyedValues {
+    for (NSString *key in keyedValues) {
+        id value = [keyedValues objectForKey:key];
+        if (value == [NSNull null])
+            value = nil;
```
msft-Jeyaram: nit: would prefer if this is wrapped in {}. Just to be safe for future changes.

triplef:
Thanks for your feedback, I’ve added the missing { } around `value = nil;`.

Where would be the best place to add a unit test for this?

rajsesh-msft:
tests\unittests\foundation\NSObject_NSKeyValueArrayAdaptersTests.mm would be a reasonable place. Thanks again for your contributions!

---
In reply to: [218444600](https://github.com/Microsoft/WinObjC/pull/484#issuecomment-218444600) [](ancestors = 218444600)

triplef:
Thanks Raj, I’ve added a basic unit test for this. Let me know if there’s a better way to test this.

Please note that while I verified the test manually, I wasn’t able to actually run the test suite: the runall.bat script [referenced in the wiki](https://github.com/Microsoft/WinObjC/wiki/How-to-Contribute#running-tests) seems to no longer exist, and the tests don’t seem to be integrated in Visual Studio as selecting "Run All Tests" here didn’t find any tests to run. It would be great if the documentation could be updated for how to run tests.

bbowman:
Please use .\tests\Run-UnitTests.ps1 to run tests now. Apologies for the out of date documentation.

triplef:
Here’s what I get when I try to run the unit tests like this:

```
PS C:\Dev\winobjc-src> .\tests\Run-UnitTests.ps1
---------------------------------------------------
SUMMARY Passed: 0 Failed: 0 Disabled: 0 Total: 0
```

I have previously done a release build of the "build" project. I’m probably missing something.

bbowman:
Sorry I don't have it in front of me to tell you the exact flag but I think you'll need to use -Configuration Release (defaults to debug builds).

rajsesh-msft:
Please add -config release to the command line. We will be sure to update the wiki today, thanks for pointing this out!

triplef:
Thanks guys, that works. The test is passing.

DHowett-MSFT:
Thanks!

PR #488, opened 2016-05-10
Fix ~ipad specific bundle loading

ehren:
~ipad should be lower case. Also fixes setStartupDisplayMode: called too late (tablet mode must be configured to load ~ipad resources).

----------

Note that reloading the main bundle in response to changing tablet mode after startup isn't supported.

### DIFF
--- a/Frameworks/Foundation/NSBundle.mm
+++ b/Frameworks/Foundation/NSBundle.mm
@@ -245,11 +245,11 @@ bool ScanFilename(BundleFile* dest, char* pDirectory, char* pFilename) {
         }
         dest->pszPlatform = copyWithoutExtension("~iphone", "");
         szTemp[strlen(szTemp) - 7] = 0;
-    } else if (hasExtension(szTemp, "~iPad")) {
+    } else if (hasExtension(szTemp, "~ipad")) {
         if (!isTabletDevice()) {
             return false;
         }
-        dest->pszPlatform = copyWithoutExtension("~iPad", "");
+        dest->pszPlatform = copyWithoutExtension("~ipad", "");
         szTemp[strlen(szTemp) - 5] = 0;
     }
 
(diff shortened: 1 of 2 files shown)


rajsesh-msft:
@Ramu-msft is added to the review. #Closed

rajsesh-msft:

Frameworks/Foundation/NSBundle.mm
```diff
@@ -867,8 +867,8 @@ - (NSString*)localizedStringForKey:(NSString*)key value:(NSString*)value table:(
             memmove(pPos, pPos + 7, strlen(pPos + 7) + 1);
         }
     } else {
-        if (strstr(szPath, "~iPad") != NULL) {
-            char* pPos = strstr(szPath, "~iPad");
+        if (strstr(szPath, "~ipad") != NULL) {
+            char* pPos = strstr(szPath, "~ipad");
             memmove(pPos, pPos + 4, strlen(pPos + 4) + 1);
```
bviglietta: It looks like this would work better with 5.

Ramu-msft:
:shipit:

ehren:
Let me know if I should squash the second commit addressing @bviglietta's comment. That code actually won't be called in either case. There is some commented out code in `pathForResource:ofType:` that would call it -- so looks like ~ipad resources won't be handled correctly when using that API. My use case is just to have ~ipad specific xib files chosen on startup which is fixed by this pull.

rajsesh-msft:
@ehren - squash is fine.  We will merge squash anyway.

ehren:
squashed

bviglietta:
:shipit:

rajsesh-msft:
Waiting for CI build to finish.

---
In reply to: [218510109](https://github.com/Microsoft/WinObjC/pull/488#issuecomment-218510109) [](ancestors = 218510109)

bdrlamb:
:shipit:

rajsesh-msft:
:shipit:

rajsesh-msft:
Thanks for your contributions! :100:

PR #492, opened 2016-05-13
Updated NSInvocation to handle up to 8 method parameters

triplef:
Previously this was limited to 6 parameters.

While this should probably be rewritten to support an arbitrary number of parameters, this update should cover some more extreme cases and allows us to use NSInvocation for our specific use case. :)


bviglietta:
:shipit:

msft-Jeyaram:
:shipit:

DHowett-MSFT:
:shipit:

rajsesh-msft:
@triplef thanks!

PR #499, opened 2016-05-16
NSSet containsObject should return NO if the object is nil.

### LINKED ISSUE #498: NSSet containsObject should return NO for nil object.

RG-J:
fix #498

### DIFF
--- a/Frameworks/Foundation/NSSet.mm
+++ b/Frameworks/Foundation/NSSet.mm
@@ -526,6 +526,9 @@ int NSSetEnumeratorGetNextObject(NSSet* set, void* enumeratorHolder, id* ret, in
  @Status Interoperable
 */
 - (id)member:(id)obj {
+    if (obj == nil) {
+        return nil;
+    }
     return NSSetTableMember(self, obj);
 }
 
--- a/build/Tests/UnitTests/Foundation/Foundation.UnitTests.vcxproj
+++ b/build/Tests/UnitTests/Foundation/Foundation.UnitTests.vcxproj
@@ -298,6 +298,7 @@
     <ClangCompile Include="$(StarboardBasePath)\tests\unittests\Foundation\NSPredicateTests.m" />
     <ClangCompile Include="$(StarboardBasePath)\tests\unittests\Foundation\NSProcessInfoTests.mm" />
     <ClangCompile Include="$(StarboardBasePath)\tests\unittests\Foundation\NSProgressTests.mm" />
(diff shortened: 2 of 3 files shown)


Frameworks/Foundation/NSSet.mm
```diff
@@ -14,15 +14,15 @@
...
-#import "Foundation/NSMutableSet.h"
 #import "Foundation/NSCountedSet.h"
 #import "Foundation/NSEnumerator.h"
-#import "NSEnumeratorInternal.h"
 #import "Foundation/NSKeyedArchiver.h"
+#import "Foundation/NSMutableSet.h"
+#import "NSEnumeratorInternal.h"
 #import "NSKeyedArchiverInternal.h"
+#import "Starboard.h"
+#import "StubReturn.h"
```
rajsesh-msft: please revert the changes to the header file ordering.

rajsesh-msft:

rajsesh-msft:

Frameworks/Foundation/NSSet.mm
```diff
@@ -370,6 +370,9 @@ + (NSSet*)set {
  @Status Interoperable
 */
 - (BOOL)containsObject:(id)obj {
+    if (obj == nil) {
+        return NO;
+    }
```
rajsesh-msft: please move this to member method instead. We are getting a new version of this class and that will make merging this easier.

rajsesh-msft:
:clock1:

RG-J:
@rajsesh-msft nil issue addressed, please take a look.

jaredhms:
:shipit:

Frameworks/Foundation/NSSet.mm
```diff
@@ -526,6 +526,9 @@ - (NSArray*)sortedArrayUsingDescriptors:(NSArray*)descriptors {
  @Status Interoperable
 */
 - (id)member:(id)obj {
+    if (obj == nil) {
+        return nil;
+    }
```
bbowman: Insufficient. You should make sure that nil can't be added to the set. That more closely matches the reference platform and then this check would not be needed.

bbowman:
:clock1:

Frameworks/Foundation/NSSet.mm
bbowman: discussed offline. we still want this check and ideally would like a nil check on adding a member in the mutable version. The change, as is, is good though.

---
In reply to: [63953430](https://github.com/Microsoft/WinObjC/pull/499#discussion_r63953430) [](ancestors = 63953430)

PR #512, opened 2016-05-24
Fix misc. WOCCatalog breaks in Xcode

s5msft:
Various (small) fixes to parts of WOCCatalog when built for iOS

### DIFF
--- a/samples/WOCCatalog/WOCCatalog/CoreMotionViewController.mm
+++ b/samples/WOCCatalog/WOCCatalog/CoreMotionViewController.mm
@@ -42,7 +42,6 @@
         [self setupAccelerometer];
     } else {
         accLabel = [[UILabel alloc] initWithFrame:CGRectMake(30, accHeight, 300, 50)];
-        [accLabel setBackgroundColor:nil];
         [accLabel setText:@"Accelerometer Not Available!"];
         [accLabel setTextAlignment:NSTextAlignmentLeft];
         [scrollView addSubview:accLabel];
@@ -54,7 +53,6 @@
         [self setupGyrometer];
     } else {
         gyroLabel = [[UILabel alloc] initWithFrame:CGRectMake(30, gyroHeight, 300, 50)];
-        [gyroLabel setBackgroundColor:nil];
         [gyroLabel setText:@"Gyrometer Not Available!"];
         [gyroLabel setTextAlignment:NSTextAlignmentLeft];
         [scrollView addSubview:gyroLabel];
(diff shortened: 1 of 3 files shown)


samples/WOCCatalog/WOCCatalog/PickersViewController.m
```diff
@@ -23,31 +23,22 @@ - (void)viewDidLoad {
...
+    [self tableView].separatorColor = [UIColor clearColor];
     [self tableView].allowsSelection = NO;
+    [self tableView].scrollEnabled = NO;
 
     self.title = @"Pickers";
     
     // PICKER VIEW
-    
-    self.datePickerView = [[UIDatePicker alloc] initWithFrame:CGRectZero];
+    self.datePickerView = [[UIDatePicker alloc] initWithFrame:[self calcFrameForPicker]];
```
DHowett-MSFT: If we're overriding the size later, there's no point in setting it here.
If there is, something's broken :)

samples/WOCCatalog/WOCCatalog/PickersViewController.m
```diff
@@ -129,20 +95,19 @@ - (void)action:(id)sender
     }
 }
 
-- (CGRect)calcFrameForPicker:(CGSize)size
+- (CGRect)calcFrameForPicker
 {
     CGRect screenRect = [[UIScreen mainScreen] applicationFrame];
```
DHowett-MSFT: This should probably be self.view.bounds -- the view should scale first, and its contents should not be based on screen size

samples/WOCCatalog/WOCCatalog/CoreMotionViewController.mm
```diff
@@ -42,7 +42,6 @@ - (void)viewDidLoad {
         [self setupAccelerometer];
     } else {
         accLabel = [[UILabel alloc] initWithFrame:CGRectMake(30, accHeight, 300, 50)];
-        [accLabel setBackgroundColor:nil];
```
DHowett-MSFT: In which colour does nil result? Is it different in iOS versus WinObjC? Should it ever be clear?

samples/WOCCatalog/WOCCatalog/PickersViewController.m
s5msft: Several of these were just to maintain some of the existing code. But, fine, I'll update and resubmit.

samples/WOCCatalog/WOCCatalog/CoreMotionViewController.mm
s5msft: nil is black, and unspecified is clear

Ramu-msft:

Ramu-msft:

Ramu-msft:

samples/WOCCatalog/WOCCatalog/CoreMotionViewController.mm
bdrlamb: Might want to use [UIColor blackColor] for more clarity

---
In reply to: [64473453](https://github.com/Microsoft/WinObjC/pull/512#discussion_r64473453) [](ancestors = 64473453)

DHowett-MSFT:
:shipit:

PR #559, opened 2016-06-14
Implemented CGPathApply.

### LINKED ISSUE #525: Implement CG Path APIs
- [x] CGPathApply
- [x] CGPathContainsPoint
- [x] CGContextCopyPath
- [ ] CGContextReplacePathWithStrokedPath

adein:
Replaced pathComponent with CGPathElement. Rewrote CGPathAddEllipseInRect,
CGPathAddArc, CGPathAddArcToPoint to support CGPathElement.
Implemented CGPathGetCurrentPoint for use in CGPathAddArcToPoint.
Removed caveats for CGPathAddArc, CGPathAddEllipseInRect.
Caveat for CGPathAddArcToPoint remains.

Fix #525 : CGPathApply.

### DIFF
--- a/Frameworks/CoreGraphics/CGPath.mm
+++ b/Frameworks/CoreGraphics/CGPath.mm
@@ -156,9 +156,9 @@ void _CGPathAddElement(
         path->_elements = (CGPathElementInternal*)IwRealloc(path->_elements, path->_max * sizeof(CGPathElementInternal));
     }
     CGPathElementInternal* element = &path->_elements[path->_count];
-    // The new elements needs its points array pointer updated
+    // The new element needs to call init
     // because it was created with alloc
-    element->updatePointsArrayPointer();
+    element->init();
     element->type = type;
     element->points[0] = p0;
     element->points[1] = p1;
@@ -191,10 +191,10 @@ CGMutablePathRef CGPathCreateMutableCopy(CGPathRef path) {
     ret->_count = path->_count;
     ret->_elements = (CGPathElementInternal*)IwRealloc(ret->_elements, ret->_max * sizeof(CGPathElementInternal));
     memcpy(ret->_elements, path->_elements, path->_count * sizeof(CGPathElementInternal));
-    // All of the new elements need their points array pointer updated
+    // All of the new elements need to call init
     // because they were created with alloc + memcpy
     for (unsigned i = 0; i < path->_count; i++) {
-        ret->_elements[i].updatePointsArrayPointer();
+        ret->_elements[i].init();
     }
     return ret;
 }
@@ -436,21 +436,21 @@ void CGPathAddPath(CGMutablePathRef path, const CGAffineTransform* m, CGPathRef 
     }
 
     for (unsigned i = 0; i < copyObj->_count; i++) {
-        CGPathElementInternal c = copyObj->_elements[i];
+        pathObj->_elements[pathObj->_count] = copyObj->_elements[i];
+        pathObj->_count++;
+        CGPathElementInternal* c = &pathObj->_elements[pathObj->_count];
+
         if (m) {
-            switch (c.type) {
+            switch (c->type) {
                 case kCGPathElementMoveToPoint:
                 case kCGPathElementAddLineToPoint:
-                    c.points[0] = CGPointApplyAffineTransform(c.points[0], *m);
+                    c->points[0] = CGPointApplyAffineTransform(c->points[0], *m);
                     break;
                 default:
                     // Append the path anyway, without transforming.
                     break;
             }
         }
-
-        pathObj->_elements[pathObj->_count] = c;
-        pathObj->_count++;
     }
 }
 
@@ -723,12 +723,12 @@ bool CGPathEqualToPath(CGPathRef path1, CGPathRef path2) {
         }
         unsigned pointCount = _CGPathPointCountForElementType(element1->type);
         for (unsigned p = 0; p < pointCount; p++) {
-            CGPoint p1 = element1->points[p];
-            CGPoint p2 = element2->points[p];
-            if (abs(p1.x - p2.x) > FLT_EPSILON) {
+            CGPoint* p1 = &element1->points[p];
+            CGPoint* p2 = &element2->points[p];
+            if (abs(p1->x - p2->x) > FLT_EPSILON) {
                 return false;
             }
-            if (abs(p1.y - p2.y) > FLT_EPSILON) {
+            if (abs(p1->y - p2->y) > FLT_EPSILON) {
                 return false;
             }
         }
--- a/Frameworks/CoreGraphics/CGPath.mm
+++ b/Frameworks/CoreGraphics/CGPath.mm
@@ -455,14 +455,12 @@ void CGPathAddPath(CGMutablePathRef path, const CGAffineTransform* m, CGPathRef 
     }
 }
 
-
 /**
  @Status Interoperable
 */
 void CGPathAddEllipseInRect(CGMutablePathRef path, const CGAffineTransform* m, CGRect rect) {
+    CGFloat offsetMultiplier = _CGPathControlPointOffsetMultiplier(M_PI_2);
 
-	CGFloat offsetMultiplier = _CGPathControlPointOffsetMultiplier(M_PI_2);// (4.0/3.0)*tan(M_PI/8.0);
-    
     CGFloat xControlPointOffset = offsetMultiplier * CGRectGetWidth(rect)/2.0;
     CGFloat yControlPointOffset = offsetMultiplier * CGRectGetHeight(rect)/2.0;
     
--- a/Frameworks/CoreGraphics/CGPath.mm
+++ b/Frameworks/CoreGraphics/CGPath.mm
@@ -59,12 +59,6 @@ public:
 
 __CGPath::~__CGPath() {
     if (_elements) {
-        for (unsigned i = 0; i < _count; i++) {
-            CGPoint* points = _elements[i].points;
-            if (points) {
-                IwFree(points);
-            }
-        }
         IwFree(_elements);
     }
 }
@@ -159,10 +153,12 @@ void _CGPathAddElement(
     CGPathRef path, CGPathElementType type, CGPoint p0 = CGPointZero, CGPoint p1 = CGPointZero, CGPoint p2 = CGPointZero) {
     if (path->_count + 1 >= path->_max) {
         path->_max += 32;
-        path->_elements = (CGPathElement*)IwRealloc(path->_elements, path->_max * sizeof(CGPathElement));
+        path->_elements = (CGPathElementInternal*)IwRealloc(path->_elements, path->_max * sizeof(CGPathElementInternal));
     }
-    CGPathElement* element = &path->_elements[path->_count];
-    element->points = (CGPoint*)IwCalloc(kCGPathMaxPointCount, sizeof(CGPoint));
+    CGPathElementInternal* element = &path->_elements[path->_count];
+    // The new elements needs its points array pointer updated
+    // because it was created with alloc
+    element->updatePointsArrayPointer();
     element->type = type;
     element->points[0] = p0;
     element->points[1] = p1;
@@ -193,11 +189,12 @@ CGMutablePathRef CGPathCreateMutableCopy(CGPathRef path) {
     auto ret = __CGPath::alloc(nil);
     ret->_max = path->_max;
     ret->_count = path->_count;
-    ret->_elements = (CGPathElement*)IwRealloc(ret->_elements, ret->_max * sizeof(CGPathElement));
-    memcpy(ret->_elements, path->_elements, path->_count * sizeof(CGPathElement));
+    ret->_elements = (CGPathElementInternal*)IwRealloc(ret->_elements, ret->_max * sizeof(CGPathElementInternal));
+    memcpy(ret->_elements, path->_elements, path->_count * sizeof(CGPathElementInternal));
+    // All of the new elements need their points array pointer updated
+    // because they were created with alloc + memcpy
     for (unsigned i = 0; i < path->_count; i++) {
-        ret->_elements[i].points = (CGPoint*)IwCalloc(kCGPathMaxPointCount, sizeof(CGPoint));
-        memcpy(ret->_elements[i].points, path->_elements[i].points, kCGPathMaxPointCount * sizeof(CGPoint));
+        ret->_elements[i].updatePointsArrayPointer();
     }
     return ret;
 }
@@ -435,14 +432,11 @@ void CGPathAddPath(CGMutablePathRef path, const CGAffineTransform* m, CGPathRef 
 
     if (pathObj->_count + copyObj->_count >= pathObj->_max) {
         pathObj->_max += copyObj->_count;
-        pathObj->_elements = (CGPathElement*)IwRealloc(pathObj->_elements, pathObj->_max * sizeof(CGPathElement));
+        pathObj->_elements = (CGPathElementInternal*)IwRealloc(pathObj->_elements, pathObj->_max * sizeof(CGPathElementInternal));
     }
 
     for (unsigned i = 0; i < copyObj->_count; i++) {
-        CGPathElement c = copyObj->_elements[i];
-        c.points = (CGPoint*)IwCalloc(kCGPathMaxPointCount, sizeof(CGPoint));
-        memcpy(c.points, copyObj->_elements[i].points, kCGPathMaxPointCount * sizeof(CGPoint));
-
+        CGPathElementInternal c = copyObj->_elements[i];
         if (m) {
             switch (c.type) {
                 case kCGPathElementMoveToPoint:
@@ -653,8 +647,7 @@ void CGPathApply(CGPathRef path, void* info, CGPathApplierFunction function) {
     // TODO: Add check for NULL and return. Add for other relevant functions in this file.
 
     for (unsigned i = 0; i < path->_count; i++) {
-        CGPathElement element = path->_elements[i];
-        function(info, &element);
+        function(info, &path->_elements[i]);
     }
 }
 
@@ -723,15 +716,15 @@ bool CGPathEqualToPath(CGPathRef path1, CGPathRef path2) {
     }
 
     for (unsigned i = 0; i < path1->_count; i++) {
-        CGPathElement element1 = path1->_elements[i];
-        CGPathElement element2 = path2->_elements[i];
-        if (element1.type != element2.type) {
+        CGPathElement* element1 = &path1->_elements[i];
+        CGPathElement* element2 = &path2->_elements[i];
+        if (element1->type != element2->type) {
             return false;
         }
-        unsigned pointCount = _CGPathPointCountForElementType(element1.type);
+        unsigned pointCount = _CGPathPointCountForElementType(element1->type);
         for (unsigned p = 0; p < pointCount; p++) {
-            CGPoint p1 = element1.points[p];
-            CGPoint p2 = element2.points[p];
+            CGPoint p1 = element1->points[p];
+            CGPoint p2 = element2->points[p];
             if (abs(p1.x - p2.x) > FLT_EPSILON) {
                 return false;
             }
@@ -748,15 +741,15 @@ bool CGPathEqualToPath(CGPathRef path1, CGPathRef path2) {
 */
 CGPoint CGPathGetCurrentPoint(CGPathRef path) {
     if (path->_count > 0) {
-        CGPathElement c = path->_elements[path->_count - 1];
-        switch (c.type) {
+        CGPathElement* c = &path->_elements[path->_count - 1];
+        switch (c->type) {
             case kCGPathElementMoveToPoint:
             case kCGPathElementAddLineToPoint:
-                return c.points[0];
+                return c->points[0];
             case kCGPathElementAddQuadCurveToPoint:
-                return c.points[1];
+                return c->points[1];
             case kCGPathElementAddCurveToPoint:
-                return c.points[2];
+                return c->points[2];
             default:
                 return CGPointZero;
         }
--- a/Frameworks/CoreGraphics/CGPath.mm
+++ b/Frameworks/CoreGraphics/CGPath.mm
@@ -59,6 +59,12 @@ public:
 
 __CGPath::~__CGPath() {
     if (_components) {
+		for (unsigned i = 0; i < _count; i++) {
+			CGPoint * points = _components[i].points;
+			if( points ){
+				IwFree(points);
+			}
+		}
         IwFree(_components);
     }
 }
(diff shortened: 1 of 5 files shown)


rajsesh-msft:

rajsesh-msft:
Are there tests for these apis that need to be part of this pull request?

adein:
@rajsesh-msft Are you requesting that unit tests be added to the pull request, or for details on how we tested the changes?  We tested the changes using a sample project; we have not created unit tests in the bridge.

rajsesh-msft:
@adein, yes please. If you added a sample to woccatalog, please include that as well. the idea is to make sure we have test collateral to guard against any future regressions.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -59,69 +59,49 @@ void addPoint(float ptX, float ptY) {
 
 __CGPath::~__CGPath() {
     if (_components) {
+		for (unsigned i = 0; i < _count; i++) {
```
bbowman: nit: please use spaces instead of tabs to match the rest of the codebase. To that end, we have setup clang-format for the project and ask that all PRs have it run prior to submission so that we can avoid silly formatting comments like this. :-)

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -59,69 +59,49 @@ void addPoint(float ptX, float ptY) {
...
+			CGPoint * points = _components[i].points;
```
jaredhms: nit: please be sure to run tools/scripts/clang-format.ps1 on all modified files

rajsesh-msft:
@jeffMeador could you please rebase the branch and push again?

jeffMeador:
@rajsesh-msft This branch now contains the test project & unit tests for CGPathApply. 

The unit tests are in tests/unittests/CoreGraphics/CGPathTests.mm
The sample code is in samples/CGCatalog

rajsesh-msft:
Could you pull latest changes from github/master your branch, do a git -rebase and resubmit the pull request?  The way it stands now, the PR shows 188 changed files and is hard to go through.

tests/unittests/CoreGraphics/CGPathTests.mm
```diff
@@ -14,7 +14,224 @@
...
+            break;
+    }
+    return pointCount;
+}
+
+// CGPathApplierFunction that attempts to modify elements inside a CGPath
+void cgPathModifyingApplierFunction(void* info, const CGPathElement* element) {
+    int pointCount = cgPathPointCountForElementType(element->type);
+
+    NSMutableArray* points = [NSMutableArray array];
```
rajsesh-msft: element->points? otherwise this seems unused...

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -655,13 +637,46 @@ void CGPathAddRoundedRect(
...
+
     for (unsigned i = 0; i < path->_count; i++) {
-        CGPathElement c = path->_components[i];
-        function(info, &c);
+        CGPathElement element = path->_components[i];
+        CGPathElement returnElement;
+        returnElement.type = element.type;
+        int pointCount = _CGPathPointCountForElementType(element.type);
+        returnElement.points = (CGPoint*)IwCalloc(pointCount, sizeof(CGPoint));
+        memcpy(returnElement.points, element.points, pointCount * sizeof(CGPoint));
```
rajsesh-msft: I am not sure if this is really needed.  Documentation for CFPathApply says that the function cannot modify the path, it implies no guarantees that if the function does indeed modify the path, it will be ignored.

jeffMeador: While the documentation doesn't explicitly state that CGPathApply discards changes, the original implementation does discard changes. I don't think it works exactly like this, but it has the same effect. I could add something to the CGCatalog sample project to show this functionality on both platforms.

tests/unittests/CoreGraphics/CGPathTests.mm
jeffMeador: Are you referring to the points NSMutableArray not being used? It's not used, I can remove it. cgPathModifyingApplierFunction just exists to show that paths cannot be manipulated in an applier function.

jeffMeador:
@rajsesh-msft This branch has been properly rebased and now only shows the 5 files that have been edited.

rajsesh-msft:
Thanks a bunch, this is a lot easier to look at.

---
In reply to: [228855101](https://github.com/Microsoft/WinObjC/pull/559#issuecomment-228855101) [](ancestors = 228855101)

Frameworks/CoreGraphics/CGPath.mm
rajsesh-msft: The reference platform behavior may be an artifact of their implementation rather than the intent.  We should not be trying to match this behavior.  Any app that relies on being able to change the path is probably behaving incorrectly anyways, so i'd rather fail fast in these scenarios than support it.  Thoughts?

---
In reply to: [68611364](https://github.com/Microsoft/WinObjC/pull/559#discussion_r68611364) [](ancestors = 68611364)

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -615,12 +637,47 @@ void CGPathAddRoundedRect(
...
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 void CGPathApply(CGPathRef path, void* info, CGPathApplierFunction function) {
-    UNIMPLEMENTED();
+    if (path == NULL) {
+        return;
+    }
+
```
rajsesh-msft: Nit: it looks like CGPathApply is the only one that is doing the null check.  fail fast (by not checking NULL) will make it consistent.

Frameworks/CoreGraphics/CGPath.mm
jeffMeador: I agree that anyone trying to manipulate paths is using this API incorrectly. However, if I don't discard changes, then this does become a spot where users can intentionally or unintentionally manipulate paths. If that is the case, the applier function would be behaving differently from the documentation as a function that "can examine (but not modify) the element" because it would be able to modify the element.

rajsesh-msft: We are interpreting "can examine (but not modify) the element" differently.  IMO, it means that the function can examine it, but it should not modify it.  There is really no utility in allowing the function to modify it but guarding against it.  Looking at the applier function: typedef void (*CGPathApplierFunction) ( void *info, const CGPathElement *element );  the const on second parameter implies its contents must not be modified by the function.  We are not really concerned about a rogue user or code here as this runs in the context of the app.  Is there a legitimate case for an app modifying this element? if not, I'd rather remove all these duplications - it will matter a lot for perf.

---
In reply to: [68664667](https://github.com/Microsoft/WinObjC/pull/559#discussion_r68664667) [](ancestors = 68664667)

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -214,8 +177,8 @@ CGMutablePathRef CGPathCreateMutableCopy(CGPathRef path) {
     auto ret = __CGPath::alloc(nil);
     ret->_max = path->_max;
     ret->_count = path->_count;
-    ret->_components = (pathComponent*)IwRealloc(ret->_components, ret->_max * sizeof(pathComponent));
-    memcpy(ret->_components, path->_components, path->_count * sizeof(pathComponent));
+    ret->_components = (CGPathElement*)IwRealloc(ret->_components, ret->_max * sizeof(CGPathElement));
+    memcpy(ret->_components, path->_components, path->_count * sizeof(CGPathElement));
```
rajsesh-msft: This copy is shallow.  the _components->points is now shared between the two instances of CGPaths and will result in double free.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -239,45 +202,117 @@ void CGPathAddLineToPoint(CGMutablePathRef path, const CGAffineTransform* m, flo
...
-        pathObj->_components = (pathComponent*)IwRealloc(pathObj->_components, pathObj->_max * sizeof(pathComponent));
+        pathObj->_components = (CGPathElement*)IwRealloc(pathObj->_components, pathObj->_max * sizeof(CGPathElement));
     }
 
-    pathObj->_components[pathObj->_count].type = pathComponentLineTo;
-    pathObj->_components[pathObj->_count].point.x = x;
-    pathObj->_components[pathObj->_count].point.y = y;
+    pathObj->_components[pathObj->_count].type = kCGPathElementAddLineToPoint;
+
+    pathObj->_components[pathObj->_count].points = (CGPoint*)IwCalloc(1, sizeof(CGPoint));
```
rajsesh-msft: This approach, while really tight on memory usage, is not really buying us a lot and will cause performance issue on two fronts: 

1. Lot of small memory allocations (1, 2, 3 or 4 points).
2. Locality of reference of path element + points.

I suggest that we allocate the buffer long enough to hold CGPathElement and the maximum number of points.  The points buffer will be at the end of the element.  While this may sound bad, it actually isn't since memory wont be fragmented anymore.  Also addresses the other shallow copy comment.

Frameworks/include/CGPathInternal.h
```diff
@@ -17,65 +17,12 @@
...
-        curveToPoint ctp;
-        ellipseInRect eir;
-        quadCurveToPoint qtp;
-        arcAngle aa;
-    };
-} pathComponent;
 
 struct __CGPath : public CFBridgeBase<__CGPath> {
-    pathComponent* _components;
+    CGPathElement* _components;
```
rajsesh-msft: In light of the other comment on allocation points inline, it probably makes it easier to have CGPathElementInternal which derives from CGPathElement and adds four points.  You can then have the constructor set the pointer as well, simplifying some boilerplate.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -239,45 +202,117 @@ void CGPathAddLineToPoint(CGMutablePathRef path, const CGAffineTransform* m, flo
...
+    pathObj->_components[pathObj->_count].points[0] = { x, y };
 
     pathObj->_count++;
 }
 
+CGFloat _CGPathControlPointOffsetMultiplier(CGFloat angle) {
```
rajsesh-msft: A line of comment explaining what this is will be helpful.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -239,45 +202,117 @@ void CGPathAddLineToPoint(CGMutablePathRef path, const CGAffineTransform* m, flo
...
+    double dx0, dy0, dx2, dy2, xl0, xl2;
+    double san, n0x, n0y, n2x, n2y, t;
+
+    x0 = curPathPosition.x;
+    y0 = curPathPosition.y;
+
+    dx0 = x0 - x1;
+    dy0 = y0 - y1;
+    xl0 = sqrt(dx0 * dx0 + dy0 * dy0);
+    if (xl0 == 0)
```
rajsesh-msft: Should be FLT_EPSILON or COMPARISON_EPSILON instead.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -239,45 +202,117 @@ void CGPathAddLineToPoint(CGMutablePathRef path, const CGAffineTransform* m, flo
...
 
+CGFloat _CGPathControlPointOffsetMultiplier(CGFloat angle) {
+    return (4.0f / 3.0f) * tan(angle / 4.0f);
+}
+
 /**
  @Status Caveat
  @Notes transform property not supported
 */
 void CGPathAddArcToPoint(CGMutablePathRef path, const CGAffineTransform* m, float x1, float y1, float x2, float y2, float radius) {
```
rajsesh-msft: The documentation on this is not very clear.  "If the current point and the first tangent point of the arc (the starting point) are not equal, Quartz appends a straight line segment from the current point to the first tangent point." On the other hand, the first tangent is from current point to x1, y1.  Does the reference platform bail out if current point is same as x1, y1 like the code below (it makes sense, but again,the reference doc is not clear).

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -239,45 +202,117 @@ void CGPathAddLineToPoint(CGMutablePathRef path, const CGAffineTransform* m, flo
...
-        pathObj->_max += 32;
-        pathObj->_components = (pathComponent*)IwRealloc(pathObj->_components, pathObj->_max * sizeof(pathComponent));
+    double x0, y0;
+    double dx0, dy0, dx2, dy2, xl0, xl2;
+    double san, n0x, n0y, n2x, n2y, t;
+
+    x0 = curPathPosition.x;
+    y0 = curPathPosition.y;
+
+    dx0 = x0 - x1;
```
rajsesh-msft: should this be x1-x0?  the sign should matter as the arc is between lines from (x0, y0)->(x1,y1) and (x1,y1) -> (x2, y2), implying x1, y1 needs to be reversed here.  It would help if you could add comments with the math behind these computations, would help with these questions immensely.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -327,12 +369,12 @@ void CGPathMoveToPoint(CGMutablePathRef path, const CGAffineTransform* m, float
 
     if (pathObj->_count + 1 >= pathObj->_max) {
```
rajsesh-msft: This entire code block is repeated in a few places, might make sense to move this to a utility function and call it instead.  Understood that this was in the existing code and not new, would still be a good improvement.

rajsesh-msft:
It would help to add some comments about the math in all the new methods.  

---
In reply to: [228855305](https://github.com/Microsoft/WinObjC/pull/559#issuecomment-228855305) [](ancestors = 228855305,228855101)

rajsesh-msft:
:clock1:

Frameworks/CoreGraphics/CGPath.mm
jeffMeador: I copied this function directly from CGContextImpl::CGContextAddArcToPoint, which appears to be copied to/from CGContextCairo::CGContextAddArcToPoint. Ideally this logic would live in CGPath and CGContext would call into CGPath so it could be removed from the other locations. I understand cairo can complicate that particular implementation. How much do you want me to handle in this pull request? If I increase readability here and fix any bugs, how would you like to handle this code in the other 2 spots?

rajsesh-msft: At the very least, comments here that this is the same code in other two spots (with a TODO to refactor them) would immensely help.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -239,45 +202,117 @@ void CGPathAddLineToPoint(CGMutablePathRef path, const CGAffineTransform* m, flo
...
+    double x0, y0;
+    double dx0, dy0, dx2, dy2, xl0, xl2;
+    double san, n0x, n0y, n2x, n2y, t;
+
+    x0 = curPathPosition.x;
+    y0 = curPathPosition.y;
+
+    dx0 = x0 - x1;
+    dy0 = y0 - y1;
+    xl0 = sqrt(dx0 * dx0 + dy0 * dy0);
```
bbowman: Would be nice to use more descriptive names across the board here. dx0 could be deltaX0 and x10 could be distance10. I know its common to be extra terse in general math but more descriptive names aids the casual reader in remembering what the components all are when the code is longer than a single screen etc.

bbowman: Actually would be nice to make dx0 deltaX10. I was confused why there was no dx1 / dy1 but that is because a single point was chosen for the deltas. 

---
In reply to: [68683240](https://github.com/Microsoft/WinObjC/pull/559#discussion_r68683240) [](ancestors = 68683240)

Frameworks/CoreGraphics/CGPath.mm
jeffMeador: If this is moved to a utility function for adding a new CGPathElement, could we also put the dynamic allocation for the points array (always of the same size) in this utility function and forgo the CGPathElementInternal derived struct? This logic occurs every time a new element is created. Let me know if you had something different in mind for CGPathElementInternal, but as I understand you just wanted it to allocate memory enough memory for the max number of CGPoints and assign it to the points pointer in CGPathElement in the constructor of CGPathElementInternal.

rajsesh-msft: Yes.  That would work. The reason i suggested the internal structure is that it would then make size calculation etc trivial. 

---
In reply to: [68820359](https://github.com/Microsoft/WinObjC/pull/559#discussion_r68820359) [](ancestors = 68820359)

Frameworks/CoreGraphics/CGPath.mm
jeffMeador: The reference platform will bail out when there is no current point, like I do here.

The documentation states that if the current point and the first tangent point are the same, then it doesn't add a line from the current point to the first tangent point. Despite providing the points that should satisfy this condition, I haven't seen this behavior on the reference platform. The other bailout further down, `if (san == 0)` ,doesn't seem to line up with documentation and might also be something that was observed in testing. That can be addressed when this is function is refactored/cleaned up.

rajsesh-msft: Thanks Jeff, would be great if we can document these as caveats - where we chose compat over documented behavior.

---
In reply to: [68831259](https://github.com/Microsoft/WinObjC/pull/559#discussion_r68831259) [](ancestors = 68831259)

Frameworks/CoreGraphics/CGPath.mm
jeffMeador: Would you consider this implementation if `returnElement` was created once outside of the loop so just `memcpy` was in the loop and not the alloc/free? I'll remove if not.

rajsesh-msft: @jeffMeador I would consider it if there is actual valid usage for the case. Otherwise I prefer to not make copies here. If we see actual apps use this pattern we can always do this in the future.

jeffMeador:
@rajsesh-msft everything we discussed should be addressed in 184e31f31f9ca6bd8918de7af31fc0d7b337093c . `CGPathAddArcToPoint` was left untouched with a note that it needs to be refactored along with the copies in the other spots.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -192,6 +155,22 @@ void _CGPathGetBoundingBoxInternal(CGPathRef pathref, CGRect* rectOut) {
...
 }
 
+void _CGPathAddElement(
+    CGPathRef path, CGPathElementType type, CGPoint p0 = CGPointZero, CGPoint p1 = CGPointZero, CGPoint p2 = CGPointZero) {
+    if (path->_count + 1 >= path->_max) {
+        path->_max += 32;
+        path->_elements = (CGPathElement*)IwRealloc(path->_elements, path->_max * sizeof(CGPathElement));
+    }
+    CGPathElement* element = &path->_elements[path->_count];
+    element->points = (CGPoint*)IwCalloc(kCGPathMaxPointCount, sizeof(CGPoint));
```
rajsesh-msft: This is not what i meant.  We are still allocating memory every time a path element is added.  perhaps i misunderstood your earlier question.  Do you want to chat over the phone so we can get on the same page here?  I really want to avoid duplicate allocations for points every time an item is added to the path.  Allocating a fixed number of points without doing it once isn't really buying us anything.

jeffMeador:
@rajsesh-msft CGPath has been updated to use CGPathElementInternal and just have 1 allocation for an array of elements instead of an allocation for an array of elements & for each array of points.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -192,6 +149,24 @@ void _CGPathGetBoundingBoxInternal(CGPathRef pathref, CGRect* rectOut) {
...
+void _CGPathAddElement(
+    CGPathRef path, CGPathElementType type, CGPoint p0 = CGPointZero, CGPoint p1 = CGPointZero, CGPoint p2 = CGPointZero) {
+    if (path->_count + 1 >= path->_max) {
+        path->_max += 32;
+        path->_elements = (CGPathElementInternal*)IwRealloc(path->_elements, path->_max * sizeof(CGPathElementInternal));
+    }
+    CGPathElementInternal* element = &path->_elements[path->_count];
+    // The new elements needs its points array pointer updated
+    // because it was created with alloc
+    element->updatePointsArrayPointer();
```
rajsesh-msft: nit: from a OO standpoint, i would call this function init() (caller doesn't need to know internal details of what this function does).  There is alternately, no need for this function, you can call placement new on the memory location like this: CGPathElementInternal* element = new (&path->_elements[path->_count]) CGPathElementInternal();

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -390,70 +432,63 @@ void CGPathAddPath(CGMutablePathRef path, const CGAffineTransform* m, CGPathRef
...
     if (pathObj->_count + copyObj->_count >= pathObj->_max) {
         pathObj->_max += copyObj->_count;
-        pathObj->_components = (pathComponent*)IwRealloc(pathObj->_components, pathObj->_max * sizeof(pathComponent));
+        pathObj->_elements = (CGPathElementInternal*)IwRealloc(pathObj->_elements, pathObj->_max * sizeof(CGPathElementInternal));
     }
 
     for (unsigned i = 0; i < copyObj->_count; i++) {
-        pathComponent c = copyObj->_components[i];
-
+        CGPathElementInternal c = copyObj->_elements[i];
```
rajsesh-msft: nit: no need to make a copy here.  You could apply the transform after you call pathObj->_elements[...] = copyObj->_elements[i];

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -680,21 +708,54 @@ CGPathRef CGPathCreateWithRoundedRect(CGRect rect, CGFloat cornerWidth, CGFloat
...
+
+    for (unsigned i = 0; i < path1->_count; i++) {
+        CGPathElement* element1 = &path1->_elements[i];
+        CGPathElement* element2 = &path2->_elements[i];
+        if (element1->type != element2->type) {
+            return false;
+        }
+        unsigned pointCount = _CGPathPointCountForElementType(element1->type);
+        for (unsigned p = 0; p < pointCount; p++) {
+            CGPoint p1 = element1->points[p];
```
rajsesh-msft: You can avoid creating copies of p1 and p2 by expanding the expressions below or using ref/pointer instead.

rajsesh-msft:
Added a few minor comments to the new changes.

---
In reply to: [229352074](https://github.com/Microsoft/WinObjC/pull/559#issuecomment-229352074) [](ancestors = 229352074)

jeffMeador:
The changes requested have been committed.

On Jun 29, 2016, at 12:44 PM, Raj Seshasankaran <notifications@github.com<mailto:notifications@github.com>> wrote:


Added a few minor comments to the new changes.

In reply to: 229352074<https://github.com/Microsoft/WinObjC/pull/559#issuecomment-229352074> <x-msg://4/ancestors%20=%20229352074>

rajsesh-msft:
In CI build. validated internal apps manually and they look good so far.

rajsesh-msft:
CI build passed.

PR #557, opened 2016-06-14
Fix an issue with searching insertion index in NSArray.

bSr43:
This PR implements the behavior that we find on OS X when searching for an insertion index in an NSArray.

PR #569, opened 2016-06-21
Fixes binary search of non-existing elements in an NSArray.

bSr43:
This PR fixes the behaviour of the **indexOfObject:inSortedRange:options:usingComparator:** method when searching for a non-existing element, with either the **NSBinarySearchingFirstEqual** or **NSBinarySearchingLastEqual** option set.
The previous implementation was throwing an exception in that case.
This one returns NSNotFound, exactly like OS X, and iOS.

### DIFF
--- a/Frameworks/Foundation/NSArray.mm
+++ b/Frameworks/Foundation/NSArray.mm
@@ -665,6 +665,10 @@ static CFComparisonResult _CFComparatorFunctionFromComparator(const void* val1, 
                                             _CFComparatorFunctionFromComparator,
                                             &comparator);
 
+    if (index >= [self count] && (options & NSBinarySearchingInsertionIndex)) {
+        return index;
+    }
+
     if (([self count] > 0) && ([[self objectAtIndex:index] isEqual:obj])) {
         if (options & NSBinarySearchingFirstEqual) {
             while (NSLocationInRange(index - 1, range)) {
--- a/Frameworks/Foundation/NSArray.mm
+++ b/Frameworks/Foundation/NSArray.mm
@@ -654,9 +654,11 @@ static CFComparisonResult _CFComparatorFunctionFromComparator(const void* val1, 
                     format:@"Both NSBinarySearchingFirstEqual and NSBinarySearchingLastEqual were specified."];
     }
 
-    if (range.location + range.length > [self count]) {
+    NSUInteger selfCount = [self count];
+
+    if (range.location + range.length > selfCount) {
         [NSException raise:NSInvalidArgumentException
-                    format:@"Range {%d, %d} larger than array of size %d.", range.location, range.length, [self count]];
+                    format:@"Range {%d, %d} larger than array of size %d.", range.location, range.length, selfCount];
     }
 
     NSUInteger index = CFArrayBSearchValues(static_cast<CFArrayRef>(self),
@@ -665,11 +667,11 @@ static CFComparisonResult _CFComparatorFunctionFromComparator(const void* val1, 
                                             _CFComparatorFunctionFromComparator,
                                             &comparator);
 
-    if (index >= [self count] && (options & NSBinarySearchingInsertionIndex)) {
-        return index;
+    if (index >= selfCount && (options & NSBinarySearchingInsertionIndex)) {
+        return selfCount;
     }
 
-    if (([self count] > 0) && ([[self objectAtIndex:index] isEqual:obj])) {
+    if ((selfCount > index) && ([[self objectAtIndex:index] isEqual:obj])) {
         if (options & NSBinarySearchingFirstEqual) {
             while (NSLocationInRange(index - 1, range)) {
                 if ([[self objectAtIndex:index - 1] isEqual:obj]) {
--- a/tests/unittests/Foundation/NSArrayTests.mm
+++ b/tests/unittests/Foundation/NSArrayTests.mm
@@ -263,3 +263,62 @@ TEST(NSArray, ExpandBeyondCapacity) {
 TEST(NSArray, AddingObjects) {
     assertArrayContents([@[@1, @2] arrayByAddingObjectsFromArray:@[@3, @4]], @1, @2, @3, @4, nil);
 }
+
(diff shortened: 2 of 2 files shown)


tests/unittests/Foundation/ReferenceFoundation/TestNSArray.mm
```diff
@@ -150,13 +150,19 @@ static int objectIndexInArray(NSArray* array, int value, int startingFrom, int l
     //        NSBinarySearchingOptions* searchForBoth = NSBinarySearchingFirstEqual | NSBinarySearchingLastEqual
     //        auto _ = objectIndexInArray(array, 9, 0, 13, searchForBoth)
 
-    auto notFound = objectIndexInArray(array, 11, 0, 13);
-    ASSERT_EQ_MSG(notFound, NSNotFound, @"NSArray return NSNotFound if object is not found.");
+    ASSERT_EQ_MSG(objectIndexInArray(array, 11, 0, array.count),
```
bbowman: Curious why this was changed. In general I agree that using magic numbers is bad but this is also changing the code path slightly. With the change we are testing the whole array, whereas before we were testing a subset. I think both are good test cases and we should probably include both (especially for the below case dealing with index of 9).

bbowman:
:clock1:

tests/unittests/Foundation/ReferenceFoundation/TestNSArray.mm
```diff
@@ -150,13 +150,19 @@ static int objectIndexInArray(NSArray* array, int value, int startingFrom, int l
     //        NSBinarySearchingOptions* searchForBoth = NSBinarySearchingFirstEqual | NSBinarySearchingLastEqual
     //        auto _ = objectIndexInArray(array, 9, 0, 13, searchForBoth)
 
-    auto notFound = objectIndexInArray(array, 11, 0, 13);
-    ASSERT_EQ_MSG(notFound, NSNotFound, @"NSArray return NSNotFound if object is not found.");
```
ms-jihua: We'd like to avoid making edits to this file, in order to stay mapped exactly to the original test at [https://github.com/apple/swift-corelibs-foundation/blob/master/TestFoundation/TestNSArray.swift](https://github.com/apple/swift-corelibs-foundation/blob/master/TestFoundation/TestNSArray.swift). Could you test these cases in a separate test case in tests/unittests/Foundation/NSArrayTests.mm, and revert this file to its original state?

tests/unittests/Foundation/ReferenceFoundation/TestNSArray.mm
bSr43: The test on a subrange is done after these 3 tests. In fact, there was a bug in the current version of the algorithm where an exception was thrown only when one was searching in the whole array. The name of the variable was let me think that the developer who wrote the test wanted to search on the whole range for this case, and on a subrange for the other test (which has been kept). Unfortunately, the original test was not able to trigger the issue, because of a wrong index. Now the tests cover both cases.

tests/unittests/Foundation/ReferenceFoundation/TestNSArray.mm
bSr43: Are you sure? Last week someone asked me to do the opposite?

msft-Jeyaram: We try to avoid adding tests/changes to tests/unittests/Foundation/**ReferenceFoundation/**
If you want to add tests they should go in the tests/unittests/**Foundation** folder. The following is the file name convention we used for tests that we created NS*Tests.mm. The imports from swift are named TestNS*.mm

ms-jihua: It looks like he said the same? [https://github.com/Microsoft/WinObjC/pull/557#discussion_r67045139](https://github.com/Microsoft/WinObjC/pull/557#discussion_r67045139)

---
In reply to: [67917856](https://github.com/Microsoft/WinObjC/pull/569#discussion_r67917856) [](ancestors = 67917856)

Frameworks/Foundation/NSArray.mm
```diff
@@ -666,10 +666,10 @@ - (NSUInteger)indexOfObject:(NSObject*)obj
                                             &comparator);
 
     if (index >= [self count] && (options & NSBinarySearchingInsertionIndex)) {
-        return index;
+        return [self count];
```
jofre-ms: Please replace all the calls to [self count] in this method with a cached local variable

tests/unittests/Foundation/ReferenceFoundation/TestNSArray.mm
bbowman: Good catch. 

---
In reply to: [67916770](https://github.com/Microsoft/WinObjC/pull/569#discussion_r67916770) [](ancestors = 67916770)

tests/unittests/Foundation/ReferenceFoundation/TestNSArray.mm
bbowman: What @msft-Jeyaram  and @ms-jihua have said is correct. That said, I definitely agree that it is *not* very clear at the moment. @ms-jihua and I talked, and she will add a header to each of the ReferenceFoundation files to make this hopefully more clear in the future. Sorry for the confusion and thank you for putting up with us :-)

---
In reply to: [67920136](https://github.com/Microsoft/WinObjC/pull/569#discussion_r67920136) [](ancestors = 67920136,67917856)

Frameworks/Foundation/NSArray.mm
bSr43: Sure! Here it is

tests/unittests/Foundation/ReferenceFoundation/TestNSArray.mm
bSr43: you're right, @ms-jihua, I'm feeling stupid (again)... it seems that my VS project is a little bit screwed, and I'm not editing the correct file... I'll commit the fix ASAP!

bSr43:
@ms-jihua the tests have been moved.

rajsesh-msft:
:shipit:

PR #586, opened 2016-06-26
Implementation of the NSPointerArray class

bSr43:
This is a proposal for the implementation of the NSPointerArray class.

### DIFF
--- a/Frameworks/Foundation/NSPointerArray.mm
+++ b/Frameworks/Foundation/NSPointerArray.mm
@@ -14,146 +14,618 @@
 //
 //******************************************************************************
 
-#import <StubReturn.h>
+#import <Foundation/NSCoder.h>
+#import <Foundation/NSException.h>
+#import <Foundation/NSMutableArray.h>
 #import <Foundation/NSPointerArray.h>
+#import <NSPointerFunctionsConcrete.h>
 
-@implementation NSPointerArray
+static NSUInteger sizeFunction(const void* item) {
+    return static_cast<NSUInteger>(sizeof(void*));
+}
+
+@implementation NSPointerArray {
+    // The number of allocated elements.
+    // This implementation tries to allocate space for the pointers ahead of time, in order
+    // to avoid allocation each time one inserts a pointer in the array. This variable
+    // reflects the size of the allocated array.
+    NSUInteger _allocated;
+    // The number of elements in the NSPointerArray.
+    // This value should always be less, or equal than _allocated variable.
+    NSUInteger _count;
+    // A C array containing all pointers of the NSPointerArray.
+    void** _container;
+};
+
++ (id)allocWithZone:(NSZone*)zone {
+    if (self != [NSPointerArray class]) {
+        return nil;
+    }
+    return [super allocWithZone:zone];
+}
+
+- (instancetype)init {
+    if (self = [super init]) {
+        _allocated = 0;
+        _count = 0;
+        _container = nullptr;
+        _pointerFunctions = nil;
+    }
+    return self;
+}
+
+- (void)dealloc {
+    // Release each non-null pointer in the array, according to the current policy.
+    auto relinquish = self.pointerFunctions.relinquishFunction;
+    if (relinquish) {
+        for (NSUInteger i = 0; i < _count; i++) {
+            void* ptr = _container[i];
+            if (ptr != nullptr) {
+                relinquish(ptr, sizeFunction);
+            }
+        }
+    }
+    free(_container);
+    [_pointerFunctions release];
+    [super dealloc];
+}
+
+// Like the reference platform, we don't handle the case of integer personality.
+static void checkPointerFunctionsPersonality(NSPointerFunctions* pointerFunctions) {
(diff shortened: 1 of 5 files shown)


Frameworks/Foundation/NSPointerArray.mm
```diff
@@ -14,146 +14,380 @@
...
+    if (relinquish) {
+        for (NSUInteger i = 0; i < _count; i++) {
+            void* ptr = _container[i];
+            if (ptr != nullptr) {
+                relinquish(ptr, sizeFunction);
+            }
+        }
+    }
+    free(_container);
+    _pointerFunctions = nil;
```
mnithish: Also needs to be released before setting it to nil

Frameworks/Foundation/NSPointerArray.mm
```diff
@@ -14,146 +14,380 @@
...
+    if (self = [super init]) {
+        _allocated = 0;
+        _count = 0;
+        _container = nullptr;
+        _pointerFunctions = nil;
+    }
+    return self;
+}
+
+- (void)dealloc {
```
mnithish: It will be nice to override allocWithZone to prevent this class from being sub classed

Frameworks/Foundation/NSPointerArray.mm
```diff
@@ -14,146 +14,380 @@
...
+
+/**
+ @Status Interoperable
 */
 - (void*)pointerAtIndex:(NSUInteger)index {
-    UNIMPLEMENTED();
-    return StubReturn();
+    if (index < _count) {
+        return _container[index];
+    } else {
```
msft-Jeyaram: nit: else is not needed

msft-Jeyaram: nit: else is not needed

Frameworks/Foundation/NSPointerArray.mm
```diff
@@ -14,146 +14,380 @@
...
 }
 
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 - (void)addPointer:(void*)pointer {
-    UNIMPLEMENTED();
+    assert(_count <= _allocated && "Inconsistent internal state");
```
msft-Jeyaram: we are moving away form using just asserts as they do not give much information.
you can use one of the THROW macros in result.h or use an exception with NSInternalInconsistencyException.
Also could you verify the behavior on reference platform?

Frameworks/Foundation/NSPointerArray.mm
```diff
@@ -14,146 +14,380 @@
...
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 - (void)compact {
-    UNIMPLEMENTED();
+    for (NSUInteger i = 0; i < _count; i++) {
+        if (_container[i] == nullptr) {
+            memmove(_container + i, _container + i + 1, sizeof(void*) * (_count - i - 1));
```
mnithish: This can be done in O(n) time complexity by keeping track of count of non-null elements seen so far -

int index = 0;
for (NSUInteger i = 0; i < _count; i++) {
        if (_container[i] != nullptr) {
           if (index != i) {
               _container[index] = _container[i];
           }
           index++;
        }
}

mnithish:
:clock1:

Frameworks/Foundation/NSPointerArray.mm
bSr43: OK, done!

Frameworks/Foundation/NSPointerArray.mm
bSr43: This bad habit of using ARC :)
Fixed!

Frameworks/Foundation/NSPointerArray.mm
bSr43: Yes, indeed. I removed the "else" statement.

Frameworks/Foundation/NSPointerArray.mm
bSr43: I replaced with an exception, but I'm not sure about what you mean by checking on the reference platform: this is just a test to make sure that my implementation is not in a bad state, and I don't know how to reproduce such a state on OS X for instance.

Frameworks/Foundation/NSPointerArray.mm
bSr43: You're right. I changed the implementation.

Frameworks/Foundation/NSPointerArray.mm
```diff
@@ -14,146 +14,396 @@
...
+    if (state->extra[0] > _count) {
+        return 0;
+    }
+
+    NSUInteger slice = MIN(len, _count - state->extra[0]);
+
+    state->itemsPtr = reinterpret_cast<id*>(_container + state->extra[0]);
+    state->extra[0] += slice;
+
+    return slice;
```
jaredhms: Please add comments throughout the file.

bSr43: I added as much comments as possible.
Also, I changed the behaviour of the "isEqual:" and "hash" method in order to match the behaviour of the reference platform.

mnithish:
@bSr43 This is now merged. Thanks for your contribution!

PR #587, opened 2016-06-27
Implementation of CGPathContainsPoint API

jparmentervf:
Resolves on issue in Issue #525. CGPathContainsPoint implementation should
be interoperable with this pull request. Also updates test project paths
for relative paths as per request from MS.

Fix part of Issue 525


Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -624,12 +630,45 @@ void CGPathApply(CGPathRef path, void* info, CGPathApplierFunction function) {
...
-    UNIMPLEMENTED();
-    return StubReturn();
+    if (m) {
+        point = CGPointApplyAffineTransform(point, *m);
+    }
+    // check if the point is outside this box already, if it is, return false
+    CGRect boundingBox = CGPathGetBoundingBox(path);
+
+    if ((boundingBox.origin.x > point.x) || (boundingBox.origin.y > point.y) ||
+        ((boundingBox.origin.x + boundingBox.size.width) < point.x) || ((boundingBox.origin.y + boundingBox.size.height) < point.y)) {
```
bviglietta: This can be simplified somewhat:

if (!CGRectContainsPoint(boundingBox, point))

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -624,12 +630,45 @@ void CGPathApply(CGPathRef path, void* info, CGPathApplierFunction function) {
...
+        CGContextFillPath(context);
+    }
+
+    CGImageRef image = context->Backing()->DestImage();
+
+    float r = 0.0f;
+    float g = 0.0f;
+    float b = 0.0f;
+    float a = 0.0f;
+    image->Backing()->GetPixel(point.x, point.y, r, g, b, a);
```
bviglietta: The Cairo library that CoreGraphics is built on already has a function to do this test: cairo_in_fill. It might be more natural, rather than doing the fill and the color test, to add something like _CGContextIsPointInPath, which itself calls through to cairo_in_fill.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -624,12 +630,45 @@ void CGPathApply(CGPathRef path, void* info, CGPathApplierFunction function) {
...
+
+    CGContextRelease(context);
+
+    // the filled path will be white, so if any component is greater than 0, we have a point inside the path
```
bviglietta: gray

bviglietta:
:shipit:

msft-Jeyaram:
:shipit:

Frameworks/CoreGraphics/CGPath.mm
jparmentervf: Updated.

On Jun 28, 2016, at 1:07 PM, Ben Viglietta <notifications@github.com<mailto:notifications@github.com>> wrote:


In Frameworks/CoreGraphics/CGPath.mm<https://github.com/Microsoft/WinObjC/pull/587#discussion_r68832605>:

>   @Notes
>  */
>  bool CGPathContainsPoint(CGPathRef path, const CGAffineTransform* m, CGPoint point, bool eoFill) {
> -    UNIMPLEMENTED();
> -    return StubReturn();
> +    if (m) {
> +        point = CGPointApplyAffineTransform(point, *m);
> +    }
> +    // check if the point is outside this box already, if it is, return false
> +    CGRect boundingBox = CGPathGetBoundingBox(path);
> +
> +    if ((boundingBox.origin.x > point.x) || (boundingBox.origin.y > point.y) ||
> +        ((boundingBox.origin.x + boundingBox.size.width) < point.x) || ((boundingBox.origin.y + boundingBox.size.height) < point.y)) {


This can be simplified somewhat:

if (!CGRectContainsPoint(boundingBox, point))

Frameworks/CoreGraphics/CGPath.mm
jparmentervf: This raises an interesting question - it seems to me implementations should try to keep to the Obj-C(++) interface that comes from Apple’s APIs as much as possible. If the Cairo implementation ever changed or was updated, this would mean those implementation changes would only have to happen in the original implementation. However - a case could also be said that this could be more efficient (though compiler optimizations may make that a moot point). What are you feelings about that?
Josh

On Jun 28, 2016, at 1:07 PM, Ben Viglietta <notifications@github.com<mailto:notifications@github.com>> wrote:


In Frameworks/CoreGraphics/CGPath.mm<https://github.com/Microsoft/WinObjC/pull/587#discussion_r68832635>:

> +    CGContextSetGrayFillColor(context, 1.0, 1.0);
> +    CGContextAddPath(context, path);
> +
> +    if (eoFill) {
> +        CGContextEOFillPath(context);
> +    } else {
> +        CGContextFillPath(context);
> +    }
> +
> +    CGImageRef image = context->Backing()->DestImage();
> +
> +    float r = 0.0f;
> +    float g = 0.0f;
> +    float b = 0.0f;
> +    float a = 0.0f;
> +    image->Backing()->GetPixel(point.x, point.y, r, g, b, a);


The Cairo library that CoreGraphics is built on already has a function to do this test: cairo_in_fill. It might be more natural, rather than doing the fill and the color test, to add something like _CGContextIsPointInPath, which itself calls through to cairo_in_fill.

bviglietta: Oh, I agree absolutely that we should stick to Apple's API on the surface. I don't propose changing the interface to CGPathContainsPoint. What I was proposing was that, after we create the CGContext here, we delegate the rest of the logic to the CGContext object (via my hypothetical _CGContextIsPointInPath function). Since our CGContext objects are implemented on top of Cairo, the CGContext object could then turn around and call cairo_in_fill.

I doubt it would make a huge difference one way or the other, but if there's already a dedicated function for this, I figure it's probably the best route.

---
In reply to: [68843448](https://github.com/Microsoft/WinObjC/pull/587#discussion_r68843448) [](ancestors = 68843448)

rajsesh-msft:

Frameworks/CoreGraphics/CGPath.mm
jaredhms: Agree with Ben; if cairo already provides this functionality, we should probably leverage it.  Can you please add a private _CGContextIsPointInPath function to our CGContextRef implementation that turns around and calls cairo_in_fill instead of hand-rolling this code?

---
In reply to: [68853001](https://github.com/Microsoft/WinObjC/pull/587#discussion_r68853001) [](ancestors = 68853001,68843448)

samples/CGCatalog/CGCatalog/Samples/CGCCGPathContainsPointViewController.m
```diff
@@ -0,0 +1,174 @@
...
+}
+
+- (void)didReceiveMemoryWarning {
+    [super didReceiveMemoryWarning];
+    // Dispose of any resources that can be recreated.
+}
+
+- (void)observeValueForKeyPath:(NSString *)keyPath ofObject:(id)object change:(NSDictionary *)change context:(void *)context
+{
+    if( context == gKVOContext && [keyPath isEqualToString: @"testResult"] )
```
jaredhms: please apply clang-formatting to your changes

Frameworks/CoreGraphics/CGPath.mm
jparmentervf: @jaredhms - I implemented the below in CGContextCairo (see code sample below), and both the CGCatalog sample app and my unit tests fail in the rounded corners. I noticed in the cairo documentation that clipping is not counted, and that may be the problem here. Do you prefer having the current PR implementation that mimics iOS more closely? Or do you prefer the behavior cairo's implementation? 

`bool CGContextCairo::CGContextIsPointInPath(bool eoFill, float x, float y) {
    ObtainLock();

    LOCK_CAIRO();
    if(eoFill){
        cairo_set_fill_rule(_drawContext, CAIRO_FILL_RULE_WINDING);
        } else {
        cairo_set_fill_rule(_drawContext, CAIRO_FILL_RULE_EVEN_ODD);
        }
    bool returnValue = cairo_in_fill(_drawContext, x, y);
    UNLOCK_CAIRO();
    return returnValue;
}`

jparmentervf: @jaredhms - perhaps disregard the note above. A fresh set of eyes this morning sees some other problems that may be causing this behavior. I'm going to dig a little more and will post an update soon.

bviglietta: We'd want to mimic iOS's behavior, but I'm surprised if cairo_fill and cairo_in_fill don't compute the same fill region. Is there a current transformation matrix that you need to apply before calling cairo_in_fill?

---
In reply to: [69068519](https://github.com/Microsoft/WinObjC/pull/587#discussion_r69068519) [](ancestors = 69068519)

jparmentervf: @bviglietta - No - transforms are handled earlier in the method before any bounds checking at all occurs. This allows for a quick return without having to check further if the point is obviously out of bounds for the path's bounding rect. 
What I found this morning when I looked again was that, for the original implementation that I sent the PR for (where I was checking pixel color), I was setting a context fill color. If we are able to use the cairo functions that shouldn't be needed any more. I removed that, and I think the cairo functions are now working as they should. Still testing some edge cases and reviewing my new implementation though.

jaredhms: Great; let me know when it's ready for another review.

---
In reply to: [69164416](https://github.com/Microsoft/WinObjC/pull/587#discussion_r69164416) [](ancestors = 69164416)

jparmentervf: @jaredhms - I have rebased to master and pushed updates to the PR. Please let me know if there is anything else you would like to see.

adein:
@rajsesh-msft Any update on this PR?  It's ready for review again.

rajsesh-msft:
in CI build.

PR #588, opened 2016-06-28
VSO 8015441: Handle NSOperations and GCD main queue post main runloop wait stage

mnithish:
This is a regression that was introduced when we moved off fibers. This does not impact functionality but will cause slight performance regression as the run loop will be woken a bit too frequently to service these queues when they have data to be serviced.


rajsesh-msft:

rajsesh-msft:
@msft-Jeyaram is added to the review. #Closed

rajsesh-msft:
:shipit:

rajsesh-msft:

jaredhms:
:shipit:

msft-Jeyaram:
:shipit:

Issue #524, opened 2016-06-03
implement Core graphics transparency APIs


rajsesh-msft:
Specifically these stubs

- [x] CGContextBeginTransparencyLayer
- [x] CGContextBeginTransparencyLayerWithRect
- [x] CGContextEndTransparencyLayer
- [x] CGContextSetPatternPhase

adein:
@rajsesh-msft These items are finished and the issue can be closed.

PR #596, opened 2016-06-30
Implemented CGContextSetPatternPhase

adein:
Implemented CGContextSetPatternPhase using the translation portion of the
pattern's transform matrix.  Also contains a fix for a missing retain in CGContextSetFillPattern.

Implements last item in Issue #524


jaredhms:
:shipit:

jaredhms:
In CI build.

PR #600, opened 2016-06-30
Fix crash when posting notification to notification center from dealloc

### LINKED ISSUE #577: Posting NSNotification during dealloc causes crash.
If an object posts a notification during dealloc passing `self` to the `object` argument it causes a crash.

This might seem an odd thing to do :-) but it turns out some of our code during dealloc and cleanup generates a few model change notifications, so thats why we experience this crash.

Example code: 

```
@implementation TestObjectA

- (instancetype)init
{
	if (self = [super init]) {
	}

	return self;
}

- (void)dealloc
{
	[[NSNotificationCenter defaultCenter] postNotificationName:@"MOOSENOTIFICATION" object:self];
}

@end
```

```
TestObjectA *object = [[TestObjectA alloc] init];

// Let ARC dealloc object on out of scope
```

rajsesh-msft:
NSNotificationCenter was using an auto released NSNotification object, which was in turn holding a reference to an object that is going away.  This change removes the usage of autorelease to make the api compatible with reference platform.

Fixes #577

### DIFF
--- a/Frameworks/Foundation/NSNotification.mm
+++ b/Frameworks/Foundation/NSNotification.mm
@@ -27,26 +27,14 @@
  @Status Interoperable
 */
 + (NSNotification*)notificationWithName:(NSString*)name object:(id)obj {
-    NSNotification* ret = [self alloc];
-
-    ret->notificationName = [name retain];
-    ret->notificationObj = [obj retain];
-    ret->userInfo = nil;
-
-    return [ret autorelease];
+    return [self notificationWithName:name object:obj userInfo:nil];
 }
 
 /**
  @Status Interoperable
 */
 + (NSNotification*)notificationWithName:(NSString*)name object:(id)obj userInfo:(NSDictionary*)info {
-    NSNotification* ret = [self alloc];
-
-    ret->notificationName = [name retain];
-    ret->notificationObj = [obj retain];
-    ret->userInfo = [info retain];
-
-    return [ret autorelease];
+    return [[[self alloc] initWithName:name object:obj userInfo:info] autorelease];
 }
 
 /**
@@ -74,8 +62,8 @@
  @Status Interoperable
 */
 - (void)dealloc {
-    [notificationName release];
     [notificationObj release];
+    [notificationName release];
     [userInfo release];
 
     [super dealloc];
@@ -85,9 +73,13 @@
  @Status Stub
  @Notes
 */
-- (instancetype)initWithName:(NSString*)aName object:(id)object userInfo:(NSDictionary*)userInfo {
-    UNIMPLEMENTED();
-    return StubReturn();
+- (instancetype)initWithName:(NSString*)aName object:(id)object userInfo:(NSDictionary*)info {
+    if ((self = [super init]) != nil) {
+        notificationObj = [object retain];
+        notificationName = [aName copy];
+        userInfo = [userInfo retain];
+    }
+    return self;
 }
 
 /**
--- a/Frameworks/Foundation/NSNotification.mm
+++ b/Frameworks/Foundation/NSNotification.mm
@@ -17,11 +17,7 @@
 #import <StubReturn.h>
 #import <Foundation/NSNotification.h>
 
-@implementation NSNotification {
-    NSString* notificationName;
-    id notificationObj;
-    id userInfo;
-}
+@implementation NSNotification
 
 /**
  @Status Interoperable
@@ -40,31 +36,10 @@
 /**
  @Status Interoperable
 */
-- (id)object {
-    return notificationObj;
-}
-
-/**
- @Status Interoperable
-*/
-- (NSString*)name {
-    return notificationName;
-}
-
-/**
- @Status Interoperable
-*/
-- (NSDictionary*)userInfo {
-    return userInfo;
-}
-
-/**
- @Status Interoperable
-*/
 - (void)dealloc {
-    [notificationObj release];
-    [notificationName release];
-    [userInfo release];
+    [_object release];
+    [_name release];
+    [_userInfo release];
 
     [super dealloc];
 }
@@ -75,9 +50,9 @@
 */
 - (instancetype)initWithName:(NSString*)aName object:(id)object userInfo:(NSDictionary*)info {
     if ((self = [super init]) != nil) {
-        notificationObj = [object retain];
-        notificationName = [aName copy];
-        userInfo = [userInfo retain];
+        _object = [object retain];
+        _name = [aName copy];
+        _userInfo = [info retain];
     }
     return self;
 }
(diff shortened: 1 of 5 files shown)


rajsesh-msft:

rajsesh-msft:

Frameworks/Foundation/NSNotification.mm
```diff
@@ -85,9 +73,13 @@ - (void)dealloc {
  @Status Stub
  @Notes
 */
-- (instancetype)initWithName:(NSString*)aName object:(id)object userInfo:(NSDictionary*)userInfo {
-    UNIMPLEMENTED();
-    return StubReturn();
+- (instancetype)initWithName:(NSString*)aName object:(id)object userInfo:(NSDictionary*)info {
+    if ((self = [super init]) != nil) {
```
msft-Jeyaram: nit: not needed

msft-Jeyaram:
>     id userInfo;

nit: could we just wrap them in StrongId? 

---
Refers to: Frameworks/Foundation/NSNotification.mm:23 in 9291c4e. [](commit_id = 9291c4ea7d40c429dd9db7728587d1b95fe07fee, deletion_comment = False)

Frameworks/Foundation/NSNotification.mm
```diff
@@ -74,8 +62,8 @@ - (NSDictionary*)userInfo {
  @Status Interoperable
 */
 - (void)dealloc {
-    [notificationName release];
     [notificationObj release];
+    [notificationName release];
```
mnithish: Recommend using StrongId<> so these can be removed

mnithish:
:shipit:

msft-Jeyaram:
:shipit:

Frameworks/Foundation/NSNotification.mm
```diff
@@ -85,9 +73,13 @@ - (void)dealloc {
...
  @Notes
 */
-- (instancetype)initWithName:(NSString*)aName object:(id)object userInfo:(NSDictionary*)userInfo {
-    UNIMPLEMENTED();
-    return StubReturn();
+- (instancetype)initWithName:(NSString*)aName object:(id)object userInfo:(NSDictionary*)info {
+    if ((self = [super init]) != nil) {
+        notificationObj = [object retain];
+        notificationName = [aName copy];
+        userInfo = [userInfo retain];
```
rajsesh-msft: bug bug!

rajsesh-msft:
>     id userInfo;

I'd like to go in the other direction instead and just use the storage provided by the property declaration. removes a bunch of unnecessary code as well.

---
In reply to: [229817516](https://github.com/Microsoft/WinObjC/pull/600#issuecomment-229817516) [](ancestors = 229817516)

---
Refers to: Frameworks/Foundation/NSNotification.mm:23 in 9291c4e. [](commit_id = 9291c4ea7d40c429dd9db7728587d1b95fe07fee, deletion_comment = False)

mnithish:
>     id userInfo;

That sounds like a better idea. Let property do the memory management

---
In reply to: [229832711](https://github.com/Microsoft/WinObjC/pull/600#issuecomment-229832711) [](ancestors = 229832711,229817516)

---
Refers to: Frameworks/Foundation/NSNotification.mm:23 in 9291c4e. [](commit_id = 9291c4ea7d40c429dd9db7728587d1b95fe07fee, deletion_comment = False)

Frameworks/Foundation/NSNotification.mm
msft-Jeyaram: info :D

---
In reply to: [69236021](https://github.com/Microsoft/WinObjC/pull/600#discussion_r69236021) [](ancestors = 69236021)

msft-Jeyaram:
>     id userInfo;

although you can still copy it and attach to it (so you aren't getting a retain), at the end it would give the auto-autorelease mechanism.
Given we have these properties backed by internal ivars, we could ease the pain and use StrongId (which comes with it's own shenanigans). 

Your direction makes sense, we can get rid of the these and directly set the properties. 

---
In reply to: [229856697](https://github.com/Microsoft/WinObjC/pull/600#issuecomment-229856697) [](ancestors = 229856697,229832711,229817516)

---
Refers to: Frameworks/Foundation/NSNotification.mm:23 in 9291c4e. [](commit_id = 9291c4ea7d40c429dd9db7728587d1b95fe07fee, deletion_comment = False)

Frameworks/Foundation/NSNotificationCenter.mm
```diff
@@ -79,8 +79,9 @@ - (instancetype)init {
  @Status Interoperable
 */
 - (void)postNotificationName:(NSString*)name object:(NSObject*)sender {
-    NSNotification* notification = [NSNotification notificationWithName:name object:sender];
+    NSNotification* notification = [[NSNotification alloc] initWithName:name object:sender userInfo: nil];
```
msft-Jeyaram: format issue

Frameworks/Foundation/NSNotification.mm
```diff
@@ -85,9 +48,13 @@ - (void)dealloc {
...
  @Notes
 */
-- (instancetype)initWithName:(NSString*)aName object:(id)object userInfo:(NSDictionary*)userInfo {
-    UNIMPLEMENTED();
-    return StubReturn();
+- (instancetype)initWithName:(NSString*)aName object:(id)object userInfo:(NSDictionary*)info {
+    if ((self = [super init]) != nil) {
+        _object = [object retain];
+        _name = [aName copy];
+        _userInfo = [info retain];
```
mnithish: Should be copy

Frameworks/Foundation/NSNotification.mm
```diff
@@ -85,9 +48,13 @@ - (void)dealloc {
  @Status Stub
```
MSFTFox: Is this still a stubbed method?

tests/unittests/Foundation/NSNotificationTests.mm
```diff
@@ -0,0 +1,32 @@
...
+@implementation TestObjectA
+- (void)dealloc {
+    [[NSNotificationCenter defaultCenter] postNotificationName:@"MOOSENOTIFICATION" object:self];
+    [super dealloc];
+}
+
+@end
+
+TEST(NSNotificationCenter, PostNotificationFromDealloc) {
+    TestObjectA* obj = [[TestObjectA new] autorelease];
```
mnithish: nit: do you want to perform this test synchronously? i.e. allocated and free NSNotificationCenter within the test and not catch it as part of the test clean-up

rajsesh-msft: I have tested it with a release, but this is the test code from the issue report, so i decided to leave the autorelease version in.

---
In reply to: [69334437](https://github.com/Microsoft/WinObjC/pull/600#discussion_r69334437) [](ancestors = 69334437)

PR #602, opened 2016-07-01
Fix #580, add support for concurrent NSOperation subclass

### LINKED ISSUE #580: Add support for custom concurrent NSOperation subclasses.
The current implementation of NSOperation/Queue only supports subclasses that override `-(void)main`
 and does not work with custom concurrent NSOperation subclasses that override `-(void)start`.

Take the following example NSOperation subclass which wraps a couple of async operations. The way this should work is when the operation is done,  it's complete block is executed, it's removed from it's queue, and any dependant operations are lined up for execution. This operation does not executed as expected with the current implementation.

```
@interface MyOperation : NSOperation

@property (assign, getter = isExecuting) BOOL executing;
@property (assign, getter = isFinished, readonly) BOOL finished;

@end

@implementation MyOperation

@synthesize executing = _executing;
@synthesize finished = _finished;

- (void)setExecuting:(BOOL)executing
{
	[self willChangeValueForKey:@"isExecuting"];
	[self willChangeValueForKey:@"isFinished"];
	
	_executing = executing;
	_finished = !executing;
	
	[self didChangeValueForKey:@"isExecuting"];
	[self didChangeValueForKey:@"isFinished"];
}

- (BOOL)isExecuting
{
	return _executing;
}

- (BOOL)isFinished
{
	return _finished;
}

- (void)start
{	
	if (self.isCancelled)
		return;
	
	self.executing = YES;
	[self doSomething];
}

- (void)doSomething
{
	// Do some async task.
	dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_LOW, 0), ^{
        
		// Do another async task
		dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_LOW, 0), ^{
			// Do some async task.
			self.executing = NO;
		});
    });
}

@end
```

This is an important feature of NSOperation/Queue as it allows you to group multiple async tasks into a single operation. It would be great if support for this could be added.

jofre-ms:
This change redesigns NSOperation for subclassability as documented for the reference platform, and adds a test to verify that the subclass no longer deadlocks in waitUntilFinished.

### DIFF
--- a/Frameworks/Foundation/NSOperation.mm
+++ b/Frameworks/Foundation/NSOperation.mm
@@ -41,17 +41,20 @@ OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 
 @implementation NSOperation {
     StrongId<NSMutableArray> _dependencies;
-    void (^_completionBlock)(void);
+    StrongId<void(^)()> _completionBlock;
     std::condition_variable_any _finishCondition;
     std::recursive_mutex _finishLock;
     std::recursive_mutex _dependenciesLock;
     std::recursive_mutex _completionBlockLock;
+    std::recursive_mutex _readyLock;
 }
 
 @synthesize cancelled = _cancelled;
 @synthesize executing = _executing;
 @synthesize finished = _finished;
 @synthesize ready = _ready;
+
+static const NSString* NSOperationContext = @"context";
 
 /**
  @Status Interoperable
@@ -62,11 +65,12 @@ OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 }
 
 - (void)_checkReady {
+    std::lock_guard<std::recursive_mutex> lock(_readyLock);
     bool newReady = YES;
 
     // If cancelled, skip this logic and set _ready to YES. Otherwise check dependencies.
     if (![self isCancelled]) {
-        _dependenciesLock.lock();
+        std::lock_guard<std::recursive_mutex> lock(_dependenciesLock);
         int count = [_dependencies count];
 
         for (int i = 0; i < count; i++) {
@@ -76,8 +80,6 @@ OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
                 newReady = NO;
             }
         }
-
-        _dependenciesLock.unlock();
     }
 
     if (_ready != newReady) {
@@ -88,7 +90,7 @@ OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 }
 
 - (void)observeValueForKeyPath:(NSString*)keyPath ofObject:(id)object change:(NSDictionary*)change context:(void*)context {
-    if ([keyPath isEqualToString:@"isFinished"]) {
+    if (context == (void*)NSOperationContext) {
         if (object == self) {
             _finishLock.lock();
             _finishCondition.notify_all();
@@ -103,7 +105,7 @@ OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
     if (self = [super init]) {
         _dependencies.attach([NSMutableArray new]);
         _ready = YES;
-        [self addObserver:self forKeyPath:@"isFinished" options:0 context:NULL];
+        [self addObserver:self forKeyPath:@"isFinished" options:0 context:(void*)NSOperationContext];
     }
 
     return self;
@@ -113,6 +115,7 @@ OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
  @Status Interoperable
 */
 - (BOOL)isReady {
+    std::lock_guard<std::recursive_mutex> lock(_readyLock);
     return _ready;
 }
 
@@ -120,11 +123,10 @@ OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
  @Status Interoperable
 */
 - (void)addDependency:(id)operation {
+    std::lock_guard<std::recursive_mutex> lock(_dependenciesLock);
     [self willChangeValueForKey:@"dependencies"];
-    _dependenciesLock.lock();
     [_dependencies addObject:operation];
-    [operation addObserver:self forKeyPath:@"isFinished" options:0 context:NULL];
-    _dependenciesLock.unlock();
+    [operation addObserver:self forKeyPath:@"isFinished" options:0 context:(void*)NSOperationContext];
     [self didChangeValueForKey:@"dependencies"];
     [self _checkReady];
 }
@@ -133,13 +135,15 @@ OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
  @Status Interoperable
 */
 - (void)removeDependency:(NSOperation*)operation {
-    [self willChangeValueForKey:@"dependencies"];
-    _dependenciesLock.lock();
-    [operation removeObserver:self forKeyPath:@"isFinished" context:NULL];
-    [_dependencies removeObject:operation];
-    _dependenciesLock.unlock();
-    [self didChangeValueForKey:@"dependencies"];
-    [self _checkReady];
+    std::lock_guard<std::recursive_mutex> lock(_dependenciesLock);
+    NSUInteger index = [_dependencies indexOfObject:operation];
+    if (index != NSNotFound) {
+        [self willChangeValueForKey:@"dependencies"];
+        [operation removeObserver:self forKeyPath:@"isFinished" context:(void*)NSOperationContext];
+        [_dependencies removeObject:operation];
+        [self didChangeValueForKey:@"dependencies"];
+        [self _checkReady];
+    }
 }
 
 
@@ -162,11 +166,9 @@ OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
  @Status Interoperable
 */
 - (void)setCompletionBlock:(void (^)(void))block {
+    std::lock_guard<std::recursive_mutex> lock(_completionBlockLock);
     [self willChangeValueForKey:@"completionBlock"];
-    _completionBlockLock.lock();
-    [_completionBlock release];
-    _completionBlock = [block copy];
-    _completionBlockLock.unlock();
+    _completionBlock = block;
     [self didChangeValueForKey:@"completionBlock"];
 }
 
@@ -174,16 +176,15 @@ OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
  @Status Interoperable
 */
 - (void (^)(void))completionBlock {
-    _completionBlockLock.lock();
-    id ret = [[_completionBlock retain] autorelease];
-    _completionBlockLock.unlock();
-    return ret;
+    std::lock_guard<std::recursive_mutex> lock(_completionBlockLock);
+    return [[_completionBlock retain] autorelease];
 }
 
 /**
  @Status Interoperable
 */
 - (BOOL)isCancelled {
+    std::lock_guard<std::recursive_mutex> lock(_finishLock);
     return _cancelled;
 }
 
@@ -276,12 +277,11 @@ OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
  @Status Interoperable
 */
 - (void)cancel {
+    std::lock_guard<std::recursive_mutex> lock(_finishLock);
     if (_cancelled == NO) {
-        _finishLock.lock();
         [self willChangeValueForKey:@"isCancelled"];
         _cancelled = YES;
         [self didChangeValueForKey:@"isCancelled"];
-        _finishLock.unlock();
         [self _checkReady];
     }
 }
@@ -308,9 +308,8 @@ OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
  @Status Interoperable
 */
 - (NSArray*)dependencies {
-    _dependenciesLock.lock();
+    std::lock_guard<std::recursive_mutex> lock(_dependenciesLock);
     NSArray* copy = [[_dependencies copy] autorelease];
-    _dependenciesLock.unlock();
     return copy;
 }
 
@@ -322,12 +321,11 @@ OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
     int count = [_dependencies count];
     for (int i = 0; i < count; i++) {
         id op = [_dependencies objectAtIndex:i];
-        [op removeObserver:self forKeyPath:@"isFinished" context:NULL];
+        [op removeObserver:self forKeyPath:@"isFinished" context:(void*)NSOperationContext];
     }
 
     _dependenciesLock.unlock();
-    [self removeObserver:self forKeyPath:@"isFinished" context:NULL];
-    [_completionBlock release];
+    [self removeObserver:self forKeyPath:@"isFinished" context:(void*)NSOperationContext];
     [super dealloc];
 }
 
--- a/Frameworks/Foundation/NSOperation.mm
+++ b/Frameworks/Foundation/NSOperation.mm
@@ -1,3 +1,22 @@
+/*
+Original Author: Michael Ash on 11/9/08.
+Copyright (c) 2008 Rogue Amoeba Software LLC
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
+documentation files (the "Software"), to deal in the Software without restriction, including without limitation the
+rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit
+persons to whom the Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all copies or substantial portions of the
+Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE
+WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+*/
+
 //******************************************************************************
 //
 // Copyright (c) 2016 Microsoft Corporation. All rights reserved.
@@ -14,24 +33,25 @@
 //
 //******************************************************************************
 
-#include "Starboard.h"
-#include "StubReturn.h"
-#include "Foundation/NSOperation.h"
-
-static const wchar_t* TAG = L"NSOperation";
+#import <Starboard.h>
+#import <StubReturn.h>
+#import <Foundation/NSOperation.h>
+#import <condition_variable>
+#import <mutex>
 
 @implementation NSOperation {
     StrongId<NSMutableArray> _dependencies;
     void (^_completionBlock)(void);
-    pthread_cond_t _finishCondition;
-    pthread_mutex_t _finishLock;
-    pthread_mutex_t _dependenciesLock;
-    pthread_mutex_t _completionBlockLock;
+    std::condition_variable_any _finishCondition;
+    std::recursive_mutex _finishLock;
+    std::recursive_mutex _dependenciesLock;
+    std::recursive_mutex _completionBlockLock;
 }
 
 @synthesize cancelled = _cancelled;
 @synthesize executing = _executing;
 @synthesize finished = _finished;
+@synthesize ready = _ready;
 
 /**
  @Status Interoperable
@@ -41,26 +61,48 @@ static const wchar_t* TAG = L"NSOperation";
     return NO;
 }
 
+- (void)_checkReady {
+    bool newReady = YES;
+
+    // If cancelled, skip this logic and set _ready to YES. Otherwise check dependencies.
+    if (![self isCancelled]) {
+        _dependenciesLock.lock();
+        int count = [_dependencies count];
+
+        for (int i = 0; i < count; i++) {
+            id op = [_dependencies objectAtIndex:i];
+
+            if (![op isFinished]) {
+                newReady = NO;
+            }
+        }
+
+        _dependenciesLock.unlock();
+    }
+
+    if (_ready != newReady) {
+        [self willChangeValueForKey:@"isReady"];
+        _ready = newReady;
+        [self didChangeValueForKey:@"isReady"];
+    }
+}
+
 - (void)observeValueForKeyPath:(NSString*)keyPath ofObject:(id)object change:(NSDictionary*)change context:(void*)context {
     if ([keyPath isEqualToString:@"isFinished"]) {
-        pthread_mutex_lock(&_finishLock);
-        pthread_cond_broadcast(&_finishCondition);
-        pthread_mutex_unlock(&_finishLock);
+        if (object == self) {
+            _finishLock.lock();
+            _finishCondition.notify_all();
+            _finishLock.unlock();
+        } else {
+            [self _checkReady];
+        }
     }
 }
 
 - (id)init {
     if (self = [super init]) {
         _dependencies.attach([NSMutableArray new]);
-        pthread_cond_init(&_finishCondition, 0);
-        pthread_mutexattr_t attr;
-        pthread_mutexattr_init(&attr);
-        pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_RECURSIVE);
-        pthread_mutex_init(&_finishLock, &attr);
-        pthread_mutexattr_destroy(&attr);
-        pthread_mutex_init(&_dependenciesLock, NULL);
-        pthread_mutex_init(&_completionBlockLock, NULL);
-
+        _ready = YES;
         [self addObserver:self forKeyPath:@"isFinished" options:0 context:NULL];
     }
 
@@ -70,44 +112,36 @@ static const wchar_t* TAG = L"NSOperation";
 /**
  @Status Interoperable
 */
+- (BOOL)isReady {
+    return _ready;
+}
+
+/**
+ @Status Interoperable
+*/
 - (void)addDependency:(id)operation {
-    pthread_mutex_lock(&_dependenciesLock);
+    [self willChangeValueForKey:@"dependencies"];
+    _dependenciesLock.lock();
     [_dependencies addObject:operation];
-    pthread_mutex_unlock(&_dependenciesLock);
+    [operation addObserver:self forKeyPath:@"isFinished" options:0 context:NULL];
+    _dependenciesLock.unlock();
+    [self didChangeValueForKey:@"dependencies"];
+    [self _checkReady];
 }
 
 /**
  @Status Interoperable
 */
 - (void)removeDependency:(NSOperation*)operation {
-    pthread_mutex_lock(&_dependenciesLock);
+    [self willChangeValueForKey:@"dependencies"];
+    _dependenciesLock.lock();
+    [operation removeObserver:self forKeyPath:@"isFinished" context:NULL];
     [_dependencies removeObject:operation];
-    pthread_mutex_unlock(&_dependenciesLock);
+    _dependenciesLock.unlock();
+    [self didChangeValueForKey:@"dependencies"];
+    [self _checkReady];
 }
 
-/**
- @Status Interoperable
-*/
-- (BOOL)isReady {
-    if ([self isCancelled]) {
-        return YES;
-    }
-
-    pthread_mutex_lock(&_dependenciesLock);
-    int count = [_dependencies count];
-
-    for (int i = 0; i < count; i++) {
-        id op = [_dependencies objectAtIndex:i];
-
-        if (![op isFinished]) {
-            pthread_mutex_unlock(&_dependenciesLock);
-            return NO;
-        }
-    }
-
-    pthread_mutex_unlock(&_dependenciesLock);
-    return YES;
-}
 
 /**
  @Status Stub
@@ -128,19 +162,21 @@ static const wchar_t* TAG = L"NSOperation";
  @Status Interoperable
 */
 - (void)setCompletionBlock:(void (^)(void))block {
-    pthread_mutex_lock(&_completionBlockLock);
-    [_completionBlock autorelease];
+    [self willChangeValueForKey:@"completionBlock"];
+    _completionBlockLock.lock();
+    [_completionBlock release];
     _completionBlock = [block copy];
-    pthread_mutex_unlock(&_completionBlockLock);
+    _completionBlockLock.unlock();
+    [self didChangeValueForKey:@"completionBlock"];
 }
 
 /**
  @Status Interoperable
 */
 - (void (^)(void))completionBlock {
-    pthread_mutex_lock(&_completionBlockLock);
+    _completionBlockLock.lock();
     id ret = [[_completionBlock retain] autorelease];
-    pthread_mutex_unlock(&_completionBlockLock);
+    _completionBlockLock.unlock();
     return ret;
 }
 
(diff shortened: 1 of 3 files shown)


Frameworks/Foundation/NSOperation.mm
```diff
@@ -1,115 +1,112 @@
...
         pthread_mutexattr_t attr;
-
         pthread_mutexattr_init(&attr);
         pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_RECURSIVE);
-        pthread_mutex_init(&finishLock, &attr);
+        pthread_mutex_init(&_finishLock, &attr);
         pthread_mutexattr_destroy(&attr);
-    }
+        pthread_mutex_init(&_dependenciesLock, NULL);
+        pthread_mutex_init(&_completionBlockLock, NULL);
```
msft-Jeyaram: we talked earlier about using NSLock and yep it's not going to be efficient as just using raw pthreads since it objc_msgSend will add addition overhead.

Although pthreads are fine, we have allot going on here, it's easier to forget to set/destroy something. Since we support C++ on these files, we should go ahead and use std::mutex, recursive mutex, etc. 

This would make the code much cleaner to maintain and read, also avoid pitfalls that we might encounter in the future maintenance.  Going forward we should use them.

Frameworks/Foundation/NSOperation.mm
```diff
@@ -1,115 +1,112 @@
...
-#include <pthread.h>
-#include <string.h>
-struct NSOperationPriv {
-    NSOperationQueuePriority priority;
-    id dependencies;
-    void (^completionBlock)(void);
+@implementation NSOperation {
+    StrongId<NSMutableArray> _dependencies;
+    void (^_completionBlock)(void);
+    pthread_cond_t _finishCondition;
```
mnithish: Can we use NSCondition please?

Frameworks/Foundation/NSOperation.mm
```diff
@@ -1,115 +1,112 @@
...
-    if (priv->dependencies == nil) {
-        priv->dependencies = [[NSMutableArray alloc] init];
+    pthread_mutex_lock(&_dependenciesLock);
+    int count = [_dependencies count];
+
+    for (int i = 0; i < count; i++) {
+        id op = [_dependencies objectAtIndex:i];
+
+        if (![op isFinished]) {
+            pthread_mutex_unlock(&_dependenciesLock);
```
mnithish: let us use std:: functions here and make the code a bit simpler to maintain with RAII

Frameworks/Foundation/NSOperation.mm
```diff
@@ -1,115 +1,112 @@
...
+// THE SOFTWARE.
+//
+//******************************************************************************
 
 #include "Starboard.h"
 #include "StubReturn.h"
 #include "Foundation/NSOperation.h"
-#include "Foundation/NSString.h"
-#include "Foundation/NSMutableArray.h"
-#include "LoggingNative.h"
```
msft-Jeyaram: use angular brackets <>

Frameworks/Foundation/NSOperation.mm
```diff
@@ -1,115 +1,112 @@
...
 
 static const wchar_t* TAG = L"NSOperation";
```
mnithish: nit: not used

msft-Jeyaram:
:clock1:

Frameworks/Foundation/NSOperation.mm
```diff
@@ -131,139 +128,109 @@ - (void)setThreadPriority:(double)priority {
  @Status Interoperable
 */
 - (void)setCompletionBlock:(void (^)(void))block {
-    id oldBlock = priv->completionBlock;
-    priv->completionBlock = [block copy];
-    [oldBlock release];
+    pthread_mutex_lock(&_completionBlockLock);
+    [_completionBlock autorelease];
```
mnithish: Curious: why should this be not release?

Frameworks/Foundation/NSOperation.mm
```diff
@@ -131,139 +128,109 @@ - (void)setThreadPriority:(double)priority {
...
+    }
 
-        [self _setFinished:YES
-            andPerformUnderLock:^{
-                if (execute) {
-                    priv->executing = 0;
-                    [self didChangeValueForKey:@"isExecuting"];
-                }
-            }];
+    THROW_NS_IF(E_INVALIDARG, (_executing || ![self isReady]));
```
mnithish: interesting this would be the behavior on the reference platform

Frameworks/Foundation/NSOperation.mm
jofre-ms: I think it's unnecessary. I'll change back to release.

---
In reply to: [69351630](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69351630) [](ancestors = 69351630)

Frameworks/Foundation/NSOperation.mm
jofre-ms: From the documentation "If the operation is currently executing or is not ready to execute, this method [start] throws an NSInvalidArgumentException exception."

---
In reply to: [69351761](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69351761) [](ancestors = 69351761)

Frameworks/Foundation/NSOperation.mm
```diff
@@ -131,139 +128,109 @@ - (void)setThreadPriority:(double)priority {
...
-        priv->completionBlock();
-        [priv->completionBlock release];
-        priv->completionBlock = nil;
+    pthread_mutex_unlock(&_finishLock);
+    
+    if (completion) {
+        completion();
     }
+
+    [completion release];
```
mnithish: This is not thread safe... theoretically  _completionBlock could get a new value and retain the new value while you are executing these statements

tests/unittests/Foundation/NSOperationTests.mm
```diff
@@ -59,3 +62,102 @@
...
+}
+
+- (BOOL)isFinished
+{
+    return _finished;
+}
+
+- (void)start
+{
+    if (self.isCancelled)
```
mnithish: nit: braces to follow coding convention

tests/unittests/Foundation/NSOperationTests.mm
```diff
@@ -59,3 +62,102 @@
...
+
+@implementation MyOperation
+
+@synthesize executing = _executing;
+@synthesize finished = _finished;
+
+- (void)setExecuting:(BOOL)executing
+{
+    [self willChangeValueForKey:@"isExecuting"];
+    [self willChangeValueForKey:@"isFinished"];
```
mnithish: Add some KVO tests around these?

Frameworks/Foundation/NSOperation.mm
mnithish: Sorry... missed the fact this is a local block. Thanks for explaining John

---
In reply to: [69352631](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69352631) [](ancestors = 69352631)

Frameworks/Foundation/NSOperation.mm
```diff
@@ -1,115 +1,112 @@
...
+    pthread_mutex_t _finishLock;
+    pthread_mutex_t _dependenciesLock;
+    pthread_mutex_t _completionBlockLock;
+}
 
-    pthread_cond_t finishCondition;
-    pthread_mutex_t finishLock;
+@synthesize cancelled = _cancelled;
+@synthesize executing = _executing;
+@synthesize finished = _finished;
```
mnithish: Can we STUB concurrent and asynchronous properties in the header file or implement them if we plan to complete the implementation for this class?

tests/unittests/Foundation/NSOperationTests.mm
```diff
@@ -59,3 +62,102 @@
...
+{
+    // Do some async task.
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_LOW, 0), ^{
+
+        // Do another async task
+        dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_LOW, 0), ^{
+            // Do some async task.
+            self.executing = NO;
+        });
+    });
```
mnithish: Shouldn't these be implemented in -main to test the real usage of this class?

tests/unittests/Foundation/NSOperationTests.mm
```diff
@@ -59,3 +62,102 @@
...
+{
+    return _executing;
+}
+
+- (BOOL)isFinished
+{
+    return _finished;
+}
+
+- (void)start
```
mnithish: Can we also have a test that does not override start but implements main?

Frameworks/Foundation/NSOperation.mm
```diff
@@ -131,139 +128,109 @@ - (void)setThreadPriority:(double)priority {
...
+    if (shouldExecute) {
+        [self willChangeValueForKey:@"isExecuting"];
+        _executing = YES;
+        [self didChangeValueForKey:@"isExecuting"];
+    }
+    pthread_mutex_unlock(&_finishLock);
+    
+    if (shouldExecute) {
+        NSAutoreleasePool* pool = [NSAutoreleasePool new];
+        [self main];
```
mnithish: try-catch? From iOS documentation - "You should never allow exceptions to propagate out of your main method." Might be worth checking the reference platform as they jump around talking about OS X in the same line

tests/unittests/Foundation/NSOperationTests.mm
jofre-ms: I don't think it's necessary to test the test class MyOperation, but I'll add KVO tests for the base class.

---
In reply to: [69353346](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69353346) [](ancestors = 69353346)

Frameworks/Foundation/NSOperation.mm
```diff
@@ -1,115 +1,112 @@
...
+    pthread_mutex_unlock(&_dependenciesLock);
 }
 
 /**
  @Status Interoperable
 */
-- (NSOperationQueuePriority)queuePriority {
-    return priv->priority;
-}
+- (BOOL)isReady {
```
mnithish: From documentation -  "you must generate KVO notifications for the isReady key path whenever the ready state of your operation object changes."

Frameworks/Foundation/NSOperation.mm
mnithish: Ah looks like NSOperation can be called directly without being added to a NSOperationQueue. Interesting...

---
In reply to: [69352263](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69352263) [](ancestors = 69352263,69351761)

tests/unittests/Foundation/NSOperationTests.mm
```diff
@@ -59,3 +62,102 @@
...
+    [queue addOperation:operation];
+
+    [operation waitUntilFinished];
+
+    ASSERT_TRUE([operation isFinished]);
+    ASSERT_FALSE([operation isExecuting]);
+    ASSERT_NO_THROW([operation release]);
+}
+
+TEST(NSOperation, NSOperationMultipleWaiters) {
```
mnithish: Tests around dependencies?

Frameworks/Foundation/NSOperation.mm
jofre-ms: According to the preceding sentence in the documentation, this is the correct behavior. " In OS X v10.6 and later, exceptions are allowed to propagate beyond this [start] method. You should never allow exceptions to propagate out of your main method."

---
In reply to: [69355325](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69355325) [](ancestors = 69355325)

Frameworks/Foundation/NSOperation.mm
```diff
@@ -131,139 +128,109 @@ - (void)setThreadPriority:(double)priority {
...
+    pthread_mutex_lock(&_completionBlockLock);
+    [_completionBlock autorelease];
+    _completionBlock = [block copy];
+    pthread_mutex_unlock(&_completionBlockLock);
 }
 
 /**
  @Status Interoperable
 */
 - (void (^)(void))completionBlock {
```
mnithish: What does it mean to add support to observe this property? According to the document this is an observable property

Frameworks/Foundation/NSOperation.mm
```diff
@@ -277,35 +244,35 @@ - (void)main {
...
+
+    pthread_mutex_unlock(&_finishLock);
 }
 
 /**
  @Status Interoperable
 */
-- (id)dependencies {
-    return priv->dependencies;
+- (NSArray*)dependencies {
```
mnithish: What does it mean to add support to observe this property? According to the document this is an observable property

Frameworks/Foundation/NSOperation.mm
```diff
@@ -131,139 +128,109 @@ - (void)setThreadPriority:(double)priority {
...
 
 /**
  @Status Interoperable
 */
 - (BOOL)isExecuting {
-    return priv->executing != 0;
+    return _executing;
 }
 
 /**
```
mnithish: Implement isAsynchronous and return NO and isConcurrent and return NO

Frameworks/Foundation/NSOperation.mm
```diff
@@ -131,139 +128,109 @@ - (void)setThreadPriority:(double)priority {
...
+    if (didExecute) {
+        _executing = NO;
+        [self didChangeValueForKey:@"isExecuting"];
     }
 
-    if (newValue == 1) {
-        pthread_cond_broadcast(&priv->finishCondition);
-    }
+    pthread_mutex_lock(&_completionBlockLock);
+    void (^completion)(void) = [_completionBlock retain];
```
jofre-ms: "In iOS 8 and later and OS X v10.10 and later, this property is set to nil after the completion block begins executing."

Frameworks/Foundation/NSOperation.mm
jofre-ms: [self willChangeValueForKey:@"dependencies"];
    [self didChangeValueForKey:@"dependencies"];
added to addDependency and removeDependency

---
In reply to: [69357321](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69357321) [](ancestors = 69357321)

mnithish:
:clock1:

Frameworks/Foundation/NSOperation.mm
jofre-ms: needs
    [self willChangeValueForKey:@"completionBlock"];
    [self didChangeValueForKey:@"completionBlock"];


---
In reply to: [69357300](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69357300) [](ancestors = 69357300)

Frameworks/Foundation/NSOperation.mm
jofre-ms: This will need a test as well

---
In reply to: [69355570](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69355570) [](ancestors = 69355570)

Frameworks/Foundation/NSOperation.mm
jofre-ms: queuePriority is used in NSOperationQueue and cannot be stub. I'm adding a "caveat" implementation that always returns NSOperationQueuePriorityNormal for that property. qualityOfService will be STUB and concurrent/asynchronous return NO.

---
In reply to: [69354091](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69354091) [](ancestors = 69354091)

Frameworks/Foundation/NSOperation.mm
```diff
@@ -131,139 +128,109 @@ - (void)setThreadPriority:(double)priority {
...
 }
 
 /**
  @Status Interoperable
 */
 - (void)cancel {
-    if (priv->cancelled == 0) {
-        pthread_mutex_lock(&priv->finishLock);
+    if (_cancelled == NO) {
+        pthread_mutex_lock(&_finishLock);
```
DHowett-MSFT: Is it safe to do this outside of lock?

Frameworks/Foundation/NSOperation.mm
DHowett-MSFT: Depending on how this works, you might have to use `{did|will}ChangeValuesAtIndexes: forKey:`

---
In reply to: [69358159](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69358159) [](ancestors = 69358159,69357321)

Frameworks/Foundation/NSOperation.mm
```diff
@@ -277,35 +244,35 @@ - (void)main {
...
+    return copy;
 }
 
 /**
  @Status Interoperable
 */
 - (void)dealloc {
-    assert(!priv->completionBlock);
-    delete priv;
+    [self removeObserver:self forKeyPath:@"isFinished" context:NULL];
```
DHowett-MSFT: nit: prefer nullptr over NULL.

Frameworks/Foundation/NSOperation.mm
jofre-ms: Actually queuePriority can just be auto-synthesized

---
In reply to: [69614305](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69614305) [](ancestors = 69614305,69354091)

Frameworks/Foundation/NSOperation.mm
```diff
@@ -277,35 +244,35 @@ - (void)main {
...
+    pthread_cond_destroy(&_finishCondition);
+    pthread_mutex_destroy(&_finishLock);
+    pthread_mutex_destroy(&_dependenciesLock);
+    pthread_mutex_destroy(&_completionBlockLock);
+    [_completionBlock release];
```
DHowett-MSFT: `_completionBlock` can be a `StrongId<void(^)()>` if you want to dispense with the formalities.

Frameworks/Foundation/NSOperation.mm
jofre-ms: Discussed offline. This matches the ref platform

---
In reply to: [69615177](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69615177) [](ancestors = 69615177,69358159,69357321)

DHowett-MSFT: It looks like the reference platform dispatches full change notifications and not indexed ones.

---
In reply to: [69615177](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69615177) [](ancestors = 69615177,69358159,69357321)

Frameworks/Foundation/NSOperation.mm
```diff
@@ -77,41 +61,88 @@ + (BOOL)automaticallyNotifiesObserversForKey:(NSString*)key {
...
-+ (id)allocWithZone:(NSZone*)zone {
-    NSOperation* ret = [super allocWithZone:zone];
+- (void)_checkReady {
+    bool newReady = YES;
 
-    ret->priv = new NSOperationPriv();
+    // If cancelled, skip this logic and set _ready to YES. Otherwise check dependencies.
+    if (![self isCancelled]) {
+        _dependenciesLock.lock();
+        int count = [_dependencies count];
```
DHowett-MSFT: nit: prefer std::lock_guard for scope-guarded mutex use

Frameworks/Foundation/NSOperation.mm
```diff
@@ -77,41 +61,88 @@ + (BOOL)automaticallyNotifiesObserversForKey:(NSString*)key {
     return NO;
 }
 
-/**
- @Status Interoperable
-*/
-+ (id)allocWithZone:(NSZone*)zone {
-    NSOperation* ret = [super allocWithZone:zone];
+- (void)_checkReady {
```
DHowett-MSFT: Is this safe for possibly concurrent access?

Frameworks/Foundation/NSOperation.mm
```diff
@@ -77,41 +61,88 @@ + (BOOL)automaticallyNotifiesObserversForKey:(NSString*)key {
...
+    if (_ready != newReady) {
+        [self willChangeValueForKey:@"isReady"];
+        _ready = newReady;
+        [self didChangeValueForKey:@"isReady"];
+    }
+}
+
+- (void)observeValueForKeyPath:(NSString*)keyPath ofObject:(id)object change:(NSDictionary*)change context:(void*)context {
+    if ([keyPath isEqualToString:@"isFinished"]) {
+        if (object == self) {
```
DHowett-MSFT: If you use context on registration for isFinished, you can avoid a string compare.

Frameworks/Foundation/NSOperation.mm
```diff
@@ -77,41 +61,88 @@ + (BOOL)automaticallyNotifiesObserversForKey:(NSString*)key {
...
 /**
  @Status Interoperable
 */
-- (void)setQueuePriority:(NSOperationQueuePriority)priority {
-    priv->priority = priority;
+- (void)removeDependency:(NSOperation*)operation {
+    [self willChangeValueForKey:@"dependencies"];
+    _dependenciesLock.lock();
+    [operation removeObserver:self forKeyPath:@"isFinished" context:NULL];
+    [_dependencies removeObject:operation];
```
DHowett-MSFT: What happens here if you remove a dependency that this operation doesn't have?

Frameworks/Foundation/NSOperation.mm
DHowett-MSFT: If you set op.completionBlock = op.completionBlock, you will likely destroy the block here./

---
In reply to: [69352043](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69352043) [](ancestors = 69352043,69351630)

tests/unittests/Foundation/NSOperationTests.mm
```diff
@@ -59,3 +62,256 @@
...
+@end
+
+TEST(NSOperation, NSOperationConcurrentSubclass) {
+    NSOperationQueue* queue = [[NSOperationQueue new] autorelease];
+
+    NSOperation* operation = [MyConcurrentOperation new];
+
+    [operation setCompletionBlock:^{
+        [operation waitUntilFinished]; // Should not deadlock, but we cannot test this
+        ASSERT_TRUE([operation isFinished]);
```
DHowett-MSFT: We can somewhat test deadlocks -- by having a watchdog thread. In the worst case, the test runs until the watchdog fails it, but then we need to abort the test run completely since we can't recover.

Frameworks/Foundation/NSOperation.mm
DHowett-MSFT: I'm upgrading this from a nit -- we should definitely use `lock_guard` and `unique_lock` everywhere. `std::condition_variable_any` actually can take a `unique_lock`, as well.

---
In reply to: [69623694](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69623694) [](ancestors = 69623694)

DHowett-MSFT:
:clock1:

Frameworks/Foundation/NSOperation.mm
jofre-ms: moved the scope of the lock outside of the if statement

---
In reply to: [69615071](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69615071) [](ancestors = 69615071)

Frameworks/Foundation/NSOperation.mm
jofre-ms: switched to strongId

---
In reply to: [69624034](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69624034) [](ancestors = 69624034,69352043,69351630)

Frameworks/Foundation/NSOperation.mm
jofre-ms: why?

---
In reply to: [69624996](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69624996) [](ancestors = 69624996,69623694)

Frameworks/Foundation/NSOperation.mm
jofre-ms: Needs another lock

---
In reply to: [69623776](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69623776) [](ancestors = 69623776)

Frameworks/Foundation/NSOperation.mm
jofre-ms: On the reference platform, nothing happens. I'll add a test.

---
In reply to: [69623943](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69623943) [](ancestors = 69623943)

Frameworks/Foundation/NSOperation.mm
mnithish: setQueuePriority should spew a warning saying only NSOperationQueuePriorityNormal is supported and have a CAVEAT keyword

---
In reply to: [69615276](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69615276) [](ancestors = 69615276,69614305,69354091)

Frameworks/Foundation/NSOperation.mm
```diff
@@ -131,139 +166,123 @@ - (void)setThreadPriority:(double)priority {
...
-        pthread_cond_broadcast(&priv->finishCondition);
+    if (didExecute) {
+        _executing = NO;
+        [self didChangeValueForKey:@"isExecuting"];
     }
 
-    pthread_mutex_unlock(&priv->finishLock);
+    _completionBlockLock.lock();
+    void (^completion)(void) = [_completionBlock retain];
+    [self setCompletionBlock:nil];
```
mnithish: Question: Does this need to trigger KVO events or should it silently set it to nil?

tests/unittests/Foundation/NSOperationTests.mm
```diff
@@ -59,3 +62,264 @@
...
+    }];
+
+    [queue addOperation:operation];
+    [operation waitUntilFinished];
+
+    ASSERT_TRUE([operation isFinished]);
+    ASSERT_FALSE([operation isExecuting]);
+    ASSERT_TRUE([operation didWork]);
```
mnithish: didWork needs to be reset after every successful test else the following test check is not really valid

Frameworks/Foundation/NSOperation.mm
jofre-ms: I checked the reference platform and it sets silently. I'll make the change here.

---
In reply to: [69660196](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69660196) [](ancestors = 69660196)

Frameworks/Foundation/NSOperation.mm
jofre-ms: The current implementation for queue priority, which is just auto-synthesized, is sufficiently interoperable

---
In reply to: [69659816](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69659816) [](ancestors = 69659816,69615276,69614305,69354091)

Frameworks/Foundation/NSOperation.mm
```diff
@@ -131,139 +166,123 @@ - (void)setThreadPriority:(double)priority {
...
-        id op = [priv->dependencies objectAtIndex:i];
-
-        if (![op isFinished]) {
-            return NO;
-        }
-    }
-    return YES;
+- (BOOL)isCancelled {
+    std::lock_guard<std::recursive_mutex> lock(_finishLock);
+    return _cancelled;
```
DHowett-MSFT: Cancelled is guarded by the finish lock, but finished and executing aren't guarded by anything?

Frameworks/Foundation/NSOperation.mm
```diff
@@ -131,139 +166,123 @@ - (void)setThreadPriority:(double)priority {
...
+    if (shouldExecute) {
+        [self willChangeValueForKey:@"isExecuting"];
+        _executing = YES;
+        [self didChangeValueForKey:@"isExecuting"];
+    }
+    _finishLock.unlock();
+    
+    if (shouldExecute) {
+        NSAutoreleasePool* pool = [NSAutoreleasePool new];
+        [self main];
```
DHowett-MSFT: NSOperatinQueue nigh-unto guarantees that operations run with their own pools -- do we have to make another one?

Frameworks/Foundation/NSOperation.mm
```diff
@@ -131,139 +166,123 @@ - (void)setThreadPriority:(double)priority {
...
-    // is cancelled.
-    if (newValue == 1 && priv->completionBlock) {
-        priv->completionBlock();
-        [priv->completionBlock release];
-        priv->completionBlock = nil;
+    _finishLock.unlock();
+    
+    if (completion) {
+        completion();
     }
```
DHowett-MSFT: Does completion fire even if it did not execute (and was cancelled)?

Frameworks/Foundation/NSOperation.mm
```diff
@@ -277,35 +296,37 @@ - (void)main {
...
 */
 - (void)dealloc {
-    assert(!priv->completionBlock);
-    delete priv;
-    [super dealloc];
-}
+    _dependenciesLock.lock();
+    int count = [_dependencies count];
+    for (int i = 0; i < count; i++) {
+        id op = [_dependencies objectAtIndex:i];
```
DHowett-MSFT: This is likely better written as `for(NSOperation* op in _dependencies)`, as it can be optimized.

tests/unittests/Foundation/NSOperationTests.mm
jofre-ms: I don't think it's worth changing to this design, although it would be better to have a watchdog thread. We can still detect the regression.

---
In reply to: [69624529](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69624529) [](ancestors = 69624529)

Frameworks/Foundation/NSOperation.mm
jofre-ms: Yes

---
In reply to: [69819051](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69819051) [](ancestors = 69819051)

jofre-ms: "A finished operation may finish either because it was cancelled or because it successfully completed its task. You should take that fact into account when writing your block code."

---
In reply to: [69831586](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69831586) [](ancestors = 69831586,69819051)

Frameworks/Foundation/NSOperation.mm
jofre-ms: NSOperation support being invoked "manually" - that is, outside of the Queue by user code. We still need an autorelease pool in that case.

---
In reply to: [69818876](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69818876) [](ancestors = 69818876)

Frameworks/Foundation/NSOperation.mm
jofre-ms: I don't think the lock is needed for the accessor. Removing.

---
In reply to: [69818734](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69818734) [](ancestors = 69818734)

Frameworks/Foundation/NSOperation.mm
```diff
@@ -131,139 +162,122 @@ - (void)setThreadPriority:(double)priority {
...
+    return _executing;
 }
 
 /**
  @Status Interoperable
 */
-- (BOOL)isExecuting {
-    return priv->executing != 0;
+- (BOOL)isConcurrent {
+    return NO;
```
mnithish: Just noticed the changelist subject reads "add support for _concurrent_ NSOperation subclass"

DHowett-MSFT: It's the _subclass_ that's concurrent; NSOperation has to have the machinery to support both even though it is itself nonconcurrent.

Frameworks/Foundation/NSOperation.mm
```diff
@@ -131,139 +162,122 @@ - (void)setThreadPriority:(double)priority {
...
-        pthread_cond_broadcast(&priv->finishCondition);
+    if (didExecute) {
+        _executing = NO;
+        [self didChangeValueForKey:@"isExecuting"];
     }
 
-    pthread_mutex_unlock(&priv->finishLock);
+    _completionBlockLock.lock();
+    void (^completion)(void) = [_completionBlock retain];
+    _completionBlock = nil;
```
rajsesh-msft: >_completionBlock [](start = 4, length = 16)

dont need to notify change here?

rajsesh-msft: will be lot simpler to use completion = self.completionBlock
self.completionBlock = nil;
completion();


---
In reply to: [69981681](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69981681) [](ancestors = 69981681)

Frameworks/Foundation/NSOperation.mm
```diff
@@ -131,139 +162,122 @@ - (void)setThreadPriority:(double)priority {
...
-    int newValue = finished ? 1 : 0;
-    pthread_mutex_lock(&priv->finishLock);
+    THROW_NS_IF(E_INVALIDARG, (_executing || ![self isReady]));
+    
+    _finishLock.lock();
+    BOOL shouldExecute = !_cancelled; // Note: in the cancelled case, [self main] is not called but isFinished will still be observable
+    if (shouldExecute) {
+        [self willChangeValueForKey:@"isExecuting"];
+        _executing = YES;
+        [self didChangeValueForKey:@"isExecuting"];
```
rajsesh-msft: can you move this to setExecution() and reuse it?

Frameworks/Foundation/NSOperation.mm
```diff
@@ -77,41 +64,85 @@ + (BOOL)automaticallyNotifiesObserversForKey:(NSString*)key {
     return NO;
 }
 
-/**
- @Status Interoperable
-*/
-+ (id)allocWithZone:(NSZone*)zone {
-    NSOperation* ret = [super allocWithZone:zone];
+- (void)_checkReady {
```
rajsesh-msft: >_checkReady [](start = 8, length = 11)

_checkReady can be changed to O(1) instead of O(N) by keeping the ready count.  since you are observing dependencies and their 'finished' state, incrementing/decrementing the counter should do the trick.

rajsesh-msft: Sycned up offline with John. Since this is an urgent fix, will follow-up on this separately.  #624

---
In reply to: [69982783](https://github.com/Microsoft/WinObjC/pull/602#discussion_r69982783) [](ancestors = 69982783)

rajsesh-msft:
:shipit:

PR #603, opened 2016-07-01
Add activation functional tests for Cortana

aballway:
Changes activation code to allow functional tests to activate with custom arguments and adds test for Cortana foreground activation.

### DIFF
--- a/Frameworks/UIKit/StarboardXaml/StarboardXaml.cpp
+++ b/Frameworks/UIKit/StarboardXaml/StarboardXaml.cpp
@@ -79,49 +79,8 @@ void App::OnLaunched(LaunchActivatedEventArgs^ args) {
 }
 
 void App::OnActivated(IActivatedEventArgs^ args) {
-    TraceVerbose(TAG, L"OnActivated event received for %d. Previous app state was %d", args->Kind, args->PreviousExecutionState);
-
-    bool initiateAppLaunch = false;
-    if ((args->PreviousExecutionState != ApplicationExecutionState::Running) &&
-        (args->PreviousExecutionState != ApplicationExecutionState::Suspended)) {
-        TraceVerbose(TAG, L"Initializing application");
-        initiateAppLaunch = true;
-    }
-
-    if (args->Kind == ActivationKind::ToastNotification) {
-        Platform::String^ argsString = safe_cast<ToastNotificationActivatedEventArgs^>(args)->Argument;
-        TraceVerbose(TAG, L"Received toast notification with argument - %s", argsString->Data());
-
-        if (initiateAppLaunch) {
-            _ApplicationMainLaunch(ActivationTypeToast, argsString);
-        }
-
-        UIApplicationMainHandleToastNotificationEvent(Strings::WideToNarrow(argsString->Data()).c_str());
-    } else if (args->Kind == ActivationKind::VoiceCommand) {
-        Windows::Media::SpeechRecognition::SpeechRecognitionResult^ argResult = safe_cast<VoiceCommandActivatedEventArgs^>(args)->Result;
-        TraceVerbose(TAG, L"Received voice command with argument - %s", argResult->Text->Data());
-
-        if (initiateAppLaunch) {
-            _ApplicationMainLaunch(ActivationTypeVoiceCommand, argResult);
-        }
-
-        UIApplicationMainHandleVoiceCommandEvent(reinterpret_cast<IInspectable*>(argResult));
-    } else if (args->Kind == ActivationKind::Protocol) {
-        Windows::Foundation::Uri^ argUri = safe_cast<ProtocolActivatedEventArgs^>(args)->Uri;
-        TraceVerbose(TAG, L"Received protocol with uri- %s", argUri->ToString()->Data());
-
-        if (initiateAppLaunch) {
-            _ApplicationMainLaunch(ActivationTypeProtocol, argUri);
-        }
-
-        UIApplicationMainHandleProtocolEvent(reinterpret_cast<IInspectable*>(argUri));
-    } else {
-        TraceVerbose(TAG, L"Received unhandled activation kind - %d", args->Kind);
-        
-        if (initiateAppLaunch) {
-            _ApplicationMainLaunch(ActivationTypeNone, nullptr);
-        }
-    }
+        _ApplicationActivate(args);
+        _RegisterEventHandlers();
 }
 
 void App::_ApplicationMainLaunch(ActivationType activationType, Platform::Object^ activationArg) {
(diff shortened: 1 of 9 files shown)


rajsesh-msft:

rajsesh-msft:

tests/functionaltests/Tests/CortanaTests.mm
```diff
@@ -106,9 +222,85 @@
...
+
+TEST(CortanaTest, ProtocolForegroundActivationDelegateMethodsCalled) {
+    ASSERT_TRUE(willFinishLaunchingWithOptionsCalled);
+    ASSERT_TRUE(didFinishLaunchingWithOptionsCalled);
+    ASSERT_TRUE(didReceiveProtocolCalled);
+
+    // Set flags to false so they can be used for other tests.
+    willFinishLaunchingWithOptionsCalled = false;
+    didFinishLaunchingWithOptionsCalled = false;
+    didReceiveProtocolCalled = false;
```
bbowman: This is a reason that we shouldn't be using static flags for test markers. Every test should be independent and static test flags like this make that easier to mess up than some scoped variables. As you can see from the MOCK_METHOD implementations, there are mechanism to dynamically control what something does at runtime. I would prefer if the CortanaProtocolForegroundTestDelegate class had similar functionality.

tests/functionaltests/Tests/CortanaTests.mm
```diff
@@ -31,15 +32,21 @@
...
 using namespace ABI::Windows::ApplicationModel::Activation;
 using namespace ABI::Windows::Media::SpeechRecognition;
+using namespace ABI::Windows::Foundation;
 using namespace Microsoft::WRL;
 
+// Method to call in tests to activate app
+extern "C" void UIApplicationActivationTest(IInspectable* args);
+
+// These are for testing that delegate methods are actually called
+static NSMutableDictionary* cortanaVoiceCommandForegroundDelegateMethodsCalled = [NSMutableDictionary new];
```
mnithish: nit: Need not be static, can be part of CortanaXXXForegroundTestDelegate test interface

tests/functionaltests/Tests/CortanaTests.mm
```diff
@@ -106,9 +221,75 @@
...
+
     fakeVoiceCommandActivatedEventArgs->Setget_PreviousExecutionState([](ApplicationExecutionState* state) {
         *state = ApplicationExecutionState_NotRunning;
         return S_OK;
     });
+
+    // Swizzle method to use our delegate for testing
+    Method originalMethod = class_getInstanceMethod([UIApplication class], @selector(setDelegate:));
+    Method swizzledMethod = class_getInstanceMethod([UIApplication class], @selector(swizzleVoiceCommandDelegate:));
+    method_exchangeImplementations(originalMethod, swizzledMethod);
```
mnithish: Nice!!

mnithish: Though this is a cool way to test, all that was required was to set g_delegateClassName with the classname of your custom delegate in StarboardXaml.cpp for it to be picked up. This is how UIApplicationDelegate gets set and would avoid test restriction to be able to call it only once per class etc.

Your call if you want to clean it up and make the test simpler.

mnithish:
:shipit:

tests/functionaltests/Tests/CortanaTests.mm
```diff
@@ -54,36 +57,150 @@
...
+    [[self methodsCalled] setObject:@(YES) forKey:NSStringFromSelector(_cmd)];
+    return true;
+}
+
+- (BOOL)application:(UIApplication*)application didFinishLaunchingWithOptions:(NSDictionary*)launchOptions {
+    ASSERT_TRUE(launchOptions[UIApplicationLaunchOptionsVoiceCommandKey]);
+    WMSSpeechRecognitionResult* result = launchOptions[UIApplicationLaunchOptionsVoiceCommandKey];
+    ASSERT_STREQ("CORTANA_TEST", [result.text UTF8String]);
+    [[self methodsCalled] setObject:@(YES) forKey:NSStringFromSelector(_cmd)];
+    return true;
```
DHowett-MSFT: Since you're in an instance method, you don't need to use the property accessor.
```_methodsCalled[NSStringFromSelector(_cmd)] = @(YES);```

File: Frameworks/UIKit/StarboardXaml/StarboardXaml.cpp [hunk 5]
```diff
@@ -223,4 +228,15 @@ void UIApplicationMainTest() {
...
 }
 
+UIKIT_EXPORT
+void UIApplicationActivationTest(IInspectable* activationArgs, wchar_t* delegateName){
+    // Initialize COM on this thread
+    ::CoInitializeEx(nullptr, COINIT_MULTITHREADED);
+
+    // Register tracelogging
+    TraceRegister();
+	g_delegateClassName = ref new Platform::String(delegateName);
```
DHowett-MSFT: nit: it looks like a tab sneaked in

Frameworks/UIKit/StarboardXaml/StarboardXaml.cpp
```diff
@@ -181,6 +140,52 @@ extern "C" void _ApplicationLaunch(ActivationType activationType, Platform::Obje
...
+
+        if (initiateAppLaunch) {
+            _ApplicationLaunch(ActivationTypeVoiceCommand, argResult);
+        }
+
+        UIApplicationMainHandleVoiceCommandEvent(reinterpret_cast<IInspectable*>(argResult));
+    } else if (args->Kind == ActivationKind::Protocol) {
+        Windows::Foundation::Uri^ argUri = safe_cast<ProtocolActivatedEventArgs^>(args)->Uri;
+        TraceVerbose(TAG, L"Received protocol with uri- %s", argUri->ToString()->Data());
+
```
DHowett-MSFT: Ideally, we'll use `%hs` and `%ls` to indicate what sort of string we're passing in the args.
Otherwise, `L"%s"` means `L"%ls"` implicitly.

tests/functionaltests/Tests/CortanaTests.mm
```diff
@@ -106,9 +223,65 @@
...
+    });
+
+    fakeProtocolActivatedEventArgs->Setget_PreviousExecutionState([](ApplicationExecutionState* state) {
+        *state = ApplicationExecutionState_NotRunning;
+        return S_OK;
+    });
+
+    // Pass activation argument to method which activates the app
+    auto args = fakeProtocolActivatedEventArgs.Detach();
+    UIApplicationActivationTest(reinterpret_cast<IInspectable*>(args), L"CortanaProtocolForegroundTestDelegate");
```
mnithish: NSStringFromClass([CortanaProtocolForegroundTestDelegate class]));

tests/functionaltests/Tests/CortanaTests.mm
```diff
@@ -106,9 +223,65 @@
...
-    
+
     fakeVoiceCommandActivatedEventArgs->Setget_PreviousExecutionState([](ApplicationExecutionState* state) {
         *state = ApplicationExecutionState_NotRunning;
         return S_OK;
     });
+
+    // Pass activation argument to method which activates the app
+    auto args = fakeVoiceCommandActivatedEventArgs.Detach();
+    UIApplicationActivationTest(reinterpret_cast<IInspectable*>(args), L"CortanaVoiceCommandForegroundTestDelegate");
```
mnithish: NSStringFromClass([CortanaVoiceCommandForegroundTestDelegate class]));

tests/functionaltests/Tests/CortanaTests.mm
DHowett-MSFT: That's a very unusual way to get the class name. Also, it looks like this doesn't take an NSString, it takes a wide char string.

---
In reply to: [69814325](https://github.com/Microsoft/WinObjC/pull/603#discussion_r69814325) [](ancestors = 69814325)

Frameworks/UIKit/StarboardXaml/StarboardXaml.cpp
mnithish: nit: check if it not null and only set it when it is valid

Frameworks/UIKit/StarboardXaml/StarboardXaml.cpp
```diff
@@ -223,4 +228,15 @@ void UIApplicationMainTest() {
     _ApplicationLaunch(ActivationTypeNone, nullptr);
 }
 
+UIKIT_EXPORT
+void UIApplicationActivationTest(IInspectable* activationArgs, wchar_t* delegateName){
```
mnithish: can we keep the interface similar to UIApplicationMain and use void* i.e. it takes a NSString* and because we cannot present that in C++/Cx file we use void*

tests/functionaltests/Tests/CortanaTests.mm
mnithish: yes, see my other comment in StarboardXaml.cpp for UIApplicationActivationTest function

---
In reply to: [69814747](https://github.com/Microsoft/WinObjC/pull/603#discussion_r69814747) [](ancestors = 69814747,69814325)

Frameworks/UIKit/StarboardXaml/StarboardXaml.cpp
aballway: #Resolved

Frameworks/UIKit/StarboardXaml/StarboardXaml.cpp
aballway: test

---
In reply to: [69815541](https://github.com/Microsoft/WinObjC/pull/603#discussion_r69815541) [](ancestors = 69815541)

bbowman:
:shipit:

rajsesh-msft:
:shipit:

aballway:
CI build is successful

PR #609, opened 2016-07-05
Fix CFAutorelease on _NSCFString and _NSCFNumber.

### LINKED ISSUE #504: Memory leak when repeatedly adding/removing KVO observers on an object.
There seems to be a memory leak when repeatedly adding/removing KVO observers on an object.

The attached sample project demonstrates the issue.

1. Unzip, build and run app
2. Press "Update Name Property"
3. Launch Task Manager and look for "KVOBug.exe". Notice the memory continuously growing as the app is executing.

A lot of the memory allocation seems to occur in `_addNestedObserversAndOptionallyDependents` and `_addKeyObserver`, but then never gets freed when the observer is removed.

[KVOBug-memoryleak.zip](https://github.com/Microsoft/WinObjC/files/270157/KVOBug-memoryleak.zip)

DHowett-MSFT:
Fixes #504.

`-autorelease` on `_NSCFString`, `_NSCFNumber`, and `_NSCFBoolean` instances was nonoperational, as we got the `IsBridgedObject` check somewhat backwards.

### DIFF
--- a/Frameworks/CoreFoundation/Base.subproj/CFInternal.h
+++ b/Frameworks/CoreFoundation/Base.subproj/CFInternal.h
@@ -707,11 +707,14 @@ CF_INLINE uintptr_t __CFISAForTypeID(CFTypeID typeID) {
 CF_INLINE bool __CF_IsBridgedObject(CFTypeRef obj) {
     CFRuntimeBase* object = (CFRuntimeBase*)obj;
     if (!object || 
-        (object->_cfisa == 0) || 
-        (object->_cfisa == (uintptr_t)(&_OBJC_CLASS__NSCFString)) || 
+        (object->_cfisa == 0)) {
+        return false;
+    }
+
+    if ((object->_cfisa == (uintptr_t)(&_OBJC_CLASS__NSCFString)) || 
         (object->_cfisa == (uintptr_t)(&_OBJC_CLASS__NSCFNumber)) || 
         (object->_cfisa == (uintptr_t)(&_OBJC_CLASS__NSCFBoolean))) {
-        return false;
+        return true;
     }
 
     for (unsigned int i = 0; i < __CFRuntimeClassTableSize; i++) {


Frameworks/CoreFoundation/Base.subproj/CFInternal.h
```diff
@@ -707,11 +707,14 @@ CF_INLINE uintptr_t __CFISAForTypeID(CFTypeID typeID) {
...
-        (object->_cfisa == 0) || 
-        (object->_cfisa == (uintptr_t)(&_OBJC_CLASS__NSCFString)) || 
+        (object->_cfisa == 0)) {
+        return false;
+    }
+
+    if ((object->_cfisa == (uintptr_t)(&_OBJC_CLASS__NSCFString)) || 
         (object->_cfisa == (uintptr_t)(&_OBJC_CLASS__NSCFNumber)) || 
         (object->_cfisa == (uintptr_t)(&_OBJC_CLASS__NSCFBoolean))) {
-        return false;
```
rajsesh-msft: >false [](start = 15, length = 5)

gasp!

rajsesh-msft:

rajsesh-msft:
:shipit:

bbowman:
>         (object->_cfisa == (uintptr_t)(&_OBJC_CLASS__NSCFBoolean)) ||

can be removed?

---
Refers to: Frameworks/CoreFoundation/Base.subproj/CFInternal.h:739 in f56cacc. [](commit_id = f56caccc0d9c0741090397783d01d9ef057da340, deletion_comment = False)

bbowman:
:shipit:

Frameworks/CoreFoundation/Base.subproj/CFInternal.h
```diff
@@ -707,11 +707,14 @@ CF_INLINE uintptr_t __CFISAForTypeID(CFTypeID typeID) {
 CF_INLINE bool __CF_IsBridgedObject(CFTypeRef obj) {
     CFRuntimeBase* object = (CFRuntimeBase*)obj;
     if (!object || 
-        (object->_cfisa == 0) || 
-        (object->_cfisa == (uintptr_t)(&_OBJC_CLASS__NSCFString)) || 
+        (object->_cfisa == 0)) {
+        return false;
+    }
+
+    if ((object->_cfisa == (uintptr_t)(&_OBJC_CLASS__NSCFString)) ||
```
msft-Jeyaram: Given this is a faster resolution does it make sense to add the other well known bridged classes here (NSCF*).
Although I assume these are the most popular classes that are used commonly.

Frameworks/CoreFoundation/Base.subproj/CFInternal.h
```diff
@@ -707,11 +707,14 @@ CF_INLINE uintptr_t __CFISAForTypeID(CFTypeID typeID) {
 CF_INLINE bool __CF_IsBridgedObject(CFTypeRef obj) {
     CFRuntimeBase* object = (CFRuntimeBase*)obj;
     if (!object || 
-        (object->_cfisa == 0) || 
-        (object->_cfisa == (uintptr_t)(&_OBJC_CLASS__NSCFString)) || 
+        (object->_cfisa == 0)) {
```
msft-Jeyaram: we are indicating that isa 0 is a CFObject in the __CF_IsCFObject and then we are covering the logic with a call into __CF_IsBridgedObject, for the __CF_IsCFObject, I assume nil is not a CFObject?

Frameworks/CoreFoundation/Base.subproj/CFInternal.h
bbowman: This is not actually a performance optimization (although it may be useful in that sense too). This is because these three classes in particular are used at static init time and these checks may happen *before* they are registered as bridge classes due to the non determinism in static init.

---
In reply to: [69767697](https://github.com/Microsoft/WinObjC/pull/609#discussion_r69767697) [](ancestors = 69767697)

Frameworks/CoreFoundation/Base.subproj/CFInternal.h
bbowman: Remember that we are manually looking into the object's isa field here. Nil is comparison made on the pointer itself. A 0 isa means its a cf object that has not been bridged yet. 

---
In reply to: [69772935](https://github.com/Microsoft/WinObjC/pull/609#discussion_r69772935) [](ancestors = 69772935)

Frameworks/CoreFoundation/Base.subproj/CFInternal.h
msft-Jeyaram: Lets add a comment there to indicate that case?

DHowett-MSFT:
This is in CI build.

Frameworks/CoreFoundation/Base.subproj/CFInternal.h
DHowett-MSFT: I think in this case, it should be alright without one. Once we revisit CF's objc interop we may not have to worry about this.

Specifically, if everything becomes a bridged internal type, everything is Objective-C-able, and we don't have to worry about many of these checks.

PR #610, opened 2016-07-05
Implement a CoreData xcdatamodel parser.

DHowett-MSFT:
The intent of this pull request is to implement the basic functionality required by NSManagedObjectModel.
In lieu of precompiling the model into a binary format, CoreData will thaw it from XML at runtime.

#### Quick Notes
* Based on CoreFoundation's SAX XML parser.
* Supports most basic entity configuration, attributes, relationships, inversion.
* Makes a best-effort attempt to create constraints for max/min values on attributes.
* **does not** support fetch requests or multiple configurations.


include/CoreData/NSManagedObjectModel.h
```diff
@@ -29,18 +29,18 @@
...
-@property (readonly, copy) NSDictionary* entitiesByName STUB_PROPERTY;
+@property (strong) NSArray* entities;
+@property (readonly, copy) NSDictionary* entitiesByName;
 @property (readonly, strong) NSArray* configurations STUB_PROPERTY;
 - (NSArray*)entitiesForConfiguration:(NSString*)configuration STUB_METHOD;
-- (void)setEntities:(NSArray*)entities forConfiguration:(NSString*)configuration STUB_METHOD;
+- (void)setEntities:(NSArray*)entities forConfiguration:(NSString*)configuration;
 @property (readonly, copy) NSDictionary* fetchRequestTemplatesByName STUB_PROPERTY;
-- (NSFetchRequest*)fetchRequestTemplateForName:(NSString*)name STUB_METHOD;
+- (NSFetchRequest*)fetchRequestTemplateForName:(NSString*)name STUB_PROPERTY;
```
DHowett-MSFT: Oops, this should not have changed from STUB_METHOD.

Frameworks/CoreData/NSAttributeDescription.mm
```diff
@@ -14,8 +14,111 @@
...
+}
+
++ (NSObject<NSCopying>*)_convertValueString:(NSString*)valueString forAttributeType:(NSAttributeType)attributeType {
+    switch (attributeType) {
+        case NSInteger16AttributeType:
+        case NSInteger32AttributeType:
+        case NSInteger64AttributeType:
+            return [NSNumber numberWithLongLong:[valueString longLongValue]];
+        case NSDecimalAttributeType:
+            return nil; // [NSDecimalNumber decimalNumberWithString:valueString]; when implemented (VSO 7346158)
```
msft-Jeyaram: nit: TODO 7346158 ......

Frameworks/CoreData/NSAttributeDescription.mm
```diff
@@ -14,8 +14,111 @@
...
+        case NSBooleanAttributeType:
+            return [NSNumber numberWithBool:[valueString boolValue]];
+        default:
+            // This type can't have a parsed default.
+            return nil;
+    }
+    return nil;
+}
+
+- (id)initWithXMLElementName:(NSString*)entityName attributes:(NSDictionary<NSString*, NSString*>*)attributes {
```
msft-Jeyaram: instancetype?

Frameworks/CoreData/NSEntityDescription.mm
```diff
@@ -14,10 +14,164 @@
...
+    return NO;
+}
+
+- (NSArray<NSPropertyDescription*>*)properties {
+    @synchronized(self) {
+        return _properties;
+    }
+}
+
+- (void)setProperties:(NSArray*)properties {
```
msft-Jeyaram: "Setting the properties raises an exception if the receiver’s model has been used by an object graph manager."
is this something to consider?

Frameworks/CoreData/NSEntityDescription.mm
```diff
@@ -14,10 +14,164 @@
...
+}
+@property (readwrite, assign) NSEntityDescription* superentity;
+@property (readwrite, copy) NSDictionary* subentitiesByName;
+@property (readwrite, copy) NSDictionary* propertiesByName;
+@property (readwrite, copy) NSDictionary* attributesByName;
+@property (readwrite, copy) NSDictionary* relationshipsByName;
+@end
 
 @implementation NSEntityDescription
+- (id)initWithXMLElementName:(NSString*)entityName attributes:(NSDictionary<NSString*, NSString*>*)attributes {
```
msft-Jeyaram: nit: instancetype

Frameworks/CoreData/NSPropertyDescription.mm
```diff
@@ -14,10 +14,64 @@
...
 
+#import <CoreData/CoreData.h>
+#import <Foundation/Foundation.h>
 #import <StubReturn.h>
-#import <CoreData/NSPropertyDescription.h>
+
+#import "NSManagedObjectModel-XMLParsing.h"
 
 @implementation NSPropertyDescription
+- (id)initWithXMLElementName:(NSString*)entityName attributes:(NSDictionary<NSString*, NSString*>*)attributes {
```
msft-Jeyaram: entityName is not being used?

Frameworks/CoreData/NSManagedObjectModel.mm
```diff
@@ -14,17 +14,106 @@
...
-@Status Stub
-@Notes
+ @Status Caveat
+ @Notes Parses all relationships, entities, attributes and ignopres configurations and fetch requests.
 */
-- (instancetype)initWithContentsOfURL:(NSURL*)url {
-    UNIMPLEMENTED();
-    return StubReturn();
+- (id)initWithContentsOfURL:(NSURL*)url {
+    [self release];
```
msft-Jeyaram: release when we enter this function? or did we intend this be released when url is nil?

Frameworks/CoreData/NSEntityDescription.mm
DHowett-MSFT: Since we do not have an object graph manager, we don't need that right now.
Every property is intended to be made readonly by the manager, so we will have to do the work there when we get one.

---
In reply to: [69756520](https://github.com/Microsoft/WinObjC/pull/610#discussion_r69756520) [](ancestors = 69756520)

Frameworks/CoreData/NSPropertyDescription.mm
DHowett-MSFT: Alas, this is for _CDXMLCoding protocol conformance. Just because this class does not use it, it doesn't mean that other protocol consumers won't.

---
In reply to: [69757738](https://github.com/Microsoft/WinObjC/pull/610#discussion_r69757738) [](ancestors = 69757738)

Frameworks/CoreData/NSManagedObjectModel.mm
DHowett-MSFT: `_NSParseManagedObjectModelFromXMLDataAtURL` returns a new instance of `NSManagedObjectModel`

---
In reply to: [69758575](https://github.com/Microsoft/WinObjC/pull/610#discussion_r69758575) [](ancestors = 69758575)

tests/unittests/CoreData/CoreDataTests.mm
```diff
@@ -0,0 +1,25 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#import <CoreData/CoreData.h>
+#import <Foundation/NSURL.h>
+#include <TestFramework.h>
+
+TEST(CoreData, NSManagedObjectModelInit) {
```
rajsesh-msft: We need better tests here, perhaps parse a real model file.  Otherwise none of the xml parsing code is hit.

Frameworks/CoreData/NSManagedObjectModel-XMLParsing.h
```diff
@@ -0,0 +1,52 @@
...
+//
+//******************************************************************************
+
+#import <CoreData/CoreData.h>
+#import <Foundation/Foundation.h>
+#import <Starboard/SmartTypes.h>
+
+@protocol _NSCDXMLCoding <NSObject>
+@required
+- (instancetype)initWithXMLElementName:(NSString*)entityName attributes:(NSDictionary<NSString*, NSString*>*)attributes;
```
rajsesh-msft: >initWithXMLElementName [](start = 16, length = 22)

nit: only init doesnt have the _ notation for private methods.  Given that the protocol is internal, maybe drop the _ everywhere to be consistent?

Frameworks/CoreData/NSManagedObjectModel-XMLParsing.mm
```diff
@@ -0,0 +1,282 @@
...
+static NSString* const _kNSManagedObjectModelEntityCompoundIndexTag = @"compoundIndex";
+static NSString* const _kNSManagedObjectModelEntityCompoundIndexesTag = @"compoundIndexes";
+static NSString* const _kNSManagedObjectModelEntityCompoundIndexIndexTag = @"index";
+static NSString* const _kNSManagedObjectModelEntityUniquenessConstraintTag = @"uniquenessConstraint";
+static NSString* const _kNSManagedObjectModelEntityUniquenessConstraintsTag = @"uniquenessConstraints";
+static NSString* const _kNSManagedObjectModelEntityUniquenessConstraintConstraintTag = @"constraint";
+
+// To save ourselves some string compares:
+static NSString* _uniqueTag(NSString* tag) {
+    static StrongId<NSSet> _tagSet = [NSSet setWithObjects:_kNSManagedObjectModelTag,
```
rajsesh-msft: Nit: NSSet<NSString*>?

Frameworks/CoreData/NSManagedObjectModel-XMLParsing.mm
```diff
@@ -0,0 +1,282 @@
...
+    _NSModelParsingStateCompoundIndexEntry, // object is a string
+
+    _NSModelParsingStateUniquenessConstraints, // object is an array<array<string>>
+    _NSModelParsingStateUniquenessConstraint, // object is an array<string>
+    _NSModelParsingStateUniquenessConstraintEntry, // object is a string
+};
+
+struct _NSModelParsingContext {
+    StrongId<NSManagedObjectModel> model;
+    std::list<std::tuple<_NSModelParsingState, StrongId<NSObject>>> states;
```
rajsesh-msft: >list [](start = 9, length = 4)

std::stack to save you some other code here?

Frameworks/CoreData/NSManagedObjectModel-XMLParsing.mm
```diff
@@ -0,0 +1,282 @@
...
+        _kNSManagedObjectModelEntityTag : [NSEntityDescription class],
+        _kNSManagedObjectModelEntityAttributeTag : [NSAttributeDescription class],
+        _kNSManagedObjectModelEntityRelationshipTag : [NSRelationshipDescription class],
+    };
+    return [_mappings objectForKey:elementName];
+}
+
+// XML data model parsing is done via recursive descent and state management. Each element
+// we encounter can push another state onto the stack. The state dictates what state transitions
+// are available and what is to be done when the state is exited.
```
rajsesh-msft: need a warning here that the states have to be in specific order for the state machine to work.  In case someone gets adventurous later on to change this code :).

Frameworks/CoreData/NSEntityDescription.mm
```diff
@@ -14,10 +14,164 @@
...
+    if (self = [super init]) {
+        _unresolvedProperties = [NSMutableArray new];
+        _name = [attributes[@"name"] copy];
+        _managedObjectClassName = [attributes[@"representedClassName"] copy];
+        _renamingIdentifier = [attributes[@"elementID"] copy]; // TODO(DH): figure.
```
rajsesh-msft: Not clear what this comment is

Frameworks/CoreData/NSEntityDescription.mm
```diff
@@ -14,10 +14,164 @@
...
+- (void)_resolveRelationships {
+    // This should only be done once the model above us is stabilized.
+    for (NSRelationshipDescription* relationship in [_relationshipsByName allValues]) {
+        relationship.destinationEntity = [_managedObjectModel.entitiesByName objectForKey:relationship._destinationEntityName];
+
+        NSEntityDescription* inverseEntity = [_managedObjectModel.entitiesByName objectForKey:relationship._inverseEntityName];
+
+        relationship.inverseRelationship = [inverseEntity.relationshipsByName objectForKey:relationship._inverseRelationshipName];
+        if (relationship._inverseEntityName != nil && relationship.inverseRelationship == nil) {
+            __debugbreak();
```
rajsesh-msft: >__debugbreak [](start = 12, length = 12)

FAIL_FAST()?

Frameworks/CoreData/NSManagedObjectModel.mm
```diff
@@ -14,17 +14,106 @@
...
+    @synchronized(self) {
+        if (entities == _entities) {
+            return;
+        }
+        [_entities autorelease];
+        _entities = [entities copy];
+
+        NSMutableDictionary<NSString*, NSEntityDescription*>* entitiesByName = [NSMutableDictionary dictionary];
+
+        // Pass 1: add all named entities.
```
rajsesh-msft: >Pass [](start = 11, length = 4)

is there a pass 2?

rajsesh-msft:
:clock1:

Frameworks/CoreData/NSManagedObjectModel-XMLParsing.mm
DHowett-MSFT: Good idea! I forgot about `std::stack`.

Frameworks/CoreData/NSEntityDescription.mm
DHowett-MSFT: Oh, I think I left it in from before I matched up elementID with renamingIdentifier. I'll cross-check to make sure it's 1) correct and 2) not necessary to have an ambiguous comment

Frameworks/CoreData/NSEntityDescription.mm
DHowett-MSFT: Oops. Thanks!

Frameworks/CoreData/NSManagedObjectModel-XMLParsing.mm
DHowett-MSFT: This is a bit of a premature optimization, but: using NSSet here gets us O(1) key uniquing. It ensures that `@"userInfo"` is pointer-identical with whatever instance of the string `"userInfo"` CFXML gives us. We avoid character-by-character string comparison for every tag we see in the XML document.

DHowett-MSFT: That is: We look up the tag we got from CFXML in this set, and it gives us back the const version.

DHowett-MSFT:
I added an in-depth single entity two-property test; one attribute and one relationship. It doesn't get into any attribute constraints, but it does make sure that a lot of the basic parseable keys are intact.

build/CoreData/dll/CoreData.vcxproj
```diff
@@ -25,6 +25,15 @@
...
   <ItemGroup>
+    <ProjectReference Include="..\..\CoreFoundation\dll\CoreFoundation.vcxproj">
+      <Project>{81F30AF6-EAC3-4DFA-929A-C25D69E8080B}</Project>
+    </ProjectReference>
+    <ProjectReference Include="..\..\Foundation\dll\Foundation.vcxproj">
+      <Project>{86127226-9A6E-439B-A070-420A572AF0C7}</Project>
+    </ProjectReference>
+    <ProjectReference Include="..\..\Starboard\dll\Starboard.vcxproj">
+      <Project>{0AC27ECF-E2AB-420B-9359-4843FFF4CBFA}</Project>
+    </ProjectReference>
```
bbowman: I don't think you need this if its using sdk-build.props.

Frameworks/CoreData/NSAttributeDescription.mm
```diff
@@ -14,8 +14,111 @@
...
 #import <StubReturn.h>
-#import <CoreData/NSAttributeDescription.h>
+
+#import <CoreData/NSManagedObjectModel-XMLParsing.h>
+
+@interface NSAttributeDescription () {
+    StrongId<id> _minValue;
+    StrongId<id> _maxValue;
+}
+@end
```
bbowman: I thought the hotness was to put this in the implementation? Also curious that _ things are publicly exposed.

Frameworks/CoreData/NSAttributeDescription.mm
```diff
@@ -14,8 +14,111 @@
...
+- (NSString*)description {
+    return [NSString stringWithFormat:@"<%@ %p: %@,%@%@%@%@%@ default=%@ userInfo=%@>",
+                                      object_getClass(self),
+                                      self,
+                                      self.name,
+                                      self.optional ? @" optional" : @"",
+                                      self.transient ? @" transient" : @"",
+                                      self.indexed ? @" indexed" : @"",
+                                      self.storedInExternalRecord ? @" external" : @"",
+                                      self.indexedBySpotlight ? @" spotlight" : @"",
```
bbowman: would %hs be more efficient here?

Frameworks/CoreData/NSEntityDescription.mm
```diff
@@ -14,10 +14,164 @@
...
+    if (self = [super init]) {
+        _unresolvedProperties = [NSMutableArray new];
+        _name = [attributes[@"name"] copy];
+        _managedObjectClassName = [attributes[@"representedClassName"] copy];
+        _renamingIdentifier = [attributes[@"elementID"] copy]; // TODO(DH): figure.
+        _versionHashModifier = [attributes[@"versionHashModifier"] copy];
+        _abstract = [attributes[@"isAbstract"] isEqualToString:@"YES"];
+
+        // _userInfo, _compoundIndexes, and _uniquenessConstraints are set out-of-band by the XML parser.
+        // _subentities, *ByName, and other collection properties are derived from sub-elements via the XML parser.
```
bbowman: two comment lines here appear disconnected / lack context to make sense of them. Please add to these / move them to be connected to the code block in question. Perhaps a header comment on the init function if you are saying these aren't getting set got some reason?

Frameworks/CoreData/NSEntityDescription.mm
```diff
@@ -14,10 +14,164 @@
...
+    [_compoundIndexes release];
+    [_uniquenessConstraints release];
+    [_versionHash release];
+    [_versionHashModifier release];
+
+    // derived properties
+    [_subentitiesByName release];
+    [_propertiesByName release];
+    [_attributesByName release];
+    [_relationshipsByName release];
```
bbowman: This makes me so sad. I suppose this is due to synthesized properties but :-( ... RAII would be much nicer.

Frameworks/CoreData/NSManagedObjectModel-XMLParsing.mm
```diff
@@ -0,0 +1,288 @@
...
+static void* SKIP = nullptr;
+static void* CONTINUE = (void*)0x01;
+
+#define EMIT_STATE(st, ...)                   \
+    do {                                      \
+        context->push_state(st, __VA_ARGS__); \
+        return CONTINUE;                      \
+    } while (0);
+
+static void* xmlCreateStructure(CFXMLParserRef parser, CFXMLNodeRef node, void* info) {
```
bbowman: >static void*  [](start = 0, length = 13)

Why a void* return?

Frameworks/CoreData/NSManagedObjectModel-XMLParsing.mm
```diff
@@ -0,0 +1,288 @@
...
+};
+
+static void* SKIP = nullptr;
+static void* CONTINUE = (void*)0x01;
+
+#define EMIT_STATE(st, ...)                   \
+    do {                                      \
+        context->push_state(st, __VA_ARGS__); \
+        return CONTINUE;                      \
+    } while (0);
```
bbowman: seems overkill here. Can you just push and call return? Whats the savings?

Frameworks/CoreData/NSManagedObjectModel-XMLParsing.mm
DHowett-MSFT: I understand now: GitHub's e-mail did not include the `<NSString*>` (because HTML :P) and I thought you were asking why it was an NSSet.
This is a great suggestion. Thanks. :)

---
In reply to: [70035761](https://github.com/Microsoft/WinObjC/pull/610#discussion_r70035761) [](ancestors = 70035761)

Frameworks/CoreData/NSManagedObjectModel-XMLParsing.mm
DHowett-MSFT: This is part of the CFXML contract. `CFXMLCreateStructure` callbacks must return a structure to continue parsing, else the element is skipped. In order, the calls become create -> add -> end. That doesn't work for us, however, because we want to realize all our elements before we add them to the hierarchy.
We return _something_ here and move the work into `endStructure` so that we get final control.

---
In reply to: [70722579](https://github.com/Microsoft/WinObjC/pull/610#discussion_r70722579) [](ancestors = 70722579)

Frameworks/CoreData/NSManagedObjectModel-XMLParsing.mm
DHowett-MSFT: It should not be possible to easily emit a new state without returning; coupling them here neatly enforces that soft contract.

---
In reply to: [70722626](https://github.com/Microsoft/WinObjC/pull/610#discussion_r70722626) [](ancestors = 70722626)

build/CoreData/dll/CoreData.vcxproj
DHowett-MSFT: Good catch, thanks.

---
In reply to: [70720366](https://github.com/Microsoft/WinObjC/pull/610#discussion_r70720366) [](ancestors = 70720366)

Frameworks/CoreData/NSAttributeDescription.mm
DHowett-MSFT: ?
Our coding convention usually places ivars in the class extension block in the implementation file, which this is.
They're not exposed, though. min/max value are exposed only through the validation predicates.

---
In reply to: [70720550](https://github.com/Microsoft/WinObjC/pull/610#discussion_r70720550) [](ancestors = 70720550)

Frameworks/CoreData/NSEntityDescription.mm
DHowett-MSFT: Alas :(

---
In reply to: [70721690](https://github.com/Microsoft/WinObjC/pull/610#discussion_r70721690) [](ancestors = 70721690)

Frameworks/CoreData/NSAttributeDescription.mm
DHowett-MSFT: Likely, but this is not intended to be used supremely often. Logging Entity, attribute, and property descriptions is not a use case that needs to be performant.

jofre-ms:
:shipit:

DHowett-MSFT:
In CI build.

PR #612, opened 2016-07-06
Fix #583: NSUserdefaults synchronize

### LINKED ISSUE #583: NSUserDefaults are not being saved across launches
User defaults settings are not being restored when re-launching the app.

Running the following code should output `Test value: 0` the first time, but `Test value: 1` for every subsequent run of the app. However, it always outputs `Test value: 0`.

```objc
NSLog(@"Test value: %d", [[NSUserDefaults standardUserDefaults] boolForKey:@"Test"]);
[[NSUserDefaults standardUserDefaults] setBool:YES forKey:@"Test"];
```

jofre-ms:
* NSUserdefaults synchronize must be called after app settings have changed.
* Added boolean to NSUSerDefaults WOCCatalog page

### DIFF
--- a/Frameworks/Foundation/NSUserDefaults.mm
+++ b/Frameworks/Foundation/NSUserDefaults.mm
@@ -48,6 +48,7 @@ FOUNDATION_EXPORT NSString* const NSUserDefaultsDidChangeNotification = @"NSUser
 
 @implementation NSUserDefaults {
     StrongId<NSMutableDictionary> _registrationDict;
+    StrongId<NSOperationQueue> _synchronizeQueue;
 }
 
 /**
@@ -57,6 +58,9 @@ FOUNDATION_EXPORT NSString* const NSUserDefaultsDidChangeNotification = @"NSUser
     [super init];
 
     _registrationDict = [NSMutableDictionary dictionary];
+
+    _synchronizeQueue = [NSOperationQueue new];
+    [_synchronizeQueue setMaxConcurrentOperationCount:1];
 
     [self setObject:[NSArray arrayWithObject:@"en"] forKey:@"AppleLanguages"];
     [self setObject:@"en_US" forKey:@"AppleLocale"];
@@ -245,6 +249,12 @@ FOUNDATION_EXPORT NSString* const NSUserDefaultsDidChangeNotification = @"NSUser
     CFPreferencesSetAppValue(static_cast<CFStringRef>(key), valueCopy, kCFPreferencesCurrentApplication);
 
     [[NSNotificationCenter defaultCenter] postNotificationName:NSUserDefaultsDidChangeNotification object:self];
+
+    if ([_synchronizeQueue operationCount] < 2) {
+        [_synchronizeQueue addOperationWithBlock:^void(void) {
+            [self synchronize];
+        }];
+    }
 }
 
 /**
--- a/Frameworks/Foundation/NSUserDefaults.mm
+++ b/Frameworks/Foundation/NSUserDefaults.mm
@@ -54,7 +54,7 @@ FOUNDATION_EXPORT NSString* const NSUserDefaultsDidChangeNotification = @"NSUser
     StrongId<NSOperationQueue> _synchronizeQueue;
 }
 
-static NSUserDefaults* _standard;
+static StrongId<NSUserDefaults> _standard = nil;
 
 /**
  @Status Interoperable
@@ -97,11 +97,18 @@ static NSUserDefaults* _standard;
  @Status Interoperable
 */
 + (NSUserDefaults*)standardUserDefaults {
-    if (_standard == nil) {
-        _standard = [self new];
+    @synchronized(self) {
+        if (nil == _standard) {
+            _standard.attach([self new]);
+        }
+        return _standard;
     }
+}
 
-    return _standard;
++ (NSUserDefaults*)_standardUserDefaultsNoInitialize {
+    @synchronized(self) {
+        return _standard;
+    }
 }
 
 /**
@@ -309,17 +316,13 @@ static NSUserDefaults* _standard;
     return _synchronizeQueue;
 }
 
-+ (void) _suspendSynchronize {
-    if (_standard != nil) {
-        [[_standard _synchronizeQueue] setSuspended:YES];
-        [[_standard _synchronizeQueue] waitUntilAllOperationsAreFinished];
-    }
+- (void) _suspendSynchronize {
+    [_synchronizeQueue setSuspended:YES];
+    [_synchronizeQueue waitUntilAllOperationsAreFinished];
 }
 
-+ (void) _resumeSynchronize {
-    if (_standard != nil) {
-        [[_standard _synchronizeQueue] setSuspended:NO];
-    }
+- (void) _resumeSynchronize {
+    [_synchronizeQueue setSuspended:NO];
 }
 
 /**
--- a/Frameworks/Foundation/NSUserDefaults.mm
+++ b/Frameworks/Foundation/NSUserDefaults.mm
@@ -25,6 +25,7 @@ OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 #import <Foundation/NSUserDefaults.h>
 #import <Foundation/NSThread.h>
 #import "NSStringInternal.h"
+#import "NSUserDefaultsInternal.h"
 #import <LoggingNative.h>
 #import <ForFoundationOnly.h>
 #import <CoreFoundation/CFPreferences.h>
@@ -53,20 +54,36 @@ FOUNDATION_EXPORT NSString* const NSUserDefaultsDidChangeNotification = @"NSUser
     StrongId<NSOperationQueue> _synchronizeQueue;
 }
 
+static NSUserDefaults* _standard;
+
 /**
  @Status Interoperable
 */
 - (instancetype)init {
-    [super init];
+    return [self initWithSuiteName:nil];
+}
 
-    _cacheDict = [NSMutableDictionary dictionary];
-    _registrationDict = [NSMutableDictionary dictionary];
+/**
+ @Status Caveat
+ @Notes supports nil only for suitename
+*/
+- (instancetype)initWithSuiteName:(NSString*)suitename {
+    if (nil != suitename) {
+        UNIMPLEMENTED();
+        return StubReturn();
+    }
 
-    _synchronizeQueue.attach([NSOperationQueue new]);
-    [_synchronizeQueue setMaxConcurrentOperationCount:1];
+    if(self = [super init]) {
 
-    [self setObject:[NSArray arrayWithObject:@"en"] forKey:@"AppleLanguages"];
-    [self setObject:@"en_US" forKey:@"AppleLocale"];
+        _cacheDict = [NSMutableDictionary dictionary];
+        _registrationDict = [NSMutableDictionary dictionary];
+
+        _synchronizeQueue.attach([NSOperationQueue new]);
+        [_synchronizeQueue setMaxConcurrentOperationCount:1];
+
+        [self setObject:[NSArray arrayWithObject:@"en"] forKey:@"AppleLanguages"];
+        [self setObject:@"en_US" forKey:@"AppleLocale"];
+    }
 
     return self;
 }
@@ -80,13 +97,11 @@ FOUNDATION_EXPORT NSString* const NSUserDefaultsDidChangeNotification = @"NSUser
  @Status Interoperable
 */
 + (NSUserDefaults*)standardUserDefaults {
-    static NSUserDefaults* standard;
-
-    if (standard == nil) {
-        standard = [self new];
+    if (_standard == nil) {
+        _standard = [self new];
     }
 
-    return standard;
+    return _standard;
 }
 
 /**
@@ -129,22 +144,26 @@ FOUNDATION_EXPORT NSString* const NSUserDefaultsDidChangeNotification = @"NSUser
 }
 
 /**
- @Status Interoperable
+ @Status Caveat
+ @Notes Writes to file only - external changes to the preferences file are overwritten.
 */
 - (BOOL)synchronize {
-    _cacheLock.lock();
-    BOOL isDirty = _cacheIsDirty;
+    BOOL isDirty;
 
-    if (isDirty) {
-        [_cacheDict enumerateKeysAndObjectsUsingBlock:^(id key, id obj, BOOL* stop) {
-            CFPreferencesSetAppValue(static_cast<CFStringRef>(key), obj, kCFPreferencesCurrentApplication);
-        }];
+    // lock scope
+    {
+        std::lock_guard<std::mutex> lock(_cacheLock);
+        isDirty = _cacheIsDirty;
 
-        [_cacheDict removeAllObjects];
-        _cacheIsDirty = NO;
+        if (isDirty) {
+            [_cacheDict enumerateKeysAndObjectsUsingBlock:^(id key, id obj, BOOL* stop) {
+                CFPreferencesSetAppValue(static_cast<CFStringRef>(key), obj, kCFPreferencesCurrentApplication);
+            }];
+
+            [_cacheDict removeAllObjects];
+            _cacheIsDirty = NO;
+        }
     }
-    _cacheLock.unlock();
-
 
     BOOL result = NO;
     if (isDirty) {
@@ -158,10 +177,13 @@ FOUNDATION_EXPORT NSString* const NSUserDefaultsDidChangeNotification = @"NSUser
  @Status Interoperable
 */
 - (id)objectForKey:(NSString*)defaultName {
-    // Check cache
-    _cacheLock.lock();
-    id obj = [_cacheDict objectForKey:defaultName];
-    _cacheLock.unlock();
+    id obj;
+
+    // Check cache under lock
+    {
+        std::lock_guard<std::mutex> lock(_cacheLock);
+        obj = [_cacheDict objectForKey:defaultName];
+    }
 
     if (!obj) {
         // Check stored app preferences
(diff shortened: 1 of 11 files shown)


samples/WOCCatalog/WOCCatalog/UserDefaultsViewController.m
```diff
@@ -51,7 +52,19 @@ - (void)viewDidLoad {
...
+    // Slider for retrieving favorite bool
+    UISwitch* switchFavoriteBool = [[UISwitch alloc] initWithFrame:frame];
+    [switchFavoriteBool setOn:[[standardDefaults objectForKey:@"FavoriteBool"] boolValue] animated:NO];
+    [_textFields addObject:switchFavoriteBool];
+    [switchFavoriteBool addTarget:self action:@selector(changeFavoriteBool) forControlEvents:UIControlEventValueChanged];
+
+    [self tableView].allowsSelection = NO;    
+}
+
+- (void) changeFavoriteBool {
```
DHowett-MSFT: nit: formatting

Frameworks/Foundation/NSUserDefaults.mm
```diff
@@ -58,6 +59,9 @@ - (instancetype)init {
 
     _registrationDict = [NSMutableDictionary dictionary];
 
+    _synchronizeQueue = [NSOperationQueue new];
```
DHowett-MSFT: This will leak, as `+new` returns a **+1**.

Frameworks/Foundation/NSUserDefaults.mm
```diff
@@ -48,6 +48,7 @@
 
 @implementation NSUserDefaults {
     StrongId<NSMutableDictionary> _registrationDict;
+    StrongId<NSOperationQueue> _synchronizeQueue;
 }
```
DHowett-MSFT: I'm somewhat wary of introducing a dependency on NSOperationQueue here; there might be some way to do this with dispatch groups or another primitive.
NSOperationQueue could spawn its own thread pool to service requests like this, but we don't need an entire thread sitting around dedicated to saving user preferences.

Frameworks/Foundation/NSUserDefaults.mm
```diff
@@ -245,6 +249,12 @@ - (void)setObject:(id)value forKey:(NSString*)key {
     CFPreferencesSetAppValue(static_cast<CFStringRef>(key), valueCopy, kCFPreferencesCurrentApplication);
 
     [[NSNotificationCenter defaultCenter] postNotificationName:NSUserDefaultsDidChangeNotification object:self];
+
+    if ([_synchronizeQueue operationCount] < 2) {
+        [_synchronizeQueue addOperationWithBlock:^void(void) {
```
DHowett-MSFT: It might be worth calling out why we're using 2 here.

DHowett-MSFT:
:clock1:

jofre-ms:

samples/WOCCatalog/WOCCatalog/UserDefaultsViewController.m
```diff
@@ -30,12 +30,13 @@ - (void)viewDidLoad {
     [super viewDidLoad];
 
     NSUserDefaults* standardDefaults = [NSUserDefaults standardUserDefaults];
-    NSDictionary* dict = @{@"FavoriteWord" : @"Blue"};
+    NSDictionary* dict = @{@"FavoriteWord" : @"Blue", @"FavoriteBool" : [NSNumber numberWithBool:YES]};
```
rajsesh-msft: >FavoriteBool [](start = 56, length = 12)

why this change?

Frameworks/Foundation/NSUserDefaults.mm
rajsesh-msft: the proper implementation of NSOperationQueue would use thread pools, but agree that our current implementation is not efficient.  

---
In reply to: [69786383](https://github.com/Microsoft/WinObjC/pull/612#discussion_r69786383) [](ancestors = 69786383)

rajsesh-msft:
:clock1:

samples/WOCCatalog/WOCCatalog/UserDefaultsViewController.m
```diff
@@ -51,7 +52,19 @@ - (void)viewDidLoad {
 
     [_textFields addObject:textField];
 
-    [self tableView].allowsSelection = NO;
+    // Slider for retrieving favorite bool
+    UISwitch* switchFavoriteBool = [[UISwitch alloc] initWithFrame:frame];
+    [switchFavoriteBool setOn:[[standardDefaults objectForKey:@"FavoriteBool"] boolValue] animated:NO];
```
jofre-ms: boolForKey

Frameworks/Foundation/NSUserDefaults.mm
```diff
@@ -245,6 +249,12 @@ - (void)setObject:(id)value forKey:(NSString*)key {
     CFPreferencesSetAppValue(static_cast<CFStringRef>(key), valueCopy, kCFPreferencesCurrentApplication);
```
rajsesh-msft: >CFPreferencesSetAppValue [](start = 4, length = 24)

we need to synchronize of the apps use CFPreferences* apis to set values.

jofre-ms: On the contrary, the documentation for  CFPreferencesSetAppValue states "You must call the CFPreferencesAppSynchronize function in order for your changes to be saved to permanent storage." - Therefore this automatic synchronize logic only needs to exist in NSUserDefaults.

---
In reply to: [69979853](https://github.com/Microsoft/WinObjC/pull/612#discussion_r69979853) [](ancestors = 69979853)

samples/WOCCatalog/WOCCatalog/UserDefaultsViewController.m
jofre-ms: Testing a bool strictly increases our test coverage. I wrote these changes because I couldn't answer the question of whether this scenario had regressed.

---
In reply to: [69818261](https://github.com/Microsoft/WinObjC/pull/612#discussion_r69818261) [](ancestors = 69818261)

Frameworks/Foundation/NSUserDefaults.mm
jofre-ms: _synchronizeQueue is a StrongId, which will release the object at the appropriate time.

---
In reply to: [69786087](https://github.com/Microsoft/WinObjC/pull/612#discussion_r69786087) [](ancestors = 69786087)

Frameworks/Foundation/NSUserDefaults.mm
jofre-ms: A quick examination of our NSOperationQueue implementation shows that we destroy the worker thread whenever the queue is empty and create a new thread when an operation is added. That's inefficient compared to using threadpools, but we should make the fix in NSOperationQueue rather than switching to a threadpool in NSUserDefaults because:

1) The performance gains in NSUserDefaults would be nominal compared to the already huge cost of writing to file
2) We would now need to write two fixes instead of one and maintain two threadpool/queue implementations instead of one.

---
In reply to: [69818599](https://github.com/Microsoft/WinObjC/pull/612#discussion_r69818599) [](ancestors = 69818599,69786383)

Frameworks/Foundation/NSUserDefaults.mm
jofre-ms: Looks like I need .attach

---
In reply to: [69994676](https://github.com/Microsoft/WinObjC/pull/612#discussion_r69994676) [](ancestors = 69994676,69786087)

Frameworks/Foundation/NSUserDefaults.mm
```diff
@@ -245,6 +249,12 @@ - (void)setObject:(id)value forKey:(NSString*)key {
...
 
     [[NSNotificationCenter defaultCenter] postNotificationName:NSUserDefaultsDidChangeNotification object:self];
+
+    if ([_synchronizeQueue operationCount] < 2) {
```
jofre-ms: Maybe this isn't the best way to determine if an operation should be added to the queue. I'm going to experiment with alternate designs and see if it makes my perf test any faster.

rajsesh-msft:
>     textField.textAlignment = UITextAlignmentRight;

Unrelated to this change, but this layout causes this page to be unusable on phone.  Can you make the text field left aligned? 

---
Refers to: samples/WOCCatalog/WOCCatalog/UserDefaultsViewController.m:49 in a37872a. [](commit_id = a37872a2d97035c48efb8527320124d8da4f5164, deletion_comment = False)

rajsesh-msft:
>     textField.backgroundColor = [UIColor blueColor];

can we just use the defaults instead?

---
Refers to: samples/WOCCatalog/WOCCatalog/UserDefaultsViewController.m:43 in a37872a. [](commit_id = a37872a2d97035c48efb8527320124d8da4f5164, deletion_comment = False)

samples/WOCCatalog/WOCCatalog/UserDefaultsViewController.m
```diff
@@ -30,12 +30,13 @@ - (void)viewDidLoad {
     [super viewDidLoad];
 
     NSUserDefaults* standardDefaults = [NSUserDefaults standardUserDefaults];
-    NSDictionary* dict = @{@"FavoriteWord" : @"Blue"};
+    NSDictionary* dict = @{@"FavoriteWord" : @"Blue", @"FavoriteBool" : [NSNumber numberWithBool:YES]};
     [standardDefaults registerDefaults:dict];
     NSString* placeHolder = [NSString stringWithFormat:@"Favorite Word: %@ (Type a new value here and relaunch!)", [standardDefaults stringForKey:@"FavoriteWord"]];
```
rajsesh-msft: >Favorite Word: %@ (Type a new value here and relaunch!) [](start = 57, length = 55)

This string will not fit on phone, please use the current value as the placeholder.

samples/WOCCatalog/WOCCatalog/UserDefaultsViewController.m
```diff
@@ -30,28 +30,41 @@ - (void)viewDidLoad {
...
     _textFields = [NSMutableArray new];
 
+    // Text box for retrieving favorite string
     CGRect frame = CGRectMake(c_originX, c_originY, c_width, c_height);
     UITextField* textField = [[UITextField alloc] initWithFrame:frame];
-    textField.textColor = [UIColor blackColor];
-    textField.backgroundColor = [UIColor blueColor];
-    textField.secureTextEntry = NO;
-    textField.font = [UIFont systemFontOfSize:17.0];
+//    textField.textColor = [UIColor blackColor];
```
jofre-ms: whoops, need to remove these

Frameworks/Foundation/NSUserDefaults.mm
```diff
@@ -56,14 +59,23 @@ @implementation NSUserDefaults {
 - (instancetype)init {
     [super init];
```
msft-Jeyaram: > [super init]; [](start = 3, length = 14)

if(self = [super init])

msft-Jeyaram: the call to super should be through the designated initializer.

samples/WOCCatalog/WOCCatalog/UserDefaultsViewController.m
```diff
@@ -30,28 +30,35 @@ - (void)viewDidLoad {
...
-    textField.textAlignment = UITextAlignmentRight;
-    textField.spellCheckingType = UITextSpellCheckingTypeNo;
+    textField.textAlignment = UITextAlignmentLeft;
     textField.delegate = self;
 
     [_textFields addObject:textField];
 
-    [self tableView].allowsSelection = NO;
+    // Slider for retrieving favorite bool
+    UISwitch* switchFavoriteBool = [[UISwitch alloc] initWithFrame:frame];
```
msft-Jeyaram: >switchFavoriteBool [](start = 14, length = 18)

needs to be released

msft-Jeyaram:
:clock1:

Frameworks/Foundation/NSUserDefaults.mm
```diff
@@ -119,16 +132,44 @@ - (void)setPersistentDomain:(id)domain forName:(NSString*)name {
  @Status Interoperable
 */
 - (BOOL)synchronize {
-    return CFPreferencesAppSynchronize(kCFPreferencesCurrentApplication);
+    _cacheLock.lock();
```
DHowett-MSFT: prefer `std::lock_guard` for scoped locking around mutices

jofre-ms: We need fine control over the lock here, as CFPreferencesAppSynchronize should be called outside of _cacheLock.

---
In reply to: [70540876](https://github.com/Microsoft/WinObjC/pull/612#discussion_r70540876) [](ancestors = 70540876)

Frameworks/Foundation/NSUserDefaults.mm
```diff
@@ -119,16 +132,44 @@ - (void)setPersistentDomain:(id)domain forName:(NSString*)name {
  @Status Interoperable
```
jofre-ms: >Interoperable [](start = 9, length = 13)

this should really be caveat - with a note that synchronize does not read from file

Frameworks/Foundation/NSUserDefaults.mm
DHowett-MSFT: Fortunately, `std::lock_guard` locks a scope, and you could introduce a scope for this.

---
In reply to: [70674087](https://github.com/Microsoft/WinObjC/pull/612#discussion_r70674087) [](ancestors = 70674087,70540876)

jofre-ms: ok then let's do it everywhere.

---
In reply to: [70694534](https://github.com/Microsoft/WinObjC/pull/612#discussion_r70694534) [](ancestors = 70694534,70674087,70540876)

Frameworks/Foundation/NSUserDefaults.mm
```diff
@@ -47,40 +47,68 @@
...
 */
 + (NSUserDefaults*)standardUserDefaults {
-    static NSUserDefaults* standard;
-
-    if (standard == nil) {
-        standard = [self new];
+    if (_standard == nil) {
+        _standard = [self new];
     }
```
mnithish: not thread safe

Frameworks/UIKit/UIApplicationMain.mm
```diff
@@ -320,6 +320,14 @@ void _UIApplicationShutdown() {
     [[UIApplication sharedApplication] _sendHighMemoryWarning];
 }
 
+extern "C" void UIApplicationMainHandleSuspendEvent() {
+    [[UIApplication sharedApplication] _sendSuspendEvent];
```
mnithish: >_sendSuspendEvent [](start = 39, length = 17)

nit: we are not sending any suspend event (unlike other methods) we are handling it.

mnithish: Same for resume below

---
In reply to: [70830139](https://github.com/Microsoft/WinObjC/pull/612#discussion_r70830139) [](ancestors = 70830139)

docs/Foundation/NSUserDefaults.md
```diff
@@ -0,0 +1,14 @@
...
+
+## **DEV DESIGN SPECIFICATION: NSUserDefaults
+
+### Feature Summary
+
+NSUserDefaults is a Foundation interface to CFPreferences that stores a preference plist in the package appdata directory.
+
+### Technology/Architecture Decisions and Caveats
+
+- Changes to standardDefaults are written to cache and synchronized to disk as soon as possible on an asyncrhonous queue.
```
mnithish: >asyncrhonous [](start = 102, length = 12)

nit

mnithish:
:clock1:

Frameworks/Foundation/NSUserDefaults.mm
```diff
@@ -47,40 +47,68 @@
...
+
 /**
  @Status Interoperable
 */
 + (NSUserDefaults*)standardUserDefaults {
-    static NSUserDefaults* standard;
-
-    if (standard == nil) {
-        standard = [self new];
+    if (_standard == nil) {
```
DHowett-MSFT: I think you should keep this as-is; use a magic static to force threadsafe init.
```
static NSUserDefaults* standard = [self new];
return standard;
```

C++11 ensures it only happens once.

Ideally, the suspend/resume methods would be per-instance; UIApplication would then call `[[NSUserDefaults standardUserDefaults] _zhuLiDoTheThing]`.

jofre-ms: The reason I avoided the [NSUserDefaults standardUserDefaults] call is to prevent initialization of standardUserDefaults when we suspend, if it hasn't already been initialized. I'll try a different workaround.

---
In reply to: [70878479](https://github.com/Microsoft/WinObjC/pull/612#discussion_r70878479) [](ancestors = 70878479,70878446)

Frameworks/Foundation/NSUserDefaults.mm
```diff
@@ -232,8 +302,32 @@ - (double)doubleForKey:(NSString*)defaultName {
...
+}
+
+- (NSOperationQueue*) _synchronizeQueue {
+    return _synchronizeQueue;
+}
+
+- (void) _suspendSynchronize {
+    [_synchronizeQueue setSuspended:YES];
+    [_synchronizeQueue waitUntilAllOperationsAreFinished];
+}
```
DHowett-MSFT: What happens if we suspend it, then another thread injects a synchronize, and then we try to wait on an operation that'll never complete (since it was suspended)?

mnithish: That should not a problem because, _suspendSynchronize is called just before the application is being suspended, so the other thread will also get suspended and when the application resumes _resumeSynchronize will get called on the UI thread and that will resume the other thread from the wait.

---
In reply to: [70900447](https://github.com/Microsoft/WinObjC/pull/612#discussion_r70900447) [](ancestors = 70900447)

DHowett-MSFT: Discussed offline; Doing this should suffice:
```
[_synchronizeQueue addOperations:@[[NSBlockOperation blockOperationWithBlock:^{ [_synchronizeQueue suspend]; }]] waitUntilFinished:YES];
```

---
In reply to: [70908191](https://github.com/Microsoft/WinObjC/pull/612#discussion_r70908191) [](ancestors = 70908191,70900447)

DHowett-MSFT:
In CI build.

PR #613, opened 2016-07-06
Add ability to suspend/resume NSOperationQueue

bbowman:
Add ability to suspend/resume NSOperationQueue

Description:
Adds the ability to use the suspended property on operation queue. This was previously implmented but not actually hooked up to stop operations from running while suspended. Adds a new UT to check this case.

### DIFF
--- a/Frameworks/Foundation/NSOperationQueue.mm
+++ b/Frameworks/Foundation/NSOperationQueue.mm
@@ -196,6 +196,12 @@ static BOOL RunOperationFromLists(NSAtomicListRef* listPtr, NSAtomicListRef* sou
         didWork = FALSE;
 
         for (int i = 0; i < NSOperationQueuePriority_Count; i++) {
+            [priv->suspendedCondition lock];
+            while (priv->isSuspended) {
+                [priv->suspendedCondition wait];
+            }
+            [priv->suspendedCondition unlock];
+
             if (RunOperationFromLists(&priv->myQueues[i], (NSAtomicListRef*)(&priv->queues[i]), &priv->curOperation)) {
                 didWork = TRUE;
             }
--- a/tests/unittests/Foundation/NSOperationTests.mm
+++ b/tests/unittests/Foundation/NSOperationTests.mm
@@ -17,8 +17,13 @@
 #include <TestFramework.h>
 #import <Foundation/Foundation.h>
 
+#import <thread>
+#import <mutex>
+#import <condition_variable>
+#import <chrono>
+
 TEST(NSOperation, NSOperationDealloc) {
-    NSOperationQueue *queue = [[NSOperationQueue alloc] init];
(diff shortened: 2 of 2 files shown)


DHowett-MSFT:
:shipit:

rajsesh-msft:
:shipit:

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -196,6 +196,12 @@ - (NSOperationQueue*)_doMainWork {
         didWork = FALSE;
 
         for (int i = 0; i < NSOperationQueuePriority_Count; i++) {
+            [priv->suspendedCondition lock];
+            while (priv->isSuspended) {
+                [priv->suspendedCondition wait];
+            }
```
MSFTFox: Is a while loop really the best way to hold onto a suspended state?

DHowett-MSFT: A while loop plus a condition variable (which this is) is the best way to hold for the suspended state. Consider that we signal on `suspendedCondition` for both suspend and resume, and we need to only hold operation processing when `isSuspended` is true.

---
In reply to: [69805759](https://github.com/Microsoft/WinObjC/pull/613#discussion_r69805759) [](ancestors = 69805759)

bbowman: [http://web.stanford.edu/class/cs140/cgi-bin/lecture.php?topic=locks](http://web.stanford.edu/class/cs140/cgi-bin/lecture.php?topic=locks) may be useful reading. A while on a condition variable is a pretty typical pattern to wait until the desired condition is met. Remember that it will broadcast on any set (not just a set to false from true).

---
In reply to: [69805759](https://github.com/Microsoft/WinObjC/pull/613#discussion_r69805759) [](ancestors = 69805759)

bbowman: Hooray saying the same things at roughly the same time.

---
In reply to: [69806160](https://github.com/Microsoft/WinObjC/pull/613#discussion_r69806160) [](ancestors = 69806160,69805759)

DHowett-MSFT: Additionally, more primitive condition variables can be prone to spurious wakes. :)

---
In reply to: [69806286](https://github.com/Microsoft/WinObjC/pull/613#discussion_r69806286) [](ancestors = 69806286,69805759)

MSFTFox: Cool, makes sense. I remember a similar situation some time ago where I was told not to use a while loop on a condition. I was just curious as to the reasoning. Thanks. :)

---
In reply to: [69806286](https://github.com/Microsoft/WinObjC/pull/613#discussion_r69806286) [](ancestors = 69806286,69805759)

MSFTFox:
:shipit:

PR #617, opened 2016-07-07
Added NULL path support to CGPath methods.

tcunning:
The currently implemented CGPath* methods now support NULL paths.  The previous implementation would crash if the supplied path was NULL.  This change is helpful in preventing crashing in Objective-C code that may pass nil paths to these methods.

### DIFF
--- a/Frameworks/CoreGraphics/CGPath.mm
+++ b/Frameworks/CoreGraphics/CGPath.mm
@@ -187,6 +187,10 @@ CGMutablePathRef CGPathCreateMutable() {
  @Notes Creates a mutable copy
 */
 CGPathRef CGPathCreateCopy(CGPathRef path) {
+    if( path == NULL ) {
+        return NULL;
+    }
+
     return CGPathCreateMutableCopy(path);
 }
 
@@ -194,6 +198,10 @@ CGPathRef CGPathCreateCopy(CGPathRef path) {
  @Status Interoperable
 */
 CGMutablePathRef CGPathCreateMutableCopy(CGPathRef path) {
+    if( path == NULL ) {
+        return NULL;
+    }
+
     auto ret = __CGPath::alloc(nil);
     ret->_max = path->_max;
     ret->_count = path->_count;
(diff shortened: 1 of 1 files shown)


mnithish:

rajsesh-msft:
:shipit:

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -187,13 +187,21 @@ CGMutablePathRef CGPathCreateMutable() {
  @Notes Creates a mutable copy
 */
 CGPathRef CGPathCreateCopy(CGPathRef path) {
+    if( path == NULL ) {
```
rajsesh-msft: >( [](start = 6, length = 1)

the formatting of all the if statements doesnt follow the conventions.  Can you please make sure you have run clang formatter?  (ex: it should be if (path == NULL) instead of if( path == NULL )

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -390,6 +406,10 @@ void CGPathAddArc(CGMutablePathRef path,
  @Status Interoperable
 */
 void CGPathMoveToPoint(CGMutablePathRef path, const CGAffineTransform* m, float x, float y) {
+    if (path == NULL) {
```
msft-Jeyaram: nit: you could use the RETURN_IF_NULL() macros.

tcunning: I didn't find a RETURN_IF_NULL macro available in the project.  I only found the following matching NULL macros which aren't appropriate for these cases as we aren't returning HRESULT:

```
  C:\WinObjC\Frameworks\include\wil\result.h(823):#define RETURN_IF_NULL_ALLOC(ptr)                  \
  C:\WinObjC\Frameworks\include\wil\result.h(904):#define RETURN_IF_NULL_ALLOC_LOG(ptr)                  \
  C:\WinObjC\Frameworks\include\wil\result.h(985):#define RETURN_IF_NULL_ALLOC_MSG(ptr, fmt, ...)                          \
  C:\WinObjC\Frameworks\include\wil\result.h(1061):#define RETURN_IF_NULL_ALLOC_EXPECTED(ptr) \
  Matching lines: 4    Matching files: 1    Total files searched: 4039
```

mnithish:
:shipit:

rajsesh-msft:
CI build passed.

PR #619, opened 2016-07-07
Basic ReplacePathWithStrokedPath Implementation

tcunning:
This initial implementation leverages cairo_copy_path_flat that returns a copy of the current path with all line elements.  The underlying elements for the path may be drastically different because all the curves will be replaced by line elements. Everything else should be similar to iOS behavior, such as order of the points, so functionality like an animation along a path should behave in a similar fashion.

Endcap styles are supported.  Line join styles are not yet implemented.

The included sample is dependent on pull request #617


rajsesh-msft:
@tcunning will there be a new PR for this?

tcunning:
Yes, this issue was resolved.

PR #625, opened 2016-07-08
Basic ReplacePathWithStrokedPath Implementation

tcunning:
This initial implementation leverages cairo_copy_path_flat that returns a
copy of the current path with all line elements. The underlying elements
for the path may be drastically different because all the curves will be
replaced by line elements. Everything else should be similar to iOS
behavior, such as order of the points, so functionality like an animation
along a path should behave in a similar fashion.

Endcap styles are supported. Line join styles are not yet implemented.


mnithish:

Frameworks/CoreGraphics/CGContextCairo.mm
```diff
@@ -1629,6 +1630,229 @@
     UNLOCK_CAIRO();
 }
 
+void CGContextCairo::_CGContextDrawStrokedPath(cairo_path_t* flatPath, int i, int prev, int total, CGFloat lineWidth) {
```
msft-Jeyaram: null check for flatPath? we are doing a memory reference.

Frameworks/include/CGContextImpl.h
```diff
@@ -154,6 +155,7 @@ __CGCONTEXTIMPL_TEST_FRIENDS;
     virtual void CGContextSetLineCap(DWORD lineCap);
     virtual void CGContextSetLineWidth(float width);
     virtual void CGContextSetShouldAntialias(DWORD shouldAntialias);
+    virtual void CGContextReplacePathWithStrokedPath(CGContextRef context);
```
msft-Jeyaram: could this not be made purely virtual to make things a bit more easier?

Frameworks/CoreGraphics/CGContextCairo.mm
DHowett-MSFT: This is an internal implementation, I think. Are we sure that the callers adequately ensure it's non-NULL/

msft-Jeyaram: ^indeed.

mnithish: Null check is not required here as the path is always obtained from cairo which guarantees returning a valid pointer, but a check for flatPath->status would be nice to have if for some reason the cairo context used to retrieve the path was in an error state

---
In reply to: [70112840](https://github.com/Microsoft/WinObjC/pull/625#discussion_r70112840) [](ancestors = 70112840)

tcunning: This is an internal method.  It's called by CGContextReplacePathWithStrokedPath which uses cairo_copy_path_flat which always returns a valid pointer, but the result may have no data (data==NULL and num_data==0).  So I think we are good w/o a NULL check.

Frameworks/CoreGraphics/CGContextCairo.mm
```diff
@@ -1629,6 +1630,229 @@
...
+    // Try to get next and previous types.
+    cairo_path_data_type_t nextType = CAIRO_PATH_CLOSE_PATH;
+    cairo_path_data_type_t prevType = CAIRO_PATH_CLOSE_PATH;
+    if (moreData) {
+        nextType = flatPath->data[nextIndex].header.type;
+    }
+    if (i > 0) {
+        prevType = flatPath->data[prev].header.type;
+    }
+    switch (data->header.type) {
```
rajsesh-msft: >switch [](start = 4, length = 6)

This function does too many things, has too many locals and too little comment, so it is hard to exactly understand what is going on.  Could you please:

1. Refactor the code here and break this function down into smaller pieces based on the path element type.
2. Add comments as needed.

rajsesh-msft:

Frameworks/CoreGraphics/CGContextCairo.mm
```diff
@@ -1629,6 +1630,229 @@
...
+            if (moreData) {
+                _CGContextDrawStrokedPath(flatPath, nextIndex, i, total, lineWidth);
+                // return the current point to the saved point.
+                curPathPosition.x = lastPos.x;
+                curPathPosition.y = lastPos.y;
+            }
+            break;
+        }
+        case CAIRO_PATH_CURVE_TO:
+            // There shouldn't be curves in this path. Move on.
```
rajsesh-msft: >There shouldn't be curves in this path. Move on [](start = 15, length = 47)

Should that be an error condition?  Perhaps a warning trace is needed.

Frameworks/CoreGraphics/CGContextCairo.mm
```diff
@@ -1629,6 +1630,229 @@
...
+            // Don't do a move here, it'll move to the center of our stroke.
+            // Save the current position for later. This will help us close the current sub path
+            CGPoint lastPos = curPathPosition;
+            // Set the current position
+            curPathPosition.x = data[1].point.x;
+            curPathPosition.y = data[1].point.y;
+
+            // If there's more data process it
+            if (moreData) {
+                _CGContextDrawStrokedPath(flatPath, nextIndex, i, total, lineWidth);
```
rajsesh-msft: >_CGContextDrawStrokedPath [](start = 16, length = 25)

I am not sure recursion is the right approach here, when iteration will work and will be lot more deterministic with stack space.

Frameworks/CoreGraphics/CGContextCairo.mm
```diff
@@ -1629,6 +1630,229 @@
...
+    if (i > 0) {
+        prevType = flatPath->data[prev].header.type;
+    }
+    switch (data->header.type) {
+        case CAIRO_PATH_MOVE_TO: {
+            // Don't do a move here, it'll move to the center of our stroke.
+            // Save the current position for later. This will help us close the current sub path
+            CGPoint lastPos = curPathPosition;
+            // Set the current position
+            curPathPosition.x = data[1].point.x;
```
rajsesh-msft: >curPathPosition.x = data[1].point.x [](start = 12, length = 35)

nit: you could simply do curPathPosition = data[1].point;

rajsesh-msft:
It is important to have some unit tests for this API.  Granted, it won't behave same on reference platform, but will be helpful to catch regressions if we make changes in the future.

samples/CGCatalog/CGCatalog/Samples/CGCCGContextReplacePathWithStrokedPath.m
```diff
@@ -50,36 +57,86 @@ - (void)drawRect:(CGRect)rect {
...
+        CGPathAddQuadCurveToPoint(path, NULL, 100, 200, 200, 300);
+    }
+
     // Log original path info
     [self.logs addObject:@"Original Path:"];
     params = @{ @"logs" : self.logs, @"points" : originalPoints };
     CGPathApply(path, (__bridge void*)params, _Applier);
 
     // Create stroked path
     // TODO: Change this to use CGContextReplacePathWithStrokedPath()
```
rajsesh-msft: >CGContextReplacePathWithStrokedPath [](start = 32, length = 35)

This TODO can be removed?

samples/CGCatalog/CGCatalog/Samples/CGCCGContextReplacePathWithStrokedPath.m
```diff
@@ -50,36 +57,86 @@ - (void)drawRect:(CGRect)rect {
...
+        case StrokedPathTypeLine: {
+            CGContextSetStrokeColorWithColor(context, [UIColor greenColor].CGColor);
+            CGContextAddPath(context, path);
+            CGContextStrokePath(context);
+            points = originalPoints;
+        } break;
+
+        case StrokedPathTypeCopyStroke: {
+            NSMutableArray<NSValue*>* strokedPoints = [NSMutableArray new];
+            CGPathRef newPath = CGPathCreateCopyByStrokingPath(path, NULL, c_LineWidth, kCGLineCapRound, kCGLineJoinMiter, 1);
```
rajsesh-msft: >CGPathCreateCopyByStrokingPath [](start = 32, length = 30)

This is returning null and crashing later on.

samples/CGCatalog/CGCatalog/Samples/CGCCGContextReplacePathWithStrokedPath.m
```diff
@@ -50,36 +57,86 @@ - (void)drawRect:(CGRect)rect {
...
+        CGPathAddArcToPoint(path, NULL, maxWidth / 2, maxHeight * 1.25, 0, maxHeight / 4, 50);
+        CGPathAddLineToPoint(path, NULL, maxWidth / 4, maxHeight / 2);
+        CGPathAddLineToPoint(path, NULL, 10, maxHeight / 2);
+    } else {
+        CGPathMoveToPoint(path, NULL, 200, 200);
+        CGPathAddLineToPoint(path, NULL, 260, 100);
+        CGPathAddArcToPoint(path, NULL, 500, 500, 450, 450, 50);
+        CGPathAddQuadCurveToPoint(path, NULL, 100, 200, 200, 300);
+    }
+
```
rajsesh-msft: You could create two paths in viewDidLoad and simply use them here, instead of recreating the default paths for every frame.

rajsesh-msft: Issues #630 => the points actually don't show.  Is the sample behavior tested on winobjc?

samples/CGCatalog/CGCatalog/Samples/CGCCGContextReplacePathWithStrokedPath.m
```diff
@@ -50,36 +57,86 @@ - (void)drawRect:(CGRect)rect {
...
+            [self.logs addObject:@"Replace Stroke:"];
+            params = @{ @"logs" : self.logs, @"points" : strokedPoints };
+            CGPathApply(newPath, (__bridge void*)params, _Applier);
+
+            if (self.strokedPathType == StrokedPathTypeReplaceStroke)
+                CGContextSetFillColorWithColor(context, [UIColor orangeColor].CGColor);
+            else
+                CGContextSetFillColorWithColor(context, [UIColor purpleColor].CGColor);
+
+            CGContextStrokePath(context);
```
rajsesh-msft: >CGContextStrokePath [](start = 12, length = 19)

nit: You need a CGContextFillPath(), or it wont be filled with the right color.

samples/CGCatalog/CGCatalog/Samples/CGCCGContextReplacePathWithStrokedPath.m
```diff
@@ -50,36 +57,86 @@ - (void)drawRect:(CGRect)rect {
...
+            CGContextReplacePathWithStrokedPath(context);
+
+            CGPathRef newPath = CGContextCopyPath(context);
+            NSMutableArray<NSValue*>* strokedPoints = [NSMutableArray new];
+            [self.logs addObject:@"Replace Stroke:"];
+            params = @{ @"logs" : self.logs, @"points" : strokedPoints };
+            CGPathApply(newPath, (__bridge void*)params, _Applier);
+
+            if (self.strokedPathType == StrokedPathTypeReplaceStroke)
+                CGContextSetFillColorWithColor(context, [UIColor orangeColor].CGColor);
```
rajsesh-msft: >CGContextSetFillColorWithColor [](start = 16, length = 30)

nit: here and elsewhere, our coding conventions require {} on conditionals.

Frameworks/CoreGraphics/CGContextCairo.mm
jeffMeador: Yes we could put a warning trace here. We know in this situation that there are no curves, but a warning could help someone in the future.

Frameworks/CoreGraphics/CGContextCairo.mm
jeffMeador: Any way I think about this it requires 2 loops because of the way I need to traverse the original path. Then I'd need to either redo or save any calculations needed in both loops. Is that a fair tradeoff?

rajsesh-msft: What you have here is a state machine, you should be able to transition state and iterate once through the points. 

---
In reply to: [70330348](https://github.com/Microsoft/WinObjC/pull/625#discussion_r70330348) [](ancestors = 70330348)

samples/CGCatalog/CGCatalog/Samples/CGCCGContextReplacePathWithStrokedPath.m
tcunning: Yea now that's everything is synced it, we can actually call CGContextReplacePathWithStrokedPath

samples/CGCatalog/CGCatalog/Samples/CGCCGContextReplacePathWithStrokedPath.m
tcunning: Yea, CGPathCreateCopyByStrokingPath isn't yet supported.   Added a NULL check and TODO to remove it once implemented.

samples/CGCatalog/CGCatalog/Samples/CGCCGContextReplacePathWithStrokedPath.m
tcunning: The points do not show as CGContextAddArc is unimplemented.  I will change this to draw little squares

samples/CGCatalog/CGCatalog/Samples/CGCCGContextReplacePathWithStrokedPath.m
tcunning: fixed!

tcunning:
The last push was to address some of the feedback for the CGCatalog example.

Frameworks/CoreGraphics/CGContextCairo.mm
jeffMeador: Every point in the array needs to be translated to its stroked point on one side of the the stroke then once we're at the end of the array we have to go back through every point in the reverse order and create the stroked point for the other side of the stroke. Creating the stroked points on either side of each point in order would produce intermediate lines like in the line breakdown example image from one of my emails.

jeffMeador: Actually it might not necessarily create intermediate lines, but the underlying order of elements will not match the reference platform and that would affect things like animations. There could be other side effects I'd have to dig up.

jeffMeador:
@rajsesh-msft I committed the current state of the looping implementation with d28fbc781becf0d6e48c0292343a56e839de9158. This should be in a similar state as the recursive implementation we had earlier.

Frameworks/CoreGraphics/CGContextCairo.mm
```diff
@@ -1633,6 +1634,358 @@
...
+    CGFloat halfStrokeWidth = (curState->lineWidth) / 2.0f;
+
+    // Calculates points on stroked path from original points
+    // Reverse flag should be true if we're traversing the path backwards
+    auto strokedPoint = [](CGPoint p, CGFloat angle, CGFloat distance, bool reverse) -> CGPoint {
+        if (reverse) {
+            angle += M_PI;
+        }
+        CGFloat x = p.x - (sin(angle) * distance);
+        CGFloat y = p.y + (cos(angle) * distance);
```
bbowman: perhaps this my lack of knowledge about Core Graphics in general but it seems like this is using a weird coordinate system for delta x to be -sin  and delta y to be cos. If you are doing something "tricky" here like automatically adjusting for the normal of the input angle (+ PI/2) , then it deserves comments.

Frameworks/CoreGraphics/CGContextCairo.mm
```diff
@@ -1633,6 +1634,358 @@
...
+            CGFloat m1 = tan(angle1);
+            CGFloat b1 = p1.y - (m1 * p1.x);
+
+            CGFloat m2 = tan(angle2);
+            CGFloat b2 = p2.y - (m2 * p2.x);
+
+            CGFloat x = (b2 - b1) / (m1 - m2);
+            CGFloat y = 0;
+            // If m1 is very large it's probably a vertical line
+            if (m1 > 1000000) {
```
bbowman: seems a little arbitrary? How was this constant and optimization decided on?

Frameworks/CoreGraphics/CGContextCairo.mm
```diff
@@ -1633,6 +1634,358 @@
...
+                y = m2 * x + b2;
+            } else {
+                y = m1 * x + b1;
+            }
+
+            // TODO: This code comes directly from CGContextAddArcToPoint
+            // Consolidate and reuse instead of copy
```
bbowman: should we file a github issue for this todo and add the number here?

bbowman:
:clock1:

Frameworks/CoreGraphics/CGContextCairo.mm
jeffMeador: The number is arbitrary, `m1 > 1000000` can be replaced with `angle1 == M_PI / 2 || angle1 == M_PI * 3 / 2` (with float approximations).

Alan-Python: Date: Tue, 26 Jul 2016 06:50:54 -0700

Frameworks/CoreGraphics/CGContextCairo.mm
jeffMeador: Right, this should be equivalent to:

`CGFloat x = p.x + (cos(angle - M_PI_2) * distance);`
`CGFloat y = p.y + (sin(angle - M_PI_2) * distance);`

Would you prefer this implementation so it appears more straightforward instead of comments?

Frameworks/CoreGraphics/CGContextCairo.mm
jeffMeador: Yes, this code is copied a couple different spots, but it also needs refactoring/commenting to make it clear what it's doing. The other spot it was copied to has a similar todo comment.

rajsesh-msft:
Given our plan to move to D2D for coregraphics, we can get D2D to do this instead of hand rolled implementation.  Closing this PR as such.

Issue #622, opened 2016-07-07
DatePicker graphical corruption


shahmeeresmail:
In WOCCATALOG on Windows, the date picker has a checkered background for date and striped background for time. On iOS it has a white background for both.

This is observed in the master and dev branches.

Below is a cropped screenshot showing the issue.
![datepickercorruption](https://cloud.githubusercontent.com/assets/17790501/16666811/1db985c0-443e-11e6-9739-4c7575df2643.png)

Ramu-msft:
The fix is available in 0.1 Preview (July 15, 2016) release.

PR #636, opened 2016-07-12
UI Image fixes

shahmeeresmail:
Fixes graphical corruption in UI elements (examples: UIDatePicker, UISegmentControl) resulting in a checkerboard pattern being drawn and optimized patched draw function. The corruption is what is seen in issue #622 and is visible in the Segments and DatePicker sections of WOCCATALOG. This also resolves issue #674 since it implements the use of blendMode and alpha for the various drawAtPoint and drawInRect functions.

Details:
- Instead of creating a pattern and calling a CB function (drawNinePatchCallback) to subdivide image into patches and draw individually, the change was made to just call a function which subdivides the images into patches and draws (drawPatches)
- Insets improperly scaled in many places
- drawPatches optimized to subdivide the image into the smallest number of patches to reduce draw setup and state management overhead
- Depending on what insets are set, 1-9 patched draws may be needed. Optimized drawPatches function to do 1 to 9 draws instead of 3 to 9 draws (effectively reducing min draws from 3 to 1) and minimize the number of draws used.
- Make makeRect and drawPatches functions inline


Frameworks/UIKit/UIImage.mm
```diff
@@ -714,102 +710,108 @@ - (void)drawAsPatternInRect:(CGRect)pos {
...
-                                            ii->size.size.width - img->_imageInsets.left - img->_imageInsets.right,
-                                            ii->size.size.height - img->_imageInsets.top - img->_imageInsets.bottom));
+                                   cgImg,
+                                   makeRect(leftCap, botCap, (src->size.width - leftCap - rightCap), (src->size.height - topCap - botCap)),
+                                   makeRect((dst->origin.x + leftCap),
+                                            (dst->origin.y + botCap),
+                                            (dst->size.width - leftCap - rightCap),
+                                            (dst->size.height - topCap - botCap)));
         } else {
             assert(0);
```
msft-Jeyaram: > assert(0); [](start = 11, length = 11)

although this is old code, it would be much better to Throw and exception with some information on why this is an issue.
It would provide much more information rather than just asserting.

Frameworks/UIKit/UIImage.mm
```diff
@@ -714,102 +710,108 @@ - (void)drawAsPatternInRect:(CGRect)pos {
...
+            CGContextDrawImageRect(context,
+                                   cgImg,
+                                   makeRect((src->size.width - rightCap), botCap, rightCap, (src->size.height - topCap - botCap)),
+                                   makeRect((dst->origin.x + dst->size.width - rightCap),
+                                            (dst->origin.y + botCap),
+                                            rightCap,
+                                            (dst->size.height - topCap - botCap)));
+        }
+    } else {
+        assert(0);
```
msft-Jeyaram: >assert(0); [](start = 8, length = 10)

same as the comment on the earlier assert.

Frameworks/UIKit/UIImage.mm
shahmeeresmail: Replaced assert with unimplemented_with_msg macro.

Frameworks/UIKit/UIImage.mm
shahmeeresmail: Replaced assert with unimplemented_with_msg macro.

rajsesh-msft:
@Ramu-msft is added to the review. #Closed

rajsesh-msft:

Frameworks/UIKit/UIImage.mm
```diff
@@ -714,102 +710,108 @@ - (void)drawAsPatternInRect:(CGRect)pos {
...
+            CGContextDrawImageRect(context,
+                                   cgImg,
+                                   makeRect((src->size.width - rightCap), botCap, rightCap, (src->size.height - topCap - botCap)),
+                                   makeRect((dst->origin.x + dst->size.width - rightCap),
+                                            (dst->origin.y + botCap),
+                                            rightCap,
+                                            (dst->size.height - topCap - botCap)));
+        }
+    } else {
+        UNIMPLEMENTED_WITH_MSG("Patched draws only supported when sum of topCap and botCap is less than the height of the UI element.");
```
davelamb: This doesn't seem to be the case on the reference platform. It seems to draw the caps as a ratio between the cap sizes.

shahmeeresmail: That's true and this is a pre-existing issue. I will probably create another pull request in the future to address this and potentially a functional test for this scenario to make sure it matches the output on the reference platform.

Frameworks/UIKit/UIImage.mm
```diff
@@ -1033,59 +1035,41 @@ - (void)drawInRect:(CGRect)pos {
...
-
-        CGPatternRelease(pat);
-    }
+    CGRect srcRect;
+    srcRect.origin = { 0 };
+    srcRect.size = { static_cast<float>(imgBacking->Width()), static_cast<float>(imgBacking->Height()) };
 
-    CGContextRestoreGState(cur);
+    // Draw image and divide into patches if necessary
+    drawPatches(ctx, self, &srcRect, &dstRect);
```
davelamb: Your comment about things being incorrectly scaled isn't exactly correct. An image named <something>@2x.<ext> should have a scale of 2.0, and when rendered should be half the size of logical pixels. Try it on iOS, something like this (photo1.png exists in WOCCatalog):

UIImage* image = [UIImage imageNamed:@"photo1.png"];
image = [UIImage imageWithCGImage:image.CGImage scale:4.0f orientation:UIImageOrientationUp];
image = [image1 resizableImageWIthCapInsets:UIEdgeInsetsMake(50, 50, 50, 50)];

Try drawing this image (from a UIImageView or drawRect) and you'll see that the image caps shrink (the size is the same, but the image is rendered smaller) according to the scale. This is the same behaviour as when using a CALayer with contents = image.CGImage, contentsGravity = kCAGravityCenter and contentsScale = 4.0.

davelamb: That is to say, in the previous iteration there was a * _scale in there that has to remain. 

---
In reply to: [70727868](https://github.com/Microsoft/WinObjC/pull/636#discussion_r70727868) [](ancestors = 70727868)

davelamb:
We've root caused the crashing issue to a missing call to EMMS (_mm_empty, clears MMX state) and the tiling issue to x/y step size being incorrect in the call to CGPatternCreate. I have a small fix that addresses both these issues that we'd like to push out that shouldn't collide (but for one line) with your change.

Frameworks/UIKit/UIImage.mm
shahmeeresmail: With the original scaling this is what the datepicker output looks like:
originalscaling

With the scale factor not applied to the dest rect this is what is displayed:
reviseddrawrect

In the first case it seems as though the image is overdrawn or scaled twice or something similar. In the second one it fits within the bounds.

I will try to replicate the scenario you described sometime tomorrow but in this particular case, keeping the original scaling does not seem to be the way to go.

jaredhms:
> }

delete and just call CGRectMake?

---
Refers to: Frameworks/UIKit/UIImage.mm:948 in 532b33e. [](commit_id = 532b33e68ad931ec90882b591204d6088b9fde9d, deletion_comment = False)

jaredhms:
> static void drawLeftCap(UIImage* self, CGContextRef cur, CGRect pos) {

Can we delete this?

---
Refers to: Frameworks/UIKit/UIImage.mm:821 in 532b33e. [](commit_id = 532b33e68ad931ec90882b591204d6088b9fde9d, deletion_comment = False)

jaredhms:
> static void drawTopCap(UIImage* self, CGContextRef cur, CGRect pos) {

Can we delete this?

---
Refers to: Frameworks/UIKit/UIImage.mm:883 in 532b33e. [](commit_id = 532b33e68ad931ec90882b591204d6088b9fde9d, deletion_comment = False)

jaredhms:
> static void drawLeftAndTopCap(UIImage* self, CGContextRef ctx, CGRect rect) {

Can we delete this?

---
Refers to: Frameworks/UIKit/UIImage.mm:950 in 532b33e. [](commit_id = 532b33e68ad931ec90882b591204d6088b9fde9d, deletion_comment = False)

shahmeeresmail:
I don't see why we can't delete the drawTopCap or drawLeftAndTopCap methods. They are unused but I originaly left it alone because I wasn't sure if it was envisioned to use it as part of a future change.

shahmeeresmail:
I will also remove makeRect and substitute existing references to it with CGRectMake

Frameworks/UIKit/UIImage.mm
```diff
@@ -671,38 +667,45 @@ - (void)draw1PartImageInRect:(CGRect)pos {
...
-    pos.size.width = ((float)getImage(self)->Backing()->Width() / _scale);
-    pos.size.height = ((float)getImage(self)->Backing()->Height() / _scale);
+    pos.size.width = ((float)imgBacking->Width());
+    pos.size.height = ((float)imgBacking->Height());
 
     srcRect.origin.x = 0;
-    srcRect.origin.y = float(getImage(self)->Backing()->Height());
-    srcRect.size.width = float(getImage(self)->Backing()->Width());
-    srcRect.size.height = -((float)getImage(self)->Backing()->Height());
+    srcRect.origin.y = float(imgBacking->Height());
```
jaredhms: >float( [](start = 23, length = 6)

nit: it's a bid odd that two of these are constructing a float rather than casting to a float

jaredhms:
In CI build.

jaredhms:
:shipit:

Issue #594, opened 2016-06-29
Update Bitmap Contexts to support more than just RGB colorspaces


jparmentervf:
The current underlying CGContextRefs only support RGB colorspaces. Any other color space causes a fast fail. Updating the underlying APIs will mean improvements to CGImageGetBitmapInfo and better support for functionality in CGBitmapContextCreate

shahmeeresmail:
I have created a [pull request ](https://github.com/Microsoft/WinObjC/pull/641)to fix other issues but it also partially resolves this one:
- Full support for: CGImageGetBitmapInfo, CGBitmapContextGetBitmapInfo, CGImageGetAlphaInfo and CGBitmapContextGetAlphaInfo, CGColorspaceGetColorspaceModel, CGColorSpaceGetNumberOfComponents amongst others
- CGBitmapContextCreate and CGImageCreate properly use colorSpace, bitmapInfo and bitsPerComponent. BytesPerRow still has issues with Cairo when the input data has padded pixels and those haven't been resolved yet but the infrastructure has been set up to at least pass the value down.
- RGB and Gray colorspaces can be created and used
- Provides some support for indexed and pattern colorspaces but more work is needed here
- ColorFormat is decoupled from the colorSpace and instead the colorSpace is coupled with CGColorspaceModel (decoupling of ColorFormat necessary to properly support bitmapInfo at create time and implement bitmapInfo and alphaInfo accessor functions)

rajsesh-msft:
We will cover this as part of the CG rewrite using D2D.  Adding additional colorspaces would be a feature work.

Issue #640, opened 2016-07-13
CoreGraphics: CGColorSpace incorrectly coupled with ColorFormat


shahmeeresmail:
Right now when a color space is created, the format is determined. The format effectively also specifies the byteOrder and alphaPosition (which are both flags of bitmapInfo). However, imageCreate or contextCreate, bitmapInfo is also specified and there may be a conflict. If there is a conflict between the colorSpace format and the bitmapInfo then the colorSpace format will prevail (incorrectly).

ColorSpace should be paired with a colorSpaceModel at ColorSpaceCreate time and the ColorFormat should be determined at imageCreate or contextCreate time after taking into consideration: colorSpace, bitsPerPixel, bitsPerComponent and bitmapInfo (effectively alphaPosition + byteOrder + floatIntSelectorBit)

rajsesh-msft:
@shahmeeresmail thanks for reporting and fixing this!

Issue #639, opened 2016-07-13
CoreGraphics: ByteOrder & AlphaPosition not implemented correctly


shahmeeresmail:
- During context or image create, byteOrder is ignored and alpha position is used incorrectly to determine how the imageBacking is created. 
- Alpha position incorrectly determined and returned in context/image accessor functions
- ColorFormat naming scheme inconsistently uses bigEndian and littleEndian to indicate byte ordering and in some cases specifies the wrong byte ordering completely.

rajsesh-msft:
@shahmeeresmail thanks for reporting and fixing this.

PR #641, opened 2016-07-13
CoreGraphicsRework

shahmeeresmail:
Set of changes to rework CoreGraphics to resolve many bugs and implement some new features. Specifically this addresses issues #639 and #640 but also partially resolves issue #594 and others.

Details:
- Renamed internal context create functions to match their use and meet formatting standards
    - Created CGSurfaceInfo structure which combines all inputs needed initialize CGImageData, CGGraphicBufferImage,  CGGraphicBufferImageBacking, CGBitmapImage, and CGBitmapImageBacking objects in one structure.
    - Decouple format from ColorSpace. CGColorSpaceModel now used to initialize __CGColorSpace objects instead of format and functions looking for an image's or context's format have been reworked to look elsewhere. This (partially) fixes CGColorSpaceCreate and colorspace property retrieval functions (CGColorSpaceCreateCalibratedRGB, CGColorSpaceCreateIndexed, CGColorSpaceCreateDeviceRGB, CGColorSpaceCreatePattern, CGColorSpaceGetModel, CGColorSpaceGetNumberOfComponents, CGColorSpaceCreateDeviceGray, CGColorSpaceCreateWithName, CGBitmapContextGetColorSpace, CGImageGetAlphaInfo, CGImageGetColorSpace, CGimageGetBitsPerComponent, CGImageGetBitsPerPixel, GBimageGetBitmapInfo)
    - Improved support for CGBitmapContextCreate and CGImageCreate. Now supports byte order, alpha position and colorspace parameters correctly.
    - Implemented CGBitmapContextGetBitmapInfo, CGBitmapContextGetAlphaInfo
    - Created helper function for getting format from image creation parameters (_CGImageGetFormat)
    - Moved surfaceFormat, renderingIntent, colorspaceModel, surfaceInfo, byteOrder, and alphaInfo enums to CoreGrpahicsExport.h so as to create a safe include file for cpp files using CoreGraphics structures and built using the MS compiler.
    - Renamed surface formats to reflect their actual storage in memory and to uniformly use big endian notation similar to that used in CoreGraphics documentation (example: _ColorARGB represents a format with byte ordering {b0 = B, b1 = G, b2 = R, b3 = A} )
 - Updated CoreGraphics test cases and AccelerateViewController in WOCCATALOG to properly use these changes


rajsesh-msft:

rajsesh-msft:

Frameworks/UIKit/UIColor.mm
```diff
@@ -504,7 +505,7 @@ - (UIColor*)initWithPatternImage:(UIImage*)image {
 callbacks.releaseInfo = 0;
 callbacks.drawPattern = __UIColorPatternFill;
 
-_pattern = (id) CGPatternCreateColorspace(self, bounds, m, bounds.size.width, bounds.size.height, 0, FALSE, &callbacks, pImg->_has32BitAlpha ? _ColorRGBA : _ColorRGB);
+_pattern = (id) CGPatternCreateColorspace(self, bounds, m, bounds.size.width, bounds.size.height, 0, FALSE, &callbacks, pImg->_has32BitAlpha ? _ColorABGR : _ColorBGR);
```
rajsesh-msft: This code is under #if 0, we should probably just delete this.

Frameworks/include/CGSurfaceInfoInternal.h
```diff
@@ -0,0 +1,57 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#ifndef __CGSURFACEINFOINTERNAL_H
+#define __CGSURFACEINFOINTERNAL_H
```
rajsesh-msft: please use #pragma once instead.

Frameworks/UIKit/UIColor.mm
shahmeeresmail: This change got hit as part of the surface enum renaming. I did it elsewhere for similar code that was commented out or compiled out via ifdefs just for consistency.

Frameworks/include/CGSurfaceInfoInternal.h
shahmeeresmail: Done.

include/CoreGraphics/CGImage.h
```diff
@@ -106,3 +86,8 @@ COREGRAPHICS_EXPORT bool CGImageIsMask(CGImageRef image) STUB_METHOD;
 
 COREGRAPHICS_EXPORT size_t CGImageGetHeight(CGImageRef image);
 COREGRAPHICS_EXPORT size_t CGImageGetWidth(CGImageRef image);
+
+COREGRAPHICS_EXPORT inline surfaceFormat _CGImageGetFormat(unsigned int bitsPerComponent,
```
rajsesh-msft: >inline [](start = 20, length = 6)

inline + export doesn't make a lot of sense, should this not be inline?

include/CoreGraphics/CoreGraphicsExport.h
```diff
@@ -66,5 +67,94 @@ typedef struct __CGPath* CGMutablePathRef;
...
+    _ColorMax,
+    _ColorRGBA = _ColorABGR
+} surfaceFormat;
+
+typedef struct {
+    CGColorSpaceModel colorSpaceModel;
+    CGBitmapInfo bitmapInfo;
+    unsigned int bitsPerComponent;
+    unsigned int bytesPerPixel;
+} __CGFormatProperties;
```
rajsesh-msft: Shouldn't these __ structs be internal, they should not be in the export header?

samples/WOCCatalog/WOCCatalog/AccelerateViewController.m
```diff
@@ -326,8 +326,10 @@ - (UIImage*)transformImage:(UIImage*)image {
     // create CGImageRef from vImage_Buffer output
     CGColorSpaceRef colorSpace = CGColorSpaceCreateDeviceRGB();
 
-    CGContextRef ctx = CGBitmapContextCreate(
-        outBuffer.data, outBuffer.width, outBuffer.height, 8, outBuffer.rowBytes, colorSpace, (CGBitmapInfo)kCGImageAlphaNoneSkipLast);
+    const CGBitmapInfo bitmapInfo = kCGImageAlphaNoneSkipLast | kCGBitmapByteOrder32Little;
```
rajsesh-msft: >kCGBitmapByteOrder32Little [](start = 64, length = 26)

This seems like a breaking change - if the apps now have to use kCGBitmapByteOrder32Little.

include/CoreGraphics/CoreGraphicsExport.h
shahmeeresmail: I kept them here because they are used internally by many different components and effectively allows other components to just import CoreGraphicsExport.h in order to make use of them. I can put them in a separate file if that works. Perhaps CoreGraphicsInternal.h?

rajsesh-msft: Yes, please! We have done the *Internal.h pattern in other places as well. 

---
In reply to: [71207581](https://github.com/Microsoft/WinObjC/pull/641#discussion_r71207581) [](ancestors = 71207581)

Frameworks/CoreGraphics/CGBitmapContext.mm
```diff
@@ -48,25 +51,40 @@ CGImageRef CGBitmapContextCreateImage(CGContextRef context) {
...
+@Notes
 */
 CGBitmapInfo CGBitmapContextGetBitmapInfo(CGContextRef context) {
-	CGImageRef imageRef = context->Backing()->DestImage();
-	CGBitmapInfo bitmapInfo = CGImageGetBitmapInfo(imageRef);
-	return bitmapInfo;
+    int ret;
+
+    if (context) {
+        ret = context->Backing()->DestImage()->Backing()->BitmapInfo();
```
rajsesh-msft: >Backing [](start = 23, length = 7)

Nit: probably makes sense to add BitMapInfo() method to Backing context so we don't have this chain here.

samples/WOCCatalog/WOCCatalog/AccelerateViewController.m
shahmeeresmail: BitmapInfo (including byteOrder flags) is intended to indicate what the underlying format is. The way it is processed is:
- A byte order of 32big is assumed (ie RGB)
- Use alphaInfo to determine: whether there is an alpha component or padding variable present and if so what its position is and place it in the pixel. In this case and at this point we would have: RGBX
- Use the byteOrder to determine if it is different from the default value and whether or not the order needs to be flipped. In this case, kCGBitmapByteOrder32Little would indicate that the byte order would need to be flipped so RGBX becomes XBGR.

Frameworks/include/CGSurfaceInfoInternal.h
```diff
@@ -0,0 +1,53 @@
...
+
+static inline __CGSurfaceInfo _CGSurfaceInfoInit(size_t width,
+                                                 size_t height,
+                                                 surfaceFormat fmt,
+                                                 void* data = NULL,
+                                                 size_t bytesPerRow = 0,
+                                                 CGBitmapInfo bitmapInfo = c_kCGBitmapInfoInvalidBits) {
+    __CGSurfaceInfo surfaceInfo;
+
+    assert(fmt < _ColorMax);
```
rajsesh-msft: nit: We prefer FAIL_FAST over assert.

samples/WOCCatalog/WOCCatalog/AccelerateViewController.m
shahmeeresmail: BitmapInfo is something apps should already be specifying when creating contexts or images.

rajsesh-msft: Understood, this code however didn't need kCGBitmapByteOrder32Little before, so this will break apps that don't specify it?

---
In reply to: [71218160](https://github.com/Microsoft/WinObjC/pull/641#discussion_r71218160) [](ancestors = 71218160)

include/CoreGraphics/CGImage.h
shahmeeresmail: Good catch. I will put this function with the other helpers in CGSurfaceInfoInternal.h

include/CoreGraphics/CoreGraphicsExport.h
```diff
@@ -66,5 +67,94 @@ typedef struct __CGPath* CGMutablePathRef;
...
+    { kCGColorSpaceModelMonochrome, (kCGImageAlphaNone | kCGBitmapByteOrderDefault), 8, 1 }, //_ColorGrayscale,
+    { kCGColorSpaceModelRGB, (kCGImageAlphaNone | kCGBitmapByteOrderDefault), 8, 3 }, //_ColorBGR,
+    { kCGColorSpaceModelPattern, (kCGImageAlphaOnly | kCGBitmapByteOrderDefault), 8, 1 }, //_ColorA8,
+    { kCGColorSpaceModelIndexed, (kCGImageAlphaNone | kCGBitmapByteOrderDefault), 8, 2 }, //_ColorIndexed,
+};
+
+struct __CGSurfaceInfo {
+    size_t width;
+    size_t height;
+    size_t bitsPerComponent;
```
rajsesh-msft: >bitsPerComponent [](start = 11, length = 16)

Some of the fields here are similar to __CGFormatProperties, does it make sense to use that struct here?  Also, size_t usage here is different from the other struct. Note that size_t is different on different architectures (ex: 64 bit).

samples/WOCCatalog/WOCCatalog/AccelerateViewController.m
shahmeeresmail: In this scenario we got lucky because there were multiple bugs which resulted in the misuse of bitmapInfo working correctly but purely by accident:
- CGBitmapContextCreate ignored byteOrdering so it didn't matter whether it was set correctly
- RGB Colorspaces were incorrectly coupled with the format BGRA and interpreted as such. So by specifying any RGB colorspace in CGBitmapContextCreate and any byteOrder value, BGRA was automatically implied no matter what. Since in the app is not using alpha and it really wants BGRX anyway, BGRA would operate almost identically to what it would expect it to.

include/CoreGraphics/CoreGraphicsExport.h
shahmeeresmail: I agree that the type should be consistent. Will make them all unsigned ints for now since the imageData backings use that type as well and it is highly unlikely we will have dimensions or individual pixel bit/byte sizes in the millions :)

shahmeeresmail: I can also use CGFormatProperties here too

include/CoreGraphics/CoreGraphicsExport.h
```diff
@@ -66,5 +67,94 @@ typedef struct __CGPath* CGMutablePathRef;
...
+    _ColorABGR,
+    _ColorBGRX,
+    _ColorXBGR,
+    _ColorGrayscale,
+    _ColorBGR,
+    _ColorA8,
+    _ColorIndexed,
+    _ColorMax,
+    _ColorRGBA = _ColorABGR
+} surfaceFormat;
```
rajsesh-msft: >surfaceFormat [](start = 2, length = 13)

Nit: This enum is an odd man out.  Could we make it __CGSurfaceFormat while we are at this?

Frameworks/include/CGNoDataImage.h
```diff
@@ -42,7 +45,11 @@ class CGNoDataImageBacking : public CGImageBacking {
     int Height();
     int BytesPerRow();
     int BytesPerPixel();
+    int BitsPerComponent();
```
rajsesh-msft: nit: would be useful to add the override annotation to the overridden methods.

Frameworks/CoreGraphics/CGGraphicBufferImage.mm
```diff
@@ -31,71 +33,49 @@
...
-CGGraphicBufferImage::CGGraphicBufferImage(DWORD width, DWORD height, surfaceFormat fmt) {
-    _img = new CGBitmapImageBacking(width, height, fmt);
+CGGraphicBufferImage::CGGraphicBufferImage(int width, int height, surfaceFormat fmt) {
+    __CGSurfaceInfo surfaceInfo = _CGSurfaceInfoInit(width, height, fmt);
+
+    _img = new CGBitmapImageBacking(&surfaceInfo);
+    _img->_parent = this;
+    _imgType = CGImageTypeBitmap;
+
+    _img->_parent = this;
```
rajsesh-msft: this can be factored into a private method and reused in the constructors.

Frameworks/CoreGraphics/CGGraphicBufferImage.mm
```diff
@@ -174,6 +179,7 @@
     int stride;
     _imageData = _nativeTextureLocking->LockWritableBitmapTexture(_nativeTexture, &stride);
     _bytesPerPixel = 4;
+    _bitsPerComponent = 8;
```
rajsesh-msft: This seems like a strange place to change _bitsPerComponent.

Frameworks/include/CGGraphicBufferImage.h
```diff
@@ -20,8 +21,9 @@
 
 class CGGraphicBufferImage : public __CGImage {
 public:
-    CGGraphicBufferImage(DWORD width, DWORD height, surfaceFormat fmt);
-    CGGraphicBufferImage(DWORD width, DWORD height, surfaceFormat fmt, DisplayTexture* nativeTexture, DisplayTextureLocking* locking);
+    CGGraphicBufferImage(int width, int height, surfaceFormat fmt);
+    CGGraphicBufferImage(__CGSurfaceInfo* surfaceInfo);
```
rajsesh-msft: >__CGSurfaceInfo [](start = 25, length = 15)

const _CGSurfaceInfo* (since you are not changing it).  Altertnatively const __CGSurfaceInfo& so you could do this like new CGGraphicsBufferImage(_CGInitSurface(...)).

include/CoreGraphics/CoreGraphicsExport.h
rajsesh-msft: __CGSurfaceInfo : public __CGFormatProperties might work out well.

---
In reply to: [71232435](https://github.com/Microsoft/WinObjC/pull/641#discussion_r71232435) [](ancestors = 71232435)

Frameworks/CoreGraphics/CGColorSpace.mm
```diff
@@ -91,70 +92,77 @@ CGColorSpaceRef CGColorSpaceCreateIndexed(id baseSpace, int lastIndex, void* col
...
- @Status Interoperable
+@Status Interoperable
 */
-CGColorSpaceRef CGColorSpaceCreateDeviceGray() {
-    return (CGColorSpaceRef) new __CGColorSpace(_ColorGrayscale);
+CGColorSpaceModel CGColorSpaceGetModel(CGColorSpaceRef colorSpace) {
+    if (colorSpace) {
+        return ((__CGColorSpace*)colorSpace)->colorSpaceModel;
+    } else {
+        TraceWarning(TAG, L"CGColorSpaceGetModel: Null colorspace. Returning RGB");
```
rajsesh-msft: >TraceWarning [](start = 8, length = 12)

This is probably the expected behavior on reference platform?  Might as well remove the warning in that case.

Frameworks/UIKit/WYPopoverController.m
```diff
@@ -70,9 +70,17 @@ - (BOOL)getValueOfRed:(CGFloat *)red green:(CGFloat *)green blue:(CGFloat *)blue
...
-    CGColorSpaceModel colorSpaceModel = CGColorSpaceGetModel(colorSpace);
-    CGColorSpaceRelease(colorSpace);
+
+    if (colorSpace != NULL) {
+        colorSpaceModel = CGColorSpaceGetModel(colorSpace);
+        CGColorSpaceRelease(colorSpace);
+    } else {
+        // Default to RGB in the absence of a colorSpace from CGColor
+        colorSpaceModel = kCGColorSpaceModelRGB;
+    }
```
rajsesh-msft: colorSpace NULL is already handled in CGColorSpaceGetModel(), so it should be safe to revert this file?

Frameworks/CoreGraphics/CGColorSpace.mm
```diff
@@ -91,70 +92,77 @@ CGColorSpaceRef CGColorSpaceCreateIndexed(id baseSpace, int lastIndex, void* col
...
+    const char* strName = static_cast<const char*>([(__bridge NSString*)name UTF8String]);
 
-        case _ColorGrayscale:
-            return 1;
-
-        default:
-            assert(0);
-            return 0;
+    if (strcmp(strName, "kCGColorSpaceGenericRGB") == 0) {
+        return (CGColorSpaceRef) new __CGColorSpace(kCGColorSpaceModelRGB);
```
rajsesh-msft: It probably makes sense to keep a dictionary (CFDictionary) of __CGColorSpace instances and return the instance instead of creating new ones everytime.  Clearly, this is just an optimization.

include/CoreGraphics/CoreGraphicsExport.h
shahmeeresmail: Absolutely :) The only reason I didn't do this earlier is because I didn't want to make this PR larger than it already was.

Frameworks/CoreGraphics/CGDiscardableImage.mm
```diff
@@ -105,9 +109,22 @@
     return _forward->BytesPerPixel();
 }
 
+int CGDiscardableImageBacking::BitsPerComponent() {
+    ConstructIfNeeded();
```
rajsesh-msft: Once we construct the image once, we should probably cache all of these properties, so if the bitmap is discarded, we can get the info without reconstructing it.  I see that the pattern is preexisting though, can the surface change after the image is discarded (and then re-used?).

rajsesh-msft:
> }

CGDiscardableImageBacking::ReleaseImageData() -> should not ConstructIfNeeded().  If _forward is null, there is nothing to release.. (Not related to your change though).

---
Refers to: Frameworks/CoreGraphics/CGDiscardableImage.mm:179 in faef665. [](commit_id = faef6650dc9ae1945d8d3a727833e64774ef4b9f, deletion_comment = False)

Frameworks/CoreGraphics/CGImage.mm
```diff
@@ -751,12 +703,14 @@ CGImageRef CGImageCreate(size_t width,
         }
 
         newImage->Backing()->ReleaseImageData();
-    } else {
-        newImage = new CGBitmapImage(width, height, ((__CGColorSpace*)colorSpace)->colorSpace, data);
     }
 
     newImage->_provider = dataProvider;
 
+    if (colorSpaceAllocated == true) {
```
rajsesh-msft: >if (colorSpaceAllocated == true) { [](start = 4, length = 34)

nit: this could simply be if (colorSpaceAllocated)

rajsesh-msft:
:clock1:

samples/WOCCatalog/WOCCatalog/AccelerateViewController.m
shahmeeresmail: I looked into this a little more and although, I have implemented the default case (where no byte order is specified), I have implemented to default to 32big where it should default to 32 little. It is a small change and I will update it and remove the change in this file since it isn't needed.

Frameworks/CoreGraphics/CGGraphicBufferImage.mm
shahmeeresmail: I verified that indeed there is no reason why bitsPerComponent nor bytesPerPixel needs to be set here or in BytesPerRow where it is also overridden.

samples/WOCCatalog/WOCCatalog/AccelerateViewController.m
rajsesh-msft: Thats perfect.  There are other files where this change as well.  

---
In reply to: [71264537](https://github.com/Microsoft/WinObjC/pull/641#discussion_r71264537) [](ancestors = 71264537)

Frameworks/ImageIO/CGImageSource.mm
```diff
@@ -1718,7 +1718,7 @@ CGImageRef CGImageSourceCreateImageAtIndex(CGImageSourceRef isrc, size_t index,
                                         32,
                                         frameWidth * 4,
                                         colorspaceRgb,
-                                        kCGImageAlphaFirst,
+                                        (kCGImageAlphaLast | kCGBitmapByteOrder32Little),
```
rajsesh-msft: >kCGBitmapByteOrder32Little [](start = 61, length = 26)

We shouldn't need this flag by default, per the other change?

shahmeeresmail: That is correct. I initially put it here to be more explicit but for consistency and thoroughness,, I reverted use of kCGBitmapByteOrder32Little here and elsewhere where it should be assumed anyway and retested the code once more.

rajsesh-msft:
in ci build.

rajsesh-msft:
@shahmeeresmail the ImageIO unittests (several of them) fail, could you please take a look at the failures?

shahmeeresmail:
Raj, the failures being reported were due to stale values of CGImageAlphaInfo and CGBitmapInfo being checked against. I updated the test with the correct values and made a few small changes in the test to make it easier to debug failures.

rajsesh-msft:
back in CI build.

rajsesh-msft:
CI build passed. Validated with test apps.  :shipit:

PR #642, opened 2016-07-13
Improving the duration of the scheduled timer tests.

msft-Jeyaram:
This needs to be verified on a different machine. As I had issue with timer not going off for the specified duration. 
@mnithish is trying it.

### DIFF
--- a/tests/unittests/Foundation/NSTimerTests.mm
+++ b/tests/unittests/Foundation/NSTimerTests.mm
@@ -136,7 +136,7 @@ TEST(NSTimer, ScheduledTimerWithTimeIntervalRepeat) {
     volatile long waitCompletion = 0;
 
     auto selectorCalledAsync = std::async(std::launch::async, [&waitCompletion, testObj]() {
-        BOOL fooChanged = [testObj waitOnCalledConditionForInterval:5] && [testObj called];
+        BOOL fooChanged = [testObj waitOnCalledConditionForInterval:4] && [testObj called];
         _InterlockedExchange(&waitCompletion, 1L);
         return fooChanged;
     });
--- a/tests/unittests/Foundation/NSTimerTests.mm
+++ b/tests/unittests/Foundation/NSTimerTests.mm
@@ -14,8 +14,8 @@
 //
 //******************************************************************************
 
-#import "Starboard.h"
-#include <TestFramework.h>
+#import <Starboard.h>
+#import <TestFramework.h>
 #import <Foundation/Foundation.h>
 #import <future>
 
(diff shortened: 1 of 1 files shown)


tests/unittests/Foundation/NSTimerTests.mm
```diff
@@ -133,19 +132,18 @@ - (void)dealloc {
...
     volatile long waitCompletion = 0;
 
-    auto selectorCalledAsync = std::async(std::launch::async,
-                                          [&waitCompletion, testObj]() {
-                                              BOOL fooChanged = [testObj waitOnCalledConditionForInterval:15] && [testObj called];
-                                              _InterlockedExchange(&waitCompletion, 1L);
-                                              return fooChanged;
-                                          });
+    auto selectorCalledAsync = std::async(std::launch::async, [&waitCompletion, testObj]() {
+        BOOL fooChanged = [testObj waitOnCalledConditionForInterval:5] && [testObj called];
```
msft-Jeyaram: I can reduce this to 3/4

DHowett-MSFT:
My only concern with this is that we're making heavy-handed tests more likely to fail if the machine is under stress. Is there a way to write these tests that doesn't require a timed waiter and a timed waitee?
If we were to have a timed waitee and an un-timed waiter (say: holding a mutex, or waiting on a condition), we might be able to avoid stressed systems failing these tests.

msft-Jeyaram:
True, a system under stress can delay the timers. Although we seem to have timed test pattern in some other tests in our code base (e.g NSObject_PerformSelector* tests). I'll look into this.

---
In reply to: [232461993](https://github.com/Microsoft/WinObjC/pull/642#issuecomment-232461993) [](ancestors = 232461993)

mnithish:
I agree with Dustin, I am more than confident these will fail in our nightly runs

---
In reply to: [232463598](https://github.com/Microsoft/WinObjC/pull/642#issuecomment-232463598) [](ancestors = 232463598,232461993)

tests/unittests/Foundation/NSTimerTests.mm
```diff
@@ -105,16 +105,15 @@ - (void)dealloc {
...
+    auto selectorCalledAsync = std::async(std::launch::async, [&waitCompletion, testObj]() {
+        BOOL fooChanged = [testObj waitOnCalledConditionForInterval:2] && [testObj called];
+        _InterlockedExchange(&waitCompletion, 1L);
+        return fooChanged;
+    });
 
     NSRunLoop* runLoop = [NSRunLoop currentRunLoop];
 
-    NSTimer* timer = [NSTimer scheduledTimerWithTimeInterval:10 target:testObj selector:@selector(testFunction) userInfo:nil repeats:NO];
+    NSTimer* timer = [NSTimer scheduledTimerWithTimeInterval:1 target:testObj selector:@selector(testFunction) userInfo:nil repeats:NO];
```
MSFTFox: Might be worth it to put some of these time intervals and other numbers at the top as statics to make updating this test easier in the future as well.

tests/unittests/Foundation/NSTimerTests.mm
```diff
@@ -105,16 +105,15 @@ - (void)dealloc {
...
     volatile long waitCompletion = 0;
 
-    auto selectorCalledAsync = std::async(std::launch::async,
-                                          [&waitCompletion, testObj]() {
-                                              BOOL fooChanged = [testObj waitOnCalledConditionForInterval:15] && [testObj called];
-                                              _InterlockedExchange(&waitCompletion, 1L);
-                                              return fooChanged;
-                                          });
+    auto selectorCalledAsync = std::async(std::launch::async, [&waitCompletion, testObj]() {
+        BOOL fooChanged = [testObj waitOnCalledConditionForInterval:2] && [testObj called];
```
bbowman: >2 [](start = 68, length = 1)

in terms of unit tests 2 seconds is still a long time. Perhaps consider something in the ms range.

bbowman:
>     }

busy waits are about my least favorite mechanism to do multithreaded tests. Especially since this is hard to get correct as coded. I forget all of the details on volatile and would much prefer us to use a condition variable, std::atomic or other syncronization object.

---
Refers to: tests/unittests/Foundation/NSTimerTests.mm:128 in db05926. [](commit_id = db059265324b4921a4a77bb73b9f324f278f4bb4, deletion_comment = False)

bbowman:
:clock1:

rajsesh-msft:
Is anything happening with this PR?

msft-Jeyaram:
this will be obsolete with the use of CFRunLoopTimer as most of the time based tests are removed to use conditional checks.

rajsesh-msft:
Thanks @msft-Jeyaram in that case, please close this pull request.

Issue #657, opened 2016-07-16
CGColorEqualsColor() always returns true


rajsesh-msft:
https://github.com/Microsoft/WinObjC/commit/8c9d1b85e30a6c3a538464d730082c8e6bbab9ff

/**
 @Status Interoperable
*/
bool CGColorEqualToColor(CGColorRef color1, CGColorRef color2) {
    ColorQuad components1{};
    ColorQuad components2{};

    [(UIColor*)color1 getColors:&components1];
    [(UIColor*)color2 getColors:&components2];

    return ((components1.r == components1.r) && (components1.g == components1.g) && (components1.b == components1.b) &&
            (components1.a == components1.a));
}

rajsesh-msft:
That's the bug.  Fix is part of the PR #658

PR #658, opened 2016-07-16
UIColor now stores ColorQuad (which has same layout as an array of 4 …

### LINKED ISSUE #652: Leak in CGColorGetComponents()
According to the Core Foundation "Get Rule", CGColorGetComponents() should not return an object that must be freed by the caller, but in fact it returns a newly allocated object:
https://github.com/Microsoft/WinObjC/blob/d9578b016df6cfc3822ff409904b69544a2a6ec6/Frameworks/CoreGraphics/CGColor.mm#L120

rajsesh-msft:
…floats).  This can then be used by CGColor.

The correct long terms solution will involve CGColor holding the colors and UIColor using it, but this is a first step towards that.  In the process, also cleaned up the internal methods and removed some unnecessary copies.

Also found and fixed bug in CGColorEqualsColor.

Fix #652, #657

### DIFF
--- a/Frameworks/CoreGraphics/CGColor.mm
+++ b/Frameworks/CoreGraphics/CGColor.mm
@@ -62,11 +62,9 @@ CGColorRef CGColorCreateCopy(CGColorRef color) {
  @Status Interoperable
 */
 CGColorRef CGColorCreateCopyWithAlpha(CGColorRef color, float alpha) {
-    ColorQuad curColor;
+    const ColorQuad* curColor = [(UIColor*)color _getColors];
 
-    [(UIColor*)color getColors:&curColor];
-
-    return static_cast<CGColorRef>([[_LazyUIColor colorWithRed:curColor.r green:curColor.g blue:curColor.b alpha:alpha] retain]);
+    return static_cast<CGColorRef>([[_LazyUIColor colorWithRed:curColor->r green:curColor->g blue:curColor->b alpha:alpha] retain]);
 }
 
 /**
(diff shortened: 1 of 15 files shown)


rajsesh-msft:

rajsesh-msft:

rajsesh-msft:

Frameworks/CoreGraphics/CGColor.mm
```diff
@@ -62,11 +62,9 @@ CGColorRef CGColorCreateCopy(CGColorRef color) {
  @Status Interoperable
 */
 CGColorRef CGColorCreateCopyWithAlpha(CGColorRef color, float alpha) {
-    ColorQuad curColor;
+    const ColorQuad* curColor = [(UIColor*)color _getColors];
 
-    [(UIColor*)color getColors:&curColor];
-
-    return static_cast<CGColorRef>([[_LazyUIColor colorWithRed:curColor.r green:curColor.g blue:curColor.b alpha:alpha] retain]);
+    return static_cast<CGColorRef>([[_LazyUIColor colorWithRed:curColor->r green:curColor->g blue:curColor->b alpha:alpha] retain]);
```
bbowman: >_LazyUIColor [](start = 37, length = 12)

:-( The inverteed dependency here still makes me so sad.

bbowman:
>     _components = new float[count * componentCount];

would prefer an RAII construct for this guy.

---
Refers to: Frameworks/CoreGraphics/CGGradient.mm:75 in a0b7360. [](commit_id = a0b73609050ebda364a6860840a09faa6fbf597b, deletion_comment = False)

bbowman:
:shipit:

mnithish:
:shipit:

Frameworks/CoreGraphics/CGColor.mm
rajsesh-msft: Understood, this will be part of the rework we do for CoreGraphics in the near future.

---
In reply to: [71058697](https://github.com/Microsoft/WinObjC/pull/658#discussion_r71058697) [](ancestors = 71058697)

rajsesh-msft:
>     _components = new float[count * componentCount];

Won't fix for now.

---
In reply to: [233101767](https://github.com/Microsoft/WinObjC/pull/658#issuecomment-233101767) [](ancestors = 233101767)

---
Refers to: Frameworks/CoreGraphics/CGGradient.mm:75 in a0b7360. [](commit_id = a0b73609050ebda364a6860840a09faa6fbf597b, deletion_comment = False)

rajsesh-msft:
In CI build.

PR #665, opened 2016-07-19
Removed Copyright notice from sample app

tcunning:
Removed Copyright notice from sample app per request.


rajsesh-msft:
In CI build.

Issue #662, opened 2016-07-18
NSFileManager needs unit tests


msft-Jeyaram:
Most of the functionality does not have unit tests, also edge cases are not being handled properly (e.g nil, etc).
- File clean up for dead code and logs.

DHowett-MSFT:
@msft-Jeyaram: Leaving this open (referenced via e9bb1b2) to track additional unit test work.

rajsesh-msft:
NSFileManager tests should use the newly added _SCOPE (compatible with OSX)

rajsesh-msft:
Closing this issue.  We will write tests as we add features here.

Issue #653, opened 2016-07-15
[NSFileManager moveItemAtURL:toURL:error] not working.


rupertdaniel:
Moving a file using `[NSFileManager moveItemAtURL:toURL:error]` does not work.

Example code:
```
NSString *path = [NSSearchPathForDirectoriesInDomains(NSApplicationSupportDirectory, NSUserDomainMask, YES) firstObject];
	
NSURL *file1 = [NSURL fileURLWithPath:[path stringByAppendingPathComponent:@"test1.txt"]];
NSURL *file2 = [NSURL fileURLWithPath:[path stringByAppendingPathComponent:@"test2.txt"]];

[@"test string" writeToURL:file1 atomically:NO encoding:NSUTF8StringEncoding error:nil];

NSError *error = nil;
BOOL result = [[NSFileManager defaultManager] moveItemAtURL:file1 toURL:file2 error:&error];
if (!result) {
	NSLog(@"Failed to move file: %@", error);
}
```

The test file `test1.txt` is written, but the move fails and the following is outputted to the console:
```
Failed to move file: (null)
```

Also `[NSFileManager moveItemAtPath:toPath:error]` does not work.

PR #669, opened 2016-07-19
Add support for moveItemAtURL and unit tests.

msft-Jeyaram:
#653  #662

### DIFF
--- a/Frameworks/Foundation/NSFileManager.mm
+++ b/Frameworks/Foundation/NSFileManager.mm
@@ -1,8 +1,8 @@
 //******************************************************************************
 //
-// Copyright (c) Microsoft Corporation. All rights reserved.
+// COPYRIGHT (C) MICROSOFT. ALL RIGHTS RESERVED.
 //
-// This code is licensed under the MIT License (MIT).
+// THIS CODE IS LICENSED UNDER THE MIT LICENSE (MIT).
 //
 // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
 // IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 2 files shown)


Frameworks/Foundation/NSFileManager.mm
```diff
@@ -436,36 +436,48 @@ - (BOOL)copyItemAtPath:(id)srcPath toPath:(id)destPath error:(NSError**)error {
...
 /**
- @Status Stub
+ @Status Caveat
+ @Notes does not support file manager move delegates for shouldMoveItem or shouldProceedAfterError.
 */
 - (BOOL)moveItemAtURL:(NSURL*)srcURL toURL:(NSURL*)dstURL error:(NSError**)error {
-    UNIMPLEMENTED();
-
-    return NO;
+    return [self moveItemAtPath:[srcURL path] toPath:[dstURL path] error:error];
```
DHowett-MSFT: This might do well to check that the file URLs are file URLs; it doesn't make sense to move anything else.

Frameworks/Foundation/NSFileManager.mm
```diff
@@ -436,36 +436,48 @@ - (BOOL)copyItemAtPath:(id)srcPath toPath:(id)destPath error:(NSError**)error {
...
+            *error = [NSError errorWithDomain:NSCocoaErrorDomain code:NSFileNoSuchFileError userInfo:nil];
+        }
+
+        return NO;
+    }
+
     const char* src = [srcPath UTF8String];
     const char* dest = [destPath UTF8String];
 
     TraceVerbose(TAG, L"Moving %hs to %hs", src, dest);
```
DHowett-MSFT: >TraceVerbose(TAG, L"Moving %hs to %hs", src, dest); [](start = 4, length = 51)

We can probably destroy this log.

tests/unittests/Foundation/NSFileManagerTests.mm
```diff
@@ -136,4 +170,76 @@
...
+    EXPECT_OBJCEQ_MSG(str, content, "FAILED: Unable to read the file content.");
+
+    [fh closeFile];
+
+    deleteFile((status ? destName : srcName));
+}
+
+TEST(NSFileManager, MoveFileViaURL) {
+    char const* srcName = "\\NSFileManagerMoveTestFile.txt";
+    char const* destName = "\\MovedFile.txt";
```
DHowett-MSFT: > "\\ [](start = 26, length = 4)

The slashes here worry me; we should be using something like `-[NSString stringByAddingPathComponent:]` and its friends rather than relying on the input string to have a \ at the beginning.

tests/unittests/Foundation/NSFileManagerTests.mm
```diff
@@ -14,18 +14,52 @@
...
+    int len = GetModuleFileNameA(NULL, fullPath, _MAX_PATH);
+    char* ptrToLastEntry = strrchr(fullPath, '\\');
+    int lengthOfLastEntry = strlen(ptrToLastEntry);
+    strncpy_s(ptrToLastEntry, _MAX_PATH - (len - lengthOfLastEntry), filePath, strlen(filePath) + 1);
+
+    return [NSString stringWithFormat:@"%s", fullPath];
+}
+
+static void createFileWithContentAndVerify(char const* filePath, NSString* content) {
+    NSString* fullPath = getPathToFile(filePath);
```
DHowett-MSFT: This entire function could be [content writeToFile:xxx atomically:NO];

tests/unittests/Foundation/NSFileManagerTests.mm
```diff
@@ -14,18 +14,52 @@
...
+#import <Foundation/NSDate.h>
+#import <Foundation/NSURL.h>
+#import <Foundation/NSFileHandle.h>
+#import <stdlib.h>
+#import <windows.h>
+
+static NSString* getPathToFile(char const* filePath) {
+    char fullPath[_MAX_PATH];
+    int len = GetModuleFileNameA(NULL, fullPath, _MAX_PATH);
+    char* ptrToLastEntry = strrchr(fullPath, '\\');
```
DHowett-MSFT: pre-cache this, and simply use `[_cachedModulePath stringByAppendingPathComponent:filePathWhichIsNowAnNSString]`

DHowett-MSFT:
:clock1:

DHowett-MSFT:
If you explicitly say `Fixes #xxx.` or `Fix #xxx.` in the commit log, it will automatically close the issue when the pull request is merged!

Frameworks/Foundation/NSFileManager.mm
```diff
@@ -302,18 +302,18 @@ - (BOOL)createDirectoryAtPath:(NSString*)pathAddr
                 bool success = EbrMkdir(curPath);
                 if (!success && errno != EEXIST) {
                     TraceError(TAG, L"Failed to make path %hs: %d", curPath, errno);
-                    // return FALSE;
+                    // return NO;
```
MSFTFox: Is there a reason this is commented out and not just removed? Also, it seems like the error parameter is unused. This is probably outside the scope of this review, but could you take a look at it and file issues if necessary or follow up with somebody?

msft-Jeyaram: there are allot of things that needs to be cleaned up in this file + most of the functionality does not have unit tests. I opened a separate bug for that.
So far this review is for algo's bug fix.

---
In reply to: [71405791](https://github.com/Microsoft/WinObjC/pull/669#discussion_r71405791) [](ancestors = 71405791)

Frameworks/Foundation/NSFileManager.mm
MSFTFox: In the wake of these logs, could you scrub this file for any unnecessary logging? It might also be worth looking into standardizing the error messages while you're in here since you seem to be touching on those a bit.

tests/unittests/Foundation/NSFileManagerTests.mm
```diff
@@ -136,4 +170,76 @@
...
+    NSFileManager* manager = [NSFileManager defaultManager];
+
+    NSError* error = nil;
+    BOOL status = [manager moveItemAtPath:srcPath toPath:destPath error:&error];
+    EXPECT_TRUE(status);
+    EXPECT_EQ(nil, error);
+
+    // Verify file exists.
+    EXPECT_TRUE([manager fileExistsAtPath:destPath]);
+    EXPECT_FALSE([manager fileExistsAtPath:srcPath]);
```
MSFTFox: Is there a reason these are expects and not asserts? Just curious.

Frameworks/Foundation/NSFileManager.mm
msft-Jeyaram: Yep, the error reflects reference platform, except for rename call, reference platform sets type to POSIX, we are setting it to WIN32. although the code is the errno that is being set.

Overall file clean up/error standardization  will be part of #662

---
In reply to: [71406908](https://github.com/Microsoft/WinObjC/pull/669#discussion_r71406908) [](ancestors = 71406908)

tests/unittests/Foundation/NSFileManagerTests.mm
```diff
@@ -136,4 +170,76 @@
...
+
+    NSString* srcPath = getPathToFile(srcName);
+    NSString* destPath = getPathToFile(destName);
+
+    NSFileManager* manager = [NSFileManager defaultManager];
+
+    NSError* error = nil;
+    BOOL status = [manager moveItemAtPath:srcPath toPath:destPath error:&error];
+    EXPECT_TRUE(status);
+    EXPECT_EQ(nil, error);
```
MSFTFox: This Expect nil, error really doesn't do a whole lot. It might be worthwhile to add a test for actual errors instead of a test that just makes sure nil comes back as nil through a function.

msft-Jeyaram: this is to make sure that although the file was moved, we don't have any error flags set.
Will add negative tests.

---
In reply to: [71407552](https://github.com/Microsoft/WinObjC/pull/669#discussion_r71407552) [](ancestors = 71407552)

msft-Jeyaram:
somehow I lost that in my squash and ended up adding it manually.

---
In reply to: [233735940](https://github.com/Microsoft/WinObjC/pull/669#issuecomment-233735940) [](ancestors = 233735940)

tests/unittests/Foundation/NSFileManagerTests.mm
```diff
@@ -136,4 +170,76 @@
...
+
+    NSData* fileData = [fh readDataToEndOfFile];
+    EXPECT_TRUE_MSG(fileData != nil, "FAILED: reading the entire file failed.");
+
+    NSString* str = [[[NSString alloc] initWithData:fileData encoding:NSUTF8StringEncoding] autorelease];
+    EXPECT_OBJCEQ_MSG(str, content, "FAILED: Unable to read the file content.");
+
+    [fh closeFile];
+
+    deleteFile((status ? destName : srcName));
```
MSFTFox: The deletes here worry me a bit. In a really bad situation, we could have files lying around. It might be best to move any deletes to a separate test so they get sweeped regularly instead of at the end of their own tests. Something like a test at the end of this file that checks if these files exist and deletes them if they do would be great.

DHowett-MSFT: Prefer a WIL scope defer, actually. RAII is king!

---
In reply to: [71407993](https://github.com/Microsoft/WinObjC/pull/669#discussion_r71407993) [](ancestors = 71407993)

tests/unittests/Foundation/NSFileManagerTests.mm
```diff
@@ -136,4 +170,76 @@
...
+    EXPECT_FALSE([manager fileExistsAtPath:[srcURL path]);
+
+    // Verify data.
+    NSFileHandle* fh = [NSFileHandle fileHandleForReadingAtPath:getPathToFile(destName)];
+    EXPECT_TRUE(fh != nil);
+
+    NSData* fileData = [fh readDataToEndOfFile];
+    EXPECT_TRUE_MSG(fileData != nil, "FAILED: reading the entire file failed.");
+
+    NSString* str = [[[NSString alloc] initWithData:fileData encoding:NSUTF8StringEncoding] autorelease];
```
MSFTFox: Might it be worthwhile to add a test for data other than UTF8?

DHowett-MSFT: This is not testing whether we can read UTF-8, and additional encodings are out-of-scope here.
It'd probably be better to avoid using NSString entirely for this and just compare two NSData instances.

---
In reply to: [71408245](https://github.com/Microsoft/WinObjC/pull/669#discussion_r71408245) [](ancestors = 71408245)

tests/unittests/Foundation/NSFileManagerTests.mm
DHowett-MSFT: `dest` should be an ASSERT, and `src` should be an EXPECT. The destination not existing will be fatal, but the source not existing will not impact the rest of the test.

---
In reply to: [71407123](https://github.com/Microsoft/WinObjC/pull/669#discussion_r71407123) [](ancestors = 71407123)

tests/unittests/Foundation/NSFileManagerTests.mm
```diff
@@ -136,4 +170,76 @@
...
+
+    // Verify data.
+    NSFileHandle* fh = [NSFileHandle fileHandleForReadingAtPath:destPath];
+    EXPECT_TRUE(fh != nil);
+
+    NSData* fileData = [fh readDataToEndOfFile];
+    EXPECT_TRUE_MSG(fileData != nil, "FAILED: reading the entire file failed.");
+
+    NSString* str = [[[NSString alloc] initWithData:fileData encoding:NSUTF8StringEncoding] autorelease];
+    EXPECT_OBJCEQ_MSG(str, content, "FAILED: Unable to read the file content.");
```
DHowett-MSFT: It seems unusual to use NSFileHandle here; NSData is equally capable of reading a file from disk in a single line, and makes for a more succinct test surface.

tests/unittests/Foundation/NSFileManagerTests.mm
msft-Jeyaram: wrapped up with RAII 

---
In reply to: [71407993](https://github.com/Microsoft/WinObjC/pull/669#discussion_r71407993) [](ancestors = 71407993)

tests/unittests/Foundation/NSFileManagerTests.mm
msft-Jeyaram: indeed. switched to NSData, will makes it easier to verify data.

---
In reply to: [71408764](https://github.com/Microsoft/WinObjC/pull/669#discussion_r71408764) [](ancestors = 71408764)

tests/unittests/Foundation/NSFileManagerTests.mm
msft-Jeyaram: ASSERT: Fails fast, aborting the current function.
    EXPECT: Continues after the failure.

Updating, also using RAII, that should make life easier :D

---
In reply to: [71408625](https://github.com/Microsoft/WinObjC/pull/669#discussion_r71408625) [](ancestors = 71408625,71407123)

Frameworks/Foundation/NSFileManager.mm
msft-Jeyaram: I was letting it fall down to the fact if it's not fileURL then we throw exception. looks like reference platform returns an error with the exception.

---
In reply to: [71392720](https://github.com/Microsoft/WinObjC/pull/669#discussion_r71392720) [](ancestors = 71392720)

Frameworks/Foundation/NSFileManager.mm
```diff
@@ -1,8 +1,8 @@
 //******************************************************************************
 //
-// Copyright (c) 2015 Microsoft Corporation. All rights reserved.
+// COPYRIGHT (C) MICROSOFT. ALL RIGHTS RESERVED.
 //
-// This code is licensed under the MIT License (MIT).
+// THIS CODE IS LICENSED UNDER THE MIT LICENSE (MIT).
```
msft-Jeyaram: oh god everything got caped.

tests/unittests/Foundation/NSFileManagerTests.mm
```diff
@@ -136,4 +161,55 @@
...
+    EXPECT_EQ(nil, error);
+
+    // Verify file exists.
+    ASSERT_TRUE([manager fileExistsAtPath:[destURL path]]);
+    EXPECT_FALSE([manager fileExistsAtPath:[srcURL path]]);
+
+    // Verify data.
+
+    ASSERT_OBJCEQ([content dataUsingEncoding:NSUTF8StringEncoding], [NSData dataWithContentsOfURL:destURL]);
+}
```
DHowett-MSFT: It might be even better to sue `ASSERT_OBJCEQ(content, [NSString stringWithContentsOfFile:...])` to avoid encoding shenanigans.

msft-Jeyaram: either way we end up encoding, but it's a single pass. changed .

---
In reply to: [71443437](https://github.com/Microsoft/WinObjC/pull/669#discussion_r71443437) [](ancestors = 71443437)

rajsesh-msft:
:shipit:

mnithish:
CI build successful

PR #671, opened 2016-07-19
Fall back on CoreFoundation for some core NSString/NSCFString behaviours

DHowett-MSFT:
This reimplements some of `_NSCFString`'s internals in terms of non-backstore-checking CF internal implementations exposed in ForFoundationOnly.h.
As a result, we now throw range/length exceptions and properly catch string mutability without exploding.

It also replaces our home-grown strings format parser with CoreFoundation's tried/true one.

Additionally, printing an object via NSString's `%@` format specifier will prefer `-description` instead of `CFShow()` for Objective-C types.


DHowett-MSFT:

DHowett-MSFT:

tests/unittests/Foundation/NSStringTests.m
```diff
@@ -576,3 +576,41 @@ void testUrlCharacterSetEncoding(NSString* decodedString, NSString* encodedStrin
...
+
+    ASSERT_OBJCEQ(@"value1", propertyList[@"key1"]);
+    ASSERT_OBJCEQ(@"value2", propertyList[@"key2"]);
+}
+
+INSTANTIATE_TEST_CASE_P(NSString,
+                        StringsFormatPropertyList,
+                        ::testing::Values(L"\uFEFFkey1=value1;\n\"key2\"=\"value2\";", // BOM
+                                          L"key1=value1;\n\"key2\"=\"value2\";" // No BOM
+                                          ));
```
bbowman: This is so cool!!!!!

Frameworks/Foundation/NSCFString.mm
```diff
@@ -89,12 +93,12 @@ - (instancetype)initWithFormat:(NSString*)format locale:(id)locale arguments:(va
     CFStringRef str;
 
     if (locale == nil) {
-        str = CFStringCreateWithFormatAndArguments(nullptr, nullptr, static_cast<CFStringRef>(format), argList);
+        str = _CFStringCreateWithFormatAndArgumentsAux(nullptr, &_NSCFStringCopyDescription, nullptr, static_cast<CFStringRef>(format), argList);
```
bbowman: >_CFStringCreateWithFormatAndArgumentsAux [](start = 14, length = 40)

do we need this in a .def?

bbowman:
:shipit:

tests/unittests/Foundation/NSStringTests.m
DHowett-MSFT: Glad you think so! I used some of the other NSString tests as inspiration.

---
In reply to: [71446495](https://github.com/Microsoft/WinObjC/pull/671#discussion_r71446495) [](ancestors = 71446495)

Frameworks/Foundation/NSCFString.mm
DHowett-MSFT: Foundation seems capable of linking this alright; I'll check on ARM.

---
In reply to: [71446639](https://github.com/Microsoft/WinObjC/pull/671#discussion_r71446639) [](ancestors = 71446639)

tests/unittests/Foundation/NSStringTests.m
```diff
@@ -576,3 +576,41 @@ void testUrlCharacterSetEncoding(NSString* decodedString, NSString* encodedStrin
     string = @"/tmp/random.foo.tiff";
     ASSERT_TRUE([string.pathExtension isEqualToString:@"tiff"]);
 }
+
+TEST(NSString, Exceptions) {
+    NSRange range{ 100, 10 };
+
+    EXPECT_ANY_THROW([@"hello" characterAtIndex:10]);
```
rajsesh-msft: >EXPECT_ANY_THROW [](start = 4, length = 16)

Do these tests work on ARM?

Frameworks/CoreFoundationAdditions/_NSCFTemporaryRootObject.mm
```diff
@@ -16,4 +16,8 @@
 #include "_NSCFTemporaryRootObject.h"
 
 @implementation _NSCFTemporaryRootObject
+- (void)doesNotRecognizeSelector:(SEL)selector {
+    // According to the reference platform documentation, this method must never return.
+    abort();
```
rajsesh-msft: >abort [](start = 4, length = 5)

FAIL_FAST(), so we get the telemetry.

tests/unittests/Foundation/NSStringTests.m
DHowett-MSFT: I've been caught! Probably not. I'll throw them into `ARM_DISABLED_TESTS`.

---
In reply to: [71574294](https://github.com/Microsoft/WinObjC/pull/671#discussion_r71574294) [](ancestors = 71574294)

Frameworks/CoreFoundationAdditions/_NSCFTemporaryRootObject.mm
DHowett-MSFT: We can't take a dependency on error reporting here, as it requires Foundation to be loaded.

---
In reply to: [71574472](https://github.com/Microsoft/WinObjC/pull/671#discussion_r71574472) [](ancestors = 71574472)

rajsesh-msft: abort doesn't translate to watson dump, so we wont know the cases where this is hit.  Perhaps dereference NULL to 'gracefully' crash if there is such a thing.

---
In reply to: [71574896](https://github.com/Microsoft/WinObjC/pull/671#discussion_r71574896) [](ancestors = 71574896,71574472)

DHowett-MSFT: We could throw a `std::runtime_error` with a known `what`, since we have access to the CRT. It would be just as catchable and generate a trace, I think. What do you think?

That said, if this is being called, something far far more egregious has failed. It'll only happen when Foundation doesn't load and stomp the superclass for the types derived herefrom, or something in CF uses Objective-C in a bad way before that happens.

---
In reply to: [71575314](https://github.com/Microsoft/WinObjC/pull/671#discussion_r71575314) [](ancestors = 71575314,71574896,71574472)

rajsesh-msft: std::runtime_error would be a resonable alternative.

---
In reply to: [71576020](https://github.com/Microsoft/WinObjC/pull/671#discussion_r71576020) [](ancestors = 71576020,71575314,71574896,71574472)

Frameworks/Foundation/NSCFString.mm
DHowett-MSFT: Apparently not.

---
In reply to: [71573534](https://github.com/Microsoft/WinObjC/pull/671#discussion_r71573534) [](ancestors = 71573534,71446639)

tests/unittests/Foundation/NSStringTests.m
DHowett-MSFT: These tests actually do work on ARM, puzzlingly.

---
In reply to: [71574787](https://github.com/Microsoft/WinObjC/pull/671#discussion_r71574787) [](ancestors = 71574787,71574294)

Frameworks/CoreFoundationAdditions/_NSCFTemporaryRootObject.mm
DHowett-MSFT: I decided to use `RaiseFailFastException`. Verified that it generates a dump via WER when Local Dumps are enabled. The exception record is intact, too!

---
In reply to: [71582230](https://github.com/Microsoft/WinObjC/pull/671#discussion_r71582230) [](ancestors = 71582230,71576020,71575314,71574896,71574472)

tests/unittests/Foundation/NSStringTests.m
rajsesh-msft: Thanks for checking.  A wise man once said "Its always good to ask".

---
In reply to: [71600230](https://github.com/Microsoft/WinObjC/pull/671#discussion_r71600230) [](ancestors = 71600230,71574787,71574294)

Frameworks/CoreFoundationAdditions/_NSCFTemporaryRootObject.mm
rajsesh-msft: code after RaiseFailFastException is unreachable, so you should simply remove the abort.

---
In reply to: [71600334](https://github.com/Microsoft/WinObjC/pull/671#discussion_r71600334) [](ancestors = 71600334,71582230,71576020,71575314,71574896,71574472)

rajsesh-msft:
:shipit:

DHowett-MSFT:
In CI Build!

DHowett-MSFT:
Failed an application launch test. `-[NSString isAbsolutePath]` is blindly pulling characters without bounds checking, and will now throw an exception when it does so.

Addressing in an additional commit.

Frameworks/Foundation/NSString.mm
```diff
@@ -1093,7 +1097,15 @@ - (NSRange)rangeOfString:(NSString*)searchString options:(NSStringCompareOptions
  @Status Interoperable
 */
 - (BOOL)isAbsolutePath {
-    return ([self hasPrefix:_NSGetSlashStr()]) || ((_isLetter([self characterAtIndex:0])) && ([self characterAtIndex:1] == ':'));
+    if ([self hasPrefix:_NSGetSlashStr()]) {
+        return YES;
+    }
+
+    if (self.length >= 2) {
+        return _isLetter([self characterAtIndex:0]) && [self characterAtIndex:1] == ':';
```
DHowett-MSFT: This function is technically wrong, because `C:Hello.txt` is a relative path, not an absolute one.

Fixing it is out-of-scope for this change.

Frameworks/Foundation/NSString.mm
```diff
@@ -537,6 +537,10 @@ - (BOOL)getBytes:(void*)buffer
  @Status Interoperable
 */
 - (void)getCharacters:(unichar*)dest range:(NSRange)range {
+    if (range.location + range.length > self.length) {
```
ms-jihua: surprised we didn't have this check already. use NSLocationInRange instead, though?

DHowett-MSFT: We don't exactly want that here; if length is 5, `NSLocationInRange(5, {1,2})` will be false but that doesn't make this invalid.

Frameworks/Foundation/NSString.mm
ms-jihua: Talked to DHowett offline, he's going to take a closer look at drive paths next iteration, and leave this slightly-wrong state for now.

Be sure to add a comment about this status.

DHowett-MSFT:
Submitting a new CI build with the updates, just to make sure.

DHowett-MSFT:
Passed CI build and application launch tests.

PR #684, opened 2016-07-25
Force the Objective-C 2-stage loader to be pinned when pulled in from a lib and add /WHOLEARCHIVE support

DHowett-MSFT:
Additionally, add support for `/WHOLEARCHIVE`, easing the consumption of Objective-C code in libraries. Objective-c classes and categories consumed via static libraries with this option turned on will no longer be stripped from the final binary.

For external projects, it is opt-in to sidestep potential build issues due to missing dependencies.
The SDK and unit tests will build with it enabled.


bbowman:
>     <ClangCompile Include="..\..\..\Frameworks\CoreFoundation\Base.subproj\CFRuntime.c" />

was this already here and just being stripped? I'm just a little confused since its not showing as an addition in iteration 9.

---
Refers to: build/CoreFoundation/lib/CoreFoundationLib.vcxproj:126 in 8ab9f86. [](commit_id = 8ab9f8655a3ec2e39bd534d9f214531055c64e65, deletion_comment = False)

DHowett-MSFT:
>     <ClangCompile Include="..\..\..\Frameworks\CoreFoundation\Base.subproj\CFRuntime.c" />

It looks like it, yes; with /WHOLEARCHIVE we were getting duplicate symbol errors. It wasn't added in this CR.

---
In reply to: [235375643](https://github.com/Microsoft/WinObjC/pull/684#issuecomment-235375643) [](ancestors = 235375643)

---
Refers to: build/CoreFoundation/lib/CoreFoundationLib.vcxproj:126 in 8ab9f86. [](commit_id = 8ab9f8655a3ec2e39bd534d9f214531055c64e65, deletion_comment = False)

build/CoreFoundation/lib/CoreFoundationLib.vcxproj
```diff
@@ -153,7 +153,6 @@
     <ClangCompile Include="..\..\..\Frameworks\CoreFoundation\Parsing.subproj\CFOldStylePList.c" />
     <ClangCompile Include="..\..\..\Frameworks\CoreFoundation\Parsing.subproj\CFPropertyList.c" />
     <ClangCompile Include="..\..\..\Frameworks\CoreFoundation\Parsing.subproj\CFXMLInputStream.c" />
-    <ClangCompile Include="..\..\..\Frameworks\CoreFoundation\Parsing.subproj\CFXMLInterface.c" />
```
DHowett-MSFT: >XMLInterf [](start = 80, length = 9)

@MSFTFox is working on libxml-related fixes that address this.

Frameworks/CoreFoundation/Base.subproj/CFSystemDirectories.c
```diff
@@ -55,7 +55,8 @@ CFSearchPathEnumerationState __CFGetNextSearchPathEnumeration(CFSearchPathEnumer
 #endif
 
 
-#if DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_EMBEDDED || DEPLOYMENT_TARGET_WINDOWS
+// WINOBJC: Disable this function only for WinObjC as we don't provide system intrinsics for it.
```
bbowman: > system intrinsics [](start = 70, length = 18)

I'm not seeing the system intrinsics here. what are we missing to allow this? Who are the callers of this? do we need to adjust a .def?

DHowett-MSFT: This was never in the .def, so it was being stripped out anyway.
The native functions I'm referring to are above, on 38 and 42. _"We use the System framework implementation on Mach."_

---
In reply to: [72318824](https://github.com/Microsoft/WinObjC/pull/684#discussion_r72318824) [](ancestors = 72318824)

DHowett-MSFT: That is, `NSGetNextSearchPathEnumeration`

---
In reply to: [72319135](https://github.com/Microsoft/WinObjC/pull/684#discussion_r72319135) [](ancestors = 72319135,72318824)

msvc/sdk-build.props
```diff
@@ -29,6 +29,7 @@
     <ApplicationTypeRevision Condition="'$(ApplicationTypeRevision)'==''">10.0</ApplicationTypeRevision>
     <UseStarboardSourceSdk Condition="'$(UseStarboardSourceSdk)'==''">true</UseStarboardSourceSdk>
     <IslandwoodDRT Condition="'$(IslandwoodDRT)'==''">false</IslandwoodDRT>
+    <StarboardLinkWholeArchive Condition="'$(StarboardLinkWholeArchive)'==''">true</StarboardLinkWholeArchive>
```
bbowman: >StarboardLinkWholeArchive [](start = 5, length = 25)

Is this something we would expect our 3rd party consumers to want by default? I imagine any of them that want categories will be sad if this is off (and they make a lib + dll I guess).... just wondering if we want this to be opt out instead of in.

bbowman:
:shipit:

msvc/sdk-build.props
DHowett-MSFT: We should definitely evaluate this, and possibly have vsimporter detect it and turn it on.

For now, it's too large a hammer to drop on 3rd-party consumers.

rajsesh-msft:
Do we have comparative binary sizes after this change?  Granted, having code in the binary is more important than having smaller binaries with missing code, but would be good to know the disk impact.

msvc/objcrt/objc_2stage_init.cpp
```diff
@@ -92,6 +92,8 @@ extern "C" const InitFunc __declspec(selectany) _SEGALLOC(".CRT$XCLB") ___objc_m
 // we would be resolving classes that have not been loaded.
 extern "C" const InitFunc __declspec(selectany) _SEGALLOC(".CRT$XCV") ___objc_legacy_post_XCU = &___objc_legacy_resolve_all;
 
+extern "C" __declspec(selectany) const void* ___pin_objc_init = 0x0;
```
rajsesh-msft: >___pin_objc_init [](start = 45, length = 16)

Could we please add a comment on this contract with the compiler?

rajsesh-msft:
:shipit:

PooyaZv:
:shipit:

PooyaZv:
@annagrin is added to the review. #Closed

PooyaZv:

DHowett-MSFT:
Passed CI Build!
Addressing feedback by adding comments to `objc_2stage_init.cpp`, and running a few builds to determine the growth in binary size.

DHowett-MSFT:
```
                      | Debug   | Release
----------------------+---------+--------
   With /WHOLEARCHIVE | 66.0MiB | 44.5MiB
Without /WHOLEARCHIVE | 65.0MiB | 43.7MiB
```

DHowett-MSFT:
This figure is for every DLL produced by our build system taken in conglomerate.

PR #685, opened 2016-07-26
UIImage bug fixes

shahmeeresmail:
drawInRect: In some cases, images may be drawn upside down due to the source image being sampled incorrectly.

drawAtPoint: Unlike drawInRect, drawAtPoint must make use of the scaling factor so revert scaling factor change for this function only.


davelamb:
I just compiled the code against our local develop branch and found the scaling issue I mentioned before still exists. You'll notice there's a discrepancy between the way UISegment is rendered and the way a plain UIImageView with resizable image renders. The latter uses a Xaml NineGrid and scales things correctly. The former doesn't take the scale into account for the source rect and so renders things too large when used with @2x images, such as our segments.

My suggestion would be to add another set of locals to drawPatches, srcTopCap, srcBotCap, etc, which is equal to img->_imageInsets.<cap> * img->_scale, and use those in the source rects of the DrawImageRect calls. 

You'll also have to change the conditionals in drawPatches to read something like if ((pos.size.width - leftCap - rightCap > 0) && (srcWidth - srcLeftCap - srcRightCap > 0)) { .. }

Finally, you'll have to change UISegment +initialize. All the resizable images should have insets of { 5, 5, 5, 5 } since with my suggested code changes some of the patches don't get rendered due to the conditional, and we'd also like to stretch the caps height-wise.

Thanks!

shahmeeresmail:
There are several issues here:
1. This code is to fix UIImage::drawInRect and UIImage::drawAtPoint. I have verified that after my fix, these functions operate exactly as they are supposed to on iOS. I can share with you the code I used to verify this offline if you like.

2. Some functions, specifically UISegment::drawRect, rely upon UIImage::drawInRect and UIImage::drawAtPoint to draw the border and background of their component. The problem is that UIImage::drawInRect is not supposed to take the UIImage scale factor into account but UISegment expects it to since as you mentioned it is using a 2x texture. DrawInRect simply takes the entire source image and translates and stretches it to fit entirely within the destination rect while taking account insets which are not intended to be stretched. DrawAtPoint on the other hand, will draw the UIElement and apply whatever scale factor has been specified for the component when defining the output rect.

3. Even with the original implementation and ignoring everything above, UISegment is not rendering correctly. Right now, WinObjC uses selected and unselected textures to draw a gray border and white background when unselected and a blue background with a gray border when selected. On iOS, an app can a specify tintColor which defines the border color when unselected and the background color when selected. Since WinObjC selects between 2 static textures to draw the border and fill the background, it can't take into account the colors specified by the developer. Further, when insets are specified, up to 9 textured draws are being done for each segment which makes it quite inefficient. The correct way would be to do a single vector draw for the border and a single colorFill for the background. This way we could avoid 9 textured draws and at the same time more closely match what the developer specified.

davelamb:
Sure, please send the test code over, since I also tested locally and found drawAtPoint and drawInRect both care about the scale of the UIImage being drawn.

In terms of the third point, I don't disagree, but we're talking orders of magnitude of difference in the amount of work required, I was just hoping for stop-gap solution for the meantime. 

---
In reply to: [237041236](https://github.com/Microsoft/WinObjC/pull/685#issuecomment-237041236) [](ancestors = 237041236)

shahmeeresmail:
So I did some more testing and found that insets and their contents are not drawn correctly in BOTH the original code and in my code. Further, in the original code, if the source image has scaling applied, the output quality is significantly reduced.

For example in a override of UIView::drawRect, this code draws incorrectly with the original code:

    //Image dim: 2500x1800
    UIImage *uiImage = [UIImage imageNamed : @"photo.jpg"];
    CGImageRef cgImage = uiImage.CGImage;

    UIImage* scaledImage = [[UIImage imageWithCGImage : cgImage scale : 3.0f orientation : UIImageOrientationUp] resizableImageWithCapInsets:UIEdgeInsetsMake(150, 150, 150, 150) resizingMode : UIImageResizingModeStretch];
    //Rect Dim: 320x480
    CGRect imageRect = [[UIScreen mainScreen] bounds];
    [scaledImage drawInRect : imageRect];

iOS: All corner caps very large, thin vertical and horizontal strips containing squashed center of majority of image.
Original WinObjC: Insets not applied at all, and the image is of poor quality. Looks like a lower resolution version of the image was stretched over the target. This is probably due to the fact that it is rendered to a low resolution image one third the size of the original image(ie the pattern) then stretched again over the destination.
WinObjC with fix1: Inset size is correct but contents do not match iOS
WinObjC with fix2: Matches iOS

shahmeeresmail:
On a similar note, since we should probably do the right thing, I have developed a fix for the issues highlighted in part 3 of my original comment. Reviewing the change internally right now.

jaredhms:
In CI build.

PR #691, opened 2016-07-27
Fix NSOperation - ready check in less time

### LINKED ISSUE #624: NSOperation ready state check should be O(1) not O(N)
_checkReady can be changed to O(1) instead of O(N) by keeping the ready count.  since you are observing dependencies and their 'finished' state, incrementing/decrementing the counter should do the trick.

jofre-ms:
This pull request fixes issue #624 by tracking operation dependencies in a set and counting the number that are ready.

### DIFF
--- a/Frameworks/Foundation/NSOperation.mm
+++ b/Frameworks/Foundation/NSOperation.mm
@@ -41,13 +41,12 @@ OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 
 @implementation NSOperation {
     StrongId<NSMutableSet> _dependencies;
-    NSUInteger _readyCount;
+    NSUInteger _outstandingDependenciesCount;
     StrongId<void(^)()> _completionBlock;
     std::condition_variable_any _finishCondition;
     std::recursive_mutex _finishLock;
     std::recursive_mutex _dependenciesLock;
     std::recursive_mutex _completionBlockLock;
-    std::recursive_mutex _readyLock;
 }
 
 @synthesize cancelled = _cancelled;
@@ -66,13 +65,12 @@ static const NSString* NSOperationContext = @"context";
 }
 
 - (void)_checkReady {
-    std::lock_guard<std::recursive_mutex> lock(_readyLock);
+    std::lock_guard<std::recursive_mutex> lock(_dependenciesLock);
     bool newReady = YES;
 
     // If cancelled, skip this logic and set _ready to YES. Otherwise check dependency ready count.
     if (![self isCancelled]) {
-        std::lock_guard<std::recursive_mutex> lock(_dependenciesLock);
-        if (_readyCount < [_dependencies count]) {
+        if (_outstandingDependenciesCount > 0) {
             newReady = NO;
         }
     }
@@ -91,10 +89,10 @@ static const NSString* NSOperationContext = @"context";
             _finishCondition.notify_all();
             _finishLock.unlock();
         } else {
-            std::lock_guard<std::recursive_mutex> lock(_readyLock);
-            if ([object isFinished]) {
-                if ([_dependencies containsObject:object]) {
-                    _readyCount++;
+            std::lock_guard<std::recursive_mutex> lock(_dependenciesLock);
+            if ([_dependencies containsObject:object]) {
+                if ([object isFinished]) {
+                    _outstandingDependenciesCount--;
                 }
             }
 
@@ -117,7 +115,7 @@ static const NSString* NSOperationContext = @"context";
  @Status Interoperable
 */
 - (BOOL)isReady {
-    std::lock_guard<std::recursive_mutex> lock(_readyLock);
+    std::lock_guard<std::recursive_mutex> lock(_dependenciesLock);
     return _ready;
 }
 
@@ -130,6 +128,7 @@ static const NSString* NSOperationContext = @"context";
     [_dependencies addObject:operation];
     [operation addObserver:self forKeyPath:@"isFinished" options:NSKeyValueObservingOptionInitial context:(void*)NSOperationContext];
     [self didChangeValueForKey:@"dependencies"];
+    _outstandingDependenciesCount++; // will be decremented if necessary in observeValueForKey, due to NSKeyValueObservingOptionInitial
     [self _checkReady];
 }
 
@@ -137,12 +136,11 @@ static const NSString* NSOperationContext = @"context";
  @Status Interoperable
 */
 - (void)removeDependency:(NSOperation*)operation {
-    std::lock_guard<std::recursive_mutex> readyLock(_readyLock);
     std::lock_guard<std::recursive_mutex> dependenciesLock(_dependenciesLock);
     if ([_dependencies containsObject:operation]) {
         [self willChangeValueForKey:@"dependencies"];
-        if ([operation isFinished]) {
-            _readyCount--;
+        if (![operation isFinished]) {
+            _outstandingDependenciesCount--;
         }
 
         [_dependencies removeObject:operation];
--- a/Frameworks/Foundation/NSOperation.mm
+++ b/Frameworks/Foundation/NSOperation.mm
@@ -44,9 +44,11 @@ OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
     NSUInteger _outstandingDependenciesCount;
     StrongId<void(^)()> _completionBlock;
     std::condition_variable_any _finishCondition;
-    std::recursive_mutex _finishLock;
-    std::recursive_mutex _dependenciesLock;
-    std::recursive_mutex _completionBlockLock;
+
+    // The locks. _finishLock should be taken before _completionBlockLock if both need to be taken. No other locks should overlap.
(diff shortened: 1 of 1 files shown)


Frameworks/Foundation/NSOperation.mm
```diff
@@ -92,14 +91,21 @@ - (void)observeValueForKeyPath:(NSString*)keyPath ofObject:(id)object change:(NS
             _finishCondition.notify_all();
             _finishLock.unlock();
         } else {
+            std::lock_guard<std::recursive_mutex> lock(_readyLock);
+            if ([object isFinished]) {
+                if ([_dependencies containsObject:object]) {
```
rajsesh-msft: >ependencies containsObject:object] [](start = 23, length = 34)

why is this check even necessary?

Frameworks/Foundation/NSOperation.mm
```diff
@@ -119,10 +125,10 @@ - (BOOL)isReady {
  @Status Interoperable
 */
 - (void)addDependency:(id)operation {
-    std::lock_guard<std::recursive_mutex> lock(_dependenciesLock);
+    std::lock_guard<std::recursive_mutex> dependenciesLock(_dependenciesLock);
     [self willChangeValueForKey:@"dependencies"];
     [_dependencies addObject:operation];
```
rajsesh-msft: >_dependencies [](start = 5, length = 13)

do you need to do 

if ([operation isFinished]) {
            _readyCount++;
        }

rajsesh-msft:
:clock1:

DHowett:
Nit: I'd prefer an `outstandingDependenciesCount` instead of a `readyCount`: It saves us having to think about the actual depth of the dep queue when we're figuring out if there are outstanding deps.

Separately, you can possibly replace isReady with `{return _outstandingDependencies==0;}` instead of storing a separate ready flag. Less state to drift out of sync!

Frameworks/Foundation/NSOperation.mm
jofre-ms: This check is necessary to prevent a race condition between removeDependency and observeValueForKeyPath resulting from that dependency finishing. In removeDependency, we check the dependency's isFinished property and decrement _readyCount if that dependency is finished. If for some reason it is not finished when we check it, but it completes and calls observeValueForKeyPath before removeObserver is called, then observeValueForKeypath will block until _readyLock can be held, and at that point the dependency will no longer be contained in the _dependencies set, ensuring that _readyCount is not modified.

---
In reply to: [72368995](https://github.com/Microsoft/WinObjC/pull/691#discussion_r72368995) [](ancestors = 72368995)

Frameworks/Foundation/NSOperation.mm
jofre-ms: NSKeyValueObservingOptionInitial will cause observeValueForKeyPath to be invoked on the dependency, which contains the logic you mentioned.

---
In reply to: [72369304](https://github.com/Microsoft/WinObjC/pull/691#discussion_r72369304) [](ancestors = 72369304)

jofre-ms:
I like the first change because we can save some calls to [_dependencies count]. I don't like the second change because it conflicts with the design: isReady needs to return YES when _cancelled is YES - to work around this we'd need more helpers and logic to avoid storing a flag to figure out when isReady has changed, and storing a flag is not expensive.

---
In reply to: [235612017](https://github.com/Microsoft/WinObjC/pull/691#issuecomment-235612017) [](ancestors = 235612017)

jofre-ms:
>     std::recursive_mutex _readyLock;

I think this can be combined with _dependenciesLock

---
Refers to: Frameworks/Foundation/NSOperation.mm:50 in 666fe36. [](commit_id = 666fe36fdcd6547c8b44cb295c94f6788cbddbb3, deletion_comment = False)

Frameworks/Foundation/NSOperation.mm
```diff
@@ -111,32 +115,36 @@ - (id)init {
...
-    [operation addObserver:self forKeyPath:@"isFinished" options:0 context:(void*)NSOperationContext];
+    [operation addObserver:self forKeyPath:@"isFinished" options:NSKeyValueObservingOptionInitial context:(void*)NSOperationContext];
     [self didChangeValueForKey:@"dependencies"];
+    _outstandingDependenciesCount++; // will be decremented if necessary in observeValueForKey, due to NSKeyValueObservingOptionInitial
     [self _checkReady];
```
DHowett-MSFT: Possibly reorder these, so that the increment always stages before the decrement.
Otherwise, you might run into an underflow situation when addObserver drops it below 0.

Frameworks/Foundation/NSOperation.mm
```diff
@@ -304,16 +312,15 @@ - (void)waitUntilFinished {
...
 }
 
 /**
  @Status Interoperable
 */
 - (void)dealloc {
     _dependenciesLock.lock();
-    for(NSOperation* op in (NSArray*)_dependencies) {
+    for(NSOperation* op in [_dependencies allObjects]) {
         [op removeObserver:self forKeyPath:@"isFinished" context:(void*)NSOperationContext];
```
DHowett-MSFT: You can still NSFastEnumerate a set!

Frameworks/Foundation/NSOperation.mm
jofre-ms: Both operations happen under _dependenciesLock. Should not be an issue.

---
In reply to: [72506986](https://github.com/Microsoft/WinObjC/pull/691#discussion_r72506986) [](ancestors = 72506986)

Frameworks/Foundation/NSOperation.mm
```diff
@@ -304,16 +312,15 @@ - (void)waitUntilFinished {
...
-    NSArray* copy = [[_dependencies copy] autorelease];
-    return copy;
+    return [_dependencies allObjects];
 }
 
 /**
  @Status Interoperable
 */
 - (void)dealloc {
     _dependenciesLock.lock();
```
bbowman: >   _dependenciesLock.lock() [](start = 1, length = 27)

can we be consistent with the style of locking? I prefer lock_guard as its RAII so there is less chance of accidentally leaving things locked.

Frameworks/Foundation/NSOperation.mm
```diff
@@ -111,32 +115,36 @@ - (id)init {
...
 */
 - (void)addDependency:(id)operation {
-    std::lock_guard<std::recursive_mutex> lock(_dependenciesLock);
+    std::lock_guard<std::recursive_mutex> dependenciesLock(_dependenciesLock);
     [self willChangeValueForKey:@"dependencies"];
     [_dependencies addObject:operation];
-    [operation addObserver:self forKeyPath:@"isFinished" options:0 context:(void*)NSOperationContext];
+    [operation addObserver:self forKeyPath:@"isFinished" options:NSKeyValueObservingOptionInitial context:(void*)NSOperationContext];
     [self didChangeValueForKey:@"dependencies"];
+    _outstandingDependenciesCount++; // will be decremented if necessary in observeValueForKey, due to NSKeyValueObservingOptionInitial
```
bbowman: >_outstandingDependenciesCount++; // will be decremented if necessary in observeValueForKey, due to NSKeyValueObservingOptionInitial [](start = 4, length = 131)

seems wrong. What if dependency is already finished? Shouldn't you only increment for unfinished ones? Does KVO send an immediate observation for the current state or only when it changes? Additionally this logic seems wrong if you add the same dependency twice. You will send a change notification for _dependencies when no change occured and you'll be double counting the same object which is unclear how the _outstandingDepCount will work out....

Frameworks/Foundation/NSOperation.mm
```diff
@@ -40,13 +40,13 @@
...
 
 @implementation NSOperation {
-    StrongId<NSMutableArray> _dependencies;
+    StrongId<NSMutableSet> _dependencies;
+    NSUInteger _outstandingDependenciesCount;
     StrongId<void(^)()> _completionBlock;
     std::condition_variable_any _finishCondition;
     std::recursive_mutex _finishLock;
     std::recursive_mutex _dependenciesLock;
     std::recursive_mutex _completionBlockLock;
```
bbowman: this is a terrifying number of locks for a single object. Can you add a comment here explaining what order locks need to be taken in?

bbowman: and also what data each is protecting as it looks like _finishLock protects some non obvious members like _cancel 

---
In reply to: [72513398](https://github.com/Microsoft/WinObjC/pull/691#discussion_r72513398) [](ancestors = 72513398)

Frameworks/Foundation/NSOperation.mm
jofre-ms: NSKeyValueObservingOptionInitial sends an immediate observation for the current state, before addObserver returns (and in the same callstack). Therefore, if the dependency is already finished, we will increment and decrement _outstandingDependenciesCount before addDependency returns, while holding the appropriate locks. According to documentation, "The same dependency should not be added more than once to the receiver, and the results of doing so are undefined."

---
In reply to: [72513161](https://github.com/Microsoft/WinObjC/pull/691#discussion_r72513161) [](ancestors = 72513161)

DHowett-MSFT:
:shipit:

bbowman:
:shipit:

PR #707, opened 2016-07-29
Fix #696 by removing the short circuit from -[NSData copyWithZone:]

### LINKED ISSUE #696: -[NSCFData copy] cannot legally take the easy way out and return [self retain];
There is no way for `NSCFData` to determine whether it was created with an unowned-free or unowned-nofree buffer; therefore, copies of these immutable data will be left with a double free or pointing to garbage memory, respectively.

DHowett-MSFT:
#696 (link for posterity)


DHowett-MSFT:
This is not the appropriate fix at the moment.

PR #718, opened 2016-07-29
Update gtest-api.h to allow OSX ASSERT_OBJCEQ macros to print correctly

### LINKED ISSUE #705: OSX ASSERT_OBJCEQ output prints object pointer instead of description
Dependent on https://github.com/Microsoft/WinObjC/pull/634

Eg:

[ RUN      ] NSBundle.Localizations
Foundation/ReferenceFoundation/TestNSBundle.mm:89: Failure
Value of: bundle.localizations
  Actual: 0x7ff2e1c04ec0
Expected: @[ @"en" ]
Which is: 0x7ff2e1c4b7d0
[  FAILED  ] NSBundle.Localizations (0 ms)

This makes diagnosing issues quite a bit more difficult - we should see why our ostream specialization for id isn't getting hit.

ms-jihua:
Update gtest-api.h to allow ASSERT_OBJCEQ macros to print descriptions instead of pointer addresses on OSX

Fixes #705

### DIFF
--- a/tests/frameworks/gtest/include/gtest-api.h
+++ b/tests/frameworks/gtest/include/gtest-api.h
@@ -97,12 +97,8 @@ std::vector<char> Format(NSString* message, Args ... args)
 }
 }
 
-namespace testing {
-namespace internal {
 inline std::ostream& operator<<(std::ostream& os, const id& object) {
     return os << (char*)([[object description] UTF8String]);
-}
-}
 }
 
 namespace woc {


tests/frameworks/gtest/include/gtest-api.h
```diff
@@ -97,13 +97,9 @@ std::vector<char> Format(NSString* message, Args ... args)
 }
 }
 
-namespace testing {
-namespace internal {
```
bbowman: can we instead change the macro definitions to use the namespace? this change adds a global operator overload in a .h which includers may not want ...

bbowman:
:clock1:

tests/frameworks/gtest/include/gtest-api.h
bbowman: also I'm pretty confused why this didn't work ...

---
In reply to: [72853566](https://github.com/Microsoft/WinObjC/pull/718#discussion_r72853566) [](ancestors = 72853566)

ms-jihua: discussed offline - we decided we were okay with this.

---
In reply to: [72853566](https://github.com/Microsoft/WinObjC/pull/718#discussion_r72853566) [](ancestors = 72853566)

DHowett-MSFT:
:shipit:

DHowett-MSFT:
Talked with Brian offline; he appeared comfortable with this change. Passed CI build!

PR #726, opened 2016-07-29
Fix some build issues with the OS X Unit Tests

DHowett-MSFT:
* Set a minimum OS X version (10.10, for now).
* Fix C++ imports for WOCStdLib consumers.
* Add a make entrypoint (`make` without any arguments will choose the topmost rule).
* Fix dependency tracking that caused UT outputs to be rebuilt every time.


tests/unittests/Makefile
```diff
@@ -4,7 +4,14 @@
 
 include $(CLEAR_VARS)
 
+TESTS_TO_BUILD := Foundation
+
+# Makefile Entrypoint: for running `make` without an argument.
+.PHONY: all
+all: $(foreach t,$(TESTS_TO_BUILD),$(t).UnitTests.exe)
+
```
ms-jihua: Thanks for taking care of this for me, it would've been the next thing to hit after looking at the current failures.

tests/unittests/Makefile
```diff
@@ -56,11 +64,11 @@ $(OUTDIR)/ReferenceFoundation_%.o : $(OUTDIR)/ReferenceFoundation_TestUtils.o $(
 	mkdir -p $(OUTDIR)
 	$(CLANG) -c -o $@ $(CFLAGS) $(word 2,$^) 
 
-.PHONY: clean Shims
```
ms-jihua: There's no .PHONY for shims anymore?

DHowett-MSFT: nah, since it's not a rule. having it being a phony rule made make think that `Foundation.UnitTests.exe` was out of date all the time.

---
In reply to: [72871776](https://github.com/Microsoft/WinObjC/pull/726#discussion_r72871776) [](ancestors = 72871776)

tests/unittests/Makefile
DHowett-MSFT: Gladly! :)

---
In reply to: [72871659](https://github.com/Microsoft/WinObjC/pull/726#discussion_r72871659) [](ancestors = 72871659)

ms-jihua:
:shipit:

rajsesh-msft:
:shipit:

DHowett-MSFT:
CI builds passed (though they do not run OS X tests, so they do not provide full coverage here.)

PR #727, opened 2016-07-29
Fix #696 by only deep-copying un-owned or mutable instances in -[NSData copyWithZone:]

### LINKED ISSUE #696: -[NSCFData copy] cannot legally take the easy way out and return [self retain];

DHowett-MSFT:
This also adds tests, which were verified on OS X.


bbowman:
:shipit:

rajsesh-msft:
:shipit:

DHowett-MSFT:
Passed CI build; Passed OS X UTs.

Issue #734, opened 2016-08-01
NSOperation completionBlock not executed for NSOperation subclass.


rupertdaniel:
If you implement your own concurrent NSOperation subclass, the completion block for the operation is not executed when the `isFinished` property is changed to `TRUE`.

It looks like the completion block handling is done in the private method `- (void)_finish:(BOOL)didExecute` and if your subclass overrides the `-(void)start` method, then the `_finish` method will never be called and the completion block will not be executed.

DHowett-MSFT:
@rupertdaniel Thanks for the report. We've root-caused the issue and come up with a tentative fix. Do you have a representative sample we can include in our test burden to make sure this doesn't regress?

DHowett-MSFT:
It appears as though we have a test for this case -- it asserts that the operation is finished when the completion block is called.
However, _it never actually asserts that the completion block was, in fact, called._

rupertdaniel:
Great, thanks for looking at this.

Here is a simple operation subclass...

```
@interface MyOperation : NSOperation

@property (assign, getter = isExecuting) BOOL executing;
@property (assign, getter = isFinished, readonly) BOOL finished;

@end

@implementation MyOperation

@synthesize executing = _executing;
@synthesize finished = _finished;

- (void)setExecuting:(BOOL)executing
{
	[self willChangeValueForKey:@"isExecuting"];
	[self willChangeValueForKey:@"isFinished"];
	
	_executing = executing;
	_finished = !executing;
	
	[self didChangeValueForKey:@"isExecuting"];
	[self didChangeValueForKey:@"isFinished"];
}

- (BOOL)isExecuting
{
	return _executing;
}

- (BOOL)isFinished
{
	return _finished;
}

- (void)start
{	
	if (self.isCancelled)
		return;
	
	self.executing = YES;
	[self doSomething];
}

- (void)doSomething
{
	// Do some async task.
	dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_LOW, 0), ^{
            self.executing = NO;
    });
}

@end
```

Then to run it...
```
self.op = [[MyOperation alloc] init];
self.op.completionBlock = ^{
  NSLog(@"FINISHED!!!");
};

[self.queue addOperation:self.op];
```

DHowett-MSFT:
Thanks.

We investigated a couple more use cases for NSOperation, and identified a couple other problem areas.
* Operations should be scheduleable without a queue and run manually.
[...]

PR #784, opened 2016-08-01
Fire NSOperation's completion block on a background thread after the isFinished notification.

DHowett-MSFT:
This is a fix for #734.

From the documentation on the reference platform,
> The exact execution context for your completion block is not guaranteed but is typically a secondary thread. Therefore, you should not use this block to do any work that requires a very specific execution context. Instead, you should shunt that work to your application’s main thread or to the specific thread that is capable of doing it. For example, if you have a custom thread for coordinating the completion of the operation, you could use the completion block to ping that thread. 

We could take an optimization and call the completion block on the operation queue's thread at the cost of further complicating the code, or always shunt it off to a background thread as we are doing in this pull request.

Every test that had a completion block with an assertion was updated to **actually make sure that completion block was called.**

**Passed OS X UTs** apart from known failures in KVO observation.

### DIFF
--- a/Frameworks/Foundation/NSOperation.mm
+++ b/Frameworks/Foundation/NSOperation.mm
@@ -39,6 +39,10 @@ OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 #import <condition_variable>
 #import <mutex>
 
+#import "NSOperationInternal.h"
+
+static wchar_t TAG[] = L"NSOperation";
+
 @implementation NSOperation {
     StrongId<NSMutableSet> _dependencies;
     NSUInteger _outstandingDependenciesCount;
@@ -55,6 +59,7 @@ OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 @synthesize executing = _executing;
 @synthesize finished = _finished;
 @synthesize ready = _ready;
+@synthesize _completionQueue = _completionQueue;
 
 static const NSString* NSOperationContext = @"context";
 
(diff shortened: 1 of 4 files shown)


Frameworks/Foundation/NSOperation.mm
```diff
@@ -84,11 +84,32 @@ - (void)_checkReady {
     }
 }
 
+- (void)_completionBlockThread {
+    decltype(_completionBlock) localCompletion;
+
+    { // _completionBlockLock scope
+        std::lock_guard<std::recursive_mutex> lock(_completionBlockLock);
+        localCompletion = std::move(_completionBlock);
```
DHowett-MSFT: `std::move` is well-defined for `AutoId`: the source is nil'd.

That said: a moved-from object is said to be in an undefined state. If we give it a defined state, we are attempting to trump the standard.

Thoughts?

rajsesh-msft: _completionBlock in this case is undefined, which doesn't mean that it cannot go to a reliable state, just that code should not assume any specific state.  What it means in this specific case is that after the std::move(), you should set _completionBlock to nil even if you know that our implementation will set it to nil.  

---
In reply to: [73055468](https://github.com/Microsoft/WinObjC/pull/784#discussion_r73055468) [](ancestors = 73055468)

Frameworks/Foundation/NSOperation.mm
```diff
@@ -84,11 +84,32 @@ - (void)_checkReady {
...
+    } // end _completionBlockLock scope
+
+    if (localCompletion) {
+        localCompletion();
+    }
+}
+
+- (void)_spawnCompletionBlockThread {
+    // Occurs under _finishLock.
+    [NSThread detachNewThreadSelector:@selector(_completionBlockThread) toTarget:self withObject:nil];
```
rajsesh-msft: >detachNewThreadSelector [](start = 14, length = 23)

We will spin off a new thread for every operation that completes, which is not good.  At the very least, we should use a thread pool, ideally we should further complicate the code and call it in the NSOperationQueue thread. :)

rajsesh-msft:
Discussed offline with Dustin on this, we will spend a little bit longer to make sure we don't have many more lingering bugs in this codebase.

rajsesh-msft:
:clock1:

DHowett-MSFT:
This pull request now has an update to fix dependency processing.

@rajsesh-msft I've removed your :shipit: shipit guy :)

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -163,7 +163,9 @@ static BOOL RunOperationFromLists(NSAtomicListRef* listPtr, NSAtomicListRef* sou
             *curOperation = nil;
         } else {
             NSAtomicListInsert(sourceListPtr, op);
-            return FALSE;
+            // We claim that we have done some work here.
+            // Otherwise, this work session is considered to be over and we clear the pending queue.
+            return TRUE;
```
DHowett-MSFT: This is the only meaningful change here. The rest is tidying.

DHowett-MSFT:

Frameworks/Foundation/NSOperation.mm
```diff
@@ -55,6 +59,7 @@ @implementation NSOperation {
 @synthesize executing = _executing;
 @synthesize finished = _finished;
 @synthesize ready = _ready;
+@synthesize _completionQueue = _completionQueue;
```
aballway: >_completionQueue = _completionQueue [](start = 12, length = 35)

Would this default to __completionQueue?

aballway:
:shipit:

Frameworks/Foundation/NSOperation.mm
DHowett-MSFT: I believe.

Frameworks/Foundation/NSOperation.mm
```diff
@@ -84,11 +89,38 @@ - (void)_checkReady {
     }
 }
 
+- (void)_dispatchCompletionBlock {
+    dispatch_queue_t completionQueue = _completionQueue;
+    if (!completionQueue) {
+        // It is legal to run an NSOperation without an operation queue.
```
mnithish: >It is legal to run an NSOperation without an operation queue. [](start = 11, length = 61)

If iOS does not allow running NSOperation without an operation queue, do we have not be this nice?

mnithish:
:shipit:

Frameworks/Foundation/NSOperation.mm
DHowett-MSFT: This behaviour was confirmed on the reference platform!

PR #786, opened 2016-08-01
AddressBook Framework -- AddressBook and Authorization

### LINKED ISSUE #675: Addressbook - addressbook and authorization
Implement the following APIs:

- [ ] ABAddressBookCreate
- [ ] ABAddressBookCreateWithOptions
- [ ] ABAddressBookGetAuthorizationStatus
- [ ] ABAddressBookRequestAccessWithCompletion

chadimoussi:
Fixes #675

Adds support for creating an AddressBook and checking
the Authorization Status (permissions) of using AddressBook.


Frameworks/AddressBook/ABAddressBook.mm
```diff
@@ -16,44 +16,87 @@
...
 ABAddressBookRef ABAddressBookCreateWithOptions(CFDictionaryRef options, CFErrorRef* error) {
-    UNIMPLEMENTED();
-    return StubReturn();
+    _ABAddressBook* addressBook = [[_ABAddressBook alloc] init];
+    if (addressBook.contactStore == nil) {
+        // There was an error setting up the backing
+        // contact store, so just return NULL to signify
+        // the error.
+        if (error != NULL) {
+			// In this case, the caller is responsible to release error.
```
MSFTFox: >		 [](start = 0, length = 2)

Please use spaces, not tabs.

Frameworks/AddressBook/_ABAddressBook.mm
```diff
@@ -0,0 +1,52 @@
...
+//******************************************************************************
+
+#import "_ABAddressBook.h"
+#import "UWP/WindowsApplicationModelContacts.h"
+
+@interface _ABAddressBook ()
+
+@end
+
+@implementation _ABAddressBook
```
MSFTFox: >_ABAddressBook [](start = 16, length = 14)

I'm not sure you need an extra class to handle the fetch for the contact store. It might be better to break it out into a static helper function and store the contact store as a private ivar in address book.

Frameworks/AddressBook/ABAddressBook.mm
```diff
@@ -16,44 +16,87 @@
...
 */
 ABAddressBookRef ABAddressBookCreateWithOptions(CFDictionaryRef options, CFErrorRef* error) {
-    UNIMPLEMENTED();
-    return StubReturn();
+    _ABAddressBook* addressBook = [[_ABAddressBook alloc] init];
+    if (addressBook.contactStore == nil) {
+        // There was an error setting up the backing
+        // contact store, so just return NULL to signify
+        // the error.
+        if (error != NULL) {
```
MSFTFox: >NULL [](start = 21, length = 4)

nullptr

Frameworks/AddressBook/ABAddressBook.mm
```diff
@@ -16,44 +16,87 @@
...
 }
 
 /**
- @Status Stub
- @Notes
+ @Status Caveat
+ @Notes Requesting access should be done by declaring Contacts capabilities in the app's package manifest -- it cannot be granted at
+ runtime.
 */
 void ABAddressBookRequestAccessWithCompletion(ABAddressBookRef addressBook, ABAddressBookRequestAccessCompletionHandler completion) {
```
bbowman: >ABAddressBookRequestAccessWithCompletion [](start = 5, length = 40)

Is there any way to deep link to the privacy page in settings for this?

Frameworks/AddressBook/_ABAddressBook.mm
chadimoussi: It was set-up as a class with future operations in mind (like adding/updating/removing contacts) -- an _ABAddressBook will need to maintain extra state and have more functionality to interact with the contact store and/or individual contacts as future features are implemented, particularly the save operation.

---
In reply to: [73073364](https://github.com/Microsoft/WinObjC/pull/786#discussion_r73073364) [](ancestors = 73073364)

MSFTFox: That makes sense then. I would prefer a more descriptive name then. Something like _ContactListManager. Having a name virtually identical to another class can create confusion going forward.

---
In reply to: [73080445](https://github.com/Microsoft/WinObjC/pull/786#discussion_r73080445) [](ancestors = 73080445,73073364)

Frameworks/AddressBook/_ABAddressBook.mm
```diff
@@ -0,0 +1,52 @@
...
+        [WACContactManager requestStoreAsyncWithAccessType:WACContactStoreAccessTypeAllContactsReadOnly
+            success:^(WACContactStore* success) {
+                completed = YES;
+                result = success;
+            }
+            failure:^(NSError* failure) {
+                completed = YES;
+            }];
+
+        // Wait until async method completes.
```
bviglietta: >// Wait until async method completes. [](start = 8, length = 37)

This isn't quite thread-safe. Just because completed has been set to YES doesn't mean that result has been set.

Just swapping the order in which completed and result are set won't necessarily fix it. You should probably wait on some kind of synchronization object here.

Frameworks/AddressBook/ABAddressBook.mm
```diff
@@ -16,44 +16,87 @@
...
+#import <StubReturn.h>
+
+#import "_ABAddressBook.h"
+
+#import "UWP/InteropBase.h"
+#import "UWP/WindowsDevicesEnumeration.h"
+#import "initguid.h"
+
+// GUID to check Windows Contacts permissions.
+DEFINE_GUID(_ABAddressBookContactsGUID, 0x7D7E8402, 0x7C54, 0x4821, 0xA3, 0x4E, 0xAE, 0xEF, 0xD6, 0x2D, 0xED, 0x93);
```
bviglietta: >DEFINE_GUID [](start = 0, length = 11)

Is this GUID available from some Windows header, rather than redefining it ourselves?

tests/testapps/AddressBookSample/AddressBookSample.vsimporter/AddressBookSample-WinStore10/Package.appxmanifest
```diff
@@ -1,49 +1,29 @@
...
+        <uap:DefaultTile Wide310x150Logo="Assets\Wide310x150Logo.png">
+        </uap:DefaultTile>
         <uap:SplashScreen Image="Assets\SplashScreen.png" />
       </uap:VisualElements>
     </Application>
   </Applications>
-
   <Capabilities>
     <Capability Name="internetClient" />
+    <uap:Capability Name="contacts" />
```
bviglietta: ><uap:Capability Name="contacts" /> [](start = 4, length = 34)

Do we need a vsimporter change to insert this into app manifests if they're using the AddressBook framework?

Frameworks/AddressBook/ABAddressBook.mm
bviglietta: >In this case, the caller is responsible to release error. [](start = 6, length = 57)

Are you sure? Error out-parameters are usually autoreleased.

Frameworks/AddressBook/ABAddressBook.mm
```diff
@@ -16,44 +16,87 @@
...
+    CFErrorRef error = NULL;
+    bool accessGranted = ABAddressBookGetAuthorizationStatus() == kABAuthorizationStatusAuthorized;
+    if (!accessGranted) {
+        error = CFErrorCreate(NULL, ABAddressBookErrorDomain, kABOperationNotPermittedByUserError, NULL);
+    }
+
+    // Since we can't request access at runtime (Contacts capabilities must be declared
+    // in the app's package manifest), immediately call the completion handler with the current
+    // authorization status.
+    completion(accessGranted, error);
```
bviglietta: >completion(accessGranted, error); [](start = 4, length = 33)

Maybe for compatibility with the reference platform we should dispatch this async.

Frameworks/AddressBook/_ABAddressBook.mm
bbowman: To tack on to what Ben said, busy loops are also not my favorite mechanism for waiting. pumping the run loop here is also questionable as you don't know what thread you are being called from. What if you are on a background thread already (that doesn't have a run loop) what does it mean to be pumping the current run loop in that case?

---
In reply to: [73199542](https://github.com/Microsoft/WinObjC/pull/786#discussion_r73199542) [](ancestors = 73199542)

Frameworks/AddressBook/_ABAddressBook.mm
```diff
@@ -0,0 +1,52 @@
...
+        __block WACContactStore* result = nil;
+
+        [WACContactManager requestStoreAsyncWithAccessType:WACContactStoreAccessTypeAllContactsReadOnly
+            success:^(WACContactStore* success) {
+                completed = YES;
+                result = success;
+            }
+            failure:^(NSError* failure) {
+                completed = YES;
+            }];
```
bbowman: Ignoring the failure here seems questionable. I would think if we failed this that we should return a nil object and release self?

Frameworks/include/MockClass.h
```diff
@@ -503,11 +527,20 @@ VERIFY_ARE_EQUAL(1, callCount);
...
+#ifdef VERIFY_IS_NOT_NULL
 template <typename TFunctor>
 void ValidateCallback(const TFunctor& functor, const wchar_t* name) {
     WEX::TestExecution::SetVerifyOutput verifySettings(WEX::TestExecution::VerifyOutputSettings::LogOnlyFailures);
     VERIFY_IS_NOT_NULL(functor, WEX::Common::String().Format(L"The callback for %s has not been set!", name));
 }
+#elif defined(ASSERT_NE_MSG)
+template <typename TFunctor>
+void ValidateCallback(const TFunctor& functor, const wchar_t* name) {
+	ASSERT_NE_MSG(functor, nullptr, "The callback for %s has not been set!", name);
```
bbowman: >	 [](start = 0, length = 1)

no tabs! please make sure you are clang formatting all code before sending a pr.

Frameworks/include/MockClass.h
```diff
@@ -503,11 +527,20 @@ VERIFY_ARE_EQUAL(1, callCount);
 #if !defined CUSTOM_MOCK_CALLBACK_VALIDATOR
 namespace Test {
 namespace Mock {
+#ifdef VERIFY_IS_NOT_NULL
 template <typename TFunctor>
```
bbowman: This seems pretty hacky. What happens if this is included before the test framework headers? Is there anything better we can do here?

tests/unittests/AddressBook/AuthorizationTests.mm
```diff
@@ -0,0 +1,143 @@
...
+
+#import <AddressBook/ABAddressBook.h>
+#import "UWP/WindowsDevicesEnumeration.h"
+
+using namespace ABI::Windows::Devices::Enumeration;
+using namespace ABI::Windows::Foundation;
+using namespace Microsoft::WRL;
+
+// Status to be returned by mocked version of get_CurrentStatus.
+static WDEDeviceAccessStatus statusToMock;
```
bbowman: >statusToMock [](start = 29, length = 12)

should probably initialize this guy to something.

tests/unittests/AddressBook/AuthorizationTests.mm
```diff
@@ -0,0 +1,143 @@
...
+                  ABAddressBookGetAuthorizationStatus(),
+                  "FAILED: Authorization Status should be authorized!\n");
+
+    statusToMock = WDEDeviceAccessStatusDeniedByUser;
+    ASSERT_EQ_MSG(kABAuthorizationStatusDenied, ABAddressBookGetAuthorizationStatus(), "FAILED: Authorization Status should be denied!\n");
+
+    statusToMock = WDEDeviceAccessStatusDeniedBySystem;
+    ASSERT_EQ_MSG(kABAuthorizationStatusRestricted,
+                  ABAddressBookGetAuthorizationStatus(),
+                  "FAILED: Authorization Status should be restricted!\n");
```
bbowman: seems like a good candidate for paramterized tests

tests/unittests/AddressBook/AuthorizationTests.mm
```diff
@@ -0,0 +1,143 @@
...
+
+    [__ABAddressBook_Swizzle tearDownMock:originalImplementation];
```
bbowman: >[__ABAddressBook_Swizzle tearDownMock:originalImplementation]; [](start = 4, length = 62)

setup and teardown methods are hard to get correct. For example what happens if your test encounters an exception? what about if it AVs (seg faults)? What state are we corrupting and how it will impact future tests? In general we want to isolate the test code itself from any setup / cleanup that needs to occur. In other test frameworks like TAEF, there is a facility known as test fixtures (for example TEST_CLASS_SETUP) that handle this nicely. You should investigate if something similar exists in gtest that you can use for this purpose.

tests/unittests/AddressBook/AuthorizationTests.mm
```diff
@@ -0,0 +1,143 @@
...
+    statusToMock = WDEDeviceAccessStatusUnspecified;
+    ABAddressBookRequestAccessWithCompletion(NULL, ^(bool granted, CFErrorRef error) {
+        ASSERT_FALSE_MSG(granted, "FAILED: Access should not be granted\n");
+    });
+
+    // Test the case where permissions are granted.
+    statusToMock = WDEDeviceAccessStatusAllowed;
+    ABAddressBookRequestAccessWithCompletion(NULL, ^(bool granted, CFErrorRef error) {
+        ASSERT_TRUE_MSG(granted, "FAILED: Access should be granted\n");
+    });
```
bbowman: ditto on this this test for all comments on the other.

include/AddressBook/ABPerson.h
```diff
@@ -26,7 +26,7 @@
 
 typedef uint32_t ABPersonSortOrdering;
 typedef uint32_t ABPersonCompositeNameFormat;
-typedef enum ABPersonImageFormat;
+typedef enum { kABPersonImageFormatThumbnail = 0, kABPersonImageFormatOriginalSize = 2 } ABPersonImageFormat;
```
bbowman: formatting? seems like it should be on separate lines?

Frameworks/include/_ABAddressBook.h
```diff
@@ -0,0 +1,25 @@
...
+//
+//******************************************************************************
+
+#pragma once
+
+#import "UWP/WindowsApplicationModelContacts.h"
+
+@interface _ABAddressBook : NSObject
+
+@property (nonatomic) WACContactStore* contactStore;
```
bbowman: >nonatomic [](start = 11, length = 9)

interesting choice. Why?

Frameworks/AddressBook/ABAddressBook.mm
```diff
@@ -16,44 +16,87 @@
...
+        // There was an error setting up the backing
+        // contact store, so just return NULL to signify
+        // the error.
+        if (error != NULL) {
+			// In this case, the caller is responsible to release error.
+            *error = CFErrorCreate(NULL, ABAddressBookErrorDomain, kABOperationNotPermittedByUserError, NULL);
+        }
+        return NULL;
+    } else {
+        return (__bridge_retained ABAddressBookRef)addressBook;
```
bbowman: >__bridge_retained [](start = 16, length = 17)

why do you need a __bridge_retained here? This is not a bridged class in the traditional sense of the term.

Frameworks/AddressBook/ABAddressBook.mm
```diff
@@ -16,44 +16,87 @@
...
 /**
- @Status Stub
- @Notes
+ @Status Caveat
+ @Notes Requesting access should be done by declaring Contacts capabilities in the app's package manifest -- it cannot be granted at
+ runtime.
 */
 void ABAddressBookRequestAccessWithCompletion(ABAddressBookRef addressBook, ABAddressBookRequestAccessCompletionHandler completion) {
-    UNIMPLEMENTED();
+    CFErrorRef error = NULL;
```
bbowman: >CFErrorRef [](start = 4, length = 10)

consider using woc::unique_cf for error here. if the code in completion throws an exception (and you don't own that code) you will leak error. If you make completion async you have other interesting memory management problems. In objective C, especially with ARC disabled, we need to think very hard about memory management and the ownership of objects.

Frameworks/AddressBook/ABAddressBook.mm
```diff
@@ -16,44 +16,87 @@
...
+
+    if (currentStatus == WDEDeviceAccessStatusAllowed) {
+        return kABAuthorizationStatusAuthorized;
+    } else if (currentStatus == WDEDeviceAccessStatusDeniedBySystem) {
+        return kABAuthorizationStatusRestricted;
+    } else if (currentStatus == WDEDeviceAccessStatusDeniedByUser) {
+        return kABAuthorizationStatusDenied;
+    } else { // WDEDeviceAccessStatusUnspecified
+        return kABAuthorizationStatusNotDetermined;
+    }
```
bbowman: switch instead?

bbowman:
:clock1:

Frameworks/AddressBook/ABAddressBook.mm
chadimoussi: If the app doesn't have the capabilities added in their manifest, the privacy page won't show up in settings. We could deep-link in the case where an app had declared the contacts capability, but a user went out of their way to turn off the permissions in the settings app; however, the examples I've seen don't use this method to request permissions if a user has explicitly denied permission (they will only try to request permission if the status is not determined).


---
In reply to: [73075277](https://github.com/Microsoft/WinObjC/pull/786#discussion_r73075277) [](ancestors = 73075277)

Frameworks/AddressBook/ABAddressBook.mm
chadimoussi: I got the GUID from a device access list from someone on the contacts team -- not sure if a header exists.


---
In reply to: [73199986](https://github.com/Microsoft/WinObjC/pull/786#discussion_r73199986) [](ancestors = 73199986)

tests/testapps/AddressBookSample/AddressBookSample.vsimporter/AddressBookSample-WinStore10/Package.appxmanifest
chadimoussi: Yes, I think that is a good idea to save developers the trouble of having to manually update the manifest.

---
In reply to: [73201248](https://github.com/Microsoft/WinObjC/pull/786#discussion_r73201248) [](ancestors = 73201248)

Frameworks/AddressBook/ABAddressBook.mm
chadimoussi: I was under the impression that it followed the create rule ([https://developer.apple.com/library/ios/documentation/CoreFoundation/Conceptual/CFMemoryMgmt/Concepts/Ownership.html](https://developer.apple.com/library/ios/documentation/CoreFoundation/Conceptual/CFMemoryMgmt/Concepts/Ownership.html))

---
In reply to: [73203025](https://github.com/Microsoft/WinObjC/pull/786#discussion_r73203025) [](ancestors = 73203025)

Frameworks/AddressBook/ABAddressBook.mm
chadimoussi: To follow the create rule ([https://developer.apple.com/library/ios/documentation/CoreFoundation/Conceptual/CFMemoryMgmt/Concepts/Ownership.html](https://developer.apple.com/library/ios/documentation/CoreFoundation/Conceptual/CFMemoryMgmt/Concepts/Ownership.html)) and give the caller ownership, since we are expecting them to call CFRelease when they are done

---
In reply to: [73208539](https://github.com/Microsoft/WinObjC/pull/786#discussion_r73208539) [](ancestors = 73208539)

include/AddressBook/ABPerson.h
chadimoussi: That was what the clang formatter did

---
In reply to: [73208203](https://github.com/Microsoft/WinObjC/pull/786#discussion_r73208203) [](ancestors = 73208203)

Frameworks/AddressBook/ABAddressBook.mm
bviglietta: Create rule in this case applies to the return value, but not to out-parameters.

---
In reply to: [73217170](https://github.com/Microsoft/WinObjC/pull/786#discussion_r73217170) [](ancestors = 73217170,73203025)

Frameworks/AddressBook/ABAddressBook.mm
bbowman: link doesn't seem to talk about __bridge_retained at all ... __bridge_retained only matters when ARC is turned on (which I don't think it is here) and will add a retain call that you don't need here. [[Waffle alloc] init] returns an object with a +1 ref count. This means you either need to (auto)release it or transfer ownership. Here you are transferring ownership to the caller by nature of returning it at a +1. This means it becomes the caller's responsibility to (auto)release it when they are done with it and the Create rule is satisfied. 

---
In reply to: [73218322](https://github.com/Microsoft/WinObjC/pull/786#discussion_r73218322) [](ancestors = 73218322,73208539)

Frameworks/include/_ABAddressBookManager.h
```diff
@@ -0,0 +1,25 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#pragma once
+
+#import "UWP/WindowsApplicationModelContacts.h"
+
+@interface _ABAddressBookManager : NSObject
```
rajsesh-msft: >_ABAddressBookManager [](start = 11, length = 21)

Nit:while the class name is fine by convention, we don't have files that start with _.  These are typically something like ABAddressBookManagerInternal.h/mm.  Granted, the interface names and file names will not match.

tests/testapps/AddressBookSample/AddressBookSample.vsimporter/AddressBookSample-WinStore10/Package.appxmanifest
rajsesh-msft: @chadimoussi Can you open an issue to track this?

---
In reply to: [73217044](https://github.com/Microsoft/WinObjC/pull/786#discussion_r73217044) [](ancestors = 73217044,73201248)

Frameworks/AddressBook/_ABAddressBookManager.mm
```diff
@@ -0,0 +1,58 @@
...
+                failed = NO;
+                dispatch_semaphore_signal(semaphore);
+            }
+            failure:^(NSError* failure) {
+                failed = YES;
+                dispatch_semaphore_signal(semaphore);
+            }];
+
+        // Wait until async method completes.
+        dispatch_semaphore_wait(semaphore, DISPATCH_TIME_FOREVER);
```
rajsesh-msft: >dispatch_semaphore_wait [](start = 8, length = 23)

have you tried this on the phone?  phone has strong requirements on not blocking the UI thread...

Frameworks/AddressBook/ABAddressBook.mm
```diff
@@ -16,44 +16,89 @@
...
 /**
- @Status Stub
+ @Status Interoperable
  @Notes
 */
 ABAddressBookRef ABAddressBookCreateWithOptions(CFDictionaryRef options, CFErrorRef* error) {
-    UNIMPLEMENTED();
-    return StubReturn();
+    _ABAddressBookManager* addressBook = [[_ABAddressBookManager alloc] init];
+    if (addressBook.contactStore == nil) {
```
rajsesh-msft: >(addressBook.contactStore  [](start = 7, length = 26)

addressBook == nil, since init will return nil on failure.

rajsesh-msft:
:clock1:

Frameworks/AddressBook/_ABAddressBookManager.mm
chadimoussi: I ran the sample app on a phone and it seems to run okay (the success callback is called, and init terminates)

---
In reply to: [73253069](https://github.com/Microsoft/WinObjC/pull/786#discussion_r73253069) [](ancestors = 73253069)

Frameworks/AddressBook/ABAddressBook.mm
DHowett-MSFT: Since an objc property get on nil returns nil, this is still valid. :)

rajsesh-msft: absolutely but unnecessary <:D

---
In reply to: [73264467](https://github.com/Microsoft/WinObjC/pull/786#discussion_r73264467) [](ancestors = 73264467)

Frameworks/AddressBook/ABAddressBook.mm
chadimoussi: ARC is turned on, so __bridge_retained will allow us to return at +1

---
In reply to: [73252102](https://github.com/Microsoft/WinObjC/pull/786#discussion_r73252102) [](ancestors = 73252102,73218322,73208539)

Frameworks/AddressBook/ABAddressBook.mm
chadimoussi: I think it does in this case after talking to Brian/Dustin

---
In reply to: [73249545](https://github.com/Microsoft/WinObjC/pull/786#discussion_r73249545) [](ancestors = 73249545,73217170,73203025)

chadimoussi:
Both the Debug and Release CI Builds passed

PR #787, opened 2016-08-02
NSPredicate copy should be done through static method.

### LINKED ISSUE #704: Foundation.NSCompoundPredicate_copy fails on OSX
Dependent on https://github.com/Microsoft/WinObjC/pull/634

[ RUN      ] Foundation.NSCompoundPredicate_copy
unknown file: Failure
[  FAILED  ] Foundation.NSCompoundPredicate_copy (0 ms)

Require deeper investigation to see what the exception actually is.

msft-Jeyaram:
It seems as on reference platform NSPredicate{*} classes are not concrete instance, thus doing an alloc/init will not return a copy-able object.
Since our backing implementation is different, it gives us the freedom to do the that. 

Changing the test to be compatible with the reference platform.

fix #704

### DIFF
--- a/tests/unittests/Foundation/NSCompoundPredicateTests.m
+++ b/tests/unittests/Foundation/NSCompoundPredicateTests.m
@@ -388,7 +388,7 @@ TEST(NSCompoundPredicate, ArchiveAndUnarchiveObject) {
 }
 
 TEST(Foundation, NSCompoundPredicate_copy) {
-    NSPredicate* predicate = [[NSPredicate alloc] init];
+    NSPredicate* predicate = [NSPredicate predicateWithFormat:@"10 >= 30"];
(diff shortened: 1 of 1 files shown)


tests/unittests/Foundation/NSCompoundPredicateTests.m
```diff
@@ -388,7 +388,7 @@
 }
 
 TEST(Foundation, NSCompoundPredicate_copy) {
-    NSPredicate* predicate = [[NSPredicate alloc] init];
+    NSPredicate* predicate = [NSPredicate predicateWithFormat:@"10 >= 30"];
```
rajsesh-msft: What about the other tests in this file?  Looks like there are other failures as well, could you please fix all the related issues?

rajsesh-msft:
:clock1:

msft-Jeyaram:
recalling this back. Sending out all of CompoundPredicate changes together.

PR #789, opened 2016-08-02
making NSCompoundPredicate tests compatible with reference platform

### LINKED ISSUE #747: NSCompoundPredicate.ArchiveAndUnarchiveObject fails on OSX
[ RUN      ] NSCompoundPredicate.ArchiveAndUnarchiveObject
unknown file: Failure
[  FAILED  ] NSCompoundPredicate.ArchiveAndUnarchiveObject (1 ms)

msft-Jeyaram:
It seems as on reference platform predicate classes are not concrete instance, thus doing an alloc/init will not return a copy-able object.
Since our backing implementation is different, it gives us the freedom to do the that.

Changing the test to be compatible with the reference platform.

fix #747 #704

### DIFF
--- a/tests/unittests/Foundation/NSCompoundPredicateTests.m
+++ b/tests/unittests/Foundation/NSCompoundPredicateTests.m
@@ -18,20 +18,17 @@
 #import <Foundation/Foundation.h>
 
 TEST(NSCompoundPredicate, Init) {
-    NSCompoundPredicate* compoundPredicate = [[NSCompoundPredicate alloc] init];
+    NSCompoundPredicate* compoundPredicate = [NSCompoundPredicate notPredicateWithSubpredicate:[NSPredicate predicateWithFormat:@"1<3"]];
     ASSERT_TRUE_MSG(compoundPredicate != nil, "FAILED: compoundPredicate should be non-null!");
-    [compoundPredicate release];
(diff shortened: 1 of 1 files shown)


ms-jihua:
:shipit:

tests/unittests/Foundation/NSCompoundPredicateTests.m
```diff
@@ -18,20 +18,17 @@
 #import <Foundation/Foundation.h>
 
 TEST(NSCompoundPredicate, Init) {
-    NSCompoundPredicate* compoundPredicate = [[NSCompoundPredicate alloc] init];
+    NSCompoundPredicate* compoundPredicate = [NSCompoundPredicate notPredicateWithSubpredicate:[NSPredicate predicateWithFormat:@"1<3"]];
```
DHowett-MSFT: >"1<3 [](start = 129, length = 4)

Is `TRUE` a valid predicate? If so, it might be better to use it.

DHowett-MSFT:
:shipit:

rajsesh-msft:
:shipit:

msft-Jeyaram:
In CI Build

msft-Jeyaram:
CI Build passed.

Issue #737, opened 2016-08-01
NSArray.Description test fails on OSX


ms-jihua:
[ RUN      ] NSArray.Description
Foundation/NSArrayTests.mm:229: Failure
Value of: [testArray description]
  Actual: (
    1,
    2,
    3
)
Expected: @"(1, 2, 3)"
Which is: (1, 2, 3)
[  FAILED  ] NSArray.Description (1 ms)

Seems like a fairly simple fix.

Issue #738, opened 2016-08-01
NSArray BinarySearch tests fail on OSX


ms-jihua:
[ RUN      ] NSArray.BinarySearchInsertionIndex
Foundation/NSArrayTests.mm:278: Failure
Value of: objectIndexInArray(array, 11, 0, array.count, 0) == NSNotFound
  Actual: false
Expected: true
NSArray return NSNotFound if object is not found.
[  FAILED  ] NSArray.BinarySearchInsertionIndex (0 ms)

[ RUN      ] NSArray.BinarySearch
Foundation/ReferenceFoundation/TestNSArray.mm:154: Failure
Value of: NSNotFound
  Actual: 9223372036854775807
Expected: notFound
Which is: -1
NSArray return NSNotFound if object is not found.
[  FAILED  ] NSArray.BinarySearch (0 ms)

[ RUN      ] NSArray.BinarySearchFringeCases
Foundation/ReferenceFoundation/TestNSArray.mm:310: Failure
Value of: NSNotFound
  Actual: 9223372036854775807
Expected: notFoundInEmptyArray
Which is: -1
Empty NSArray return NSNotFound for any valid arguments.
[  FAILED  ] NSArray.BinarySearchFringeCases (0 ms)

Seems like they are all related to expecting NSNotFound and not getting it.

ms-jihua:
This, from the build warnings, seems to have something to do with it:

Foundation/NSArrayTests.mm:287:37: warning: comparison of constant 9223372036854775807 with expression of type 'int' is always false [-Wtautological-constant-out-of-range-compare]
    ASSERT_TRUE_MSG(notFoundInRange == NSNotFound, @"NSArray return NSNotFound if object is not found.");

might be a signed/unsigned thing?

PR #790, opened 2016-08-02
NSArray Fixes for Reference platform tests

aballway:
Makes NSArray's description method match that of reference platform and changes a helper method's return type to prevent reference platform test failures.

Fixes issues #737 and #738


DHowett-MSFT:
:shipit:

rajsesh-msft:
:shipit:

aballway:
Passed CI Build

PR #701, opened 2016-07-28
Bridge NSTimeZone and CFTimeZone

ms-jihua:

PR #792, opened 2016-08-02
Correct NSPropertyListSerializationTest

### LINKED ISSUE #785: NSPropertyListSerializationTests fail on OSX
Off by eight hours, same as if the test is run on Windows with the CFTimeZone changes. Uncertain whether this test is actually valid - the off by 8 seems to be expected behavior, because one of the values takes into account local time and the other does not.

[ RUN      ] NSPropertyListSerialization.PropertyListForDate
Foundation/NSPropertyListSerializationTests.mm:37: Failure
Value of: plistDict[@"Date Modified"]
  Actual: 2014-12-15 19:48:38 +0000
Expected: date
Which is: 2014-12-16 03:48:38 +0000
[  FAILED  ] NSPropertyListSerialization.PropertyListForDate (0 ms)
[----------] 1 test from NSPropertyListSerialization (1 ms total)

ms-jihua:
- Test was written incorrectly.
   [NSDateFormatter dateFromString] interprets the string in local time,
   (this error also repros on Windows with https://github.com/Microsoft/WinObjC/pull/701)
   causing the dates to be off by 8 hrs when run in PST.

   Corrected by switching to using an NSDate initializer that takes an absolute time.

Fixes #785

### DIFF
--- a/docs/Testing/OSXWinObjCTests.md
+++ b/docs/Testing/OSXWinObjCTests.md
@@ -47,9 +47,9 @@ While it's possible to checkout a separate git enlistment on an OSX machine and 
 #### Current known failures:
 
 
-    * [==========] 821 tests from 74 test cases ran. (56608 ms total)
-    * [  PASSED  ] 698 tests.
-    * [  FAILED  ] 123 tests, listed below:
+    * [==========] 833 tests from 74 test cases ran. (58373 ms total)
+    * [  PASSED  ] 707 tests.
+    * [  FAILED  ] 126 tests, listed below:
     * [  FAILED  ] Archival.NSKeyedUnarchiver_Secure
     * [  FAILED  ] KVO.DerivedKeyOnSubpath1
     * [  FAILED  ] KVO.RemoveWithoutContext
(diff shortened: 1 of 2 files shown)


tests/unittests/Foundation/NSPropertyListSerializationTests.mm
```diff
@@ -29,10 +29,7 @@
     });
     ASSERT_NE(nil, plistDict);
 
-    NSDateFormatter* formatter = [[NSDateFormatter new] autorelease];
-    ASSERT_NE(nil, formatter);
-    [formatter setDateFormat:@"yyyy-MM-dd'T'HH:mm:ss'Z'"];
-    NSDate* date = [formatter dateFromString:@"2014-12-15T19:48:38Z"];
+    NSDate* date = [NSDate dateWithTimeIntervalSinceReferenceDate:440365718]; // Mon Dec 15 2014 19:48:38 GMT
```
MSFTFox: >dateWithTimeIntervalSinceReferenceDate [](start = 27, length = 38)

nit; could use calendar [dateWithEra:1 year:2014 hour:19 minute:48 second:38] if you wanted to avoid the magic number there.

MSFTFox:
:shipit:

tests/unittests/Foundation/NSPropertyListSerializationTests.mm
ms-jihua: Eh, it's a bit of a wash - I think I prefer not adding extraneous elements by bringing in Calendar here.

---
In reply to: [73243150](https://github.com/Microsoft/WinObjC/pull/792#discussion_r73243150) [](ancestors = 73243150)

docs/Testing/OSXWinObjCTests.md
```diff
@@ -47,9 +47,9 @@ While it's possible to checkout a separate git enlistment on an OSX machine and
 #### Current known failures:
 
 
-    * [==========] 821 tests from 74 test cases ran. (56608 ms total)
-    * [  PASSED  ] 698 tests.
-    * [  FAILED  ] 123 tests, listed below:
+    * [==========] 833 tests from 74 test cases ran. (58373 ms total)
```
rajsesh-msft: >* [](start = 4, length = 1)

I don't think we should update this file for every fix.  Lets accumulate some fixes and do one update at the end of the week.

rajsesh-msft: That said, *this* PR is okay.

---
In reply to: [73249616](https://github.com/Microsoft/WinObjC/pull/792#discussion_r73249616) [](ancestors = 73249616)

ms-jihua: Agreed on both fronts.

---
In reply to: [73249644](https://github.com/Microsoft/WinObjC/pull/792#discussion_r73249644) [](ancestors = 73249644,73249616)

DHowett-MSFT:
:shipit:

DHowett-MSFT:
Passed CI build!

Issue #759, opened 2016-08-01
NSObject tests fail on OSX


ms-jihua:
[----------] 12 tests from NSObject
[ RUN      ] NSObject.InstanceMethodSignature
[       OK ] NSObject.InstanceMethodSignature (0 ms)
[ RUN      ] NSObject.DynamicRespondsToSelector
[       OK ] NSObject.DynamicRespondsToSelector (0 ms)
[ RUN      ] NSObject.DynamicClassResolution
2016-08-01 10:46:03.286 Foundation.UnitTests.exe[4412:468597] +[Foundation_NSObject_Tests_DynamicResolution dynamicInstanceMethod]: unrecognized selector sent to class 0x10d35fb30
[       OK ] NSObject.DynamicClassResolution (1 ms)
[ RUN      ] NSObject.DynamicResolution
2016-08-01 10:46:03.287 Foundation.UnitTests.exe[4412:468597] -[Foundation_NSObject_Tests_DynamicResolution dynamicClassMethod]: unrecognized selector sent to instance 0x7fae98c2f040
[       OK ] NSObject.DynamicResolution (1 ms)
[ RUN      ] NSObject.PerformSelectorAfterDelaySanity
2016-08-01 10:46:03.289 Foundation.UnitTests.exe[4412:468626] *** -[NSCondition waitUntilDate:]: condition (<NSCondition: 0x7fae9b203b30> '(null)') not locked
2016-08-01 10:46:03.289 Foundation.UnitTests.exe[4412:468626] *** Break on _NSLockError() to debug.
2016-08-01 10:46:03.290 Foundation.UnitTests.exe[4412:468626] *** -[NSCondition waitUntilDate:]: condition (<NSCondition: 0x7fae9b203b30> '(null)') not locked by current thread
2016-08-01 10:46:03.290 Foundation.UnitTests.exe[4412:468626] *** Break on _NSLockError() to debug.
2016-08-01 10:46:03.613 Foundation.UnitTests.exe[4412:468597] *** -[NSCondition dealloc]: condition (<NSCondition: 0x7fae9b203b30> '(null)') deallocated while still in use
2016-08-01 10:46:03.614 Foundation.UnitTests.exe[4412:468597] *** Break on _NSLockError() to debug.
[       OK ] NSObject.PerformSelectorAfterDelaySanity (325 ms)
[ RUN      ] NSObject.CancelPreviousPerformRequestsWithTarget
[...]

PR #795, opened 2016-08-03
Fix the mutation rules in NSObject's mutableArrayValueForKeyPath: adapter.

DHowett-MSFT:
Previously, we would mutate a field if it was accessed via
ivar (instead of via getter or valueForKey:).

On the reference platform, instance variables accessed directly
are not copied and mutated.

This pull request also splits the tests into more fine-grained units, all of which pass on both platforms.


DHowett-MSFT:
Fixes #759.

rajsesh-msft:
:shipit:

ms-jihua:
Don't forget to add 'Fixes #759' to the actual commit message.

ms-jihua:
:shipit:

DHowett-MSFT:
Passed CI build!

PR #796, opened 2016-08-03
 Disable NSMutableIndexSet shiftIndexesStartingAtIndex tests from running on OSX due to inconsistent API behavior

mnithish:


mnithish:
@DHowett is added to the review. #Closed

mnithish:

mnithish:

DHowett-MSFT:

tests/unittests/Foundation/NSIndexSetTests.mm
```diff
@@ -27,18 +27,20 @@ void verifyIndexSetEqualsArray(NSArray* expected, NSIndexSet* actual) {
     }];
 
     ASSERT_EQ([expected count], callCount);
-    ASSERT_EQ(YES, [actual isEqualToIndexSet:expectedIndexSet]);
+    ASSERT_TRUE([actual isEqualToIndexSet:expectedIndexSet]);
```
DHowett-MSFT: >T_TRUE([actual is [](start = 9, length = 17)

This could also be ASSERT_OBJCEQ, right?

tests/unittests/Foundation/NSIndexSetTests.mm
```diff
@@ -27,18 +27,20 @@ void verifyIndexSetEqualsArray(NSArray* expected, NSIndexSet* actual) {
...
 }
 
 TEST(NSIndexSet, AddIndexesInRange) {
-    NSMutableIndexSet* indexSet = [NSMutableIndexSet new];
+    NSMutableIndexSet* indexSet = [[NSMutableIndexSet new] autorelease];
+    NSMutableIndexSet* expectedIndexSet = [[NSMutableIndexSet new] autorelease];
```
DHowett-MSFT: >ableIndexSet new] autoreleas [](start = 49, length = 28)

`[[NSMutableIndexSet new] autorelease]` => `[NSMutableIndexSet indexSet]`

tests/unittests/Foundation/NSIndexSetTests.mm
```diff
@@ -47,33 +49,51 @@ void verifyIndexSetEqualsArray(NSArray* expected, NSIndexSet* actual) {
...
     LOG_INFO("Left overlap addIndexesInRange operation tests of NSIndexSet");
     [indexSet addIndexesInRange:NSMakeRange(1, 2)]; // [1-2]
-    expected = @[ @1, @2, @3, @4, @5, @7, @8, @11, @12, @13, @14, @15 ];
-    verifyIndexSetEqualsArray(expected, indexSet);
+    // The resulting set should contain ranges - [1-5], [7-8] and [11-15].
+    // verify the ranges in the set.
+    [expectedIndexSet removeAllIndexes];
+    [expectedIndexSet addIndexesInRange:NSMakeRange(1, 5)]; // [1-5]
+    [expectedIndexSet addIndexesInRange:NSMakeRange(7, 2)]; // [7-8]
+    [expectedIndexSet addIndexesInRange:NSMakeRange(11, 5)]; // [11-15]
```
ms-jihua: nit: this would be neater with a helper function that takes a list/array/va_list of NSRanges.

ms-jihua:
:shipit:

tests/unittests/Foundation/NSIndexSetTests.mm
mnithish: True, it makes sense not to reply on our implementation of equality check for UT

---
In reply to: [73382204](https://github.com/Microsoft/WinObjC/pull/796#discussion_r73382204) [](ancestors = 73382204)

tests/unittests/Foundation/NSIndexSetTests.mm
mnithish: I found moving the asserts to helper methods does not really help much when investigating test failure as the asserts always point to the helper function and not the actual point in the test that caused the failure.
Said that I do agree helper methods are useful but in this case I don't think it helps much

---
In reply to: [73384439](https://github.com/Microsoft/WinObjC/pull/796#discussion_r73384439) [](ancestors = 73384439)

tests/unittests/Foundation/NSIndexSetTests.mm
mnithish: Oh well looks like we store std::vector inside the object and that does not pass ASSERT_OBJCEQ check

---
In reply to: [73382204](https://github.com/Microsoft/WinObjC/pull/796#discussion_r73382204) [](ancestors = 73382204)

tests/unittests/Foundation/NSIndexSetTests.mm
mnithish: Will use this for few tests so we have coverage

---
In reply to: [73382312](https://github.com/Microsoft/WinObjC/pull/796#discussion_r73382312) [](ancestors = 73382312)

tests/unittests/Foundation/NSIndexSetTests.mm
DHowett-MSFT: That worries me, because isEqual: should definitely call through to isEqualToIndexSet:

mnithish: Good point. We have not overridden isEqual: method 

---
In reply to: [73635392](https://github.com/Microsoft/WinObjC/pull/796#discussion_r73635392) [](ancestors = 73635392)

DHowett-MSFT:
:shipit:

PR #802, opened 2016-08-03
Issue 770: NSThread.Priority tests fail on OSX

MSFTFox:
This issue is due to the fact that thread priorities can change. When setting a
thread priority, the system actually decides what the thread's priority will be.
Because of this, using a hard equal in the test will fail. The solution is to
check if the priority is within a certain acceptable margin. This was acceptable
on Windows before since Windows pigeon holes the thread priorities into certain
groups that can be counted on to return as a certain value.

Verified by running unit tests and passing.

### DIFF
--- a/tests/unittests/Foundation/NSThreadTests.m
+++ b/tests/unittests/Foundation/NSThreadTests.m
@@ -42,6 +42,12 @@ static const double sExpectedHighThreadPriority = .87;
 static const double sExpectedLowThreadPriority = 0;
 static const double sExpectedMidThreadPriority = .5;
 
+// A thread never has a set priority. The system can always change it, in fact it will most likely change the priority.
+// Even still, the thread priority should be close enough to the actual requested value so we need to make sure it's sufficiently close.
+// The largest difference between thread priorities in Windows should be .07 units in difference as Windows pigeon holes priorities into
+// certain values.
+static const double sMarginOfError = .07;
+
 TEST(NSThread, Priority) {
     ThreadTestController* testController = [[[ThreadTestController alloc] init] autorelease];
     testController.stop = NO;
@@ -57,15 +63,15 @@ TEST(NSThread, Priority) {
     // THREAD_PRIORITY_HIGHEST and THREAD_PRIORITY_TIME_CRITICAL. Setting priority to half way between these makes it clear
     // what the priority is when conveyed back to the user.
 
-    EXPECT_EQ([myThread threadPriority], sExpectedHighThreadPriority);
+    EXPECT_NEAR([myThread threadPriority], sExpectedHighThreadPriority, sMarginOfError);
 
     // Change priority again.
     [myThread setThreadPriority:sOriginalLowThreadPriority];
-    EXPECT_EQ([myThread threadPriority], sExpectedLowThreadPriority);
+    EXPECT_NEAR([myThread threadPriority], sExpectedLowThreadPriority, sMarginOfError);
 
     // Once more.
     [myThread setThreadPriority:sOriginalMidThreadPriority];
-    EXPECT_EQ([myThread threadPriority], sExpectedMidThreadPriority);
+    EXPECT_NEAR([myThread threadPriority], sExpectedMidThreadPriority, sMarginOfError);
 
     // Let the thread die.
     testController.stop = YES;


tests/unittests/Foundation/NSThreadTests.m
```diff
@@ -57,15 +63,15 @@ - (void)countingloop {
     // THREAD_PRIORITY_HIGHEST and THREAD_PRIORITY_TIME_CRITICAL. Setting priority to half way between these makes it clear
     // what the priority is when conveyed back to the user.
 
-    EXPECT_EQ([myThread threadPriority], sExpectedHighThreadPriority);
+    EXPECT_LT([myThread threadPriority] - sExpectedHighThreadPriority, sMarginOfError);
```
aballway: >[myThread threadPriority] - sExpectedHighThreadPriority [](start = 14, length = 55)

Should this be absolute value or are we always going to be higher than these values?

ms-jihua: +1

---
In reply to: [73436715](https://github.com/Microsoft/WinObjC/pull/802#discussion_r73436715) [](ancestors = 73436715)

ms-jihua:
:shipit:

aballway:
:shipit:

rajsesh-msft:
:shipit:

tests/unittests/Foundation/NSThreadTests.m
MSFTFox: Good catch.

tests/unittests/Foundation/NSThreadTests.m
```diff
@@ -57,15 +63,15 @@ - (void)countingloop {
...
```
DHowett-MSFT: Chatted; Prefer `EXPECT_NEAR`.

DHowett-MSFT:
:clock1:

tests/unittests/Foundation/NSThreadTests.m
```diff
@@ -42,6 +42,12 @@ - (void)countingloop {
 static const double sExpectedLowThreadPriority = 0;
 static const double sExpectedMidThreadPriority = .5;
 
+// A thread never has a set priority. The system can always change it, in fact it will most likely change the priority.
+// Even still, the thread priority should be close enough to the actual requested value so we need to make sure it's sufficiently close.
+// The largest difference between thread priorities in Windows should be .07 units in difference as Windows pigeon holes priorities into
+// certain values.
+static const double sMarginOfError = .07;
```
bbowman: This is interesting. Can the system both increase and decrease the priority? I'm wondering if there are stronger statements we can make or if this is it. Also  how exactly did you arrive at .07? given that this needs to run on OSX and Windows we want the max of the two's range of movement right?

bbowman:
:clock1:

tests/unittests/Foundation/NSThreadTests.m
MSFTFox: Yes, the system can essentially do whatever it wants with the thread priority. From the reference platform's documentation: "...but because the priority is determined by the kernel, there is no guarantee what this value actually will be." That said, I took the larger of the two values that I've seen the priority change by. In this case, it's Windows since the system priorities are bucketized to particular values that I've mapped to more expected values. If the priority changes by more than .07 then that indicates the priority of the thread was changed by more than the size of one of these buckets which means the priority was not preserved.

I've seen the OSX values change significantly less on an average basis however they change more often. Usually by a couple decimals in either direction. Seems like the reference platform is very hands on with thread priorities.

---
In reply to: [73447706](https://github.com/Microsoft/WinObjC/pull/802#discussion_r73447706) [](ancestors = 73447706)

tests/unittests/Foundation/NSThreadTests.m
```diff
@@ -42,6 +42,12 @@ - (void)countingloop {
...
+
 TEST(NSThread, Priority) {
     ThreadTestController* testController = [[[ThreadTestController alloc] init] autorelease];
```
MSFTFox: CI Builds succeed on builds Release: Build 20160803.6  and Debug: Build 20160803.7. No new tests have been added but old tests have been fixed.

rajsesh-msft:
#Fix 770

PR #768, opened 2016-08-03
NSString tests fail on OSX (integer formatting-related)

bbowman:
[ RUN      ] NSString.NSStringTests
Foundation/NSStringTests.m:101: Failure
Value of: actualString
  Actual: ffffffffffffffff 2147483648
Expected: @"ffffffff -2147483648"
Which is: ffffffff -2147483648
[  FAILED  ] NSString.NSStringTests (0 ms)

[ RUN      ] NSString.IntegerValue
Foundation/ReferenceFoundation/TestNSString.mm:88: Failure
Value of: 2147483647
Expected: [string9 integerValue]
Which is: 9223372036854775807
[  FAILED  ] NSString.IntegerValue (1 ms)
[ RUN      ] NSString.IntValue
Foundation/ReferenceFoundation/TestNSString.mm:120: Failure
Value of: 9223372036854775807L
  Actual: 9223372036854775807
Expected: [string9 intValue]
Which is: 2147483647
[  FAILED  ] NSString.IntValue (1 ms)

### DIFF
--- a/Frameworks/Foundation/NSString.mm
+++ b/Frameworks/Foundation/NSString.mm
@@ -779,7 +779,7 @@ BASE_CLASS_REQUIRED_IMPLS(NSString, NSStringPrototype, CFStringGetTypeID);
 /**
  @Status Interoperable
 */
-- (int)intValue {
+- (int32_t)intValue {
     char* str = (char*)[self UTF8String];
     return strtol(str, nullptr, 10);
 }
@@ -788,7 +788,8 @@ BASE_CLASS_REQUIRED_IMPLS(NSString, NSStringPrototype, CFStringGetTypeID);
  @Status Interoperable
 */
 - (NSInteger)integerValue {
-    return [self intValue];
+    char* str = (char*)[self UTF8String];
+    return strtol(str, nullptr, 10);
 }
 
 /**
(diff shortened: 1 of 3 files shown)


bbowman:
libobjc2 should not have changed. Investigating.

DHowett-MSFT:
:shipit:

Frameworks/Foundation/NSString.mm
```diff
@@ -779,7 +779,7 @@ - (NSComparisonResult)compare:(NSString*)compStrAddr options:(NSStringCompareOpt
 /**
  @Status Interoperable
 */
-- (int)intValue {
+- (int32_t)intValue {
```
ms-jihua: >int32_t [](start = 3, length = 7)

! this is scary, does this mean it was interpreting 'int' as non-32bit?

Frameworks/Foundation/NSString.mm
```diff
@@ -788,7 +788,8 @@ - (int)intValue {
  @Status Interoperable
 */
 - (NSInteger)integerValue {
-    return [self intValue];
```
ms-jihua: >return [self intValue]; [](start = 4, length = 23)

NSInteger isn't def'd to int??!

PR #804, opened 2016-08-03
Issue 786: NSString tests fail on OSX (integer formatting-related)

bbowman:
Description:
Fix issue with NSString intValue needing to always return 32bit values.

How Verified: passes on both WinObjC and reference platform.

Reviewed-by: DHowett-MSFT, ms-jihua


DHowett-MSFT:
:shipit:

msft-Jeyaram:
:shipit:

ms-jihua:
:shipit:

rajsesh-msft:
:shipit:

PR #820, opened 2016-08-04
Fix NSRangeFromString to match reference platform

aballway:
Our NSRangeFromString was much more strict than the reference platform's.  This changes NSRangeFromString to accept the first two integers it finds in the string.

### DIFF
--- a/Frameworks/Foundation/NSRange.mm
+++ b/Frameworks/Foundation/NSRange.mm
@@ -49,15 +49,12 @@ NSString* NSStringFromRange(NSRange range) {
  @Status Interoperable
 */
 NSRange NSRangeFromString(NSString* s) {
-    static NSCharacterSet* numbers = [NSCharacterSet characterSetWithCharactersInString:@"1234567890"];
-    NSScanner* scanner = [NSScanner scannerWithString:s];
     unsigned long long position = 0;
     unsigned long long length = 0;
 
-    [scanner scanUpToCharactersFromSet:numbers intoString:nullptr];
-    if ([scanner scanUnsignedLongLong:&position]) {
-        [scanner scanUpToCharactersFromSet:numbers intoString:nullptr];
-        [scanner scanUnsignedLongLong:&length];
+    const char* input = strpbrk([s UTF8String], "0123456789");
+    if (input) {
+        sscanf(input, "%llu%*[^0-9]%llu", &position, &length);
     }
 
     return NSMakeRange(position, length);
--- a/Frameworks/Foundation/NSRange.mm
+++ b/Frameworks/Foundation/NSRange.mm
@@ -49,33 +49,18 @@ NSString* NSStringFromRange(NSRange range) {
  @Status Interoperable
 */
 NSRange NSRangeFromString(NSString* s) {
-    static NSCharacterSet* trimChars = [NSCharacterSet characterSetWithCharactersInString:@"{}- "];
+    static NSCharacterSet* numbers = [NSCharacterSet characterSetWithCharactersInString:@"1234567890"];
+    NSScanner* scanner = [NSScanner scannerWithString:s];
+    unsigned long long position = 0;
+    unsigned long long length = 0;
 
-    // Valid string begins with '{' and ends with '}'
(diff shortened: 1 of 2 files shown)


DHowett-MSFT:
Thanks.
In general, I think using NSScanner instead of sscanf here was not the right choice. Alas!
Things were done the way they were done.

Frameworks/Foundation/NSRange.mm
```diff
@@ -49,33 +49,18 @@ static BOOL NSLocationInRange(NSUInteger idx, NSRange r) {
...
-            return NSMakeRange(0, 0);
-        }
-
-        return NSMakeRange(position, length);
+    static NSCharacterSet* numbers = [NSCharacterSet characterSetWithCharactersInString:@"1234567890"];
+    NSScanner* scanner = [NSScanner scannerWithString:s];
+    unsigned long long position = 0;
+    unsigned long long length = 0;
+
+    [scanner scanUpToCharactersFromSet:numbers intoString:nullptr];
```
rajsesh-msft: >scanUpToCharactersFromSet [](start = 13, length = 25)

is " 1, 2" without {} valid range?

tests/unittests/Foundation/NSRangeTests.mm
```diff
@@ -19,19 +19,19 @@
...
+    testStringFromRange(@"1, 2}", { 1, 2 });
+    testStringFromRange(@"{1, 2", { 1, 2 });
+    testStringFromRange(@"{a, 2}", { 2, 0 });
+    testStringFromRange(@"{1,, 2}", { 1, 2 });
+    testStringFromRange(@"{1.3, 2}", { 1, 3 });
     testStringFromRange(@"{2147483648, 2}", { 2147483648, 2 });
-    testStringFromRange(@"\r\n{1, 2}", { 0, 0 });
+    testStringFromRange(@"\r\n{1, 2}", { 1, 2 });
     testStringFromRange(@"{  2, 4  }", { 2, 4 });
     testStringFromRange(@"{-1, -2}", { 1, 2 });
```
ms-jihua: Can we add an extra test case with three numbers?

ms-jihua:
:shipit:

rajsesh-msft:
:clock1:

Frameworks/Foundation/NSRange.mm
aballway: Yes, it will just try and grab the first two numbers it finds and make a range out of them.

---
In reply to: [73620628](https://github.com/Microsoft/WinObjC/pull/820#discussion_r73620628) [](ancestors = 73620628)

Frameworks/Foundation/NSRange.mm
```diff
@@ -49,33 +49,15 @@ static BOOL NSLocationInRange(NSUInteger idx, NSRange r) {
...
-        unsigned long long length;
-        scanner = [NSScanner scannerWithString:[components[1] stringByTrimmingCharactersInSet:trimChars]];
-        if (![scanner scanUnsignedLongLong:&length]) {
-            return NSMakeRange(0, 0);
-        }
-
-        return NSMakeRange(position, length);
+    const char* input = strpbrk([s UTF8String], "0123456789");
+    if (input) {
+        sscanf(input, "%llu%*[^0-9]%llu", &position, &length);
```
rajsesh-msft: >sscanf [](start = 8, length = 6)

if you must use sscanf, use sscanf_s instead.  Also, you will let string such as 1, 2 (without braces) pass.  Is that the expected behavior?

rajsesh-msft: could we add a test that does validate the case?

---
In reply to: [73719314](https://github.com/Microsoft/WinObjC/pull/820#discussion_r73719314) [](ancestors = 73719314)

rajsesh-msft: spoke too soon about the tests :).  Please change to use sscanf_s() instead.

---
In reply to: [73719408](https://github.com/Microsoft/WinObjC/pull/820#discussion_r73719408) [](ancestors = 73719408,73719314)

msft-Jeyaram:
:shipit:

DHowett-MSFT:
:shipit:

Issue #821, opened 2016-08-05
UIKit: Properly draw UISegment, UIDatePicker and other UI elements


shahmeeresmail:
These and possibly other UI elements are not drawn correctly on WinObjC

On iOS, an app can a specify tintColor which defines the border and text color when unselected and the background color when selected. 

On WinObjC, static textures are used to draw the border and background for selected and unselected states using textured draw calls. Since there are only two textures (one for each state) and they are fixed, the colors are static as well:
- If the element is unselected, the element will always have a gray border and white background
- If the element is selected, the element will have a gray border and blue background. 

Further, when insets are specified, it is drawn using up to 9 textured draws for each segment which makes it quite inefficient. 

It could be drawn with a single vector draw for the border and a single colorFill for the background. This way we could avoid 9 textured draws (27 if drawing a segment with 3 parts) and at the same time more closely match what the developer specified and the output on iOS.

rajsesh-msft:
This has to be reconciled with the xaml backed solution for most UIKit controls.

shahmeeresmail:
Absolutely, ideally we should map as many UIKit items to xaml type controls. However, some are not mapped including UIDatePicker and UISegment which have custom draw calls to display their components. In those cases, we should do our best to match what the app developer is asking for.

jaredhms:
Fixed with https://github.com/Microsoft/WinObjC/pull/822.

PR #822, opened 2016-08-05
UIKit drawing fixes and performance optimizations

shahmeeresmail:
Optimize performance and fix drawing of UI elements like UISegment and UIDatepicker
- Performance & Correctness: Use a single vector draw and a single color fill to draw border and background instead of 9 seperate textured draw calls. Also only draw border when element unselected. When unselected, increase size of fillRect to cover space normally filled with the border.
- Fixes: Give effect to app specified selected/unselected text colors and tint colors. Previously, 2 static textures were used and so border was always black, unselected background was always white and selected background was always blue.
- WOCCATALOG: UIColor::colorWithRed not implemented correctly so replace with explicit creation of greenColor as intended (original colorWithRed call creates a green like color on iOS)

This resolves issue #821


Frameworks/UIKit/UISegment.mm
```diff
@@ -192,6 +185,38 @@ - (id)drawRect:(CGRect)inRect {
 
     bounds = [self bounds];
 
+    id bgColor, textColor;
+
+    if (isOSTarget(@"7.0") && _selected == 0 && _tintColor != nil) {
+        textColor = _tintColor;
+        bgColor = [UIColor whiteColor];
+    } else {
```
jaredhms: can you add more comments as to why the behavior differs?

shahmeeresmail: I left it there since I was unsure why it was there to start with.

I had built and run woccatalog targetting ios7 and ios9 and it seemed to look visually identical but I didn't feel comfortable removing it just in case there was a non-obvious reason that ios7 was singled out.

jaredhms:
:shipit:

jaredhms:
Thanks for your contribution!

PR #823, opened 2016-08-05
Issue 745: NSCharacterSet tests fail on OSX

MSFTFox:
NSCharacterSet tests are failing in a few places. First and foremost, there's
an issue with NSMutableCharacterSet where many of the static initializers
are actually meant to return an immutable version that the user calls mutablecopy
on themselves. Our solution by default was to return a mutable copy. This is
incorrect.

There were a couple tests that were just invalid cases, specifically
characterSetWithBitmapRepresentation and characterSetWithContentsOfFile which
are not valid cases when passed nil.

Finally, a test was failing due to an incorrect encoding being used for
unicode characters. NSUnicodeStringEncoding appears to default to UTF16
Little Endian encoding as where it defaults to UTF16 Big Endian encoding on
the reference platform. This results in the same data representing two
different characters.

Validated by running unit tests and passing.

### DIFF
--- a/tests/unittests/Foundation/NSCharacterSetTests.m
+++ b/tests/unittests/Foundation/NSCharacterSetTests.m
@@ -219,10 +219,4 @@ TEST(NSCharacterSet, Polymorphic_Creators) {
 
     set = [NSMutableCharacterSet characterSetWithRange:NSMakeRange(0, 32)];
     ASSERT_TRUE([set isKindOfClass:[NSMutableCharacterSet class]]);
-
-    set = [NSMutableCharacterSet characterSetWithBitmapRepresentation:nil];
-    ASSERT_TRUE([set isKindOfClass:[NSMutableCharacterSet class]]);
-
-    set = [NSMutableCharacterSet characterSetWithContentsOfFile:nil];
-    ASSERT_TRUE([set isKindOfClass:[NSMutableCharacterSet class]]);
 }
--- a/tests/unittests/Foundation/ReferenceFoundation/TestNSCharacterSet.mm
+++ b/tests/unittests/Foundation/ReferenceFoundation/TestNSCharacterSet.mm
@@ -97,12 +97,14 @@ TEST(NSCharacterSet, Bitmap) {
 
 TEST(NSCharacterSet, Mutables) {
     auto attachmentCharacterUnichar = static_cast<unichar>(0xFFFC);
-
-    NSRange attachmentCharacterRange =
-        NSMakeRange(attachmentCharacterUnichar, (attachmentCharacterUnichar - attachmentCharacterUnichar) + 1);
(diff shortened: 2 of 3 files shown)


rajsesh-msft:
In the commit message, please use Fix #<issue number> to automatically close issue on merge.

DHowett-MSFT:
_NB:_ Since GitHub lets you edit the emssage when squashing, it could be placed there too. :smile:

DHowett:
Each `mutableCopy` here requires an autorelease. Even the existing factory methods that you haven't changed.

I noticed that you removed only half of the `NSMutableCharacterSet` factory methods; are all the other ones specified to return mutable versions of the respective character sets?

If so, is it more valid for us to make the cases you're deleting work natively, and say that the reference platform is _incorrect_?

tests/unittests/Foundation/ReferenceFoundation/TestNSCharacterSet.mm
```diff
@@ -98,11 +98,10 @@
 TEST(NSCharacterSet, Mutables) {
     auto attachmentCharacterUnichar = static_cast<unichar>(0xFFFC);
```
DHowett: You can only assume this will be little endian on little endian platforms. If you want to be 100% sure, please encode it as separate LE bytes, `const unsigned char attachmentCharacterUnicharBytes[] = {0xFC, 0xFF};`

ms-jihua: alternatively, do an ifdef on __BIG_ENDIAN__, __LITTLE_ENDIAN__

---
In reply to: [73653571](https://github.com/Microsoft/WinObjC/pull/823#discussion_r73653571) [](ancestors = 73653571)

rajsesh-msft:
:shipit:

DHowett-MSFT:
:shipit:

tests/unittests/Foundation/NSCharacterSetTests.m
```diff
@@ -219,10 +219,4 @@ void testCharacter(NSCharacterSet* charSet, NSString* testString) {
 
     set = [NSMutableCharacterSet characterSetWithRange:NSMakeRange(0, 32)];
     ASSERT_TRUE([set isKindOfClass:[NSMutableCharacterSet class]]);
-
-    set = [NSMutableCharacterSet characterSetWithBitmapRepresentation:nil];
-    ASSERT_TRUE([set isKindOfClass:[NSMutableCharacterSet class]]);
-
-    set = [NSMutableCharacterSet characterSetWithContentsOfFile:nil];
```
DHowett-MSFT: Why remove these? What's it mean?

MSFTFox: These tests are entirely invalid. You're not allowed to call characterSetWithBitmapRepresentation:nil. The argument must be non-nil. If it's nil it causes a failure on OSX.

PR #824, opened 2016-08-05
vImage API implementation with SSE2 optimizations (Accelerate framework)

shahmeeresmail:
The following vImage APIs have been implemented and have both SSE2 optimized and regular code paths (for non x86 based systems) where applicable:
          vImageConvert_ARGB8888toPlanar8
          vImageConvert_BGRA8888toPlanar8
          vImageConvert_RGBA8888toPlanar8
          vImageConvert_Planar8toRGB888
          vImageConvert_Planar8toARGB8888
          vImageConvert_Planar8toPlanarF
          vImageConvert_PlanarFtoPlanar8
          vImageUnpremultiplyData_ARGB8888
          vImageUnpremultiplyData_RGBA8888
          vImageUnpremultiplyData_BGRA8888
          vImageBuffer_Init
          vImageBuffer_InitWithCGImage
          vImageCreateCGImageFromBuffer

Also updated Woccatalog's ImagesViewController and added unit tests to validate implementation. Validated that the output matched that produced on iOS.

WOCCATALOG cmykStripes function:
- Loads image loaded into pixel buffer
- Loads packed pixel buffer into 4 buffers each containing a color or alpha plane
- Screen divided into 3 segment vertically and one color component removed from each segment by memsetting the planar buffer for each row of planar data. When pixel data combined, effectively creates 3 highlighted strips containing cyan, magenta, yellow tinted data
- Alpha planar data set to 127 (effectively 0.5)
- Planar data recombined into packed pixel buffer to produce image with cyan, magenta and yellow highlights
- Pixelbuffer unpremultiplied with alpha to brighten the image
- Unpremultiplied pixelbuffer output to a cgImage


samples/WOCCatalog/WOCCatalog/ImagesViewController.m
```diff
@@ -38,6 +40,180 @@ + (UIImage*)scaleImage:(CGImageRef)imageRef scaledRect:(CGRect)rect quality:(CGI
     return scaledImage;
 }
 
++ (void)initCGImageFormat:(CGImageRef)imageRef formatInfo:(vImage_CGImageFormat*)formatInfo {
```
rajsesh-msft: >initCGImageFormat [](start = 8, length = 17)

nit: I would call this getCGImageFormat instead.

shahmeeresmail: If format was an enum or another dword sized value, then I figure a 'get' prefix would be appropriate.  However, in this case all values in a multi-element structure is being set so I feel that an 'init' prefix would be more appropriate here.

Frameworks/Accelerate/vImage.cpp
```diff
@@ -951,4 +935,876 @@ vImage_Error vImageMatrixMultiply_ARGB8888(const vImage_Buffer* src,
...
+@Notes Packed buffer byte order notation is little endian
+*/
+vImage_Error vImageUnpremultiplyData_ARGB8888(const vImage_Buffer* src, const vImage_Buffer* dest, vImage_Flags flags) {
+    return _vImageUnpremultiplyData_8888<0>(src, dest, flags);
+}
+
+/**
+@Status Interoperable
+*/
+vImage_Error vImageBuffer_Init(
```
rajsesh-msft: >vImageBuffer_Init [](start = 13, length = 17)

I am not seeing any documentation for this, is this private for WinOBJC? If so, these methods should not require annotations.  Also, would make sense to export vImage_free() instead of relying on application calling free.  Expecting free to be called by app would imply that we cannot run debug version of app with release version of the SDK.

samples/WOCCatalog/WOCCatalog/ImagesViewController.m
rajsesh-msft: For the purposes of this feature, it would make sense to add this to a new page instead.  My plan is to move all the non UX related work to their own samples, and having it in a separate page would make it easier.  Also will help keep the image view page relatively simple.

Frameworks/Accelerate/vImage.cpp
```diff
@@ -951,4 +935,876 @@ vImage_Error vImageMatrixMultiply_ARGB8888(const vImage_Buffer* src,
...
+        if (buffer->data != nullptr) {
+            // SIMD load/store instructions operate best on 16byte aligned memory and in blocks of 16bytes
+            // Padding to ensure that for unaligned start addresses, at the very least:
+            // 1. A single 16pixel block can be operated on per row
+            // 2. Every second row will be 16byte aligned
+            const uint32_t bytePaddingForIdealAlignment =
+                (c_vImageBlockSizeBytes - ((uint32_t)(buffer->data) & (c_vImageBlockSizeBytes - 1))) % c_vImageBlockSizeBytes;
+            buffer->rowBytes = alignedPitch + bytePaddingForIdealAlignment;
+        } else {
+            // Try again without additional padding
```
rajsesh-msft: >y again without additional padding [](start = 17, length = 34)

I am not sure if this is necessary.  For an image that is 2K rows tall, we will add ~32K memory, which given the size of the image itself, should be trivial.  Could simplify logic elsewhere to assume padding is always present.

Frameworks/Accelerate/vImage.cpp
```diff
@@ -951,4 +935,876 @@ vImage_Error vImageMatrixMultiply_ARGB8888(const vImage_Buffer* src,
...
+            returnCode = kvImageMemoryAllocationError;
+        }
+    }
+
+    return returnCode;
+}
+
+// Alias vImageConvert_ARGB8888toPlanar8 with vImageConvert_BGRA8888toPlanar8 and vImageConvert_RGBA8888toPlanar8
+_vImageFunctionInterfaceCvt8888ToPlanar8 vImageConvert_BGRA8888toPlanar8 = vImageConvert_ARGB8888toPlanar8;
+_vImageFunctionInterfaceCvt8888ToPlanar8 vImageConvert_RGBA8888toPlanar8 = vImageConvert_ARGB8888toPlanar8;
```
rajsesh-msft: >vImageConvert_RGBA8888toPlanar8 [](start = 41, length = 31)

Won't this result in B and R being swapped?

Frameworks/Accelerate/vImage.cpp
shahmeeresmail: This is an API that does not appear in the official documentation but is certainly part of the API since many samples use it and it shows up on the iOS headers. Some documentation is available here:
https://developer.apple.com/library/mac/releasenotes/Performance/RN-vecLib/

Frameworks/Accelerate/vImage.cpp
shahmeeresmail: This is a quick check and only pads when necessary. The other APIs which use vImage_Buffer structures will know whether or not padding is present based on the rowBytes field.

Frameworks/Accelerate/vImage.cpp
shahmeeresmail: It won't because the inputs will also be swapped (ie: the interface to the function is vImageConvert_RGBA8888toPlanar8(redPlaneBuffer, greenPlaneBuffer, bluePlaneBuffer, alphaPlaneBuffer) )

The function vImageConvert_ARGB8888toPlanar8 interleaves the planes in the same order as the parameters are passed in so if the inputs order is changed, the component order in the resulting pixel buffer will change with it.

Frameworks/Accelerate/vImage.cpp
shahmeeresmail: Also, this isn't the only way a client can create a vImage_Buffer. There is nothing stopping a developer from allocating memory and filling out the data, height, width, rowBytes fields themselves. Therefore, padding can't be assumed and needs to be checked where ever this structure is passed in.

Frameworks/Accelerate/vImage.cpp
shahmeeresmail: Allocating and freeing a vImage_Buffer structure is up to the client. They can use vImageBuffer_Init to allocate a structure or provide alignment information to allocate the buffer optimally themselves by using functions like alligned_malloc with the provided allocation information or they could choose to just use malloc and set the rowbytes equal to bpp*width with no padding or alignment.

Similarly, the client must free the allocation. The framework does not track which allocations are allocated through vImageBuffer_Init and if it wasn't, it doesn't know whether aligned_malloc, mm_malloc, malloc or some other memory allocation function was used. Since it doesn't know how it was allocated, it can't free the allocation either.

rajsesh-msft: Thanks Shameer, I'll do some manual validations on this as well as CI build. In the mean time, the only remaining comment I have is moving the new tests out of the ImagesView into a new view.

---
In reply to: [75983555](https://github.com/Microsoft/WinObjC/pull/824#discussion_r75983555) [](ancestors = 75983555)

rajsesh-msft:
@shahmeeresmail both release and debug builds failed..
016-08-29T18:16:15.0767454Z          ..\..\..\Frameworks\Accelerate\vImage.cpp(1538): error C2059: syntax error: 'return' [D:\Build\Agent1\_work\1\s\build\Accelerate\lib\AccelerateLib.vcxproj]
2016-08-29T18:16:15.0767454Z          ..\..\..\Frameworks\Accelerate\vImage.cpp(1539): error C2059: syntax error: '}' [D:\Build\Agent1\_work\1\s\build\Accelerate\lib\AccelerateLib.vcxproj]
2016-08-29T18:16:15.0767454Z          ..\..\..\Frameworks\Accelerate\vImage.cpp(1539): error C2143: syntax error: missing ';' before '}' [D:\Build\Agent1\_work\1\s\build\Accelerate\lib\AccelerateLib.vcxproj]
2016-08-29T18:16:15.0767454Z          ..\..\..\Frameworks\Accelerate\vImage.cpp(1542): error C2143: syntax error: missing ';' before '{' [D:\Build\Agent1\_work\1\s\build\Accelerate\lib\AccelerateLib.vcxproj]
2016-08-29T18:16:15.0923655Z          ..\..\..\Frameworks\Accelerate\vImage.cpp(1542): error C2447: '{': missing function header (old-style formal list?) [D:\Build\Agent1\_work\1\s\build\Accelerate\lib\AccelerateLib.vcxproj]
2016-08-29T18:16:15.0923655Z          ..\..\..\Frameworks\Accelerate\vImage.cpp(1694): error C2059: syntax error: 'return' [D:\Build\Agent1\_work\1\s\build\Accelerate\lib\AccelerateLib.vcxproj]
2016-08-29T18:16:15.0923655Z          ..\..\..\Frameworks\Accelerate\vImage.cpp(1695): error C2059: syntax error: '}' [D:\Build\Agent1\_work\1\s\build\Accelerate\lib\AccelerateLib.vcxproj]
2016-08-29T18:16:15.0923655Z          ..\..\..\Frameworks\Accelerate\vImage.cpp(1695): error C2143: syntax error: missing ';' before '}' [D:\Build\Agent1\_work\1\s\build\Accelerate\lib\AccelerateLib.vcxproj]
2016-08-29T18:16:15.0923655Z          ..\..\..\Frameworks\Accelerate\vImage.cpp(1702): error C2988: unrecognizable template declaration/definition [D:\Build\Agent1\_work\1\s\build\Accelerate\lib\AccelerateLib.vcxproj]
2016-08-29T18:16:15.0923655Z          ..\..\..\Frameworks\Accelerate\vImage.cpp(1702): error C2143: syntax error: missing ';' before '<' [D:\Build\Agent1\_work\1\s\build\Accelerate\lib\AccelerateLib.vcxproj]
2016-08-29T18:16:15.0923655Z          ..\..\..\Frameworks\Accelerate\vImage.cpp(1702): error C2182: '_vImageUnpremultiplyPixel': illegal use of type 'void' [D:\Build\Agent1\_work\1\s\build\Accelerate\lib\AccelerateLib.vcxproj]
2016-08-29T18:16:15.0923655Z          ..\..\..\Frameworks\Accelerate\vImage.cpp(1702): error C2433: '_vImageUnpremultiplyPixel': 'inline' not permitted on data declarations [D:\Build\Agent1\_work\1\s\build\Accelerate\lib\AccelerateLib.vcxproj]
2016-08-29T18:16:15.0923655Z          ..\..\..\Frameworks\Accelerate\vImage.cpp(1702): error C2059: syntax error: '<' [D:\Build\Agent1\_work\1\s\build\Accelerate\lib\AccelerateLib.vcxproj]
2016-08-29T18:16:15.0923655Z          ..\..\..\Frameworks\Accelerate\vImage.cpp(1710): error C2143: syntax error: missing ';' before '{' [D:\Build\Agent1\_work\1\s\build\Accelerate\lib\AccelerateLib.vcxproj]
2016-08-29T18:16:15.0923655Z          ..\..\..\Frameworks\Accelerate\vImage.cpp(1710): error C2447: '{': missing function header (old-style formal list?) [D:\Build\Agent1\_work\1\s\build\Accelerate\lib\AccelerateLib.vcxproj]
2016-08-29T18:16:15.0923655Z          ..\..\..\Frameworks\Accelerate\vImage.cpp(1722): error C2065: '_vImageUnpremultiplyData_8888': undeclared identifier [D:\Build\Agent1\_work\1\s\build\Accelerate\lib\AccelerateLib.vcxproj]
2016-08-29T18:16:15.0923655Z          ..\..\..\Frameworks\Accelerate\vImage.cpp(1730): error C2065: '_vImageUnpremultiplyData_8888': undeclared identifier [D:\Build\Agent1\_work\1\s\build\Accelerate\lib\AccelerateLib.vcxproj]
2016-08-29T18:16:15.1079964Z          ..\..\..\Frameworks\Accelerate\vImage.cpp(1738): error C2065: '_vImageUnpremultiplyData_8888': undeclared identifier [D:\Build\Agent1\_work\1\s\build\Accelerate\lib\AccelerateLib.vcxproj]

shahmeeresmail:
I am not able to repro this failure, however, I am resyncing and rebuilding my local repo to see if I can repro it that way.

shahmeeresmail:
I resynced and rebuilt and still cannot repro the issue. Based on the first error, it looks as though vImage.h might be missing from the copy that is being built on your end. I checked all the commits in this branch though and vImage.cpp is always paired with vImage.h and it looks like you have the latest copy of vImage.cpp (line numbers match) and Accelerate.vcxproj (otherwise vImage.cpp wouldn't build). Therefore, it seems as though the most likely thing that has happened here is that vImage.h was not pulled prior to the build being generated. Can you ensure that when syncing to the branch that the latest copy of include/vImage.h was pulled?

rajsesh-msft:
FAILED  ] 1 test, listed below:
[  FAILED  ] Mach.MachTaskInfoThreadTimes

This test has been known to fail sometimes (probably needs to be removed).  Overriding it.

PR #833, opened 2016-08-05
Fix NSPointerFunctions tests to match reference platform

### LINKED ISSUE #762: NSPointerFunctions tests fail on OSX
[ RUN      ] NSPointerFunctions.OpaquePersonality
Foundation/NSPointerFunctionsTests.mm:37: Failure
Expected: (hashFunction(item1, sizeFunction)) != (hashFunction(item3, sizeFunction)), actual: 564006119 vs 564006119
[  FAILED  ] NSPointerFunctions.OpaquePersonality (0 ms)

Note that the unsupported config warnings below are stated to be expected:

[ RUN      ] NSPointerFunctions.InvalidConfig
[  INFO ] Invalid config messages during this test are expected:
2016-08-01 10:46:14.109 Foundation.UnitTests.exe[4412:468597] *** An unsupported PointerFunctions configuration was requested, probably for use by NSMapTable, NSHashTable, or NSPointerArray.  The requested configuration fails due to objects with none of strong, weak, or opaque memory
2016-08-01 10:46:14.110 Foundation.UnitTests.exe[4412:468597] *** An unsupported PointerFunctions configuration was requested, probably for use by NSMapTable, NSHashTable, or NSPointerArray.  The requested configuration would create only a partially initialized NSPointerFunctions object
Foundation/NSPointerFunctionsTests.mm:102: Failure
Expected: [NSPointerFunctions pointerFunctionsWithOptions:7] throws an exception.
  Actual: it doesn't.
[  FAILED  ] NSPointerFunctions.InvalidConfig (2 ms)

mnithish:
Fixes #762

### DIFF
--- a/tests/unittests/Foundation/NSPointerFunctionsTests.mm
+++ b/tests/unittests/Foundation/NSPointerFunctionsTests.mm
@@ -80,7 +80,8 @@ TEST(NSPointerFunctions, OpaquePersonality) {
     const void* ptr3 = (void*)((long)ptr1 + 1);
 
     testEqualFunction(functions, ptr1, ptr2, ptr3);
-    testHashFunction(functions, ptr1, ptr2, ptr3);
+    // Disabling hash function check here as performing hash on opaque values cannot be accuratly matched with the reference platform.
+    // testHashFunction(functions, ptr1, ptr2, ptr3);
 }
 
 TEST(NSPointerFunctions, IntegerPersonality) {
@@ -95,7 +96,9 @@ TEST(NSPointerFunctions, IntegerPersonality) {
     testHashFunction(functions, (void*)int1, (void*)int2, (void*)int3);
 }
 
-TEST(NSPointerFunctions, InvalidConfig) {
+// Disabling test from running on OSX as the reference platform behavior is to log error and then return a nil when an invalid config
+// is used. We have decided to be a little more strict on our platform by throwing an exception instead.
+OSX_DISABLED_TEST(NSPointerFunctions, InvalidConfig) {
     LOG_INFO("Invalid config messages during this test are expected:");
 
     // Not a value of NSPointerFunctionsOptions
--- a/tests/unittests/Foundation/NSPointerFunctionsTests.mm
+++ b/tests/unittests/Foundation/NSPointerFunctionsTests.mm
@@ -80,7 +80,7 @@ TEST(NSPointerFunctions, OpaquePersonality) {
     const void* ptr3 = (void*)((long)ptr1 + 1);
 
     testEqualFunction(functions, ptr1, ptr2, ptr3);
-    // Disabling hash function check here as performing hash on opaque values cannot be accuratly matched with the reference platform.
+    // Disabling hash function check here as performing hash on opaque values cannot be accurately matched with the reference platform.
     // testHashFunction(functions, ptr1, ptr2, ptr3);
 }
 


mnithish:

mnithish:

tests/unittests/Foundation/NSPointerFunctionsTests.mm
```diff
@@ -80,7 +80,8 @@ NSUInteger NSPointerFunctionsTestStructSizeFunction(const void* item) {
     const void* ptr3 = (void*)((long)ptr1 + 1);
 
     testEqualFunction(functions, ptr1, ptr2, ptr3);
-    testHashFunction(functions, ptr1, ptr2, ptr3);
+    // Disabling hash function check here as performing hash on opaque values cannot be accuratly matched with the reference platform.
```
ms-jihua: >tl [](start = 94, length = 2)

nit: typo

tests/unittests/Foundation/NSPointerFunctionsTests.mm
```diff
@@ -80,7 +80,8 @@ NSUInteger NSPointerFunctionsTestStructSizeFunction(const void* item) {
...
+    // testHashFunction(functions, ptr1, ptr2, ptr3);
```
ms-jihua: >// testHashFunction(functions, ptr1, ptr2, ptr3); [](start = 4, length = 49)

What was the thinking behind commenting this out, instead of a) keeping it in an #ifdef WINOBJC, or b) deleting it?

ms-jihua:
Title for this PR seems incorrect :)

mnithish:
Thanks github. I have no idea how it populated this

---
In reply to: [237972440](https://github.com/Microsoft/WinObjC/pull/833#issuecomment-237972440) [](ancestors = 237972440)

tests/unittests/Foundation/NSPointerFunctionsTests.mm
mnithish: This is kind of a test for which we need to know how it has been implemented for it to pass. Our implementation of hash for opaque types is by left shifting the value by 3, so this test will not pass for opaque types that differ only by the first three significant bits. Reference platform seems to be doing a right shift. It is commented so people know what check has specifically been disabled when they look.

I can put it under WINOBJC but don't see a real value with this test. I am open to the idea

---
In reply to: [73762989](https://github.com/Microsoft/WinObjC/pull/833#discussion_r73762989) [](ancestors = 73762989)

DHowett-MSFT:
You can edit it!

---
In reply to: [237972693](https://github.com/Microsoft/WinObjC/pull/833#issuecomment-237972693) [](ancestors = 237972693,237972440)

DHowett-MSFT:
:shipit:

tests/unittests/Foundation/NSPointerFunctionsTests.mm
ms-jihua: Hmm, I think there's value in knowing that our impl for this hash function doesn't collide. The scenario you describe where the first three bits get wiped out is _possible_ but shouldn't really ever happen with the way allocation typically works.

Does it make sense for us to change our hash to a right shift instead and leave this commented out (as it'll be much more prone to failure)?

---
In reply to: [73764075](https://github.com/Microsoft/WinObjC/pull/833#discussion_r73764075) [](ancestors = 73764075,73762989)

DHowett-MSFT: I would generally agree, but we can't assume that people are only putting allocated memory into these things. There's stuff like function pointers from different modules (different load addresses), section/segment headers, etc.

The problem is that it doesn't even make sense to "hash" pointers. They're already 32-64 full bits of entropy.

---
In reply to: [73764838](https://github.com/Microsoft/WinObjC/pull/833#discussion_r73764838) [](ancestors = 73764838,73764075,73762989)

ms-jihua: Fair enough with regard to value of testing. I'm still wondering whether we should change to a right shift to match the reference platform.

---
In reply to: [73765315](https://github.com/Microsoft/WinObjC/pull/833#discussion_r73765315) [](ancestors = 73765315,73764838,73764075,73762989)

ms-jihua: Discussed offline, we'll keep this as-is.

---
In reply to: [73766353](https://github.com/Microsoft/WinObjC/pull/833#discussion_r73766353) [](ancestors = 73766353,73765315,73764838,73764075,73762989)

ms-jihua:
:shipit:

PR #840, opened 2016-08-05
Fix NSSet ObjectsPassingTest unit test

### LINKED ISSUE #767: NSSet.ObjectsPassingTest fails on OSX
[ RUN      ] NSSet.ObjectsPassingTest
Foundation/NSSetTests.mm:65: Failure
Value of: actual
  Actual: {(
    6,
    2
)}
Expected: expectedEvensLessThanFive
Which is: {(
    4,
    2
)}
[  FAILED  ] NSSet.ObjectsPassingTest (1 ms)

mnithish:
Fixes #767

### DIFF
--- a/tests/unittests/Foundation/NSSetTests.mm
+++ b/tests/unittests/Foundation/NSSetTests.mm
@@ -36,8 +36,8 @@ TEST(NSSet, ContainsObject) {
 
 TEST(NSSet, SetValueForKey) {
     NSMutableDictionary* dict1 = [[@{} mutableCopy] autorelease];
-    NSMutableDictionary* dict2 = [[@{@"a" : @"1"} mutableCopy] autorelease];
-    NSMutableDictionary* dict3 = [[@{@"b" : @"2"} mutableCopy] autorelease];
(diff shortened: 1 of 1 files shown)


mnithish:

DHowett-MSFT:
LOL.

DHowett-MSFT:
Why did this work on our platform?

DHowett-MSFT:
:shipit:

ms-jihua:
:shipit:

ms-jihua:
Just guessing here:

Our platform happens to order the set in such a way that the test passes. The ref plat makes no such guarantee, so we hit the 6 early and stop.

---
In reply to: [237985558](https://github.com/Microsoft/WinObjC/pull/840#issuecomment-237985558) [](ancestors = 237985558)

mnithish:
Correct - reference platform seems to store the value in BST so when we enumerate they seem to be give the value in the level order of how they were stored, so we hit 5 before we hit 2

---
In reply to: [237985796](https://github.com/Microsoft/WinObjC/pull/840#issuecomment-237985796) [](ancestors = 237985796,237985558)

PR #844, opened 2016-08-06
Fix NSCalendar tests that fail on OSX

### LINKED ISSUE #744: NSCalendar tests fail on OSX
https://github.com/Microsoft/WinObjC/issues/743 may be related.

[ RUN      ] NSCalendar.GetDates
unknown file: Failure
[  FAILED  ] NSCalendar.GetDates (61 ms)

[ RUN      ] NSCalendar.CompareDates
Foundation/NSCalendarTests.mm:240: Failure
Value of: NSOrderedAscending
  Actual: -1
Expected: [calendar compareDate:date1 toDate:date2 toUnitGranularity:NSCalendarUnitNanosecond]
Which is: 0
[  FAILED  ] NSCalendar.CompareDates (0 ms)

[ RUN      ] NSCalendar.PreserveSmallerUnitsMatching
Foundation/NSCalendarTests.mm:345: Failure
Value of: actualDate
  Actual: 2012-02-29 08:00:00 +0000
Expected: expectedDate
Which is: 2012-02-29 11:04:05 +0000
[  FAILED  ] NSCalendar.PreserveSmallerUnitsMatching (5 ms)

[ RUN      ] NSCalendar.weekOfUnitMatching
Foundation/NSCalendarTests.mm:405: Failure
Value of: actualDate
  Actual: 2000-01-30 08:00:00 +0000
Expected: expectedDate
Which is: 2000-02-05 08:00:00 +0000
[  FAILED  ] NSCalendar.weekOfUnitMatching (7 ms)

[ RUN      ] NSCalendar.NSRangeValidation
Foundation/NSCalendarTests.mm:486: Failure
Value of: 5
Expected: range.length
Which is: 6
[  FAILED  ] NSCalendar.NSRangeValidation (0 ms)

MSFTFox:
Fixes #744

NSCalendar Tests have some failures as well as some unexpected behavior.
Some of the tests actually produce the correct results as described in the
documentation despite OSX actually producing incorrect results. These are
when the nanoseconds of two dates are compared using a calendar, when using
a matching flag that preserves smaller units when matching a leap date such as
February 29th, as well as the range of weeks in months always returning 6
despite not every month having 6 weeks.

These tests have been disabled for OSX.

### DIFF
--- a/Frameworks/CoreFoundation/Locale.subproj/CFCalendar.c
+++ b/Frameworks/CoreFoundation/Locale.subproj/CFCalendar.c
@@ -744,14 +744,14 @@ static CFRange __CFCalendarGetRangeOfUnit2(CFCalendarRef calendar, CFCalendarUni
             }
         break;
     // WINOBJC: Add cases for weekOf units. ICU will handle them below.
-    case kCFCalendarUnitWeekOfYear:
     case kCFCalendarUnitWeekOfMonth:
-    // WINOBJC: Asking for the range of WeekOfMonth in Year always returns 6.
+        // WINOBJC: Asking for the range of WeekOfMonth in Year always returns 6.
         if (biggerUnit == kCFCalendarUnitYear) {
             range.location = 1;
             range.length = 6;
             break;
         }
+    case kCFCalendarUnitWeekOfYear:
     case kCFCalendarUnitWeek:
             switch (biggerUnit) {
             case kCFCalendarUnitMonth:
--- a/Frameworks/CoreFoundation/Locale.subproj/CFCalendar.c
+++ b/Frameworks/CoreFoundation/Locale.subproj/CFCalendar.c
@@ -746,6 +746,12 @@ static CFRange __CFCalendarGetRangeOfUnit2(CFCalendarRef calendar, CFCalendarUni
     // WINOBJC: Add cases for weekOf units. ICU will handle them below.
     case kCFCalendarUnitWeekOfYear:
     case kCFCalendarUnitWeekOfMonth:
+    // WINOBJC: Asking for the range of WeekOfMonth in Year always returns 6.
+        if (biggerUnit == kCFCalendarUnitYear) {
+            range.location = 1;
+            range.length = 6;
+            break;
+        }
     case kCFCalendarUnitWeek:
             switch (biggerUnit) {
             case kCFCalendarUnitMonth:
@@ -961,6 +967,13 @@ Boolean _CFCalendarAddComponentsV(CFCalendarRef calendar, /* inout */ CFAbsolute
     ucal_setMillis(calendar->_cal, udate, &status);
     char ch = *componentDesc;
     while (ch) {
+        // WINOBJC: ICU function ucal_add does not accept Nanoseconds. Skip this component as it causes failure.
+        if (ch == '#') {
+            vector++;
+            componentDesc++;
+            ch = *componentDesc;
+            continue;
+        }
         UCalendarDateFields field = __CFCalendarGetICUFieldCodeFromChar(ch);
             int amount = *vector;
         if (options & kCFCalendarComponentsWrap) {
--- a/Frameworks/CoreFoundation/Locale.subproj/CFCalendar.c
+++ b/Frameworks/CoreFoundation/Locale.subproj/CFCalendar.c
@@ -744,14 +744,8 @@ static CFRange __CFCalendarGetRangeOfUnit2(CFCalendarRef calendar, CFCalendarUni
             }
         break;
     // WINOBJC: Add cases for weekOf units. ICU will handle them below.
-    case kCFCalendarUnitWeekOfMonth:
-        // WINOBJC: Asking for the range of WeekOfMonth in Year always returns 6.
-        if (biggerUnit == kCFCalendarUnitYear) {
-            range.location = 1;
-            range.length = 6;
-            break;
-        }
     case kCFCalendarUnitWeekOfYear:
+    case kCFCalendarUnitWeekOfMonth:
     case kCFCalendarUnitWeek:
             switch (biggerUnit) {
             case kCFCalendarUnitMonth:
@@ -782,6 +776,7 @@ static CFRange __CFCalendarGetRangeOfUnit2(CFCalendarRef calendar, CFCalendarUni
     }
     return range;
 
+    // TODO 8463791 : rangeOfUnit can return incorrect results
     calculate:;
     ucal_clear(calendar->_cal);
     UCalendarDateFields smallField = __CFCalendarGetICUFieldCode(smallerUnit);
--- a/Frameworks/Foundation/NSCFCalendar.mm
+++ b/Frameworks/Foundation/NSCFCalendar.mm
@@ -670,11 +670,14 @@ static NSUInteger s_retryAttempts = 55;
     // Issue 650: NSCalendar Certain Units Need Optimized and Combination of Interpretive Units Match Incorrect Dates
     NSCalendarUnit interpretiveUnit = 0;
 
+    NSDateComponents* smallerComps = [[[NSDateComponents alloc] init] autorelease];
     for (int i = 0; i < _countof(s_NSDateComponentsIndividualFlags); i++) {
         // If a largest unit exists and the matching options does not specify preserving smaller units
-        if (mathematicalUnit != 0 && i < NSDateComponentsMathematicalUnitCutoff &&
-            !((opts & NSCalendarMatchPreviousTimePreservingSmallerUnits) || (opts & NSCalendarMatchNextTimePreservingSmallerUnits))) {
-            // Set this smaller unit to undefined.
+        if (mathematicalUnit != 0 && i < NSDateComponentsMathematicalUnitCutoff) {
+            if (((opts & NSCalendarMatchPreviousTimePreservingSmallerUnits) || (opts & NSCalendarMatchNextTimePreservingSmallerUnits))) {
+                [smallerComps setValue:[startComps valueForComponent:s_NSDateComponentsIndividualFlags[i]]
+                          forComponent:s_NSDateComponentsIndividualFlags[i]];
+            }
             [newComps setValue:NSUndefinedDateComponent forComponent:s_NSDateComponentsIndividualFlags[i]];
         }
 
@@ -761,8 +764,8 @@ static NSUInteger s_retryAttempts = 55;
                 numIters++;
             }
 
-            // If matchnext is specified and an exact match is not found after the initial try.
-            if (matchNext && !([comps _componentsMatch:newComps forUnits:unitFlags])) {
+            // If stopOnFirstMatch is specified and an exact match is not found after the initial try.
+            if (stopOnFirstMatch && !([comps _componentsMatch:newComps forUnits:unitFlags])) {
                 // Find the unit that made this match fail.
                 NSCalendarUnit highestUnitThatFailed = 0;
 
@@ -779,7 +782,17 @@ static NSUInteger s_retryAttempts = 55;
 
                 NSDateComponents* oneUnit = [[[NSDateComponents alloc] init] autorelease];
 
-                [oneUnit setValue:(searchForwards ? 1 : -1) forComponent:highestUnitThatFailed];
+                // If matching does not want the next value then do not update the highest unit.
+                [oneUnit setValue:(matchNext ? 1 : 0) forComponent:highestUnitThatFailed];
+
+                if ((opts & NSCalendarMatchPreviousTimePreservingSmallerUnits) || (opts & NSCalendarMatchNextTimePreservingSmallerUnits)) {
+                    for (int i = 0; i < _countof(s_NSDateComponentsIndividualFlags); i++) {
+                        NSUInteger value = [smallerComps valueForComponent:s_NSDateComponentsIndividualFlags[i]];
+                        if (value != NSUndefinedDateComponent) {
+                            [oneUnit setValue:value forComponent:s_NSDateComponentsIndividualFlags[i]];
+                        }
+                    }
+                }
 
                 bestdate = [self dateByAddingComponents:oneUnit toDate:bestdate options:opts];
             }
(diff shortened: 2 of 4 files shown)


Frameworks/Foundation/NSCFCalendar.mm
```diff
@@ -594,6 +594,42 @@ - (void)enumerateDatesStartingAfterDate:(NSDate*)start
...
+
+    // This is a valid solo option as well as when used in conjunction with the others.
+    if (!isOptionsValid) {
+        isOptionsValid = ((opts & NSCalendarMatchStrictly) != 0);
+    }
+
+    THROW_NS_IF_FALSE_MSG(E_INVALIDARG,
+                          (isOptionsValid && !optionsAreInContrast),
+                          "Options must specify exactly one of these matching options: NSCalendarMatchPreviousTimePreservingSmallerUnits, "
+                          "NSCalendarMatchNextTimePreservingSmallerUnits, NSCalendarMatchNextTime");
```
rajsesh-msft: This whole block could be simplified as:

if ( opts & NSCalendarMatchStrictly) != 0 && opts & NSCalendarMatchPreviousTimePreservingSmallerUnits) != 0 && (opts & NSCalendarMatchNextTimePreservingSmallerUnits != 0 || opts & NSCalendarMatchNextTime != 0)) {
         THROW_....
}

rajsesh-msft: or not.. you do want one of these flags present, i see it now :)

---
In reply to: [73946729](https://github.com/Microsoft/WinObjC/pull/844#discussion_r73946729) [](ancestors = 73946729)

tests/unittests/Foundation/NSCalendarTests.mm
```diff
@@ -481,15 +472,69 @@
...
+        // No matching options
+        NSDate* date = [calendar nextDateAfterDate:constantDate matchingComponents:comps options:0];
+
+        ASSERT_TRUE_MSG(false, "Should not have gotten here. nextDateAfterDate should throw an INVALID_ARG exception");
+    } @catch (NSException* e) {
+        ASSERT_OBJCEQ(e.name, @"NSInvalidArgumentException");
+    }
+}
+
+OSX_DISABLED_TEST(NSCalendar, DisabledCalendarTests) {
```
rajsesh-msft: >DisabledCalendarTests [](start = 30, length = 21)

I suggest you name this something else.

tests/unittests/Foundation/NSCalendarTests.mm
```diff
@@ -481,15 +472,75 @@
...
+        NSDate* expectedNextDay = [calendar dateWithEra:1 year:2010 month:1 day:2 hour:0 minute:0 second:0 nanosecond:0];
+
+        NSDateComponents* comps = [[NSDateComponents alloc] init];
+        comps.day = 2;
+
+        // No matching options
+        NSDate* date = [calendar nextDateAfterDate:constantDate matchingComponents:comps options:0];
+
+        ASSERT_TRUE_MSG(false, "Should not have gotten here. nextDateAfterDate should throw an INVALID_ARG exception");
+    } @catch (NSException* e) {
```
bviglietta: >@catch (NSException* e) [](start = 6, length = 23)

Have you validated this test on ARM? I don't think we're able to catch exceptions there.

tests/unittests/Foundation/NSCalendarTests.mm
```diff
@@ -481,15 +472,75 @@
...
+}
+
+OSX_DISABLED_TEST(NSCalendar, RangeOfWeekOfMonth) {
+    // Retrieving the range of a WeekOfMonth unit should return the number of weeks within the month for that date.
+    // The behavior on OSX however, is that the returned range is always 6 weeks regardless of the actual number of
+    // weeks in the given month.
+
+    NSCalendar* calendar = [NSCalendar calendarWithIdentifier:NSCalendarIdentifierGregorian];
+    NSDate* date = [calendar dateWithEra:1 year:2010 month:3 day:31 hour:3 minute:4 second:5 nanosecond:0];
+    NSRange range = [calendar rangeOfUnit:NSCalendarUnitWeekOfMonth inUnit:NSCalendarUnitYear forDate:date];
```
bviglietta: >rangeOfUnit:NSCalendarUnitWeekOfMonth inUnit:NSCalendarUnitYear [](start = 30, length = 63)

You're asking for the range within a year, not the range within a month. It sounds like OS X is doing the right thing by returning 6.

tests/unittests/Foundation/NSCalendarTests.mm
```diff
@@ -481,15 +472,75 @@
...
+    ASSERT_EQ(range.location, 1);
+    ASSERT_EQ(range.length, 5);
+}
+
+OSX_DISABLED_TEST(NSCalendar, NSCalendarMatchNextTimePreservingSmallerUnitsLeapDay) {
+    // NSCalendarMatchNextTimePreservingSmallerUnits as a matching option does not reflect the documentation accurately.
+    // When matching a date, this option should preserve all smaller values when finding the next date. The behavior on OSX
+    // has instead stripped the date of these smaller values. The date matched from OSX is 2/29/2012 00:00:00 when it should be
+    // 2/29/2012 03:04:05. Thus, this test is disabled.
```
bviglietta: This behavior on OSX is not limited to leap days. It always trims the smaller units when it finds an exact match. The documentation is a little ambiguous, but I think this is the intended behavior, and we should probably match it.

bviglietta:
:clock1:

Frameworks/CoreFoundation/Locale.subproj/CFCalendar.c
```diff
@@ -746,6 +746,12 @@ static CFRange __CFCalendarGetRangeOfUnit2(CFCalendarRef calendar, CFCalendarUni
     // WINOBJC: Add cases for weekOf units. ICU will handle them below.
     case kCFCalendarUnitWeekOfYear:
```
bviglietta: >case kCFCalendarUnitWeekOfYear: [](start = 4, length = 31)

Beware fallthrough.

Frameworks/CoreFoundation/Locale.subproj/CFCalendar.c
```diff
@@ -746,6 +746,12 @@ static CFRange __CFCalendarGetRangeOfUnit2(CFCalendarRef calendar, CFCalendarUni
...
     case kCFCalendarUnitWeekOfMonth:
+    // WINOBJC: Asking for the range of WeekOfMonth in Year always returns 6.
```
bviglietta: >Asking for the range of WeekOfMonth in Year always returns 6. [](start = 16, length = 61)

Why do we need a special case here? The logic in calculate: looks sound, as far as I can tell.

MSFTFox: Calculate provides the number of weeks within that particular month if the larger unit is the year. As you pointed out previously, this is incorrect. The behavior is to always return 6 weeks. Thus we can return 6 here instead of relying on calculate.

---
In reply to: [74336418](https://github.com/Microsoft/WinObjC/pull/844#discussion_r74336418) [](ancestors = 74336418)

bviglietta: OK, but that sounds like a bug in the calculate logic, or maybe in ICU underneath it. Do you know why you were getting that result? I would prefer to fix the error at the source if possible.

---
In reply to: [74342094](https://github.com/Microsoft/WinObjC/pull/844#discussion_r74342094) [](ancestors = 74342094,74336418)

bviglietta: For example, maybe ucal_add doesn't do what we want when told to add one WeekOfMonth to a date? If that's the case, maybe calculate should set fieldToAdd to Week instead of WeekOfMonth. Just speculating.

---
In reply to: [74354002](https://github.com/Microsoft/WinObjC/pull/844#discussion_r74354002) [](ancestors = 74354002,74342094,74336418)

MSFTFox: It seems to be an issue with ICU. Since ICU is operating on a particular date, and that date happens to not have 6 weeks in it, ICU will not update weekOfMonth to 6. It's essentially an out of bounds date component. Setting the smaller unit to week would instead return how many weeks are in the year. Even if we changed the unit to add to week, it would still not update to 6 weeks if the month does not have that because again, it's an out of bounds value for the unit of WeekOfMonth. To begin with, it doesn't make much sense to ask how many weeks are in a month in a year. I'm merely matching what the reference platform does in this case. I'd consider this a very low priority fringe case that doesn't provide much of a basis for changing core foundation code.

---
In reply to: [74354002](https://github.com/Microsoft/WinObjC/pull/844#discussion_r74354002) [](ancestors = 74354002,74342094,74336418)

bviglietta: I think the question makes sense: it's asking what is the range of WeekOfMonth values within a year. Normally (always?) it can range from 1 to 6.

---
In reply to: [74356593](https://github.com/Microsoft/WinObjC/pull/844#discussion_r74356593) [](ancestors = 74356593,74354002,74342094,74336418)

bviglietta: OK, I see the problem here. The calculate logic works by setting the starting date to the first moment in the larger unit, then advancing in increments of the smaller unit to find the min and max values. The problem is that this may not work correctly when the smaller unit doesn't divide the larger unit evenly.

For example, try getting the range of WeekOfMonth in Month for July 2016. The answer should be [1..6], but you will actually get [1..5].

This also explains why you sometimes get the wrong answer when asking for the range of WeekOfMonth in Year. It's not returning the number of weeks in the current month; it's advancing from Jan 1 week by week through the year, querying the WeekOfMonth each time.

The logic in calculate should be fixed.

---
In reply to: [74359393](https://github.com/Microsoft/WinObjC/pull/844#discussion_r74359393) [](ancestors = 74359393,74356593,74354002,74342094,74336418)

MSFTFox: Yeah, it definitely seems like calculate has a few problems in it. Fixing that is beyond the scope of fixing the tests for OSX though. Could you file another issue for fixing the calculate function?

---
In reply to: [74461625](https://github.com/Microsoft/WinObjC/pull/844#discussion_r74461625) [](ancestors = 74461625,74359393,74356593,74354002,74342094,74336418)

bviglietta: I opened bug 8463791 for this.

But I don't think we should put in the workaround that you have now for WeekOfMonth in Year. It's insufficient and not the right way to address the problem.

---
In reply to: [74468177](https://github.com/Microsoft/WinObjC/pull/844#discussion_r74468177) [](ancestors = 74468177,74461625,74359393,74356593,74354002,74342094,74336418)

MSFTFox: It begs the question as to why other units are fast tracked to return constant values here though.  Everything from second up through hour seems incorrect to then return constant values. Especially given that there hopefully aren't 60 seconds in a year. These should also be addressed in the work done for fixing calculate.

I think the workaround to return a constant 6 is sufficient to match the reference platform on this one case for the time being. I've yet to find a case where it's incorrect even if it's a faulty mechanism for determining the actual values. Especially given the other short cuts that core foundation has already taken for those smaller values I think it's acceptable until the actual calculations are fixed.

---
In reply to: [74473193](https://github.com/Microsoft/WinObjC/pull/844#discussion_r74473193) [](ancestors = 74473193,74468177,74461625,74359393,74356593,74354002,74342094,74336418)

MSFTFox: Looking into this further, I agree we shouldn't patch this with a constant 6 value. There's simply too many cases where this function returns the incorrect range and adding any additional patches just creates a bigger mess. Removing.

---
In reply to: [74478358](https://github.com/Microsoft/WinObjC/pull/844#discussion_r74478358) [](ancestors = 74478358,74473193,74468177,74461625,74359393,74356593,74354002,74342094,74336418)

tests/unittests/Foundation/NSCalendarTests.mm
```diff
@@ -481,15 +481,63 @@
...
+
+        NSDateComponents* comps = [[NSDateComponents alloc] init];
+        comps.day = 2;
+
+        // No matching options
+        NSDate* date = [calendar nextDateAfterDate:constantDate matchingComponents:comps options:0];
+
+        ASSERT_TRUE_MSG(false, "Should not have gotten here. nextDateAfterDate should throw an INVALID_ARG exception");
+    } @catch (NSException* e) {
+        ASSERT_OBJCEQ(e.name, @"NSInvalidArgumentException");
```
DHowett-MSFT: You could make this test work on ARM with ASSERT_ANY_THROW. That would remove the need for both of these and the explicit catch.

DHowett-MSFT: If you're going to check the name of the exception, however, please use the constant `NSInvalidArgumentException`.

---
In reply to: [74656977](https://github.com/Microsoft/WinObjC/pull/844#discussion_r74656977) [](ancestors = 74656977)

tests/unittests/Foundation/NSCalendarTests.mm
```diff
@@ -481,15 +481,63 @@
...
+    // is the same.
+
+    NSCalendar* calendar = [NSCalendar calendarWithIdentifier:NSCalendarIdentifierGregorian];
+    NSDate* date1 = [NSDate dateWithTimeIntervalSinceReferenceDate:.1];
+    NSDate* date2 = [NSDate dateWithTimeIntervalSinceReferenceDate:.9];
+    ASSERT_EQ([calendar compareDate:date1 toDate:date2 toUnitGranularity:NSCalendarUnitNanosecond], NSOrderedAscending);
+    ASSERT_EQ([calendar compareDate:date2 toDate:date1 toUnitGranularity:NSCalendarUnitNanosecond], NSOrderedDescending);
+}
+
+DISABLED_TEST(NSCalendar, rangeOfWeekOfMonthTest) {
```
DHowett-MSFT: >DISABLED_TEST [](start = 0, length = 13)

Does this test work on OS X?

DHowett-MSFT:
:shipit:

tests/unittests/Foundation/NSCalendarTests.mm
MSFTFox: Yes, I'll enable on OSX.

---
In reply to: [74657115](https://github.com/Microsoft/WinObjC/pull/844#discussion_r74657115) [](ancestors = 74657115)

ms-jihua:
:shipit:

PR #850, opened 2016-08-08
Fix NSJSONSerialization JSONObjectWithData tests to only verify error code  and not compare description string

mnithish:


mnithish:

mnithish:

DHowett-MSFT:
:shipit:

tests/unittests/Foundation/NSJSONSerializationTests.m
```diff
@@ -106,12 +97,11 @@ void VerifyDataWithJSONObjectThrows(id testObject) {
     VerifyJSONObjectWithDataSucceeds(testString6, 1, @{ @"foo" : @"1", @"bar" : @"2" });
 
     // error cases
-    NSString* fragmentDescription = @"JSON text did not start with array or object and option to allow fragments not set.";
```
mnithish: >@"JSON text did not start with array or object and option to allow fragments not set." [](start = 36, length = 86)

Reference platform does not document these strings and it seems like an over kill to compare error description in UT. Error codes should be sufficient

Note: An error code can return different descriptive descriptions, so error code to description is not an 1:1 mapping always

rajsesh-msft:
:shipit:

PR #851, opened 2016-08-08
NSDictionary allObjects has a leak

bbowman:
should be autoreleased.

### DIFF
--- a/Frameworks/Foundation/NSDictionary.mm
+++ b/Frameworks/Foundation/NSDictionary.mm
@@ -101,7 +101,7 @@ static int _NSDict_SortedKeysHelper(id key1, id key2, void* context) {
 
     NSMutableArray* toRet = [NSMutableArray array];
 
-    for (id key in _keyEnumerator) {
+    for (id key in static_cast<NSEnumerator*>(_keyEnumerator)) {
         [toRet addObject:[_dictionary objectForKey:key]];
     }
 
--- a/Frameworks/Foundation/NSDictionary.mm
+++ b/Frameworks/Foundation/NSDictionary.mm
@@ -99,8 +99,11 @@ static int _NSDict_SortedKeysHelper(id key1, id key2, void* context) {
         return nil;
     }
 
-    StrongId<NSArray> keyArray = [_keyEnumerator allObjects];
-    StrongId<NSMutableArray> toRet = [NSMutableArray new];
+    StrongId<NSArray> keyArray;
+    keyArray.attach([_keyEnumerator allObjects]);
+
+    StrongId<NSMutableArray> toRet;
+    toRet.attach([NSMutableArray new]);
 
     for (id key in static_cast<NSArray*>(keyArray)) {
         [toRet addObject:[_dictionary objectForKey:key]];
@@ -718,11 +721,12 @@ BASE_CLASS_REQUIRED_IMPLS(NSDictionary, NSDictionaryPrototype, CFDictionaryGetTy
 */
 - (NSArray*)keysSortedByValueWithOptions:(NSSortOptions)opts usingComparator:(NSComparator)cmptr {
     NSMutableArray* ret = [[[self allKeys] mutableCopy] autorelease];
-    [ret sortWithOptions:opts usingComparator:^NSComparisonResult(id key1, id key2) {
-        id val1 = [self objectForKey:key1];
-        id val2 = [self objectForKey:key2];
-        return cmptr(val1, val2);
-    }];
+    [ret sortWithOptions:opts
+         usingComparator:^NSComparisonResult(id key1, id key2) {
+             id val1 = [self objectForKey:key1];
+             id val2 = [self objectForKey:key2];
+             return cmptr(val1, val2);
+         }];
 
     return ret;
 }
@@ -799,10 +803,12 @@ BASE_CLASS_REQUIRED_IMPLS(NSDictionary, NSDictionaryPrototype, CFDictionaryGetTy
         return;
     }
 
-    _enumerateWithBlock([self keyEnumerator], options, ^(id key, BOOL* stop) {
-        id value = [self objectForKey:key];
-        block(key, value, stop);
-    });
+    _enumerateWithBlock([self keyEnumerator],
+                        options,
+                        ^(id key, BOOL* stop) {
+                            id value = [self objectForKey:key];
+                            block(key, value, stop);
+                        });
 }
 
 /**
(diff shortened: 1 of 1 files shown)


bbowman:

bbowman:

Frameworks/Foundation/NSDictionary.mm
```diff
@@ -99,8 +99,11 @@ - (NSArray*)allObjects {
         return nil;
     }
 
-    StrongId<NSArray> keyArray = [_keyEnumerator allObjects];
-    StrongId<NSMutableArray> toRet = [NSMutableArray new];
+    StrongId<NSArray> keyArray;
```
msft-Jeyaram: >NSArray [](start = 13, length = 7)

NSArray*?

Frameworks/Foundation/NSDictionary.mm
```diff
@@ -99,8 +99,11 @@ - (NSArray*)allObjects {
...
+    keyArray.attach([_keyEnumerator allObjects]);
+
+    StrongId<NSMutableArray> toRet;
```
msft-Jeyaram: >NSMutableArray [](start = 13, length = 14)

should be NSMutableArray*?

msft-Jeyaram: >toRet [](start = 29, length = 5)

is this really need to be wrapped in StrongId? given it's already auto released (as per [NSMutableArray new]) and it's being returned.

msft-Jeyaram:
:clock1:

Frameworks/Foundation/NSDictionary.mm
```diff
@@ -99,8 +99,11 @@ - (NSArray*)allObjects {
...
+    toRet.attach([NSMutableArray new]);
 
     for (id key in static_cast<NSArray*>(keyArray)) {
```
msft-Jeyaram: >static_cast<NSArray*> [](start = 19, length = 21)

can be removed if the above changes are made.

Frameworks/Foundation/NSDictionary.mm
bbowman: yes. Its a good habit.  Also [Waffle new] is *not* autoreleased.

---
In reply to: [73973293](https://github.com/Microsoft/WinObjC/pull/851#discussion_r73973293) [](ancestors = 73973293)

msft-Jeyaram: ops. [NSMutableArray array] should be autoreleased.

---
In reply to: [73973752](https://github.com/Microsoft/WinObjC/pull/851#discussion_r73973752) [](ancestors = 73973752,73973293)

bbowman: No it shouldn't. Please see the implementation of StrongId. essentially all smart ptr types remove the * for convenience. 

---
In reply to: [73973051](https://github.com/Microsoft/WinObjC/pull/851#discussion_r73973051) [](ancestors = 73973051)

bbowman: true that would be autoreleased but you are adding complication to the object lifetime by doing an autorelease on copy rule object. There is no need to autorelease and its not necessarily clear from the callee perspective when an autorelease will actually take place which makes the object lifetime longer than needed and may in fact cause some issues with resource de-init (see Java's or other garabage collected language's issues around closing files deterministically etc.)

---
In reply to: [73973850](https://github.com/Microsoft/WinObjC/pull/851#discussion_r73973850) [](ancestors = 73973850,73973752,73973293)

msft-Jeyaram: NSMutableArray* toRet =[NSMutableArray array];
..
return toRet; should make things more clear?

Having StrongId is also fine.

---
In reply to: [73973850](https://github.com/Microsoft/WinObjC/pull/851#discussion_r73973850) [](ancestors = 73973850,73973752,73973293)

msft-Jeyaram: True. I went with the assumption of creator ownership as per ios documentation, given this class creates it and it's not the known (copy,alloc) methods. it's naturally autoreleased. 

---
In reply to: [73974242](https://github.com/Microsoft/WinObjC/pull/851#discussion_r73974242) [](ancestors = 73974242,73973850,73973752,73973293)

bbowman: apologies. I mis-remembered what the copy attribute is signifying. Its copy-on-set not copy-on-get. I detest languages that use conventions as the primary mechanism to describe ownership.

---
In reply to: [73974316](https://github.com/Microsoft/WinObjC/pull/851#discussion_r73974316) [](ancestors = 73974316,73973850,73973752,73973293)

Frameworks/Foundation/NSDictionary.mm
```diff
@@ -99,8 +99,11 @@ - (NSArray*)allObjects {
         return nil;
     }
 
-    StrongId<NSArray> keyArray = [_keyEnumerator allObjects];
-    StrongId<NSMutableArray> toRet = [NSMutableArray new];
+    StrongId<NSArray> keyArray;
+    keyArray.attach([_keyEnumerator allObjects]);
```
DHowett-MSFT: >keyArray.attach([_keyEnumerator allObjects]) [](start = 4, length = 44)

This will almost certainly explode later; allObjects returns an autoreleased and attach does not add a +1.

DHowett-MSFT: StrongId isn't the right choice for these, since autorelease takes care of it delightfully well.

---
In reply to: [73979408](https://github.com/Microsoft/WinObjC/pull/851#discussion_r73979408) [](ancestors = 73979408)

DHowett-MSFT:
:clock1:

DHowett-MSFT:
:shipit:

rajsesh-msft:
:shipit:

Frameworks/Foundation/NSDictionary.mm
```diff
@@ -99,14 +99,14 @@ - (NSArray*)allObjects {
         return nil;
     }
 
-    StrongId<NSArray> keyArray = [_keyEnumerator allObjects];
-    StrongId<NSMutableArray> toRet = [NSMutableArray new];
+    NSArray* keyArray = [_keyEnumerator allObjects];
+    NSMutableArray* toRet = [NSMutableArray array];
 
-    for (id key in static_cast<NSArray*>(keyArray)) {
+    for (id key in keyArray) {
```
DHowett: Actually, there's a pointless copy here!
NSEnumerator can be fast-enumerated using `for(...in...)` syntax just as well as an array can.

We can kill the temporary, the retain on all the keys, and all their various friends by using `for (id key in _keyEnumerator)`.

Issue #689, opened 2016-07-27
EntryPoint.cpp needs a scopeguard


ms-jihua:
tests/unittests/EntryPoint.cpp::30
// TODO: move to a scopeguard when we have one in our codebase

The initialize and uninitialize should be done in a scopeguard. This would also reduce the size of the ifdef introduced in https://github.com/Microsoft/WinObjC/pull/634

rajsesh-msft:
Not super duper important. Won't fix.

PR #855, opened 2016-08-09
Move EntryPoint CoUninitialize to be under a scope guard. #689

msft-Jeyaram:


msft-Jeyaram:
Ping!

DHowett-MSFT:
Can we use RoInitializeWrapper or some other existing COM/WRL scope guard here instead of writing our own?

msft-Jeyaram:
Going to take this back and send a new one.

PR #740, opened 2016-08-10
NSBundle tests fail on OSX (WinObjC implementation details?)

bbowman:
Looks like we were testing for something WinObjC-exclusive? If so, should move to WindowsOnly tests:

[ RUN      ] NSBundle.ClassNamed
Foundation/NSBundleTests.mm:35: Failure
Value of: [[NSBundle mainBundle] classNamed:@"NSString"]
  Actual: NULL
Expected: [NSString class]
Which is: NSString
[  FAILED  ] NSBundle.ClassNamed (1 ms)

[ RUN      ] NSBundle.Localizations
Foundation/ReferenceFoundation/TestNSBundle.mm:89: Failure
Value of: bundle.localizations
  Actual: (
)
Expected: @[ @"en" ]
Which is: (
    en
)
[  FAILED  ] NSBundle.Localizations (0 ms)

### DIFF
--- a/Frameworks/Logging/LoggingNative.cpp
+++ b/Frameworks/Logging/LoggingNative.cpp
@@ -33,10 +33,10 @@ std::mutex s_isRegisteredMutex;
 
 // TODO: WIL logging hook
 // This is where we store the WIL hook
-//namespace wil {
-//namespace details {
+// namespace wil {
+// namespace details {
 //
-//void(__stdcall* g_pfnLoggingCallback)(wil::FailureInfo const& failure) WI_NOEXCEPT;
+// void(__stdcall* g_pfnLoggingCallback)(wil::FailureInfo const& failure) WI_NOEXCEPT;
 //
 //}
 //}
@@ -77,7 +77,7 @@ void TraceCritical(const wchar_t* tag, const wchar_t* format, ...) {
 }
 
 // TODO: WIL logging hook
-//void __stdcall _wilLoggingCallback(wil::FailureInfo const& failure) {
+// void __stdcall _wilLoggingCallback(wil::FailureInfo const& failure) {
 //    wchar_t debugString[2048];
 //    wil::GetFailureLogString(debugString, _countof(debugString), failure);
 //    TraceVerbose(L"WIL", debugString);
@@ -90,7 +90,7 @@ void TraceRegister() {
     }
 
     // TODO: WIL logging hook
-    //if (wil::details::g_pfnLoggingCallback == nullptr) {
+    // if (wil::details::g_pfnLoggingCallback == nullptr) {
     //    wil::details::g_pfnLoggingCallback = &_wilLoggingCallback;
     //}
 }
@@ -102,7 +102,7 @@ void TraceUnregister() {
     }
 
     // TODO: WIL logging hook
-    //if (wil::details::g_pfnLoggingCallback == &_wilLoggingCallback) {
+    // if (wil::details::g_pfnLoggingCallback == &_wilLoggingCallback) {
     //    wil::details::g_pfnLoggingCallback = nullptr;
     //}
 }
--- a/tests/frameworks/OSXShims/src/LoggingNative.mm
+++ b/tests/frameworks/OSXShims/src/LoggingNative.mm
@@ -15,12 +15,36 @@
 //******************************************************************************
 
 #import <LoggingNative.h>
+#import <Starboard/SmartTypes.h>
 #import <Foundation/NSString.h>
 
 #import <cstdarg>
 
+// NOTE: this function is assumed to only run in Little Endian environments like OSX
+static NSStringEncoding _getWcharEncoding() {
+    switch (sizeof(wchar_t)) {
+        case 2:
+            return NSUTF16LittleEndianStringEncoding;
+        case 4:
+            return NSUTF32LittleEndianStringEncoding;
+        default:
+            assert(false);
+    }
+}
+
+static NSStringEncoding getWcharEncoding() {
+    static NSStringEncoding encoding = _getWcharEncoding();
+    return encoding;
+}
+
 void TraceVerbose(const wchar_t* tag, const wchar_t* format, ...) {
-    NSString* formatString = [NSString stringWithFormat:@"[%ls] %ls", tag, format];
+    StrongId<NSString> nsTag;
+    nsTag.attach([[NSString alloc] initWithBytes:tag length:sizeof(wchar_t) * wcslen(tag) encoding:getWcharEncoding()]);
+
+    StrongId<NSString> nsFormat;
+    nsFormat.attach([[NSString alloc] initWithBytes:format length:sizeof(wchar_t) * wcslen(format) encoding:getWcharEncoding()]);
+
+    NSString* formatString = [NSString stringWithFormat:@"[%@] %@", static_cast<NSString*>(nsTag), static_cast<NSString*>(nsFormat)];
     va_list varArgs;
     va_start(varArgs, format);
     NSLogv(formatString, varArgs);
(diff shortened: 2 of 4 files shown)


bbowman:

bbowman:

Frameworks/Logging/LoggingNative.cpp
```diff
@@ -33,10 +33,10 @@ std::mutex s_isRegisteredMutex;
 
 // TODO: WIL logging hook
 // This is where we store the WIL hook
-//namespace wil {
```
ms-jihua: Looks like just formatting changes?

tests/frameworks/OSXShims/src/LoggingNative.mm
```diff
@@ -15,44 +15,92 @@
...
+        case 2:
+            return NSUTF16LittleEndianStringEncoding;
+        case 4:
+            return NSUTF32LittleEndianStringEncoding;
+        default:
+            assert(false);
+    }
+}
+
+static NSStringEncoding getWcharEncoding() {
```
ms-jihua: Hm - should we unify on one solution for this on OSX (should we remove my #define from the shim windows.h)?

tests/frameworks/OSXShims/src/LoggingNative.mm
```diff
@@ -15,44 +15,92 @@
...
+    static NSStringEncoding encoding = _getWcharEncoding();
+    return encoding;
+}
+
 void TraceVerbose(const wchar_t* tag, const wchar_t* format, ...) {
-    NSString* formatString = [NSString stringWithFormat:@"[%ls] %ls", tag, format];
+    StrongId<NSString> nsTag;
```
ms-jihua: >StrongId<NSString> nsTag; [](start = 4, length = 25)

Use of StrongId seems excessive here, especially given that you cast it back to NSString* below.

tests/unittests/Foundation/ReferenceFoundation/TestNSBundle.mm
```diff
@@ -97,68 +98,75 @@
...
-    for (NSString* n in _bundleResourceNames) {
-        [[NSFileManager defaultManager] createFileAtPath:[[bundlePath stringByAppendingString:@"/"] stringByAppendingString:n]
+protected:
+    virtual void SetUp() {
+        NSArray* _bundleResourceNames = @[ @"hello.world", @"goodbye.world", @"swift.org" ];
+
+        // Make sure the directory is uniquely named
+        // WINOBJC: test environment is different, '/tmp/' is not a top-level directory
+        // NSString* tempDir =
+        //     [[@"/tmp/TestFoundation_Playground_" stringByAppendingString:[NSUUID UUID].UUIDString] stringByAppendingString:@"/"];
```
ms-jihua: >/tmp/TestFoundation_Playground_ [](start = 19, length = 31)

update this comment also?

tests/unittests/Foundation/ReferenceFoundation/TestNSBundle.mm
```diff
@@ -97,68 +98,75 @@
...
+        NSString* tempDir =
+            [[@"./tmp_TestFoundation_Playground_" stringByAppendingString:[NSUUID UUID].UUIDString] stringByAppendingString:@"/"];
```
ms-jihua: >./tmp_TestFoundation_Playground_ [](start = 16, length = 32)

if the intent is for cleanup to be reliable, does it make sense for us to remove /tmp/ from our gitignore?

tests/unittests/Foundation/NSBundleTests.mm
```diff
@@ -31,6 +31,13 @@
     ASSERT_EQ_MSG(nil, [NSBundle bundleWithPath:nil], @"Bundle with nil path somehow not nil");
 }
 
+// Add a new class that will definitely be part of the mainBundle for this UT executable.
+@interface Waffle : NSObject
+@property (nonatomic, assign) float bacon;
```
ms-jihua: >bacon [](start = 36, length = 5)

nit: does not go on waffles

ms-jihua:
:clock1:

Frameworks/Logging/LoggingNative.cpp
bbowman: yep.

---
In reply to: [74168773](https://github.com/Microsoft/WinObjC/pull/740#discussion_r74168773) [](ancestors = 74168773)

tests/frameworks/OSXShims/src/LoggingNative.mm
bbowman: maybe? I don't really care one way or another but I don't think many other devs are going to stumble on WCHAR_ENCODING. I also don't think there are big endian OSX environments? We can chat in person about what we want.

---
In reply to: [74168895](https://github.com/Microsoft/WinObjC/pull/740#discussion_r74168895) [](ancestors = 74168895)

tests/frameworks/OSXShims/src/LoggingNative.mm
bbowman: not fixing. 

---
In reply to: [74169017](https://github.com/Microsoft/WinObjC/pull/740#discussion_r74169017) [](ancestors = 74169017)

tests/unittests/Foundation/ReferenceFoundation/TestNSBundle.mm
bbowman: no. the comment is the old code that was commented out. 

---
In reply to: [74169117](https://github.com/Microsoft/WinObjC/pull/740#discussion_r74169117) [](ancestors = 74169117)

tests/unittests/Foundation/ReferenceFoundation/TestNSBundle.mm
bbowman: probably not. Someone may still have that laying around.

---
In reply to: [74169170](https://github.com/Microsoft/WinObjC/pull/740#discussion_r74169170) [](ancestors = 74169170)

tests/unittests/Foundation/NSBundleTests.mm
bbowman: lol

---
In reply to: [74169190](https://github.com/Microsoft/WinObjC/pull/740#discussion_r74169190) [](ancestors = 74169190)

bbowman:
@ms-jihua addressed most feedback. Is there anything besides the wchar thing that is blocking this going in for you?

tests/frameworks/OSXShims/src/LoggingNative.mm
ms-jihua: justification? it would be half as long.

---
In reply to: [74170322](https://github.com/Microsoft/WinObjC/pull/740#discussion_r74170322) [](ancestors = 74170322,74169017)

ms-jihua:
I'm still of the opinion that you should take the StrongIds out, but besides that and the wchar thing, no.

---
In reply to: [238736424](https://github.com/Microsoft/WinObjC/pull/740#issuecomment-238736424) [](ancestors = 238736424)

tests/frameworks/OSXShims/src/LoggingNative.mm
bbowman: This is a file that will get compiled into all cross plat test binaries if I understand the build you set up ... some of those may have ARC on. This suggests we should use a mechanism that easily handles arc vs not which strong id does. Additionally it will be the same length as the alternative requires release statements in place of the attaches. The static cast is meh to me. I wouldn't mind a .get() method on a strong id so its more concise. 

---
In reply to: [74295731](https://github.com/Microsoft/WinObjC/pull/740#discussion_r74295731) [](ancestors = 74295731,74170322,74169017)

ms-jihua: okay, accepted

---
In reply to: [74333836](https://github.com/Microsoft/WinObjC/pull/740#discussion_r74333836) [](ancestors = 74333836,74295731,74170322,74169017)

ms-jihua:
:shipit:

tests/frameworks/OSXShims/src/LoggingNative.mm
DHowett-MSFT: @bbowman ARC is per-file, though, not per-module. Additionally, `.attach` is invalid in ARC mode, since `StrongId` resolves to `__strong id`.

Our options here are minimal; either we use `stringWithCString:encoding:` which does not allow us to specify length, or we use `autorelease` and make 100% certain that this file is built with ARC off. My vote is for the second, since it makes it significantly easier to reason about what's happening.

Issue #801, opened 2016-08-03
Implement NS*Collection* methods to filter using predicates for middleware


rajsesh-msft:
NSMutableSet filterUsingPredicate and NSSet filteredSetUsingPredicate.

PR #857, opened 2016-08-10
Supporting fliterbyPredicate for NSSet and NSMutableSet. #801

msft-Jeyaram:


tests/unittests/Foundation/NSSetTests.mm
```diff
@@ -60,4 +60,24 @@
...
+}
+
+TEST(NSSet, FilterUsingPredicateString) {
+    NSSet* americanMakes = [NSSet setWithObjects:@"Chrysler", @"Ford", @"General Motors", nil];
+    NSPredicate* predicate = [NSPredicate predicateWithFormat:@"SELF BEGINSWITH[c] %@", @"G"];
+
+    NSSet* filteredSet = [americanMakes filteredSetUsingPredicate:predicate];
+
+    ASSERT_NE(nil, filteredSet);
+    ASSERT_EQ(1, [filteredSet count]);
```
aballway: >ASSERT_EQ(1, [filteredSet count]) [](start = 4, length = 33)

Should also test to make sure it contains the correct elements

msft-Jeyaram: This should be sufficient given the element is not mutable. 

---
In reply to: [74267425](https://github.com/Microsoft/WinObjC/pull/857#discussion_r74267425) [](ancestors = 74267425)

msft-Jeyaram: but it would be a nice to have. adding.

---
In reply to: [74289845](https://github.com/Microsoft/WinObjC/pull/857#discussion_r74289845) [](ancestors = 74289845,74267425)

Frameworks/Foundation/NSMutableSet.mm
```diff
@@ -139,11 +139,21 @@ - (void)removeObject:(id)object {
 }
 
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 - (void)filterUsingPredicate:(NSPredicate*)predicate {
-    UNIMPLEMENTED();
+    if (predicate == NULL) {
```
msft-Jeyaram: >NULL [](start = 21, length = 4)

nit: nil

tests/unittests/Foundation/NSSetTests.mm
```diff
@@ -60,4 +60,24 @@
...
+}
+
+TEST(NSSet, FilterUsingPredicateNumbers) {
+    NSSet* americanMakes = [NSSet setWithObjects:@2, @4, @8, @16, @32, nil];
```
msft-Jeyaram: >americanMakes [](start = 11, length = 13)

lol rename.

aballway:
:clock1:

Frameworks/Foundation/NSMutableSet.mm
```diff
@@ -139,11 +139,21 @@ - (void)removeObject:(id)object {
...
+        return;
+    }
+    @synchronized(self) {
+        NSArray* objects = [self allObjects];
```
DHowett-MSFT: Is this API guaranteed to be thread-safe?
If not, there's no reason to `@synchronized` here. None of the other APIs do.

Frameworks/Foundation/NSMutableSet.mm
```diff
@@ -139,11 +139,21 @@ - (void)removeObject:(id)object {
...
+
+        for (id obj in objects) {
+            if (![predicate evaluateWithObject:obj]) {
+                [self removeObject:obj];
```
DHowett-MSFT: >self removeObject:obj]; [](start = 17, length = 23)

It is not legal to mutate a collection while enumerating it.
Please build up a second set of removals and then use `-minusSet:` with it.

tests/unittests/Foundation/NSSetTests.mm
```diff
@@ -60,4 +60,24 @@
...
+    NSPredicate* predicate = [NSPredicate predicateWithFormat:@"(SELF >= 4) AND (SELF < 32)"];
```
DHowett-MSFT: >eWithFormat:@"(SELF >= 4) AND (SELF < 32)"]; [](start = 50, length = 44)

please test with an always-false predicate (removes all objects) and an always-true predicate.

DHowett-MSFT:
:clock1:

Frameworks/Foundation/NSMutableSet.mm
msft-Jeyaram: As mentioned we aren't going through the set. I am using the allobjects which is the snapshot (copy)

---
In reply to: [74296016](https://github.com/Microsoft/WinObjC/pull/857#discussion_r74296016) [](ancestors = 74296016)

DHowett-MSFT: Nevermind: you're using `-allObjects`.

---
In reply to: [74296016](https://github.com/Microsoft/WinObjC/pull/857#discussion_r74296016) [](ancestors = 74296016)

Frameworks/Foundation/NSMutableSet.mm
msft-Jeyaram: Nope it's not. i'll go ahead with the removal. I thought i can guard the copy->removal phase

---
In reply to: [74295924](https://github.com/Microsoft/WinObjC/pull/857#discussion_r74295924) [](ancestors = 74295924)

Issue #859, opened 2016-08-10
Need PNG/JPG texture rendering in OpenGL test app


rajsesh-msft:
This would have caught the regression #847 before the final validation.

rajsesh-msft:
This was primarily to catch pixel format issues. #1220 has unit tests that do this much more effectively, so this is not useful anymore.

PR #860, opened 2016-08-10
Fix surface format for RGBA surfaces

### LINKED ISSUE #847: The new Colorspace implementation doesn't quite support RGBA, and that breaks jpg and png rendering
Both our JPG and PNG decoders set surface format as RGBA.  This breaks some validation apps (they use cocos2d), where the colors of the PNG and JPG images are off.

rajsesh-msft:
The correct surface format for Big Endian + Alpha Last is _ColorRGBA. We were incorrectly setting it to ARGB, resulting in apps that use this combination to get blue and red swapped and broke some validation apps.

To uncover future regressions, also modified one of the images to be PNG in WOCCatalog.  We still need additional validation through OpenGL samples.  Opened Issue #859 to track this.

Fixes #847

### DIFF
--- a/Frameworks/include/CGSurfaceInfoInternal.h
+++ b/Frameworks/include/CGSurfaceInfoInternal.h
@@ -88,10 +88,12 @@ static inline __CGSurfaceFormat _CGImageGetFormat(unsigned int bitsPerComponent,
                         fmt = _ColorARGB;
                     }
                     break;
-                case kCGImageAlphaNoneSkipLast:
                 case kCGImageAlphaLast:
                 case kCGImageAlphaPremultipliedLast:
                     fmt = _ColorRGBA;
+                    break;
+                case kCGImageAlphaNoneSkipLast:
+                    fmt = _ColorXBGR;
                     break;
                 default:
                     UNIMPLEMENTED_WITH_MSG("Unknown pixel format, assuming ARGB");
--- a/Frameworks/include/CGSurfaceInfoInternal.h
+++ b/Frameworks/include/CGSurfaceInfoInternal.h
@@ -91,8 +91,10 @@ static inline __CGSurfaceFormat _CGImageGetFormat(unsigned int bitsPerComponent,
                 case kCGImageAlphaNoneSkipLast:
                 case kCGImageAlphaLast:
                 case kCGImageAlphaPremultipliedLast:
+                    fmt = _ColorRGBA;
+                    break;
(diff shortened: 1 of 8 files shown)


rajsesh-msft:
@shahmeeresmail #Closed

jaredhms:
>                         fmt = _ColorARGB;

are we able to verify the rest of these?  it seems like we should have a UT for _CGImageGetFormat for the various supported/unsupported formats.

---
Refers to: Frameworks/include/CGSurfaceInfoInternal.h:80 in a64e395. [](commit_id = a64e395cdb36e55afb50b118cffd8139dfdb7560, deletion_comment = False)

Frameworks/include/CGSurfaceInfoInternal.h
```diff
@@ -91,8 +91,10 @@ static inline __CGSurfaceFormat _CGImageGetFormat(unsigned int bitsPerComponent,
                 case kCGImageAlphaNoneSkipLast:
```
rajsesh-msft: Should be RGBX.

shahmeeresmail:
How is this hit in the current code path? I searched through the code base and even with the new changes, I could not find an instance where we set both kCGBitmapByteOrder32Big and kCGImageAlphaNoneSkipLast simultaneously as the bitmapInfo for an object. Perhaps it is set somewhere as a magic number?

rajsesh-msft:
This was found in some internal apps we use for validation.  The apps use cocos2d.  The app creates CGBitmapContext with these surface spec, draws into it, and uses the bytes through openGL.  

---
In reply to: [238941568](https://github.com/Microsoft/WinObjC/pull/860#issuecomment-238941568) [](ancestors = 238941568)

rajsesh-msft:
>                         fmt = _ColorARGB;

No.  validating this function is of no real use.  What we need really are true e2e validation for different apis that go through this.  #859 should really track it.

---
In reply to: [238881925](https://github.com/Microsoft/WinObjC/pull/860#issuecomment-238881925) [](ancestors = 238881925)

---
Refers to: Frameworks/include/CGSurfaceInfoInternal.h:80 in a64e395. [](commit_id = a64e395cdb36e55afb50b118cffd8139dfdb7560, deletion_comment = False)

jaredhms:
:shipit:

mnithish:
:shipit:

bbowman:
:shipit:

PR #866, opened 2016-08-10
Undef Windows UMD macros that conflict with iOS

mnithish:


mnithish:
@memontic-ms is added to the review. #Closed

mnithish:

mnithish:

include/WOCStdLib/winnt.h
```diff
@@ -0,0 +1,23 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#pragma once
+
+#include_next <winnt.h>
+#ifdef __OBJC__
+#undef DELETE
```
memontic-ms: Check #ifdef DELETE beforehand

memontic-ms:
:shipit:

include/WOCStdLib/winnt.h
mnithish: Good point. Done

---
In reply to: [74333399](https://github.com/Microsoft/WinObjC/pull/866#discussion_r74333399) [](ancestors = 74333399)

DHowett-MSFT:
:shipit:

include/WOCStdLib/winnt.h
bbowman: the "new hotness" for this is to use a pragma push which will push the old def onto to macro stack. You can then undef / redef things to your heart's content and pop at the end. This a convenient way to make sure the window's header gets what it wants but doesn't corrupt state down the line without the big hammer of killing anyone else's DELETE macro ... 

---
In reply to: [74333570](https://github.com/Microsoft/WinObjC/pull/866#discussion_r74333570) [](ancestors = 74333570,74333399)

bbowman:
:clock1:

include/WOCStdLib/winnt.h
mnithish: Good to know. Will change to that Brian

---
In reply to: [74335367](https://github.com/Microsoft/WinObjC/pull/866#discussion_r74335367) [](ancestors = 74335367,74333570,74333399)

PR #868, opened 2016-08-10
Align NSDateComponents selectors with OSX 

aballway:
NSDateComponents does not have a public "week" property, but it does have deprecated week and setWeek: methods, which do not interact with other values in NSDateComponents.

### DIFF
--- a/include/Foundation/NSDateComponents.h
+++ b/include/Foundation/NSDateComponents.h
@@ -38,7 +38,6 @@ FOUNDATION_EXPORT_CLASS
 @property NSInteger minute;
 @property NSInteger second;
 @property NSInteger nanosecond;
-@property NSInteger week;
 @property NSInteger weekday;
 @property NSInteger weekdayOrdinal;
 @property NSInteger quarter;
@@ -48,4 +47,6 @@ FOUNDATION_EXPORT_CLASS
 @property (getter=isLeapMonth) BOOL leapMonth;
 - (NSInteger)valueForComponent:(NSCalendarUnit)unit;
 - (void)setValue:(NSInteger)value forComponent:(NSCalendarUnit)unit;
+- (NSInteger)week;
+- (void)setWeek:(NSInteger)v;
 @end
--- a/tests/unittests/Foundation/NSDateComponentsTests.m
+++ b/tests/unittests/Foundation/NSDateComponentsTests.m
@@ -75,15 +75,15 @@ TEST(NSDateComponents, IndependentWeekTest) {
     dateComponents.day = 1;
     dateComponents.month = 1;
     dateComponents.year = 1999;
-    EXPECT_EQ_MSG([dateComponents week], LONG_MAX, "FAILED: week default value should be LONG_MAX");
+    EXPECT_EQ_MSG([dateComponents week], NSUndefinedDateComponent, "FAILED: week default value should be NSUndefinedDateComponent");
 
     dateComponents.calendar = [[NSCalendar alloc] initWithCalendarIdentifier:NSCalendarIdentifierGregorian];
-    EXPECT_EQ_MSG([dateComponents week], LONG_MAX, "FAILED: week value should not change because of calendar");
+    EXPECT_EQ_MSG([dateComponents week], NSUndefinedDateComponent, "FAILED: week value should not change because of calendar");
 
     dateComponents.weekOfMonth = 1;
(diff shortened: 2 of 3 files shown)


rajsesh-msft:
:shipit:

DHowett-MSFT:
Having a week property is the same as having `setWeek:` and `week`. Perhaps we just need an explicit `@synthesize week = _week;` at the top of the implementation?

DHowett-MSFT:
I worry somewhat that legacy code will be relying on `week` to actually do something; I don't think that it was always deprecated, and I don't think it always did nothing.
Are there facets that consumer NSDateComponents that could rely on the week property?

rajsesh-msft:
This is merely to not be flagged by our analysis tools as a mismatch.  week never worked on our platform, there is no expectation that it should work at this point.  Given that it *is* deprecated we could consider it if someone really needs it.

---
In reply to: [239017245](https://github.com/Microsoft/WinObjC/pull/868#issuecomment-239017245) [](ancestors = 239017245)

DHowett-MSFT:
:shipit:

tests/unittests/Foundation/NSDateComponentsTests.m
```diff
@@ -67,5 +67,24 @@
     NSDate* cstDate = [calendar dateFromComponents:dateComponents];
 
     NSTimeInterval secondsBetween = [cstDate timeIntervalSinceDate:estDate];
-    ASSERT_EQ_MSG(secondsBetween, 3600, "FAILED: dates should differ by one hour");
+    EXPECT_EQ_MSG(secondsBetween, 3600, "FAILED: dates should differ by one hour");
+}
+
+TEST(NSDateComponents, IndependentWeekTest) {
+    NSDateComponents* dateComponents = [[NSDateComponents alloc] init];
```
msft-Jeyaram: >dateComponents [](start = 22, length = 14)

needs to be released

msft-Jeyaram:
>     NSDateComponents* dateComponents = [[NSDateComponents alloc] init];

needs to be released.

---
Refers to: tests/unittests/Foundation/NSDateComponentsTests.m:26 in 8f72d34. [](commit_id = 8f72d3464940b334b6066f0654aaff1c7e82e2d3, deletion_comment = False)

msft-Jeyaram:
> #include "gtest-api.h"

nit: replace it by TestFramework.h

---
Refers to: tests/unittests/Foundation/NSDateComponentsTests.m:17 in 8f72d34. [](commit_id = 8f72d3464940b334b6066f0654aaff1c7e82e2d3, deletion_comment = False)

msft-Jeyaram:
:clock1:

include/Foundation/NSDateComponents.h
```diff
@@ -48,4 +47,6 @@ FOUNDATION_EXPORT_CLASS
 @property (getter=isLeapMonth) BOOL leapMonth;
 - (NSInteger)valueForComponent:(NSCalendarUnit)unit;
 - (void)setValue:(NSInteger)value forComponent:(NSCalendarUnit)unit;
+- (NSInteger)week;
+- (void)setWeek:(NSInteger)v;
```
MSFTFox: Hoenstly, I'm pretty confused by this. What benefit is there to remove the week property and replacing it with an ivar and some selectors? This does the same thing as having a property for week does it not?

MSFTFox:
That's not true, week works fine on our platform. I have some tests in NSCalendar verifying so, not as many as the other calendar units since it is deprecated however. If the concern is our analysis tools, a better solution may be to flag this in some way so the analyzer doesn't mark it as invalid.

---
In reply to: [239018515](https://github.com/Microsoft/WinObjC/pull/868#issuecomment-239018515) [](ancestors = 239018515,239017245)

include/Foundation/NSDateComponents.h
aballway: It's just to align it with the OSX class which does not have a property but the two very propertylike methods.  Though it's a bit silly, it makes more sense to align it with the reference platform rather than have it be different and either modify our analysis tools or flag the code for a deprecated method, especially since this change doesn't affect functionality at all.

---
In reply to: [74354479](https://github.com/Microsoft/WinObjC/pull/868#discussion_r74354479) [](ancestors = 74354479)

MSFTFox:
:shipit:

PR #870, opened 2016-08-10
implementing lastPathComponent for NSString and unit tests.

msft-Jeyaram:


tests/unittests/Foundation/NSStringTests.m
```diff
@@ -607,4 +607,59 @@ void testUrlCharacterSetEncoding(NSString* decodedString, NSString* encodedStrin
...
+    NSString* string = @"";
+    EXPECT_OBJCEQ(@"", string.lastPathComponent);
+
+    string = @"hello.world";
+    EXPECT_OBJCEQ(@"hello.world", string.lastPathComponent);
+
+    string = @"/";
+    EXPECT_OBJCEQ(@"/", string.lastPathComponent);
+
+#if TARGET_OS_WIN32
```
ms-jihua: nit: comment here about why this is Windows-only (I get that it's because of the drive char, but the comment still helps)

Frameworks/Foundation/NSString.mm
```diff
@@ -680,7 +680,25 @@ - (instancetype)pathExtension {
...
+
+    CFIndex index = _CFStartOfLastPathComponent2(static_cast<CFStringRef>(self));
+    NSString* substring = [self substringFromIndex:index];
+
+    if (IS_SLASH([self characterAtIndex:index])) {
+        if ([self length] == 1) {
+            return substring;
+        }
+
+        // only way we could of ended up here is, if we have a string with multiple leading slashes.
```
ms-jihua: >of [](start = 29, length = 2)

nit: have

Frameworks/Foundation/NSString.mm
```diff
@@ -680,7 +680,25 @@ - (instancetype)pathExtension {
...
+        NSArray<NSString*>* tokens = [self componentsSeparatedByString:_NSGetSlashStr()];
+        return ([tokens[0] isEqualToString:@""]) ? _NSGetSlashStr() : tokens[0];
```
ms-jihua: >tokens[0] [](start = 70, length = 9)

wouldn't this return the _first_ path component, then?

msft-Jeyaram: this is based of the substring. so yes.

/foo/bar///


---
In reply to: [74351965](https://github.com/Microsoft/WinObjC/pull/870#discussion_r74351965) [](ancestors = 74351965)

ms-jihua: how is it based on the substring? the split is being called on self

---
In reply to: [74352126](https://github.com/Microsoft/WinObjC/pull/870#discussion_r74352126) [](ancestors = 74352126,74351965)

msft-Jeyaram: i take that back, it's not based on the substring. if we have only following flashes, we have to use the actual string itself.

---
In reply to: [74352319](https://github.com/Microsoft/WinObjC/pull/870#discussion_r74352319) [](ancestors = 74352319,74352126,74351965)

ms-jihua: right, but if you ran this code on '/foo/bar/////', this would return 'foo' instead of 'bar'

---
In reply to: [74352571](https://github.com/Microsoft/WinObjC/pull/870#discussion_r74352571) [](ancestors = 74352571,74352319,74352126,74351965)

Frameworks/Foundation/NSString.mm
ms-jihua: >leading [](start = 84, length = 7)

I think you mean trailing?

Frameworks/Foundation/NSString.mm
msft-Jeyaram: true, at the moment it would actually end up returning a /. i'll fix this. lol string parsing :(

---
In reply to: [74353252](https://github.com/Microsoft/WinObjC/pull/870#discussion_r74353252) [](ancestors = 74353252,74352571,74352319,74352126,74351965)

Frameworks/Foundation/NSString.mm
```diff
@@ -680,7 +680,25 @@ - (instancetype)pathExtension {
...
+    }
+
+    NSArray<NSString*>* tokens = [substring componentsSeparatedByString:_NSGetSlashStr()];
+    return tokens[0];
```
ms-jihua: What case was this designed to handle?

Frameworks/Foundation/NSString.mm
ms-jihua: remember to add a test case covering it

---
In reply to: [74353599](https://github.com/Microsoft/WinObjC/pull/870#discussion_r74353599) [](ancestors = 74353599,74353252,74352571,74352319,74352126,74351965)

msft-Jeyaram: can't implement without test :p

---
In reply to: [74353658](https://github.com/Microsoft/WinObjC/pull/870#discussion_r74353658) [](ancestors = 74353658,74353599,74353252,74352571,74352319,74352126,74351965)

msft-Jeyaram: definitely 

---
In reply to: [74353703](https://github.com/Microsoft/WinObjC/pull/870#discussion_r74353703) [](ancestors = 74353703,74353658,74353599,74353252,74352571,74352319,74352126,74351965)

tests/unittests/Foundation/NSStringTests.m
```diff
@@ -607,4 +607,59 @@ void testUrlCharacterSetEncoding(NSString* decodedString, NSString* encodedStrin
...
+
+    string = @"/tmp/random.foo.tiff";
+    EXPECT_OBJCEQ(@"random.foo.tiff", string.lastPathComponent);
+
+    string = @"/temp//foo//bar.txt";
+    EXPECT_OBJCEQ(@"bar.txt", string.lastPathComponent);
+
+    string = @"/.";
+    EXPECT_OBJCEQ(@".", string.lastPathComponent);
+}
```
ms-jihua: I realize these probably work already but I would like to add a few test cases with

multiple leading slashes: ///foo/bar/test
uri schemes at the front: [file://foo/bar/test](file://foo/bar/test)
a dot name within a directory /tmp/foo/.tiff

Frameworks/Foundation/NSString.mm
```diff
@@ -680,7 +680,25 @@ - (instancetype)pathExtension {
  @Status Interoperable
 */
 - (instancetype)lastPathComponent {
-    return [self substringFromIndex:_CFStartOfLastPathComponent2(static_cast<CFStringRef>(self))];
+    if ([self length] == 0) {
+        return @"";
+    }
+
+    CFIndex index = _CFStartOfLastPathComponent2(static_cast<CFStringRef>(self));
```
msft-Jeyaram: >_CFStartOfLastPathComponent2 [](start = 20, length = 28)

_CFStartOfLastPathComponent2 will not be helpful in case of trailing slashes as this would end up pointing to the slash before the last.

tests/unittests/Foundation/NSStringTests.m
msft-Jeyaram: more the merrier 

---
In reply to: [74354034](https://github.com/Microsoft/WinObjC/pull/870#discussion_r74354034) [](ancestors = 74354034)

Frameworks/Foundation/NSString.mm
msft-Jeyaram: /temp/

---
In reply to: [74353642](https://github.com/Microsoft/WinObjC/pull/870#discussion_r74353642) [](ancestors = 74353642)

ms-jihua: what was the result when it reached this point?

---
In reply to: [74354338](https://github.com/Microsoft/WinObjC/pull/870#discussion_r74354338) [](ancestors = 74354338,74353642)

msft-Jeyaram: I wish the result from CF was good. then we wouldn't have to do much of this extra painful parsing. it seems it was assumed during implementation. 

---
In reply to: [74354338](https://github.com/Microsoft/WinObjC/pull/870#discussion_r74354338) [](ancestors = 74354338,74353642)

msft-Jeyaram: _CFStartOfLastPathComponent2 would return 1, meaning the result would be temp/

---
In reply to: [74354418](https://github.com/Microsoft/WinObjC/pull/870#discussion_r74354418) [](ancestors = 74354418,74354338,74353642)

ms-jihua: Okay - please add a comment that this step removes the trailing slashes still in the substring, with the example you just gave me :D

---
In reply to: [74354542](https://github.com/Microsoft/WinObjC/pull/870#discussion_r74354542) [](ancestors = 74354542,74354418,74354338,74353642)

tests/unittests/Foundation/NSStringTests.m
```diff
@@ -607,4 +607,60 @@ void testUrlCharacterSetEncoding(NSString* decodedString, NSString* encodedStrin
...
+    EXPECT_OBJCEQ(@"/", string.lastPathComponent);
+
+    string = @"/tmp/scratch..tiff";
+    EXPECT_OBJCEQ(@"scratch..tiff", string.lastPathComponent);
+
+    string = @"/tmp/random.foo.tiff";
+    EXPECT_OBJCEQ(@"random.foo.tiff", string.lastPathComponent);
+
+    string = @"/temp//foo//bar.txt";
+    EXPECT_OBJCEQ(@"bar.txt", string.lastPathComponent);
```
MSFTFox: >EXPECT_OBJCEQ [](start = 4, length = 13)

Any particular reason for EXPECT over ASSERT?

Also, if the justification is for EXPECT (Which I agree with) then could you change the other tests to use EXPECTS if it makes sense for them while you're in here?

msft-Jeyaram: we are testing multiple cases here and they are independent of each other.
so expect is appropriate. 

---
In reply to: [74357619](https://github.com/Microsoft/WinObjC/pull/870#discussion_r74357619) [](ancestors = 74357619)

MSFTFox: Yup, so should we then modify the other tests in this file to use EXPECT instead of ASSERT while you're updating this file?

---
In reply to: [74357771](https://github.com/Microsoft/WinObjC/pull/870#discussion_r74357771) [](ancestors = 74357771,74357619)

msft-Jeyaram: I will update it for the ones that seems independent. Although I'm sure we can go through each test case and see if they can be converted to EXPECT, but this is not the scope of this CR.

---
In reply to: [74357911](https://github.com/Microsoft/WinObjC/pull/870#discussion_r74357911) [](ancestors = 74357911,74357771,74357619)

bbowman: Also a good candidate for parameterized tests.

---
In reply to: [74358052](https://github.com/Microsoft/WinObjC/pull/870#discussion_r74358052) [](ancestors = 74358052,74357911,74357771,74357619)

Frameworks/Foundation/NSString.mm
```diff
@@ -680,7 +680,25 @@ - (instancetype)pathExtension {
...
+    NSString* substring = [self substringFromIndex:index];
+
+    if (IS_SLASH([self characterAtIndex:index])) {
+        if ([self length] == 1) {
+            return substring;
+        }
+
+        // only way we could have ended up here is, if we have a string with multiple leading slashes.
```
bbowman: > is, if [](start = 47, length = 7)

punctuation of this sentence is incorrect.

Frameworks/Foundation/NSString.mm
```diff
@@ -680,7 +680,25 @@ - (instancetype)pathExtension {
...
-    return [self substringFromIndex:_CFStartOfLastPathComponent2(static_cast<CFStringRef>(self))];
+    if ([self length] == 0) {
+        return @"";
+    }
+
+    CFIndex index = _CFStartOfLastPathComponent2(static_cast<CFStringRef>(self));
+    NSString* substring = [self substringFromIndex:index];
+
+    if (IS_SLASH([self characterAtIndex:index])) {
+        if ([self length] == 1) {
```
bbowman: >if ([self length] == 1) { [](start = 8, length = 25)

may be useful to comment what this case if for. "/" right?

Frameworks/Foundation/NSString.mm
```diff
@@ -680,7 +680,25 @@ - (instancetype)pathExtension {
...
+            return substring;
+        }
+
+        // only way we could have ended up here is, if we have a string with multiple leading slashes.
+        NSArray<NSString*>* tokens = [self componentsSeparatedByString:_NSGetSlashStr()];
+        return ([tokens[0] isEqualToString:@""]) ? _NSGetSlashStr() : tokens[0];
```
bbowman: >tokens[0] [](start = 17, length = 9)

blind index here seems risky? Any concerns for that / comments we need to reassure the reader that this is ok?

bbowman:
:clock1:

Frameworks/Foundation/NSString.mm
DHowett-MSFT: Comments would be good. `[@"x" componentsSeparatedByString:@"y"]` will return `@[@"x"]` if there's no `y`. We already made sure the length was safe above.

---
In reply to: [74492304](https://github.com/Microsoft/WinObjC/pull/870#discussion_r74492304) [](ancestors = 74492304)

rajsesh-msft:
This has been completed by @aballway through #934

PR #871, opened 2016-08-11
Revert "Force the ObjC 2-stage loader to be pinned & add /WHOLEARCHIV…

rajsesh-msft:
…E support (#684)"

This reverts commit b4f568cc41a69f43433b25a562303b7a4f6d4e25.


rajsesh-msft:
Not reverting the change, we will attempt to fix the regressions and push this next week.

PR #838, opened 2016-08-11
[Win32] Foundation.UnitTests project doesn't cause a dep rebuild on Foundation.dll

bbowman:
This causes library changes to not be picked up by the unit test build.

### DIFF
--- a/msvc/sbclang.targets
+++ b/msvc/sbclang.targets
@@ -486,6 +486,7 @@
     <ItemGroup Condition="'@(_StarboardReferencesRequiringObjectiveC)'!='' and '$(StarboardLinkWholeArchive)'=='true'">
       <Link>
         <AdditionalOptions>%(Link.AdditionalOptions) "/WHOLEARCHIVE:@(_StarboardReferencesRequiringObjectiveC, '" /WHOLEARCHIVE:"')"</AdditionalOptions>
+        <LinkIncremental>false</LinkIncremental>
       </Link>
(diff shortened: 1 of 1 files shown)


bbowman:

bbowman:
Would love it if you guys would try it out to confirm. This should do the trick though.

DHowett-MSFT:
Alternatively, #871.

rajsesh-msft:
This fixes part (but not all) of the issues uncovered in the /wholearchive change.  I'll try this out as well.

ms-jihua:
This does not fix the issue for me :( should I also be trying with #871 below?

---
In reply to: [239043624](https://github.com/Microsoft/WinObjC/pull/838#issuecomment-239043624) [](ancestors = 239043624)

ms-jihua:
We looked at the issue offline - for some reason this didn't propagate to my Foundation.dll properties? (/INCREMENTAL was still set to yes) bbowman's investigating

---
In reply to: [239045051](https://github.com/Microsoft/WinObjC/pull/838#issuecomment-239045051) [](ancestors = 239045051,239043624)

bbowman:
This is sort of a dupe of issue 871. We have seen other /WHOLEARCHIVE issues and will fix this issue when we fix that. Closing for now. @DHowett-MSFT or @rajsesh-msft  will make sure incremental linking works or is turned off when the other issues are sorted out.

PR #874, opened 2016-08-12
Fix a handful of issues in /WHOLEARCHIVE support.

DHowett-MSFT:
* There was a breakage in the ARM unit tests due to a missing dependency pulled in through /WHOLEARCHIVE.
  * That dependency is now included in the test CRT dependency set.
* The library path for our ARM unit tests was being set incorrectly.
  * In setting it correctly, we found that Starboard had a dependency on the concurrency runtime, which we cannot link. That dependency was found to be in error.
* Linking a project with both `/INCREMENTAL` and `/WHOLEARCHIVE` can lead to incomplete object merging.
* We would occasionally emit `/WHOLEARCHIVE` and `___pin_objc_init` for referenced libraries containing only **excluded** Objective-C files.

Local tests:
* Passed ARM Unit Tests (Debug/Release).
* Passed ARM application validation.
* Passed x86 Unit Tests (Debug/Release)

Remote tests:
* Currently running a full soak (ARM+x86) test.

### DIFF
--- a/msvc/ut-build-early.props
+++ b/msvc/ut-build-early.props
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup Label="Starboard OneCore Overrides">
+    <_SB_Local_VCInstallDir Condition="'$(_SB_Local_VCInstallDir)' == ''">$(Registry:HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\VisualStudio\14.0\Setup\VC@ProductDir)</_SB_Local_VCInstallDir>
+    <_SB_Local_VCInstallDir Condition="'$(_SB_Local_VCInstallDir)' == ''">$(Registry:HKEY_LOCAL_MACHINE\SOFTWARE\Wow6432Node\Microsoft\VisualStudio\14.0\Setup\VC@ProductDir)</_SB_Local_VCInstallDir>
+    <_SB_Local_VCInstallDir Condition="'$(_SB_Local_VCInstallDir)' == ''">$(Registry:HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\VSWinExpress\14.0\Setup\VC@ProductDir)</_SB_Local_VCInstallDir>
+    <_SB_Local_VCInstallDir Condition="'$(_SB_Local_VCInstallDir)' == ''">$(Registry:HKEY_LOCAL_MACHINE\SOFTWARE\Wow6432Node\Microsoft\VSWinExpress\14.0\Setup\VC@ProductDir)</_SB_Local_VCInstallDir>
+
+    <!-- We want to link the ARM unit test binaries against the store libraries without turning on WindowsAppContainer.
+         Therefore, we have to override the VC ARM library path early - before Microsoft.Cpp.Common.props is included.
+    -->
+    <StarboardUnitTestAPISet Condition="'$(StarboardUnitTestAPISet)'==''">onecore</StarboardUnitTestAPISet>
+    <VC_LibraryPath_VC_ARM>$(_SB_Local_VCInstallDir)lib\$(StarboardUnitTestAPISet)\ARM;</VC_LibraryPath_VC_ARM>
+  </PropertyGroup>
+</Project>
+
--- a/Frameworks/Starboard/pthread.cpp
+++ b/Frameworks/Starboard/pthread.cpp
@@ -24,7 +24,6 @@
 #include <cmath>
 #include <map>
 #include <mutex>
-#include <concrt.h>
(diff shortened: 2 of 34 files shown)


DHowett-MSFT:

DHowett-MSFT:

DHowett-MSFT:

DHowett-MSFT:

msvc/ut-build-early.props
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup Label="Starboard OneCore Overrides">
+    <_SB_Local_VCInstallDir Condition="'$(_SB_Local_VCInstallDir)' == ''">$(Registry:HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\VisualStudio\14.0\Setup\VC@ProductDir)</_SB_Local_VCInstallDir>
+    <_SB_Local_VCInstallDir Condition="'$(_SB_Local_VCInstallDir)' == ''">$(Registry:HKEY_LOCAL_MACHINE\SOFTWARE\Wow6432Node\Microsoft\VisualStudio\14.0\Setup\VC@ProductDir)</_SB_Local_VCInstallDir>
+    <_SB_Local_VCInstallDir Condition="'$(_SB_Local_VCInstallDir)' == ''">$(Registry:HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\VSWinExpress\14.0\Setup\VC@ProductDir)</_SB_Local_VCInstallDir>
+    <_SB_Local_VCInstallDir Condition="'$(_SB_Local_VCInstallDir)' == ''">$(Registry:HKEY_LOCAL_MACHINE\SOFTWARE\Wow6432Node\Microsoft\VSWinExpress\14.0\Setup\VC@ProductDir)</_SB_Local_VCInstallDir>
+
```
bbowman: no other build variables that already this stuff?

DHowett-MSFT:
One commit message needs to have its wording fixed during merge.

msvc/ut-build-early.props
DHowett-MSFT: Not this early, no. The file that sets these sets the others as well.

bbowman:
:shipit:

deps/3rdparty/libobjc2
```diff
@@ -1 +1 @@
-Subproject commit 7bafddf63787ef2a99bb7492bdb5cdb3834c79be
+Subproject commit 9c2455830e9bca9c33e5fb38bffc704414d44a6a
```
DHowett-MSFT: This will be removed.

rajsesh-msft:
:shipit:

msvc/ut-build-early.props
```diff
@@ -0,0 +1,16 @@
...
+    <!-- We want to link the ARM unit test binaries against the store libraries without turning on WindowsAppContainer.
```
pradipd: > store libraries  [](start = 63, length = 17)

I think you mean onecore libraries.

msvc/ut-build-early.props
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup Label="Starboard OneCore Overrides">
+    <_SB_Local_VCInstallDir Condition="'$(_SB_Local_VCInstallDir)' == ''">$(Registry:HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\VisualStudio\14.0\Setup\VC@ProductDir)</_SB_Local_VCInstallDir>
```
pradipd: >14.0 [](start = 136, length = 4)

What will happen in the next VS version? 15.0?  
Is it possible to use /NoDefaultLib in each UT and specify the libs with some VS variable that is set in MS.Cpp.Defaults.props

pradipd: They other reason I point this is out is, when VS 15 comes out, some poor sap, let's call him Artem, will be tasked with validating VS 15.  And, it will probably work because he'll have VS14 and VS15 side by side.  We'll sign off, and then somebody who only has VS15 installed will hit this.

But, if this is the only way, it's OK.  The VS targets/props are quite complex.  

---
In reply to: [74666079](https://github.com/Microsoft/WinObjC/pull/874#discussion_r74666079) [](ancestors = 74666079)

bbowman: I agree with Pradip here. THis is a bit scary. We might want to talk to VS folks to see what else we could do?

---
In reply to: [74666494](https://github.com/Microsoft/WinObjC/pull/874#discussion_r74666494) [](ancestors = 74666494,74666079)

DHowett-MSFT:
Revisiting this later.

PR #876, opened 2016-08-12
Adding scoped windows multi threaded runtime initialization #689.

msft-Jeyaram:


tests/unittests/EntryPoint.cpp
```diff
@@ -17,19 +17,23 @@
...
+#include <wrl\wrappers\corewrappers.h>
+using namespace Microsoft::WRL::Wrappers;
+#endif
+
 int main(int argc, char** argv) {
-// ::CoInitialize, Uninitialize, don't exist on OSX, aren't needed
 #ifdef WIN32
-    // Initialize COM for all of the tests
-    ::CoInitializeEx(NULL, COINIT_MULTITHREADED);
+    // Initialize the windows runtime, with uninitialized upon destructor invocation.
```
msft-Jeyaram: remove

msft-Jeyaram: >with [](start = 39, length = 4)

remove

tests/unittests/EntryPoint.cpp
```diff
@@ -1,6 +1,6 @@
 //******************************************************************************
 //
-// Copyright (c) 2015 Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
```
shahmeeresmail: Just curious, why remove the year instead of updating it to 2016? Is this a new policy going forward?

tests/unittests/EntryPoint.cpp
```diff
@@ -17,19 +17,23 @@
...
-    // Initialize COM for all of the tests
-    ::CoInitializeEx(NULL, COINIT_MULTITHREADED);
+    // Initialize the windows runtime, with uninitialized upon destructor invocation.
+    RoInitializeWrapper initialize(RO_INIT_MULTITHREADED);
+    if (FAILED(initialize)) {
+        return -1;
+    }
 
     testing::InitGoogleTest(&argc, argv);
     LOG_INFO("Starting unit tests...");
```
MSFTFox: >LOG_INFO [](start = 4, length = 8)

nit: if the message doesn't have to come after the InitGoogle call you could move the init and result lines out of the ifdef blocks.

MSFTFox:
:shipit:

DHowett-MSFT:
:shipit:

bbowman:
:shipit:

PR #877, opened 2016-08-12
Add Invalid Argument Throws to NSCharacterSet Initializers to match Reference Platform

MSFTFox:
NSCharacterSet characterSetWithBitmapRepresentation:nil and characterSetWithContentsOfFile:nil should throw illegal argument exceptions.

Two tests have been added to ensure this.

### DIFF
--- a/tests/unittests/Foundation/NSCharacterSetTests.m
+++ b/tests/unittests/Foundation/NSCharacterSetTests.m
@@ -225,15 +225,14 @@ ARM_DISABLED_TEST(NSCharacterSet, ExceptionTests) {
     NSMutableCharacterSet* set;
     @try {
         set = [NSMutableCharacterSet characterSetWithBitmapRepresentation:nil];
-        EXPECT_TRUE_MSG(false, "NSCharacterSet should have thrown invalid argument exception.");
+        FAIL();
     } @catch (NSException* e) {
         ASSERT_OBJCEQ(e.name, @"NSInvalidArgumentException");
     }
 
-    @
-    try {
+    @try {
         set = [NSMutableCharacterSet characterSetWithContentsOfFile:nil];
-        EXPECT_TRUE_MSG(false, "NSCharacterSet should have thrown invalid argument exception.");
+        FAIL();
     } @catch (NSException* e) {
         ASSERT_OBJCEQ(e.name, @"NSInvalidArgumentException");
     }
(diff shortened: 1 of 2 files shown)


tests/unittests/Foundation/NSCharacterSetTests.m
```diff
@@ -152,71 +152,89 @@ void testCharacter(NSCharacterSet* charSet, NSString* testString) {
...
+}
+
+ARM_DISABLED_TEST(NSCharacterSet, ExceptionTests) {
+    NSMutableCharacterSet* set;
+    @try {
+        set = [NSMutableCharacterSet characterSetWithBitmapRepresentation:nil];
+        EXPECT_TRUE_MSG(false, "NSCharacterSet should have thrown invalid argument exception.");
+    } @catch (NSException* e) {
+        ASSERT_OBJCEQ(e.name, @"NSInvalidArgumentException");
+    }
```
msft-Jeyaram: although for arm expect any throw seems to work.

msft-Jeyaram:
:shipit:

tests/unittests/Foundation/NSCharacterSetTests.m
MSFTFox: I'm not a fan of catching any throw. Sure it's expected that it's the correct throw but not guaranteed. It's more accurate to verify which exception was thrown.

msft-Jeyaram: resonable

---
In reply to: [74654704](https://github.com/Microsoft/WinObjC/pull/877#discussion_r74654704) [](ancestors = 74654704)

ms-jihua:
:shipit:

aballway:
:shipit:

tests/unittests/Foundation/NSCharacterSetTests.m
```diff
@@ -152,71 +152,89 @@ void testCharacter(NSCharacterSet* charSet, NSString* testString) {
...
+    @try {
+        set = [NSMutableCharacterSet characterSetWithBitmapRepresentation:nil];
+        FAIL();
+    } @catch (NSException* e) {
+        ASSERT_OBJCEQ(e.name, @"NSInvalidArgumentException");
+    }
+
+    @
+    try {
+        set = [NSMutableCharacterSet characterSetWithContentsOfFile:nil];
```
bbowman: weird formatting. Double check that you don't have anything else weird preventing this from being correct. (clang-format uses a syntax tree form to evaluate how format which means weird formatting can sometimes be indicative of other errors).

bbowman:
:shipit:

PR #883, opened 2016-08-16
Property synthesis issues causing use after release in UIDatePicker.mm

ehren:
There are some use after releases that occur when setting minimumDate/maximumDate in an application due to UIDatePicker.mm using the autosynthesized _minimumDate (_maximumDate) instead of the StrongId __minDate. There's something funny with auto property synthesis at the root of this - just removing one of the double underscores from the ivar results in:

    type of property 'minimumDate' ('NSDate *') does not match type of instance variable '_minimumDate' ('AutoId<NSDate, LifetimeRetain>')

No idea why this works for @property date / _date.

Commit just switches to manual memory management for minimumDate/maximumDate (unless someone knows what's going on with the property synthesis + smart types here).

### DIFF
--- a/Frameworks/UIKit/UIDatePicker.mm
+++ b/Frameworks/UIKit/UIDatePicker.mm
@@ -36,7 +36,7 @@ static const int MAX_DATEPICKER_COMPONENTS = 4;
     StrongId<NSMutableArray> _weekdays;
     StrongId<NSDate> _dayStartTime;
     UIDatePickerMode _mode;
-    StrongId<NSDate> __minimumDate, __maximumDate;
+    StrongId<NSDate> _minimumDate, _maximumDate;
     StrongId<UILabel> _hoursLabel, _minutesLabel;
     bool _didSnapTime;
 }
@@ -813,10 +813,24 @@ static void resetPickerPositions(UIDatePicker* self) {
 /**
  @Status Interoperable
 */
+- (NSDate*)minimumDate {
+    return _minimumDate;
+}
+
+/**
+ @Status Interoperable
+*/
 - (void)setMinimumDate:(NSDate*)date {
     _minimumDate = date;
     resetPickerPositions(self);
     [_pickerView _invalidateAllComponents];
+}
+
+/**
+ @Status Interoperable
+*/
+- (NSDate*)maximumDate {
+    return _maximumDate;
 }
 
 /**


mnithish:
:shipit:

Frameworks/UIKit/UIDatePicker.mm
```diff
@@ -814,6 +813,8 @@ static void resetPickerPositions(UIDatePicker* self) {
  @Status Interoperable
 */
 - (void)setMinimumDate:(NSDate*)date {
+    [date retain];
+    [_minimumDate release];
     _minimumDate = date;
```
bbowman: Can we change these ivars to use StrongId? That seems like it would be more concise and avoid other issues like this in the future.

bbowman:
:clock1:

Frameworks/UIKit/UIDatePicker.mm
DHowett-MSFT: This change backs out the strongid use for these properties..

ehren: That's the issue, one can't use StrongId because of the error: 

"type of property 'minimumDate' ('NSDate *') does not match type of instance variable '_minimumDate' ('AutoId<NSDate, LifetimeRetain>')

Note that the existing StrongIds are named __minimumDate, __maximumDate (two underscores). But the code is using the auto synthesized variables _minimumDate, _maximumDate resulting in use after frees.

The alternative fix would be to keep the __minimumDate, __maximumDate StrongIds and change the .mm file to use these (but then you would still have "extra" _minimumDate, _maximumDate ivars due to autosynthesis).

bbowman: Yeah I suppose its helpful if I actually read your change description :-p ... I need to check the .h but if we specify getters and setters in the property decl it *should* turn off synthesis barring compiler bugs right?

---
In reply to: [74983076](https://github.com/Microsoft/WinObjC/pull/883#discussion_r74983076) [](ancestors = 74983076)

bbowman: Actually I think if we just define our own getter, the compiler will turn off synthesis. That is the problem. The getter isn't defined so the compiler has to synthesize one with an ivar (one underscore and raw*). If we instead define it ourselves and return the strong id, no extra ivar will be produced, but we will have to change internal uses of self->_min/maxdate as those won't exist now. Basically we either need to go full custom with StrongId or full synthesized and I'm arguing for Strong id since it cleans up the memory management and we don't get the atomicity benefits of synthesized since they are declared as nonatomic anyway. How does that sound?

---
In reply to: [74986350](https://github.com/Microsoft/WinObjC/pull/883#discussion_r74986350) [](ancestors = 74986350,74983076)

bbowman: @DHowett-MSFT May have other comments on the general convention we want here?

---
In reply to: [74988055](https://github.com/Microsoft/WinObjC/pull/883#discussion_r74988055) [](ancestors = 74988055,74986350,74983076)

DHowett-MSFT:
@ehren The problem here is that we can't auto-@synthesize a property backed by an AutoId -- partly because our compiler isn't enlightened as to whether that should be possible, and partly because _maybe it shouldn't be possible._

Everywhere we've wanted to do that, we either
* Explicitly back the property with the right setter and getter that interact with the AutoId.
  * This is a little cumbersome, but gives us the most expressiveness.
  * This does not work with atomic properties, because we lack the runtime affordance for doing the same sort of property-scoped locking that the compiler does.

or
* Remove the AutoId and fall back to manual memory management.

DHowett-MSFT:
In short, @bbowman is right and also beat me to it. We lose atomicity, but we didn't need it in this case.

ehren:
@bbowman, @DHowett-MSFT makes sense. I think @bbowman is right that adding explicit setter=, getter= to the property declaration will turn off autosynthesis. I'll try rebasing with that change.

DHowett-MSFT:
Adding an explicit `setter=` and `getter=` shouldn't stop autosynthesis. Having an instance variable and writing the getters/setters directly, however, will!

ehren:
@DHowett-MSFT that's right, just adding an explicit getter was enough to disable the synthesis. Fix now in rebased commit.

Frameworks/UIKit/UIDatePicker.mm
```diff
@@ -843,6 +857,8 @@ - (void)dealloc {
     _dayStartDate = nil;
     _pickerView = nil;
     _months = nil;
+    _minimumDate = nil;
```
DHowett-MSFT: Since the `AutoId` destructor does the release for you, it's not necessary to stomp it with nil in dealloc! :+1: on this change though!

DHowett-MSFT:
:shipit:

Frameworks/UIKit/UIDatePicker.mm
ehren: removed those in final rebase

mnithish:
Thanks for contributing @ehren

Issue #712, opened 2016-07-29
Implement NSDecimalNumber


rajsesh-msft:

rajsesh-msft:
This is not yet ready to be merged, will have to wait until @msft-Jeyaram is back.

PR #896, opened 2016-08-19
NSDecimal + NSDecimalNumber Implementation.

msft-Jeyaram:
Holding off one the divide and multiplication as i'm still working on fixing some cases.
Will send out it as part of this CR. ( force squash will make it appear as one iteration in Codeflow)


msft-Jeyaram:
#712

Frameworks/Foundation/NSDecimal.mm
```diff
@@ -0,0 +1,592 @@
...
+    NSComparisonResult comparison = NSDecimalCompare(leftOperand, rightOperand);
+    if (comparison == NSOrderedSame) {
+        SET_TO_ZERO(result);
+        return NSCalculationNoError;
+    }
+
+    if (IS_NEGATIVE(leftOperand) && IS_NEGATIVE(rightOperand)) {
+        NEGATIVE_FLAG((&left)) = 0;
+        NEGATIVE_FLAG((&right)) = 0;
+        if ((error = _evalSubtractSameNumbers(result, &right, &left, comparison)) != NSCalculationNoError) {
```
msft-Jeyaram: will add note on the switch here.

rajsesh-msft:

rajsesh-msft:

rajsesh-msft:

include/Foundation/NSDecimal.h
```diff
@@ -1,48 +1,55 @@
...
+// THE SOFTWARE.
+//
+//******************************************************************************
+#pragma once
 
 #include <limits.h>
 #import <Foundation/NSObjCRuntime.h>
 
-#define NSDecimalMaxSize (8)
+ enum {
```
rajsesh-msft: >  [](start = 0, length = 1)

nit: clang formatting..?

include/Foundation/NSDecimal.h
```diff
@@ -1,48 +1,55 @@
...
-        unsigned int _length : 4;
-        unsigned int _isNegative : 1;
-        unsigned int _isCompact : 1;
-        unsigned int _reserved : 18;
-        unsigned short _mantissa[NSDecimalMaxSize];
-    };
-
-    double val;
+      signed int _exponent:8;
+      unsigned int _length:4;
```
rajsesh-msft: >_length [](start = 19, length = 7)

narrative of the fields will help.

Frameworks/Foundation/NSDecimal.mm
```diff
@@ -0,0 +1,592 @@
...
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#import <Foundation/NSDecimal.h>
+#import <StubReturn.h>
+#include <algorithm>
+
+#define EXPONENT_MAX 127
+#define EXPONENT_MIN -128
```
aballway: >#define [](start = 0, length = 7)

why not constants?

Frameworks/Foundation/NSDecimalNumber.mm
```diff
@@ -14,170 +14,191 @@
...
-    if (strstr(pStr, ".") != NULL) {
-        double fVal = strtod(pStr, NULL);
-        return [[self alloc] initWithDouble:fVal];
-    } else {
-        int64_t val = _strtoi64(pStr, NULL, 10);
-        return [[self alloc] initWithLongLong:val];
+- (instancetype)initWithDecimal:(NSDecimal)decimal {
+    if (self = [super init]) {
+        _decimal = new NSDecimal{ 0 };
+        NSDecimal* foo = new NSDecimal{ 0 };
```
msft-Jeyaram: >NSDecimal* foo = new NSDecimal{ 0 }; [](start = 8, length = 36)

removed;

Frameworks/Foundation/NSDecimal.mm
```diff
@@ -0,0 +1,592 @@
...
+
+    if (LENGTH(number) == 0) {
+        SET_TO_ZERO(number);
+    }
+}
+
+/**
+ @Status Interoperable
+*/
+NSComparisonResult NSDecimalCompare(const NSDecimal* leftOperand, const NSDecimal* rightOperand) {
```
rajsesh-msft: >NSDecimalCompare [](start = 19, length = 16)

No tests for this method.

Frameworks/Foundation/NSDecimal.mm
```diff
@@ -0,0 +1,592 @@
...
+
+    EXPONENT(result) = exponent;
+    return NSCalculationNoError;
+}
+
+/**
+ @Status Interoperable
+*/
+NSCalculationError NSDecimalNormalize(NSDecimal* number1, NSDecimal* number2, NSRoundingMode roundingMode) {
+    THROW_NS_IF_FALSE(E_INVALIDARG, (number1 != NULL) || (number2 != NULL));
```
aballway: >|| [](start = 54, length = 2)

should this be && ?

Frameworks/Foundation/NSDecimal.mm
msft-Jeyaram: this is internal, and I don't need to be presented as a var.
preprocessor placement is fine for it's limited use.

---
In reply to: [75560569](https://github.com/Microsoft/WinObjC/pull/896#discussion_r75560569) [](ancestors = 75560569)

Frameworks/Foundation/NSDecimal.mm
```diff
@@ -0,0 +1,592 @@
...
+        return IS_NEGATIVE(rightOperand) ? NSOrderedAscending : NSOrderedDescending;
+    }
+
+    return NSOrderedSame;
+}
+
+/**
+ @Status Interoperable
+*/
+void NSDecimalCopy(NSDecimal* destination, const NSDecimal* source) {
```
rajsesh-msft: no tests for this NSDecimalCopy.

Frameworks/Foundation/NSDecimal.mm
msft-Jeyaram: This is being tested via EXPECT_OBJCEQ when we compare two decimals in the tests.
It ends up calling NSDecimalNumber:compare, which calls this.

---
In reply to: [75560855](https://github.com/Microsoft/WinObjC/pull/896#discussion_r75560855) [](ancestors = 75560855)

Frameworks/Foundation/NSDecimal.mm
msft-Jeyaram: The method is being tested via NSDecimal functions that call into it, which are being called by NSDecimalNumber methods.

---
In reply to: [75560999](https://github.com/Microsoft/WinObjC/pull/896#discussion_r75560999) [](ancestors = 75560999)

tests/unittests/Foundation/NSDecimalNumberTests.mm
```diff
@@ -0,0 +1,269 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#import <TestFramework.h>
+#import <Foundation/Foundation.h>
+
+static void assertArrayEqual(unsigned short* a, unsigned short* b, int len) {
```
rajsesh-msft: >assertArrayEqual [](start = 12, length = 16)

should be a macro, or when this fails you wont see a proper line number.

Frameworks/Foundation/NSDecimal.mm
msft-Jeyaram: bah, yes. Also at other places.
I switched to this wearily named macro from a normal if check, looks like i messed it up. 

---
In reply to: [75560872](https://github.com/Microsoft/WinObjC/pull/896#discussion_r75560872) [](ancestors = 75560872)

Frameworks/Foundation/NSDecimal.mm
```diff
@@ -0,0 +1,592 @@
...
+        }
+        MANTISSA(decimal)[index] = 0;
+    }
+}
+
+/**
+ @Status Interoperable
+*/
+void NSDecimalRound(NSDecimal* result, const NSDecimal* number, NSInteger scale, NSRoundingMode roundingMode) {
+    THROW_NS_IF_FALSE(E_INVALIDARG, (result != NULL) || (number != NULL));
```
msft-Jeyaram: >|| [](start = 53, length = 2)

&&

tests/unittests/Foundation/NSDecimalNumberTests.mm
```diff
@@ -0,0 +1,269 @@
...
+
+    //(x) - (-x)
+    testSubtract([NSDecimalNumber decimalNumberWithMantissa:10123456789 exponent:-18 isNegative:NO],
+                 [NSDecimalNumber decimalNumberWithMantissa:10123456789 exponent:-18 isNegative:YES],
+                 [NSDecimalNumber decimalNumberWithMantissa:20246913578 exponent:-18 isNegative:NO]);
+
+    testSubtract([NSDecimalNumber decimalNumberWithMantissa:2 exponent:0 isNegative:NO], [NSDecimalNumber one], [NSDecimalNumber one]);
+}
+
+static void testPowerOf10(NSDecimalNumber* x, short power, NSDecimalNumber* expected) {
```
rajsesh-msft: >testPowerOf10 [](start = 12, length = 13)

macro here as well.

Frameworks/Foundation/NSDecimal.mm
```diff
@@ -0,0 +1,592 @@
...
+    }
+
+    THROW_NS_IF_FALSE(E_INVALIDARG, (leftOperand != NULL) || (rightOperand != NULL));
+
+    if (NEGATIVE_FLAG(leftOperand) != NEGATIVE_FLAG(rightOperand)) {
+        // miss match in sign.
+        return IS_NEGATIVE(rightOperand) ? NSOrderedDescending : NSOrderedAscending;
+    }
+
+    int leftValue = EXPONENT(leftOperand) + LENGTH(leftOperand);
```
rajsesh-msft: >leftValue [](start = 8, length = 9)

deceptive variable name.  This is really the number of digits > 0? leftWidth more appropriate (or leftLength)?

msft-Jeyaram: i'll change it to leftWidth;

---
In reply to: [75561578](https://github.com/Microsoft/WinObjC/pull/896#discussion_r75561578) [](ancestors = 75561578)

Frameworks/Foundation/NSDecimal.mm
```diff
@@ -0,0 +1,592 @@
...
+    int rightValue = EXPONENT(rightOperand) + LENGTH(rightOperand);
+    if (leftValue > rightValue) {
+        return IS_NEGATIVE(rightOperand) ? NSOrderedAscending : NSOrderedDescending;
+    }
+    if (leftValue < rightValue) {
+        return IS_NEGATIVE(rightOperand) ? NSOrderedDescending : NSOrderedAscending;
+    }
+
+    // same length
```
rajsesh-msft: >same length [](start = 7, length = 11)

length is again confusing here, the effective length (matissa + exp) is same, but the length of mantissa may be different.

Frameworks/Foundation/NSDecimal.mm
aballway: I think in general constants are better style than #defines, which I think even internally we should keep to the standard and avoid.  This file seems very macro heavy in general, many of which seem unnecessary and hurt readability.  

---
In reply to: [75560920](https://github.com/Microsoft/WinObjC/pull/896#discussion_r75560920) [](ancestors = 75560920,75560569)

Frameworks/Foundation/NSDecimal.mm
```diff
@@ -0,0 +1,592 @@
...
+
+#define VALIDATE_ARITHMETIC_INPUT(result, leftOperand, rightOperand)                \
+    if (NSDecimalIsNotANumber(leftOperand) || NSDecimalIsNotANumber(leftOperand)) { \
+        if (result) {                                                               \
+            INVALIDATE_DECIMAL(result);                                             \
+        }                                                                           \
+        return NSCalculationNoError;                                                \
+    }
+
+#define CHECK_DIVIDE_BY0(decimal)                           \
```
aballway: >CHECK_DIVIDE_BY0 [](start = 8, length = 16)

nit: CHECK_DIVIDE_BY_ZERO would be better

Frameworks/Foundation/NSDecimal.mm
msft-Jeyaram: that comment should be worded better. 

---
In reply to: [75561709](https://github.com/Microsoft/WinObjC/pull/896#discussion_r75561709) [](ancestors = 75561709)

Frameworks/Foundation/NSDecimal.mm
```diff
@@ -0,0 +1,592 @@
...
+        return IS_NEGATIVE(rightOperand) ? NSOrderedAscending : NSOrderedDescending;
+    }
+    if (leftValue < rightValue) {
+        return IS_NEGATIVE(rightOperand) ? NSOrderedDescending : NSOrderedAscending;
+    }
+
+    // same length
+    int index = std::min(LENGTH(leftOperand), LENGTH(rightOperand)) - 1;
+    while (index >= 0) {
+        if (MANTISSA(leftOperand)[index] < MANTISSA(rightOperand)[index]) {
```
rajsesh-msft: >( [](start = 20, length = 1)

I dont think this is quite correct.. you should only compare matissa after adjusting for exponent.

Frameworks/Foundation/NSDecimal.mm
```diff
@@ -0,0 +1,592 @@
...
+        NSDecimalCopy(result, x);
+        shortest = y;
+    } else {
+        NSDecimalCopy(result, y);
+        shortest = x;
+    }
+
+    int carryOver = 0;
+    int offset = (LENGTH(result) - LENGTH(shortest));
+    for (int shortIndex = LENGTH(shortest) - 1; shortIndex >= 0; --shortIndex) {
```
aballway: >LENGTH(shortest) [](start = 26, length = 16)

is this guaranteed to be > 0?

Frameworks/Foundation/NSDecimal.mm
aballway: We should have tests confirming this behaviour

---
In reply to: [75561368](https://github.com/Microsoft/WinObjC/pull/896#discussion_r75561368) [](ancestors = 75561368,75560872)

Frameworks/Foundation/NSDecimal.mm
msft-Jeyaram: it could be 0, in that case you will not enter the loop and exit with the result as y.

---
In reply to: [75562017](https://github.com/Microsoft/WinObjC/pull/896#discussion_r75562017) [](ancestors = 75562017)

Frameworks/Foundation/NSDecimal.mm
msft-Jeyaram: I think this could use more tests. will add more.

---
In reply to: [75561008](https://github.com/Microsoft/WinObjC/pull/896#discussion_r75561008) [](ancestors = 75561008,75560855)

Frameworks/Foundation/NSDecimal.mm
aballway: oh, I misread the comparison, nvm

---
In reply to: [75562328](https://github.com/Microsoft/WinObjC/pull/896#discussion_r75562328) [](ancestors = 75562328,75562017)

Frameworks/Foundation/NSDecimal.mm
msft-Jeyaram: uh I forgot to normalize. Looks like i didn't hit this unique case. 
Will add more tests to cover this.

---
In reply to: [75561877](https://github.com/Microsoft/WinObjC/pull/896#discussion_r75561877) [](ancestors = 75561877)

Frameworks/Foundation/NSDecimal.mm
```diff
@@ -0,0 +1,592 @@
...
+}
+
+/**
+ @Status Interoperable
+*/
+NSCalculationError NSDecimalMultiplyByPowerOf10(NSDecimal* result, const NSDecimal* number, short power, NSRoundingMode roundingMode) {
+    THROW_NS_IF_FALSE(E_INVALIDARG, (result != NULL) || (number != NULL));
+
+    NSDecimalCopy(result, number);
+    int exponent;
```
rajsesh-msft: >exponent [](start = 8, length = 8)

nit: initialize exponent = EXPONENT(result) + power here instead of in the if statement.

Frameworks/Foundation/NSDecimal.mm
```diff
@@ -0,0 +1,592 @@
...
+    if ((exponent = EXPONENT(result) + power) > EXPONENT_MAX) {
```
rajsesh-msft: >exponent [](start = 9, length = 8)

This is not entirely correct as you are allowed to use the entire mantissa and exponents.

Frameworks/Foundation/NSDecimal.mm
```diff
@@ -0,0 +1,592 @@
...
+    } else {
+        NSDecimalCopy(result, y);
+        shortest = x;
+    }
+
+    int carryOver = 0;
+    int offset = (LENGTH(result) - LENGTH(shortest));
+    for (int shortIndex = LENGTH(shortest) - 1; shortIndex >= 0; --shortIndex) {
+        int temp = (MANTISSA(shortest)[shortIndex] + MANTISSA(result)[offset + shortIndex] + carryOver);
+        MANTISSA(result)[offset + shortIndex] = temp % 10;
```
ms-jihua: I'm really confused about the role the 10 serves here, and why you're working in base10 at all. The carryover is much easier if you just do:

*reinterpret_cast<int*>(result[i+1]) = static_cast<int>(left[i]) + static_cast<int>(right[i])

If you do this I think you can skip the complicated steps below? The carryover bits will just naturally spill over.

tests/unittests/Foundation/NSDecimalNumberTests.mm
```diff
@@ -0,0 +1,269 @@
...
+    decimal = [price decimalValue];
+    ASSERT_EQ(4, decimal._length);
+    ASSERT_EQ(0, decimal._exponent);
+    ASSERT_EQ(1, decimal._isNegative);
+    assertArrayEqual(mantissa, decimal._mantissa, decimal._length);
+    ASSERT_EQ(-1599.0, [price doubleValue]);
+}
+
+TEST(NSDecimalNumber, decimalNumberAddition) {
+    NSDecimalNumber* d1 = [NSDecimalNumber decimalNumberWithMantissa:1599 exponent:-2 isNegative:NO];
```
ms-jihua: You also need:

- negative positive mixed test
- test values with much larger mantissas (greater than 2^32, so it takes up more than 2 shorts' worth)

Frameworks/Foundation/NSDecimal.mm
```diff
@@ -0,0 +1,592 @@
...
+        NSDecimalCopy(result, rightOperand);
+        return NSCalculationNoError;
+    }
+
+    if (IS_ZERO(rightOperand)) {
+        NSDecimalCopy(result, leftOperand);
+        // flip the sign
+        NEGATIVE_FLAG(result) = !NEGATIVE_FLAG(result);
+        return NSCalculationNoError;
+    }
```
bviglietta: These two conditions are backwards.

Frameworks/Foundation/NSDecimal.mm
```diff
@@ -0,0 +1,592 @@
...
+}
+
+/**
+ @Status Interoperable
+*/
+NSCalculationError NSDecimalSubtract(NSDecimal* result,
+                                     const NSDecimal* leftOperand,
+                                     const NSDecimal* rightOperand,
+                                     NSRoundingMode roundingMode) {
+    THROW_NS_IF_FALSE(E_INVALIDARG, (leftOperand != NULL) || (rightOperand != NULL) || (result != NULL));
```
bviglietta: >|| [](start = 58, length = 2)

&&

Frameworks/Foundation/NSDecimal.mm
```diff
@@ -0,0 +1,592 @@
...
+#define VALIDATE_ARITHMETIC_INPUT(result, leftOperand, rightOperand)                \
+    if (NSDecimalIsNotANumber(leftOperand) || NSDecimalIsNotANumber(leftOperand)) { \
+        if (result) {                                                               \
+            INVALIDATE_DECIMAL(result);                                             \
+        }                                                                           \
+        return NSCalculationNoError;                                                \
+    }
+
+#define CHECK_DIVIDE_BY0(decimal)                           \
+    NSDecimal zero = { 0 };                                 \
```
bviglietta: >NSDecimal zero [](start = 4, length = 14)

This is creating a variable called "zero" in the scope of wherever this macro is used. That means, for instance, that you can't use the macro twice in the same function.

Frameworks/Foundation/NSDecimal.mm
```diff
@@ -0,0 +1,592 @@
...
+    }
+
+    if (IS_ZERO(rightOperand)) {
+        NSDecimalCopy(result, leftOperand);
+        return NSCalculationNoError;
+    }
+
+    if (NEGATIVE_FLAG(leftOperand) != NEGATIVE_FLAG(rightOperand)) {
+        // We have sign miss match.
+        return (IS_NEGATIVE(rightOperand) ? NSDecimalSubtract(result, leftOperand, rightOperand, roundingMode) :
```
bviglietta: >NSDecimalSubtract(result, leftOperand, rightOperand, roundingMode) [](start = 44, length = 66)

The caller has asked for x+y and you're instead returning x-y.

Frameworks/Foundation/NSDecimal.mm
```diff
@@ -0,0 +1,592 @@
...
+#define EXPONENT_MAX 127
+#define EXPONENT_MIN -128
+
+#define MANTISSA(decimal) decimal->_mantissa
+
+#define LENGTH(decimal) decimal->_length
+
+#define EXPONENT(decimal) decimal->_exponent
+
+#define NEGATIVE_FLAG(decimal) decimal->_isNegative
```
bviglietta: >decimal [](start = 31, length = 7)

If you put the macro argument in parentheses here, you wouldn't have to use double parentheses when calling the macro, as you do later on.

Frameworks/Foundation/NSDecimal.mm
```diff
@@ -0,0 +1,592 @@
...
+ @Status Interoperable
+*/
+void NSDecimalCompact(NSDecimal* number) {
+    if ((number == NULL) || NSDecimalIsNotANumber(number)) {
+        return;
+    }
+
+    int indexer = 0;
+    // remove all the leading and trailing 0s of the mantissa.
+    while ((indexer < number->_length) && (MANTISSA(number)[indexer] == 0)) {
```
bviglietta: >number->_length [](start = 22, length = 15)

LENGTH(number)

Frameworks/Foundation/NSDecimal.mm
```diff
@@ -0,0 +1,592 @@
...
+    if (!source || !destination) {
+        return;
+    }
+
+    NEGATIVE_FLAG(destination) = NEGATIVE_FLAG(source);
+    EXPONENT(destination) = EXPONENT(source);
+    LENGTH(destination) = LENGTH(source);
+    destination->_isCompact = source->_isCompact;
+    destination->_reserved = source->_reserved;
+    memcpy(&MANTISSA(destination), &MANTISSA(source), NSDecimalMaxSize * sizeof(unsigned short));
```
bviglietta: *destination = *source;   ?

Frameworks/Foundation/NSDecimalNumber.mm
```diff
@@ -14,170 +14,191 @@
...
- @Status Stub
- @Notes
-*/
-NSCalculationError NSDecimalAdd(NSDecimal* result,
-                                const NSDecimal* leftOperand,
-                                const NSDecimal* rightOperand,
-                                NSRoundingMode roundingMode) {
-    UNIMPLEMENTED();
-    return StubReturn();
+typedef NS_ENUM(NSInteger, _DecimalOperand) { _Addition, _Subtract, _Multiply, _Divide };
```
bviglietta: >_DecimalOperand [](start = 27, length = 15)

Naming: these are operations, not operands.

Frameworks/Foundation/NSDecimalNumber.mm
```diff
@@ -186,214 +207,326 @@ + (NSDecimalNumber*)decimalNumberWithDecimal:(NSDecimal)decimal {
...
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
-+ (NSDecimalNumber*)one {
-    UNIMPLEMENTED();
-    return StubReturn();
+- (BOOL)isEqual:(NSObject*)objAddr {
+    if ([objAddr isKindOfClass:[NSDecimalNumber class]]) {
```
bviglietta: >if ([objAddr isKindOfClass:[NSDecimalNumber class]]) { [](start = 4, length = 54)

An NSDecimalNumber should compare equal to other NSNumbers with the same value, even if they aren't NSDecimalNumbers.

bviglietta:
>     NSDecimal ret = {[self doubleValue] };

This doesn't make much sense.

---
Refers to: Frameworks/Foundation/NSNumber.mm:523 in a31027d. [](commit_id = a31027d403f8e265ee5ceef5e44b11cb6c6ae2bb, deletion_comment = False)

Frameworks/Foundation/NSDecimalNumber.mm
```diff
@@ -14,170 +14,191 @@
...
- @Notes
+ @Status Interoperable
 */
-NSCalculationError NSDecimalPower(NSDecimal* result, const NSDecimal* number, NSUInteger power, NSRoundingMode roundingMode) {
-    UNIMPLEMENTED();
-    return StubReturn();
++ (NSDecimalNumber*)minimumDecimalNumber {
+    static dispatch_once_t __minNumberOnce;
+    dispatch_once(&__minNumberOnce, ^{
+        _minNumber = [[NSDecimalNumber alloc] initWithMantissa:std::numeric_limits<unsigned long long>::min() exponent:127 isNegative:YES];
```
bviglietta: >min() [](start = 104, length = 5)

Should be max(). The minimum unsigned long long value is zero.

Frameworks/Foundation/NSDecimalNumber.mm
```diff
@@ -14,170 +14,191 @@
...
+typedef NS_ENUM(NSInteger, _DecimalOperand) { _Addition, _Subtract, _Multiply, _Divide };
+
+static NSDecimalNumber* _zero;
+static NSDecimalNumber* _one;
+static NSDecimalNumber* _notANumber;
+static NSDecimalNumber* _maxNumber;
+static NSDecimalNumber* _minNumber;
+
+@implementation NSDecimalNumber {
+    NSDecimal* _decimal;
```
bviglietta: >NSDecimal* _decimal; [](start = 4, length = 20)

Why a pointer instead of storing the decimal by value?

Frameworks/Foundation/NSDecimalNumber.mm
```diff
@@ -14,170 +14,191 @@
...
  @Status Interoperable
 */
-- (const char*)objCType {
-    return "d";
+- (instancetype)initWithMantissa:(unsigned long long)mantissa exponent:(short)exponent isNegative:(BOOL)flag {
+    NSDecimal decimal = { 0 };
+    decimal._exponent = exponent;
+    decimal._isNegative = flag;
+
+    std::vector<unsigned short> digits;
```
bviglietta: >std::vector<unsigned short> digits; [](start = 4, length = 35)

You don't need a separate vector for this. Store the digits directly into decimal._mantissa, then reverse them in place.

Frameworks/Foundation/NSDecimalNumber.mm
```diff
@@ -428,9 +565,86 @@ + (NSDecimalNumber*)maximumDecimalNumber {
...
+- (instancetype)initWithFloat:(float)num {
     UNIMPLEMENTED();
     return StubReturn();
 }
 
+/**
+ @Status Interoperable
+*/
+- (instancetype)initWithInt:(int)num {
+    return [self initWithMantissa:((num < 0) ? -num : num) exponent:0 isNegative:((num < 0) ? YES : NO)];
```
bviglietta: >-num [](start = 47, length = 4)

If I do initWithInt:INT_MIN, this will overflow and produce undefined behavior. Depending on how the compiler feels, it might do what you want and it might not.

Frameworks/Foundation/NSDecimalNumber.mm
bviglietta: >_Addition [](start = 46, length = 9)

This is a noun and the other enum values are verbs.

Frameworks/Foundation/NSDecimal.mm
```diff
@@ -0,0 +1,592 @@
...
+        return;
+    }
+    if (scale == NSDecimalNoScale || (LENGTH(number) == 0)) {
+        NSDecimalCopy(result, number);
+        return;
+    }
+
+    int adjustedLengthOfMantissa = EXPONENT(number) + LENGTH(number) + scale;
+    if (adjustedLengthOfMantissa <= 0) {
+        // We pretty much rounded it off fully.
```
bviglietta: >// We pretty much rounded it off fully. [](start = 8, length = 39)

Not necessarily. If we round 7 to a scale of -1 with default rounding, it should round up to 10, not down to 0.

Frameworks/Foundation/NSDecimal.mm
```diff
@@ -0,0 +1,592 @@
...
+/**
+ @Status Stub
+ @Notes
+*/
+NSCalculationError NSDecimalPower(NSDecimal* result, const NSDecimal* number, NSUInteger power, NSRoundingMode roundingMode) {
+    UNIMPLEMENTED();
+    return StubReturn();
+}
+
+static void _roundUp(NSDecimal* decimal) {
```
bviglietta: >_roundUp [](start = 12, length = 8)

_roundUp(999) == 0 ?

Frameworks/Foundation/NSDecimal.mm
```diff
@@ -0,0 +1,592 @@
...
+ @Notes
+*/
+NSString* NSDecimalString(const NSDecimal* dcm, id locale) {
+    UNIMPLEMENTED();
+    return StubReturn();
+}
+
+static NSCalculationError _addSameSign(NSDecimal* result, NSDecimal* x, NSDecimal* y) {
+    // add starting at the end of the larger number.
+    // Note: we will have different placements of digits, since the decimal would of been compacted.
```
bviglietta: >would of [](start = 76, length = 8)

This is the worst've all possible errors.

Frameworks/Foundation/NSDecimal.mm
```diff
@@ -0,0 +1,592 @@
...
+
+    // If we still have a carry over, meaning we need to add one more place to the result and shift it.
+    // We need to validate if we have enough space for the placement and shift.
+    if (LENGTH(result) == NSDecimalMaxSize) {
+        // we don't have any space left.
+        INVALIDATE_DECIMAL(result);
+        return NSCalculationLossOfPrecision;
+    }
+
+    // do the shiffting by one and append the carryover.
```
bviglietta: >shiffting [](start = 14, length = 9)

shifting

Frameworks/Foundation/NSDecimalNumber.mm
```diff
@@ -14,170 +14,191 @@
...
 
 /**
  @Status Interoperable
 */
-- (const char*)objCType {
-    return "d";
+- (instancetype)initWithMantissa:(unsigned long long)mantissa exponent:(short)exponent isNegative:(BOOL)flag {
+    NSDecimal decimal = { 0 };
+    decimal._exponent = exponent;
+    decimal._isNegative = flag;
```
bviglietta: >decimal._isNegative = flag; [](start = 4, length = 27)

It's unclear whether you intend to support negative zero as a value. This initializer will allow you to create it, but other arithmetic functions like NSDecimalCompare will not handle it correctly.

include/Foundation/NSDecimal.h
bviglietta: >_length:4 [](start = 19, length = 9)

How are you going to represent values up to 38 in a 4-bit field?

include/Foundation/NSDecimal.h
```diff
@@ -1,48 +1,55 @@
...
+ enum {
+   NSCalculationNoError = 0,
+   NSCalculationLossOfPrecision,
+   NSCalculationUnderflow,
+   NSCalculationOverflow,
+   NSCalculationDivideByZero 
+};
+typedef NSUInteger NSCalculationError;
+
+#define NSDecimalMaxSize (38)
```
shahmeeresmail: Doesn't apple define this to be 8?
https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Miscellaneous/Foundation_Constants/#//apple_ref/doc/constant_group/NSDecimal_Constants

msft-Jeyaram:
Closing this out and starting a new CR

PR #900, opened 2016-08-20
NSUndoManager - Basic Undo/Redo Actions

### LINKED ISSUE #808: NSUndoManager - Basic Undo/Redo Registering
Provide basic functionality of NSUndoManager by allowing events to be registered with a basic selector to undo an action.
Examples:
[NSUndoManager* registerUndoWithTarget:target selector:@selector(functionForUndo) object:singleParameterForUndo];

[NSUndoManager* setActionName:NSLocalizedString(@"group.name", @"action name")];

[NSUndoManager* undo] and [NSUndoManager* redo].

MSFTFox:
Implement basic undo and redo operations for NSUndoManager with simple top level Undo Grouping.

Verified by new unit tests.

Fix #808


MSFTFox:
These changes do not support manual creation of undo groups nor nested groups. Additionally, this does not cover forwardInvocation undo operations.

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -16,16 +16,69 @@
...
- @Notes
+ @Status Interoperable
+*/
+- (id)init {
+    if (self = [super init]) {
+        _groupsByEvent = YES;
+        _NSUndoGroup* startingGroup = [[[_NSUndoGroup alloc] initWithOwner:self] autorelease];
+        _NSUndoGroup* startingRedoGroup = [[[_NSUndoGroup alloc] initWithOwner:self] autorelease];
+
+        _undoStack = [NSMutableArray array];
```
aballway: >_undoStack = [NSMutableArray array]; [](start = 8, length = 36)

nit: can this be  [NSMutableArray arrayAWithObject:startingGroup] ?

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -16,16 +16,69 @@
...
+ @Status Interoperable
 */
 - (void)registerUndoWithTarget:(id)target selector:(SEL)aSelector object:(id)anObject {
-    UNIMPLEMENTED();
+    if (!(self.isUndoRegistrationEnabled)) {
+        return;
+    }
+
+    _NSUndoCall* undoCall = [[[_NSUndoCall alloc] _initWithTarget:target selector:aSelector object:anObject] autorelease];
+    [self _registerUndoAction:undoCall];
```
aballway: nit: since this is so short, it'd probably be better to just flip the if and put this inside rather than return

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -143,4 +223,18 @@ - (void)setActionIsDiscardable:(BOOL)discardable {
     UNIMPLEMENTED();
 }
 
+/**
+ @Status Interoperable
+*/
+- (BOOL)canUndo {
+    return [((_NSUndoGroup*)[_undoStack lastObject])canUndo];
```
aballway: >((_NSUndoGroup*)[_undoStack lastObject])canUndo [](start = 12, length = 47)

nit: c++ cast here? also a space before canUndo

Frameworks/Foundation/_NSUndoCall.mm
```diff
@@ -0,0 +1,43 @@
...
+#import "_NSUndoCall.h"
+#import "_NSUndoObject.h"
+
+NSString* const NSUndoManagerGroupIsDiscardableKey = @"NSUndoManagerGroupIsDiscardableKey";
+
+@implementation _NSUndoCall : _NSUndoObject
+
+- (id)_initWithTarget:(id)target selector:(SEL)aSelector object:(id)anObject {
+    if (self = [super init]) {
+        self.undoAction = aSelector;
```
aballway: >self. [](start = 8, length = 5)

is there a reason these can't be _undoAction?

Frameworks/Foundation/_NSUndoCall.mm
```diff
@@ -0,0 +1,43 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#import <StubReturn.h>
+#import <Foundation/NSUndoManager.h>
+#import "_NSUndoCall.h"
+#import "_NSUndoObject.h"
```
aballway: is this import needed?

tests/unittests/Foundation/NSUndoManagerTests.mm
```diff
@@ -0,0 +1,161 @@
...
+    EXPECT_OBJCEQ(comparisonArray, object.arrayOfNumbers);
+
+    [object undo];
+
+    EXPECT_OBJCEQ([NSMutableArray array], object.arrayOfNumbers);
+
+    [object redo];
+
+    EXPECT_OBJCEQ(comparisonArray, object.arrayOfNumbers);
+}
```
aballway: should probably add tests for trying to undo/redo when there's nothing to undo/redo to guarantee the objects aren't modified

build/Foundation/lib/FoundationLib.vcxproj
```diff
@@ -237,6 +237,9 @@
     <ClangCompile Include="$(MSBuildThisFileDirectory)..\..\..\Frameworks\Foundation\_NSKeyValueCodingAggregateFunctions.mm" />
     <ClangCompile Include="$(MSBuildThisFileDirectory)..\..\..\Frameworks\Foundation\NSCFTimeZone.mm" />
     <ClangCompile Include="$(MSBuildThisFileDirectory)..\..\..\Frameworks\Foundation\NSString+HSTRING.mm" />
+    <ClangCompile Include="..\..\..\Frameworks\Foundation\_NSUndoCall.mm" />
```
DHowett-MSFT: `$(MSBuildThisFileDirectory)`

build/Tests/UnitTests/Foundation/Foundation.UnitTests.vcxproj.filters
```diff
@@ -134,6 +134,8 @@
     <ClangCompile Include="$(StarboardBasePath)\tests\unittests\Foundation\ReferenceFoundation\TestNSTimeZone.mm">
       <Filter>ReferenceFoundation</Filter>
     </ClangCompile>
+    <ClangCompile Include="$(StarboardBasePath)\tests\unittests\Foundation\NSURLResponseTests.mm" />
+    <ClangCompile Include="..\..\..\..\tests\unittests\Foundation\NSUndoManagerTests.mm" />
```
DHowett-MSFT: `$(StarboardBasePath)`

Frameworks/include/_NSUndoCall.h
```diff
@@ -0,0 +1,31 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+#pragma once
+
+#import <Foundation/Foundation.h>
```
DHowett-MSFT: Are these files intended to be visible to Foundation-external consumers? If not, please move them to Frameworks/Foundation

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -16,16 +16,69 @@
...
+        } else {
+            // Any new actions invalidate any redo actions
+            [self _removeRedoActions];
+            if ([_undoStack count] == 0) {
+                _NSUndoGroup* startingUndoGroup = [[[_NSUndoGroup alloc] initWithOwner:self] autorelease];
+                [_undoStack addObject:startingUndoGroup];
+            }
+            [[_undoStack lastObject] addUndoCallToUndoGroup:undoCall];
+        }
+    } else {
```
DHowett-MSFT: >else [](start = 6, length = 4)

Else: UNIMPLEMENTED_WITH_MSG()

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -16,16 +16,69 @@
...
+    _NSUndoCall* undoCall = [[[_NSUndoCall alloc] _initWithTarget:target selector:aSelector object:anObject] autorelease];
+    [self _registerUndoAction:undoCall];
+}
+
+// Internal function for proper registration of the deferred call.
+- (void)_registerUndoAction:(_NSUndoCall*)undoCall {
+    // If groupsByEvent is set then have a top level undo group to catch all undo calls.
+    if (_groupsByEvent) {
+        if (self.isUndoing) {
+            if ([_redoStack count] == 0) {
```
DHowett-MSFT: >if ([_redoStack count] == 0) { [](start = 12, length = 30)

I'm worried about who controls whether these groups exist: you create them in `init` AND you make sure they exist when registering an action.

Frameworks/Foundation/_NSUndoGroup.mm
```diff
@@ -0,0 +1,114 @@
...
+        [topObjectInGroup addUndoCallToUndoGroup:undoCall];
+    }
+    [_undoGrouping insertObject:undoCall atIndex:0];
+}
+
+- (void)callUndoGroup {
+    while ([_undoGrouping count] > 0) {
+        _NSUndoObject* object = [_undoGrouping firstObject];
+        if ([object isKindOfClass:[_NSUndoGroup class]]) {
+            [_owningManager callNestedGroup];
```
DHowett-MSFT: If you make all of these conform to an internal interface such as `_NSUndoing`, you don't need to switch on the class.

`_NSUndoGroup`'s implementation of `_NSUndoing` would just call the `_NSUndoing` implementation on every one of its subjugates.

With Objective-C, class checking is a code smell.

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -38,11 +91,19 @@ - (id)prepareWithInvocationTarget:(id)target {
...
 */
 - (void)undo {
-    UNIMPLEMENTED();
+    if (![self canUndo]) {
+        return;
+    }
+    _undoing = YES;
+
+    _NSUndoGroup* undoGroup = (_NSUndoGroup*)[_undoStack lastObject];
+    [undoGroup callUndoGroup];
```
DHowett-MSFT: The cast should be unnecessary with generic syntax. If it's necessary, please file a bug to get NSArray fixed.

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -38,11 +91,19 @@ - (id)prepareWithInvocationTarget:(id)target {
...
- @Notes
+ @Status Interoperable
 */
 - (void)undo {
-    UNIMPLEMENTED();
+    if (![self canUndo]) {
+        return;
+    }
+    _undoing = YES;
+
```
DHowett-MSFT: Is this intended to be threadsafe?

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -78,27 +147,38 @@ - (void)endUndoGrouping {
...
  @Notes
 */
 - (void)removeAllActions {
-    UNIMPLEMENTED();
+    [self _removeUndoActions];
+    [self _removeRedoActions];
+}
+
+- (void)_removeUndoActions {
+    _undoStack = [NSMutableArray array];
```
DHowett-MSFT: You're stomping a strong referenced array with an autoreleased one.

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -78,27 +147,38 @@ - (void)endUndoGrouping {
...
+    _NSUndoGroup* startingGroup = [[[_NSUndoGroup alloc] initWithOwner:self] autorelease];
+    [_undoStack addObject:startingGroup];
+}
+
+- (void)_removeRedoActions {
+    _redoStack = [NSMutableArray array];
```
DHowett-MSFT: You're stomping a strong referenced array with an autoreleased one.

Frameworks/Foundation/NSUndoManager.mm
DHowett-MSFT: Casts should not be necesasry.

---
In reply to: [75709894](https://github.com/Microsoft/WinObjC/pull/900#discussion_r75709894) [](ancestors = 75709894)

Frameworks/Foundation/_NSUndoCall.mm
```diff
@@ -0,0 +1,43 @@
...
+
+NSString* const NSUndoManagerGroupIsDiscardableKey = @"NSUndoManagerGroupIsDiscardableKey";
+
+@implementation _NSUndoCall : _NSUndoObject
+
+- (id)_initWithTarget:(id)target selector:(SEL)aSelector object:(id)anObject {
+    if (self = [super init]) {
+        self.undoAction = aSelector;
+        self.target = target;
+        self.object = [[anObject retain] autorelease];
```
DHowett-MSFT: >elf.object = [[anObject retain] autorelease]; [](start = 9, length = 45)

What.

The pool is almost guaranteed to drain before this _NSUndoCall is used.

Frameworks/Foundation/_NSUndoCall.mm
```diff
@@ -0,0 +1,43 @@
...
+
+#import <StubReturn.h>
+#import <Foundation/NSUndoManager.h>
+#import "_NSUndoCall.h"
+#import "_NSUndoObject.h"
+
+NSString* const NSUndoManagerGroupIsDiscardableKey = @"NSUndoManagerGroupIsDiscardableKey";
+
+@implementation _NSUndoCall : _NSUndoObject
+
```
DHowett-MSFT: No superclasses in @implementation.

Frameworks/Foundation/_NSUndoCall.mm
```diff
@@ -0,0 +1,43 @@
...
+//******************************************************************************
+
+#import <StubReturn.h>
+#import <Foundation/NSUndoManager.h>
+#import "_NSUndoCall.h"
+#import "_NSUndoObject.h"
+
+NSString* const NSUndoManagerGroupIsDiscardableKey = @"NSUndoManagerGroupIsDiscardableKey";
+
+@implementation _NSUndoCall : _NSUndoObject
```
DHowett-MSFT: >_NSUndoCall [](start = 16, length = 11)

`_NSBasicUndoAction`?

Frameworks/include/_NSUndoObject.h
```diff
@@ -0,0 +1,25 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+#pragma once
+
+#import <Foundation/Foundation.h>
+
+@interface _NSUndoObject : NSObject
```
DHowett-MSFT: With an appropriate `@protocol`, you won't need the base class.

DHowett-MSFT: (also: is both consumers override it, there's no reason to have it.)

---
In reply to: [75725333](https://github.com/Microsoft/WinObjC/pull/900#discussion_r75725333) [](ancestors = 75725333)

Frameworks/Foundation/_NSUndoCall.mm
```diff
@@ -0,0 +1,43 @@
...
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#import <StubReturn.h>
+#import <Foundation/NSUndoManager.h>
+#import "_NSUndoCall.h"
+#import "_NSUndoObject.h"
+
+NSString* const NSUndoManagerGroupIsDiscardableKey = @"NSUndoManagerGroupIsDiscardableKey";
```
DHowett-MSFT: >NSUndoManagerGroupIsDiscardableKey [](start = 55, length = 34)

Since this is a `Key`, you don't need `Key` in the name.

DHowett-MSFT: Also, why's that here and not in `NSUndoGroup` or `NSUndoManager`?

---
In reply to: [75725453](https://github.com/Microsoft/WinObjC/pull/900#discussion_r75725453) [](ancestors = 75725453)

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -16,16 +16,69 @@
...
 #import <StubReturn.h>
 #import <Foundation/NSUndoManager.h>
+#import <Foundation/NSMutableArray.h>
+#import "_NSUndoCall.h"
+#import "_NSUndoGroup.h"
 
-NSString* const NSUndoManagerGroupIsDiscardableKey = @"NSUndoManagerGroupIsDiscardableKey";
+@implementation NSUndoManager {
+    NSMutableArray<_NSUndoGroup*>* _undoStack;
+    NSMutableArray<_NSUndoGroup*>* _redoStack;
```
DHowett-MSFT: You could likely get away with making these `StrongId`

DHowett-MSFT:
:clock1:

Frameworks/Foundation/_NSUndoCall.mm
MSFTFox: Shouldn't be, removed it.

---
In reply to: [75725498](https://github.com/Microsoft/WinObjC/pull/900#discussion_r75725498) [](ancestors = 75725498,75725453)

Frameworks/Foundation/NSUndoManager.mm
MSFTFox: No, NSUndoManager is not thread safe. 

---
In reply to: [75724843](https://github.com/Microsoft/WinObjC/pull/900#discussion_r75724843) [](ancestors = 75724843)

Frameworks/Foundation/NSUndoManager.mm
MSFTFox: It's a strange situation. As called out in the design doc, a top level group will be created if _groupsByEvent is true. It will also automatically close the top level group when an undo message is sent.

---
In reply to: [75724410](https://github.com/Microsoft/WinObjC/pull/900#discussion_r75724410) [](ancestors = 75724410)

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -38,67 +89,99 @@ - (id)prepareWithInvocationTarget:(id)target {
...
 }
 
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 - (void)beginUndoGrouping {
-    UNIMPLEMENTED();
+    _NSUndoGroup* undoGroup = [_redoStack lastObject];
```
DHowett: This doesn't seem like it should be operating on the `redoStack`

tests/unittests/Foundation/NSUndoManagerTests.mm
```diff
@@ -0,0 +1,239 @@
+//******************************************************************************
+//
+// Copyright (c) 2016 Microsoft Corporation. All rights reserved.
```
rajsesh-msft: >Copyright [](start = 3, length = 9)

Please use new copyright header in the files you are touching.  This line should read "Copyright (c) Microsoft. All rights reserved."

tests/unittests/Foundation/NSUndoManagerTests.mm
```diff
@@ -0,0 +1,239 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#import <Foundation/Foundation.h>
+#import <TestFramework.h>
+
+@interface someObject : NSObject
```
rajsesh-msft: >someObject [](start = 11, length = 10)

Nit: SomeObject.

tests/unittests/Foundation/NSUndoManagerTests.mm
```diff
@@ -0,0 +1,239 @@
...
+- (void)updateNumberSingleUndoStep:(NSNumber*)value;
+- (void)addToMutableArray:(NSNumber*)value;
+- (void)removeFromMutableArray:(NSNumber*)value;
+@end
+
+@implementation someObject
+
+- (id)initWithValue:(NSNumber*)num {
+    self = [super init];
+    _undoManager = [[[NSUndoManager alloc] init] autorelease];
```
rajsesh-msft: >autorelease [](start = 49, length = 11)

There is no retain when you assign to _undoManager, so autorelease should not be there.

tests/unittests/Foundation/NSUndoManagerTests.mm
```diff
@@ -0,0 +1,239 @@
...
+    _value = num;
```
rajsesh-msft: >num [](start = 13, length = 3)

[num copy]?

tests/unittests/Foundation/NSUndoManagerTests.mm
```diff
@@ -0,0 +1,239 @@
...
+    self = [super init];
+    _undoManager = [[[NSUndoManager alloc] init] autorelease];
+    _value = nil;
+    _arrayOfNumbers = [NSMutableArray array];
+    return self;
+}
+
+- (void)updateNumber:(NSNumber*)value {
+    [_undoManager registerUndoWithTarget:self selector:@selector(updateNumber:) object:_value];
+    _value = value;
```
rajsesh-msft: >_value [](start = 4, length = 6)

self.value = value?

MSFTFox: I would rather do something like, _value = newValue, to stay consistent with using the ivar directly instead of the getters/setters.

---
In reply to: [76114880](https://github.com/Microsoft/WinObjC/pull/900#discussion_r76114880) [](ancestors = 76114880)

rajsesh-msft: As long as you maintain the ref counting 'contract'.  I'd rather stick to one convention - which is to use the ivars only in init methods (with careful attention to ref count) and use the properties everywhere else.  That way, you dont have to worry about _value = [newValue retain].

---
In reply to: [76117372](https://github.com/Microsoft/WinObjC/pull/900#discussion_r76117372) [](ancestors = 76117372,76114880)

Frameworks/Foundation/_NSUndoManagerInternal.h
```diff
@@ -0,0 +1,45 @@
...
+- (id)_initWithTarget:(id)target selector:(SEL)aSelector object:(id)anObject;
+
+@end
+
+@interface _NSUndoGroup : NSObject <_NSUndoable>
+
+- (id)initWithLevel:(NSInteger)level;
+- (void)addUndoCallToUndoGroup:(_NSUndoBasicAction*)undoCall;
+- (NSInteger)createUndoGroupWithLevel:(NSInteger)level;
+- (BOOL)closeUndoGroup;
```
aballway: >BOOL [](start = 3, length = 4)

should this be void?

Frameworks/Foundation/_NSUndoGroup.mm
```diff
@@ -0,0 +1,93 @@
...
+    [_undoGrouping insertObject:newGroup atIndex:0];
+    return level;
+}
+
+- (void)closeUndoGroup {
+    id<_NSUndoable> topObjectInGroup = [_undoGrouping firstObject];
+    if (topObjectInGroup != nil && ![topObjectInGroup isClosed]) {
+        [topObjectInGroup closeUndoGroup];
+    }
+    _depth--;
```
aballway: >_depth [](start = 4, length = 6)

Should check to make sure !isClosed before doing this or _depth would go negative

Frameworks/Foundation/_NSUndoBasicAction.mm
```diff
@@ -0,0 +1,54 @@
...
+    StrongId<id> _object;
+    id _target;
+    SEL _undoAction;
+}
+
+- (id)_initWithTarget:(id)target selector:(SEL)aSelector object:(id)anObject {
+    if (self = [super init]) {
+        _undoAction = aSelector;
+        _target = objc_storeWeak(&_target, target);
+        _object = [anObject retain];
```
rajsesh-msft: >retain [](start = 28, length = 6)

StrongId<> does a retain, you are leaking anObject as a result.

Frameworks/Foundation/_NSUndoBasicAction.mm
```diff
@@ -0,0 +1,54 @@
...
+- (void)undo {
+    [objc_loadWeak(&_target) performSelector:_undoAction withObject:_object];
+}
+
+- (BOOL)canUndo {
+    return YES;
+}
+
+- (void)dealloc {
+    [_object release];
```
rajsesh-msft: no need to do this, StrongId<> does it for you.  Use _object = nil if you must.

rajsesh-msft:
Can you please add your design doc to the PR?

Frameworks/Foundation/_NSUndoGroup.mm
```diff
@@ -0,0 +1,93 @@
...
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#import <Starboard.h>
+#import <Foundation/NSUndoManager.h>
+#import "_NSUndoManagerInternal.h"
+
+@implementation _NSUndoGroup {
+    StrongId<NSMutableArray*> _undoGrouping;
```
rajsesh-msft: Given that this is a stack, using std::stack<> would be much more efficient for insert and remove at index 0.

tests/unittests/Foundation/NSUndoManagerTests.mm
```diff
@@ -0,0 +1,239 @@
...
+}
+
+- (void)removeFromMutableArray:(NSNumber*)value {
+    [_undoManager registerUndoWithTarget:self selector:@selector(addToMutableArray:) object:value];
+    [_arrayOfNumbers removeObject:value];
+}
+
+@end
+
+TEST(NSUndoManager, BasicUndoTest) {
```
rajsesh-msft: stating the obvious here, these tests do pass on reference platform, correct?

Frameworks/Foundation/_NSUndoGroup.mm
```diff
@@ -0,0 +1,93 @@
...
+    id<_NSUndoable> topObjectInGroup = [_undoGrouping firstObject];
+    if (topObjectInGroup != nil && ![topObjectInGroup isClosed]) {
+        [topObjectInGroup closeUndoGroup];
+    }
+    _depth--;
+}
+
+- (BOOL)canUndo {
+    // Top level undo group does not need to be closed.
+    _NSUndoBasicAction* topObjectInGroup = [_undoGrouping firstObject];
```
rajsesh-msft: >_NSUndoBasicAction [](start = 4, length = 18)

nit: id<_NSUndoable> is sufficient (and possibly more correct).

rajsesh-msft:
:clock1:

Frameworks/Foundation/_NSUndoGroup.mm
MSFTFox: We will however have to traverse this "stack" for undo operations on particular targets though with the -removeAllActionsWithTarget:(id)target method. I figured it would be significantly worse to use std::stack in that situation.

---
In reply to: [76124040](https://github.com/Microsoft/WinObjC/pull/900#discussion_r76124040) [](ancestors = 76124040)

tests/unittests/Foundation/NSUndoManagerTests.mm
MSFTFox: retain. There's no reason to copy a number.

---
In reply to: [76114710](https://github.com/Microsoft/WinObjC/pull/900#discussion_r76114710) [](ancestors = 76114710)

tests/unittests/Foundation/NSUndoManagerTests.mm
MSFTFox: Correct. I've been testing thoroughly.

---
In reply to: [76124255](https://github.com/Microsoft/WinObjC/pull/900#discussion_r76124255) [](ancestors = 76124255)

Frameworks/Foundation/_NSUndoGroup.mm
rajsesh-msft: Fair enough, using a list would be more appropriate then, with O(N) traversal (same as array) and O(1) insert/remove.  In normal operation, adding operations and performing undo will be far more frequent than removing specific target.

---
In reply to: [76125991](https://github.com/Microsoft/WinObjC/pull/900#discussion_r76125991) [](ancestors = 76125991,76124040)

Frameworks/Foundation/_NSUndoManagerStack.mm
```diff
@@ -0,0 +1,47 @@
...
+- (id<_NSUndoable>)peek {
+    if ([self count] > 0) {
+        return _stack.front();
+    }
+    return nil;
+}
+
+- (void)push:(id<_NSUndoable>)undoObject {
+    _stack.push_front(undoObject);
+    [undoObject release];
```
DHowett-MSFT: My advice to release this only applied if you were pushing it onto the std::list in the same place you were allocating it.

If you think about it, this means that `push:` is now a consumer for a +1 reference. That's not what something named push should do.

Why not just have the `std::list`s right on `NSUndoManager`? It seems like we could avoid the dynamic dispatch here.

Frameworks/Foundation/_NSUndoGroup.mm
```diff
@@ -0,0 +1,83 @@
...
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#import <Starboard.h>
+#import <Foundation/NSUndoManager.h>
+#import "_NSUndoManagerInternal.h"
+
+@implementation _NSUndoGroup {
+    StrongId<_NSUndoManagerStack*> _undoGrouping;
```
DHowett-MSFT: >_NSUndoManagerStack [](start = 13, length = 19)

Don't use `*` with StrongId; while it will work, it doesn't match the rest of the code.

tests/unittests/Foundation/NSUndoManagerTests.mm
```diff
@@ -0,0 +1,266 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#import <Foundation/Foundation.h>
+#import <TestFramework.h>
+
+@interface SomeObject : NSObject
```
DHowett-MSFT: >SomeObject [](start = 11, length = 10)

Can you give this a name that won't necessarily collide with other tests?

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -38,67 +102,159 @@ - (id)prepareWithInvocationTarget:(id)target {
...
 }
 
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 - (void)disableUndoRegistration {
-    UNIMPLEMENTED();
+    _disableCount++;
```
DHowett-MSFT: >_disableCount [](start = 4, length = 13)

Why have both a counter and a flag that says whether the counter is 0? That just screams "we can drift out of sync!"

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -38,67 +102,159 @@ - (id)prepareWithInvocationTarget:(id)target {
...
+ @Status Interoperable
 */
 - (void)removeAllActions {
-    UNIMPLEMENTED();
+    [self _removeUndoActions];
+    [self _removeRedoActions];
+}
+
+- (void)_removeUndoActions {
+    _undoStack = [[_NSUndoManagerStack alloc] init];
```
DHowett-MSFT: > [[_NSUndoManagerStack alloc] init]; [](start = 16, length = 36)

You are putting a +1 reference into a `StrongId`; you have to use `.attach()`

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -38,67 +102,159 @@ - (id)prepareWithInvocationTarget:(id)target {
...
 */
 - (void)redo {
-    UNIMPLEMENTED();
+    _groupingLevel = [_redoStack count];
+    if (![self canRedo]) {
+        return;
+    }
+
+    if (_undoing) {
+        [NSException raise:NSInternalInconsistencyException format:@"Cannot invoke redo while undoing is active."];
```
DHowett-MSFT: >Cannot invoke redo while undoing is active. [](start = 69, length = 43)

Phrase exceptions more as a statement of fact; "-redo called during an undo operation." or something.

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -38,67 +102,159 @@ - (id)prepareWithInvocationTarget:(id)target {
...
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 - (void)redo {
-    UNIMPLEMENTED();
+    _groupingLevel = [_redoStack count];
+    if (![self canRedo]) {
+        return;
+    }
```
DHowett-MSFT: Might as well check this before you get the grouping level!

DHowett-MSFT: And/or throw the exception.

---
In reply to: [76160709](https://github.com/Microsoft/WinObjC/pull/900#discussion_r76160709) [](ancestors = 76160709)

DHowett-MSFT: Because you don't want to adjust the grouping level if you're about to bail out.

---
In reply to: [76160720](https://github.com/Microsoft/WinObjC/pull/900#discussion_r76160720) [](ancestors = 76160720,76160709)

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -38,67 +102,159 @@ - (id)prepareWithInvocationTarget:(id)target {
...
+    _NSUndoGroup* undoGroup = [_undoStack peek];
+
+    // Record redo operation as a single redo group
+    [self _beginRedoGrouping];
+    [undoGroup undo];
+    [self _endRedoGrouping];
+    [_undoStack pop];
+
+    _groupingLevel = [_undoStack count];
+    ;
```
DHowett-MSFT: >  ; [](start = 2, length = 3)

`;`

Frameworks/Foundation/NSUndoManager.mm
MSFTFox: That's actually the point believe it or not. Multiple disable calls are recorded. Enable calls must equal the amount of disable calls when invoking any undo operation. Otherwise, an inconsistency exception is raised.

---
In reply to: [76160566](https://github.com/Microsoft/WinObjC/pull/900#discussion_r76160566) [](ancestors = 76160566)

DHowett-MSFT: That doesn't change the fact that _disableCount==0 means "It's not disabled"; you're using a second flag to store whether this is 0... when you could just check that it is 0.

---
In reply to: [76288487](https://github.com/Microsoft/WinObjC/pull/900#discussion_r76288487) [](ancestors = 76288487,76160566)

MSFTFox: I thought that was correct since there's a bool property for registration but the implementation expects something else. Should we just forego the property here then?

---
In reply to: [76291490](https://github.com/Microsoft/WinObjC/pull/900#discussion_r76291490) [](ancestors = 76291490,76288487,76160566)

Frameworks/Foundation/_NSUndoManagerStack.mm
MSFTFox: The reason I broke it out into this separate class was to avoid cluttering undo manager. It turned out to be a smaller class than I expected. Particularly I wanted the the remove call to be done in here to keep the logic for walking this tree isolated.

---
In reply to: [76160405](https://github.com/Microsoft/WinObjC/pull/900#discussion_r76160405) [](ancestors = 76160405)

rajsesh-msft: There isn't enough in this class to break this out, i second Dustin's comment to keep the list directly in NSUndoManager.

---
In reply to: [76294970](https://github.com/Microsoft/WinObjC/pull/900#discussion_r76294970) [](ancestors = 76294970,76160405)

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -38,67 +102,159 @@ - (id)prepareWithInvocationTarget:(id)target {
...
 }
 
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 - (void)enableUndoRegistration {
-    UNIMPLEMENTED();
+    if (_disableCount == 0) {
```
rajsesh-msft: is threadsafety a concern?

MSFTFox: It is not, NSUndoManager makes no promises of thread safety. However, I've read that a common pattern is for multiple objects to call enable/disable. Making this counter thread safe couldn't hurt.

---
In reply to: [76338299](https://github.com/Microsoft/WinObjC/pull/900#discussion_r76338299) [](ancestors = 76338299)

Frameworks/Foundation/_NSUndoManagerStack.mm
MSFTFox: However, the same code is reused many many times. Removing this class would clutter the NSUndoManager. In addition, it makes sense to have the remove logic that will be added in this class as it deals strictly with the stack.

---
In reply to: [76338176](https://github.com/Microsoft/WinObjC/pull/900#discussion_r76338176) [](ancestors = 76338176,76294970,76160405)

MSFTFox:
This turned into implementing NSUndoManager in full. Apart from the forward invocation calls and UI involved changes (such as action and menu item names) this class should be working as intended.

Frameworks/Foundation/_NSUndoManagerInternal.h
```diff
@@ -0,0 +1,53 @@
...
+@interface _NSUndoBasicAction : NSObject <_NSUndoable>
+
+- (id)_initWithTarget:(id)target selector:(SEL)aSelector object:(id)anObject modes:(NSArray<NSString*>*)runLoopModes;
+
+@end
+
+@interface _NSUndoGroup : NSObject <_NSUndoable>
+
+- (id)init;
+- (void)addUndoCallToUndoGroup:(_NSUndoBasicAction*)undoCall;
```
DHowett-MSFT: Why not:

1. Use `id<_NSUndoable>` for the argument type
2. Rename it to `addUndoAction:` (rationale: `-[NSUndoGroup addUndoCallToUndoGroup:]` is unnecessarily redundant, since you're already operating on an undo group

Frameworks/Foundation/_NSUndoManagerInternal.h
```diff
@@ -0,0 +1,53 @@
...
+- (void)createUndoGroup;
```
DHowett-MSFT: >createUndoGroup [](start = 8, length = 15)

These method names are unclear to me. What does it mean for an undo group to create an undo grou?

Frameworks/Foundation/_NSUndoManagerStack.mm
```diff
@@ -0,0 +1,67 @@
...
+- (void)push:(id<_NSUndoable>)undoObject {
+    _stack.push_front(undoObject);
+}
+
+- (NSUInteger)count {
+    return _stack.size();
+}
+
+- (void)removeAllWithTarget:(id)target {
+    std::list<StrongId<id<_NSUndoable>>>::const_iterator iterator = _stack.begin();
```
DHowett-MSFT: >std::list<StrongId<id<_NSUndoable>>>::const_iterator [](start = 4, length = 52)

These are great candidates for `auto`

Frameworks/Foundation/_NSUndoManagerInternal.h
DHowett-MSFT: (Naming it `addUndoAction:` also matches the kind of thing you're adding to it: an action.)

---
In reply to: [76707255](https://github.com/Microsoft/WinObjC/pull/900#discussion_r76707255) [](ancestors = 76707255)

DHowett-MSFT: Or even, `addUndoable:`)

---
In reply to: [76707650](https://github.com/Microsoft/WinObjC/pull/900#discussion_r76707650) [](ancestors = 76707650,76707255)

Frameworks/Foundation/_NSUndoBasicAction.mm
```diff
@@ -0,0 +1,60 @@
...
+        _modes = runLoopModes;
+    }
+    return self;
+}
+
+- (void)undo:(BOOL)undoAll {
+    [objc_loadWeak(&_target) performSelector:_undoAction withObject:_object];
+}
+
+- (void)dealloc {
```
DHowett-MSFT: >dealloc [](start = 8, length = 7)

Make certain you call super.

Frameworks/Foundation/_NSUndoBasicAction.mm
```diff
@@ -0,0 +1,60 @@
...
+    SEL _undoAction;
+    StrongId<NSArray<NSString*>*> _modes;
+}
+
+- (id)_initWithTarget:(id)target selector:(SEL)aSelector object:(id)anObject modes:(NSArray<NSString*>*)runLoopModes {
+    if (self = [super init]) {
+        _undoAction = aSelector;
+        _target = objc_storeWeak(&_target, target);
+        _object = anObject;
+        _modes = runLoopModes;
```
DHowett-MSFT: >runLoopModes [](start = 17, length = 12)

Arrays should be copied in; otherwise, they could be mutated by the caller.

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -14,18 +14,84 @@
 //
 //******************************************************************************
 
+#import <Starboard.h>
 #import <StubReturn.h>
 #import <Foundation/NSUndoManager.h>
+#import "_NSUndoManagerInternal.h"
 
 NSString* const NSUndoManagerGroupIsDiscardableKey = @"NSUndoManagerGroupIsDiscardableKey";
```
DHowett-MSFT: >NSUndoManagerGroupIsDiscardableKey [](start = 55, length = 34)

nit: Key should only be in the name, not in the value.

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -14,18 +14,84 @@
...
+}
+
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
+*/
+- (id)init {
+    if (self = [super init]) {
+        _runLoopModes = @[ NSDefaultRunLoopMode ];
```
DHowett-MSFT: >_runLoopModes [](start = 8, length = 13)

_runLoopModes is a +1 property, by virtue of being marked `copy`. Please make sure it is holding onto a +1 object.

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -38,83 +104,231 @@ - (id)prepareWithInvocationTarget:(id)target {
...
+ @Status Interoperable
 */
 - (void)beginUndoGrouping {
-    UNIMPLEMENTED();
+    if (_undoing) {
+        return;
+    }
+    // Over the limit on undo operations
+    if (self.levelsOfUndo != 0 && _undoGroupingLevel == 0 && [_undoStack count] >= self.levelsOfUndo) {
+        [_undoStack popBack];
```
DHowett-MSFT: >popBack [](start = 20, length = 7)

`-popBack` is not documented in the public interface for _NSUndoManagerStack.

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -38,83 +104,231 @@ - (id)prepareWithInvocationTarget:(id)target {
...
- @Notes
+ @Status Interoperable
 */
 - (void)removeAllActions {
-    UNIMPLEMENTED();
+    [self _removeUndoActions];
+    [self _removeRedoActions];
+}
+
+- (void)_removeUndoActions {
```
DHowett-MSFT: >_removeUndoActions [](start = 8, length = 18)

Should these also reset the depths? You are keeping track of a lot of state, but you're not resetting it all when you call `removeAllActions`

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -38,83 +104,231 @@ - (id)prepareWithInvocationTarget:(id)target {
...
 
 /**
- @Status Stub
+ @Status Interoperable
  @Notes
 */
 - (void)removeAllActionsWithTarget:(id)target {
-    UNIMPLEMENTED();
+    [_undoStack removeAllWithTarget:target];
+    [_redoStack removeAllWithTarget:target];
```
DHowett-MSFT: What if a group becomes empty while you do this??

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -38,83 +104,231 @@ - (id)prepareWithInvocationTarget:(id)target {
...
- @Status Stub
+ @Status Interoperable
  @Notes
 */
 - (void)setActionName:(NSString*)actionName {
-    UNIMPLEMENTED();
+    if (_undoing) {
+        _redoActionName = [actionName copy];
+    } else {
+        _undoActionName = [actionName copy];
```
DHowett-MSFT:
```
_x = [x copy];
_x = [x copy]; // The old value of _x is leaked.
```

Please be careful with memory management :)

Frameworks/Foundation/NSUndoManager.mm
ms-jihua: how do Good Practices feel about using the setter in the initializer here to implicitly ensure that this has the right retain level?

---
In reply to: [76707964](https://github.com/Microsoft/WinObjC/pull/900#discussion_r76707964) [](ancestors = 76707964)

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -14,18 +14,84 @@
...
+
+- (void)_createAutomaticUndoGroup {
+    if (!_isAutomaticGroupOpen && _undoGroupingLevel == 0 && _groupsByEvent) {
+        _isAutomaticGroupOpen = YES;
+        [self beginUndoGrouping];
+    }
+}
+
+// Internal function for proper registration of the deferred call.
+- (void)_registerUndoAction:(_NSUndoBasicAction*)undoCall {
```
ms-jihua: nit: This should probably be above its call (does it generate a warning because this symbol hasn't been declared yet?). This ordering also, to me at least, feels more immediately intuitive

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -14,18 +14,84 @@
...
+    if (_undoing) {
+        if ([_redoStack count] == 0) {
+            _NSUndoGroup* startingRedoGroup = [[_NSUndoGroup alloc] init];
+            [_redoStack push:startingRedoGroup];
+            [startingRedoGroup release];
```
ms-jihua: >[startingRedoGroup release]; [](start = 12, length = 28)

nit: would autoreleasing above cause any issues? (do you depend on the -1 being done already by the time this function exits?)

Frameworks/Foundation/_NSUndoGroup.mm
```diff
@@ -0,0 +1,91 @@
...
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#import <Starboard.h>
+#import <Foundation/NSUndoManager.h>
+#import "_NSUndoManagerInternal.h"
+
+@implementation _NSUndoGroup {
+    StrongId<_NSUndoManagerStack> _undoGrouping;
```
ms-jihua: >_undoGrouping; [](start = 34, length = 14)

not sure how I feel about this variable name basically being a repeat of the class name - got me confused several times

Frameworks/Foundation/NSUndoManager.mm
MSFTFox: Most likely not, but it's more clear this way that the stack has the object now.

---
In reply to: [76881492](https://github.com/Microsoft/WinObjC/pull/900#discussion_r76881492) [](ancestors = 76881492)

Frameworks/Foundation/NSUndoManager.mm
MSFTFox: It's now removed from the stack if it's empty.

---
In reply to: [76708572](https://github.com/Microsoft/WinObjC/pull/900#discussion_r76708572) [](ancestors = 76708572)

Frameworks/Foundation/NSUndoManager.mm
MSFTFox: I thought calling any setter in the init was always a bad idea.

---
In reply to: [76852007](https://github.com/Microsoft/WinObjC/pull/900#discussion_r76852007) [](ancestors = 76852007,76707964)

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -38,83 +104,247 @@ - (id)prepareWithInvocationTarget:(id)target {
...
 */
 - (void)setActionName:(NSString*)actionName {
-    UNIMPLEMENTED();
+    if (_undoing) {
+        if (_redoActionName) {
+            [_redoActionName release];
+        }
+        _redoActionName = [actionName copy];
+    } else {
+        if (_undoActionName) {
```
DHowett-MSFT: >if (_undoActionName) { [](start = 8, length = 22)

`[nil release]` is valid. There's no reason to do these checks.

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -14,18 +14,84 @@
...
 
+#import <Starboard.h>
 #import <StubReturn.h>
 #import <Foundation/NSUndoManager.h>
+#import "_NSUndoManagerInternal.h"
 
-NSString* const NSUndoManagerGroupIsDiscardableKey = @"NSUndoManagerGroupIsDiscardableKey";
+NSString* const NSUndoManagerGroupIsDiscardableKey = @"NSUndoManagerGroupIsDiscardable";
+
+@implementation NSUndoManager {
```
DHowett-MSFT: >implementation [](start = 1, length = 14)

This class has `retain` and `copy` properties but it does not have a `dealloc` :|

Frameworks/Foundation/_NSUndoManagerStack.mm
```diff
@@ -0,0 +1,74 @@
...
+            ++iterator;
+            _stack.erase(toRemove);
+        } else {
+            [*iterator removeAllWithTarget:target];
+            if ([*iterator count] == 0 && [*iterator isClosed]) {
+                auto toRemove = iterator;
+                ++iterator;
+                _stack.erase(toRemove);
+            } else {
+                // Must do this in else as ordering of iterator must be incremented before removing.
```
DHowett-MSFT: >Must do this in else as ordering of iterator must be incremented before removing. [](start = 19, length = 81)

This comment needs to be reworded to make sense.

Frameworks/Foundation/NSUndoManager.mm
DHowett-MSFT: Why is it `removeAllActions`' responsibility to set the grouping level to 0, but there's a special function called `_removeUndoActions` that doesn't?

It seems like you'd either:
1) have one function that clears all states
or
2) have one function per stack that resets _stack-specific_ state.

Right now, you have two functions, one that clears half the states and then individual ones that clear the other halves of the states.

---
In reply to: [76708543](https://github.com/Microsoft/WinObjC/pull/900#discussion_r76708543) [](ancestors = 76708543)

Frameworks/Foundation/NSUndoManager.mm
```diff
@@ -38,83 +104,247 @@ - (id)prepareWithInvocationTarget:(id)target {
...
- @Status Stub
+ @Status Interoperable
  @Notes
 */
 - (void)removeAllActionsWithTarget:(id)target {
-    UNIMPLEMENTED();
+    [_undoStack removeAllWithTarget:target];
+    [_redoStack removeAllWithTarget:target];
+
+    _undoGroupingLevel = [[_undoStack peek] updateLevel];
```
DHowett-MSFT: >updateLevel [](start = 44, length = 11)

Is this called `updateLevel` because it is a level of updates, or because it will update the level.

If the first, why are we talking about updates? If the second, it doesn't look like it's setting anything... it isn't performing an update.

MSFTFox: If neither, make my method names actually describe what they do.

---
In reply to: [76904743](https://github.com/Microsoft/WinObjC/pull/900#discussion_r76904743) [](ancestors = 76904743)

tests/unittests/Foundation/NSUndoManagerTests.mm
```diff
@@ -0,0 +1,718 @@
...
+    _value = nil;
+    _arrayOfNumbers = [[NSMutableArray array] retain];
+    return self;
+}
+
+- (void)dealloc {
+    [_undoManager release];
+    [_value release];
+    [_arrayOfNumbers release];
+}
```
DHowett-MSFT: Call `[super dealloc]`. . .

DHowett-MSFT: Can you search for other deallocs in this code review and make sure they call `super` too?

---
In reply to: [77097364](https://github.com/Microsoft/WinObjC/pull/900#discussion_r77097364) [](ancestors = 77097364)

tests/unittests/Foundation/NSUndoManagerTests.mm
```diff
@@ -0,0 +1,718 @@
...
+@end
+
+TEST(NSUndoManager, BasicUndoTest) {
+    NSNumber* oldValue = [NSNumber numberWithInt:5];
+    NSNumber* newValue = [NSNumber numberWithInt:6];
+
+    UndoTestObject* object = [[[UndoTestObject alloc] initWithValue:oldValue] autorelease];
+    EXPECT_EQ([object.undoManager canUndo], NO);
+
+    EXPECT_EQ(object.value.intValue, oldValue.intValue);
```
DHowett-MSFT: >EXPECT_EQ [](start = 4, length = 9)

nit: You can compare NSNumber instances with `OBJCEQ` instead of querying their primitive values...

tests/unittests/Foundation/NSUndoManagerTests.mm
```diff
@@ -0,0 +1,718 @@
...
+    UndoTestObject* object = [[[UndoTestObject alloc] initWithValue:val1] autorelease];
+    EXPECT_EQ([object.undoManager canUndo], NO);
+
+    // Each undo operation here gets added to the single default undo group.
+    [object updateNumber:val2];
+    [object updateNumber:val3];
+    [object updateNumber:val4];
+    EXPECT_EQ(object.value.intValue, val4.intValue);
+    [object.undoManager undo];
+    EXPECT_EQ(object.value.intValue, val1.intValue);
```
DHowett-MSFT: >EXPECT_EQ(object.value.intValue, val1.intValue); [](start = 4, length = 48)

I am surprised that this undo undoes until 5.

tests/unittests/Foundation/NSUndoManagerTests.mm
```diff
@@ -0,0 +1,718 @@
...
+    EXPECT_EQ([object.undoManager canUndo], NO);
+}
+
+TEST(NSUndoManager, undoRedoTests) {
```
DHowett-MSFT: >undoRedoTests [](start = 20, length = 13)

your test names lost their titlecase at this point.

tests/unittests/Foundation/NSUndoManagerTests.mm
```diff
@@ -0,0 +1,718 @@
...
+    EXPECT_EQ(object.value.intValue, val1.intValue);
+    [object.undoManager redo];
+    EXPECT_EQ(object.value.intValue, val4.intValue);
+    EXPECT_EQ([object.undoManager canRedo], NO);
+    [object.undoManager undo];
+    EXPECT_EQ(object.value.intValue, val1.intValue);
+    EXPECT_EQ([object.undoManager canRedo], YES);
+}
+
+TEST(NSUndoManager, undoRedoTestsWithMutableObject) {
```
DHowett-MSFT: >undoRedoTestsWithMutableObject [](start = 20, length = 30)

I do not know what this is testing. You're using a mutable array in the object?

tests/unittests/Foundation/NSUndoManagerTests.mm
```diff
@@ -0,0 +1,718 @@
...
+    [object.undoManager undo];
+
+    EXPECT_OBJCEQ([NSMutableArray array], object.arrayOfNumbers);
+
+    [object.undoManager redo];
+
+    EXPECT_OBJCEQ(comparisonArray, object.arrayOfNumbers);
+}
+
+TEST(NSUndoManager, ExtraUndoOperations) {
```
DHowett-MSFT: >ExtraUndoOperations [](start = 20, length = 19)

`UndoWithoutAnyUndoableActions`

tests/unittests/Foundation/NSUndoManagerTests.mm
```diff
@@ -0,0 +1,718 @@
...
+    EXPECT_EQ(object.value.intValue, val3.intValue);
+
+    [object.undoManager undoNestedGroup];
+    EXPECT_EQ(object.value.intValue, val2.intValue);
+
+    [object.undoManager undoNestedGroup];
+    EXPECT_EQ(object.value.intValue, val1.intValue);
+}
+
+TEST(NSUndoManager, IndividualUndoOperationsUndoCall) {
```
DHowett-MSFT: >IndividualUndoOperationsUndoCall [](start = 20, length = 32)

This test is identical with the one above it.

tests/unittests/Foundation/NSUndoManagerTests.mm
```diff
@@ -0,0 +1,718 @@
...
+
+// Each update call is wrapped in its own group.
+- (void)updateValue:(NSNumber*)value {
+    [self.undoManager beginUndoGrouping];
+    [self.undoManager registerUndoWithTarget:self selector:@selector(updateValue:) object:self.value];
+    self.value = value;
+    [self.undoManager endUndoGrouping];
+}
+
+- (void)dealloc {
```
DHowett-MSFT: >dealloc [](start = 8, length = 7)

.. such as this one.

tests/unittests/Foundation/NSUndoManagerTests.mm
```diff
@@ -0,0 +1,718 @@
...
+    NSNumber* target1val2 = [NSNumber numberWithInt:16];
+    NSNumber* target1val3 = [NSNumber numberWithInt:17];
+    NSNumber* target2val1 = [NSNumber numberWithInt:25];
+    NSNumber* target2val2 = [NSNumber numberWithInt:26];
+    NSNumber* target2val3 = [NSNumber numberWithInt:27];
+    NSNumber* target3val1 = [NSNumber numberWithInt:35];
+    NSNumber* target3val2 = [NSNumber numberWithInt:36];
+    NSNumber* target3val3 = [NSNumber numberWithInt:37];
+
+    NSUndoManager* manager = [[NSUndoManager alloc] init];
```
DHowett-MSFT: >manager [](start = 19, length = 7)

This needs to be released.

tests/unittests/Foundation/NSUndoManagerTests.mm
```diff
@@ -0,0 +1,718 @@
...
+// Each update call is wrapped in its own group.
+- (void)updateValue:(NSNumber*)value {
+    [self.undoManager beginUndoGrouping];
+    [self.undoManager registerUndoWithTarget:self selector:@selector(updateValue:) object:self.value];
+    self.value = value;
+    [self.undoManager endUndoGrouping];
+}
+
+- (void)dealloc {
+    [_undoManager release];
```
DHowett-MSFT: >_undoManager [](start = 5, length = 12)

**You did not retain this, you cannot release it.**

tests/unittests/Foundation/NSUndoManagerTests.mm
```diff
@@ -0,0 +1,718 @@
...
+@property (retain) NSUndoManager* undoManager;
+@property (retain) NSNumber* value;
+- (id)initWithValue:(NSNumber*)num;
+- (void)updateTarget:(NSNumber*)value;
+@end
+
+@implementation UndoTestObjectWithSharedManager
+- (id)initWithValue:(NSNumber*)num undoManager:(NSUndoManager*)manager {
+    self = [super init];
+    _undoManager = manager;
```
DHowett-MSFT: >manager [](start = 19, length = 7)

**You are releasing this later, but you did not retain it.**

DHowett-MSFT: Heck, it says it's a `retain` property.

---
In reply to: [77098028](https://github.com/Microsoft/WinObjC/pull/900#discussion_r77098028) [](ancestors = 77098028)

tests/unittests/Foundation/NSUndoManagerTests.mm
DHowett-MSFT: >manager [](start = 19, length = 7)

Why not autorelease this?

tests/unittests/Foundation/NSUndoManagerTests.mm
```diff
@@ -0,0 +1,718 @@
...
+    EXPECT_EQ([object.undoManager canUndo], YES);
+    [object.undoManager undo];
+    EXPECT_EQ([object.undoManager canUndo], YES);
+    [object.undoManager undo];
+    EXPECT_EQ([object.undoManager canUndo], YES);
+    [object.undoManager undo];
+    EXPECT_EQ([object.undoManager canUndo], NO);
+
+    // Should do nothing.
+    [object.undoManager undo];
```
DHowett-MSFT: >undoManager [](start = 12, length = 11)

EXPECT_NO_THROW?

tests/unittests/Foundation/NSUndoManagerTests.mm
```diff
@@ -0,0 +1,718 @@
...
+    manager.groupsByEvent = NO;
+    [manager beginUndoGrouping];
+    [manager endUndoGrouping];
+
+    // An empty undo group will remain in the stack even when closed. This is why canUndo reports yes.
+    EXPECT_EQ(manager.canUndo, YES);
+}
+
+// Test that when a target is removed, all undo groups are updated properly. If an undo group remains,
+TEST(NSUndoManager, RemovingOneTarget) {
```
DHowett-MSFT: If an undo group remains, what?

tests/unittests/Foundation/NSUndoManagerTests.mm
MSFTFox: All of these undo calls are part of a single undo group. [manager undo] undoes everything in a single top level group.

---
In reply to: [77097544](https://github.com/Microsoft/WinObjC/pull/900#discussion_r77097544) [](ancestors = 77097544)

MSFTFox: Because groupsByEvent is on by default, a top level undo group is opened and closed automatically.

---
In reply to: [77102647](https://github.com/Microsoft/WinObjC/pull/900#discussion_r77102647) [](ancestors = 77102647,77097544)

tests/unittests/Foundation/NSUndoManagerTests.mm
MSFTFox: Just a slightly different test of using undo operations to mutate an object instead of assigning it.

---
In reply to: [77097663](https://github.com/Microsoft/WinObjC/pull/900#discussion_r77097663) [](ancestors = 77097663)

Frameworks/Foundation/_NSUndoManagerStack.mm
```diff
@@ -0,0 +1,75 @@
...
+    return _stack.size();
+}
+
+- (void)removeAllWithTarget:(id)target {
+    auto iterator = _stack.begin();
+
+    while (iterator != _stack.end()) {
+        if ([*iterator targets:target]) {
+            auto toRemove = iterator;
+            // Incrememting the iterator must happen before removing the object that iterator represents from the list.
```
ms-jihua: nit: typo

Frameworks/Foundation/_NSUndoGroup.mm
```diff
@@ -0,0 +1,106 @@
...
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#import <Starboard.h>
+#import <Foundation/NSUndoManager.h>
+#import "_NSUndoManagerInternal.h"
+
+@implementation _NSUndoGroup {
+    StrongId<_NSUndoManagerStack> _undoObjectsInGroup;
```
ms-jihua: >_undoObjectsInGroup; [](start = 34, length = 20)

nit: this is better, but this seems like it can probably be narrowed down to '_undoablesInGroup'? 'undoObjects' just adds ambiguity (is there an undo object that's NOT an NSUndoable, etc)

ms-jihua:
:shipit:

PR #903, opened 2016-08-23
Rewrite NSInvocation to use architecture-specific argument allocators

DHowett-MSFT:
This pullrequest:
* Removes the arbitrary 16-argument limit on NSInvocation.
* Adds support for methods that take multiple arguments and return structs.
* Fixes `retainArguments` to also copy `char*` arguments (_yeah, the docs say it does this..._).
* Is a proper base implementation for `forwardInvocation:`, which will work with buffers laid out like the `_NSInvocationCallFrame` internal buffer.
* Adds support for properly passing floating point arguments on ARM (previously, they would all be extended to 64-bit and passed on the stack.)
  * Adds support for passing uniform vfp structures in floating point registers, accounting for their use, and spilling arguments onto the stack (if necessary.)
* Is a source of eternal visceral terror virtue of being partially written in assembly.
* Adds some tests that fail on OS X.

What remains to be done:
* Investigation of struct alignment requirements w.r.t the stack on ARM and x86.
* Exceptions must be thrown for methods taking unions or variadic arguments: NSInvocation does not officially support either of those.
  * It looks like the reference platform does this in NSMethodSignature. We can punt this.

This has passed ARM and x86 UTs, as well as rounds of manually sitting around looking at the argument layouts.

The assembly thunks are inspired heavily by those of [libffi](https://github.com/libffi/libffi) (_MIT licensed_), which does all this and more and more than we need.

I hope for @bviglietta to be especially hard on this one :smile:.


DHowett-MSFT:
It's liable that there are testable cases that aren't covered here; please try to think of them.
What we have:
* 1-4 homogenous floats/doubles ad arguments and return values
* Disparate structs as arguments and return values
* Tests with arguments intended to vex the allocator:
  * float/double/float/double (etc.) -- ARM will allocate these out-of-order, and then fill in the holes!
  * char/maxalign/char/maxalign -- maxalign is the largest alignment type; OS X fails this test (?)
  * multiple homogenous VFP structs followed by a stack-spilled float, followed by a struct that will be forced to spill onto the stack as well since it won't fit into registers.

DHowett:
Note to self: calloc the return value, don't malloc it; we want to zero-initialize it for the `forwardInvocation:` case where a receiver may not invoke it.

DHowett:
Optimization: ARM allocator could determine whether VFP is used before allocating the arena to save 64 bytes.

Frameworks/Foundation/_NSInvocation.arm.mm
```diff
@@ -0,0 +1,486 @@
...
+    while (**type != _C_STRUCT_E) {
+        if (**type == '"') {
+            // Occasionally, struct encodings store field names as in {name="x"f"y"f}. Skip those.
+            do {
+                ++(*type);
+            } while ('"' != **type);
+            ++(*type);
+        }
+        char v = **type;
+        switch (v) {
```
aballway: >switch (v) { [](start = 8, length = 12)

why switch here rather than if/else?

DHowett-MSFT:
`impl::~impl` is necessary to delete `_buffer`

MSFTFox:

Update copyright.

---
Refers to: Frameworks/Foundation/NSInvocation.mm:3 in 4043661. [](commit_id = 40436617cda16a9cca350882a75aeb13316c4d3b, deletion_comment = False)

DHowett-MSFT:
@bviglietta is added to the review. #Closed

DHowett-MSFT:

DHowett-MSFT:

Frameworks/Foundation/_NSInvocation.arm.mm
DHowett-MSFT: No particular reason; almost every other type-switch we do us a switch

---
In reply to: [75930104](https://github.com/Microsoft/WinObjC/pull/903#discussion_r75930104) [](ancestors = 75930104)

DHowett-MSFT:
Buffers

DHowett-MSFT:
[Here](https://github.com/DHowett-MSFT/WinObjC/blob/94c9910ecc0d0b0b3a181fa6dde4e53150ce0a46/docs/Foundation/NSInvocation.md) is a markdown preview link for the design document.

rajsesh-msft:
> struct DisparateAggregateDF {

Would be good to have some disparate aggregates with more than just one double and float. 

---
Refers to: tests/unittests/Foundation/NSInvocationTests.mm:84 in 3e6cf87. [](commit_id = 3e6cf87aac665c046d0a18bf735347f6fa3866f8, deletion_comment = False)

docs/Foundation/NSInvocation.md
```diff
@@ -0,0 +1,127 @@
...
+
+1. A conformant implementation of `_NSInvocationCallFrame`.
+2. An assembly thunk that will take the platform internal representation of `_NSInvocationCallFrame` and dispatch a call to a function pointer with the stored arguments.
+
+### X86 details
+
+X86 is the easier of the two currently-supported architectures for which to allocate arguments.
+
+* Every argument is placed on the stack, in stack-ascending order.
+* Arguments less than four bytes in width are sign- or zero-extended and stored in one doubleword and emitted onto the stack.
```
rajsesh-msft: >doubleword [](start = 87, length = 10)

doubleword == 32 bits for x86? it may be better to call out the actual width here to avoid any confusions with Win32 DWORD.

Frameworks/Foundation/NSInvocation.mm
```diff
@@ -14,71 +14,93 @@
...
+        _methodSignature = [methodSignature retain];
+
+        _callFrame.reset(new _NSInvocationCallFrame(methodSignature));
 
-    return [ret autorelease];
+        // true _non-promoted_ return length.
+        // Used only for copying the argument back to the caller.
+        _returnLength = methodSignature.methodReturnLength;
+
+        auto promotedReturnLength = _callFrame->getReturnLength();
```
rajsesh-msft: >auto [](start = 8, length = 4)

nit: not a good use of auto (type is known and easy to mention here).

Frameworks/Foundation/NSInvocation.mm
```diff
@@ -14,71 +14,93 @@
...
+        if (promotedReturnLength > _countof(_smallReturnValueOptimization)) {
```
rajsesh-msft: >_countof [](start = 35, length = 8)

nit: sizeof may be more flexible here in case _smallReturnValueOptimization ever changes to anything other than a byte array.  _countof usage here threw me off a bit.

Frameworks/Foundation/NSInvocation.mm
```diff
@@ -14,71 +14,93 @@
...
+            _returnValue = IwMalloc(promotedReturnLength); // promoted return length
```
rajsesh-msft: >IwMalloc [](start = 27, length = 8)

Do we need any tracing in case we try to optimize perf here later on?

Frameworks/Foundation/NSInvocation.mm
```diff
@@ -160,55 +180,41 @@ - (void)retainArguments {
...
         for (unsigned int i = 0; i < numArgs; i++) {
-            char* type = (char*)[_methodSignature getArgumentTypeAtIndex:i];
-
-            if (args[i] != NULL) {
-                if (strcmp(type, "@") == 0) {
-                    id arg = *((id*)args[i]);
-                    [arg release];
-                }
+            auto type = [_methodSignature getArgumentTypeAtIndex:i];
+            if (type[0] == '@') {
```
rajsesh-msft: >type [](start = 16, length = 4)

These blocks are repeated many times.  Does it make sense to implement a utility method that takes 'action' blocks to act on arguments?  There would be an action for each type (object, string)

docs/Foundation/NSInvocation.md
```diff
@@ -0,0 +1,127 @@
...
+It will then consume the 4 register-width slots, each of which represents one general purpose register from `R0` to `R3`. It does so with `LDMIA ..., {R0-R3}`.
+
+At this point, all that remains in the arena is the stack. The thunk temporarily redirects the stack pointer so that it points here, and calls the function. Every argument
+is precisely where the destination method expects it.
+
+### X86 Dispatching
+
+X86 proves to be the easy case; the stack pointer is directed at the arena, and the function is called.
+
+## Rationale
```
rajsesh-msft: very useful write up, thanks :)

tests/unittests/Foundation/NSInvocationTests.mm
```diff
@@ -137,56 +178,414 @@ @interface NSIT_InvocationTestClass : NSObject
...
+
+    uint32_t arg2 = 0xABABABABUL;
+    uint64_t arg3 = 0x123409871234ABABULL;
+    [invocation setArgument:&arg2 atIndex:2];
+    [invocation setArgument:&arg3 atIndex:3];
+
+    [invocation invoke];
 
     [tester release];
 }
```
rajsesh-msft: We are not testing any retainArguments codepath.

Frameworks/Foundation/NSInvocation.mm
```diff
@@ -92,65 +114,63 @@ - (NSMethodSignature*)methodSignature {
...
 }
 
 /**
  @Status Interoperable
 */
 - (void)retainArguments {
-    if (!retainArguments) {
-        retainArguments = true;
+    if (!_retainArguments) {
+        _retainArguments = true;
```
rajsesh-msft: >true [](start = 27, length = 4)

Nit: YES, not true.

Frameworks/Foundation/NSInvocationInternal.h
```diff
@@ -0,0 +1,54 @@
...
+// It is a _NSInvocationCallFrame's job to carve out allocation extents from a buffer.
+class _NSInvocationCallFrame {
+private:
+	struct impl; // pImpl idiom
+	impl* _impl;
+public:
+	_NSInvocationCallFrame(NSMethodSignature* methodTypeEncoding);
+	~_NSInvocationCallFrame();
+
+    void storeArgument(const void* value, unsigned int index);
```
rajsesh-msft: >storeArgument [](start = 9, length = 13)

super nit: could just be struct impl* _impl;

Frameworks/Foundation/_NSInvocation.x86.mm
```diff
@@ -0,0 +1,238 @@
...
+        _returnType = _getReturnType([_methodSignature methodReturnType]);
+        _returnLength = [_methodSignature methodReturnLength];
+
+        // 1, 2, 4, and 8-byte structs are returned in registers.
+        if (_returnType == RETURN_TYPE_STRUCT) {
+            _stret = true;
+            _stretExtent = _allocateArgument("^v");
+        } else if (_returnType != RETURN_TYPE_NONE) {
+            // Promote all non-stret return lengths to one machine word.
+            _returnLength = std::max(sizeof(uintptr_t), _returnLength);
```
rajsesh-msft: >std::max [](start = 28, length = 8)

seems like we can do without the std::max here.

Frameworks/Foundation/_NSInvocation.x86.mm
```diff
@@ -0,0 +1,238 @@
...
+        }
+
+        auto nArguments = [_methodSignature numberOfArguments];
```
rajsesh-msft: Would be better to use the actual type here.  Also, for loop below should use unsigned int, or will produce warnings..

Frameworks/Foundation/_NSInvocation.x86.mm
```diff
@@ -0,0 +1,238 @@
...
+    size_t _returnLength;
+    return_type _returnType;
+
+    bool _stret;
+    allocation_extent _stretExtent;
+    std::vector<allocation_extent> _allocationExtents;
+};
+
+_NSInvocationCallFrame::_NSInvocationCallFrame(NSMethodSignature* methodSignature)
+    : _impl(new _NSInvocationCallFrame::impl(methodSignature)) {
```
rajsesh-msft: It probably makes sense to consider moving the struct definition to platform specific header and using the impl struct as a member (instead of heap allocated).  This will help with perf/fragmentation.

Frameworks/Foundation/_NSInvocation.x86.asm
```diff
@@ -0,0 +1,129 @@
...
+    je _CASE_RETURN_TYPE_INT32
+    cmp ebx, 2
+    je _CASE_RETURN_TYPE_UINT16
+    cmp ebx, 1
+    je _CASE_RETURN_TYPE_UINT8
+    ; everything else has already been handled by stret out; unwind
+    jmp LOCAL_ret
+
+JUMPTABLE:
+    ; MUST BE KEPT IN THE ORDER FROM NSInvocationInternal.h
```
rajsesh-msft: >NSInvocationInternal [](start = 37, length = 20)

NSInvocation.x86.mm?

rajsesh-msft:
:clock1:

rajsesh-msft:
Didn't review the asm code, hopefully @bviglietta can.

Frameworks/Foundation/_NSInvocation.arm.asm
```diff
@@ -0,0 +1,82 @@
...
+; r1 = frame
+; r2 = fn
+; r3 = vfp count
+    ALIGN
+|_CallFrameInternal_VFP| PROC
+    GLOBAL _CallFrameInternal_VFP
+    GLOBAL _CallFrameInternal
+    CMP r3, 1
+    VLDREQ.64 d0, [r0]
+    VLDMGT.64 r0, {d0-d7}
```
bviglietta: Conditional VFP instructions are deprecated as of ARMv8 and not officially supported on Windows. You should probably prefer to use a branch here.

Frameworks/Foundation/_NSInvocation.arm.asm
```diff
@@ -0,0 +1,82 @@
...
+    MOV ip, #0
+    LDMIA sp!, {r0-r3}
+    BLX lr
+
+    LDR r2, [fp, #8] ; return pointer
+    LDR r3, [fp, #12] ; return type
+    MOV sp, fp
+
+    ADR ip, %F0
+    LDR pc, [ip, r3, lsl #2]
```
bviglietta: >LDR pc, [ip, r3, lsl #2] [](start = 4, length = 24)

Minor optimization: your jump table will be a quarter the size if you use TBB [ip, r3] here.

DHowett-MSFT: That is awesome. Thanks!

---
In reply to: [76449948](https://github.com/Microsoft/WinObjC/pull/903#discussion_r76449948) [](ancestors = 76449948)

Frameworks/Foundation/_NSInvocation.arm.asm
```diff
@@ -0,0 +1,82 @@
...
+;
+;******************************************************************************
+
+    AREA |.text|,ALIGN=4,THUMB,CODE,READONLY
+; r0 = arena start
+; r1 = frame
+; r2 = fn
+; r3 = vfp count
+    ALIGN
+|_CallFrameInternal_VFP| PROC
```
bviglietta: >|_CallFrameInternal_VFP| [](start = 0, length = 24)

Do you expect to be able to propagate exceptions thrown by invocation targets? If so, I expect we'll have to generate .xdata for these functions.

DHowett-MSFT: I would like to, and any advice you have for that would be exceptionally useful :)

---
In reply to: [76459746](https://github.com/Microsoft/WinObjC/pull/903#discussion_r76459746) [](ancestors = 76459746)

DHowett-MSFT: I bet I should be using kxarm.h and the procedure macros.. hm

---
In reply to: [76459820](https://github.com/Microsoft/WinObjC/pull/903#discussion_r76459820) [](ancestors = 76459820,76459746)

docs/Foundation/NSInvocation.md
```diff
@@ -0,0 +1,127 @@
...
+* Non-floating-point arguments less than or equal to four bytes in width are sign- or zero-extended and stored preferentially in registers `R0` through `R3`.
+  * If there are no unallocated registers, these arguments are placed on the stack.
+* Non-floating-point arguments eight bytes in width requiring 8-byte alignment are stored preferentially in `R0,R1` or `R2,R3`.
+  * If these registers are not available, the entire argument is placed on the stack.
+* Single-precision floating point arguments are preferentially allocated to single-precision registers `S0` through `S15`.
+  * If there are no unallocated single-precision registers, these arguments are placed on the stack.
+* Double-precision floating point arguments are preferentially allocated to double-precision registers `D0` through `D7`.
+  * If there are no unallocated double-precision registers, these arguments are aligned to 8 bytes and on the stack.
+  * Each `Dx` register consumes two `Sx` registers. They will be unavailable for floating-point arguments.
+* Aggregates of 1-4 single- or double-precision values are allocated to available floating-point registers.
```
bviglietta: >Aggregates of 1-4 single- or double-precision values are allocated to available floating-point registers. [](start = 2, length = 105)

This also applies to aggregates of vector types, where a single vector can be up to 128 bits long and may contain either floating-point or integer data. I don't know if we're intending to support those types now.

DHowett-MSFT: I don't believe we support them elsewhere right now, but it'll be good to keep in mind. Are vector types vended into q0/q1/q2/q3 even if they contain integer data?

---
In reply to: [76460844](https://github.com/Microsoft/WinObjC/pull/903#discussion_r76460844) [](ancestors = 76460844)

DHowett-MSFT: For what it's worth, objc does not have a type encoding for vector types, IIRC

---
In reply to: [76461957](https://github.com/Microsoft/WinObjC/pull/903#discussion_r76461957) [](ancestors = 76461957,76460844)

DHowett-MSFT: I take it back, gcc specifies one! ``

---
In reply to: [76461989 [](ancestors = 76461989,76461957,76460844)

Frameworks/Foundation/_NSInvocation.x86.asm
```diff
@@ -0,0 +1,129 @@
...
+    mov esp, edx
+
+    mov eax, [ebp + 24]
+    call eax
+
+    mov ecx, [ebp + 20] ; return out pointer
+    mov ebx, [ebp + 12] ; return type
+    shl ebx, 2
+    add ebx, JUMPTABLE
+    mov ebx, [ebx]
```
DHowett-MSFT: jmp JUMPTABLE[ebx*4]

DHowett-MSFT:
We can do this at a later date, if we determine that we are spending too much time or memory on this.

---
In reply to: [241630798](https://github.com/Microsoft/WinObjC/pull/903#issuecomment-241630798) [](ancestors = 241630798)

Frameworks/Foundation/NSInvocation.mm
DHowett-MSFT: I'm worried it'd add a little more overhead than I'm comfortable with; what were you thinking it would look like/

---
In reply to: [76333507](https://github.com/Microsoft/WinObjC/pull/903#discussion_r76333507) [](ancestors = 76333507)

Frameworks/Foundation/_NSInvocation.x86.mm
DHowett-MSFT: The method signature's return length can be less than one register in width, so we need to pin it to at least uintptr_t

---
In reply to: [76335845](https://github.com/Microsoft/WinObjC/pull/903#discussion_r76335845) [](ancestors = 76335845)

DHowett-MSFT:
> struct DisparateAggregateDF {

We do, down below; Large, Small8, Small4; technically, HighlyAligned is also disparate since max_align_t is a double on x86/arm

---
In reply to: [242551185](https://github.com/Microsoft/WinObjC/pull/903#issuecomment-242551185) [](ancestors = 242551185)

---
Refers to: tests/unittests/Foundation/NSInvocationTests.mm:84 in 3e6cf87. [](commit_id = 3e6cf87aac665c046d0a18bf735347f6fa3866f8, deletion_comment = False)

docs/Foundation/NSInvocation.md
bviglietta: Yes, vectors are passed in floating-point registers + the stack regardless of the type of data they contain.

---
In reply to: [76461957](https://github.com/Microsoft/WinObjC/pull/903#discussion_r76461957) [](ancestors = 76461957,76460844)

Frameworks/Foundation/_NSInvocation.arm.asm
bviglietta: Normally, yes, you use those macros: NESTED_ENTRY, NESTED_END, etc., and they take care of emitting the pdata and xdata into the image. With the stack pointer shenanigans going on here, though, it might not be so simple. Without investigating more deeply, I'm not sure exactly what has to be done.

---
In reply to: [76460282](https://github.com/Microsoft/WinObjC/pull/903#discussion_r76460282) [](ancestors = 76460282,76459820,76459746)

Frameworks/Foundation/_NSInvocation.arm.asm
```diff
@@ -0,0 +1,82 @@
...
+    VSTR s0, [r2]
+    POP {fp, pc}
+12 ; VFP_D
+    VSTR d0, [r2]
+    POP {fp, pc}
+13 ; VFP_HOMOGENOUS
+    VSTM r2, {d0-d3}
+    POP {fp, pc}
+
+15 ; INT32
```
bviglietta: >INT32 [](start = 5, length = 5)

INT64

Frameworks/Foundation/_NSInvocation.arm.asm
DHowett-MSFT: Talked with @rajsesh-msft about prioritizing ARM exception handling and stack unwinding. For now, we'll kick it down the road a little bit.

---
In reply to: [76474680](https://github.com/Microsoft/WinObjC/pull/903#discussion_r76474680) [](ancestors = 76474680,76460282,76459820,76459746)

Frameworks/Foundation/NSInvocationInternal.h
```diff
@@ -0,0 +1,54 @@
...
+    void loadArgument(void* value, unsigned int index) const;
+
+	size_t getReturnLength() const;
+	bool getRequiresStructReturn() const;
```
DHowett-MSFT: >	 [](start = 0, length = 1)

There are some tabs in here, somehow.

Frameworks/Foundation/NSInvocation.mm
DHowett-MSFT: Good eye! That makes more sense.

---
In reply to: [76332700](https://github.com/Microsoft/WinObjC/pull/903#discussion_r76332700) [](ancestors = 76332700)

Frameworks/Foundation/NSInvocation.mm
DHowett-MSFT: We may like to consider it; we can emit events based on allocation sizes, and determine what we're hitting most often.

---
In reply to: [76332892](https://github.com/Microsoft/WinObjC/pull/903#discussion_r76332892) [](ancestors = 76332892)

Frameworks/Foundation/_NSInvocation.arm.asm
```diff
@@ -0,0 +1,87 @@
...
+    BLX lr
+
+    LDR r2, [fp, #8] ; return pointer
+    LDR r3, [fp, #12] ; return type
+    MOV sp, fp
+
+    ADR ip, %F101
+    TBB [ip, r3]
+
+100
```
bviglietta: Minor: the code might be a little clearer with text labels instead of numeric labels. JUMPTABLE, CASE_INT32, etc.

bviglietta:
:shipit:

Frameworks/Foundation/_NSInvocation.arm.asm
DHowett-MSFT: @bviglietta I couldn't determine how to do this and maintain the ability to use proc/routine-specific locals :(

DHowett-MSFT:
This refs #711. #Resolved

docs/Foundation/NSInvocation.md
DHowett-MSFT: We already don't support vectors; I'll update this document when we do. Thanks!

---
In reply to: [76474190](https://github.com/Microsoft/WinObjC/pull/903#discussion_r76474190) [](ancestors = 76474190,76461957,76460844)

DHowett-MSFT:
The KVC/KVO test cases hit a missing case, and it's been fixed.

---
In reply to: [241597563](https://github.com/Microsoft/WinObjC/pull/903#issuecomment-241597563) [](ancestors = 241597563)

DHowett-MSFT:
Verified separately on ARM by @chadimoussi. Thanks!

Frameworks/Foundation/NSInvocation.mm
```diff
@@ -217,291 +239,23 @@ - (void)dealloc {
...
-
-    //  structure return type?
-    if (type[0] == '{') {
-        pReturnVal = (char*)IwMalloc(returnSize);
-        totalLength += 4;
-        pMsgFunc = "_objc_msgSend_stret";
+    void* pfn = &objc_msgSend;
+    if (_callFrame->getRequiresStructReturn()) {
+        pfn = &objc_msgSend_stret;
     }
```
jaredhms: nit: light comments here would be nice

Frameworks/Foundation/NSInvocation.mm
```diff
@@ -14,71 +14,99 @@
...
+        [NSException raise:NSInvalidArgumentException format:@"-[NSInvocation initWithMethodSignature:]: method signature must not be nil."];
+        [self release];
+        return nil;
+    }
+
+    if (self = [super init]) {
+        _methodSignature = [methodSignature retain];
 
-    return [ret autorelease];
+        _callFrame.reset(new _NSInvocationCallFrame(methodSignature));
```
jaredhms: >.reset(new _NSInvocationCallFrame(methodSignature)); [](start = 18, length = 52)

nit: _callFrame = make_unique?

Frameworks/Foundation/NSInvocation.mm
```diff
@@ -14,71 +14,99 @@
...
-    int size = objc_sizeof_type(type);
+    std::unique_ptr<_NSInvocationCallFrame> _callFrame;
 
-    void* ret = (void*)IwMalloc(size);
-    memcpy(ret, buf, size);
+    size_t _returnLength;
+    BOOL _retainArguments;
 
-    return ret;
+    uint8_t _smallReturnValueOptimization[16];
```
jaredhms: guaranteed to be aligned properly?

DHowett-MSFT: Always `memcpy`'d in and out bytewise; alignment should not matter here. It looks like for struct returns the compiler emits a `memcpy` before the epilogue as well.

Frameworks/Foundation/NSInvocation.mm
```diff
@@ -92,65 +120,73 @@ - (NSMethodSignature*)methodSignature {
  @Status Interoperable
 */
 - (void)setArgument:(void*)buf atIndex:(int)index {
-    if ((index < 0) || (index >= [_methodSignature numberOfArguments]) || (index >= MAX_ARGS)) {
-        TraceVerbose(TAG, L"index = %d, MAX_ARGS = %d, MethodSig arguments = %d", index, MAX_ARGS, [_methodSignature numberOfArguments]);
+    if ((index < 0) || (index >= [_methodSignature numberOfArguments])) {
+        TraceVerbose(TAG, L"index = %d, MethodSig arguments = %d", index, [_methodSignature numberOfArguments]);
         [NSException raise:NSInvalidArgumentException format:@"The number of arguments exceeds the allowed limit."];
```
jaredhms: >exceeds [](start = 87, length = 7)

super-nit: or it's negative

Frameworks/Foundation/NSInvocation.mm
```diff
@@ -14,71 +14,99 @@
...
+#include <vector>
+#include <memory>
+
 static const wchar_t* TAG = L"NSInvocation";
 
-@implementation NSInvocation
-static void* copyArgument(NSInvocation* self, void* buf, int index) {
-    char* type = (char*)[self->_methodSignature getArgumentTypeAtIndex:index];
+@implementation NSInvocation {
+    NSMethodSignature* _methodSignature;
```
jaredhms: >NSMethodSignature [](start = 4, length = 17)

why not StrongId so we don't have to retain/release?

Frameworks/Foundation/_NSInvocation.x86.h
```diff
@@ -0,0 +1,53 @@
...
+#import <Foundation/NSMethodSignature.h>
+
+#import <cstdint>
+#import <algorithm>
+#import <vector>
+
+// It is a _NSInvocationCallFrame's job to carve out allocation extents from a buffer.
+class _NSInvocationCallFrame {
+private:
+    StrongId<NSMethodSignature> _methodSignature;
```
jaredhms: >StrongId<NSMethodSignature> [](start = 4, length = 27)

nit: ...whereas an extra retain/release on this copy probably isn't necessary since _NSInvocationCallFrame can never outlive its parent NSInvocation

Frameworks/Foundation/_NSInvocation.x86.h
```diff
@@ -0,0 +1,53 @@
...
+
+    uint8_t* _buffer;
+    off_t _offset;
+    size_t _returnLength;
+    unsigned int _returnType; /* platform-specific/opaque */
+
+    bool _stret;
+    allocation_extent _stretExtent;
```
jaredhms: nit: I realize that 'stret' is documented *somewhere* in obcjmsgsend docs, but can we use more descriptive variable names here and elsewhere?

Frameworks/Foundation/NSInvocationInternal.h
```diff
@@ -0,0 +1,33 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#pragma once
+
+struct allocation_extent {
```
jaredhms: >allocation_extent [](start = 7, length = 17)

nit: inconsistent/non-conformant struct name

Frameworks/Foundation/NSInvocationInternal.h
```diff
@@ -0,0 +1,33 @@
...
+    off_t offset;
```
jaredhms: >off_t [](start = 4, length = 5)

nit: interesting - i've never used off_t before.  the docs say it's for 'file offsets', but i suppose the type is the same....i'd prob just use size_t

DHowett-MSFT: Actually, `ptrdiff_t` may be better still

Frameworks/Foundation/_NSInvocation.x86.mm
```diff
@@ -0,0 +1,207 @@
...
+    if (_returnType == RETURN_TYPE_STRUCT) {
+        _structReturn = true;
+        _structReturnExtent = _allocateArgument("^v");
+    } else if (_returnType != RETURN_TYPE_NONE) {
+        // Promote all non-stret return lengths to one machine word.
+        _returnLength = std::max(sizeof(uintptr_t), _returnLength);
+    }
+
+    unsigned int nArguments = [_methodSignature numberOfArguments];
+    for (unsigned int i = 0; i < nArguments; ++i) {
```
jaredhms: >nArguments [](start = 33, length = 10)

nit: no hungarian

jaredhms: (in a couple spots)

---
In reply to: [77061000](https://github.com/Microsoft/WinObjC/pull/903#discussion_r77061000) [](ancestors = 77061000)

jaredhms:
:shipit:

DHowett-MSFT:
Nit: what determined whether we included the umbrella header `Foundation.h` or the class header `NSObject.h`?

DHowett-MSFT:
Test failure post-mortem: The test for arguments above limit was expecting `NSInvalidArgumentException` instead of `NSRangeException`.

PR #911, opened 2016-08-25
Create file for NSData write tests and NSFileHandle WritingAtPath test

### LINKED ISSUE #899: NSData.WriteToURL, NSFileHandle.WritingAtPath fail on OSX
Missed these two when fixing the file-related OSX test issues.

[ RUN      ] NSData.WriteToURL
NSDataTests.mm:192: Failure
Value of: [expectedData writeToURL:fileURL options:0 error:nullptr]
  Actual: false
Expected: true
NSDataWriteToURL should succeed
[  FAILED  ] NSData.WriteToURL (2 ms)

[ RUN      ] NSFileHandle.WritingAtPath
NSFileHandleTests.mm:257: Failure
Value of: fh != nil
  Actual: false
Expected: true
[  FAILED  ] NSFileHandle.WritingAtPath (2 ms)

aballway:
Our platform is more forgiving when writing to files, creating them when they do not exist rather than failing, so these tests were failing on the reference platform because the files were not created beforehand.

Fixes #899

### DIFF
--- a/Frameworks/Foundation/NSData.mm
+++ b/Frameworks/Foundation/NSData.mm
@@ -1,6 +1,6 @@
 //******************************************************************************
 //
-// Copyright (c) 2015 Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
 //
 // This code is licensed under the MIT License (MIT).
 //
--- a/Frameworks/Foundation/NSData.mm
+++ b/Frameworks/Foundation/NSData.mm
@@ -352,6 +352,7 @@ BASE_CLASS_REQUIRED_IMPLS(NSData, NSDataPrototype, CFDataGetTypeID);
         return TRUE;
     } else {
         TraceVerbose(TAG, L"NSData couldn't open %hs for write (with options)", fname);
+        *error = [NSError errorWithDomain:NSCocoaErrorDomain code:4 userInfo:@{ NSFilePathErrorKey : filename }];
         return FALSE;
     }
 }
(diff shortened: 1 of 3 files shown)


rajsesh-msft:

Nit: update to new copyright banner.

---
Refers to: tests/unittests/Foundation/NSFileHandleTests.mm:3 in de40cc4. [](commit_id = de40cc4f14af0b6f71d0952f9d6dcd4953304dfc, deletion_comment = False)

rajsesh-msft:
:shipit:

DHowett-MSFT:
`-[NSData write...]` should definitely be creating files that do not exist; do you mean directories?

DHowett-MSFT:
:shipit:

DHowett-MSFT:
If it _is_ directories: technically, we should not be creating the directories (and instead, throw a file not found/illegal/etc. exception.)

aballway:
somehow I am now able to create files with it on OSX where I was not before, but it certainly cannot create directories.  Will change to give error when trying to do so.  I have no idea why it wasn't working for me before, I tried it quite a bit

---
In reply to: [242566261](https://github.com/Microsoft/WinObjC/pull/911#issuecomment-242566261) [](ancestors = 242566261)

bbowman:
🕐

aballway:
Other than not creating the error on failure, our implementation functions the same as the reference platform.  The failure was due to the write expecting to run in a directory containing Library, which does not seem to happen on the reference platform.  I'm still not sure why I wasn't able to create files earlier when testing.

Frameworks/Foundation/NSData.mm
```diff
@@ -352,6 +352,7 @@ - (BOOL)writeToFile:(NSString*)filename options:(NSDataWritingOptions)options er
         return TRUE;
     } else {
         TraceVerbose(TAG, L"NSData couldn't open %hs for write (with options)", fname);
+        *error = [NSError errorWithDomain:NSCocoaErrorDomain code:4 userInfo:@{ NSFilePathErrorKey : filename }];
```
MSFTFox: Check if error is nil before using it.

PR #927, opened 2016-08-26
Revert NSMutableCharacterSet test changes.

### LINKED ISSUE #897: NSCharacterSet.ExceptionTests fails on OSX
This was a new test added recently.

[ RUN      ] NSCharacterSet.ExceptionTests
NSCharacterSetTests.m:228: Failure
Failed
[  FAILED  ] NSCharacterSet.ExceptionTests (0 ms)

MSFTFox:
There was a misunderstanding that the expected nonnull argument to
two factory methods in NSCharacterSet should be enforced by
throwing an exception. This is not correct. 

One test case has still been removed. set = [NSMutableCharacterSet characterSetWithContentsOfFile:nil]; is not a valid statement as searching for a nil file will throw an exception.

Fix #897

### DIFF
--- a/tests/unittests/Foundation/NSCharacterSetTests.m
+++ b/tests/unittests/Foundation/NSCharacterSetTests.m
@@ -1,6 +1,6 @@
 //******************************************************************************
 //
-// Copyright (c) 2015 Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
 //
 // This code is licensed under the MIT License (MIT).
 //
--- a/tests/unittests/Foundation/NSCharacterSetTests.m
+++ b/tests/unittests/Foundation/NSCharacterSetTests.m
@@ -219,22 +219,7 @@ TEST(NSCharacterSet, Polymorphic_Creators) {
 
     set = [NSMutableCharacterSet characterSetWithRange:NSMakeRange(0, 32)];
     EXPECT_TRUE([set isKindOfClass:[NSMutableCharacterSet class]]);
-}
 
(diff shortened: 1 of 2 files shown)


rajsesh-msft:
:shipit:

aballway:

nit

---
Refers to: tests/unittests/Foundation/NSCharacterSetTests.m:3 in f6f0cba. [](commit_id = f6f0cbabbca8cf883186864c956bbfd5ac4e51c9, deletion_comment = False)

tests/unittests/Foundation/NSCharacterSetTests.m
```diff
@@ -219,22 +219,7 @@ void testCharacter(NSCharacterSet* charSet, NSString* testString) {
...
-    @try {
-        set = [NSMutableCharacterSet characterSetWithBitmapRepresentation:nil];
-        FAIL();
-    } @catch (NSException* e) {
-        ASSERT_OBJCEQ(e.name, @"NSInvalidArgumentException");
-    }
 
-    @
-    try {
-        set = [NSMutableCharacterSet characterSetWithContentsOfFile:nil];
```
aballway: >set = [NSMutableCharacterSet characterSetWithContentsOfFile:nil]; [](start = 8, length = 65)

can we test this as well?

MSFTFox: It merely calls characterSetWithBitmap but with the contents of the file. It would merely be testing file lookup. As characterSetWithBitmap isn't tested either, there's no point in exploring this function before it gets any actual attention. A real test needs to be developed for this function which is not the priority of this pull request.

DHowett-MSFT:
:shipit:

PR #935, opened 2016-08-29
AddressBook Framework -- Modifying, Adding, and Removing Contacts

### LINKED ISSUE #888: Addressbook - modify contacts and add/remove to addressbook
Implement the following APIs:

- [ ] ABRecordSetValue
- [ ] ABPersonCreate
- [ ] ABRecordRemoveValue
- [ ] ABAddressBookRemoveRecord
- [ ] ABAddressBookAddRecord
- [ ] ABAddressBookHasUnsavedChanges
- [ ] ABAddressBookSave
- [ ] ABAddressBookRevert

chadimoussi:
Fixes #888
Added support for modifying a contact, as well as adding, removing,
and saving the AddressBook. Added new method to retrieve contacts
that can be modified or deleted.

### DIFF
--- a/Frameworks/AddressBook/ABAddressBook.mm
+++ b/Frameworks/AddressBook/ABAddressBook.mm
@@ -104,47 +104,44 @@ void ABAddressBookRequestAccessWithCompletion(ABAddressBookRef addressBook, ABAd
 }
 
 /**
- @Status Stub
+ @Status Interoperable
  @Notes
 */
 bool ABAddressBookHasUnsavedChanges(ABAddressBookRef addressBook) {
-    UNIMPLEMENTED();
-    return StubReturn();
+    return [(__bridge _ABAddressBookManager*)addressBook hasUnsavedChanges];
 }
 
 /**
- @Status Stub
+ @Status Interoperable
  @Notes
 */
 bool ABAddressBookSave(ABAddressBookRef addressBook, CFErrorRef* error) {
-    UNIMPLEMENTED();
-    return StubReturn();
+    return [(__bridge _ABAddressBookManager*)addressBook save];
 }
 
 /**
- @Status Stub
+ @Status Interoperable
  @Notes
 */
 void ABAddressBookRevert(ABAddressBookRef addressBook) {
-    UNIMPLEMENTED();
+    [(__bridge _ABAddressBookManager*)addressBook revert];
 }
 
 /**
- @Status Stub
+ @Status Interoperable
  @Notes
 */
 bool ABAddressBookAddRecord(ABAddressBookRef addressBook, ABRecordRef record, CFErrorRef* error) {
-    UNIMPLEMENTED();
-    return StubReturn();
+    return [(__bridge _ABAddressBookManager*)addressBook addContact:record error:error];
 }
 
 /**
- @Status Stub
- @Notes
+ @Status Caveat
+ @Notes Records can only be removed if they are retrieved with ABAddressBookCopyArrayOfAllUserAppPeople
+ -- contacts from ABAddressBookCopyArrayOfAllPeople are read-only and cannot be removed.
 */
 bool ABAddressBookRemoveRecord(ABAddressBookRef addressBook, ABRecordRef record, CFErrorRef* error) {
-    UNIMPLEMENTED();
-    return StubReturn();
+    return [(__bridge _ABAddressBookManager*)addressBook removeContact:record error:error];
 }
 
 /**
--- a/Frameworks/AddressBook/ABAddressBook.mm
+++ b/Frameworks/AddressBook/ABAddressBook.mm
@@ -1,6 +1,6 @@
 //******************************************************************************
 //
-// Copyright (c) 2016 Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
 //
 // This code is licensed under the MIT License (MIT).
 //
@@ -116,7 +116,16 @@ bool ABAddressBookHasUnsavedChanges(ABAddressBookRef addressBook) {
  @Notes
 */
 bool ABAddressBookSave(ABAddressBookRef addressBook, CFErrorRef* error) {
-    return [(__bridge _ABAddressBookManager*)addressBook save];
+    bool result = [(__bridge _ABAddressBookManager*)addressBook save];
+    if ((!result) && error) {
+        NSDictionary* userInfo = @{
+            NSLocalizedDescriptionKey : NSLocalizedString(@"Error saving AddressBook.\n", nil),
+        };
+        *error = (__bridge_retained CFErrorRef)[NSError errorWithDomain:(__bridge NSString*)ABAddressBookErrorDomain
+                                                                   code:kABOperationNotPermittedByStoreError
+                                                               userInfo:userInfo];
+    }
+    return result;
 }
 
 /**
--- a/Frameworks/AddressBook/ABAddressBookManagerInternal.h
+++ b/Frameworks/AddressBook/ABAddressBookManagerInternal.h
@@ -17,10 +17,17 @@
 #pragma once
 
 #import "UWP/WindowsApplicationModelContacts.h"
+#import <AddressBook/ABRecord.h>
 
 @interface _ABAddressBookManager : NSObject
 
 @property WACContactStore* contactStore;
 - (NSArray*)getListOfContacts;
-
+- (NSArray*)getListOfModifiableContacts;
+- (bool)addContact:(ABRecordRef)record error:(CFErrorRef*)error;
+- (bool)removeContact:(ABRecordRef)record error:(CFErrorRef*)error;
+- (void)modifyContact:(ABRecordRef)record;
+- (bool)hasUnsavedChanges;
+- (void)revert;
+- (bool)save;
 @end
--- a/Frameworks/AddressBook/ABAddressBookManagerInternal.h
+++ b/Frameworks/AddressBook/ABAddressBookManagerInternal.h
@@ -1,6 +1,6 @@
 //******************************************************************************
 //
-// Copyright (c) 2016 Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
 //
 // This code is licensed under the MIT License (MIT).
 //
@@ -30,4 +30,12 @@
 - (bool)hasUnsavedChanges;
 - (void)revert;
 - (bool)save;
+@end
+
+@interface __ABContactOperation : NSObject
+
+@property BOOL shouldDelete;
+@property WACContact* contact;
+- (id)initWithContact:(WACContact*)contact shouldDelete:(BOOL)shouldDelete;
+
 @end
--- a/Frameworks/AddressBook/ABAddressBookManagerInternal.mm
+++ b/Frameworks/AddressBook/ABAddressBookManagerInternal.mm
@@ -43,6 +43,17 @@ static void _setError(CFErrorRef* error, NSString* message) {
     }
 }
 
+static void _copyWrappedContacts(NSArray* /* WACContact* */ source,
+                                 NSMutableArray* /* _ABContact* */ destination,
+                                 ABRecordContactType type,
+                                 ABRecordRef manager) {
+    for (WACContact* contact in source) {
+        _ABContact* wrappedContact = [[_ABContact alloc] initWithContact:contact andType:type];
+        wrappedContact.manager = manager;
+        [destination addObject:wrappedContact];
+    }
+}
+
 @implementation _ABAddressBookManager {
     NSMutableSet<WACContact*>* _toAdd;
     NSMutableDictionary<NSString*, __ABContactOperation*>* _operations;
@@ -169,11 +180,7 @@ static void _setError(CFErrorRef* error, NSString* message) {
         result = [NSMutableArray arrayWithCapacity:[success count]];
 
         // Copy over the contacts wrapped in _ABContacts.
-        for (int i = 0; i < [success count]; i++) {
-            _ABContact* contact = [[_ABContact alloc] initWithContact:success[i] andType:kAddressBookReadOnlyContact];
-            contact.manager = (__bridge ABRecordRef)self;
-            result[i] = contact;
-        }
+        _copyWrappedContacts(success, result, kAddressBookReadOnlyContact, (__bridge ABRecordRef)self);
 
         dispatch_semaphore_signal(semaphore);
     }
@@ -202,11 +209,7 @@ static void _setError(CFErrorRef* error, NSString* message) {
                 shouldContinue = NO;
             } else {
                 shouldContinue = YES;
-                for (WACContact* contact in success.contacts) {
-                    _ABContact* wrappedContact = [[_ABContact alloc] initWithContact:contact andType:kAddressBookReadWriteContact];
-                    wrappedContact.manager = (__bridge ABRecordRef)self;
-                    [result addObject:wrappedContact];
-                }
+                _copyWrappedContacts(success.contacts, result, kAddressBookReadWriteContact, (__bridge ABRecordRef)self);
             }
             dispatch_semaphore_signal(semaphore);
         }
@@ -253,7 +256,7 @@ static void _setError(CFErrorRef* error, NSString* message) {
         }
         case kAddressBookReadOnlyContact:
         default:
-            _setError(error, @"This contact already exists in the AddressBook.\n");
+            _setError(error, @"This contact cannot be added to the AddressBook.\n");
             return false;
     }
 }
--- a/Frameworks/AddressBook/ABAddressBookManagerInternal.mm
+++ b/Frameworks/AddressBook/ABAddressBookManagerInternal.mm
@@ -18,19 +18,52 @@
 #import "ABContactInternal.h"
 #import "UWP/WindowsApplicationModelContacts.h"
 
-@interface _ABAddressBookManager ()
+@interface __ABContactOperation : NSObject
+
+@property BOOL shouldDelete;
+@property WACContact* contact;
+
+- (id)initWithContact:(WACContact*)contact shouldDelete:(BOOL)shouldDelete;
 
 @end
 
-@implementation _ABAddressBookManager
+@implementation __ABContactOperation
+
+- (id)initWithContact:(WACContact*)contact shouldDelete:(BOOL)shouldDelete {
+    if (self = [super init]) {
+        self.shouldDelete = shouldDelete;
+        self.contact = contact;
+    }
+
+    return self;
+}
+
+@end
+
+@interface _ABAddressBookManager ()
+
+- (void)_setError:(CFErrorRef*)error message:(NSString*)message;
+
+@end
+
+@implementation _ABAddressBookManager {
+    NSMutableSet<WACContact*>* _toAdd;
+    NSMutableDictionary<NSString*, __ABContactOperation*>* _operations;
+    WACContactStore* _writableStore;
+    WACContactList* _contactList;
+}
 
 - (id)init {
     self = [super init];
     if (self) {
+        self->_toAdd = [[NSMutableSet alloc] init];
+        self->_operations = [[NSMutableDictionary alloc] init];
+
         __block WACContactStore* result = nil;
         __block BOOL failed = YES;
         dispatch_semaphore_t semaphore = dispatch_semaphore_create(0);
 
+        // Get the contact store for contacts that are read-only.
         [WACContactManager requestStoreAsyncWithAccessType:WACContactStoreAccessTypeAllContactsReadOnly
             success:^(WACContactStore* success) {
                 result = success;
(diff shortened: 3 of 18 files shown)


Frameworks/AddressBook/ABContactInternal.mm
```diff
@@ -15,14 +15,17 @@
 //******************************************************************************
 
 #import "ABContactInternal.h"
+#import "ABAddressBookManagerInternal.h"
 #import "UWP/WindowsApplicationModelContacts.h"
 
 @implementation _ABContact
 
-- (id)initWithContact:(WACContact*)contact {
+- (id)initWithContact:(WACContact*)contact andType:(ABRecordContactType)type {
```
aballway: >andType [](start = 43, length = 7)

nit: just type?

Frameworks/AddressBook/ABRecord.mm
```diff
@@ -80,18 +150,458 @@ ABRecordType ABRecordGetRecordType(ABRecordRef record) {
...
+            *error = (__bridge_retained CFErrorRef)[NSError errorWithDomain:(__bridge NSString*)ABAddressBookErrorDomain
+                                                                       code:kABOperationNotPermittedByStoreError
+                                                                   userInfo:userInfo];
+        }
+
+        return false;
+    }
+
+    // Cases for various name properties.
+    if (contactProperty == kABPersonFirstNameProperty) {
```
aballway: > if (contactProperty == kABPersonFirstNameProperty) { [](start = 3, length = 53)

nit: this might be better as a switch

Frameworks/AddressBook/ABRecord.mm
```diff
@@ -80,18 +150,458 @@ ABRecordType ABRecordGetRecordType(ABRecordRef record) {
...
+
+        NSMutableArray* jobInfo = person.contact.jobInfo;
+        if ([jobInfo count] == 0) {
+            [jobInfo addObject:[WACContactJobInfo make]];
+        }
+
+        WACContactJobInfo* job = jobInfo[0];
+        job.title = jobTitle;
+        _updateManager(person);
+        return true;
```
aballway: nit: can just do this at the end of the method rather than in every if{}

tests/unittests/AddressBook/ModifyContactsTests.mm
```diff
@@ -0,0 +1,241 @@
...
+
+    ASSERT_TRUE(ABRecordSetValue(_record, property, multiValue, NULL));
+    ABMultiValueRef result = ABRecordCopyValue(_record, property);
+
+    for (int i = 0; i < [labels count]; i++) {
+        NSString* expectedLabel = (__bridge NSString*)ABMultiValueCopyLabelAtIndex(multiValue, i);
+        NSString* actualLabel = (__bridge NSString*)ABMultiValueCopyLabelAtIndex(result, i);
+        NSString* expectedValue = (__bridge NSString*)ABMultiValueCopyValueAtIndex(multiValue, i);
+        NSString* actualValue = (__bridge NSString*)ABMultiValueCopyValueAtIndex(result, i);
+        ASSERT_OBJCEQ(expectedValue, actualValue);
```
aballway: >ASSERT_OBJCEQ [](start = 8, length = 13)

should make these EXPECT

tests/unittests/AddressBook/ModifyContactsTests.mm
```diff
@@ -0,0 +1,241 @@
...
+    dateComponents2.month = 6;
+    dateComponents2.year = 1935;
+    NSDate* date2 = [calendar dateFromComponents:dateComponents2];
+    [dateComponents1 release];
+    [dateComponents2 release];
+
+    ABMultiValueAddValueAndLabel(multiValue, (__bridge CFDateRef)date1, (__bridge CFStringRef) @"hi", NULL);
+    ABMultiValueAddValueAndLabel(multiValue, (__bridge CFDateRef)date2, (__bridge CFStringRef) @"yo", NULL);
+
+    ASSERT_TRUE(ABRecordSetValue(person, kABPersonDateProperty, multiValue, NULL));
```
aballway: >NULL [](start = 76, length = 4)

should have an error here to test against

MSFTFox:
> //

Update copyright: // Copyright (c) Microsoft. All rights reserved.

---
Refers to: Frameworks/AddressBook/ABAddressBook.mm:4 in 41f8a17. [](commit_id = 41f8a177377ccfd2914fb91c95ba4e02ce69d7e9, deletion_comment = False)

Frameworks/AddressBook/ABAddressBook.mm
```diff
@@ -104,47 +104,44 @@ void ABAddressBookRequestAccessWithCompletion(ABAddressBookRef addressBook, ABAd
...
-    return StubReturn();
+    return [(__bridge _ABAddressBookManager*)addressBook hasUnsavedChanges];
 }
 
 /**
- @Status Stub
+ @Status Interoperable
  @Notes
 */
 bool ABAddressBookSave(ABAddressBookRef addressBook, CFErrorRef* error) {
```
MSFTFox: >CFErrorRef* error [](start = 53, length = 17)

@Status Caveat? Seems like error is unused.

Frameworks/AddressBook/ABAddressBookManagerInternal.mm
```diff
@@ -18,19 +18,52 @@
 #import "ABContactInternal.h"
 #import "UWP/WindowsApplicationModelContacts.h"
 
+@interface __ABContactOperation : NSObject
```
MSFTFox: >__ABContactOperation [](start = 11, length = 20)

You should keep all interfaces in your header file. Especially for internal classes.

Frameworks/AddressBook/ABAddressBookManagerInternal.mm
```diff
@@ -18,19 +18,52 @@
...
+    NSMutableSet<WACContact*>* _toAdd;
+    NSMutableDictionary<NSString*, __ABContactOperation*>* _operations;
+    WACContactStore* _writableStore;
+    WACContactList* _contactList;
+}
 
 - (id)init {
     self = [super init];
     if (self) {
+        self->_toAdd = [[NSMutableSet alloc] init];
```
MSFTFox: >self [](start = 8, length = 4)

nit: "self->" is unnecessary.

Frameworks/AddressBook/ABAddressBookManagerInternal.mm
```diff
@@ -18,19 +18,52 @@
...
+- (void)_setError:(CFErrorRef*)error message:(NSString*)message;
+
 @end
 
-@implementation _ABAddressBookManager
+@implementation _ABAddressBookManager {
+    NSMutableSet<WACContact*>* _toAdd;
+    NSMutableDictionary<NSString*, __ABContactOperation*>* _operations;
+    WACContactStore* _writableStore;
+    WACContactList* _contactList;
```
MSFTFox: It seems like you could be leaking some of these objects. I would recommend using StrongId for the NS* objects and double check WAC* objects.

Frameworks/AddressBook/ABAddressBookManagerInternal.mm
```diff
@@ -18,19 +18,52 @@
...
+    }
+
+    return self;
+}
+
+@end
+
 @interface _ABAddressBookManager ()
 
+- (void)_setError:(CFErrorRef*)error message:(NSString*)message;
```
MSFTFox: >_setError [](start = 8, length = 9)

Just curious why this method is declared here instead of in the header file's interface declaration.

Frameworks/AddressBook/ABAddressBookManagerInternal.mm
```diff
@@ -82,4 +192,183 @@ - (NSArray*)getListOfContacts {
...
+    return (self->_toAdd).count + (self->_operations).count > 0;
+}
+
+- (void)revert {
+    [self->_toAdd removeAllObjects];
+    [self->_operations removeAllObjects];
+}
+
+- (bool)save {
+    // TODO: When the projections for WACContactList's saveContactAsync and deleteContactAsync
```
MSFTFox: >TODO [](start = 7, length = 4)

Is this TODO associated with another task? If so, place the task's issue ID here for proper tracking. If not, please create one. :)

Frameworks/AddressBook/ABContactInternal.mm
```diff
@@ -15,14 +15,17 @@
...
     self = [super init];
     if (self) {
```
MSFTFox: nit: condense to if(self = [super init]) {

Frameworks/AddressBook/ABContactInternal.mm
```diff
@@ -15,14 +15,17 @@
...
         self.contact = contact;
+        self.manager = nil;
```
MSFTFox: Don't call messages on a currently initializing object. Use the ivars directly to avoid bad messages. Setters and getters should not be used in init and dealloc methods.

Frameworks/AddressBook/ABRecord.mm
```diff
@@ -20,12 +20,82 @@
...
+// contact is modified to ensure changes persist if
+// the address book is saved.
+static void _updateManager(_ABContact* person) {
+    if (person.type == kAddressBookReadWriteContact) {
+        [(__bridge _ABAddressBookManager*)person.manager modifyContact:(__bridge ABRecordRef)person];
+    }
+}
+
+// The maximum allowed String length of various Windows Contacts properties.
+static const NSUInteger kABPersonFirstNameLength = 64;
```
MSFTFox: Is there documentation of this? It might be helpful to link to where you got these numbers so anybody looking at this can understand why these limits are set.

MSFTFox:
:clock1:

Frameworks/AddressBook/ABAddressBookManagerInternal.mm
DHowett-MSFT: AddressBook is an ARC-enabled framework

---
In reply to: [76673982](https://github.com/Microsoft/WinObjC/pull/935#discussion_r76673982) [](ancestors = 76673982)

Frameworks/AddressBook/ABRecord.mm
DHowett-MSFT: Since they're exported constant symbols, not defines or enums, they cannot be case labels :(

---
In reply to: [76669898](https://github.com/Microsoft/WinObjC/pull/935#discussion_r76669898) [](ancestors = 76669898)

Frameworks/AddressBook/ABRecord.mm
chadimoussi: The default case is to return false, which is at the end of the method

---
In reply to: [76670284](https://github.com/Microsoft/WinObjC/pull/935#discussion_r76670284) [](ancestors = 76670284)

Frameworks/AddressBook/ABAddressBookManagerInternal.mm
chadimoussi: I put it here since I only wanted _ABAddressBookManager to know of this class -- which header should I move it to? _ABAddressBookManager's header, or should I create its own separate header?

---
In reply to: [76672379](https://github.com/Microsoft/WinObjC/pull/935#discussion_r76672379) [](ancestors = 76672379)

Frameworks/AddressBook/ABAddressBookManagerInternal.mm
chadimoussi: It is a private method that was designed to only be used internally to this class (the rest of the methods are called from ABAddressBook.mm and such, but don't need to call this one)

---
In reply to: [76674447](https://github.com/Microsoft/WinObjC/pull/935#discussion_r76674447) [](ancestors = 76674447)

Frameworks/AddressBook/ABAddressBookManagerInternal.mm
MSFTFox: Well it seems like a lot of the classes are internal only here. The pattern I've been following has been that any private classes can go in the internal header for the file that needs to know of it.

---
In reply to: [76687605](https://github.com/Microsoft/WinObjC/pull/935#discussion_r76687605) [](ancestors = 76687605,76672379)

Frameworks/AddressBook/ABAddressBookManagerInternal.mm
MSFTFox: Since it's a private helper function you can make it a normal function call instead of an objective c message. This will avoid a dynamic look up.

---
In reply to: [76687800](https://github.com/Microsoft/WinObjC/pull/935#discussion_r76687800) [](ancestors = 76687800,76674447)

Frameworks/AddressBook/ABAddressBookManagerInternal.mm
MSFTFox: Well, you use 'self->' in quite a few places. In my opinion, it should only be used when the name of the ivar and the name of what it's being set to are significantly similar.

---
In reply to: [76672822](https://github.com/Microsoft/WinObjC/pull/935#discussion_r76672822) [](ancestors = 76672822)

Frameworks/AddressBook/ABRecord.mm
chadimoussi: Updated to avoid multiple return true's

---
In reply to: [76670284](https://github.com/Microsoft/WinObjC/pull/935#discussion_r76670284) [](ancestors = 76670284)

tests/testapps/AddressBookSample/AddressBook/ViewAllContactsViewController.m
```diff
@@ -41,10 +41,30 @@ - (void)viewDidLoad {
 - (void)viewWillAppear:(BOOL)animated {
     [super viewWillAppear:animated];
 
+#ifdef WINOBJC
```
ms-jihua: Can you comment here why these ifdefs are necessary between platforms? From what I can see the split might be justified, but I'm usually suspicious of platform divergences, especially in test.

chadimoussi: I will document this in the code as well, but it's because we added a new method (ABAddressBookCopyArrayOfAllUserAppPeople) to get modifiable/delete-able contacts due to the Windows permissions issue (but this method doesn't exist on the iOS side, so the ifdef is so the code can still work on iOS)

---
In reply to: [77093056](https://github.com/Microsoft/WinObjC/pull/935#discussion_r77093056) [](ancestors = 77093056)

tests/unittests/AddressBook/ModifyContactsTests.mm
ms-jihua: I'm personally fine with not using an error out-value here, but it bothers me that there aren't test cases here where you do get an error out (what would be an example of such a case?)

---
In reply to: [76670902](https://github.com/Microsoft/WinObjC/pull/935#discussion_r76670902) [](ancestors = 76670902)

chadimoussi: Will add cases for this, but errors would occur if a String property's length was longer than the allowed Windows length, or if an unknown property was attempted to be set

---
In reply to: [77093837](https://github.com/Microsoft/WinObjC/pull/935#discussion_r77093837) [](ancestors = 77093837,76670902)

Frameworks/AddressBook/ABAddressBookManagerInternal.h
```diff
@@ -17,10 +17,25 @@
 #pragma once
 
 #import "UWP/WindowsApplicationModelContacts.h"
+#import <AddressBook/ABRecord.h>
 
 @interface _ABAddressBookManager : NSObject
```
ms-jihua: now that this interface actually has functions, it may be useful to declare them here so that it's easy to see what the interface does at a glance

ms-jihua: this is not a good name for a class. what does a 'manager' do, in particular?

(consider just changing it to _ABAddressBook? it seems to serve a similar purpose as your previous _ classes, and the 'manager' bit adds unnecessary ambiguity, unless there's some exception to the pattern that I'm not seeing)

chadimoussi: The (roughly) equivalent Windows class is ContactManager (Also, one of the first code reviews for AddressBook mentioned that _ABAddressBook was a confusing name). Similarly, ABPerson's backing is called _ABContact (rather than _ABPerson) to match the Windows name of Contact

---
In reply to: [77094758](https://github.com/Microsoft/WinObjC/pull/935#discussion_r77094758) [](ancestors = 77094758)

Frameworks/AddressBook/ABAddressBookManagerInternal.mm
```diff
@@ -82,4 +189,171 @@ - (NSArray*)getListOfContacts {
     return result;
 }
 
+- (NSArray*)getListOfModifiableContacts {
```
DHowett-MSFT: >getListOfModifiableContacts [](start = 12, length = 27)

These two functions look ripe for turning into a helper that turns a contactList into a list of _ABContacts

Frameworks/AddressBook/ABAddressBookManagerInternal.mm
```diff
@@ -82,4 +189,171 @@ - (NSArray*)getListOfContacts {
...
+    return result;
+}
+
+- (bool)addContact:(ABRecordRef)record error:(CFErrorRef*)error {
+    _ABContact* contact = (__bridge _ABContact*)record;
+
+    switch (contact.type) {
+        case kAddressBookNewContact: {
+            // New contacts can't be added twice.
+            if ([_toAdd containsObject:contact.contact]) {
```
DHowett-MSFT: >_toAdd [](start = 17, length = 6)

Since it's in _toAdd: What if you add, save, add again/

DHowett-MSFT: Ah, I see. Below.

---
In reply to: [77099735](https://github.com/Microsoft/WinObjC/pull/935#discussion_r77099735) [](ancestors = 77099735)

Frameworks/AddressBook/ABAddressBookManagerInternal.mm
```diff
@@ -82,4 +189,171 @@ - (NSArray*)getListOfContacts {
...
+                _setError(error, @"This contact already exists in the AddressBook.\n");
+                return false;
+            }
+
+            operation.shouldDelete = NO;
+            return true;
+        }
+        case kAddressBookReadOnlyContact:
+        default:
+            _setError(error, @"This contact already exists in the AddressBook.\n");
```
DHowett-MSFT: >already [](start = 44, length = 7)

This error does not seem right for ReadOnly/Default

Frameworks/AddressBook/ABAddressBookManagerInternal.mm
```diff
@@ -82,4 +189,171 @@ - (NSArray*)getListOfContacts {
...
+    // TODO #942: When the projections for WACContactList's saveContactAsync and deleteContactAsync
+    // get success/fail blocks, use a dispatch_group_t to spawn off all async calls at the same
+    // time rather than the blocking WFIASyncAction getResults, and return false in the case
+    // that the fail block is called (rather than just returning true always).
+
+    // Go through all contacts that were newly created and need to be added,
+    // and save them to this app's contact list.
+    for (WACContact* contact in _toAdd) {
+        RTObject<WFIAsyncAction>* asyncAction = [_contactList saveContactAsync:contact];
+        [asyncAction getResults];
```
DHowett-MSFT: >tion getResults]; [](start = 16, length = 17)

Can this fail? Should we return true even if it does?

DHowett-MSFT: Oh, I see. TODO above.

---
In reply to: [77100271](https://github.com/Microsoft/WinObjC/pull/935#discussion_r77100271) [](ancestors = 77100271)

Frameworks/AddressBook/ABContactInternal.mm
DHowett-MSFT: but also there's no need to use `self->`

---
In reply to: [76677390](https://github.com/Microsoft/WinObjC/pull/935#discussion_r76677390) [](ancestors = 76677390)

Frameworks/AddressBook/ABRecord.mm
```diff
@@ -20,12 +20,84 @@
...
+static void _updateManager(_ABContact* person) {
+    if (person.type == kAddressBookReadWriteContact) {
+        [(__bridge _ABAddressBookManager*)person.manager modifyContact:(__bridge ABRecordRef)person];
+    }
+}
+
+// The maximum allowed String length of various Windows Contacts properties.
+// Values obtained from https://msdn.microsoft.com/en-us/library/windows/apps/windows.applicationmodel.contacts.contact
+// (click on a given property to see any potential length limitations).
+static const NSUInteger kABPersonFirstNameLength = 64;
```
DHowett-MSFT: >64 [](start = 51, length = 2)

Darn. Did they expose any constants for these that we could have used?
If not, we might ask them to, since that's silly.

Frameworks/AddressBook/ABRecord.mm
```diff
@@ -80,18 +152,404 @@ ABRecordType ABRecordGetRecordType(ABRecordRef record) {
...
+    } else if (contactProperty == kABPersonSuffixProperty) {
+        NSString* suffix = (__bridge NSString*)value;
+        if (!_checkLength(suffix, kABPersonSuffixLength, error, @"suffix")) {
+            return false;
+        }
+
+        person.contact.honorificNameSuffix = suffix;
+    } else if (contactProperty == kABPersonNicknameProperty) {
+        NSString* nickname = (__bridge NSString*)value;
+        person.contact.nickname = nickname;
```
DHowett-MSFT: >nickname [](start = 23, length = 8)

Does nickname not have a length limit?

Frameworks/AddressBook/ABRecord.mm
```diff
@@ -80,18 +152,404 @@ ABRecordType ABRecordGetRecordType(ABRecordRef record) {
...
+            NSString* value = (__bridge_transfer NSString*)[multiValue copyValueAtIndex:i];
+            if (!_checkLength(value, kABPersonPhoneLength, error, @"phone")) {
+                return false;
+            }
+        }
+
+        NSMutableArray* phones = person.contact.phones;
+        [phones removeAllObjects];
+
+        NSDictionary* dict = @{
```
DHowett-MSFT: >NSDictionary [](start = 8, length = 12)

You can, if you want, save these dictionaries by making them static.

`static NSDictionary* abWACTypeMappingDictionary = [@{ ... } retain];`

DHowett-MSFT: That'll avoid creating them repeatedly. The choice, however, is yours. :)

---
In reply to: [77101566](https://github.com/Microsoft/WinObjC/pull/935#discussion_r77101566) [](ancestors = 77101566)

Frameworks/AddressBook/ABRecord.mm
```diff
@@ -80,18 +152,404 @@ ABRecordType ABRecordGetRecordType(ABRecordRef record) {
...
+
+        for (int i = 0; i < [multiValue getCount]; i++) {
+            NSString* label = (__bridge_transfer NSString*)[multiValue copyLabelAtIndex:i];
+            NSString* value = (__bridge_transfer NSString*)[multiValue copyValueAtIndex:i];
+            WACContactPhone* phone = [WACContactPhone make];
+            phone.number = value;
+            phone.kind = dict[label] == nil ? WACContactPhoneKindOther : [dict[label] integerValue];
+            [phones addObject:phone];
+        }
+    } else if (contactProperty == kABPersonURLProperty) {
```
DHowett-MSFT: >kABPersonURLProperty [](start = 34, length = 20)

length limit?

DHowett-MSFT:
:shipit:

Frameworks/AddressBook/ABRecord.mm
chadimoussi: I wasn't able to find any in their documentation, unfortunately

---
In reply to: [77100716](https://github.com/Microsoft/WinObjC/pull/935#discussion_r77100716) [](ancestors = 77100716)

Frameworks/AddressBook/ABRecord.mm
chadimoussi: No (at least not a documented one)

---
In reply to: [77101290](https://github.com/Microsoft/WinObjC/pull/935#discussion_r77101290) [](ancestors = 77101290)

Frameworks/AddressBook/ABRecord.mm
chadimoussi: The documentation doesn't list a length limit for the String of a URL.

---
In reply to: [77101644](https://github.com/Microsoft/WinObjC/pull/935#discussion_r77101644) [](ancestors = 77101644)

Frameworks/AddressBook/ABAddressBookManagerInternal.h
ms-jihua: I guess maybe _ABContactManager to match the windows name then? (I still am of the opinion that 'manager' is a terrible thing to name a class, but that's a bit out of our hands I guess)

---
In reply to: [77095804](https://github.com/Microsoft/WinObjC/pull/935#discussion_r77095804) [](ancestors = 77095804,77094758)

ms-jihua:
:shipit:

PR #938, opened 2016-08-29
Fix regular expression test on long running operations.

### LINKED ISSUE #766: NSRegularExpression.NSMatchingOptionsTest fails on OSX
[ RUN      ] NSRegularExpression.NSMatchingOptionsTest
Foundation/NSRegularExpressionTests.m:285: Failure
Value of: expected
  Actual: 20
Expected: count
Which is: 25
[  FAILED  ] NSRegularExpression.NSMatchingOptionsTest (1 ms)

MSFTFox:
Regular expression matching can potentially take longer to run.
On a matching option of type NSMatchingReportProgress, this will
call the block passed in to be called more times than expected.
Modify this test case so that the test calls the block at least
an expected number of times.

Fix #766

### DIFF
--- a/tests/unittests/Foundation/NSRegularExpressionTests.m
+++ b/tests/unittests/Foundation/NSRegularExpressionTests.m
@@ -296,7 +296,14 @@ void testOption(
                          usingBlock:^void(NSTextCheckingResult* textResult, NSMatchingFlags flags, BOOL* stop) {
                              count++;
                          }];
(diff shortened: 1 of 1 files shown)


aballway:
:shipit:

DHowett-MSFT:
:shipit:

rajsesh-msft:
:shipit:

PR #944, opened 2016-08-30
Fix URLByAppendingPathComponent for empty strings

### LINKED ISSUE #880: [NSURL] URLByAppendingPathComponent fails when trying to append empty string
`CFStringGetCharacterAtIndex` supports out of bound access but `CFStringGetCharacterAtIndex(NSConstantString)` throws and exception.

aballway:
To match the reference platform, URLByAppendingPathComponent with an empty path component will append a slash if the URL does not end with one, or return a copy if it does

Fixes #880


chadimoussi:
:shipit:

rajsesh-msft:
:shipit:

PR #948, opened 2016-09-01
Fix headers to address breaks in the api analyzer tool

rajsesh-msft:
An earlier fix d9e19bcf88b187e07359dcb2bcad0e0e9ab48f0a removed dependencies on Foundation from CoreGraphics.  Unfortunately, other modules were counting on foundation headers being included by core graphics.  While this passes build, this fails the API analyzer.

This change is to fix the headers so the api analyzer remains happy.

### DIFF
--- a/include/AVFoundation/AVAssetWriter.h
+++ b/include/AVFoundation/AVAssetWriter.h
@@ -1,6 +1,6 @@
 //******************************************************************************
 //
-// Copyright (c) 2016 Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
 //
 // This code is licensed under the MIT License (MIT).
 //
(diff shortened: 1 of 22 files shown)


DHowett-MSFT:
:shipit:

DHowett-MSFT:
Nit: what determined whether we included the umbrella header `Foundation.h` or the class header `NSObject.h`?

rajsesh-msft:
usage of dispatch_queue_t.  I briefly thought about breaking these dependencies down, but that is a project we can fight another day (given that the build has been failing due to the api analyzer break for a while now).

---
In reply to: [243944720](https://github.com/Microsoft/WinObjC/pull/948#issuecomment-243944720) [](ancestors = 243944720)

PR #949, opened 2016-09-01
Accelerate (vImage) SIMD test addition and bug fixes

shahmeeresmail:
Feature additions, Bug fixes and Unit Test addition:

1. AlphaUnpremultiply (SSE2 path) bug fix: Since some apps pass in vImageBuffers that are subsets of larger vImageBuffers, the padded area at the end of each row in a vImageBuffer must not be touched and code needs to be updated to ensure this. Specifically, there is an optimization that rounds the width up to a multiple of the vector read/write block size that the entire row can be processed using vector instructions in lieu of using scalar instructions for pixels that don't fit within vectorized blocks.

2. In _vImageDivideAndSaturateUint8: avoid unnecessary floating point computations when unpremultiplying alpha and force output component to zero instead of 255 when the input alpha value is zero. Note: Only affects non-SSE2 path.

3. Allow toggling of SIMD instruction usage on and off (primary use cases are for testing and validation).

4. Created new unit test for unpremultiplyAlpha that compares SIMD output against non-SIMD output.


Frameworks/Accelerate/vImage.cpp
```diff
@@ -1809,6 +1756,10 @@ vImage_Error vImageBuffer_Init(
     return returnCode;
 }
 
+void _vImageSetSimdOptmizationsState(const bool newState) {
```
rajsesh-msft: >const [](start = 37, length = 5)

nit: const isnt required for scalar types.

tests/unittests/Accelerate/vImageTest.mm
```diff
@@ -585,8 +812,33 @@ static void vImageFillBufferWithRandomDataF(vImage_Buffer* buffer) {
...
+    vImage_Buffer unpremultipliedBufferSimd, unpremultipliedBufferNormal;
+    const uint8_t alphaVal = 0x80;
+
+    _vImageSetSimdOptmizationsState(false);
+    vImageTestSetAlphaAndUnpremultiply(photo.CGImage, &unpremultipliedBufferNormal, alphaVal);
+
+    _vImageSetSimdOptmizationsState(true);
+    vImageTestSetAlphaAndUnpremultiply(photo.CGImage, &unpremultipliedBufferSimd, alphaVal);
+
+    vImageTestCompare8888Buffers(&unpremultipliedBufferSimd, &unpremultipliedBufferNormal);
```
rajsesh-msft: >vImageTestCompare8888Buffers [](start = 4, length = 28)

This should really be a macro, or you wont have the line number information when a specific test fails.

tests/unittests/Accelerate/vImageTest.mm
```diff
@@ -444,6 +450,227 @@ static void vImageFillBufferWithRandomDataF(vImage_Buffer* buffer) {
...
+
+    if (indexA == 3) {
+        vImageUnpremultiplyData_RGBA8888(&imageBuffer8888, imageBufferUnPremultiplied8888, 0);
+    } else {
+        vImageUnpremultiplyData_ARGB8888(&imageBuffer8888, imageBufferUnPremultiplied8888, 0);
+    }
+
+    CGColorSpaceRelease(format.colorSpace);
+
+    free(imageBuffer8888.data);
```
rajsesh-msft: >free [](start = 4, length = 4)

why not vImageTestBufferFree()?

tests/unittests/Accelerate/vImageTest.mm
```diff
@@ -444,6 +450,227 @@ static void vImageFillBufferWithRandomDataF(vImage_Buffer* buffer) {
...
+                                                     &planeBuffer[indexR],
+                                                     &planeBuffer[indexA],
+                                                     0);
+        }
+    }
+
+    ASSERT_TRUE(result == kvImageNoError);
+    const uint32_t height = (uint32_t)imageBuffer8888.height;
+
+    unsigned char* colorData;
```
rajsesh-msft: >colorData [](start = 19, length = 9)

nit: Prefer declaring and initializing these variables at the same time unless it is unavoidable.

tests/unittests/Accelerate/vImageTest.mm
shahmeeresmail: You're right, a macro would ensure that the line numbers matched up more so than before but it brings with it the potential for naming conflicts when the function exceeds more than a couple of lines. As an alternative to the original implementation and to using a macro, I modified the function to return a bool that says whether or not the buffers match and wrapped the function call in the ASSERT_TRUE macro in the main function so that the line number matched up. The function also prints out a log message indicating the first point at which the buffers don't match.

PR #950, opened 2016-09-01
Basic infrastructure that moves metrics calculations in CoreText to use DirectWrite

mnithish:
This change contains the fundamental pieces required for CoreText to use DirectWrite. This change regresses some of the functionality that was working before with FreeType and will fixed in the coming iterations


mnithish:

mnithish:

mnithish:

mnithish:

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -0,0 +1,395 @@
...
+// TODO::
+// These should be removed once CTFont and UIFont are bridged.
+static IWLazyClassLookup s_lazyUIFont("UIFont");
+static const float c_defaultSystemFontSize = 15.0f;
+
+static ComPtr<IDWriteFactory> CreateDWriteFactoryInstance() {
+    ComPtr<IDWriteFactory> dwriteFactory;
+    THROW_IF_FAILED(DWriteCreateFactory(DWRITE_FACTORY_TYPE_SHARED,
+                                        __uuidof(IDWriteFactory),
+                                        reinterpret_cast<IUnknown**>(dwriteFactory.GetAddressOf())));
```
ms-jihua: >reinterpret_cast<IUnknown**>(dwriteFactory.GetAddressOf()) [](start = 40, length = 58)

can't you just do '&dwriteFactory' here?

Frameworks/CoreText/CTFramesetter.mm
```diff
@@ -16,96 +16,55 @@
...
+        sizeOut->width = frameSize.size.width;
+
+        // Calculate heigth
+        _CTLine* firstLine = static_cast<_CTLine*>([lines objectAtIndex:0]);
+        _CTLine* lastLine = static_cast<_CTLine*>([lines objectAtIndex:[lines count] - 1]);
+        // TODO::
+        // Modify this code once we move away from Cairo Freetype for rasterizing text.
+        CGFloat originYFirstLine = frameSize.size.height - firstLine->_lineOrigin.y - frameSize.origin.y;
+        CGFloat originYLastLine = frameSize.size.height - lastLine->_lineOrigin.y - frameSize.origin.y;
+        sizeOut->height = min(originYLastLine + originYFirstLine, frameSize.size.height);
```
aballway: >originYLastLine + originYFirstLine [](start = 30, length = 34)

Why can't this be firstLine->_lineOrigin.y - lastLine->_lineOrigin.y?

Frameworks/CoreText/CTRun.mm
```diff
@@ -256,11 +261,15 @@ void CTRunDraw(CTRunRef run, CGContextRef ctx, CFRange textRange) {
         range.length = [string length];
     }
 
-    int numGlyphs = curRun->_characters.size();
+    int numGlyphs = [string length];
```
aballway: >numGlyphs [](start = 8, length = 9)

This won't be true for handling ligatures or other instances where there's not a 1-1 matching between a and a glyph

Frameworks/include/CoreTextInternal.h
```diff
@@ -32,38 +36,49 @@
...
-    idretaintype(NSString) _stringFragment;
-    std::vector<CGPoint> _glyphOrigins;
+    float _yPos;
+    StrongId<NSString> _stringFragment;
     std::vector<CGSize> _glyphAdvances;
-    std::vector<WORD> _characters;
+
+    // TODO::
+    // Do we need these anymore?
+    std::vector<CGPoint> _glyphOrigins;
```
aballway: >std::vector<CGPoint> _glyphOrigins; [](start = 4, length = 35)

We'll definitely want vectors for the Get.*Ptr methods, so a _glyphPositions and _stringIndices

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -0,0 +1,395 @@
...
+static ComPtr<IDWriteTextFormat> _createDWriteTextFormat(_CTTypesetter* ts, CFRange range) {
+    // Get the direct write factory instance
+    ComPtr<IDWriteFactory> dwriteFactory = GetDWriteFactoryInstance();
+
+    // TODO::
+    // Get font family name details so that can be used here. For now this is hardcoded to "Gabriola".
+
+    // Note here we only look at attribute value at first index of the specified range as we can get a default faont size to use here.
+    // Per string range attribute handling will be done in _createDWriteTextLayout.
+    NSRange curRange = NSMakeRangeFromCF(range);
```
ms-jihua: bit silly here - you remade the CFRange into an NSRange just to look at its location

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -0,0 +1,395 @@
...
+        wcharString, ts->_charactersLen, textFormat.Get(), frameSize.size.width, frameSize.size.height, &textLayout));
+
+    return textLayout;
+}
+
+/**
+ * Custom IDWriteTextRenderer class that cpatures the glyph runs generated by DWrite for the given TextLayout and TextRenderer constrains.
+ */
+class CustomTargetTextRenderer
+    : public Microsoft::WRL::RuntimeClass<Microsoft::WRL::RuntimeClassFlags<Microsoft::WRL::WinRtClassicComMix>, IDWriteTextRenderer> {
```
ms-jihua: >Microsoft::WRL:: [](start = 13, length = 16)

you're already using this namespace

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -0,0 +1,395 @@
...
+#import <wrl/implements.h>
+#include <COMIncludes_End.h>
+#import <mutex>
+#import <vector>
+#import <algorithm>
+#import <CoreFoundation/CFBase.h>
+#import "CoreTextInternal.h"
+#import "DWriteWrapper.h"
+
+using namespace std;
```
ms-jihua: nit: I don't feel super strongly about this, but I tend to not like using std

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -0,0 +1,395 @@
...
+
+    ComPtr<IDWriteTextLayout> textLayout;
+    THROW_IF_FAILED(dwriteFactory->CreateTextLayout(
+        wcharString, ts->_charactersLen, textFormat.Get(), frameSize.size.width, frameSize.size.height, &textLayout));
+
+    return textLayout;
+}
+
+/**
+ * Custom IDWriteTextRenderer class that cpatures the glyph runs generated by DWrite for the given TextLayout and TextRenderer constrains.
```
ms-jihua: >cpatures [](start = 41, length = 8)

typo

ms-jihua: >constrains [](start = 127, length = 10)

constraints?

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -0,0 +1,395 @@
...
+    // Call custom renderer to get all glyph run details
+    _DWriteGlyphRunDetails glyphRunDetails;
+    _callCustomRender(ts, range, frameSize, &glyphRunDetails);
+
+    // Create _CTLine objects from the the obtained glyph run details
+    int numOfGlyphRuns = glyphRunDetails._glyphRuns.size();
+    int i = 0;
+    int j = 0;
+
+    NSMutableArray<_CTLine*>* lines = [[NSMutableArray alloc] init];
```
ms-jihua: >alloc] init [](start = 55, length = 11)

why not array here?

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -0,0 +1,395 @@
...
+
+    while (j < numOfGlyphRuns) {
```
ms-jihua: >j [](start = 11, length = 1)

interesting that you end up using j first...

I think it's probably actually worth giving them real names here (lineCount, runCount?)

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -0,0 +1,395 @@
...
+/**
+ * Helper method to create _CTLine objects given a _CTTypesetter, attributed string range to use and frame size to fit in.
+ *
+ * @parameter ts _CTTypesetter object to use.
+ * @parameter range attributed string range to use.
+ * @parameter frameSize size parameters of the frame to fit the text into.
+ *
+ * @return Unmutable array of _CTLine objects created with the requested parameters.
+ */
+static NSArray<_CTLine*>* _dwriteGetLines(_CTTypesetter* ts, CFRange range, CGRect frameSize) {
```
ms-jihua: nit: this feels like something we should possibly unit test, maybe add a todo/issue for it

Frameworks/CoreText/CTFramesetter.mm
```diff
@@ -16,96 +16,55 @@
...
 }
 @end
 
-static id _createFrame(_CTFrameSetter* frameSetter, CGRect frameSize, CGSize* sizeOut, bool createFrame) {
+static _CTFrame* _createFrame(_CTFrameSetter* frameSetter, CGRect frameSize, CGSize* sizeOut, bool createFrame) {
     _CTFrame* ret = nil;
 
     if (createFrame) {
-        ret = [_CTFrame alloc];
+        ret = [_CTFrame new];
```
ms-jihua: This looks like it's leaking? (I don't think CoreText is ARC-enabled, is it?)

ms-jihua: elsewhere in this file too

---
In reply to: [77269955](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77269955) [](ancestors = 77269955)

Frameworks/CoreText/CTFramesetter.mm
```diff
@@ -16,96 +16,55 @@
...
+    if (createFrame) {
+        ret->_lines = lines;
     }
 
-    if (ret) {
-        ret->_totalSize = *sizeOut;
+    if (sizeOut) {
+        // TODO::
+        // Is there a better way to do this - Investigate.
+        // For now the below logic is ued to calculate the frame size required to fit the text in -
```
ms-jihua: typo

Frameworks/CoreText/CTFramesetter.mm
```diff
@@ -16,96 +16,55 @@
...
+        //     width = frame width as it is the constraint passed to DWrite to calculate renderable area width
+        //     height = min(_lineOrigin.y of the last line in the frame + the _lineOrigin.y of the first line (so we have some buffer
+        //              space after the last line), frameSize.size.height)
+
+        // Calculate width
+        sizeOut->width = frameSize.size.width;
+
+        // Calculate heigth
```
ms-jihua: typo

ms-jihua:
:clock1:

Frameworks/CoreText/CTFrame.mm
```diff
@@ -83,28 +83,23 @@ void CTFrameGetLineOrigins(CTFrameRef frame, CFRange range, CGPoint origins[]) {
         range.length = 0x7FFFFFF;
     }
 
-    unsigned count = [((_CTFrame*)frame)->_lines count];
-    int idx = 0;
-
-    for (unsigned i = range.location; i < count && i < range.location + range.length; i++) {
-        origins[idx] = ((_CTFrame*)frame)->_lineOrigins[i];
-        idx++;
+    for (uint32_t index = 0; index < [((_CTFrame*)frame)->_lines count]; index++) {
```
aballway: >uint32_t index = 0; index < [((_CTFrame*)frame)->_lines count] [](start = 9, length = 62)

index = range.location; index < range.length

Frameworks/CoreText/CTFrame.mm
```diff
@@ -83,28 +83,23 @@ void CTFrameGetLineOrigins(CTFrameRef frame, CFRange range, CGPoint origins[]) {
         range.length = 0x7FFFFFF;
```
aballway: >0x7FFFFFF [](start = 23, length = 9)

[frame->_lines count]

Frameworks/CoreText/CTFrame.mm
aballway: > for [](start = 3, length = 4)

now this only works if they want all of the line origins, why change it?

Frameworks/CoreText/CTLine.mm
```diff
@@ -275,7 +275,7 @@ CFRange CTLineGetStringRange(CTLineRef line) {
     }
 
     CFRange ret;
-    NSRange range = ((_CTLine*)line)->_strRange;
+    CFRange range = ((_CTLine*)line)->_strRange;
```
aballway: >CFRange range [](start = 4, length = 13)

can we just do ret = line->_strRange?

Frameworks/CoreText/CTRun.mm
```diff
@@ -106,8 +106,7 @@ void CTRunGetGlyphs(CTRunRef run, CFRange range, CGGlyph buffer[]) {
 
 /**
  @Status Caveat
- @Notes Only font face, size and text color attributes are supported.  Background fill is
-        always white.
+ @Notes Only the starting position of the glyph run is handled.
```
aballway: >Only the starting position of the glyph run is handled. [](start = 8, length = 55)

Should just be able to use _glyphOrigins

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -0,0 +1,395 @@
...
+    if (!exists) {
+        index = 0;
+    }
+
+    // Get the string length.
+    uint32_t length = 0;
+    THROW_IF_FAILED(localizedString->GetStringLength(index, &length));
+
+    // Get the string.
+    auto wcharString = make_unique<wchar_t[]>(length + 1);
```
aballway: >auto wcharString [](start = 4, length = 16)

why not std::vector?

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -0,0 +1,395 @@
...
+    float fontSize = [font pointSize];
+
+    ComPtr<IDWriteTextFormat> textFormat;
+    THROW_IF_FAILED(dwriteFactory->CreateTextFormat(L"Gabriola",
+                                                    NULL,
+                                                    DWRITE_FONT_WEIGHT_REGULAR,
+                                                    DWRITE_FONT_STYLE_NORMAL,
+                                                    DWRITE_FONT_STRETCH_NORMAL,
+                                                    fontSize,
+                                                    L"en-us",
```
aballway: >en-us [](start = 54, length = 5)

nit: make this a constant, it's used other places as well

aballway:
:clock1:

Frameworks/CoreText/CTFramesetter.mm
mnithish: It cannot be  firstLine->_lineOrigin.y - lastLine->_lineOrigin.y as firstLine can be equal to lastLine if we only have one line.

That said I think my comment above and the code here does not match. Will fix that

---
In reply to: [77259329](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77259329) [](ancestors = 77259329)

Frameworks/CoreText/CTRun.mm
mnithish: I could not find a documentation to understand this so implemented the code as it was behaving before. I can just use _glyphAdvances here to be safe

---
In reply to: [77259915](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77259915) [](ancestors = 77259915)

Frameworks/include/CoreTextInternal.h
mnithish: Not sure if I understand this comment. I will sync with you offline

---
In reply to: [77261301](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77261301) [](ancestors = 77261301)

Frameworks/CoreText/DWriteWrapper.mm
mnithish: Agreed. With this TODO it does not make sense anymore.

---
In reply to: [77261366](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77261366) [](ancestors = 77261366)

Frameworks/CoreText/DWriteWrapper.mm
mnithish: I am going to leave it as is so I do not have to explicitly call std:: every place

---
In reply to: [77264108](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77264108) [](ancestors = 77264108)

Frameworks/CoreText/DWriteWrapper.mm
mnithish: This code is trying to find glyph runs that make up a line. I do not think we can give these variables any useful name

---
In reply to: [77268929](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77268929) [](ancestors = 77268929)

Frameworks/CoreText/DWriteWrapper.mm
mnithish: Yes, these are in plan to be tested in the next iteration of this code

---
In reply to: [77269567](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77269567) [](ancestors = 77269567)

Frameworks/CoreText/CTFramesetter.mm
mnithish: Welcome to the *ref world - all reference (CTFrameRef/_CTFrame*) are all +1. It is the responsibility of the caller to call CFRelease() on them.

The convention I have followed in the code is - if an API returns a NS* type it follows the regular NS memory management rules. If an API returns a *Ref type it is always +1, so the caller is responsible for freeing them

---
In reply to: [77270014](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77270014) [](ancestors = 77270014,77269955)

Frameworks/CoreText/CTFrame.mm
mnithish: I think this needed a TODO that I missed. I will revert this function back to the original implementation for now

---
In reply to: [77378972](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77378972) [](ancestors = 77378972)

Frameworks/CoreText/CTRun.mm
mnithish: With DWrite we do not have access to _glyphOrigins. I tried calculating it myself with whatever DWrite provided back to me but it was way off.

---
In reply to: [77379732](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77379732) [](ancestors = 77379732)

Frameworks/CoreText/CTFramesetter.mm
ms-jihua: Yeah my bad, I didn't notice that this was a create function

---
In reply to: [77455236](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77455236) [](ancestors = 77455236,77270014,77269955)

Frameworks/CoreText/CTFramesetter.mm
```diff
@@ -16,96 +16,55 @@
...
     }
 
-    if (ret) {
-        ret->_totalSize = *sizeOut;
+    if (sizeOut) {
+        // TODO::
+        // Is there a better way to do this - Investigate.
+        // For now the below logic is used to calculate the frame size required to fit the text in -
+        //     width = frame width as it is the constraint passed to DWrite to calculate renderable area width
+        //     height = min(_lineOrigin.y of the last line in the frame + the _lineOrigin.y of the first line (so we atlease get the
```
ms-jihua: >lease [](start = 119, length = 5)

typo

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -0,0 +1,393 @@
...
+    // Create _CTLine objects from the the obtained glyph run details
+    int numOfGlyphRuns = glyphRunDetails._glyphRuns.size();
+    int i = 0;
+    int j = 0;
+
+    NSMutableArray<_CTLine*>* lines = [NSMutableArray array];
+
+    while (j < numOfGlyphRuns) {
+        _CTLine* line = [[_CTLine new] autorelease];
+        NSMutableArray<_CTRun*>* runs = [[NSMutableArray array] retain];
```
ms-jihua: >NSMutableArray<_CTRun*>* runs = [[NSMutableArray array] retain]; [](start = 8, length = 64)

I see that this isn't leaking because you use an attach below, but to me it seems a lot more comfortable/less prone to error if you don't retain here, and use an = below

Frameworks/CoreText/CTRun.mm
```diff
@@ -256,11 +261,15 @@ void CTRunDraw(CTRunRef run, CGContextRef ctx, CFRange textRange) {
         range.length = [string length];
     }
 
-    int numGlyphs = curRun->_characters.size();
+    int numGlyphs = curRun->_glyphAdvances.size();
     WORD* glyphs = (WORD*)IwMalloc(sizeof(WORD) * numGlyphs);
 
+    WORD* characters = (WORD*)IwMalloc(sizeof(WORD) * numGlyphs);
```
aballway: >characters [](start = 10, length = 10)

Does this need to be freed?

Frameworks/include/CoreTextInternal.h
```diff
@@ -32,38 +36,48 @@
...
     std::vector<CGPoint> _lineOrigins;
 }
 @end
 
 @interface _CTLine : NSObject {
 @public
-    NSRange _strRange;
+    _CTTypesetter* _typeSetter;
+    CFRange _strRange;
+    CGPoint _lineOrigin;
```
aballway: >_lineOrigin [](start = 12, length = 11)

Will DWrite give origins for the line or for each glyphrun? If the latter we should keep this variable in the runs

Frameworks/include/CoreTextInternal.h
```diff
@@ -32,38 +36,48 @@
...
     CGSize _totalSize;
-    idretaintype(NSMutableArray) _lines;
     std::vector<CGPoint> _lineOrigins;
 }
 @end
 
 @interface _CTLine : NSObject {
 @public
-    NSRange _strRange;
+    _CTTypesetter* _typeSetter;
```
aballway: >_typeSetter [](start = 19, length = 11)

Do CTLines need a typesetter?

Frameworks/CoreText/DWriteWrapper.mm
mnithish: Sure will do

---
In reply to: [77680658](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77680658) [](ancestors = 77680658)

Frameworks/include/CoreTextInternal.h
mnithish: Not any mote. I have these changed to DWrite objects in the next CR. I will clean these up as part of that change

---
In reply to: [77732812](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77732812) [](ancestors = 77732812)

Frameworks/CoreText/CTRun.mm
mnithish: Yes it needs to do. Good catch

---
In reply to: [77732196](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77732196) [](ancestors = 77732196)

Frameworks/include/CoreTextInternal.h
mnithish: It would give it for every glyph run but it is easy to calculate it for a line and we need line origins in CoreText so instead of calculating it on every access I cached it in the CTLine object

---
In reply to: [77732628](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77732628) [](ancestors = 77732628)

Frameworks/CoreText/DWriteWrapper.h
```diff
@@ -0,0 +1,47 @@
...
+};
+
+struct _DWriteGlyphRunDetails {
+    std::vector<float> _baselineOriginX;
+    std::vector<float> _baselineOriginY;
+    std::vector<_DWriteGlyphRun> _glyphRuns;
+    std::vector<_DWriteGlyphRunDescription> _glyphRunDescriptions;
+};
+
+NSArray<NSString*>* _dwriteGetFamilyNames();
```
rajsesh-msft: >_dwriteGetFamilyNames [](start = 20, length = 21)

nit: for "C" style functions, we should start with caps.

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -0,0 +1,392 @@
...
+static IWLazyClassLookup s_lazyUIFont("UIFont");
+static const float c_defaultSystemFontSize = 15.0f;
+
+static ComPtr<IDWriteFactory> CreateDWriteFactoryInstance() {
+    ComPtr<IDWriteFactory> dwriteFactory;
+    THROW_IF_FAILED(DWriteCreateFactory(DWRITE_FACTORY_TYPE_SHARED, __uuidof(IDWriteFactory), &dwriteFactory));
+    return dwriteFactory;
+}
+
+static ComPtr<IDWriteFactory> GetDWriteFactoryInstance() {
```
rajsesh-msft: >IDWriteFactory [](start = 14, length = 14)

Nit: If this method is called a lot, it would make sense to simply return the IDWriteFactory* instead.  The caller will save it into a ComPtr, holding reference correctly.  As is, we will have extra addref/release due to the temporary object.

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -0,0 +1,392 @@
...
+    if (!exists) {
+        index = 0;
+    }
+
+    // Get the string length.
+    uint32_t length = 0;
+    THROW_IF_FAILED(localizedString->GetStringLength(index, &length));
+
+    // Get the string.
+    vector<wchar_t> wcharString = std::vector<wchar_t>(length + 1);
```
rajsesh-msft: >vector<wchar_t> wcharString = std::vector<wchar_t>(length + 1); [](start = 4, length = 63)

vector<wchar_t> wcharString(length+1);  Its a shame that std::wstring() doesnt have a constructor that just allocates the buffer.

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -0,0 +1,392 @@
...
+}
+
+/**
+ * Helper method to return the user set default locale string.
+ *
+ * @return use set locale string as wstring.
+ */
+static wstring _getUserDefaultLocaleName() {
+    wchar_t localeName[LOCALE_NAME_MAX_LENGTH];
+    int defaultLocaleSuccess = GetUserDefaultLocaleName(localeName, LOCALE_NAME_MAX_LENGTH);
```
rajsesh-msft: use [NSLocale currentLocale] instead?

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -0,0 +1,392 @@
...
+    HRESULT RuntimeClassInitialize();
+
+    HRESULT STDMETHODCALLTYPE DrawGlyphRun(_In_ void* clientDrawingContext,
+                                           _In_ float baselineOriginX,
+                                           _In_ float baselineOriginY,
+                                           DWRITE_MEASURING_MODE measuringMode,
+                                           _In_ DWRITE_GLYPH_RUN const* glyphRun,
+                                           _In_ DWRITE_GLYPH_RUN_DESCRIPTION const* glyphRunDescription,
+                                           _In_ IUnknown* clientDrawingEffects) throw() {
+        _DWriteGlyphRunDetails* glyphs = static_cast<_DWriteGlyphRunDetails*>(clientDrawingContext);
```
rajsesh-msft: >static_cast [](start = 41, length = 11)

nit: reinterpret_cast<>

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -0,0 +1,392 @@
...
+        glyphs->_baselineOriginX.push_back(baselineOriginX);
+        glyphs->_baselineOriginY.push_back(baselineOriginY);
+
+        _DWriteGlyphRun glyphRunInfo;
+        glyphRunInfo._glyphCount = glyphRun->glyphCount;
+        for (int index = 0; index < glyphRun->glyphCount; index++) {
+            glyphRunInfo._glyphIndices.push_back(glyphRun->glyphIndices[index]);
+            glyphRunInfo._glyphAdvances.push_back(CGSizeMake(glyphRun->glyphAdvances[index], 0));
+        }
+        glyphs->_glyphRuns.push_back(glyphRunInfo);
```
rajsesh-msft: >glyphRunInfo [](start = 37, length = 12)

I don't like this much from a perf standpoint.  This essentially will make a copy of the glyphRunInfo, copying all the vectors again.  How big are these vectors?

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -0,0 +1,392 @@
...
+                                               _In_ float originY,
+                                               _In_ IDWriteInlineObject* inlineObject,
+                                               _In_ BOOL isSideways,
+                                               _In_ BOOL isRightToLeft,
+                                               _In_ IUnknown* clientDrawingEffects) throw() {
+        return S_OK;
+    };
+
+    HRESULT STDMETHODCALLTYPE IsPixelSnappingDisabled(_In_opt_ void* clientDrawingContext, _Out_ BOOL* isDisabled) throw() {
+        *isDisabled = false;
```
rajsesh-msft: >false [](start = 22, length = 5)

nit: FALSE.

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -0,0 +1,392 @@
...
+        return S_OK;
+    };
+
+private:
+    ComPtr<IDWriteFactory> _dwriteFactory;
+};
+
+CustomTargetTextRenderer::CustomTargetTextRenderer() {
+    _dwriteFactory = GetDWriteFactoryInstance();
+}
```
rajsesh-msft: seems super redundant to cache the factory here, why not always call the static method?

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -0,0 +1,392 @@
...
+    THROW_IF_FAILED(dwriteFactory->CreateTextLayout(
+        wcharString, ts->_charactersLen, textFormat.Get(), frameSize.size.width, frameSize.size.height, &textLayout));
+
+    return textLayout;
+}
+
+/**
+ * Custom IDWriteTextRenderer class that captures the glyph runs generated by DWrite for the given TextLayout and TextRenderer constraints.
+ */
+class CustomTargetTextRenderer : public RuntimeClass<RuntimeClassFlags<WinRtClassicComMix>, IDWriteTextRenderer> {
```
rajsesh-msft: >CustomTargetTextRenderer [](start = 6, length = 24)

MeasuringDWriteTextRenderer will clarify the intent of this class.

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -0,0 +1,392 @@
...
+        // Get the font family.
+        ComPtr<IDWriteFontFamily> fontFamily;
+        THROW_IF_FAILED(fontCollection->GetFontFamily(i, &fontFamily));
+
+        // Get a list of localized strings for the family name.
+        ComPtr<IDWriteLocalizedStrings> familyNames;
+        THROW_IF_FAILED(fontFamily->GetFamilyNames(&familyNames));
+
+        [fontFamilyNames addObject:_convertLocalizedStringToNSString(familyNames.Get())];
+    }
```
rajsesh-msft: Presumably, this code block could use the new CTFont* apis that @ms-jihua is writing?  If so, could you please add a TODO here to change that?

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -0,0 +1,392 @@
...
+ * @parameter ts _CTTypesetter object to use.
+ * @parameter range attributed string range to use.
+ * @parameter frameSize size parameters of the frame to fit the text into.
+ *
+ * @return Unmutable array of _CTLine objects created with the requested parameters.
+ */
+static NSArray<_CTLine*>* _dwriteGetLines(_CTTypesetter* ts, CFRange range, CGRect frameSize) {
+    // Call custom renderer to get all glyph run details
+    _DWriteGlyphRunDetails glyphRunDetails;
+    _callCustomRender(ts, range, frameSize, &glyphRunDetails);
```
rajsesh-msft: >_callCustomRender [](start = 4, length = 17)

_GetGlyphRunDetails may be more appropriate name.  _callCustomRender breaks abstraction by exposing implementation details.

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -0,0 +1,392 @@
...
+        // Glyph runs that have the same _baselineOriginY value are part of the the same Line.
+        while ((j < numOfGlyphRuns) && (glyphRunDetails._baselineOriginY[i] == glyphRunDetails._baselineOriginY[j])) {
+            j++;
+        }
+        while (i < j) {
+            // Create _CTRun objects and make them part of _CTLine
+            _CTRun* run = [[_CTRun new] autorelease];
+            run->_range.location = glyphRunDetails._glyphRunDescriptions[i]._textPosition;
+            run->_range.length = glyphRunDetails._glyphRunDescriptions[i]._stringLength;
+            run->_stringFragment = [ts->_string substringWithRange:NSMakeRangeFromCF(run->_range)];
```
rajsesh-msft: >t = [ts->_string substringWithRange:NSMakeRangeFromCF(run->_range)] [](start = 31, length = 67)

We are making _stringFragments when we may not actually use them.  It probably makes sense to cache the ts in this object and create the fragment only as needed (if accessed).

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -0,0 +1,392 @@
...
+            run->_xPos = frameSize.origin.x + glyphRunDetails._baselineOriginX[i];
+            // TODO::
+            // This is a temp work around to get DWrite measurements to be usable with Cairo FreeType rasterizer. Remove this when
+            // we move away from Cairo Freetype. We need this workaround because DWrite coordinate system begins from top left but
+            // Cairo FreeType has its from bottom left.
+            run->_yPos = frameSize.origin.y + frameSize.size.height - glyphRunDetails._baselineOriginY[i];
+            run->_attributes = [ts->_attributedString attributesAtIndex:run->_range.location effectiveRange:NULL];
```
rajsesh-msft: same comments here as with _stringFragments.

Frameworks/include/CoreTextInternal.h
```diff
@@ -32,38 +36,47 @@
...
     std::vector<CGPoint> _lineOrigins;
 }
 @end
 
 @interface _CTLine : NSObject {
 @public
-    NSRange _strRange;
+    CFRange _strRange;
+    CGPoint _lineOrigin;
+    StrongId<NSMutableArray> _runs;
```
rajsesh-msft: >NSMutableArray [](start = 13, length = 14)

NSMutableArray<_CTRun*> to avoid all the static casts everywhere else?

rajsesh-msft: Other places too, the code will be lot nicer to read.

---
In reply to: [77757492](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77757492) [](ancestors = 77757492)

Frameworks/CoreText/CTRun.mm
```diff
@@ -117,7 +116,13 @@ void CTRunGetPositions(CTRunRef run, CFRange runRange, CGPoint* outPositions) {
         runRange.length = curRun->_range.length;
     }
 
-    memcpy(outPositions, &curRun->_glyphOrigins[runRange.location], sizeof(CGPoint) * runRange.length);
+    if (runRange.location != 0) {
+        UNIMPLEMENTED();
```
rajsesh-msft: >UNIMPLEMENTED [](start = 8, length = 13)

UNIMPLEMENTED_WITH_MSG()

Frameworks/CoreText/CTRun.mm
```diff
@@ -117,7 +116,13 @@ void CTRunGetPositions(CTRunRef run, CFRange runRange, CGPoint* outPositions) {
...
+        return;
+    }
+
+    outPositions->x = curRun->_xPos;
```
rajsesh-msft: >outPositions [](start = 4, length = 12)

I am not even sure if this is useful information.  We are essentially returning one glyph position.  Need a TODO to implement the real version.  What didn't work here with DWrite?

Frameworks/CoreText/CTRun.mm
```diff
@@ -130,7 +135,7 @@ void CTRunGetPositions(CTRunRef run, CFRange runRange, CGPoint* outPositions) {
 }
 
 /**
- @Status Interoperable
+ @Status Stub
```
rajsesh-msft: >Stub [](start = 9, length = 4)

Regression - I presume this will be addressed in #961?

Frameworks/CoreText/CTTypesetter.mm
```diff
@@ -84,6 +84,7 @@ static float fractionPixels(float pixels) {
 
 static CFIndex _DoWrap(CTTypesetterRef ts, CFRange range, WidthFinderFunc widthFunc, void* widthParam, double offset, CTLineRef* outLine) {
```
rajsesh-msft: Why do we still need this method? Presumably for Draw until we move it to DWrite?

rajsesh-msft:
:clock1:

Frameworks/CoreText/CTRun.mm
aballway: It would probably be better to just use std::vector characters(numGlyphs); than malloc and freeing manually style-wise 

---
In reply to: [77747600](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77747600) [](ancestors = 77747600,77732196)

mnithish: This code is going to be thrown away so kept with the existing coding style

---
In reply to: [77862776](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77862776) [](ancestors = 77862776,77747600,77732196)

Frameworks/CoreText/DWriteWrapper.mm
mnithish: They need to be copied as the memory for glyphIndices and glyphAdvances are owned by DWrite and is not will be freed once this callback returns

---
In reply to: [77756323](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77756323) [](ancestors = 77756323)

Frameworks/CoreText/DWriteWrapper.mm
ms-jihua: We both ended up writing this helper actually - I'll defer to @mnithish 's since I like the placement better, and because mine is messier due to trying to stick to CF functions at the time.

---
In reply to: [77756735](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77756735) [](ancestors = 77756735)

Frameworks/CoreText/DWriteWrapper.mm
mnithish: If I remember correctly NSLocale used to return locale string of the format "en_US" which does not match with what Windows expects. I will verify this again

---
In reply to: [77756000](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77756000) [](ancestors = 77756000)

Frameworks/CoreText/DWriteWrapper.mm
mnithish: Yes this code is not used any more. Will remove

---
In reply to: [77756443](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77756443) [](ancestors = 77756443)

Frameworks/CoreText/DWriteWrapper.mm
mnithish: It is not just for measuring, we also get glyph runs that we will use to render now. I will re-name is to something *DWrite*

---
In reply to: [77756507](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77756507) [](ancestors = 77756507)

Frameworks/CoreText/DWriteWrapper.mm
mnithish: We might need the fragments for draw so I cached it in the object instead of sending a message to NS to calculate every time. Also it was very useful for debugging purposes. Will not over optimize now given we are still developing and this is not a huge memory over head

---
In reply to: [77757111](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77757111) [](ancestors = 77757111)

Frameworks/CoreText/DWriteWrapper.mm
mnithish: <<see above comment>>

---
In reply to: [77757337](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77757337) [](ancestors = 77757337)

Frameworks/CoreText/CTRun.mm
mnithish: Correct

---
In reply to: [77757935](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77757935) [](ancestors = 77757935)

Frameworks/CoreText/CTRun.mm
mnithish: DWrite only supports providing the points for the first glyph in the glyph run and not for every glyph in the run. We need so find a way to calculate this but the few things I tried did not work so need to get DWrite team's recommendation on what to do here.

I will add a TODO

---
In reply to: [77757825](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77757825) [](ancestors = 77757825)

Frameworks/CoreText/DWriteWrapper.h
mnithish: CoreText has not followed this convention so tried to keep up with the existing convention

---
In reply to: [77755487](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77755487) [](ancestors = 77755487)

Frameworks/CoreText/CTTypesetter.mm
mnithish: This is the entry point for NSLayout*

---
In reply to: [77757993](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77757993) [](ancestors = 77757993)

Frameworks/include/CoreTextInternal.h
mnithish: It is not going to help with the static casts as  NSMutableArray only returns an id type but I agree from readability of this file given we now have support for generics

---
In reply to: [77757580](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77757580) [](ancestors = 77757580,77757492)

Frameworks/CoreText/DWriteWrapper.mm
mnithish: I am converting from a void* to _DWriteGlyphRunDetails* here so static_cast seems appropriate here

---
In reply to: [77756151](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77756151) [](ancestors = 77756151)

Frameworks/include/CoreTextInternal.h
mnithish: Dustin stopped by and looks like this is a bug with StongId

---
In reply to: [77873760](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77873760) [](ancestors = 77873760,77757580,77757492)

Frameworks/CoreText/DWriteWrapper.mm
mnithish: Will consider it if we plan to use it a lot but I don't think that given it is called only at couple of places today

---
In reply to: [77755593](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77755593) [](ancestors = 77755593)

Frameworks/CoreText/DWriteWrapper.mm
mnithish: Yup I was right. Cannot use NSLocale because of the above reason

---
In reply to: [77869413](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77869413) [](ancestors = 77869413,77756000)

Frameworks/CoreFoundation/Base.subproj/CFBase.h
```diff
@@ -498,6 +498,7 @@ CF_INLINE CFRange CFRangeMake(CFIndex loc, CFIndex len) {
 CF_EXPORT
 CFRange __CFRangeMake(CFIndex loc, CFIndex len);
 
+#define NSMakeRangeFromCF(cfr) NSMakeRange(cfr.location == kCFNotFound ? NSNotFound : cfr.location, cfr.length)
```
DHowett-MSFT: >NSMakeRange [](start = 31, length = 11)

I don't think this should be here; it introduces a dependency from CF headers to Foundation headers.

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -0,0 +1,392 @@
...
+
+static ComPtr<IDWriteFactory> CreateDWriteFactoryInstance() {
+    ComPtr<IDWriteFactory> dwriteFactory;
+    THROW_IF_FAILED(DWriteCreateFactory(DWRITE_FACTORY_TYPE_SHARED, __uuidof(IDWriteFactory), &dwriteFactory));
+    return dwriteFactory;
+}
+
+static ComPtr<IDWriteFactory> GetDWriteFactoryInstance() {
+    static ComPtr<IDWriteFactory> s_dwriteFactory = CreateDWriteFactoryInstance();
+    return s_dwriteFactory;
```
DHowett-MSFT: Is it safe to keep the factory around for the lifetime of the application?

Frameworks/CoreText/DWriteWrapper.mm
DHowett-MSFT: it does, though. unfortunately, just like `vector`, it will construct the entire range (or here, memset it)

```
template <typename CharT, ...>
basic_string(size_t count, CharT ch, ...);
```

---
In reply to: [77755880](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77755880) [](ancestors = 77755880)

Frameworks/CoreText/DWriteWrapper.mm
mnithish: Yes as this is the same design pattern UWP uses

---
In reply to: [77901157](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77901157) [](ancestors = 77901157)

Frameworks/CoreFoundation/Base.subproj/CFBase.h
mnithish: Good point. Let me move this to NS*

---
In reply to: [77900982](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77900982) [](ancestors = 77900982)

mnithish: Going to keep it local to CT for now until I find a good place for it

---
In reply to: [77923909](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77923909) [](ancestors = 77923909,77900982)

mnithish:
Ping!

Frameworks/CoreText/DWriteWrapper.mm
rajsesh-msft: static_cast is only useful where types are related.  void * isnt really related.

---
In reply to: [77875281](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77875281) [](ancestors = 77875281,77756151)

Frameworks/include/CoreTextInternal.h
rajsesh-msft: Lets get the bug on StrongId<> filed.  I am even okay removing StrongID in this case, since it would make the readability a *lot* better in all the other places.

---
In reply to: [77875386](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77875386) [](ancestors = 77875386,77873760,77757580,77757492)

Frameworks/CoreText/CTRun.mm
rajsesh-msft: could you open a github issue for this so this doesnt fall through the cracks? 

---
In reply to: [77871237](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77871237) [](ancestors = 77871237,77757825)

Frameworks/CoreText/DWriteWrapper.mm
rajsesh-msft: The issue I am concerned about is the double copy.  Once into glyphRunInfo and once when it is added to the vector.  this is somewhat problematic. 

---
In reply to: [77864667](https://github.com/Microsoft/WinObjC/pull/950#discussion_r77864667) [](ancestors = 77864667,77756323)

Frameworks/CoreText/DWriteWrapper.mm
mnithish: Synced offline - static_cast is appropriate here

---
In reply to: [78073085](https://github.com/Microsoft/WinObjC/pull/950#discussion_r78073085) [](ancestors = 78073085,77875281,77756151)

Frameworks/CoreText/DWriteWrapper.mm
mnithish: Synced offline - as this code is going away in the next iteration leaving it as is

---
In reply to: [78079893](https://github.com/Microsoft/WinObjC/pull/950#discussion_r78079893) [](ancestors = 78079893,77864667,77756323)

mnithish:
In CI build

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -0,0 +1,387 @@
...
+    ComPtr<CustomDWriteTextRenderer> textRenderer = Make<CustomDWriteTextRenderer>();
+    textLayout->Draw(glyphDetails, textRenderer.Get(), 0, 0);
+}
+
+/**
+ * Helper method to retrieve font fmaly names installed in the system.
+ *
+ * @return Unmutable array of font family name strings that are installed in the system.
+ */
+static NSArray<NSString*>* _dwriteGetFamilyNames() {
```
ms-jihua: >_dw [](start = 27, length = 3)

Note that this currently doesn't match the capitalization in the header.

mnithish: Weird I fixed all of these to be compliant with our coding style. Let me make sure that change got pushed

---
In reply to: [78095476](https://github.com/Microsoft/WinObjC/pull/950#discussion_r78095476) [](ancestors = 78095476)

ms-jihua:
didn't you say it would fail though?

---
In reply to: [245745333](https://github.com/Microsoft/WinObjC/pull/950#issuecomment-245745333) [](ancestors = 245745333)

mnithish:
Yup some CT UTs will but the build is to make sure there are not any unknown regressions

---
In reply to: [245756330](https://github.com/Microsoft/WinObjC/pull/950#issuecomment-245756330) [](ancestors = 245756330,245745333)

PR #956, opened 2016-09-02
Disabling test that causes a crash in Visual Studio.

### LINKED ISSUE #955: AddressBook ModifyContactsTests Crashes Visual Studio Project
The ModifyContactsTests crashes Visual Studio when one of its tools tries to parse the file. The root cause still needs to be identified. Until this is fixed, the file will be disabled.

MSFTFox:
This test causes a crash when visual studio attempts to parse
the content of this test. It has been commented out.

### DIFF
--- a/tests/unittests/AddressBook/ModifyContactsTests.mm
+++ b/tests/unittests/AddressBook/ModifyContactsTests.mm
@@ -84,6 +84,10 @@ TEST(AddressBookModify, ContactBirthday) {
     [dateComponents release];
 }
 
+// TODO: 955 AddressBook ModifyContactsTests Crashes Visual Studio Project #955
+// This test has been disabled due to an issue in Visual Studio attempting to parse the contents of this test.
+
+/*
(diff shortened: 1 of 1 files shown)


DHowett-MSFT:
:clock1:

aballway:
:shipit:

memontic-ms:
:shipit:

tests/unittests/AddressBook/ModifyContactsTests.mm
```diff
@@ -84,6 +84,10 @@ virtual void TearDown() {
     [dateComponents release];
 }
 
+// TODO: 955 AddressBook ModifyContactsTests Crashes Visual Studio Project #955
```
DHowett-MSFT: > 955 AddressBook ModifyContactsTests Crashes Visual Studio Project #955 [](start = 8, length = 71)

why `955 #955`?

DHowett-MSFT:
:shipit:

bbowman:
Uh what? This doesn't seem like the correct solution. What about this test is crashing vs?

PR #975, opened 2016-09-08
Fixing a crash when calling CGContextAddPath with a NULL path

### LINKED ISSUE #966: CGContextAddPath crashes with NULL path parameter
CGContextAddPath will crash when a null path parameter is passed in.  On iOS is does not crash and the Apple documentation implies that a null input value is ok.

pittmab:
The Apple docs for CGContextAddPath imply that passing a NULL path does nothing and is a valid API call.

Fix #966

### DIFF
--- a/Frameworks/CoreGraphics/CGContext.mm
+++ b/Frameworks/CoreGraphics/CGContext.mm
@@ -452,7 +452,11 @@ void CGContextFillEllipseInRect(CGContextRef ctx, CGRect rct) {
  @Status Interoperable
 */
 void CGContextAddPath(CGContextRef ctx, CGPathRef path) {
-    ctx->Backing()->CGContextAddPath(path);
+    // The Apple SDK docs imply that passing a NULL path is valid.
+    // So avoid calling into the backing if NULL.
+    if (path) {
+        ctx->Backing()->CGContextAddPath(path);
+    }
 }
 
 /**


Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -452,6 +452,11 @@ void CGContextFillEllipseInRect(CGContextRef ctx, CGRect rct) {
  @Status Interoperable
 */
 void CGContextAddPath(CGContextRef ctx, CGPathRef path) {
+    // The Apple SDK docs imply that passing a NULL path is valid
+	if (!path) {
```
aballway: >	 [](start = 0, length = 1)

nit: spaces, not tabs

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -452,6 +452,11 @@ void CGContextFillEllipseInRect(CGContextRef ctx, CGRect rct) {
...
+        return;
+	}
```
aballway: nit: this would be better with the work in the if rather than preemptively returning, especially since it's void

aballway:
:clock1:

Frameworks/CoreGraphics/CGContext.mm
pittmab: Updated to use spaces.

Frameworks/CoreGraphics/CGContext.mm
pittmab: Updated the code to reflect your preferences.  Thanks

pittmab: @aballway I have updated the code to reflect your comments.  Thanks

rajsesh-msft:
:shipit:

PR #977, opened 2016-09-08
Fixed a compilation issue where a struct (CVTimeStamp) was being forw…

pittmab:
…ard declared as an Objective C class

Fix: 976

### DIFF
--- a/include/QuartzCore/CARenderer.h
+++ b/include/QuartzCore/CARenderer.h
@@ -20,7 +20,6 @@
 #import <Foundation/NSObject.h>
 
 @class CALayer;
-@class CVTimeStamp;
 
 @interface CARenderer : NSObject {
     void* _cglContext;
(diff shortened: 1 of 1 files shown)


include/QuartzCore/CARenderer.h
```diff
@@ -15,12 +15,12 @@
 //******************************************************************************
 #pragma cone
 
+#import <CoreVideo/CVTime.h>
```
rajsesh-msft: >CoreVideo [](start = 9, length = 9)

rather than introducing this dependency, we should simply change the forward declaration below to be struct CVTimeStamp.

rajsesh-msft:
:clock1:

include/QuartzCore/CARenderer.h
pittmab: Hi Raj,

I originally tried to forward declare the struct but couldn't get it to work given how the original CVTimeStamp is declared.

typedef struct {
    uint32_t version;
    int32_t videoTimeScale;
    int64_t videoTime;
    uint64_t hostTime;
    double rateScalar;
    int64_t videoRefreshPeriod;
    CVSMPTETime smpteTime;
    uint64_t flags;
    uint64_t reserved;
} CVTimeStamp;

Do you know how to forward declare it?

Thanks,

Brent

rajsesh-msft: Brent, you could entirely remove the @class CVTimeStamp and change the method to timeStamp:(struct CVTimeStamp*)timeStamp. 

---
In reply to: [78206753](https://github.com/Microsoft/WinObjC/pull/977#discussion_r78206753) [](ancestors = 78206753)

pittmab:
@rajsesh-msft made the change to avoid the include and use the forward declare.

PR #979, opened 2016-09-08
Fixed a compilation issue where the typedef for GLsizei does not matc…

pittmab:
…h the OpenGL header type

Fix: 978

### DIFF
--- a/include/CoreVideo/CVOpenGLESTexture.h
+++ b/include/CoreVideo/CVOpenGLESTexture.h
@@ -27,7 +27,7 @@ typedef unsigned int GLuint;
 typedef signed int GLint;
 typedef char GLchar;
 typedef float GLfloat;
-typedef size_t GLsizei;
+typedef int GLsizei;
 
 typedef CVImageBufferRef CVOpenGLESTextureRef;
 


DHowett-MSFT:
Curiously, it looks like our GL headers may be the incorrect ones:
[This](https://www.opengl.org/wiki/OpenGL_Type) specifies that it should be unsigned.

MSFTFox:
There's a bit of interesting information on this and I believe int should be correct the type here.

http://oss.sgi.com/archives/ogl-sample/2005-07/msg00003.html

rajsesh-msft:
Deceptive commentary in the docs there.  "GLsizei​ 32 A non-negative binary integer, for sizes." vs "GLuint64​ 64 Unsigned binary integer".  Implies this is actually an integer.

---
In reply to: [245742064](https://github.com/Microsoft/WinObjC/pull/979#issuecomment-245742064) [](ancestors = 245742064)

rajsesh-msft:
:shipit:

MSFTFox:
:shipit:

DHowett-MSFT:
:shipit:

PR #983, opened 2016-09-08
Fixed a bug where the Pinch gesture recognizer wasn't resetting it's …

pittmab:
…stored data properly

There was a bug in which the pinch gesture would keep track of a touch from a previous interaction and incorrectly be recognized in later interactions.

Fix: 982

### DIFF
--- a/Frameworks/UIKit/UIPinchGestureRecognizer.mm
+++ b/Frameworks/UIKit/UIPinchGestureRecognizer.mm
@@ -55,6 +55,15 @@ static void deleteTouch(UIPinchGestureRecognizer* self, UITouch* touch) {
     }
 }
 
+- (void)reset {
+    [super reset];
+
+    _curScale = 1.0f;
+    _startScale = 1.0f;
+    _startedPinch = false;
+    _numTouches = 0;
+}
+
 static void deleteAllTouches(UIPinchGestureRecognizer* self) {
     ::memset(&self->_touches, 0, sizeof(self->_touches));
     self->_numTouches = 0;


rajsesh-msft:
@jaredhms please take a look.

jaredhms:
Thanks for the fix, but this was fixed within the ~past week on github master in this commit: https://github.com/Microsoft/WinObjC/commit/fb61e282ad3ef92184ec5b8324222ade4611dacc.

pittmab:
@jaredhms The issue I found happened after a successful pinch gesture was recognized.  The next interaction was locked to a pinch gesture since the reset method was not implemented in the pinch gesture.  The commit you referenced only cleans up data on touchesCancelled, did it address this issue where the Pinch Gesture needs to be reset?

jaredhms:
@davelamb, please take a look.

Frameworks/UIKit/UIPinchGestureRecognizer.mm
```diff
@@ -55,6 +55,15 @@ static void deleteTouch(UIPinchGestureRecognizer* self, UITouch* touch) {
     }
 }
 
+- (void)reset {
+    [super reset];
+
+    _curScale = 1.0f;
+    _startScale = 1.0f;
+    _startedPinch = false;
+    _numTouches = 0;
```
jaredhms: should we call deleteAllTouches(self) here instead?

jaredhms:
Apologies for the confusion; I agree that it looks like we should be implementing reset.  

---
In reply to: [246369916](https://github.com/Microsoft/WinObjC/pull/983#issuecomment-246369916) [](ancestors = 246369916)

jaredhms:
:shipit:

jaredhms:

davelamb:
:shipit:

DHowett-MSFT:
Thanks!

PR #986, opened 2016-09-08
Add a script that purports to submit CI builds.

DHowett-MSFT:
This script attempts to be intelligent about remote branch name matching and stop you from submitting a build off an un-synced branch.

It will likely run amok, so please use `Submit-CIBuilds -Verbose -WhatIf` to inspect the submitted build definitions before takeoff.

By default, it does the following:
1. Look up your current branch.
2. Look up your current branch's remote.
3. Look up your current branch's tracking branch.
4. Check the most recent local commit against the most recent remot commit.
5. Convert your remote into a URL.
6. Submit that `url + ref` combination to VSO.

If your remote is wrong, pass `-Remote name`.
If you want to CI build a different branch, pass `-Ref name|sha1`
If you want to CI build a different branch on a different remote, and skip automatic resolution, pass `-Remote xxx -Ref yyy`.

This script has the capacity for building only certain configurations (`-Config Debug`), and submitting different build types for different remote names. It is configured by a build specification, `.ci-repo.xml`, in the root of the repository.

On first launch, it will bombard you with information about token creation and ask you to create a VSO access token.
If your token breaks, or you want to switch to a different account, `-Reconfigure`. Reconfiguration will not submit a build.

### DIFF
--- a/tools/build/Submit-CIBuilds.ps1
+++ b/tools/build/Submit-CIBuilds.ps1
@@ -1,0 +1,310 @@
+<#
+.SYNOPSIS
+    Submits a CI build to Visual Studio Online given a set of repository-specific criteria.
+#>
+[CmdletBinding(SupportsShouldProcess=$true)]
+Param (
+    [string]$RemoteURL,
+    [string]$Remote,
+    [string]$Ref,
+    [string]$Config,
+    [switch]$Reconfigure,
+    [switch]$SkipRemoteCheck,
+    [string]$BuildFilter,
+
+    [string]$AuthUsername,
+    [string]$AuthToken
+)
+
+Function Coalesce {
+    ForEach($a in $Args) {
+        If($null -ne $a) {
+            return $a;
+        }
+    }
+}
+
+Function Get-GitVersion {
+    $version = (&git.exe --version 2>$null)
+    If ($LASTEXITCODE -NE 0) {
+        Throw "Don't you have git? git --version failed."
+    }
+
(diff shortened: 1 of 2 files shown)


rajsesh-msft:
Finally!

rajsesh-msft:
I am not power shell expert, but having used this for a while, this is good to go IMO.

---
In reply to: [245777362](https://github.com/Microsoft/WinObjC/pull/986#issuecomment-245777362) [](ancestors = 245777362)

rajsesh-msft:
:shipit:

tools/build/Submit-CIBuilds.ps1
```diff
@@ -0,0 +1,310 @@
...
+        If($null -ne $a) {
+            return $a;
+        }
+    }
+}
+
+Function Get-GitVersion {
+    $version = (&git.exe --version 2>$null)
+    If ($LASTEXITCODE -NE 0) {
+        Throw "Don't you have git? git --version failed."
```
aballway: >Throw "Don't you have git? git --version failed." [](start = 8, length = 49)

nit: no rick roll on failure

aballway:
:shipit:

ms-jihua:
:shipit:

DHowett-MSFT:
This does not touch any buildable assets, so I am forgoing CI build.

DHowett-MSFT:
There is an irony in not CI-building a CI-building script.

Issue #729, opened 2016-07-29
Support CTFont and CTFontCollection with proper bridging support


rajsesh-msft:
- [x] Interoperable	CTFontCopyFamilyName
- [x] Interoperable	CTFontGetAscent
- [x] Interoperable	CTFontGetDescent
- [x] Interoperable	CTFontGetSize
- [x] Caveat	CTFontCopyFullName
- [x] Caveat	CTFontCopyName
- [x] Caveat	CTFontCopyPostScriptName
- [x] Caveat	CTFontCreateWithName

ms-jihua:
Note that CTFontCollection has no bridged equivalent.

PR #987, opened 2016-09-09
Initial bridging of CTFontRef and UIFont

ms-jihua:
- Rewrite big chunk of CTFont to use DWrite instead of FreeType
    - Add to DWriteWrapper file
 - Bridge CTFontRef and UIFont, UICTFont introduced
 - Add tests for UIFont and CTFont
    - Moved two UIKit test files that were in an incorrect location
 - Changed instances of 'SegoeUI' in our codebase to 'Segoe UI' - the latter is what DWrite uses for its names

Addresses #729


Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -36,14 +41,16 @@
 static IWLazyClassLookup s_lazyUIFont("UIFont");
 static const float c_defaultSystemFontSize = 15.0f;
 
-static ComPtr<IDWriteFactory> CreateDWriteFactoryInstance() {
+// Private helper for creating a DWriteFactory
+static ComPtr<IDWriteFactory> __CreateDWriteFactoryInstance() {
```
ms-jihua: >__ [](start = 30, length = 2)

Note: Proposing a naming convention here for CT/DW stuff:

two underscores for file-private
one underscore for framework-private

CF I think used a similar pattern, and I think it makes things a little easier to read.

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -14,15 +14,20 @@
...
+
+#import <CoreText/DWriteWrapper.h>
+#import <Starboard.h>
+
 #import <CoreFoundation/CFBase.h>
 #import "CoreTextInternal.h"
-#import "DWriteWrapper.h"
+
+#import <LoggingNative.h>
+#import <mutex>
```
ms-jihua: seems to not be used?

build/CoreText/dll/CoreText.def
```diff
@@ -33,6 +33,17 @@ LIBRARY CoreText
...
         CTGetCoreTextVersion
 
+        ; DWriteWrapper Additions
+        _DWriteGetFontFamilyNames
+        _DWriteGetFontNamesForFamilyName
+        _DWriteGetFamilyNameForFontName
+
+        ; Core Text Constant additions
+        kCTFontLabelFontSize
+        kCTFontButtonFontSize
```
rajsesh-msft: >kCTFontButtonFontSize [](start = 8, length = 21)

should these be DATA?

build/CoreText/lib/CoreTextLib.vcxproj
```diff
@@ -37,7 +37,7 @@
     <ClangCompile Include="$(MSBuildThisFileDirectory)..\..\..\Frameworks\CoreText\DWriteWrapper.mm" />
   </ItemGroup>
   <ItemGroup>
-    <ClInclude Include="$(MSBuildThisFileDirectory)..\..\..\Frameworks\CoreText\DWriteWrapper.h" />
+    <ClInclude Include="$(MSBuildThisFileDirectory)..\..\..\Frameworks\include\CoreText\DWriteWrapper.h" />
```
rajsesh-msft: >DWriteWrapper [](start = 88, length = 13)

Why are we making this header public?  This should be internal only?

rajsesh-msft: never mind, it is still private.

---
In reply to: [78118093](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78118093) [](ancestors = 78118093)

rajsesh-msft:
> static IWLazyClassLookup _LazyUIFont("UIFont");

I was looking forward to this going away, will that happen with #730?

---
Refers to: Frameworks/CoreText/CTFont.mm:49 in 4368641. [](commit_id = 436864149c26fc1a62fd690afaa1c02995ff1b59, deletion_comment = False)

Frameworks/CoreText/CTFont.mm
```diff
@@ -101,25 +110,74 @@
...
+struct __CTFont {
+    CFRuntimeBase _base;
+    ComPtr<IDWriteFont> _dwriteFont;
+    CGFloat _size;
+};
+
+static Boolean __CTFontEqual(CFTypeRef cf1, CFTypeRef cf2) {
+    struct __CTFont* font1 = (struct __CTFont*)cf1;
+    struct __CTFont* font2 = (struct __CTFont*)cf2;
+    return (font1->_size == font2->_size) && (static_cast<IDWriteFont3*>(font1->_dwriteFont.Get())->Equals(font2->_dwriteFont.Get()));
```
rajsesh-msft: >IDWriteFont3 [](start = 58, length = 12)

This is incorrect, you should use QI (ComPtr<IDWriteFont3> f3; font1->_dwriteFount.As(&f3)...)

Frameworks/UIKit/UIFontDescriptor.mm
```diff
@@ -51,7 +51,7 @@
 NSString* const UIFontFeatureTypeIdentifierKey = @"UIFontFeatureTypeIdentifierKey";
 NSString* const UIFontFeatureSelectorIdentifierKey = @"UIFontFeatureSelectorIdentifierKey";
 
-NSString* const SystemFontName = @"SegoeUI";
```
rajsesh-msft: :D. How did this ever work?

Frameworks/UIKit/UIFont.mm
```diff
@@ -362,99 +65,35 @@ + (UIFont*)preferredFontForTextStyle:(NSString*)style {
     return StubReturn();
 }
 
+/**
+ @Status Stub
+*/
 + (UIFont*)fontWithData:(NSData*)data {
```
rajsesh-msft: What is the plan to unstub this?  This would be a regression...

Frameworks/CoreText/CTFont.mm
```diff
@@ -101,25 +110,74 @@
...
+ All these values are fixed number in IOS no matter on iphone* or ipad*.
+ */
+const CGFloat kCTFontLabelFontSize = 17.0f;
+const CGFloat kCTFontButtonFontSize = 14.0f;
+const CGFloat kCTFontSmallSystemFontSize = 12.0f;
+const CGFloat kCTFontSystemFontSize = 14.0f;
+
+using namespace Microsoft::WRL;
+
+struct __CTFont {
```
aballway: >struct __CTFont [](start = 0, length = 15)

nit: typedef so we don't have struct everywhere

Frameworks/CoreText/CTFont.mm
```diff
@@ -101,25 +110,74 @@
...
 /**
- @Status Stub
- @Notes
+ @Status Caveat
+ @Notes matrix parameter not supported
 */
 CTFontRef CTFontCreateWithNameAndOptions(CFStringRef name, CGFloat size, const CGAffineTransform* matrix, CTFontOptions options) {
-    UNIMPLEMENTED();
-    return StubReturn();
+    uint32_t memSize = sizeof(struct __CTFont) - sizeof(CFRuntimeBase);
```
aballway: >uint32_t [](start = 4, length = 8)

nit: size_t

Frameworks/CoreText/CTFont.mm
```diff
@@ -101,25 +110,74 @@
...
+    CTFontRef ret = (CTFontRef)_CFRuntimeCreateInstance(kCFAllocatorDefault, CTFontGetTypeID(), memSize, NULL);
+
+    THROW_IF_FAILED(_DWriteCreateFontWithName(name, &(((struct __CTFont*)ret)->_dwriteFont)));
```
aballway: >(struct __CTFont*) [](start = 55, length = 18)

nit: can these be c++ casts?

Frameworks/CoreText/CTFont.mm
```diff
@@ -225,10 +283,8 @@ CFTypeRef CTFontCopyAttribute(CTFontRef font, CFStringRef attribute) {
 /**
  @Status Interoperable
 */
-float CTFontGetSize(CTFontRef font) {
-    float ret = [(UIFont*)font pointSize];
-
-    return ret;
+CGFloat CTFontGetSize(CTFontRef font) {
+    return font->_size;
```
aballway: >font->_size [](start = 11, length = 11)

check if nil before dereferencing

Frameworks/CoreText/CTFont.mm
```diff
@@ -415,21 +442,27 @@ CFArrayRef CTFontCopySupportedLanguages(CTFontRef font) {
     return StubReturn();
 }
 
+// Private helper for getting a DWRITE_FONT_METRICS struct, to avoid code duplication
+static struct DWRITE_FONT_METRICS __CTFontGetDWriteMetrics(CTFontRef font) {
+    struct DWRITE_FONT_METRICS metrics;
+    font->_dwriteFont->GetMetrics(&metrics);
```
aballway: >font->_dwriteFont->GetMetrics(&metrics) [](start = 4, length = 39)

can we cache these?

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -384,4 +356,269 @@ static void _GetGlyphRunDetails(_CTTypesetter* ts, CFRange range, CGRect frameSi
...
     return lines;
+}
+
+/**
+ * Helper method to retrieve font fmaly names installed in the system.
+ *
+ * @return Unmutable array of font family name strings that are installed in the system.
+ */
+static NSArray<NSString*>* _DWriteGetFontFamilyNames() {
+    NSMutableArray<NSString*>* fontFamilyNames = [NSMutableArray<NSString*> array];
```
aballway: >NSMutableArray<NSString*>* fontFamilyNames [](start = 4, length = 42)

Should we cache this?

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -384,4 +356,269 @@ static void _GetGlyphRunDetails(_CTTypesetter* ts, CFRange range, CGRect frameSi
...
+    return fontFamily;
+}
+
+// Private helper that parses a font name, and returns appropriate weight, stretch, and style values for DWrite accordingly
+static void __InitDWriteFontPropertiesFromName(CFStringRef fontName,
+                                               DWRITE_FONT_WEIGHT* weight,
+                                               DWRITE_FONT_STRETCH* stretch,
+                                               DWRITE_FONT_STYLE* style) {
+    // Set some defaults for when weight/stretch/style are not mentioned in the name
+    *weight = DWRITE_FONT_WEIGHT_NORMAL;
```
aballway: >weight [](start = 5, length = 6)

check if null before dereferencing

tests/unittests/CoreText/CTFontTests.mm
```diff
@@ -88,4 +88,79 @@ virtual void TearDown() {
...
+    CTFontRef font1 = CTFontCreateWithName(CFSTR("Times New Roman"), 12.0, NULL);
+    CTFontRef font2 = CTFontCreateWithName(CFSTR("Times New Roman"), 12.0, NULL);
+    CTFontRef font3 = CTFontCreateWithName(CFSTR("Times New Roman Bold"), 12.0, NULL);
+    CTFontRef font4 = CTFontCreateWithName(CFSTR("Times New Roman"), 14.0, NULL);
+    CFAutorelease(font1);
+    CFAutorelease(font2);
+    CFAutorelease(font3);
+    CFAutorelease(font4);
+
+    ASSERT_TRUE(CFEqual(font1, font2));
```
aballway: >ASSERT [](start = 4, length = 6)

nit: these can be EXPECT

Frameworks/CoreText/DWriteWrapper.mm
mnithish: Fonts can be installed at runtime so technical we should query for it every time

---
In reply to: [78207690](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78207690) [](ancestors = 78207690)

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -384,4 +356,269 @@ static void _GetGlyphRunDetails(_CTTypesetter* ts, CFRange range, CGRect frameSi
...
+    NSMutableArray<NSString*>* fontNames = [NSMutableArray<NSString*> array];
+
+    ComPtr<IDWriteFactory> dwriteFactory = __GetDWriteFactoryInstance();
+
+    // Get the system font collection.
+    ComPtr<IDWriteFontCollection> fontCollection;
+    THROW_IF_FAILED(dwriteFactory->GetSystemFontCollection(&fontCollection));
+
+    // Get the font family.
+    UINT32 index = 0;
```
mnithish: >UINT32 [](start = 4, length = 6)

Coding convention mandates no shouty types

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -384,4 +356,269 @@ static void _GetGlyphRunDetails(_CTTypesetter* ts, CFRange range, CGRect frameSi
...
+    }
+
+    return fontNames;
+}
+
+NSString* _DWriteGetFamilyNameForFontName(CFStringRef fontName) {
+    static NSDictionary<NSString*, NSString*>* fontToFamilyMap = nil;
+
+    static dispatch_once_t initOnce = 0;
+    dispatch_once(&initOnce,
```
mnithish: >dispatch_once(&initOnce, [](start = 4, length = 24)

Are we not going to handle runtime fonts added to the system? May be a note here would be nice

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -384,4 +356,269 @@ static void _GetGlyphRunDetails(_CTTypesetter* ts, CFRange range, CGRect frameSi
...
+        TraceError(TAG, L"Unable to find font family \"%ws\"", unicharFamilyName.data());
+        return fontFamily;
+    }
+
+    THROW_IF_FAILED(systemFontCollection->GetFontFamily(fontFamilyIndex, &fontFamily));
+    return fontFamily;
+}
+
+// Private helper that parses a font name, and returns appropriate weight, stretch, and style values for DWrite accordingly
+static void __InitDWriteFontPropertiesFromName(CFStringRef fontName,
```
mnithish: >__InitDWriteFontPropertiesFromName [](start = 12, length = 34)

Where is this used?

Frameworks/CoreText/CTFont.mm
```diff
@@ -101,25 +110,74 @@
...
+ */
+const CGFloat kCTFontLabelFontSize = 17.0f;
+const CGFloat kCTFontButtonFontSize = 14.0f;
+const CGFloat kCTFontSmallSystemFontSize = 12.0f;
+const CGFloat kCTFontSystemFontSize = 14.0f;
+
+using namespace Microsoft::WRL;
+
+struct __CTFont {
+    CFRuntimeBase _base;
```
DHowett-MSFT: >CFRuntimeBase _base; [](start = 4, length = 20)

This is beautiful.

Frameworks/CoreText/CTFont.mm
```diff
@@ -101,25 +110,74 @@
...
+                                    CTFontGetSize(font));
+}
+
+static CFTypeID __kCTFontTypeID = _kCFRuntimeNotATypeID;
+
+static const CFRuntimeClass __CTFontClass = { 0,
+                                              "CTFont",
+                                              NULL, // init
+                                              NULL, // copy
+                                              NULL, // deallocate (ComPtr takes care of RAII)
```
DHowett-MSFT: >NULL [](start = 46, length = 4)

However, we do need a deallocator here. `~__CTFont()` only gets called if this `__CTFont` is `delete`d.
You'll probably have to implement a deallocator that explicitly placement-deletes the ComPtr and any other non-trivial types:

```
void __CTFontDeallocate(...) {
    thing->_dwriteFont->~ComPtr();
    // ...
}
```

Frameworks/CoreText/CTFont.mm
```diff
@@ -101,25 +110,74 @@
...
+                                    cf,
+                                    [[static_cast<NSString*>(CTFontCopyFamilyName(font)) autorelease] UTF8String],
+                                    CTFontGetSize(font));
+}
+
+static CFTypeID __kCTFontTypeID = _kCFRuntimeNotATypeID;
+
+static const CFRuntimeClass __CTFontClass = { 0,
+                                              "CTFont",
+                                              NULL, // init
```
DHowett-MSFT: >NULL [](start = 46, length = 4)

Since you need a deleter, you need an init:

```
new (&thing->_dwriteFont) ComPtr<IDWriteFont>();
```

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -384,4 +356,269 @@ static void _GetGlyphRunDetails(_CTTypesetter* ts, CFRange range, CGRect frameSi
...
+    *stretch = DWRITE_FONT_STRETCH_NORMAL;
+    *style = DWRITE_FONT_STYLE_NORMAL;
+
+    NSString* familyName = _DWriteGetFamilyNameForFontName(fontName);
+
+    // Relationship of family name -> font name not always consistent
+    // Usually, properties are added to the end (eg: Arial -> Arial Narrow Bold)
+    // However, this is not always the case (eg: Eras ITC -> Eras Bold ITC)
+    // In addition, some fonts with properties are occasionally placed into their own family (eg: Segoe WP SemiLight -> Segoe WP SemiLight)
+    // Try to be more prudent about these edge cases, by looking only at the difference between the font name and family name
```
mnithish: Interesting correlation - wish DWrite had some APIs to do this

Frameworks/CoreText/CTFont.mm
```diff
@@ -101,25 +110,74 @@
...
+    return CFHash(CFAutorelease(CTFontCopyFullName(font))) * CTFontGetSize(font);
+}
+
+static CFStringRef __CTFontCopyDescription(CFTypeRef cf) {
+    CTFontRef font = (CTFontRef)cf;
+    return CFStringCreateWithFormat(kCFAllocatorSystemDefault,
+                                    NULL,
+                                    CFSTR("<CTFont %p>{font-family: %hs; font-size: %.2f}"),
+                                    cf,
+                                    [[static_cast<NSString*>(CTFontCopyFamilyName(font)) autorelease] UTF8String],
```
DHowett-MSFT: > [[static_cast<NSString*>(CTFontCopyFamilyName(font)) autorelease] UTF8String], [](start = 35, length = 79)

You should be able to do this with `%@`, and you won't need to go through Objective-C or UTF8String

build/Tests/UnitTests/UIKit/UIKit.UnitTests.vcxproj
```diff
@@ -258,16 +258,16 @@
...
+    <ClangCompile Include="$(StarboardBasePath)\tests\unittests\UIKit\NSCoder+UIKitAdditionsTest.mm" />
+    <ClangCompile Include="$(StarboardBasePath)\tests\unittests\UIKit\NSLayoutConstraint.mm" />
+    <ClangCompile Include="$(StarboardBasePath)\tests\unittests\UIKit\UIApplication.m" />
+    <ClangCompile Include="$(StarboardBasePath)\tests\unittests\UIKit\UIViewTest.mm" />
+    <ClangCompile Include="$(StarboardBasePath)\tests\unittests\UIKit\NSValue+UIKitAdditionsTests.mm" />
+    <ClangCompile Include="$(StarboardBasePath)\tests\unittests\UIKit\UIColorTests.mm" />
+    <ClangCompile Include="$(StarboardBasePath)\tests\unittests\UIKit\UIFontTests.mm" />
   </ItemGroup>
   <ItemGroup>
     <ClInclude Include="..\..\..\..\tests\unittests\UIKit\NullCompositor.h" />
```
mnithish: >..\..\..\.. [](start = 24, length = 11)

nit: $(StarboardBasePath)

Frameworks/CoreText/CTFont.mm
mnithish: +1

---
In reply to: [78205733](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78205733) [](ancestors = 78205733)

DHowett-MSFT:
> static const wchar_t* c_defaultFontName = L"Gabriola";

whoa, Gabriola? Is it possible to get the system's default UI font name?

---
Refers to: Frameworks/CoreText/DWriteWrapper.mm:37 in 4368641. [](commit_id = 436864149c26fc1a62fd690afaa1c02995ff1b59, deletion_comment = False)

Frameworks/CoreText/CTFont.mm
```diff
@@ -101,25 +110,74 @@
...
- @Status Stub
- @Notes
+ @Status Caveat
+ @Notes matrix parameter not supported
 */
 CTFontRef CTFontCreateWithNameAndOptions(CFStringRef name, CGFloat size, const CGAffineTransform* matrix, CTFontOptions options) {
-    UNIMPLEMENTED();
-    return StubReturn();
+    uint32_t memSize = sizeof(struct __CTFont) - sizeof(CFRuntimeBase);
+    CTFontRef ret = (CTFontRef)_CFRuntimeCreateInstance(kCFAllocatorDefault, CTFontGetTypeID(), memSize, NULL);
```
mnithish: >(CTFontRef) [](start = 20, length = 11)

static_cast<>

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -384,4 +356,269 @@ static void _GetGlyphRunDetails(_CTTypesetter* ts, CFRange range, CGRect frameSi
...
+}
+
+/**
+ * Helper method to retrieve font fmaly names installed in the system.
+ *
+ * @return Unmutable array of font family name strings that are installed in the system.
+ */
+static NSArray<NSString*>* _DWriteGetFontFamilyNames() {
+    NSMutableArray<NSString*>* fontFamilyNames = [NSMutableArray<NSString*> array];
+
```
DHowett-MSFT: If you implement this in terms of CFArray, we _might_ eventually be able to break the Foundation dependency fro CoreText

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -384,4 +356,269 @@ static void _GetGlyphRunDetails(_CTTypesetter* ts, CFRange range, CGRect frameSi
...
+    // Relationship of family name -> font name not always consistent
+    // Usually, properties are added to the end (eg: Arial -> Arial Narrow Bold)
+    // However, this is not always the case (eg: Eras ITC -> Eras Bold ITC)
+    // In addition, some fonts with properties are occasionally placed into their own family (eg: Segoe WP SemiLight -> Segoe WP SemiLight)
+    // Try to be more prudent about these edge cases, by looking only at the difference between the font name and family name
+    NSArray<NSString*>* fontNameTokens = [(__bridge NSString*)fontName componentsSeparatedByString:@" "];
+    NSMutableSet<NSString*>* propertyTokens = [NSMutableSet<NSString*> setWithArray:fontNameTokens];
+
+    if (familyName) {
+        NSArray<NSString*>* familyNameTokens = [familyName componentsSeparatedByString:@" "];
```
DHowett-MSFT: >familyName [](start = 48, length = 10)

Edge case: There may be a font called something silly like "Black Light", but it's not black- or light-weighted.

Also, I didn't realize "Light" was also a weight when I was making that example :(

Frameworks/CoreText/DWriteWrapper.mm
aballway: I think the plan was to get it breathing first, then cut the cord

---
In reply to: [78220974](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78220974) [](ancestors = 78220974)

Frameworks/CoreText/DWriteWrapper.mm
DHowett-MSFT: Or well, Black Light Light, if it is light-weighted. I think that would superexplode the weight guesser.

---
In reply to: [78221211](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78221211) [](ancestors = 78221211)

Frameworks/UIKit/UIFont.mm
```diff
@@ -856,102 +321,33 @@ - (instancetype)copyWithZone:(NSZone*)zone {
...
 - (uint32_t)_sizingFontHandle {
-    return (uint32_t)_sizingFont;
+    UNIMPLEMENTED();
+    return StubReturn();
 }
 
 // Internal methods
 
 // Private message sent from CTFontManager for the implementation of CTFontManagerRegisterGraphicsFont
 - (bool)_CTFontManagerRegisterGraphicsFont:(CGFontRef)font withError:(CFErrorRef*)error {
```
mnithish: >_CTFontManagerRegisterGraphicsFont [](start = 8, length = 34)

Can we remove these private methods and update their callers instead so we know what public facing APIs have regressed with this?

Frameworks/UIKit/UIFont.mm
mnithish: May be we should have a TODO for such things that we can revisit and have a work item or bug to fix

---
In reply to: [78118792](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78118792) [](ancestors = 78118792)

mnithish:
> static const wchar_t* c_defaultFontName = L"Gabriola";

I have to be blamed for that. There is no API to get the default font name as that is unique for every application. Most Windows applications use Calibri or Segoe UI

---
In reply to: [245987714](https://github.com/Microsoft/WinObjC/pull/987#issuecomment-245987714) [](ancestors = 245987714)

---
Refers to: Frameworks/CoreText/DWriteWrapper.mm:37 in 4368641. [](commit_id = 436864149c26fc1a62fd690afaa1c02995ff1b59, deletion_comment = False)

Frameworks/CoreText/DWriteWrapper.mm
mnithish: makes sense to follow this convention. May be this should be added to our coding style guided after formal approval from the team

---
In reply to: [78114123](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78114123) [](ancestors = 78114123)

Frameworks/UIKit/UIFont.mm
ms-jihua: I think I'd rather do that in a separate CR? For example, for this function, it probably makes more sense to do the full stripping at the same time as bringing up CTFontManager. Doing it in the current state would probably generate a lot of unnecessary extra work just to keep things building.

---
In reply to: [78222243](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78222243) [](ancestors = 78222243)

Frameworks/CoreText/DWriteWrapper.mm
ms-jihua: If that is the case, the family name would be something like 'Black' or 'Black Light', and the remaining tokens would be '', 'Light', or 'Light Light' (have not seen anything like the last one). The weight guesser would still handle those cases correctly (though it would handle 1x Light the same as 2x Light)

---
In reply to: [78221297](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78221297) [](ancestors = 78221297,78221211)

ms-jihua: If that is the case, the family name would be something like 'Black' or 'Black Light', and the remaining tokens would be '', 'Light', or 'Light Light' (have not seen anything like the last one). The weight guesser would still handle those cases correctly (though it would handle 1x Light the same as 2x Light)

---
In reply to: [78221297](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78221297) [](ancestors = 78221297,78221211)

Frameworks/UIKit/UIFont.mm
mnithish: Np but we should remember that now some of the public methods are actual STUBs and you have kept track of these then we are good

---
In reply to: [78224805](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78224805) [](ancestors = 78224805,78222243)

Frameworks/CoreText/DWriteWrapper.mm
ms-jihua: I actually spent a bunch of effort converting this from CFArray _to_ NSArray so that this file would be consistent T____T.

I do have versions of the functions I added that are more CF-style. Is our long term goal to make this file more CF-y? (do we care about that?) If so, do I want to also change the existing functions in this file to be more CFy?

---
In reply to: [78221279](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78221279) [](ancestors = 78221279,78220974)

Frameworks/CoreText/CTFont.mm
ms-jihua: These are in the same style as Create functions from CoreFoundation - I think there's probably merit in keeping it consistent with CF?

---
In reply to: [78206115](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78206115) [](ancestors = 78206115)

Frameworks/CoreText/CTFont.mm
ms-jihua: These are in the same style as Create functions from CoreFoundation - I think there's probably merit in keeping it consistent with CF?

---
In reply to: [78220932](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78220932) [](ancestors = 78220932)

Frameworks/CoreText/CTFont.mm
ms-jihua: These are in the same style as Create functions from CoreFoundation - I think there's probably merit in keeping it consistent with CF?

---
In reply to: [78206408](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78206408) [](ancestors = 78206408)

ms-jihua:
> static IWLazyClassLookup _LazyUIFont("UIFont");

That's more or less the plan - it's still a bit entrenched right now and it'd be extra work to leave a building state but remove just this.

---
In reply to: [245794924](https://github.com/Microsoft/WinObjC/pull/987#issuecomment-245794924) [](ancestors = 245794924)

---
Refers to: Frameworks/CoreText/CTFont.mm:49 in 4368641. [](commit_id = 436864149c26fc1a62fd690afaa1c02995ff1b59, deletion_comment = False)

Frameworks/CoreText/CTFont.mm
ms-jihua: - Is the same style as CF
- For most purposes, since this is already def'd to CTFontRef, you can just use that
- The only times when 'struct' is used is when directly modifying the struct members, and I don't dislike the emphasis there

---
In reply to: [78220566](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78220566) [](ancestors = 78220566,78205733)

Frameworks/CoreText/CTFont.mm
ms-jihua: Thanks but this is mostly just a copy of CF structs... ^^;;

---
In reply to: [78219486](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78219486) [](ancestors = 78219486)

Frameworks/CoreText/CTFont.mm
ms-jihua: Ohhh. I wasn't sure what the correct way was to access the 1,2,3 subclasses. Thanks!

---
In reply to: [78118540](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78118540) [](ancestors = 78118540)

Frameworks/CoreText/CTFont.mm
ms-jihua: Oops, I was trying to stay away from %@ since TraceError has problems with it, and I accidentally hit this I guess.

---
In reply to: [78220394](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78220394) [](ancestors = 78220394)

Frameworks/CoreText/CTFont.mm
ms-jihua: good point

---
In reply to: [78206516](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78206516) [](ancestors = 78206516)

Frameworks/CoreText/CTFont.mm
```diff
@@ -259,124 +315,95 @@ CFDictionaryRef CTFontCopyTraits(CTFontRef font) {
...
-
-    // Style name is not present in the name table associated with the font.
-    // It is stored in FT_Face.
-    if (nameKey == kCTFontStyleNameKey) {
-        return static_cast<CFStringRef>([[NSString stringWithCString:face->style_name] retain]);
+    ComPtr<IDWriteLocalizedStrings> name;
+    BOOL exists;
+    DWRITE_INFORMATIONAL_STRING_ID informationalStringId;
+
+    if (CFEqual(nameKey, kCTFontCopyrightNameKey)) {
```
ms-jihua: should I use a dictionary for this instead?

Frameworks/CoreText/CTFont.mm
ms-jihua: yeah, I think so

---
In reply to: [78206815](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78206815) [](ancestors = 78206815)

Frameworks/CoreText/DWriteWrapper.mm
ms-jihua: It's a file-private helper, we can assume that the pointers are good. I'll add a comment.

---
In reply to: [78207940](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78207940) [](ancestors = 78207940)

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -384,4 +356,269 @@ static void _GetGlyphRunDetails(_CTTypesetter* ts, CFRange range, CGRect frameSi
...
+        [propertyTokens minusSet:[NSSet setWithArray:familyNameTokens]];
+    }
+
+    for (NSString* propertyToken in propertyTokens) {
+        NSString* upperPropertyToken = [propertyToken uppercaseString];
+
+        if ([upperPropertyToken isEqual:@"THIN"]) {
```
ms-jihua: >if ([upperPropertyToken isEqual:@"THIN"]) { [](start = 8, length = 43)

Do I want to use a dictionary to do this instead?

Frameworks/CoreText/CTFont.mm
mnithish: Sorry for being a pain here but we won't follow CF* style in our files

---
In reply to: [78225659](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78225659) [](ancestors = 78225659,78206115)

build/CoreText/dll/CoreText.def
ms-jihua: oops yes

---
In reply to: [78118009](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78118009) [](ancestors = 78118009)

Frameworks/UIKit/UIFontDescriptor.mm
ms-jihua: I think Segoe UI was the only font we supported, so we hardcoded it in to work at some lower level when interfacing with FreeType? In any case, DWrite tells us the names have spaces, so spaces it is.

---
In reply to: [78118727](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78118727) [](ancestors = 78118727)

Frameworks/CoreText/CTFont.mm
ms-jihua: Very well

---
In reply to: [78228222](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78228222) [](ancestors = 78228222,78225659,78206115)

mnithish:
:clock1:

Frameworks/CoreText/CTFont.mm
ms-jihua: Actually no I changed my mind. If the caller is using this very plain C-style function on a null pointer, I think they probably deserve the AV that they get?

---
In reply to: [78227303](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78227303) [](ancestors = 78227303,78206516)

Frameworks/CoreText/CTFont.mm
ms-jihua: discussed offline, leaving as-is for now - it's working and we can't actually figure out the syntax for a new

---
In reply to: [78219961](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78219961) [](ancestors = 78219961)

Frameworks/UIKit/UIFont.mm
ms-jihua: This turns out to be a WinObjC-only function anyway - it's not something we have to support. It is used in UIApplicationMain and CGFont as what looks like a convenience method for their own purposes?

---
In reply to: [78222890](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78222890) [](ancestors = 78222890,78118792)

tests/unittests/CoreText/CTFontTests.mm
```diff
@@ -88,4 +88,97 @@ virtual void TearDown() {
...
+    // Test for reasonable bounds on number of unique glyphs in a font
+    EXPECT_LE(3000, CTFontGetGlyphCount(font));
+    EXPECT_GE(8000, CTFontGetGlyphCount(font));
+}
+
+TEST(CTFont, UnknownName) {
+    EXPECT_OBJCEQ(nil, (id)CFAutorelease(CTFontCreateWithName(CFSTR("DoesNotExistFont"), 12.0, NULL)));
+}
+
+// TODO: Do we care enough about fixing this?
```
ms-jihua: Calling this out explicitly - do we care about this?

Frameworks/CoreText/DWriteWrapper.mm
aballway: seconded

---
In reply to: [78223447](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78223447) [](ancestors = 78223447,78114123)

Frameworks/CoreText/DWriteWrapper.mm
aballway: probably.  I imagine lookup would be faster than all of the string comparisons it would take to be italic

---
In reply to: [78228119](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78228119) [](ancestors = 78228119)

Frameworks/CoreText/DWriteWrapper.mm
aballway: Long-term we should probably cut out all NS stuff and convert everything to CF, but for now I'm doing whatever works best, as converting should be reasonably painless

---
In reply to: [78225402](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78225402) [](ancestors = 78225402,78221279,78220974)

Frameworks/CoreText/CTFont.mm
aballway: probably, same reason

---
In reply to: [78227559](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78227559) [](ancestors = 78227559)

Frameworks/CoreText/DWriteWrapper.mm
ms-jihua: well I mean obv, but it'd also
a) be a memory hit for a feature the user may never use
b) possibly be more difficult to read
currently leaning towards 'no' actually, but I'd like to see what a few more people think

---
In reply to: [78610324](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78610324) [](ancestors = 78610324,78228119)

Frameworks/CoreText/DWriteWrapper.mm
ms-jihua: If we're committed to that for the long term, I guess I'll bring back in the CF versions of the code I wrote.

---
In reply to: [78610729](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78610729) [](ancestors = 78610729,78225402,78221279,78220974)

Frameworks/CoreText/DWriteWrapper.mm
mnithish: I did not find one while looking for it. Put a note here and I don't think this is a mad assumption to make

---
In reply to: [78232112](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78232112) [](ancestors = 78232112,78218544)

Frameworks/CoreText/DWriteWrapper.mm
mnithish: I do not think moving to CF is a pressing requirement now. We have a lot more work to get CT standing before diverting our focus to this

---
In reply to: [78611079](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78611079) [](ancestors = 78611079,78610729,78225402,78221279,78220974)

Frameworks/CoreText/DWriteWrapper.mm
mnithish: This is more readable and I don't think moving to dictionary will give any noticeable runtime performance 

---
In reply to: [78610958](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78610958) [](ancestors = 78610958,78610324,78228119)

tests/unittests/CoreText/CTFontTests.mm
mnithish: Let us keep the note and we can take a call when we obsoletely know we cannot implement this interface

---
In reply to: [78609035](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78609035) [](ancestors = 78609035)

Frameworks/CoreText/DWriteWrapper.mm
ms-jihua: There's already one in the function comment :)

---
In reply to: [78616872](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78616872) [](ancestors = 78616872,78232112,78218544)

Frameworks/CoreText/DWriteWrapper.mm
ms-jihua: It was more that I _already_ have the CF versions of some of these functions in another branch, so I might as well move them over now. It won't be too much effort.

---
In reply to: [78617853](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78617853) [](ancestors = 78617853,78611079,78610729,78225402,78221279,78220974)

tests/unittests/CoreText/CTFontTests.mm
ms-jihua: It was less 'cannot' and more 'this seems annoying to implement for very little gain'. Keeping the note here, then.

---
In reply to: [78618340](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78618340) [](ancestors = 78618340,78609035)

Frameworks/UIKit/UIFont.mm
```diff
@@ -490,41 +129,21 @@ + (UIFont*)systemFontOfSize:(float)size {
...
 /**
  @Status Interoperable
 */
-+ (UIFont*)boldSystemFontOfSize:(float)size {
++ (UIFont*)boldSystemFontOfSize:(CGFloat)size {
     // TODO 5785385: Using clumsy fontWithDescriptor to initialize here, so that _descriptor is initialized
     // Clean this up a bit once fontDescriptor gets better support
     UIFontDescriptor* fontDes =
-        [[UIFontDescriptor fontDescriptorWithName:@"Helvetica Bold" size:size] fontDescriptorWithSymbolicTraits:UIFontDescriptorTraitBold];
+        [[UIFontDescriptor fontDescriptorWithName:@"Arial Bold" size:size] fontDescriptorWithSymbolicTraits:UIFontDescriptorTraitBold];
```
DHowett-MSFT: >Arial [](start = 52, length = 5)

Should we store the system font names somewhere consolidated?

DHowett-MSFT: Oh, I see: we have one. Arial doesn't match though o_O

---
In reply to: [78639264](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78639264) [](ancestors = 78639264)

ms-jihua: Doesn't match? The default font in this file seems to be Arial?

---
In reply to: [78639398](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78639398) [](ancestors = 78639398,78639264)

ms-jihua: Or wait you probably mean the Gabriola in DWriteWrapper?

---
In reply to: [78640310](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78640310) [](ancestors = 78640310,78639398,78639264)

DHowett-MSFT: And Segoe in UIFontDescriptor

---
In reply to: [78640452](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78640452) [](ancestors = 78640452,78640310,78639398,78639264)

DHowett-MSFT:
:shipit:

Frameworks/UIKit/UIFont.mm
mnithish: I have fixed the default font to be Helvetica as it used to be.

---
In reply to: [78640513](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78640513) [](ancestors = 78640513,78640452,78640310,78639398,78639264)

ms-jihua: Bit concerned about that actually - the reason why I switched off of Helvetica was because I queried my list of installed fonts through DWrite, and Helvetica actually wasn't in it? So I thought we should probably pick a different default.

---
In reply to: [78640879](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78640879) [](ancestors = 78640879,78640513,78640452,78640310,78639398,78639264)

ms-jihua: Let me file an issue on myself so I don't forget, and I'll address consolidation in a separate CR.

---
In reply to: [78641706](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78641706) [](ancestors = 78641706,78640879,78640513,78640452,78640310,78639398,78639264)

ms-jihua: #1006

---
In reply to: [78648377](https://github.com/Microsoft/WinObjC/pull/987#discussion_r78648377) [](ancestors = 78648377,78641706,78640879,78640513,78640452,78640310,78639398,78639264)

PR #990, opened 2016-09-09
Fix bad merge

rajsesh-msft:
There was a bad merge to the functional tests filters project that caused the project to be not usable from solution explorer.


DHowett-MSFT:
:shipit:

PR #992, opened 2016-09-09
Adding a FontKey class for indexing the global font cache

pittmab:
- the global font cache was never finding the UIFont in the list since the isEqual method of UIFont compares the fontDescriptor and returns false if a descriptor is nil.

Fix: 991

### DIFF
--- a/Frameworks/UIKit/UIFont.mm
+++ b/Frameworks/UIKit/UIFont.mm
@@ -73,6 +73,47 @@ void* FTReallocFunc(FT_Memory memory, long curSize, long newSize, void* ptr) {
     return IwRealloc(ptr, newSize);
 }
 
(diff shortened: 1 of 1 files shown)


rajsesh-msft:

ms-jihua:
Hi Brent, we're actually in the middle of a more in-depth refactor of CoreText and UIFont right now, including a fix to the isEqual: function. Those changes are going to sit in the CT-DW branch for a while until they stabilize, so feel free to keep tabs on that.

Regarding this PR, from my understanding, you don't actually need a fix for this issue urgently? If that's the case, could you wait for the more in-depth fix we've got going in the other branch?

pittmab:
We do not need a fix for this immediately.  If it is going to be fixed by the CT-DW branch then we can keep the change locally and discard it once that branch makes it into master.

PR #994, opened 2016-09-09
Fixing a bug with the memory of the elements for a CGPath

pittmab:
- When we hit the bounds of the elements and call resize we need to reinitialize the elements pointers.

Fix:  993

### DIFF
--- a/Frameworks/CoreGraphics/CGPath.mm
+++ b/Frameworks/CoreGraphics/CGPath.mm
@@ -493,6 +493,12 @@ void CGPathAddPath(CGMutablePathRef path, const CGAffineTransform* m, CGPathRef 
             }
         }
     }
+
+    // we need to re init the elements to setup the pointers propertly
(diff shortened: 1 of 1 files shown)


Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -493,6 +493,12 @@ void CGPathAddPath(CGMutablePathRef path, const CGAffineTransform* m, CGPathRef
             }
         }
     }
+
+	// we need to re init the elements to setup the pointers propertly
```
rajsesh-msft: nit: could you please convert tabs to spaces? change looks good otherwise.

pittmab: @rajsesh-msft Converted the tab to spaces.  Thanks

rajsesh-msft:
:shipit:

PR #1002, opened 2016-09-13
Adding the ability to subclass UIGestureRecognizer

pittmab:
- removing the hard coded preference of gestures
- now enumerating the list of gestures in the order in which they are added
- supporting some of the callbacks for UIGestureRecognizerDelegate
- Adding a callback to check if the first active gesture can prevent other potential gestures in the gesture recognizer list.

Fix: 489
Fix: 475

---

This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1002)

### DIFF
--- a/Frameworks/UIKit/UIView.mm
+++ b/Frameworks/UIKit/UIView.mm
@@ -25,6 +25,7 @@
 #import "UIWindowInternal.h"
 #import "UIViewControllerInternal.h"
 #import "UIGestureRecognizerInternal.h"
+#import "UIKit/UIGestureRecognizerSubclass.h"
 #import "CALayerInternal.h"
 #import "CAAnimationInternal.h"
 #import <AutoLayout.h>
@@ -198,20 +199,67 @@ BOOL g_resetAllTrackingGestures = TRUE;
         }
     }
 
-    // gesture priority list
-    const static id s_gesturesPriority[] = {[UIPinchGestureRecognizer class],
-                                            [UISwipeGestureRecognizer class],
-                                            [UIPanGestureRecognizer class],
-                                            [UILongPressGestureRecognizer class],
-                                            [UITapGestureRecognizer class] };
+    //
+    // Remove any gestures that should not simulatanously be recognized
+    //
 
-    const static int s_numGestureTypes = sizeof(s_gesturesPriority) / sizeof(s_gesturesPriority[0]);
+    // First find the active gesture
+    UIGestureRecognizer* activeGesture = nil;
 
-    //  Process all gestures
-    for (int i = 0; i < s_numGestureTypes; i++) {
-        id curgestureClass = s_gesturesPriority[i];
-        id gestures = [g_curGesturesDict objectForKey:curgestureClass];
-        if ([curgestureClass _fireGestures:gestures]) {
+    int count = [g_currentlyTrackingGesturesList count];
+    for (UIGestureRecognizer* curGesture in [g_currentlyTrackingGesturesList reverseObjectEnumerator]) {
+        UIGestureRecognizerState state = curGesture.state;
+        if (state != UIGestureRecognizerStatePossible && state != UIGestureRecognizerStateCancelled) {
+            activeGesture = curGesture;
+            break;
+        }
+    }
+
+    if (activeGesture) {
+        BOOL activeResponds =
+            [activeGesture.delegate respondsToSelector:@selector(gestureRecognizer:shouldRecognizeSimultaneouslyWithGestureRecognizer:)];
+
+        // now test it against the other gestures, removing if we do not handle simultaneous gestures
+        NSMutableSet* gesturesToRemove = [NSMutableSet set];
+        for (UIGestureRecognizer* curGesture in g_currentlyTrackingGesturesList) {
+            if (curGesture == activeGesture) {
+                continue;
+            }
+            BOOL currentResponds =
+                [curGesture.delegate respondsToSelector:@selector(gestureRecognizer:shouldRecognizeSimultaneouslyWithGestureRecognizer:)];
+
+            if (currentResponds == NO ||
+                [curGesture.delegate gestureRecognizer:curGesture shouldRecognizeSimultaneouslyWithGestureRecognizer:activeGesture] == NO) {
+                [gesturesToRemove addObject:curGesture];
+            } else if (activeResponds == NO ||
+                       [activeGesture.delegate gestureRecognizer:curGesture
+                           shouldRecognizeSimultaneouslyWithGestureRecognizer:activeGesture] == NO) {
+                [gesturesToRemove addObject:curGesture];
+            }
+
+            BOOL canPrevent = [activeGesture canPreventGestureRecognizer:curGesture];
+            if (canPrevent == YES) {
+                [gesturesToRemove addObject:curGesture];
+            }
+        }
+
+        // remove those gesturee
+        for (UIGestureRecognizer* curGesture in gesturesToRemove) {
+            [curGesture reset];
+            if (DEBUG_GESTURES) {
+                TraceVerbose(TAG, L"Removing gesture %hs %x state=%d", object_getClassName(curGesture), curGesture);
+            }
+            [g_currentlyTrackingGesturesList removeObject:curGesture];
+            id gesturesArr = [g_curGesturesDict objectForKey:[curGesture class]];
+            [gesturesArr removeObject:curGesture];
+        }
+    }
+
+    // to support the subclassing of gestures, fire the gestures based on their order in the tracking list
+    count = [g_currentlyTrackingGesturesList count];
+    for (UIGestureRecognizer* curGesture in [g_currentlyTrackingGesturesList reverseObjectEnumerator]) {
+        id curgestureClass = [curGesture class];
+        if ([curgestureClass _fireGestures:@[ curGesture ]]) {
             handled = true;
             if (handled && DEBUG_GESTURES) {
                 TraceVerbose(TAG, L"Gesture (%hs) handled.", object_getClassName(curgestureClass));


Frameworks/UIKit/UIView.mm
```diff
@@ -198,27 +199,68 @@ - (bool)_processGesturesForTouch:(UITouch*)touch event:(UIEvent*)event touchEven
...
-        id gestures = [g_curGesturesDict objectForKey:curgestureClass];
-        if ([curgestureClass _fireGestures:gestures]) {
-            handled = true;
-            if (handled && DEBUG_GESTURES) {
-                TraceVerbose(TAG, L"Gesture (%hs) handled.", object_getClassName(curgestureClass));
+    // First find the active gesture
+    UIGestureRecognizer* activeGesture = nil;
+
+    int count = [g_currentlyTrackingGesturesList count];
+    for (int i = count - 1; i >= 0; i--) {
```
aballway: nit: this could be converted into a for in loop

Frameworks/UIKit/UIView.mm
```diff
@@ -198,27 +199,68 @@ - (bool)_processGesturesForTouch:(UITouch*)touch event:(UIEvent*)event touchEven
...
+            activeGesture = curGesture;
+            break;
+        }
+    }
+
+    if ( activeGesture ) {
+        BOOL activeResponds = [activeGesture.delegate respondsToSelector:@selector(gestureRecognizer:shouldRecognizeSimultaneouslyWithGestureRecognizer:)];
+
+        // now test it against the other gestures, removing if we do not handle simultaneous gestures
+        NSMutableSet* gesturesToRemove = [[NSMutableSet alloc] init];
```
aballway: >[NSMutableSet alloc] init [](start = 42, length = 25)

nit: [NSMutableSet set]

Frameworks/UIKit/UIView.mm
```diff
@@ -198,27 +199,68 @@ - (bool)_processGesturesForTouch:(UITouch*)touch event:(UIEvent*)event touchEven
...
+            TraceVerbose(TAG, L"Removing gesture %hs %x state=%d", object_getClassName(curGesture), curGesture );
+            [g_currentlyTrackingGesturesList removeObject:curGesture];
+            id gesturesArr = [g_curGesturesDict objectForKey:[curGesture class]];
+            [gesturesArr removeObject:curGesture];
         }
     }
 
+    // to support the subclassing of gestures, fire the gestures based on their order in the tracking list
+    count = [g_currentlyTrackingGesturesList count];
+    for (int i = count - 1; i >= 0; i--) {
```
aballway: same as above

Frameworks/UIKit/UIView.mm
pittmab: @aballway I am guessing that you would prefer that we use a reverseObjectEnumerator to traverse in reverse order.  I have noticed that the code in this file uses a mixture of old for counting loops and for in loops.  Should all new code be for in when it can?  I initially used the style I found in the existing code.

aballway: I think it's better style to use for in loops unless the index is needed or the entire file is in counting style.  Given this file is mixed either one would work, but I believe using for/in is clearer.

---
In reply to: [78584341](https://github.com/Microsoft/WinObjC/pull/1002#discussion_r78584341) [](ancestors = 78584341)

pittmab: @aballway I agree that for in is quicker to read so I will make the changes.  Just curious as to what your team prefers given the mix in this file.  I will make the changes.

jaredhms:
@davelamb, please take a look.

Frameworks/UIKit/UIView.mm
```diff
@@ -198,27 +199,68 @@ - (bool)_processGesturesForTouch:(UITouch*)touch event:(UIEvent*)event touchEven
...
+        UIGestureRecognizer* curGesture = [g_currentlyTrackingGesturesList objectAtIndex:i];
+        UIGestureRecognizerState state = curGesture.state;
+        if ( state != UIGestureRecognizerStatePossible && state != UIGestureRecognizerStateCancelled) {
+            activeGesture = curGesture;
+            break;
+        }
+    }
+
+    if ( activeGesture ) {
```
jaredhms: > activeGesture  [](start = 8, length = 15)

super-nit: several spots with inconsistent spacing; it may be beneficial to run pre_commit(.ps1) or pre_push(.ps1) which would run clang-format over your changes.

Frameworks/UIKit/UIView.mm
```diff
@@ -198,27 +199,68 @@ - (bool)_processGesturesForTouch:(UITouch*)touch event:(UIEvent*)event touchEven
...
+            BOOL canPrevent = [activeGesture canPreventGestureRecognizer:curGesture] ;
+            if ( canPrevent == YES ) {
+                [gesturesToRemove addObject:curGesture];
+            }
+        }
+
+        // remove those gesturee
+        for ( UIGestureRecognizer* curGesture in gesturesToRemove ) {
+            [curGesture reset];
+            TraceVerbose(TAG, L"Removing gesture %hs %x state=%d", object_getClassName(curGesture), curGesture );
```
jaredhms: >   TraceVerbose(TAG, L"Removing gesture %hs %x state=%d", object_getClassName(curGesture), curGesture ); [](start = 9, length = 104)

can you please wrap this in an 'if (DEBUG_GESTURES) {}' ?

Frameworks/UIKit/UIView.mm
pittmab: @jaredhms I just ran it and it doesn't indicate that any changes are necessary.  Is there something that I am missing?  Thanks

jaredhms: do you have this file staged with changes?  

---
In reply to: [78603657](https://github.com/Microsoft/WinObjC/pull/1002#discussion_r78603657) [](ancestors = 78603657)

Frameworks/UIKit/UIView.mm
```diff
@@ -198,27 +199,68 @@ - (bool)_processGesturesForTouch:(UITouch*)touch event:(UIEvent*)event touchEven
...
         }
     }
 
+    // to support the subclassing of gestures, fire the gestures based on their order in the tracking list
+    count = [g_currentlyTrackingGesturesList count];
+    for (int i = count - 1; i >= 0; i--) {
+        UIGestureRecognizer* curGesture = [g_currentlyTrackingGesturesList objectAtIndex:i];
+            id curgestureClass = [curGesture class];
+            if ([curgestureClass _fireGestures:@[curGesture]]) {
+                handled = false;
```
jaredhms: >false [](start = 26, length = 5)

This should still be 'true', shouldn't it?

Also can we please add the old tracing back?

            if (handled && DEBUG_GESTURES) {
                TraceVerbose(TAG, L"Gesture (%hs) handled.", object_getClassName(curgestureClass));
            }

Frameworks/UIKit/UIView.mm
pittmab: I edited the file and ran the pre-commit check.  The output was Performing pre-commit validation...  Commit approved.

jaredhms: Strange; it *should have* modified the file in place to remove the spacing inconsistencies.  it might be worth trying .\tools\scripts\clang-format.ps1 -File <path to file> to see if you get different results?

---
In reply to: [78604452](https://github.com/Microsoft/WinObjC/pull/1002#discussion_r78604452) [](ancestors = 78604452)

pittmab: Yes.  I have to run clang-format.  Thanks

jaredhms: Thanks - that's a bit unfortunate.  Might be worth trying pre_push next time (after you commit, before you push)...

---
In reply to: [78605310](https://github.com/Microsoft/WinObjC/pull/1002#discussion_r78605310) [](ancestors = 78605310)

Frameworks/UIKit/UIView.mm
pittmab: @jaredhms Added and other changes have been made.  Please take a look.

Frameworks/UIKit/UIView.mm
```diff
@@ -198,24 +199,68 @@ - (bool)_processGesturesForTouch:(UITouch*)touch event:(UIEvent*)event touchEven
...
+            [gesturesArr removeObject:curGesture];
+        }
+    }
+
+    // to support the subclassing of gestures, fire the gestures based on their order in the tracking list
+    count = [g_currentlyTrackingGesturesList count];
+    for (UIGestureRecognizer* curGesture in [g_currentlyTrackingGesturesList reverseObjectEnumerator]) {
+        id curgestureClass = [curGesture class];
+        if ([curgestureClass _fireGestures:@[ curGesture ]]) {
+            handled = false;
```
jaredhms: This should still be 'true', shouldn't it?

Also can we please add the old tracing back?

            if (handled && DEBUG_GESTURES) {
                TraceVerbose(TAG, L"Gesture (%hs) handled.", object_getClassName(curgestureClass));
            }

Frameworks/UIKit/UIView.mm
```diff
@@ -198,24 +199,68 @@ - (bool)_processGesturesForTouch:(UITouch*)touch event:(UIEvent*)event touchEven
         }
     }
 
-    // gesture priority list
-    const static id s_gesturesPriority[] = {[UIPinchGestureRecognizer class],
-                                            [UISwipeGestureRecognizer class],
-                                            [UIPanGestureRecognizer class],
-                                            [UILongPressGestureRecognizer class],
-                                            [UITapGestureRecognizer class] };
```
davelamb: Removing this probably breaks gesture chaining for UIPanGestureRecognizer and UITapGestureRecognizer. That is to say, you haven't implemented the delegates on the system gesture recognizers.

pittmab: So I previously wrote some simple tests to see what happens on iOS with the gesture lists.  From what I can tell the order in the list gives the priority.  The gesture delegates and other callbacks (e.g. simultaneous recognizing, prevention, etc) help control the recognition.  I haven't implemented the full delegate/coordination api across all gestures.  We do not use all of them, do you have a full test suite?  I think this implementation is a potential step in the right direction since we rely on creating our own gestures.  Any thoughts on next steps?

Frameworks/UIKit/UIView.mm
pittmab: @jaredhms Fixed and the original debugging code put back.  Thanks

jaredhms:
:shipit:

Frameworks/UIKit/UIView.mm
davelamb: You're not wrong re: the order of things (even though the iOS docs say the order in which they get serviced is undefined behaviour, it does appear to be deterministic). We don't have a test suite, since our UI automation story hasn't been fleshed out yet, so everything is manual testing for the time being. 

After a little bit of messing around with iOS, it looks like returning NO from `canPreventGestureRecognizer` and `canBePreventedByGestureRecognizer` preempts the calls to the delegate.. that is to say, you should be nesting your calls to the delegate inside a check to see if the gesture can be prevented by other gestures. 

I bring this up specifically because I'd like an avenue to keep the old behaviour while this work gets finished up, since as of this change only one gesture will be fired in most cases (where previously we'd fire all gestures). To enable that, I'd like you to nest that logic, and add a default implementation to `UIGestureRecognizer.mm` to return `NO` for `canPreventGestureRecognizer:` and `canBePreventedByGestureRecognizer:`, so all gestures fire simultaneously (unless of course someone were to subclass or add exclusion logic in their gesture handler selectors).

Thanks! Let me know if I need to clarify anything.

bbowman:
Seems like this would merit some sort of UT to validate. I know @davelamb mentioned not having UI automation yet but I don't think the test needs to be that extreme to prove out that the subclassing works and that the code you changed is operating correctly / not regressing things etc.

Frameworks/UIKit/UIView.mm
bbowman: sounds like things we should have tests for. What do we need to be able to do this? Can we avoid fullblown UI automation via mocking?

---
In reply to: [78622666](https://github.com/Microsoft/WinObjC/pull/1002#discussion_r78622666) [](ancestors = 78622666)

bbowman:
:clock1:

yiyang-msft:
Hi, @pittmab, 

  I merge your code into a new pull request #1909, would you mind taking a look?  The gesture code has gone through significant changes since this request.  It would be a loteasier for me to pull your code into latest code and working on top of it. 

Thanks
Yi

tadam-msft:
As seen in the thread the pull request was incorporated into another pull request and checked in.  This pull request is no longer needed.

PR #1004, opened 2016-09-13
Fix for loading images on non-retina devices

pittmab:
- no longer passing in nil, but adding a png image extension for the non retina cases

Fix: 1003

---

This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1004)

### DIFF
--- a/Frameworks/UIKit/UIImage.mm
+++ b/Frameworks/UIKit/UIImage.mm
@@ -382,7 +382,7 @@ static bool loadTIFF(UIImage* dest, void* bytes, int length) {
 
         if (EbrAccess(pathStr, 0) == -1) {
             id pathFind =
-                [bundle pathForResource:[NSString stringWithCString:newStr] ofType:nil inDirectory:nil forLocalization:@"English"];
+                [bundle pathForResource:[NSString stringWithCString:newStr] ofType:@"png" inDirectory:nil forLocalization:@"English"];
 
             if (pathFind != nil) {
                 path = (char*)[pathFind UTF8String];
@@ -403,7 +403,7 @@ static bool loadTIFF(UIImage* dest, void* bytes, int length) {
         pathStr = IwStrDup(path);
 
         if (EbrAccess(pathStr, 0) == -1) {
-            NSString* pathFind = [bundle pathForResource:pathAddr ofType:nil inDirectory:nil forLocalization:@"English"];
+            NSString* pathFind = [bundle pathForResource:pathAddr ofType:@"png" inDirectory:nil forLocalization:@"English"];
 
             if (pathFind != nil) {
                 path = [pathFind UTF8String];


DHowett-MSFT:
>         strcat_s(newStr, newStrSize, pathEnd);

I'm a little worried -- this should be stitching together `filename`, `@2x` and `.extension` already.

---
Refers to: Frameworks/UIKit/UIImage.mm:379 in 2518fe7. [](commit_id = 2518fe7ef69f111cda64e04a6bce904f15d2a255, deletion_comment = False)

DHowett-MSFT:

jaredhms:

jaredhms:
:shipit:

Frameworks/UIKit/UIImage.mm
```diff
@@ -382,7 +382,7 @@ - (instancetype)initWithContentsOfFile:(NSString*)pathAddr {
 
         if (EbrAccess(pathStr, 0) == -1) {
             id pathFind =
-                [bundle pathForResource:[NSString stringWithCString:newStr] ofType:nil inDirectory:nil forLocalization:@"English"];
+                [bundle pathForResource:[NSString stringWithCString:newStr] ofType:@"png" inDirectory:nil forLocalization:@"English"];
```
davelamb: This seems really strange. If I read this right, we test to see if 1) there's an extension separator, 2) we're a retina device, then tack on an `@2x`, and if that file isn't found we try again by tacking on a .png extension via pathForResource. If that isn't found, we try to access the file verbatim, and if that isn't found, we try again by tacking on a .png. 

I can think of some situations where it does what you don't want. For instance it appears it will never return an image when given a full _bundle_ path. (ie, `[UIImage initWithContentsOfFile:@"abc.jpg"]` where abc.jpg is only part of en.lproj. It seems like it'll look in the bundle first for abc@2x.jpg.png and then abc.jpg.png and of course never find it). And if the intent is to return a bundled png given a partial path, it also seems like it'll never return an `@2x` image. (ie, if this is to enable `[UIImage initWithContentsOfFile:@"abc"]` it will never return abc@2x.png, since there's no extension separator). 

Can you explain a little more the motivation behind this change? I'd be happy to give suggestions. :)

Dave

bbowman:
>         memcpy(newStr, path, pathEnd - path);

not your change but this string manipulation isn't exception safe, makes potentially unsafe assumption about the sizes of some datatypes, doesn't use the secure versions of certain common buffer overrun attack vectors, has a lot of duplicated code that could be factored out with RAII constructs, and doesn't follow WinObjC coding conventions. I'd be over the moon if we took the opportunity to improve the codebase where possible.

---
Refers to: Frameworks/UIKit/UIImage.mm:376 in 2518fe7. [](commit_id = 2518fe7ef69f111cda64e04a6bce904f15d2a255, deletion_comment = False)

Frameworks/UIKit/UIImage.mm
bbowman: Yeah anytime we are hardcoding a file extension, I'm a little dubious if that is the correct approach.

---
In reply to: [78836323](https://github.com/Microsoft/WinObjC/pull/1004#discussion_r78836323) [](ancestors = 78836323)

bbowman:
:clock1:

tadam-msft:
Can't repro issue, closing pull request.

Issue #904, opened 2016-08-23
Improve performance of [NSObject valueForKeyPath:]


rupertdaniel:
When using `[NSObject valueForKeyPath:]` there seems to be a heavy performance impact. This is very noticeable for example when using an NSSortDescriptor to sort a large array.

When profiling the app it seems a lot of time in `[NSObject valueForKeyPath:]` is spent processing the actual NSString of the key path.

Attached is a screen grab of the relevant profile section.

<img width="1108" alt="screenshot 2016-08-23 13 02 02" src="https://cloud.githubusercontent.com/assets/237035/17891170/134720b6-6932-11e6-89c7-401957031c06.png">

DHowett-MSFT:
This is fixed as of the merge of #1005.

PR #1005, opened 2016-09-13
Improve the performance of -[NSObject valueForKeyPath:].

DHowett-MSFT:
This replaces `_NSKVCSplitKeypath`, a key-splitting workhorse used in KVC and KVO, with a significantly more svelte version.

The old version:
* Used `dataUsingEncoding:` to take a full copy of the buffer at every step.
* Used ICU to find `.`

The new version:
* Trusts that CFString is capable of finding strings itself.

### Average per-invocation times.

```
  Old  |  New   | Workload
--------------------------
4208ns | 2762ns | "A.B.C.D" (2MM hits)
4985ns |  293ns | empty strings
4568ns | 1415ns | real-world profiling in application (56,000 hits @ string lengths 0-65)
```

Because both `valueForKey:` and `valueForKeyPath:` can be overridden by NSObject subclasses, full string copies of keypath components must be taken at every depth. This provides for a nearly-negligible performance impact (~799ns p/i in the average case.)

This pull request also fixes a regression in `valueForKey:` that caused the NSInvocation codepath to be triggered when the returned value would otherwise truly be `nil`.

#904.

### DIFF
--- a/Frameworks/Foundation/NSKeyValueCoding.mm
+++ b/Frameworks/Foundation/NSKeyValueCoding.mm
@@ -26,6 +26,7 @@
 #import "NSValueTransformers.h"
 #import "Starboard.h"
 #import "StubReturn.h"
+#import "type_encoding_cases.h"
 #import <_NSKeyValueCodingAggregateFunctions.h>
 
 #import <unicode/utf8.h>
@@ -162,6 +163,10 @@ id quickGet<Class>(id self, SEL getter) {
     return ((id (*)(id, SEL))objc_msgSend)(self, getter);
 }
 
+#define QUICK_GET_CASE(type, name, capitalizedName, encodingChar) \
+    case encodingChar: \
+        *ret = quickGet<type>(self, getter); \
+        return true;
 bool KVCGetViaAccessor(NSObject* self, SEL getter, id* ret) {
     if (!getter) {
         return false;
@@ -176,11 +181,9 @@ bool KVCGetViaAccessor(NSObject* self, SEL getter, id* ret) {
     }
 
     switch (valueType[0]) {
-#define APPLY_TYPE(type, name, capitalizedName, encodingChar) case encodingChar: *ret = quickGet<type>(self, getter); return true;
-        APPLY_TYPE(id, object, Object, '@')
-        APPLY_TYPE(Class, class, Class, '#')
-#include "type_encoding_cases.h"
-#undef APPLY_TYPE
+        OBJC_APPLY_NUMERIC_TYPE_ENCODINGS(QUICK_GET_CASE);
+        QUICK_GET_CASE(id, object, Object, '@');
+        QUICK_GET_CASE(Class, class, Class, '#');
     }
 
     NSInvocation* invocation = [NSInvocation invocationWithMethodSignature:sig];
@@ -395,6 +398,12 @@ bool quickSet<Class>(id self, SEL setter, id value, const char* valueType) {
     return true;
 }
 
+#define QUICK_SET_CASE(type, name, capitalizedName, encodingChar) \
+    case encodingChar: \
+        if (quickSet<type>(self, setter, value, valueType)) { \
+            return true; \
+        } \
+        break;
 bool KVCSetViaAccessor(NSObject* self, SEL setter, id value) {
     if (!setter) {
         return false;
@@ -407,16 +416,9 @@ bool KVCSetViaAccessor(NSObject* self, SEL setter, id value) {
         const char* valueType = [sig getArgumentTypeAtIndex:2];
 
         switch (valueType[0]) {
-#define APPLY_TYPE(type, name, capitalizedName, encodingChar) \
-            case encodingChar: \
-                if (quickSet<type>(self, setter, value, valueType)) { \
-                    return true; \
-                } \
-                break;
-            APPLY_TYPE(id, object, Object, '@')
-            APPLY_TYPE(Class, class, Class, '#')
-#include "type_encoding_cases.h"
-#undef APPLY_TYPE
+            OBJC_APPLY_NUMERIC_TYPE_ENCODINGS(QUICK_SET_CASE);
+            QUICK_SET_CASE(id, object, Object, '@');
+            QUICK_SET_CASE(Class, class, Class, '#');
         }
 
         uint8_t* data = static_cast<uint8_t*>(_alloca(objc_sizeof_type(valueType)));
--- a/Frameworks/Foundation/NSKeyValueCoding.mm
+++ b/Frameworks/Foundation/NSKeyValueCoding.mm
@@ -54,22 +54,12 @@ NSString* const NSUnionOfObjectsKeyValueOperator = @"NSUnionOfObjectsKeyValueOpe
 NSString* const NSUnionOfSetsKeyValueOperator = @"NSUnionOfSetsKeyValueOperator";
 
 NSString* _NSKVCSplitKeypath(NSString* keyPath, NSString* __autoreleasing* pRemainder) {
-    NSData* utf8String = [keyPath dataUsingEncoding:NSUTF8StringEncoding];
-    const char* buffer = static_cast<const char*>(utf8String.bytes);
-    NSInteger length = utf8String.length;
-    int i = 0;
-    UChar32 currentCharacter = 0;
-    while (i < length) {
-        U8_NEXT(buffer, i, length, currentCharacter);
-        if (currentCharacter == '.') {
-            *pRemainder = [[[NSString alloc] initWithBytesNoCopy:const_cast<char*>(buffer + i)
-                                                          length:length - i
-                                                        encoding:NSUTF8StringEncoding
-                                                    freeWhenDone:NO] autorelease];
-            return
-                [[[NSString alloc] initWithBytesNoCopy:const_cast<char*>(buffer) length:i - 1 encoding:NSUTF8StringEncoding freeWhenDone:NO]
-                    autorelease];
-        }
+    static woc::unique_cf<CFCharacterSetRef> dot{CFCharacterSetCreateWithCharactersInRange(nullptr, (CFRange){'.', 1})};
+    CFRange result;
+    CFIndex length = CFStringGetLength((CFStringRef)keyPath);
+    if (length > 0 && CFStringFindCharacterFromSet((CFStringRef)keyPath, dot.get(), (CFRange){0, length}, 0, &result)) {
+        *pRemainder = [(NSString*)CFStringCreateWithSubstring(nullptr, (CFStringRef)keyPath, (CFRange){result.location + 1, length - (result.location + 1)}) autorelease];
+        return [(NSString*)CFStringCreateWithSubstring(nullptr, (CFStringRef)keyPath, (CFRange){0, result.location}) autorelease];
     }
     *pRemainder = nil;
     return keyPath;
--- a/Frameworks/Foundation/NSKeyValueCoding.mm
+++ b/Frameworks/Foundation/NSKeyValueCoding.mm
@@ -175,13 +175,12 @@ bool KVCGetViaAccessor(NSObject* self, SEL getter, id* ret) {
         return false;
     }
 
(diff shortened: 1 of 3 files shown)


DHowett-MSFT:

DHowett-MSFT:

Frameworks/Foundation/NSKeyValueCoding.mm
```diff
@@ -54,22 +54,12 @@
...
-        if (currentCharacter == '.') {
-            *pRemainder = [[[NSString alloc] initWithBytesNoCopy:const_cast<char*>(buffer + i)
-                                                          length:length - i
-                                                        encoding:NSUTF8StringEncoding
-                                                    freeWhenDone:NO] autorelease];
-            return
-                [[[NSString alloc] initWithBytesNoCopy:const_cast<char*>(buffer) length:i - 1 encoding:NSUTF8StringEncoding freeWhenDone:NO]
-                    autorelease];
-        }
+    static woc::unique_cf<CFCharacterSetRef> dot{CFCharacterSetCreateWithCharactersInRange(nullptr, (CFRange){'.', 1})};
```
bbowman: >CFRange [](start = 101, length = 7)

CFMakeRange?

Frameworks/Foundation/NSKeyValueCoding.mm
```diff
@@ -54,22 +54,12 @@
...
+    CFRange result;
+    CFIndex length = CFStringGetLength((CFStringRef)keyPath);
+    if (length > 0 && CFStringFindCharacterFromSet((CFStringRef)keyPath, dot.get(), (CFRange){0, length}, 0, &result)) {
+        *pRemainder = [(NSString*)CFStringCreateWithSubstring(nullptr, (CFStringRef)keyPath, (CFRange){result.location + 1, length - (result.location + 1)}) autorelease];
```
bbowman: >CFStringRef [](start = 72, length = 11)

static cast

Frameworks/Foundation/NSKeyValueCoding.mm
```diff
@@ -54,22 +54,12 @@
...
+        return [(NSString*)CFStringCreateWithSubstring(nullptr, (CFStringRef)keyPath, (CFRange){0, result.location}) autorelease];
     }
     *pRemainder = nil;
```
bbowman: >*pRemainder [](start = 4, length = 11)

incredi-nit: I prefer for "multiple" return functions like this (using out ptr and a return val) to not use early return as it makes setting the out param harder to keep track of. A centralized place to set it and return may be nicer? I'm totally ok with you ignoring this though and making me suck it up.

Frameworks/Foundation/NSKeyValueCoding.mm
```diff
@@ -54,22 +54,12 @@
...
-                                                        encoding:NSUTF8StringEncoding
-                                                    freeWhenDone:NO] autorelease];
-            return
-                [[[NSString alloc] initWithBytesNoCopy:const_cast<char*>(buffer) length:i - 1 encoding:NSUTF8StringEncoding freeWhenDone:NO]
-                    autorelease];
-        }
+    static woc::unique_cf<CFCharacterSetRef> dot{CFCharacterSetCreateWithCharactersInRange(nullptr, (CFRange){'.', 1})};
+    CFRange result;
+    CFIndex length = CFStringGetLength((CFStringRef)keyPath);
+    if (length > 0 && CFStringFindCharacterFromSet((CFStringRef)keyPath, dot.get(), (CFRange){0, length}, 0, &result)) {
```
bbowman: >length > 0 [](start = 8, length = 10)

curious if this is actually needed or if CFStringFindCharacterFromSet will do the right thing. Either way is fine though.

Frameworks/Foundation/NSKeyValueCoding.mm
```diff
@@ -185,13 +175,12 @@ bool KVCGetViaAccessor(NSObject* self, SEL getter, id* ret) {
...
 
-    if (0
-#define APPLY_TYPE(type, name, capitalizedName, encodingChar) || (valueType[0] == encodingChar && (*ret = quickGet<type>(self, getter)))
-        APPLY_TYPE(id, object, Object, '@') APPLY_TYPE(Class, class, Class, '#')
+    switch (valueType[0]) {
+#define APPLY_TYPE(type, name, capitalizedName, encodingChar) case encodingChar: *ret = quickGet<type>(self, getter); return true;
+        APPLY_TYPE(id, object, Object, '@')
+        APPLY_TYPE(Class, class, Class, '#')
 #include "type_encoding_cases.h"
 #undef APPLY_TYPE
```
bbowman: well this is mildly terrifying. I'm pretty positive we shouldn't be knowingly including a .h smack dab in the middle of a function. That is asking for trouble and abusing the preprocessor in a way I'm not ok with. Having that header define macros etc to make this easier would be acceptable but not this.

Frameworks/Foundation/NSKeyValueCoding.mm
```diff
@@ -417,14 +406,17 @@ bool KVCSetViaAccessor(NSObject* self, SEL setter, id value) {
...
-            APPLY_TYPE(id, object, Object, '@') APPLY_TYPE(Class, class, Class, '#')
+            case encodingChar: \
+                if (quickSet<type>(self, setter, value, valueType)) { \
+                    return true; \
+                } \
+                break;
+            APPLY_TYPE(id, object, Object, '@')
+            APPLY_TYPE(Class, class, Class, '#')
 #include "type_encoding_cases.h"
 #undef APPLY_TYPE
```
bbowman: ditto.

bbowman:
:clock1:

Frameworks/Foundation/NSKeyValueCoding.mm
DHowett-MSFT: I believe this is a perfectly acceptable use of a literal

DHowett-MSFT:
There are tabs in one of the files.
I hate Visual Studio at times.

DHowett-MSFT:
@bbowman your concerns regarding `type_encoding_cases` have been addressed.

Frameworks/Foundation/NSKeyValueCoding.mm
```diff
@@ -54,22 +55,12 @@
...
-            return
-                [[[NSString alloc] initWithBytesNoCopy:const_cast<char*>(buffer) length:i - 1 encoding:NSUTF8StringEncoding freeWhenDone:NO]
-                    autorelease];
-        }
+    static woc::unique_cf<CFCharacterSetRef> dot{CFCharacterSetCreateWithCharactersInRange(nullptr, (CFRange){'.', 1})};
+    CFRange result;
+    CFIndex length = CFStringGetLength((CFStringRef)keyPath);
+    if (length > 0 && CFStringFindCharacterFromSet((CFStringRef)keyPath, dot.get(), (CFRange){0, length}, 0, &result)) {
+        *pRemainder = [(NSString*)CFStringCreateWithSubstring(nullptr, (CFStringRef)keyPath, (CFRange){result.location + 1, length - (result.location + 1)}) autorelease];
+        return [(NSString*)CFStringCreateWithSubstring(nullptr, (CFStringRef)keyPath, (CFRange){0, result.location}) autorelease];
```
msft-Jeyaram: >[(NSString*) [](start = 15, length = 12)

I assume C-style cast vs static_cast to reduce runtime operation ?

msft-Jeyaram:
:shipit:

Frameworks/Foundation/NSKeyValueCoding.mm
aballway: You should be able to just do CFRange{...} w/o cast parens

aballway:
:shipit:

Frameworks/Foundation/NSKeyValueCoding.mm
```diff
@@ -54,22 +55,12 @@
...
-                                                        encoding:NSUTF8StringEncoding
-                                                    freeWhenDone:NO] autorelease];
-            return
-                [[[NSString alloc] initWithBytesNoCopy:const_cast<char*>(buffer) length:i - 1 encoding:NSUTF8StringEncoding freeWhenDone:NO]
-                    autorelease];
-        }
+    static woc::unique_cf<CFCharacterSetRef> dot{CFCharacterSetCreateWithCharactersInRange(nullptr, (CFRange){'.', 1})};
+    CFRange result;
+    CFIndex length = CFStringGetLength((CFStringRef)keyPath);
+    if (length > 0 && CFStringFindCharacterFromSet((CFStringRef)keyPath, dot.get(), (CFRange){0, length}, 0, &result)) {
```
mnithish: >CFStringRef [](start = 52, length = 11)

nit: C++ style casts as per our coding style document

All through this CR

mnithish:
:shipit:

ms-jihua:
:shipit:

MSFTFox:
> //

nit: update copyright.

---
Refers to: Frameworks/Foundation/NSKeyValueCoding.mm:4 in f8418ea. [](commit_id = f8418ea3c0f1e7e1b5ccf625acd934351fa658e4, deletion_comment = False)

MSFTFox:
> //

nit: update copyright.

---
Refers to: Frameworks/Foundation/NSKVOSwizzling.mm:4 in f8418ea. [](commit_id = f8418ea3c0f1e7e1b5ccf625acd934351fa658e4, deletion_comment = False)

Frameworks/Foundation/NSKeyValueCoding.mm
```diff
@@ -54,22 +55,12 @@
...
+        *pRemainder = [(NSString*)CFStringCreateWithSubstring(nullptr, (CFStringRef)keyPath, (CFRange){result.location + 1, length - (result.location + 1)}) autorelease];
```
MSFTFox: >{result.location + 1, length - (result.location + 1)} [](start = 102, length = 53)

A comment about what substring is being created here and what it should look like could prove useful. Is there a situation where this substring is incorrect as in some malformed path?

MSFTFox: >pRemainder [](start = 9, length = 10)

This is an internal function and I expect this case won't be hit but should you check if pRemainder isn't a null pointer first?

MSFTFox:
:shipit:

Frameworks/Foundation/NSKeyValueCoding.mm
DHowett-MSFT: `pRemainder != nullptr` is an invariant.

Frameworks/Foundation/NSKeyValueCoding.mm
DHowett-MSFT: I'd rather be safe.

---
In reply to: [79024603](https://github.com/Microsoft/WinObjC/pull/1005#discussion_r79024603) [](ancestors = 79024603)

Frameworks/Foundation/NSKeyValueCoding.mm
DHowett-MSFT: (I'm choosing to ignore this one :))

---
In reply to: [79024385](https://github.com/Microsoft/WinObjC/pull/1005#discussion_r79024385) [](ancestors = 79024385)

Frameworks/Foundation/NSKeyValueCoding.mm
DHowett-MSFT: I am choosing to reduce the visual noise.

---
In reply to: [79666944](https://github.com/Microsoft/WinObjC/pull/1005#discussion_r79666944) [](ancestors = 79666944)

Frameworks/Foundation/NSKeyValueCoding.mm
DHowett-MSFT: At the very worst case, this will produce `a` `.b` from `a..b`; The rest will work properly (`.` in any position)

---
In reply to: [79673352](https://github.com/Microsoft/WinObjC/pull/1005#discussion_r79673352) [](ancestors = 79673352)

Issue #918, opened 2016-08-26
Implement critical CTFontDescriptor methods


rajsesh-msft:
Specifically: 
- [x] CTFontDescriptorCreateWithNameAndSize
- [x] CTFontDescriptorCopyLocalizedAttribute
- [x] Interoperable	CTFontDescriptorCreateWithAttributes
- [x] Caveat	CTFontCreateCopyWithSymbolicTraits
- [x] Caveat	CTFontGetAdvancesForGlyphs

PR #1008, opened 2016-09-14
Bridge CTFontDescriptor and UIFontDescriptor

### LINKED ISSUE #730: Bridge CTFontDescriptor and UIFontDescriptor

ms-jihua:
- Initial pass at these classes touching just the basic functionality
     - Updated constants in both classes to be closer to those on the reference platform
 - Separated out a portion of CFInternal.h and CFFoundationInternal.h
   into a new header, CFBridgeUtilities.h, for use in non-CF Core* frameworks
 - Unit tests
 - Misc fixes

Fixes #730


build/CoreFoundation/dll/CoreFoundation.def
```diff
@@ -29,6 +29,8 @@ LIBRARY CoreFoundation
         _CFAppendPathExtension2
         _CFApplicationPreferencesCopyRepresentation
         _CFRuntimeBridgeTypeToClass
+		__CFISAForTypeID
```
aballway: >		 [](start = 0, length = 2)

nit: tabs

ms-jihua: oops, saw this earlier, thought I had fixed it

---
In reply to: [78851772](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r78851772) [](ancestors = 78851772)

Frameworks/UIKit/UIFontDescriptor.mm
```diff
@@ -29,233 +36,176 @@
...
-}
-
-/**
  @Status Interoperable
 */
-- (UIFontDescriptor*)fontDescriptorWithSymbolicTraits:(UIFontDescriptorSymbolicTraits)symbolicTraits {
-    UIFontDescriptor* result = [UIFontDescriptor fontDescriptorWithDescriptor:self size:self->_size];
-    result->_traits = symbolicTraits;
-    return result;
++ (UIFontDescriptor*)fontDescriptorWithFontAttributes:(NSDictionary*)attributes {
```
ms-jihua: I ended up reordering the functions in this file to match that of the documentation - the old ordering sort of made sense in the old impl, but is very disorganized in the new one.

aballway:
:shipit:

Frameworks/CoreFoundation/Base.subproj/CFInternal.h
```diff
@@ -363,10 +359,9 @@ CF_PRIVATE Boolean __CFProcessIsRestricted();
 
 // WINOBJC: The reference platform uses a special compiler option not available to us to 
 // get __CFConstantStringClassReferencePtr filled out with the correct "Class". Since this isn't
```
DHowett-MSFT: Can you propagate this comment to CFBridgeUtilities.h? It might be useful to pretend CFBridgeUtilities is an upstream header with our own modifications (for change-tracking purposes)

Frameworks/CoreFoundation/Base.subproj/CFRuntime.c
```diff
@@ -275,6 +275,10 @@ void _CFRuntimeBridgeTypeToClass(CFTypeID cf_typeID, const void *cls_ref) {
     __CFUnlock(&__CFBigRuntimeFunnel);
 }
 
+uintptr_t __CFISAForTypeID(CFTypeID typeID) {
```
DHowett-MSFT: >__CFISAForTypeID [](start = 10, length = 16)

I'm not totally happy that this was made non-inline. Is there a way you can preserve the inlined version for CF internal consumers and keep an out-of-line version for external ones?

Frameworks/UIKit/UIFontDescriptor.mm
```diff
@@ -29,233 +36,176 @@
...
 }
 
 /**
  @Status Interoperable
 */
-- (void)dealloc {
-    _name = nil;
-    [super dealloc];
+- (UIFontDescriptor*)fontDescriptorByAddingAttributes:(NSDictionary*)attributes {
+    NSMutableDictionary* newAttributes = [[self fontAttributes] mutableCopy];
```
DHowett-MSFT: >mutableCopy [](start = 64, length = 11)

This returns a +1

ms-jihua: oops

---
In reply to: [78862671](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r78862671) [](ancestors = 78862671)

Frameworks/CoreFoundation/Base.subproj/CFInternal.h
ms-jihua: Most of this comment is in CFBridgeUtilities? Confused which part you're asking for

---
In reply to: [78862194](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r78862194) [](ancestors = 78862194)

DHowett-MSFT: oh, i see. sorry.
I think we should use "WINOBJC" just as liberally in CFBridgeUtilities -- to make sure we know what we "changed" versus what we "moved" :)
We could probably kill this whole comment block in here, and move the commented-out CFEXPORTs to CFBridgeUtilities

---
In reply to: [78864285](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r78864285) [](ancestors = 78864285,78862194)

Frameworks/CoreFoundation/Base.subproj/CFRuntime.c
ms-jihua: seems like CF_INLINE still works just fine!

---
In reply to: [78862337](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r78862337) [](ancestors = 78862337)

Frameworks/CoreFoundation/Base.subproj/CFInternal.h
ms-jihua: got it

---
In reply to: [78864587](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r78864587) [](ancestors = 78864587,78864285,78862194)

Frameworks/CoreFoundation/Base.subproj/CFRuntime.c
DHowett-MSFT: Like: in CT, CF_INLINE works fine? Even though this refers to __CFRuntimeObjCClassTable?

---
In reply to: [78866010](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r78866010) [](ancestors = 78866010,78862337)

ms-jihua: Yes - __CFRuntimeObjCClassTable is usable from other frameworks. If it weren't, we wouldn't be able to get it from Foundation, and the bridging tests we have so far for UICTFont shouldn't have worked.

---
In reply to: [79022634](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r79022634) [](ancestors = 79022634,78866010,78862337)

Frameworks/CoreFoundation/Base.subproj/CFBridgeUtilities.h
```diff
@@ -0,0 +1,68 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+// WINOBJC: The following functions/macros are separated into this header, so that their code can be more broadly reused among the other
```
bbowman: >separated [](start = 47, length = 9)

might mention the original location here.

Frameworks/CoreFoundation/Base.subproj/CFBridgeUtilities.h
```diff
@@ -0,0 +1,68 @@
...
+    } while (0)
+
+#define CF_OBJC_CALLV(obj, ...) [(obj)__VA_ARGS__]
+
+// easier to read without clang-format here
+// clang-format off
+#define CF_IS_OBJC(typeID, obj) ( \
+    (!obj) || \
+    ((((CFRuntimeBase*)(obj))->_cfisa != 0) && \
+    (((CFRuntimeBase*)(obj))->_cfisa != (uintptr_t)(&_OBJC_CLASS__NSCFString)) && \
```
bbowman: >_OBJC_CLASS__NSCFString [](start = 53, length = 23)

Didn't we also need to do this with Number or did that get removed with a different change?

ms-jihua: I vaguely remember adding something like that but it doesn't seem to be there anymore, so I'm guessing the latter? Lemme dig into the git history a bit.

---
In reply to: [79028498](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r79028498) [](ancestors = 79028498)

Frameworks/CoreFoundation/Base.subproj/CFBridgeUtilities.h
```diff
@@ -0,0 +1,68 @@
...
+    (__CFISAForTypeID(typeID) != ((CFRuntimeBase*)(obj))->_cfisa) && \
+    (![(id)(((CFRuntimeBase*)(obj)))->_cfisa isSubclassOfClass:(Class)__CFISAForTypeID(typeID)])))
```
bbowman: To help readability here we might add a comment after each case explaining (I should have done this initially):

1. // Treat nil as ObjC
2. // Any isa means ObjC but make sure it isn't a bridge
3. // Special handling for String due to static ordering and registration (see above)
4. // Make sure its not the concrete impl that is bridged
5. // Make sure its not a subclass (should only be KVO) of the concrete bridged impl

Frameworks/CoreFoundation/Base.subproj/CFBridgeUtilities.h
```diff
@@ -0,0 +1,68 @@
...
+// clang-format on
+
+// For debugging C-style classes
+#if defined(DEBUG)
+extern void __CFGenericValidateType_(CFTypeRef cf, CFTypeID type, const char* func);
+#define __CFGenericValidateType(cf, type) __CFGenericValidateType_(cf, type, __PRETTY_FUNCTION__)
+#else
+#define __CFGenericValidateType(cf, type) ((void)0)
```
bbowman: >__CFGenericValidateType [](start = 8, length = 23)

is this actually necessary for CT / CG?

Frameworks/CoreText/CTFont.mm
```diff
@@ -249,25 +249,16 @@ CTFontRef CTFontCreateCopyWithAttributes(CTFontRef font, CGFloat size, const CGA
...
-    UIFontDescriptor* newFontDescriptor =
-        [[(UIFont*)font fontDescriptor] fontDescriptorWithSymbolicTraits:(existingTraits | symTraitValue) & symTraitMask];
-
-    if (size == 0.0f) {
-        size = ((UIFont*)font).pointSize;
-    }
-    UIFont* ret = [[_LazyUIFont fontWithDescriptor:newFontDescriptor size:size] retain];
-
-    return (CTFontRef)ret;
+    UNIMPLEMENTED();
```
mnithish: >UNIMPLEMENTED [](start = 4, length = 13)

Given we have CT/UIFont bridged we can implement parts of this API now? Are you tracking this with a different task?

Frameworks/CoreFoundation/Base.subproj/CFBridgeUtilities.h
ms-jihua: Hmm, for some reason I only added Number to __CF_IsBridgedObject() and not here... how should I check if Number is currently doing the right thing? just static NSNumber* foo = @2413, break on it, see if it's okay?

---
In reply to: [79029288](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r79029288) [](ancestors = 79029288,79028498)

Frameworks/CoreText/CTFont.mm
ms-jihua: Yeah, #918 is supposed to be the full, fleshed out impl. This is the initial small chunk.

---
In reply to: [79029817](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r79029817) [](ancestors = 79029817)

Frameworks/CoreFoundation/Base.subproj/CFBridgeUtilities.h
ms-jihua: No, but CF was peppering it in so many places that it seemed like a good thing to match? Do we wanna just get rid of it in our code and have smaller function bodies?

---
In reply to: [79029560](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r79029560) [](ancestors = 79029560)

Frameworks/CoreFoundation/Base.subproj/CFRuntime.c
bbowman: I'm not sure how I feel about that. Seems like we shouldn't be exporting that guy if possible?

---
In reply to: [79025442](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r79025442) [](ancestors = 79025442,79022634,78866010,78862337)

tests/unittests/UIKit/UIFontDescriptorTests.mm
```diff
@@ -0,0 +1,77 @@
...
+    ASSERT_OBJCEQ(name, [fd objectForKey:UIFontDescriptorNameAttribute]);
+    ASSERT_EQ(size, fd.pointSize);
+}
+
+TEST(UIFontDescriptor, BadPointSize) {
+    UIFontDescriptor* fd = [UIFontDescriptor fontDescriptorWithFontAttributes:@{}];
+    ASSERT_OBJCNE(nil, fd);
+    ASSERT_EQ(0, [fd pointSize]);
+
+    fd = fd = [UIFontDescriptor fontDescriptorWithFontAttributes:@{ UIFontDescriptorSizeAttribute : @"shouldn't be a string" }];
```
mnithish: >fd = fd  [](start = 4, length = 8)

nit

ms-jihua: woops, how did that happen.

---
In reply to: [79031872](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r79031872) [](ancestors = 79031872)

Frameworks/CoreText/CTFontDescriptor.mm
```diff
@@ -15,59 +15,124 @@
...
+const CFStringRef kCTFontOrientationAttribute = static_cast<CFStringRef>(@"NSFontOrientationAttribute");
+const CFStringRef kCTFontFormatAttribute = static_cast<CFStringRef>(@"NSFontFormatAttribute");
+const CFStringRef kCTFontRegistrationScopeAttribute = static_cast<CFStringRef>(@"NSFontRegistrationScopeAttribute");
+const CFStringRef kCTFontPriorityAttribute = static_cast<CFStringRef>(@"NSFontPriorityAttribute");
+const CFStringRef kCTFontEnabledAttribute = static_cast<CFStringRef>(@"NSFontEnabledAttribute");
+
+const CFStringRef kCTFontSymbolicTrait = static_cast<CFStringRef>(@"NSCTFontSymbolicTrait");
+const CFStringRef kCTFontWeightTrait = static_cast<CFStringRef>(@"NSCTFontWeightTrait");
+const CFStringRef kCTFontWidthTrait = static_cast<CFStringRef>(@"NSCTFontWidthTrait");
+const CFStringRef kCTFontSlantTrait = static_cast<CFStringRef>(@"NSCTFontSlantTrait");
```
bbowman: Maybe would be a little nicer if we used CFSTR instead? Our NSConstantString impl isn't the *nicest*

ms-jihua: I remember there being some weird reason why we couldn't do this (ran into it in other files too) but I don't remember what it was...

---
In reply to: [79032599](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r79032599) [](ancestors = 79032599)

mnithish:
:shipit:

Frameworks/CoreText/CTFontDescriptor.mm
```diff
@@ -15,59 +15,124 @@
...
 */
 CTFontDescriptorRef CTFontDescriptorCreateCopyWithAttributes(CTFontDescriptorRef original, CFDictionaryRef attributes) {
-    UNIMPLEMENTED();
-    return StubReturn();
+    // New attributes override old ones
+    CFMutableDictionaryRef newAttributes = CFDictionaryCreateMutableCopy(kCFAllocatorDefault, 0, original->_dictionary);
+
+    CFDictionaryApplyFunction(attributes,
+                              [](const void* key, const void* value, void* context) {
+                                  CFDictionarySetValue(reinterpret_cast<CFMutableDictionaryRef>(context), key, value);
```
bbowman: >context [](start = 96, length = 7)

would prefer it if we didn't smuggle things into context if possible I think? we can pass newAttributes as a lambda capture to get more correct typing instead of falling down to void*

Frameworks/CoreFoundation/Base.subproj/CFRuntime.c
DHowett-MSFT: I didn't know we accessed it from Foundation; I thought we went through _CFRuntimeBridgeTypeToClass! Cool.

---
In reply to: [79025442](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r79025442) [](ancestors = 79025442,79022634,78866010,78862337)

Frameworks/CoreText/CTFontDescriptor.mm
DHowett-MSFT: Smuggling things is _literally the point of a void* context_

---
In reply to: [79033322](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r79033322) [](ancestors = 79033322)

DHowett-MSFT: Also, I do not believe you can use a capturing lambda as a bare function pointer. `CFDictionaryApplyFunction` expects a function pointer, and only capture-free lambdas can be used as function pointers.

---
In reply to: [79033675](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r79033675) [](ancestors = 79033675,79033322)

Frameworks/CoreText/CTFontDescriptor.mm
```diff
@@ -15,59 +15,124 @@
...
 /**
- @Status Stub
+ @Status Interoperable
  @Notes
 */
 CTFontDescriptorRef CTFontDescriptorCreateCopyWithAttributes(CTFontDescriptorRef original, CFDictionaryRef attributes) {
-    UNIMPLEMENTED();
-    return StubReturn();
+    // New attributes override old ones
+    CFMutableDictionaryRef newAttributes = CFDictionaryCreateMutableCopy(kCFAllocatorDefault, 0, original->_dictionary);
```
DHowett-MSFT: >CFDictionaryCreateMutableCopy [](start = 43, length = 29)

This looks like it'll leak.

Frameworks/CoreFoundation/Base.subproj/CFRuntime.c
DHowett-MSFT: Wait wait wait..
CF_INLINE in a .c file doesn't make sense.
The intent of having it in the header is for every internal CF consumer to inline it -- not just the ones in this file. The compiler can't inline from one .c file to another (or, LTCG can, but it is difficult).

COMDAT folding might help this, but we still have an unnecessary branch to the function impl.

---
In reply to: [79033450](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r79033450) [](ancestors = 79033450,79025442,79022634,78866010,78862337)

DHowett-MSFT: My proposal is to have it in the header for corefoundation-internal consumers, and have it forward-declared and exported for corefoundation-external consumers.

---
In reply to: [79034377](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r79034377) [](ancestors = 79034377,79033450,79025442,79022634,78866010,78862337)

Frameworks/CoreFoundation/Base.subproj/CFBridgeUtilities.h
bbowman: its probably ok as is since it no-ops in release

---
In reply to: [79030373](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r79030373) [](ancestors = 79030373,79029560)

Frameworks/CoreFoundation/Base.subproj/CFBridgeUtilities.h
bbowman: yeah you'd need to do both a static number and a static init time call to CF_IS_OBJC (preferably in the same file and higher than the definition of the number so that the IS_OBJC is executed first) and see what it says. true would be wrong, false would be correct but I'm guessing it will be wrong as currently written we just don't have a call pattern in the real world to hit it?

---
In reply to: [79029831](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r79029831) [](ancestors = 79029831,79029288,79028498)

Frameworks/CoreText/CTFontDescriptor.mm
ms-jihua: What Dustin said.

---
In reply to: [79033804](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r79033804) [](ancestors = 79033804,79033675,79033322)

Frameworks/CoreText/CTFontDescriptor.mm
ms-jihua: I keep missing these somehow...

---
In reply to: [79033908](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r79033908) [](ancestors = 79033908)

Frameworks/CoreText/CTFontDescriptor.mm
```diff
@@ -111,21 +176,28 @@ CTFontDescriptorRef CTFontDescriptorCreateMatchingFontDescriptor(CTFontDescripto
...
 */
 CFTypeRef CTFontDescriptorCopyAttribute(CTFontDescriptorRef descriptor, CFStringRef attribute) {
-    UNIMPLEMENTED();
-    return StubReturn();
+    CF_OBJC_FUNCDISPATCHV(CTFontDescriptorGetTypeID(), CFTypeRef, (UIFontDescriptor*)descriptor, objectForKey
+                          : static_cast<NSString*>(attribute));
+    __CFGenericValidateType(descriptor, CTFontDescriptorGetTypeID());
+
+    // Use CFRetain instead of a true copy here
+    //  - A generic 'CFCopy' is not available, as not all types are copyable
```
bbowman: >CFCopy [](start = 21, length = 6)

not sure if thats true. I seem to remember a copy being there but in a confusing way / name. But if this matches the ref plat. #yolo?

Frameworks/UIKit/UIFontDescriptor.mm
```diff
@@ -29,233 +36,176 @@
...
-}
+const CGFloat UIFontWeightUltraLight = -0.80f;
+const CGFloat UIFontWeightThin = -0.60f;
+const CGFloat UIFontWeightLight = -0.40f;
+const CGFloat UIFontWeightRegular = 0.00f;
+const CGFloat UIFontWeightMedium = 0.23f;
+const CGFloat UIFontWeightSemibold = 0.30f;
+const CGFloat UIFontWeightBold = 0.40f;
+const CGFloat UIFontWeightHeavy = 0.56f;
+const CGFloat UIFontWeightBlack = 0.62f;
```
bbowman: add a comment on these magic numbers about where these came from / how you determined?

Frameworks/CoreText/CTFontDescriptor.mm
ms-jihua: I mean, if such a thing existed, what would be the point of functions like CFStringCreateCopy then...

Yes yolo

---
In reply to: [79035712](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r79035712) [](ancestors = 79035712)

Frameworks/UIKit/UIFontDescriptor.mm
ms-jihua: just copied the ref plat, will comment

---
In reply to: [79036080](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r79036080) [](ancestors = 79036080)

Frameworks/CoreText/CTFontDescriptor.mm
bbowman: oh didn't know the impl was a func pointer that makes sense.

---
In reply to: [79035467](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r79035467) [](ancestors = 79035467,79033804,79033675,79033322)

Frameworks/CoreText/CTFontDescriptor.mm
DHowett-MSFT: __CFRuntimeClass has a "copy" field -- it looks like there's supposed to be a generic copy?

---
In reply to: [79036813](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r79036813) [](ancestors = 79036813,79035712)

bbowman:
:clock1:

Frameworks/CoreText/CTFontDescriptor.mm
ms-jihua: Hm, seems to be fine.

---
In reply to: [79032846](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r79032846) [](ancestors = 79032846,79032599)

Frameworks/CoreFoundation/Base.subproj/CFBridgeUtilities.h
ms-jihua: yup, currently wrong

---
In reply to: [79035359](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r79035359) [](ancestors = 79035359,79029831,79029288,79028498)

ms-jihua: nvm, I wrote it wrong - it seems to be fine and the problem seems unique to string

---
In reply to: [79054184](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r79054184) [](ancestors = 79054184,79035359,79029831,79029288,79028498)

Frameworks/CoreText/CTFontDescriptor.mm
DHowett-MSFT: It looks like `__CFRuntimeClass` has a `copy` field, and that generic copies should be possible. I'm not sure it's surfaced anywhere, though?

ms-jihua: So it does... seems difficult to find though, so I'm not gonna invest the effort right now x___x

---
In reply to: [79066713](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r79066713) [](ancestors = 79066713)

Frameworks/CoreFoundation/Base.subproj/CFRuntime.c
ms-jihua: Discussed offline, decided that this optimization wasn't worth the effort/complexity.

---
In reply to: [79034493](https://github.com/Microsoft/WinObjC/pull/1008#discussion_r79034493) [](ancestors = 79034493,79034377,79033450,79025442,79022634,78866010,78862337)

MSFTFox:
:shipit:

PR #1011, opened 2016-09-20
Do not use a recursive MSBuild rule to determine Objective-C residency.

### LINKED ISSUE #973: Target "_ComputeStarboardReferencedProjectRequiresObjectiveC missing when building a project with C++ static lib
Projects don't build anymore when adding a C++ static lib following the [wiki page instructions](https://github.com/Microsoft/WinObjC/wiki/Creating-and-Linking-Against-C---Static-Libraries), with **error MSB4057 The target "_ComputeStarboardReferencedProjectRequiresObjectiveC" does not exist in the project**. This seems to be a regression of the August 23rd release, since it works with the July 28th release (the September 6 release still gives the error).

To repro, just add a C++ static lib to the HelloUI sample following the wiki instructions and try to build the project.

DHowett-MSFT:
The recursive rule was causing us some trouble when projects did not include `sbclang.targets`.

This works just as well because any static library with an object containing this directive will spread `___pin_objc_init` in a fairly insidious manner.

Fixes #973.

Verified by:
* x86 debug app launch
* x86 debug UTs
* x86 release app launch
* x86 release UTs
* ARM release app launch
* ARM release UTs

### DIFF
--- a/build/Package/Package.nativeproj
+++ b/build/Package/Package.nativeproj
@@ -47,6 +47,7 @@
       <MSVCFiles Include="..\..\msvc\*.dll" />
       <MSVCFiles Include="..\..\msvc\*.cpp" />
       <MSVCObjCRTFiles Include="..\..\msvc\objcrt\*.cpp" />
+      <MSVCObjCRTFiles Include="..\..\msvc\objcrt\*.h" />
       <SBIncludes Include="..\..\include\**\*" />
       <SBIncludes Include="..\..\deps\prebuilt\include\zlib.h" />
       <SBIncludes Include="..\..\deps\prebuilt\include\zconf.h" />
--- a/msvc/objcrt/objc_2stage_init.cpp
+++ b/msvc/objcrt/objc_2stage_init.cpp
@@ -14,6 +14,8 @@
 //
 //******************************************************************************
 
+#pragma comment(lib, "libobjc2")
+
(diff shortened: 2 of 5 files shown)


bbowman:
:shipit:

DHowett-MSFT:
This build failure may be transient; I'm going to submit another one:
```
2016-09-20T01:06:05.2685043Z MSBUILD : error MSB4017: The build stopped unexpectedly because
 of an unexpected logger failure.
2016-09-20T01:06:05.2685043Z Microsoft.Build.Exceptions.InternalLoggerException: The build
 stopped unexpectedly because of an unexpected logger failure.
 ---> System.ArgumentException: Illegal characters in path.
```

pradipd:
That means there is an issue with parallel builds.

---
In reply to: [248383261](https://github.com/Microsoft/WinObjC/pull/1011#issuecomment-248383261) [](ancestors = 248383261)

pradipd:
Err. Maybe not.

---
In reply to: [248384868](https://github.com/Microsoft/WinObjC/pull/1011#issuecomment-248384868) [](ancestors = 248384868,248383261)

pradipd:
:shipit:

DHowett-MSFT:
Tested this with a non-sample application with subproject references, and it looks good.

PR #1025, opened 2016-09-22
Fix the audiotoolbox sample page layout for phone.

rajsesh-msft:
The current layout causes horizontal scroll, which is not ideal.


aballway:
:shipit:

PR #1026, opened 2016-09-22
Remove spurious define of TARGET_IPHONE_SIMULATOR.  

rajsesh-msft:
This was originally added according to comments for cocos2d compat.  Looking at the sources, this isnt required anymore.  Worse, it interferes with suspend/resume handling on apps where due to a workaround in the TARGET_IPHONE_SIMULATOR codepath, we are trying to resume a track while suspending, resulting in the track not playing on resume.

We no longer pretend to be TARGET_IPHONE, so there is no need to pretend to be the simulator either.

### DIFF
--- a/include/StarboardAdditionalDefines.h
+++ b/include/StarboardAdditionalDefines.h
@@ -17,10 +17,6 @@
 #ifndef __STARBOARD_ADDITIONAL_DEFINES
 #define __STARBOARD_ADDITIONAL_DEFINES
 
-#ifndef TARGET_IPHONE_SIMULATOR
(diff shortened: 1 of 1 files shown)


rajsesh-msft:

rajsesh-msft:

rajsesh-msft:
I can provide the details of the test app offline.

DHowett-MSFT:
:shipit:

bbowman:
:shipit:

MSFTFox:
:shipit:

msft-Jeyaram:
:shipit:

PR #1027, opened 2016-09-22
implement CTLineGetTypographicBounds and CTFramesetterSuggestFrameSizeWithConstraints

### LINKED ISSUE #1019: Support CTLineGetTypographicBounds
With move to DWrite this method is now stubbed.

Also enable CTTypeSetter - LinesShouldDefaultToZeroOffset test code once this API is supported

aballway:
Fixes #1019

### DIFF
--- a/Frameworks/CoreText/CTFramesetter.mm
+++ b/Frameworks/CoreText/CTFramesetter.mm
@@ -30,48 +30,19 @@ using namespace std;
 }
 @end
 
-static _CTFrame* _createFrame(_CTFramesetter* framesetter, CGRect frameSize, CGSize* sizeOut, bool createFrame) {
-    _CTFrame* ret = nil;
-
-    if (createFrame) {
-        ret = [_CTFrame new];
-        ret->_framesetter = framesetter;
-        ret->_frameRect = frameSize;
+static _CTFrame* _createFrame(_CTFramesetter* framesetter, CGRect frameSize, CFRange range) {
+    if (!framesetter) {
+        return nil;
     }
 
     // Call _DWriteWrapper to get _CTLine object list that makes up this frame
     _CTTypesetter* typesetter = static_cast<_CTTypesetter*>(framesetter->_typesetter);
-    CFRange range = CFRangeMake(0, typesetter->_characters.size());
-    NSArray<_CTLine*>* lines = _DWriteGetLines(typesetter, range, frameSize);
-    if (createFrame) {
-        ret->_lines = lines;
-        for (_CTLine* line in static_cast<id<NSFastEnumeration>>(ret->_lines)) {
-            ret->_lineOrigins.emplace_back(line->_lineOrigin);
-        }
+    if (range.length == 0) {
+        range.length = typesetter->_characters.size();
     }
-
-    if (sizeOut && range.length != 0) {
-        // TODO::
-        // Is there a better way to do this - Investigate.
-        // For now the below logic is used to calculate the frame size required to fit the text in -
-        //     width = frame width as it is the constraint passed to DWrite to calculate renderable area width
-        //     height = min(_lineOrigin.y of the last line in the frame + the _lineOrigin.y of the first line (so we atleast get the
-        //              empty space that the first line got), frameSize.size.height)
-
-        // Calculate width
-        sizeOut->width = frameSize.size.width;
-
-        // Calculate height
-        _CTLine* firstLine = static_cast<_CTLine*>([lines objectAtIndex:0]);
-        _CTLine* lastLine = static_cast<_CTLine*>([lines objectAtIndex:[lines count] - 1]);
-        // TODO::
-        // Modify this code once we move away from Cairo Freetype for rasterizing text.
-        CGFloat originYFirstLine = frameSize.size.height - firstLine->_lineOrigin.y - frameSize.origin.y;
-        CGFloat originYLastLine = frameSize.size.height - lastLine->_lineOrigin.y - frameSize.origin.y;
-        sizeOut->height = min(originYLastLine + originYFirstLine, frameSize.size.height);
-    }
-
-    return ret;
+    _CTFrame* ret = _DWriteGetFrame(typesetter, range, frameSize);
+    ret->_framesetter = framesetter;
+    return [ret retain];
 }
 
 /**
@@ -90,8 +61,7 @@ CTFrameRef CTFramesetterCreateFrame(CTFramesetterRef framesetter, CFRange string
     CGRect frameSize;
     _CGPathGetBoundingBoxInternal(path, &frameSize);
 
-    CGSize sizeOut;
-    _CTFrame* ret = _createFrame((_CTFramesetter*)framesetter, frameSize, &sizeOut, true);
+    _CTFrame* ret = _createFrame(static_cast<_CTFramesetter*>(framesetter), frameSize, stringRange);
 
     return static_cast<CTFrameRef>(ret);
 }
@@ -105,19 +75,22 @@ CTTypesetterRef CTFramesetterGetTypesetter(CTFramesetterRef framesetter) {
 }
 
 /**
- @Status Caveat
- @Notes Always sets out parameter fitRange to stringRange value
+ @Status Interoperable
+ @Notes
 */
 CGSize CTFramesetterSuggestFrameSizeWithConstraints(
     CTFramesetterRef framesetter, CFRange stringRange, CFDictionaryRef frameAttributes, CGSize constraints, CFRange* fitRange) {
-    CGSize ret;
-    if (fitRange)
-        *fitRange = stringRange;
-
-    CGRect frameSize = { 0, 0, 0, 0 };
+    CGRect frameSize = { 0 };
     frameSize.size = constraints;
 
-    _createFrame(static_cast<_CTFramesetter*>(framesetter), frameSize, &ret, false);
+    _CTFrame* frame = _createFrame(static_cast<_CTFramesetter*>(framesetter), frameSize, stringRange);
+    CGSize ret = frame ? frame->_frameRect.size : CGSize{ 0, 0 };
+
+    if (fitRange) {
+        *fitRange = CTFrameGetVisibleStringRange(static_cast<CTFrameRef>(frame));
+    }
+
+    [frame release];
     return ret;
 }
 
--- a/Frameworks/CoreText/CTFramesetter.mm
+++ b/Frameworks/CoreText/CTFramesetter.mm
@@ -42,7 +42,7 @@ static _CTFrame* _createFrame(_CTFramesetter* framesetter, CGRect frameSize, CFR
     }
     _CTFrame* ret = _DWriteGetFrame(typesetter, range, frameSize);
     ret->_framesetter = framesetter;
-    return [ret retain];
+    return ret;
 }
 
 /**
@@ -63,7 +63,7 @@ CTFrameRef CTFramesetterCreateFrame(CTFramesetterRef framesetter, CFRange string
 
     _CTFrame* ret = _createFrame(static_cast<_CTFramesetter*>(framesetter), frameSize, stringRange);
 
-    return static_cast<CTFrameRef>(ret);
+    return static_cast<CTFrameRef>([ret retain]);
 }
 
 /**
@@ -90,7 +90,6 @@ CGSize CTFramesetterSuggestFrameSizeWithConstraints(
         *fitRange = CTFrameGetVisibleStringRange(static_cast<CTFrameRef>(frame));
     }
 
-    [frame release];
     return ret;
 }
 
--- a/Frameworks/CoreText/CTFramesetter.mm
+++ b/Frameworks/CoreText/CTFramesetter.mm
@@ -30,7 +30,7 @@ using namespace std;
 }
 @end
 
-static _CTFrame* _createFrame(_CTFramesetter* framesetter, CGRect frameSize, CFRange range) {
+static _CTFrame* __CreateFrame(_CTFramesetter* framesetter, CGRect frameSize, CFRange range) {
     if (!framesetter) {
         return nil;
     }
@@ -40,9 +40,10 @@ static _CTFrame* _createFrame(_CTFramesetter* framesetter, CGRect frameSize, CFR
     if (range.length == 0) {
         range.length = typesetter->_characters.size();
     }
+
     _CTFrame* ret = _DWriteGetFrame(typesetter, range, frameSize);
     ret->_framesetter = framesetter;
-    return ret;
+    return [ret retain];
 }
 
 /**
@@ -61,9 +62,9 @@ CTFrameRef CTFramesetterCreateFrame(CTFramesetterRef framesetter, CFRange string
     CGRect frameSize;
     _CGPathGetBoundingBoxInternal(path, &frameSize);
 
-    _CTFrame* ret = _createFrame(static_cast<_CTFramesetter*>(framesetter), frameSize, stringRange);
+    _CTFrame* ret = __CreateFrame(static_cast<_CTFramesetter*>(framesetter), frameSize, stringRange);
 
-    return static_cast<CTFrameRef>([ret retain]);
+    return static_cast<CTFrameRef>(ret);
 }
 
 /**
@@ -83,13 +84,14 @@ CGSize CTFramesetterSuggestFrameSizeWithConstraints(
     CGRect frameSize = { 0 };
     frameSize.size = constraints;
 
-    _CTFrame* frame = _createFrame(static_cast<_CTFramesetter*>(framesetter), frameSize, stringRange);
+    _CTFrame* frame = __CreateFrame(static_cast<_CTFramesetter*>(framesetter), frameSize, stringRange);
     CGSize ret = frame ? frame->_frameRect.size : CGSize{ 0, 0 };
 
     if (fitRange) {
         *fitRange = CTFrameGetVisibleStringRange(static_cast<CTFrameRef>(frame));
     }
 
+    [frame release];
     return ret;
 }
 
--- a/Frameworks/CoreText/CTLine.mm
+++ b/Frameworks/CoreText/CTLine.mm
@@ -310,14 +310,15 @@ double CTLineGetTypographicBounds(CTLineRef lineRef, CGFloat* ascent, CGFloat* d
 
             DWRITE_FONT_METRICS fontMetrics;
             run->_dwriteGlyphRun.fontFace->GetMetrics(&fontMetrics);
-            CGFloat pointsPerDesignUnit = pointSize / fontMetrics.designUnitsPerEm;
 
             for (size_t i = 0; i < run->_dwriteGlyphRun.glyphCount; ++i) {
-                line->_ascent = std::max(line->_ascent, _CoreTextScaleMetric(glyphMetrics[i].verticalOriginY, pointsPerDesignUnit));
-                line->_descent = std::max(line->_descent, _CoreTextScaleMetric(glyphMetrics[i].bottomSideBearing, pointsPerDesignUnit));
+                line->_ascent =
+                    std::max(line->_ascent, _CoreTextScaleMetric(glyphMetrics[i].verticalOriginY, pointSize, fontMetrics.designUnitsPerEm));
+                line->_descent = std::max(line->_descent,
+                                          _CoreTextScaleMetric(glyphMetrics[i].bottomSideBearing, pointSize, fontMetrics.designUnitsPerEm));
             }
 
-            line->_leading = std::max(line->_leading, _CoreTextScaleMetric(fontMetrics.lineGap, pointsPerDesignUnit));
+            line->_leading = std::max(line->_leading, _CoreTextScaleMetric(fontMetrics.lineGap, pointSize, fontMetrics.designUnitsPerEm));
         }
     }
 
--- a/Frameworks/CoreText/CTLine.mm
+++ b/Frameworks/CoreText/CTLine.mm
@@ -311,14 +311,15 @@ double CTLineGetTypographicBounds(CTLineRef lineRef, CGFloat* ascent, CGFloat* d
             DWRITE_FONT_METRICS fontMetrics;
             run->_dwriteGlyphRun.fontFace->GetMetrics(&fontMetrics);
 
+            // Need scaling factor to convert from design units to pointSize
+            CGFloat scalingFactor = pointSize / fontMetrics.designUnitsPerEm;
+
             for (size_t i = 0; i < run->_dwriteGlyphRun.glyphCount; ++i) {
-                line->_ascent =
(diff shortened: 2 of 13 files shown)


tests/unittests/CoreText/CTLineTests.mm
```diff
@@ -257,4 +257,31 @@
...
+    CFAttributedStringRef variableString = (__bridge CFAttributedStringRef)getAttributedString(@"gH");
+    CTLineRef baseLine = CTLineCreateWithAttributedString(baseString);
+    CFAutorelease(baseLine);
+    CTLineRef variableLine = CTLineCreateWithAttributedString(variableString);
+    CFAutorelease(variableLine);
+
+    CGFloat baseAscent, variableAscent, baseDescent, variableDescent, baseLeading, variableLeading;
+    CGFloat baseWidth = CTLineGetTypographicBounds(baseLine, &baseAscent, &baseDescent, &baseLeading);
+    EXPECT_LT(0, baseWidth);
+    EXPECT_EQ_MSG(CTFontGetLeading(font), baseLeading, "Leading should always be the same as the font");
```
rajsesh-msft: Do these assertions hold good in reference platform?

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -493,6 +507,42 @@ static void __GetGlyphRunDetails(_CTTypesetter* ts, CFRange range, CGRect frameS
...
+            DWRITE_FONT_METRICS fontMetrics;
+            run->_dwriteGlyphRun.fontFace->GetMetrics(&fontMetrics);
+
+            line->_ascent =
+                std::accumulate(glyphMetrics,
+                                glyphMetrics + run->_dwriteGlyphRun.glyphCount,
+                                line->_ascent,
+                                [&](CGFloat ascent, DWRITE_GLYPH_METRICS metrics) {
+                                    return std::max(ascent,
+                                                    _CoreTextScaleMetric(metrics.verticalOriginY, pointSize, fontMetrics.designUnitsPerEm));
```
rajsesh-msft: >pointSize, fontMetrics.designUnitsPerEm [](start = 98, length = 39)

from perf standpoint it would be better to compute pointSize/fontMetrics.designUnitsPerEm once and use it.

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -493,6 +507,42 @@ static void __GetGlyphRunDetails(_CTTypesetter* ts, CFRange range, CGRect frameS
                 line->_width += glyphRunDetails._dwriteGlyphRun[i].glyphAdvances[index];
             }
 
+            DWRITE_GLYPH_METRICS glyphMetrics[run->_dwriteGlyphRun.glyphCount];
```
rajsesh-msft: Is there any reason to compute these upfront?  If the caller is not using these values, it would be a waste of time.  Rather initialize these from CTLineGetTypographicBounds?

rajsesh-msft:
:clock1:

tests/unittests/CoreText/CTLineTests.mm
aballway: According to their documentation on their typography stuff it should, and it has from what I've seen testing.

Frameworks/CoreText/DWriteWrapper.mm
aballway: They'll be used when determining the frame size when creating a frame as well, and we don't really have .

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -52,6 +53,17 @@ void __InitDWriteFontPropertiesFromName(
     return s_dwriteFactory;
 }
 
+// Private helper for converting a DWRITE_FONT_METRIC unit to a CTFont API unit
+// DWRITE_FONT_METRICS measures its metrics in 'design units'
+// CTFont APIs generally return in terms of point size
+CGFloat _CoreTextScaleMetric(CGFloat metric, CGFloat pointSizeUnitsPerEm) {
```
rajsesh-msft: nit: some of our private methods are __ and others are _.  what gives?

aballway: @ms-jihua suggested a naming convention of double underscore for private to the file, single underscore for private to the project

Frameworks/CoreText/DWriteWrapper.mm
rajsesh-msft: We should be able to get the frame bounds directly from dwrite instead of doing an O(N) operation like we do here.  How did that work before this change anyway?

---
In reply to: [80313171](https://github.com/Microsoft/WinObjC/pull/1027#discussion_r80313171) [](ancestors = 80313171)

aballway: Before we just got the maximum ascent and descent values from the font, which we could do here, but to be interoperable with the reference platform we need to only get the ascent and descent of the glyphs used in the line

---
In reply to: [80317746](https://github.com/Microsoft/WinObjC/pull/1027#discussion_r80317746) [](ancestors = 80317746,80313171)

bbowman:
FYI your review title doesn't make sense. Our version of CT* will never be interoperable with anything on the reference platform (they have their own version).

rajsesh-msft:
Never be interoperable in the sense of providing exact values, or functioning 'similarly'?  It should certainly do the later.

---
In reply to: [249301240](https://github.com/Microsoft/WinObjC/pull/1027#issuecomment-249301240) [](ancestors = 249301240)

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -52,6 +53,17 @@ void __InitDWriteFontPropertiesFromName(
...
+    // design units * (pt / em) / (design units / em)
+    // = design units * (pt / em) * (em / design units)
+    // = pt * (design units / design units) * (em / em)
+    // = pt
```
bbowman: commented out "code" is generally not my favorite. I understand this is "showing your work" but it may benefit from a line describing that instead of what just looks like commented out arithmetic.

bbowman:
compatible / matching / etc are fine. Interoperable means able to work with (specifically to communicate / pass data between) I believe. Mostly just a semantic difference that may mislead reviewers into thinking this review is trying to build and function on the reference platform and use / communicate components from it.

---
In reply to: [249301613](https://github.com/Microsoft/WinObjC/pull/1027#issuecomment-249301613) [](ancestors = 249301613,249301240)

bbowman:
Our "interoperable" tag on apis and other discussions of interoperability have been in reference to app code not to reference platform code I guess. I've never thought our other uses were weird and its not a big deal obviously.

---
In reply to: [249302879](https://github.com/Microsoft/WinObjC/pull/1027#issuecomment-249302879) [](ancestors = 249302879,249301613,249301240)

bbowman:
If we wanted to be crazy, i think the term AJ is really meaning would be "idemoperable" or something?

---
In reply to: [249303282](https://github.com/Microsoft/WinObjC/pull/1027#issuecomment-249303282) [](ancestors = 249303282,249302879,249301613,249301240)

bbowman:
:shipit:

rajsesh-msft:
Understood.  The pull request should simply say "implement CTLineGetTypographicBounds".  +1 for idemoperable.

---
In reply to: [249303752](https://github.com/Microsoft/WinObjC/pull/1027#issuecomment-249303752) [](ancestors = 249303752,249303282,249302879,249301613,249301240)

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -493,6 +507,29 @@ static void __GetGlyphRunDetails(_CTTypesetter* ts, CFRange range, CGRect frameS
...
+
+            CTFontRef font = static_cast<CTFontRef>([run->_attributes objectForKey:static_cast<NSString*>(kCTFontAttributeName)]);
+            CGFloat pointSize = kCTFontSystemFontSize;
+            if (font) {
+                pointSize = CTFontGetSize(font);
+            }
+
+            DWRITE_FONT_METRICS fontMetrics;
+            run->_dwriteGlyphRun.fontFace->GetMetrics(&fontMetrics);
+            CGFloat pointSizeUnitsPerEm = pointSize / fontMetrics.designUnitsPerEm;
```
ms-jihua: ...?

pointSize / designUnitsPerEm
= (pt / em) / (design unit/em)
= pt / design unit

this float is not what you labeled it...

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -52,6 +53,17 @@ void __InitDWriteFontPropertiesFromName(
...
 
+// Private helper for converting a DWRITE_FONT_METRIC unit to a CTFont API unit
+// DWRITE_FONT_METRICS measures its metrics in 'design units'
+// CTFont APIs generally return in terms of point size
+CGFloat _CoreTextScaleMetric(CGFloat metric, CGFloat pointSizeUnitsPerEm) {
+    // design units * (pt / em) / (design units / em)
+    // = design units * (pt / em) * (em / design units)
+    // = pt * (design units / design units) * (em / em)
+    // = pt
+    return metric * pointSizeUnitsPerEm;
```
ms-jihua: >return metric * pointSizeUnitsPerEm; [](start = 4, length = 36)

uhh, this just evaluates to

design unit * pt/em

which doesn't reduce to anything... what final unit were you hoping to get out here?

Frameworks/CoreText/DWriteWrapper.mm
ms-jihua: these comments don't actually apply to the math here anymore also...

---
In reply to: [80326649](https://github.com/Microsoft/WinObjC/pull/1027#discussion_r80326649) [](ancestors = 80326649)

ms-jihua:
:clock1:

aballway:
Due to some overlapping changes, I added CTFramesetterSuggestFrameSizeWithConstraints to this commit. 

---
In reply to: [249305565](https://github.com/Microsoft/WinObjC/pull/1027#issuecomment-249305565) [](ancestors = 249305565,249303752,249303282,249302879,249301613,249301240)

Frameworks/CoreText/CTLine.mm
```diff
@@ -283,16 +283,44 @@ CGRect CTLineGetImageBounds(CTLineRef line, CGContextRef context) {
...
 double CTLineGetTypographicBounds(CTLineRef lineRef, CGFloat* ascent, CGFloat* descent, CGFloat* leading) {
-    if (!lineRef) {
+    _CTLine* line = static_cast<_CTLine*>(lineRef);
+
+    if (!line || [line->_runs count] == 0) {
         return 0;
     }
 
-    _CTLine* line = static_cast<_CTLine*>(lineRef);
+    // Created with impossible values FLT_MIN which signify they need to be populated
```
ms-jihua: >// Created with impossible values FLT_MIN which signify they need to be populated [](start = 4, length = 81)

I'm not so sure about this... 
1) what's your reasoning here that FLT_MIN is an impossible, rather than merely improbable value?
2) it's not super evident that this was the intended purpose of FLT_MIN in DWriteWrapper

Frameworks/CoreText/CTLine.mm
```diff
@@ -283,16 +283,44 @@ CGRect CTLineGetImageBounds(CTLineRef line, CGContextRef context) {
...
+
+            CTFontRef font = static_cast<CTFontRef>([run->_attributes objectForKey:static_cast<NSString*>(kCTFontAttributeName)]);
+            CGFloat pointSize = kCTFontSystemFontSize;
+            if (font) {
+                pointSize = CTFontGetSize(font);
+            }
+
+            DWRITE_FONT_METRICS fontMetrics;
+            run->_dwriteGlyphRun.fontFace->GetMetrics(&fontMetrics);
+            CGFloat pointsPerDesignUnit = pointSize / fontMetrics.designUnitsPerEm;
```
ms-jihua: If you end up having to do this pointSize / designUnitsPerEm here as well as everywhere else, it'll probably be better if you make _CoreTextScaleMetric a three-parameter function instead of two-and-the-caller-is-expected-to-know-to-divide-first.

Frameworks/CoreText/CTLine.mm
aballway: I'll add a comment. I thought FLT_MIN was along the lines of INT_MIN and the most negative value, not the smallest absolute value. I'll change it to FLT_MAX (which should be so far beyond a reasonable value that it can be considered impossible)

Frameworks/CoreText/CTLine.mm
aballway: This was done to keep from having to do this division every time the helper was called by Raj's comment.

---
In reply to: [80528495](https://github.com/Microsoft/WinObjC/pull/1027#discussion_r80528495) [](ancestors = 80528495)

Frameworks/CoreText/CTLine.mm
ms-jihua: that sounds much better : )

---
In reply to: [80529650](https://github.com/Microsoft/WinObjC/pull/1027#discussion_r80529650) [](ancestors = 80529650)

Frameworks/CoreText/CTLine.mm
ms-jihua: Raj's comment was made at a time when the division result was already outside the function, so you might as well have cached it. In general I have the following issues with _CoreTextScaleMetric:

- The name is not very informative. For CTFont, __CTFontScaleMetric was more forgivable since the context was smaller, and the scope of what it could be was smaller. _CoreTextScaleMetric on the other hand is in DWriteWrapper, and could be any number of things. A better name might be something like _CoreTextDesignMetricToPoints()
- Changing the name to that though, makes it pretty obvious that the function just hides a multiplication, in which case it feels a bit silly to have a dedicated helper for it?
- It feels extra silly to have a dedicated helper for it and also assume that the caller will have the knowledge to do this division outside, rather than shifting the burden to the function.
- In general, caching that result is a pretty small optimization, and we'd prefer easier-to-read/maintain code over the small gains that caching would get us. I'm fine with leaving this function the way it is if you think this is the most readable state, but I don't want optimization to be your foremost reason.

---
In reply to: [80529943](https://github.com/Microsoft/WinObjC/pull/1027#discussion_r80529943) [](ancestors = 80529943,80528495)

aballway: Changed.  I think you're right, and the performance gain doesn't compare to how less useful the helper was made.

---
In reply to: [80534274](https://github.com/Microsoft/WinObjC/pull/1027#discussion_r80534274) [](ancestors = 80534274,80529943,80528495)

Frameworks/CoreText/CTFramesetter.mm
```diff
@@ -105,19 +75,22 @@ CTTypesetterRef CTFramesetterGetTypesetter(CTFramesetterRef framesetter) {
...
 
-    _createFrame(static_cast<_CTFramesetter*>(framesetter), frameSize, &ret, false);
+    _CTFrame* frame = _createFrame(static_cast<_CTFramesetter*>(framesetter), frameSize, stringRange);
+    CGSize ret = frame ? frame->_frameRect.size : CGSize{ 0, 0 };
+
+    if (fitRange) {
+        *fitRange = CTFrameGetVisibleStringRange(static_cast<CTFrameRef>(frame));
+    }
+
+    [frame release];
```
mnithish: >[frame release]; [](start = 4, length = 16)

This is an extra release. _createFrame returns an autorelease object

Frameworks/CoreText/CTFramesetter.mm
```diff
@@ -30,48 +30,19 @@ - (void)dealloc {
...
-        // Modify this code once we move away from Cairo Freetype for rasterizing text.
-        CGFloat originYFirstLine = frameSize.size.height - firstLine->_lineOrigin.y - frameSize.origin.y;
-        CGFloat originYLastLine = frameSize.size.height - lastLine->_lineOrigin.y - frameSize.origin.y;
-        sizeOut->height = min(originYLastLine + originYFirstLine, frameSize.size.height);
-    }
-
-    return ret;
+    _CTFrame* ret = _DWriteGetFrame(typesetter, range, frameSize);
+    ret->_framesetter = framesetter;
+    return [ret retain];
```
mnithish: >[ret retain] [](start = 11, length = 12)

OK this is a bit confusing - ideally only *Ref objects returned by a function should be retained. All others should be auto released.

The reason why I say this gets confusing is, this function returns a  _CTFrame* object that is retained but _DWriteGetFrame also returns a _CTFrame* object that is not retained. Let keep the memory management consistent so it is easy to read

Frameworks/CoreText/CTFramesetter.mm
```diff
@@ -30,47 +30,18 @@ - (void)dealloc {
...
-        // TODO::
-        // Modify this code once we move away from Cairo Freetype for rasterizing text.
-        CGFloat originYFirstLine = frameSize.size.height - firstLine->_lineOrigin.y - frameSize.origin.y;
-        CGFloat originYLastLine = frameSize.size.height - lastLine->_lineOrigin.y - frameSize.origin.y;
-        sizeOut->height = min(originYLastLine + originYFirstLine, frameSize.size.height);
-    }
-
+    _CTFrame* ret = _DWriteGetFrame(typesetter, range, frameSize);
+    ret->_framesetter = framesetter;
     return ret;
```
mnithish: >ret [](start = 11, length = 3)

[[ret retain] autorelease] - this will guarantee the autorelease object stays until the function that is calling this returns

mnithish:
:shipit:

Frameworks/UIKit/NSLayoutManager.mm
```diff
@@ -271,10 +271,11 @@ - (NSTextStorage*)textStorage {
 }
 
 /**
- @Status Interoperable
```
rajsesh-msft: :D

Frameworks/CoreText/DWriteWrapper.mm
```diff
@@ -430,17 +423,26 @@ static void __GetGlyphRunDetails(_CTTypesetter* ts, CFRange range, CGRect frameS
...
+    ComPtr<CustomDWriteTextRenderer> textRenderer = Make<CustomDWriteTextRenderer>();
     _DWriteGlyphRunDetails glyphRunDetails = {};
-    __GetGlyphRunDetails(ts, range, frameSize, &glyphRunDetails);
+    textLayout->Draw(&glyphRunDetails, textRenderer.Get(), 0, 0);
+    DWRITE_TEXT_METRICS textMetrics;
+    THROW_IF_FAILED(textLayout->GetMetrics(&textMetrics));
+    frame->_frameRect = frameSize;
+
+    // TODO:: find more precise value than 1.0 to increase width by to fully enclose frame
+    frame->_frameRect.size.width = std::min(frameSize.size.width, textMetrics.widthIncludingTrailingWhitespace + 1.0f);
```
rajsesh-msft: >1 [](start = 113, length = 1)

what is the 1.0 accomplishing?

Frameworks/CoreText/CTFramesetter.mm
rajsesh-msft: Oh gosh, this is not right.  per convention, create functions must retain.  and Get functions are auto released.  The previous iteration was correct.

Frameworks/CoreText/CTLine.mm
rajsesh-msft: I don't agree that performance in tight loop doesn't matter.  This is a case of having a helper function where one is not needed.  You need a multiplication factor, which you compute once and use everywhere.  floating point divisions are expensive and when done many times (as in this case, we would do it with every glyph) will result in "why the hell is it performing so bad".  Granted, premature optimization is bad, but this is a case where we aren't gaining anything from a helper function. 

---
In reply to: [80552492](https://github.com/Microsoft/WinObjC/pull/1027#discussion_r80552492) [](ancestors = 80552492,80534274,80529943,80528495)

ms-jihua: Then, I would say that we just get rid of the helper function except in CTFont where it makes sense still :P

---
In reply to: [80596585](https://github.com/Microsoft/WinObjC/pull/1027#discussion_r80596585) [](ancestors = 80596585,80552492,80534274,80529943,80528495)

Frameworks/CoreText/CTFramesetter.mm
aballway: But this is returning an object not a .*Ref, and creation methods return autoreleased objects no?

---
In reply to: [80596227](https://github.com/Microsoft/WinObjC/pull/1027#discussion_r80596227) [](ancestors = 80596227)

Frameworks/CoreText/DWriteWrapper.mm
aballway: Dwrite won't write all of the text in the frame if it is within some small margin of the width required, but I'm not sure what the precise value is nor have I found any documentation on what it should be.

---
In reply to: [80596031](https://github.com/Microsoft/WinObjC/pull/1027#discussion_r80596031) [](ancestors = 80596031)

Frameworks/CoreText/CTLine.mm
aballway: I don't think the redundant divisions we're doing here will cause a noticeable performance hit.  The number of glyphs in a line will be limited so we don't need to worry about extremely large N, and I think it makes sense to have this as a helper so when someone goes to change something in the future/an outside developer wants to peek in they won't be faced with more optimized but less understandable arithmetic 

---
In reply to: [80598919](https://github.com/Microsoft/WinObjC/pull/1027#discussion_r80598919) [](ancestors = 80598919,80596585,80552492,80534274,80529943,80528495)

Frameworks/CoreText/CTFramesetter.mm
ms-jihua: If it has 'Create' in the name, it'll generally indicate that it returns retained. ObjC-style convenience constructors are generally named something like "frameWith...", without a 'create'.

---
In reply to: [80719399](https://github.com/Microsoft/WinObjC/pull/1027#discussion_r80719399) [](ancestors = 80719399,80596227)

Frameworks/CoreText/CTFramesetter.mm
```diff
@@ -90,10 +61,9 @@ CTFrameRef CTFramesetterCreateFrame(CTFramesetterRef framesetter, CFRange string
     CGRect frameSize;
     _CGPathGetBoundingBoxInternal(path, &frameSize);
 
-    CGSize sizeOut;
-    _CTFrame* ret = _createFrame((_CTFramesetter*)framesetter, frameSize, &sizeOut, true);
+    _CTFrame* ret = _createFrame(static_cast<_CTFramesetter*>(framesetter), frameSize, stringRange);
 
-    return static_cast<CTFrameRef>(ret);
+    return static_cast<CTFrameRef>([ret retain]);
```
DHowett-MSFT: ><CTFrameRef>([ret retain]); [](start = 22, length = 27)

If you move the retain into `_createFrame`, you'll get this for free.

DHowett-MSFT:
:clock1:

PR #1043, opened 2016-09-27
Make NSJSON test more lenient for reference platform

### LINKED ISSUE #969: NSJSON.DataWithJSONObjectTests fails on reference platform
NSJSONSerializationTests.m:56: Failure
Value of: actualResult
  Actual: {"key":"value","nullkey":null,"array":[{"subnullkey":null}]}
Expected: expectedResult
Which is: {"key":"value","array":[{"subnullkey":null}],"nullkey":null}
FAILED: expected ({"key":"value","array":[{"subnullkey":null}],"nullkey":null}) != actual ({"key":"value","nullkey":null,"array":[{"subnullkey":null}]})
[  FAILED  ] NSJSON.DataWithJSONObjectTests (0 ms)

The windows JSON serializer does not serialize the json in the same order as reference platform.  This in itself is not an issue, just need to fix the test to not be so strict.

aballway:
One test was too strict and caused failure because of implementation differences

Fixes #969

### DIFF
--- a/tests/unittests/Foundation/NSJSONSerializationTests.m
+++ b/tests/unittests/Foundation/NSJSONSerializationTests.m
@@ -53,7 +53,7 @@ void VerifyDataWithJSONObjectSucceeds(id testObject, NSString* expectedResult) {
     NSString* actualResult = [[[NSString alloc] initWithData:result encoding:NSUTF8StringEncoding] autorelease];
 
     ASSERT_EQ_MSG(nil, err, "FAILED: err != nil");
-    ASSERT_OBJCEQ_MSG(expectedResult, actualResult, "FAILED: expected (%@) != actual (%@)", expectedResult, actualResult);
+    ASSERT_OBJCEQ(expectedResult, actualResult);
 }
 
 void VerifyDataWithJSONObjectFails(id testObject, NSInteger expectedErrorCode, NSString* expectedErrorDescription) {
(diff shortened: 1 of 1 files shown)


ms-jihua:
:shipit:

tests/unittests/Foundation/NSJSONSerializationTests.m
```diff
@@ -53,7 +53,7 @@ void VerifyDataWithJSONObjectSucceeds(id testObject, NSString* expectedResult) {
     NSString* actualResult = [[[NSString alloc] initWithData:result encoding:NSUTF8StringEncoding] autorelease];
 
     ASSERT_EQ_MSG(nil, err, "FAILED: err != nil");
-    ASSERT_OBJCEQ_MSG(expectedResult, actualResult, "FAILED: expected (%@) != actual (%@)", expectedResult, actualResult);
```
mnithish: >ASSERT_OBJCEQ_MSG [](start = 4, length = 17)

Why was MSG macro removed here?

DHowett-MSFT:
:clock:

rajsesh-msft:
We do support NULL key.  The order of the keys is not the same in the serialized JSON as it is in the reference platform.  

---
In reply to: [250037368](https://github.com/Microsoft/WinObjC/pull/1043#issuecomment-250037368) [](ancestors = 250037368)

tests/unittests/Foundation/NSJSONSerializationTests.m
```diff
@@ -124,13 +124,13 @@ void VerifyDataWithJSONObjectThrows(id testObject) {
     id testObject6 = @{ @"foo" : @"1", @"bar" : @"2" };
     id testObject7 = @{ @"foo" : @{ @"bar" : @"1" }, @"foo2" : @{ @"bar2" : @"2" } };
     id testObject8 = [NSUUID UUID];
-    id testObject9 = @{ @"key" : @"value", @"nullkey" : [NSNull null], @"array" : @[ @{ @"subnullkey" : [NSNull null] } ] };
```
rajsesh-msft: >NSNull [](start = 57, length = 6)

I am not sure if we should remove this.  Would be good to see if there is a bug in our recursion that ends up adding the elements to the JSON serializer in the wrong order?

rajsesh-msft:
:clock1:

tests/unittests/Foundation/NSJSONSerializationTests.m
aballway: It's determined by the order in which CFDictionary returns the keys for fastenumeration, which I a) really don't want to mess with and b) think is far outside of the scope of this issue.

---
In reply to: [80824620](https://github.com/Microsoft/WinObjC/pull/1043#discussion_r80824620) [](ancestors = 80824620)

rajsesh-msft: We need NSNull to work correctly as a key here.  Lets due the diligence here..

---
In reply to: [80960041](https://github.com/Microsoft/WinObjC/pull/1043#discussion_r80960041) [](ancestors = 80960041,80824620)

DHowett-MSFT: Alternatively, we could reconstruct these test cases to test single-element dictionaries. We can't lose fidelity, though, so we need to test everything we would have otherwise tested for a multi-member dictionary.

---
In reply to: [80983725](https://github.com/Microsoft/WinObjC/pull/1043#discussion_r80983725) [](ancestors = 80983725,80978644,80960041,80824620)

rajsesh-msft: We are already using a fully tested third party json parser - the windows json serializer, so I think you are both right - we could live with the simplification of the test.

---
In reply to: [80983725](https://github.com/Microsoft/WinObjC/pull/1043#discussion_r80983725) [](ancestors = 80983725,80978644,80960041,80824620)

rajsesh-msft: That is a fine idea. I think we have sufficient coverage as is, so we could punt it until we need to revamp this.

---
In reply to: [80984188](https://github.com/Microsoft/WinObjC/pull/1043#discussion_r80984188) [](ancestors = 80984188,80983725,80978644,80960041,80824620)

Issue #970, opened 2016-09-07
Implement NSIndexSet indexLessThanIndex 


rajsesh-msft:
This is required by app partner.

PR #1051, opened 2016-09-29
Implement NSIndexSet secure coding, greater than, and less than methods

### LINKED ISSUE #936: NSIndexSet needs to implement secure coding protocol
encodeWithCode and initWithCoder are stubbed and required by some middleware.

aballway:
Implements initWithCoder and encodeWithCoder for NSCoding and the index(Greater/Less)Than* methods.

Fixes #936 and #970

### DIFF
--- a/Frameworks/Foundation/NSIndexSet.mm
+++ b/Frameworks/Foundation/NSIndexSet.mm
@@ -1,6 +1,6 @@
 //******************************************************************************
 //
-// Copyright (c) 2015 Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
 //
 // This code is licensed under the MIT License (MIT).
 //
@@ -29,7 +29,11 @@
 #include "NSIndexSetInternal.h"
 #include "Etc.h"
 
+#include <algorithm>
+
 @implementation NSIndexSet
+
+static NSString* c_archiveKey = @"_NS.IS";
 
 /**
  @Status Interoperable
@@ -64,7 +68,7 @@
 */
 - (instancetype)initWithIndexesInRange:(NSRange)indexRange {
     if (self = [super init]) {
-        [(NSMutableIndexSet*)self _addItem:indexRange];
+        [static_cast<NSMutableIndexSet*>(self) _addItem:indexRange];
     }
     return self;
 }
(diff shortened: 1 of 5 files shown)


Frameworks/Foundation/NSIndexSet.mm
```diff
@@ -64,7 +68,7 @@ - (instancetype)initWithIndex:(NSUInteger)index {
 */
 - (instancetype)initWithIndexesInRange:(NSRange)indexRange {
     if (self = [super init]) {
-        [(NSMutableIndexSet*)self _addItem:indexRange];
+        [static_cast<NSMutableIndexSet*>(self) _addItem:indexRange];
```
DHowett-MSFT: >static_cast [](start = 9, length = 11)

This seems like a surprising amount of churn for little gain.

Frameworks/Foundation/NSIndexSet.mm
```diff
@@ -471,26 +498,44 @@ - (void)enumerateIndexesInRange:(NSRange)range options:(NSEnumerationOptions)opt
...
+ @Status Interoperable
 */
 - (void)encodeWithCoder:(NSCoder*)coder {
-    UNIMPLEMENTED();
+    unsigned int count = [static_cast<NSMutableIndexSet*>(self) _count];
+    if (count > 0) {
+        NSRangePointer ranges = [static_cast<NSMutableIndexSet*>(self) _allRanges];
+        if ([coder isKindOfClass:[NSKeyedArchiver class]]) {
+            [coder encodeBytes:reinterpret_cast<uint8_t*>(ranges) length:count * sizeof(NSRange) forKey:c_archiveKey];
+        } else {
```
DHowett-MSFT: This will all but guarantee that range archives serialized on 32-bit platforms are unusable on 64-bit ones, and vice versa. As archives are intended for data portability, this may not be great.

Unfortunately, I don't have a proper solution here; we may want to discuss it.

rajsesh-msft: I am not sure if *this* is the only place where we will fail in archival compat across architecture types.  

---
In reply to: [81209959](https://github.com/Microsoft/WinObjC/pull/1051#discussion_r81209959) [](ancestors = 81209959)

Frameworks/Foundation/NSIndexSet.mm
aballway: wrt changing casts? I ended up just changing them all when I fixed mine

---
In reply to: [81208628](https://github.com/Microsoft/WinObjC/pull/1051#discussion_r81208628) [](ancestors = 81208628)

Frameworks/Foundation/NSIndexSet.mm
aballway: discussed offline - will keep as is until a more thorough fix of archiving is done

---
In reply to: [81215648](https://github.com/Microsoft/WinObjC/pull/1051#discussion_r81215648) [](ancestors = 81215648,81209959)

DHowett-MSFT:
:shipit:

rajsesh-msft:
:shipit:

Issue #649, opened 2016-07-14
Binary property lists and keyed archives from the reference platform may behave badly


DHowett-MSFT:
Due to the lack of a generic umbrella bridged type, `CFKeyedArchiverUID`s, which is used in keyed archives from iOS and OS X, will not be deserialized properly. This will result in a crash when you attempt to use `NSKeyedUnarchiver` or `NSPropertyListSerialization` with archives created on the reference platform.

Normal property lists from iOS and OS X are not affected, as they will not contain UID objects.

XML archives are not affected, as they encode UIDs via `NSDictionary` (as `<dict><key>CF$UID</key><integer>1</integer></dict>`).

Reproduction case:

``` objc
// The only way to represent a raw CFKeyedArchiverUID is via binary property list.
// This property list contains:
// 0x00:        u8[8] header                       (bplist00)
// -- object table --
// 0x08: UID (0x80 0xFF)
// -- offset table --
// 0x0A: 0x08
// -- plist trailer --
// 0x0B - 0x0F: u8[5] unused                       (0x00 * 5)
// 0x10:        u8    sort version                 (0x00)
// 0x11:        u8    size of offset table entries (0x01)
// 0x12:        u8    size of embedded objects     (0x01)
// 0x13 - 0x1A: u64   number of objects            (0x01)
// 0x1B - 0x22: u64   toplevel object              (0x00)
// 0x23 - 0x2A: u64   offset to offset table       (0x0A)

const char rawPlist[] = {
    0x62, 0x70, 0x6c, 0x69, 0x73, 0x74, 0x30, 0x30, 0x80, 0xff, 0x08, 0x00,
    0x00, 0x00, 0x00, 0x00, 0x00, 0x01, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00,
    0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x0a
};
NSData* plistData = [NSData dataWithBytes:rawPlist length:_countof(rawPlist)];
NSObject* UID = [NSPropertyListSerialization propertyListFromData:plistData mutabilityOption:0 format:nullptr errorDescription:nullptr];
```

DHowett-MSFT:
This is no longer a crash, but it still represents a small compliance issue (some number of binary property lists, which we claim to support, will not act appropriately.)

[...]

PR #1059, opened 2016-09-30
Add an ObjC class to act as the parent of all CF types.

DHowett-MSFT:
This pull request intends to both complete #719 and fix the crash behind #649 (#649 is both a compliance bug _and_ a crash bug.).

True Objective-C memory management for CoreFoundation (as seen [here](https://github.com/DHowett-MSFT/WinObjC/compare/201609-nscftype?expand=1)) is infeasible right now. We have too few fiddly bitfields and subclasses to determine with any accuracy whether an object was statically-allocated or heap-allocated, and we almost certainly don't want to manipulate the nonexistent reference count on non-heap-allocated objects.

Herein, therefore, we try to solve two of the major Objective-C <-> CF bridging compliance issues:
* Not every CF type has an Objective-C parallel (and as such cannot live in an Objective-C collection.)
* CF objects cannot be zombified.

### DIFF
--- a/Frameworks/CoreFoundation/Base.subproj/CFRuntime.c
+++ b/Frameworks/CoreFoundation/Base.subproj/CFRuntime.c
@@ -1129,6 +1129,7 @@ void __CFInitialize(void) {
         memset(__CFRuntimeObjCClassTable, 0, sizeof(__CFRuntimeObjCClassTable));
 
 #if WINOBJC
+        // WINOBJC: Under the WinObjC runtime, all CFTypeRefs are _NSCFTypes or a toll-free bridged type
         for (CFIndex idx = 1; idx < __CFRuntimeClassTableSize; ++idx) {
             __CFRuntimeObjCClassTable[idx] = (uintptr_t)&_OBJC_CLASS__NSCFType;
         }
--- a/Frameworks/CoreFoundation/Base.subproj/CFRuntime.c
+++ b/Frameworks/CoreFoundation/Base.subproj/CFRuntime.c
@@ -291,9 +291,19 @@ void _CFRuntimeUnregisterClassWithTypeID(CFTypeID typeID) {
 CF_PRIVATE uint8_t __CFZombieEnabled = 0;
 CF_PRIVATE uint8_t __CFDeallocateZombies = 0;
 
-extern void __CFZombifyNSObject(void);  // from NSObject.m
+// WINOBJC: Since NSObject lives in Foundation, we have to invert the dependency here;
+// like in libobjc2, Foundation will set a callback to fire on object deallocation when
+// zombies are turned on.
+// extern void __CFZombifyNSObject(void);  // from NSObject.m
+void (*__CFZombifyNSObjectHook)(id);
 
 void _CFEnableZombies(void) {
+    __CFZombieEnabled = 1;
+}
+
+// WINOBJC: Zombies can be disabled.
+void _CFDisableZombies(void) {
+    __CFZombieEnabled = 0;
 }
 
 #endif /* DEBUG */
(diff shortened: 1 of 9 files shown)


ms-jihua:
* Not every CF type has an Objective-C parallel (and as such cannot live in an Objective-C collection.)

Thank you for this this has been a thorn in my side for forever

Frameworks/CoreFoundation/Base.subproj/CFRuntime.c
```diff
@@ -1118,6 +1128,11 @@ void __CFInitialize(void) {
         memset(__CFRuntimeClassTable, 0, sizeof(__CFRuntimeClassTable));
         memset(__CFRuntimeObjCClassTable, 0, sizeof(__CFRuntimeObjCClassTable));
 
+#if WINOBJC
```
ms-jihua: WINOBJC comment?

Frameworks/CoreFoundation/Base.subproj/CFRuntime.c
```diff
@@ -1118,6 +1128,11 @@ void __CFInitialize(void) {
...
+        for (CFIndex idx = 1; idx < __CFRuntimeClassTableSize; ++idx) {
+            __CFRuntimeObjCClassTable[idx] = (uintptr_t)&_OBJC_CLASS__NSCFType;
+        }
```
ms-jihua: depending on how you feel about stl here, you could also use std::fill

ms-jihua:
:shipit:

tests/unittests/Foundation/CoreFoundationTests.m
```diff
@@ -56,3 +57,19 @@ static uint64_t walltime(const struct mach_timebase_info& timebase, uint64_t val
...
+    CFTreeContext treeContext{};
+    CFTypeRef tree = static_cast<CFTypeRef>(CFTreeCreate(kCFAllocatorDefault, &treeContext));
+
+    NSArray* array = nil;
+
+    // The ',' in this method call upsets EXPECT_NO_THROW(...), so we need to hide it behind a lambda.
+    auto create = [&tree](){
+        return [[NSArray alloc] initWithObjects:static_cast<id>(tree), nil];
+    };
+    EXPECT_NO_THROW(array = create());
```
rajsesh-msft: >EXPECT_NO_THROW [](start = 4, length = 15)

Not sure if these are ARM safe.

rajsesh-msft:
:shipit:

mnithish:
:shipit:

Frameworks/Foundation/NSObject.mm
```diff
@@ -705,11 +708,36 @@ void WinObjC_SetMissingSelectorFatal(BOOL fatal) {
...
++ (BOOL)isSubclassOfClass:(Class)classRef {
+    Class curClass = self;
+
+    while (curClass != nil) {
+        if (curClass == classRef) {
+            return YES;
+        }
+
+        curClass = class_getSuperclass(curClass);
+    }
```
aballway: would this work as a for loop?

aballway:
:shipit:

Frameworks/Foundation/NSObject.mm
DHowett-MSFT: would this matter at all?

aballway: nope, just think it looks a bit cleaner. Shoulda been a nit, my b

---
In reply to: [81378408](https://github.com/Microsoft/WinObjC/pull/1059#discussion_r81378408) [](ancestors = 81378408)

Frameworks/CoreFoundation/Base.subproj/CFRuntime.c
DHowett-MSFT: I'd rather not get my peanut butter in my chocolate over here :)

---
In reply to: [81260899](https://github.com/Microsoft/WinObjC/pull/1059#discussion_r81260899) [](ancestors = 81260899)

tests/unittests/Foundation/CoreFoundationTests.m
DHowett-MSFT: `EXPECT_*_THROW` are ARM-safe; it's `@catch(NSException*)` that's not ARM-safe :)

---
In reply to: [81263134](https://github.com/Microsoft/WinObjC/pull/1059#discussion_r81263134) [](ancestors = 81263134)

PR #1061, opened 2016-09-30
Do not re-create text views on every text draw in CTCatalog test app

### LINKED ISSUE #1046: CTCatalog testapp is re-creating text views on every text draw causing the view to flicker
The bad flicker in the test app when we slide the slider to change fonts is due to this. We seem to be deleting and re-creating text views on every text draw call.

mnithish:
Fixes #1046


mnithish:

rajsesh-msft:
:shipit:

ms-jihua:
:shipit:

DHowett-MSFT:
:shipit:

aballway:
:shipit:

aballway:
Wouldn't you know there's a better way of updating views than murdering and replacing them

PR #1085, opened 2016-10-05
D2DFactory Wrapper

MSFTFox:
This is to provide a static helper function for creating factories to create the D2D objects that CG will be using very frequently.


build/CoreGraphics/lib/CoreGraphicsLib.vcxproj
```diff
@@ -55,6 +55,10 @@
     <ClangCompile Include="$(MSBuildThisFileDirectory)..\..\..\Frameworks\CoreGraphics\CGPDFStream.mm" />
     <ClangCompile Include="$(MSBuildThisFileDirectory)..\..\..\Frameworks\CoreGraphics\CGPDFString.mm" />
     <ClangCompile Include="$(MSBuildThisFileDirectory)..\..\..\Frameworks\CoreGraphics\CGShading.mm" />
+    <ClangCompile Include="..\..\..\Frameworks\CoreGraphics\D2DWrapper.mm" />
+  </ItemGroup>
+  <ItemGroup>
+    <ClInclude Include="..\..\..\Frameworks\include\CoreGraphics\D2DWrapper.h" />
```
MSFTFox: Fix include paths.

Frameworks/include/CoreGraphics/D2DWrapper.h
```diff
@@ -0,0 +1,30 @@
...
+//******************************************************************************
+
+#pragma once
+
+#include <COMIncludes.h>
+#import <wrl/implements.h>
+#import <D2d1.h>
+#include <COMIncludes_End.h>
+
+#import <Starboard.h>
```
msft-Jeyaram: is it wise to have starboard include in headers?

Frameworks/include/CoreGraphics/D2DWrapper.h
```diff
@@ -0,0 +1,30 @@
...
+
+using namespace std;
+using namespace Microsoft::WRL;
```
msft-Jeyaram: Should we just put this in the header file, it would be safer.

msft-Jeyaram:
:shipit:

Frameworks/CoreGraphics/D2DWrapper.mm
```diff
@@ -0,0 +1,33 @@
...
+// Helper for creating a D2DFactory
+static ComPtr<ID2D1Factory> _CreateD2DFactoryInstance(D2D1_FACTORY_TYPE type) {
+    ComPtr<ID2D1Factory> d2dFactory;
+    THROW_IF_FAILED(D2D1CreateFactory(type, d2dFactory.ReleaseAndGetAddressOf()));
+    return d2dFactory;
+}
+
+// Helper for accessing a single static multithreaded D2DFactory instance
+static ComPtr<ID2D1Factory> _GetD2DFactoryInstance() {
+    static ComPtr<ID2D1Factory> s_d2dFactory = _CreateD2DFactoryInstance(D2D1_FACTORY_TYPE_MULTI_THREADED);
```
aballway: >D2D1_FACTORY_TYPE_MULTI_THREADED [](start = 73, length = 32)

will we ever need a single-threaded d2dfactory? If not, it would be better to just have the Get method rather than another create one

Frameworks/include/CoreGraphics/D2DWrapper.h
```diff
@@ -0,0 +1,28 @@
...
+
+#pragma once
+
+#include <COMIncludes.h>
+#import <wrl/implements.h>
+#import <D2d1.h>
+#include <COMIncludes_End.h>
+
+using namespace std;
+using namespace Microsoft::WRL;
```
DHowett-MSFT: Headers shouldn't leak namespaces into their consumers; ideally, we would fully qualify every type in the header, and use the `using` directive in the `.mm` or `.cpp` files.

Frameworks/include/CoreGraphics/D2DWrapper.h
```diff
@@ -0,0 +1,28 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#pragma once
+
+#include <COMIncludes.h>
+#import <wrl/implements.h>
```
DHowett-MSFT: This should be `wrl/client.h`, since we're only using `ComPtr`

Frameworks/CoreGraphics/D2DWrapper.mm
DHowett-MSFT: I'm torn on this. CGPath can be manipulated from any thread, I believe, but CGContext cannot.
I'm not certain whether objects need to be derived from the same factory to interoperate.

---
In reply to: [82000836](https://github.com/Microsoft/WinObjC/pull/1085#discussion_r82000836) [](ancestors = 82000836)

DHowett-MSFT: Ah, I see. [This document](https://msdn.microsoft.com/en-us/library/windows/desktop/dd756757%28v=vs.85%29.aspx) clarifies cross-factory and cross-device use.

---
In reply to: [82023742](https://github.com/Microsoft/WinObjC/pull/1085#discussion_r82023742) [](ancestors = 82023742,82000836)

build/CoreGraphics/lib/CoreGraphicsLib.vcxproj
```diff
@@ -55,6 +55,10 @@
     <ClangCompile Include="$(MSBuildThisFileDirectory)..\..\..\Frameworks\CoreGraphics\CGPDFStream.mm" />
     <ClangCompile Include="$(MSBuildThisFileDirectory)..\..\..\Frameworks\CoreGraphics\CGPDFString.mm" />
     <ClangCompile Include="$(MSBuildThisFileDirectory)..\..\..\Frameworks\CoreGraphics\CGShading.mm" />
+    <ClangCompile Include="$(MSBuildThisFileDirectory)..\..\..\Frameworks\CoreGraphics\D2DWrapper.mm" />
+  </ItemGroup>
+  <ItemGroup>
+    <ClInclude Include="$(MSBuildThisFileDirectory)..\..\..\Frameworks\include\CoreGraphics\D2DWrapper.h" />
```
ms-jihua: >Frameworks\include\ [](start = 60, length = 19)

Any reason this is in Frameworks\include? DWriteWrapper is in Frameworks\include because we have some lower-level tests that exercise it - do you have similar plans for this file?

Frameworks/CoreGraphics/D2DWrapper.mm
```diff
@@ -0,0 +1,33 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#import <CoreGraphics/D2DWrapper.h>
+#import <Starboard.h>
+
+static const wchar_t* TAG = L"_D2DWrapper";
```
ms-jihua: I guess you can leave it in if you plan to use it eventually :/ (but I'd prefer c_logTag in that case)

---
In reply to: [82025722](https://github.com/Microsoft/WinObjC/pull/1085#discussion_r82025722) [](ancestors = 82025722)

DHowett-MSFT:
:clock1:

Frameworks/CoreGraphics/D2DWrapper.mm
MSFTFox: Based on the reading it seems like a singlethreaded factory really isn't needed.

---
In reply to: [82024085](https://github.com/Microsoft/WinObjC/pull/1085#discussion_r82024085) [](ancestors = 82024085,82023742,82000836)

Frameworks/CoreGraphics/D2DWrapper.mm
```diff
@@ -0,0 +1,33 @@
...
+
+// Helper for creating a D2DFactory
+static ComPtr<ID2D1Factory> _CreateD2DFactoryInstance(D2D1_FACTORY_TYPE type) {
+    ComPtr<ID2D1Factory> d2dFactory;
+    THROW_IF_FAILED(D2D1CreateFactory(type, d2dFactory.ReleaseAndGetAddressOf()));
```
aballway: nit: why not just &d2dFactory?

Frameworks/include/CoreGraphics/D2DWrapper.h
aballway: Is this using anything from std::?

Frameworks/include/CoreGraphics/D2DWrapper.h
```diff
@@ -0,0 +1,28 @@
...
+#import <D2d1.h>
+#include <COMIncludes_End.h>
+
+using namespace std;
+using namespace Microsoft::WRL;
+
+static ComPtr<ID2D1Factory> _CreateD2DFactoryInstance(D2D1_FACTORY_TYPE type);
+static ComPtr<ID2D1Factory> _GetD2DFactoryInstance();
```
rajsesh-msft: >static [](start = 0, length = 6)

static here doesn't make much sense.  use static on methods if they are only supposed to be used in one file, in this case, it will be used across different files, so this doesnt apply.

Frameworks/CoreGraphics/D2DWrapper.mm
MSFTFox: D2D1CreateFactory doesn't work well with a ComPtr since it's not typed correctly. There are two ways to work with this then, tell D2D1 the actual type of what's being passed in with __uuidof(ID2D1Factory) or to pass something with the proper type. It might be better to just use the uuid here though for clarity.

---
In reply to: [82041278](https://github.com/Microsoft/WinObjC/pull/1085#discussion_r82041278) [](ancestors = 82041278)

Frameworks/CoreGraphics/D2DWrapper.mm
```diff
@@ -0,0 +1,33 @@
...
+// Private helper for creating a D2DFactory
+static ComPtr<ID2D1Factory> __CreateD2DFactoryInstance() {
+    ComPtr<ID2D1Factory> d2dFactory;
+    THROW_IF_FAILED(D2D1CreateFactory(D2D1_FACTORY_TYPE_MULTI_THREADED, __uuidof(ID2D1Factory), &d2dFactory));
+    return d2dFactory;
+}
+
+// Helper for accessing a single static multithreaded D2DFactory instance
+ComPtr<ID2D1Factory> _GetD2DFactoryInstance() {
+    static ComPtr<ID2D1Factory> s_d2dFactory = __CreateD2DFactoryInstance();
```
aballway: nit: could be dispatch_once or lambda instead of having a helper method

build/CoreGraphics/lib/CoreGraphicsLib.vcxproj
ms-jihua: shouldn't it be in Frameworks\CoreGraphics if it's a private header, then?

---
In reply to: [82040170](https://github.com/Microsoft/WinObjC/pull/1085#discussion_r82040170) [](ancestors = 82040170,82025651)

ms-jihua:
:shipit:

Frameworks/CoreGraphics/D2DWrapper.mm
```diff
@@ -0,0 +1,33 @@
...
+
+#import <CoreGraphics/D2DWrapper.h>
+#import <Starboard.h>
+
+using namespace Microsoft::WRL;
+
+// Private helper for creating a D2DFactory
+static ComPtr<ID2D1Factory> __CreateD2DFactoryInstance() {
+    ComPtr<ID2D1Factory> d2dFactory;
+    THROW_IF_FAILED(D2D1CreateFactory(D2D1_FACTORY_TYPE_MULTI_THREADED, __uuidof(ID2D1Factory), &d2dFactory));
```
rajsesh-msft: This should be D2D1_FACTORY_TYPE_SINGLE_THREADED.  We dont need any locking from D2D as all our drawing is on UI thread.  Specifying multi threaded here is counter intuitive and will result in worse performance.  Note: even if the factory is single threaded, when we have hardware offloading, D2D will parallelize the tasks on the GPU.  https://msdn.microsoft.com/en-us/library/windows/desktop/dd368104(v=vs.85).aspx has all the details.

rajsesh-msft:
:clock1:

Frameworks/CoreGraphics/D2DWrapper.mm
DHowett-MSFT: My concern here is that a CGPath, which can be manipulated from multiple threads, will be backed by a thread-bound ID2D1Geometry.

rajsesh-msft: Any code that needs the multi-threaded access should lock what it needs.  making all D2D methods thread safe is not the way to do that.  Note that singlethreaded factory doesnt mean the objects cant be used in multiple threads - it just means that the caller has to ensure thread safety.  

---
In reply to: [82073552](https://github.com/Microsoft/WinObjC/pull/1085#discussion_r82073552) [](ancestors = 82073552)

build/CoreGraphics/lib/CoreGraphicsLib.vcxproj
rajsesh-msft: +1, this should be a private header.  Include\CoreGraphics should only have headers that are public (and in reference platform).  

---
In reply to: [82067697](https://github.com/Microsoft/WinObjC/pull/1085#discussion_r82067697) [](ancestors = 82067697,82040170,82025651)

MSFTFox: I thought I had fixed this in the previous iterations. Something must have happened. Fixed now though.

---
In reply to: [82096404](https://github.com/Microsoft/WinObjC/pull/1085#discussion_r82096404) [](ancestors = 82096404,82067697,82040170,82025651)

PR #1093, opened 2016-10-05
Make protocols for which we have backing implementation public.

mukhole:
For protocols like IPropertyValue we have the underlying implementation, but these were private.
Getting value out of a runtime class (returned from an API) which implements this interface was difficult as the interface itself was private.

This change makes all the protocols for which we have backing implementation public.


DHowett-MSFT:
:shipit:

asimonov-msft:
:shipit:

mukhole:
Update projection binaries: Make protocols with backing implementation publicly available.

PR #1097, opened 2016-10-06
Fix bug where multiple runs in a line were drawn incorrectly

aballway:
When multiple runs were present in a line we did not correctly adjust the x offset and were drawing them on top of eachother


rajsesh-msft:
:shipit:

Frameworks/CoreText/CTLine.mm
```diff
@@ -232,9 +232,14 @@ void _CTLineDraw(CTLineRef lineRef, CGContextRef ctx, bool adjustTextPosition) {
         CGContextSetTextPosition(ctx, curTextPos.x + line->_relativeXOffset, curTextPos.y + line->_relativeYOffset);
     }
 
-    for (_CTRun* curRun in static_cast<id<NSFastEnumeration>>(line->_runs)) {
-        CFRange range = { 0 };
-        _CTRunDraw(static_cast<CTRunRef>(curRun), ctx, range, false);
+    for (size_t i = 0; i < [line->_runs count]; ++i) {
+        _CTRun* curRun = [line->_runs objectAtIndex:i];
+        if (i > 0) {
```
ms-jihua: >if (i > 0) { [](start = 8, length = 12)

just start i at 1 instead, then?

ms-jihua: add a comment here explaining why this is necessary?

ms-jihua:
:shipit:

MSFTFox:
:shipit:

PR #1100, opened 2016-10-06
Bbowman/file activation

bbowman:
Adds the ability to activate an islandwood app via a file.

### DIFF
--- a/Frameworks/UIKit/StarboardXaml/StarboardXaml.cpp
+++ b/Frameworks/UIKit/StarboardXaml/StarboardXaml.cpp
@@ -165,6 +165,11 @@ void App::OnActivated(IActivatedEventArgs^ args) {
     UIApplicationActivated(args);
 }
 
+void App::OnFileActivated(FileActivatedEventArgs^ args)
+{
+    UIApplicationActivated(args);
+}
+
 #ifdef ENABLE_BACKGROUND_TASK
 void App::OnBackgroundActivated(BackgroundActivatedEventArgs^ args) {
     __super::OnBackgroundActivated(args);
@@ -229,6 +234,15 @@ void UIApplicationActivated(IActivatedEventArgs^ args) {
         }
 
         UIApplicationMainHandleProtocolEvent(reinterpret_cast<IInspectable*>(argUri), caller);
+    } else if (args->Kind == ActivationKind::File) {
+        FileActivatedEventArgs^ result = safe_cast<FileActivatedEventArgs^>(args);
+        TraceVerbose(TAG, L"Received file activation");
+
+        if (initiateAppLaunch) {
+            _ApplicationLaunch(ActivationTypeFile, result);
+        }
+
+        UIApplicationMainHandleFileEvent(reinterpret_cast<IInspectable*>(result));
     } else {
         TraceWarning(TAG, L"Received unhandled activation kind - %d", args->Kind);
 
--- a/Frameworks/include/MockClass.h
+++ b/Frameworks/include/MockClass.h
@@ -529,12 +529,12 @@ namespace Mock {
 template <typename TFunctor>
 void ValidateCallback(const TFunctor& functor, const wchar_t* name) {
     WEX::TestExecution::SetVerifyOutput verifySettings(WEX::TestExecution::VerifyOutputSettings::LogOnlyFailures);
-    VERIFY_IS_NOT_NULL(functor, WEX::Common::String().Format(L"The callback for %s has not been set!", name));
+    VERIFY_IS_NOT_NULL(functor, WEX::Common::String().Format(L"The callback for %ls has not been set!", name));
 }
-#elif defined(ASSERT_NE_MSG)
+#elif defined(ASSERT_TRUE_MSG)
 template <typename TFunctor>
 void ValidateCallback(const TFunctor& functor, const wchar_t* name) {
-    ASSERT_NE_MSG(functor, nullptr, "The callback for %s has not been set!", name);
+    ASSERT_TRUE_MSG(functor, "The callback for %S has not been set.", name);
 }
 #else
 #error "Unknown Test Framework!"
@@ -645,11 +645,11 @@ void ValidateCallback(const TFunctor& functor, const wchar_t* name) {
 // Internally used by MOCK_* macros; do not call directly.
 /////////////////////////////////////////////////////////////
 #define BUILD_MOCK_METHOD_IMPL(METHOD_NAME, PARAMETER_COUNT, IS_CONST, ...)                                                               \
-    \
+                                                                                                                                          \
 private:                                                                                                                                  \
     std::function<WEX::Common::ParameterTypes<decltype(&MOCK_CLASS_NAME::METHOD_NAME)>::ReturnType(                                       \
         BUILD_VARIABLE_NAME(MOCK_NAMED_ARGS_, PARAMETER_COUNT)(METHOD_NAME))> BUILD_VARIABLE_NAME(m_, METHOD_NAME);                       \
-    \
+                                                                                                                                          \
 public:                                                                                                                                   \
     virtual WEX::Common::ParameterTypes<decltype(&MOCK_CLASS_NAME::METHOD_NAME)>::ReturnType __VA_ARGS__ METHOD_NAME(                     \
         BUILD_VARIABLE_NAME(MOCK_NAMED_ARGS_, PARAMETER_COUNT)(METHOD_NAME)) BUILD_VARIABLE_NAME(MEMBER_PARAMETER_TYPES_CONST_, IS_CONST) \
@@ -770,7 +770,7 @@ struct MockInterfaceCalled {
     typedef WEX::Common::ParameterTypes<decltype(&INTERFACE_NAME::METHOD_NAME)> BUILD_VARIABLE_NAME(METHOD_NAME, PARAMETER_TYPES);  \
     static_assert(PARAMETER_COUNT == BUILD_VARIABLE_NAME(METHOD_NAME, PARAMETER_TYPES)::ParameterCount,                             \
                   "Incorrect macro called for function; parameter count doesn't match!");                                           \
-    \
+                                                                                                                                    \
 private:                                                                                                                            \
     std::function<BUILD_VARIABLE_NAME(METHOD_NAME, PARAMETER_TYPES)::ReturnType(                                                    \
         BUILD_VARIABLE_NAME(MOCK_NAMED_ARGS_, PARAMETER_COUNT)(INTERFACE_NAME::METHOD_NAME))> BUILD_VARIABLE_NAME(m_, METHOD_NAME); \
@@ -804,7 +804,7 @@ private:                                                                        
         BUILD_VARIABLE_NAME(BUILD_VARIABLE_NAME(PROPERTY_NAME, set), PARAMETER_TYPES);                                        \
     static_assert(1 == BUILD_VARIABLE_NAME(BUILD_VARIABLE_NAME(PROPERTY_NAME, set), PARAMETER_TYPES)::ParameterCount,         \
                   "Incorrect macro called for function; parameter count doesn't match!");                                     \
-    \
+                                                                                                                              \
 private:                                                                                                                      \
     std::function<BUILD_VARIABLE_NAME(BUILD_VARIABLE_NAME(PROPERTY_NAME, get), PARAMETER_TYPES)::ReturnType()>                \
         BUILD_VARIABLE_NAME(m_, BUILD_VARIABLE_NAME(PROPERTY_NAME, get));                                                     \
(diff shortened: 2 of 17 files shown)


tests/functionaltests/Tests/CortanaTests.mm
```diff
@@ -125,79 +125,79 @@
...
 
-- (BOOL)application:(UIApplication*)application didReceiveProtocol:(WFUri*)uri {
+- (void)application:(UIApplication*)application didReceiveProtocol:(WFUri*)uri {
     // Delegate method should only be called once
     EXPECT_EQ([[self methodsCalled] objectForKey:NSStringFromSelector(_cmd)], nil);
     EXPECT_OBJCEQ(@"CORTANA_TEST", uri.toString);
-    _methodsCalled[NSStringFromSelector(_cmd)] = @(YES);
-    return true;
+    [_methodsCalled setObject:@(YES) forKey:NSStringFromSelector(_cmd)];
+    return;
```
aballway: nit: don't need return

tests/functionaltests/Tests/CortanaTests.mm
```diff
@@ -125,79 +125,79 @@
...
+    EXPECT_NE(nullptr, launchOptions[UIApplicationLaunchOptionsProtocolKey]);
     WFUri* uri = launchOptions[UIApplicationLaunchOptionsProtocolKey];
     EXPECT_OBJCEQ(@"CORTANA_TEST", uri.toString);
-    _methodsCalled[NSStringFromSelector(_cmd)] = @(YES);
+    [_methodsCalled setObject:@(YES) forKey:NSStringFromSelector(_cmd)];
     return true;
 }
 
-- (BOOL)application:(UIApplication*)application didReceiveProtocol:(WFUri*)uri {
+- (void)application:(UIApplication*)application didReceiveProtocol:(WFUri*)uri {
```
aballway: interesting that this still worked before, thanks

tests/functionaltests/Tests/FileActivationTests.mm
```diff
@@ -0,0 +1,194 @@
+//******************************************************************************
+//
+// Copyright (c) 2016 Microsoft Corporation. All rights reserved.
```
aballway: nit: update copyright

tests/functionaltests/Tests/FileActivationTests.mm
```diff
@@ -0,0 +1,194 @@
...
+           });
+
+// Delegate for testing File foreground activation
+@interface FileActivationForegroundTestDelegate : NSObject <UIApplicationDelegate>
+@property (nonatomic, readonly) StrongId<NSMutableDictionary> methodsCalled;
+@end
+
+@implementation FileActivationForegroundTestDelegate
+- (id)init {
+    self = [super init];
```
aballway: extra nit: put in if()

tests/functionaltests/Tests/CortanaTests.mm
```diff
@@ -125,79 +125,79 @@
 // Delegate for testing Cortana foreground activation
 // Use text value to guarantee it is the same argument we create
 @interface CortanaVoiceCommandForegroundTestDelegate : NSObject <UIApplicationDelegate>
-@property (nonatomic, readonly) NSMutableDictionary* methodsCalled;
+@property (nonatomic, readonly) StrongId<NSMutableDictionary> methodsCalled;
```
DHowett-MSFT: 😒 this does not seem ideal

Frameworks/include/MockClass.h
```diff
@@ -529,12 +529,12 @@ namespace Mock {
 template <typename TFunctor>
 void ValidateCallback(const TFunctor& functor, const wchar_t* name) {
     WEX::TestExecution::SetVerifyOutput verifySettings(WEX::TestExecution::VerifyOutputSettings::LogOnlyFailures);
-    VERIFY_IS_NOT_NULL(functor, WEX::Common::String().Format(L"The callback for %s has not been set!", name));
+    VERIFY_IS_NOT_NULL(functor, WEX::Common::String().Format(L"The callback for %ls has not been set!", name));
 }
-#elif defined(ASSERT_NE_MSG)
+#elif defined(ASSERT_TRUE_MSG)
```
bbowman: no

---
In reply to: [82293757](https://github.com/Microsoft/WinObjC/pull/1100#discussion_r82293757) [](ancestors = 82293757)

bbowman: this is the easiest way to prefer gtest if present but not break back compat

---
In reply to: [82296551](https://github.com/Microsoft/WinObjC/pull/1100#discussion_r82296551) [](ancestors = 82296551,82293757)

Frameworks/include/UIApplicationInternal.h
```diff
@@ -16,6 +16,9 @@
 #pragma once
 
 #import <UIKit/UIApplication.h>
+#import <UWP/WindowsFoundation.h>
```
bbowman: well you use WMSSpeechRecognitionResult so you were getting lucky with the include ordering before .... 

---
In reply to: [82294162](https://github.com/Microsoft/WinObjC/pull/1100#discussion_r82294162) [](ancestors = 82294162)

tests/functionaltests/Tests/CortanaTests.mm
bbowman: sweet emoji. What part seems non ideal? Having a strongID prop seems fine to me. I don't particularly like the methodsCalled doodad but it works and follows AJs pattern

---
In reply to: [82296469](https://github.com/Microsoft/WinObjC/pull/1100#discussion_r82296469) [](ancestors = 82296469)

tests/functionaltests/Tests/FileActivationTests.mm
bbowman: I'll update yours too I guess

---
In reply to: [82295491](https://github.com/Microsoft/WinObjC/pull/1100#discussion_r82295491) [](ancestors = 82295491)

Frameworks/include/UIApplicationInternal.h
aballway: IIRC not including them in a header was intentional, something funky happened with the linker because of how our modules take ownership of headers or something to that effect.  I'd prefer sticking to forward declarations until that's resolved so someone doesn't accidentally stumble into that wasp nest

tests/functionaltests/Tests/FileActivationTests.mm
aballway: Probably <3

tests/functionaltests/Tests/CortanaTests.mm
rajsesh-msft: But why? I am not seeing what this is actually addressing.  

---
In reply to: [82297024](https://github.com/Microsoft/WinObjC/pull/1100#discussion_r82297024) [](ancestors = 82297024,82296469)

bbowman: Memory leaks from _methodsCalled = [NSMutableDictionary new]; in the old implementation. There was no corresponding release in the dtor. The options were 1. add a dtor 2. use strong id 3. turn on arc. I chose 2 but @DHowett-MSFT  seems less enthused with that choice.

---
In reply to: [82302619](https://github.com/Microsoft/WinObjC/pull/1100#discussion_r82302619) [](ancestors = 82302619,82297024,82296469)

Frameworks/include/UIApplicationInternal.h
bbowman: I prefer doing the right thing and fixing issues preventing us from doing the right thing rather than leaving scary things hidden for someone else to deal with. If there was "something funky" lets address that rather than running from it. Additionally this is an Internal header and thus shouldn't be subjected to modules. 

---
In reply to: [82297724](https://github.com/Microsoft/WinObjC/pull/1100#discussion_r82297724) [](ancestors = 82297724)

include/UIKit/UIApplicationDelegate.h
```diff
@@ -38,10 +38,12 @@
  * App developers should import the following files to use methods requiring projected classes
  * #import <UWP/WindowsFoundation.h>
  * #import <UWP/WindowsMediaSpeechRecognition.h>
+ * #import <UWP/WindowsApplicationModelActivation.h>
 */
```
bbowman: this is probably what you were thinking of @aballway and this has an todo so we are at least tracking the work here. It likely is due to circular deps between Projections and Foundation/Other frameworks creating the issue which I'd love us to solve at some point.

include/UIKit/UIApplicationDelegate.h
```diff
@@ -129,6 +131,7 @@ UIKIT_EXPORT NSString* const UIApplicationLaunchOptionsToastActionUserInputKey;
 - (void)application:(UIApplication*)application didReceiveRemoteNotification:(NSDictionary*)userInfo;
 - (void)application:(UIApplication*)application didReceiveToastAction:(NSDictionary*)action;
 - (void)application:(UIApplication*)application didReceiveVoiceCommand:(WMSSpeechRecognitionResult*)result;
+- (void)application:(UIApplication*)application didReceiveFile:(WAAFileActivatedEventArgs*)result;
```
jaredhms: I think I'd much prefer using the existing iOS API surface for this (appdidfinishlaunchingwithoptions); why force the app devs to change their code?

tests/functionaltests/Tests/CortanaTests.mm
rajsesh-msft: Since this is test code, I dont have a strong preference.  In framework code, we have been doing 1 consistently, we should probably do the same here.  (adding -dealloc is quite simple and less intrusive).

---
In reply to: [82303181](https://github.com/Microsoft/WinObjC/pull/1100#discussion_r82303181) [](ancestors = 82303181,82302619,82297024,82296469)

rajsesh-msft:
:shipit:

tests/frameworks/gtest/include/gtest/gtest.h
```diff
@@ -295,7 +296,7 @@ class GTEST_API_ AssertionResult {
   // assertion's expectation). When nothing has been streamed into the
   // object, returns an empty string.
   const char* message() const {
-    return message_.get() != NULL ?  message_->c_str() : "";
+    return success_ ? "" : (message_.get() != NULL ?  message_->c_str() : "");
```
ms-jihua: nit: this is kind of gross

I guess you could change it to (!success_ && message_.get()) ? message_->c_str() : "";   ? does that even look better?

ms-jihua:
:shipit:

Frameworks/UIKit/StarboardXaml/StarboardXaml.cpp
```diff
@@ -165,6 +165,11 @@ void App::OnActivated(IActivatedEventArgs^ args) {
     UIApplicationActivated(args);
 }
 
+void App::OnFileActivated(FileActivatedEventArgs^ args)
```
mnithish: You need to also edit  msvc\vsimporter-templates\WinStore10-App\App.xaml.cpp and msvc\vsimporter-templates\WinStore10-App\App.xaml.h file

mnithish:
:clock1:

Frameworks/UIKit/StarboardXaml/StarboardXaml.cpp
mnithish: And to provide more context, we updated vsimporter to use App.Xaml.cpp few iterations back so StarboardXaml is not used any more if someone tries to vsimport their iOS project. We still have to maintain both these code paths so we do not have to force existing developer from having to re-import their project

---
In reply to: [82463737](https://github.com/Microsoft/WinObjC/pull/1100#discussion_r82463737) [](ancestors = 82463737)

include/UIKit/UIApplicationDelegate.h
bbowman: it seems they are different patterns with the ability for multiple files and Verbs (open vs post vs edit, etc) and the ability to be called while active as Nithish points out. In light of that, I think the pattern here should be used rather than reducing fidelity / trying to cram extra parameters into an options dictionary that probably wouldn't be checked and be incongruous with the other delegate call.

---
In reply to: [82463028](https://github.com/Microsoft/WinObjC/pull/1100#discussion_r82463028) [](ancestors = 82463028,82305070)

msvc/vsimporter-templates/WinStore10-App/App.xaml.h
```diff
@@ -13,16 +13,17 @@ namespace $safeprojectname$ {
...
 protected:
-    virtual void OnLaunched(Windows::ApplicationModel::Activation::LaunchActivatedEventArgs^ e) override;
-    virtual void OnActivated(Windows::ApplicationModel::Activation::IActivatedEventArgs^ e) override;
+    virtual void OnLaunched(Windows::ApplicationModel::Activation::LaunchActivatedEventArgs ^ e) override;
+    virtual void OnActivated(Windows::ApplicationModel::Activation::IActivatedEventArgs ^ e) override;
 #ifdef ENABLE_BACKGROUND_TASK
-    virtual void OnBackgroundActivated(Windows::ApplicationModel::Activation::BackgroundActivatedEventArgs^ e) override;
+    virtual void OnBackgroundActivated(Windows::ApplicationModel::Activation::BackgroundActivatedEventArgs ^ e) override;
 #endif
+    void OnFileActivated(Windows::ApplicationModel::Activation::FileActivatedEventArgs ^ args) override;
```
aballway: nit: clang format doesn't like c++/cx

msvc/vsimporter-templates/WinStore10-App/App.xaml.cpp
```diff
@@ -52,6 +54,11 @@ void App::OnActivated(Windows::ApplicationModel::Activation::IActivatedEventArgs
     UIApplicationActivated(e);
 }
 
+void App::OnFileActivated(FileActivatedEventArgs^ args)
+{
```
aballway: you need to call main here to go through UIApplicationMain

msvc/vsimporter-templates/WinStore10-App/App.xaml.h
bbowman: oops. I fixed it with the .cpp but not .h


---
In reply to: [82850185](https://github.com/Microsoft/WinObjC/pull/1100#discussion_r82850185) [](ancestors = 82850185)

PR #1102, opened 2016-10-07
UI Automation - WinAppDriver test framework exercising basic UITextField tests within XAMLCatalog

oliversa-msft:
WinAppDriver is the Microsoft-authored OSS version of the Appium driver for UI automation (github.com/Microsoft/WinAppDriver). This is the UI automation test framework that will eventually replace all UI automation within MS.

You'll need to install the WinAppDriver.MSI local and run the EXE since the host service is still under development. The UI automation tests in this pull request will target XAMLCatalog so it needs to be deployed on your local machine (127.0.0.1)

The UITextField tests are pretty minimal tests (find element, send keys or change content on element, verify content is different). We will expand on these tests but I'm sharing them early so others can UX tests can follow this pattern. 

We are working with the WAD , Accessilibility and MITA teams to extract more XAML properties through this framework in the near future. This will enable us to verify the underlying XAML controls have the correct properties and states as set through UIKit controls.

CHANGES
=======
1. New XAMLCatalogTest DLL created target the XAMLCatalog app using WinAppDriver framework
2. Migrated UITextField test entries from WOCCatalog to XAMLCatalog
3. Enabled UIView's accessibilityIdentifier/setAccessibilityIdentifier property so automation IDs can be used

TESTS
====
1. Validated XAMLCatalog/WOCCatalog build and run under simulators on PC/OSX
2. XAMLCatalogTest DLL can run and target a local WinAppDriver session provided XAMLCatalog is installed locally as well.


rajsesh-msft:
You win the 🏆 for this.

Frameworks/UIKit/UIView.mm
```diff
@@ -3561,6 +3563,26 @@ - (void)setXamlElement:(WXFrameworkElement*)xamlElement {
     [self layer].contentsElement = xamlElement;
 }
 
+// Retrieve the backing XAML element's Automation Id
+- (NSString*)accessibilityIdentifier {
+    WXFrameworkElement* layerContentElement = [self layer].contentsElement;
+    WXFrameworkElement* xamlElement = layerContentElement ? layerContentElement : priv->_xamlInputElement.get();
+    if (xamlElement) {
```
amit6257: not sure but you might not need this if() condition if [WUXAAutomationProperties getAutomationId] returns nil for a  nil argument.

PR #1104, opened 2016-10-07
Convert sbAssert and sbValidate into sbAssertWithTelemetry and sbValidateWithTelemetry

asimonov-msft:


tools/vsimporter/src/query.cpp
```diff
@@ -76,7 +76,7 @@ static bool parseListSelection(const String& selection, size_t upperBound, std::
 
 void queryListSelection(const StringVec& values, const String& queryMessage, const String& itemName, std::vector<size_t>& ret)
 {
-  sbAssert(!values.empty());
+  sbAssertWithTelemetry(!values.empty(), "Unable to query empty list of values");
```
kellymor: >sbAssertWithTelemetry(!values.empty(), "Unable to query empty list of values"); [](start = 2, length = 79)

Would the query message be helpful here?  I'm not sure if the query would contain Personally Identifiable Information (PII) or not.

tools/vsimporter/src/utils/xc2vs.cpp
```diff
@@ -95,7 +95,7 @@ static VCProjectItem* addFileToVSInternal(const PBXFile* file, VCProject& proj,
   // Take into account path override
   bool useRelativePath = true;
   if (itemHint && !itemHint->pathOverride.empty()) {
-    sbAssert(!isVariant, "Unexpected path override for variant file: " + realPath);
+    sbAssertWithTelemetry(!isVariant, "Unexpected path override for variant file: " + realPath);
```
kellymor: >realPath [](start = 86, length = 8)

Could this path contain PII, like a user name, e.g. /mymac/kellymor/projects/ ?  Or is this handled by the call to sbAssertWithTelemetry?  If the latter, then disregard this comment.

asimonov-msft: The path does contain PI, but it will be stripped off since it comes after the colon.

---
In reply to: [82424365](https://github.com/Microsoft/WinObjC/pull/1104#discussion_r82424365) [](ancestors = 82424365)

tools/vsimporter/src/query.cpp
asimonov-msft: I moved the assert to the calling functions.

---
In reply to: [82420134](https://github.com/Microsoft/WinObjC/pull/1104#discussion_r82420134) [](ancestors = 82420134)

kellymor:
:shipit:

Issue #1105, opened 2016-10-07
UINibUnarchiver should not use alloc without init, nor assume all classes have init


ms-jihua:
UINibUnarchiver.mm
pObj->cachedId = [classId alloc];

alloc without init is considered invalid, and the way constructObject() is currently written may leave pObj->cachedId as an invalid object (this would be the case with any of our bridged classes, where alloc alone would return just our prototype object)

Additionally, not all classes respond to [[class alloc] init]. (eg: [[UIFont alloc] init] crashes on the reference platform, and our new implementation on the CT-DW branch also does not respect alloc->init)

I ran into this issue with UIFont specifically, so I left a quick special-case there for it on CT-DW to unblock myself, but someone should take a look at this and make a proper fix for it.

PR #1106, opened 2016-10-07
Quick fixes to unblock internal test apps

ms-jihua:
- Reduces crashes in internal test apps on CT-DW to known issues

### DIFF
--- a/Frameworks/UIKit/UICTFont.h
+++ b/Frameworks/UIKit/UICTFont.h
@@ -17,5 +17,8 @@
 
 #import <UIKit/UIFont.h>
 
+@interface UIFontPrototype : UIFont
+@end
+
 @interface UICTFont : UIFont
 @end
--- a/Frameworks/UIKit/UICTFont.mm
+++ b/Frameworks/UIKit/UICTFont.mm
@@ -18,6 +18,16 @@
 #import <CFFoundationInternal.h>
 #import <BridgeHelpers.h>
 
+// Only exists to support the WinObjC-only [UIFont initWithCoder:]
+#pragma region UIFontPrototype
+@implementation UIFontPrototype
+
+PROTOTYPE_CLASS_REQUIRED_IMPLS(UICTFont)
+
+@end
+#pragma endregion
+
+#pragma region UICTFont
 @implementation UICTFont
 
 BRIDGED_CLASS_REQUIRED_IMPLS(CTFontRef, CTFontGetTypeID, UIFont, UICTFont)
@@ -83,3 +93,4 @@ BRIDGED_CLASS_REQUIRED_IMPLS(CTFontRef, CTFontGetTypeID, UIFont, UICTFont)
 }
 
 @end
+#pragma endregion
--- a/Frameworks/UIKit/UIFont.mm
+++ b/Frameworks/UIKit/UIFont.mm
@@ -45,7 +45,7 @@ using namespace Microsoft::WRL;
 
 @implementation UIFont
 
-BASE_CLASS_REQUIRED_IMPLS(UIFont, UICTFont, CTFontGetTypeID);
+BASE_CLASS_REQUIRED_IMPLS(UIFont, UIFontPrototype, CTFontGetTypeID);
 
 /**
  @Status Interoperable
@@ -285,7 +285,7 @@ BASE_CLASS_REQUIRED_IMPLS(UIFont, UICTFont, CTFontGetTypeID);
 }
 
 /**
-@Status Interoperable
+ @Status Interoperable
 */
 - (instancetype)copyWithZone:(NSZone*)zone {
     return [self retain];
@@ -294,6 +294,27 @@ BASE_CLASS_REQUIRED_IMPLS(UIFont, UICTFont, CTFontGetTypeID);
 - (uint32_t)_sizingFontHandle {
     UNIMPLEMENTED();
     return StubReturn();
+}
+
+// WinObjC-only extension for UINibUnarchiver
+- (instancetype)initWithCoder:(NSCoder*)coder {
+    NSString* name = [coder decodeObjectForKey:@"UIFontName"];
+    if ([name length] < 1) {
+        // fallback to default if could not find a font name
+        name = (__bridge NSString*)kCTFontDefaultFontName;
+    }
+
+    CGFloat size = [coder decodeFloatForKey:@"UIFontPointSize"];
+
+    UIFont* font = [UIFont fontWithName:name size:size];
+
+    [self release];
+    if (font) {
+        self = font;
+        return font;
+    } else {
+        return nil;
+    }
 }
 
 @end


Frameworks/UIKit/UINibUnarchiver.mm
```diff
@@ -203,7 +203,13 @@ static id constructObject(UINibUnarchiver* self, Object* pObj) {
         id classId = pObj->classType; // objc_getClass(pObj->className);
         assert(classId != nil);
 
-        pObj->cachedId = [classId alloc];
+        // TODO: #1105 alloc without init is invalid, and not all classes support alloc-init anyway
```
bbowman: >and not all classes support alloc-init anyway [](start = 54, length = 45)

all nsobject deriving ones should?

bbowman:
:clock1:

bbowman:

Frameworks/UIKit/UINibUnarchiver.mm
bbowman: >// TODO: #1105 [](start = 8, length = 14)

@davelamb  should probably take a look here ...

Frameworks/UIKit/UINibUnarchiver.mm
```diff
@@ -203,7 +203,13 @@ static id constructObject(UINibUnarchiver* self, Object* pObj) {
...
+        // look for a proper fix to this issue, rather than just special-casing for UIFont
```
rajsesh-msft: >proper [](start = 22, length = 6)

Not sure if this is an issue, we already have several special cases in this method.

Frameworks/UIKit/UIRuntimeOutletCollectionConnection.mm
```diff
@@ -30,16 +30,18 @@ - (instancetype)initWithCoder:(NSCoder*)coder {
 
 - (void)_makeConnection {
     //  Grab existing collection from the target
-    NSMutableArray* collection = [source valueForKey:label];
+    if (source) {
```
rajsesh-msft: This should not be required, you can send messages to nil object.

rajsesh-msft:
:clock1:

Frameworks/UIKit/UINibUnarchiver.mm
```diff
@@ -203,7 +203,13 @@ static id constructObject(UINibUnarchiver* self, Object* pObj) {
...
+        if ([classId isEqual:[UIFont class]]) {
+            pObj->cachedId = [[UIFont defaultFont] retain];
```
ms-jihua: we weren't before, since UIFont doesn't support NSCoding, and I wanted to maintain the existing functionality. I'll wait for dave to comment.

---
In reply to: [82631100](https://github.com/Microsoft/WinObjC/pull/1106#discussion_r82631100) [](ancestors = 82631100)

Frameworks/UIKit/UINibUnarchiver.mm
davelamb: Can't we replace -alloc with NSAllocateObject?

ms-jihua: Seems... undesirable?
[https://developer.apple.com/reference/foundation/1587930-nsallocateobject](https://developer.apple.com/reference/foundation/1587930-nsallocateobject)
"This function is deprecated and unavailable for use with ARC."

---
In reply to: [82673565](https://github.com/Microsoft/WinObjC/pull/1106#discussion_r82673565) [](ancestors = 82673565)

Frameworks/UIKit/UINibUnarchiver.mm
davelamb: Hm. After a little investigation, it looks like our Xib2Nib and UIFont implementation are a little out of spec. NSFont does indeed implement NSCoding, but it does seem like UIFont doesn't, and so my only conclusion is we're only ever supposed to use UIFontDescriptor to encode/decode on iOS. IE, in `UILabel -initWithCoder`, `_font` should be populated with `[UIFont fontWithDescriptor:[_coder decodeObjectForKey:@"UIFont[Descriptor]" pointSize:0.0f]` and of course we'd have to update Xib2Nib to output the right thing. 

Since we're no longer investing in Xib2Nib, I'd say for this is a good enough solution, so long as the new UIFont does still implements initWithCoder.

Frameworks/UIKit/UINibUnarchiver.mm
davelamb: Well it's just a wrapper for the runtime call to class_createInstance, so deprecated it may be, its behavior is very well defined. We use it elsewhere for this specific case: we want to allocate and zero out a class, but not call `+ alloc`. After a little more digging, we definitely **should not** call NSAllocateObject, specifically because we break contract in Islandwood by doing some initialization inside `+ alloc` in a bunch of classes in UIKit.

I'm fine with the change as is, it's just a little silly to be putting one-offs (that aren't primitives that need special handling) in the unarchiver code. I'd almost rather see something like a new internal function on NSObject, allocForDecoding, that by default just calls alloc, but can return the default font like you have here for UIFont. Too much to ask for a stop-gap solution, I know. :)

Frameworks/UIKit/UINibUnarchiver.mm
ms-jihua: Hmm, if it's the case that the UIFont still has to support initWithCoder despite of its doc (was not previously aware of this, and thus current CT-DW UIFont doesn't), then I can just change up UIFont to support alloc and initWithCoder and we won't need the special case here.

---
In reply to: [82917523](https://github.com/Microsoft/WinObjC/pull/1106#discussion_r82917523) [](ancestors = 82917523)

Frameworks/UIKit/UINibUnarchiver.mm
ms-jihua: As commented below, I can remove the one-off, but I'm still concerned about leaving [classId alloc] here:

- alloc without init is considered invalid
- we have a code path here (if classId doesn't support initWithCoder) where we effectively call [[classId alloc] autorelease].
- we appear to have been DEPENDING on this alloc autorelease code path, since if I change this to
    if ([classId instancesRespondToSelector:@selector(initWithCoder:)])
        pObj->cachedId = [[classId alloc] initWithCoder...]
  then things start exploding all over the place with an infinite recursion

we can take care of this separately in #1105, but I wanted to call it out while we're here and make sure you're aware of it

---
In reply to: [82918436](https://github.com/Microsoft/WinObjC/pull/1106#discussion_r82918436) [](ancestors = 82918436)

Frameworks/UIKit/UIRuntimeOutletCollectionConnection.mm
ms-jihua: I definitely broke on this the other day but can't repro it anymore. Weird.

---
In reply to: [82630600](https://github.com/Microsoft/WinObjC/pull/1106#discussion_r82630600) [](ancestors = 82630600)

PR #1107, opened 2016-10-07
Implement [NSString sizeWith...] functions using CoreText

### LINKED ISSUE #1014: Investigate back button image width regressions in CTCatalog with DWrite move
The back button image width seems to have reduced with the latest move to DWrite. Investigate what caused this to happen only CGFont has moved over to using DWrite

ms-jihua:
These functions began to return {0,0} sizes as a result of CGFont becoming stubbed,
but were incorrect previously as they made a draw call even though they did not have responsibility for drawing.

- Fixes visual issues in CTCatalog
  - Back button was incorrect width
  - Most text labels were not correctly displaying
- Remove an unused WinObjC-only function in NSString+UIKitAdditions
- Ensure that UIKit line break mode constants align with CoreText's
- Misc fix to change some UIFont default fonts to use UIFont default sizes

Fixes #1014

### DIFF
--- a/Frameworks/UIKit/NSString+UIKitAdditions.mm
+++ b/Frameworks/UIKit/NSString+UIKitAdditions.mm
@@ -592,7 +592,7 @@ static NSDictionary* _getDefaultUITextAttributes() {
 }
 
 // Private helper that converts a UILineBreakMode -> NSParagraphStyle
-// TODO #: NS/CT ParagraphStyle are not properly bridged, and ParagraphStyle is not currently read anywhere
+// TODO #1108: NS/CT ParagraphStyle are not properly bridged, and ParagraphStyle is not currently read anywhere
 static inline NSParagraphStyle* _paragraphStyleWithLineBreakMode(UILineBreakMode lineBreakMode) {
     NSMutableParagraphStyle* ret = [NSMutableParagraphStyle new];
     ret.lineBreakMode = lineBreakMode;
--- a/Frameworks/UIKit/NSString+UIKitAdditions.mm
+++ b/Frameworks/UIKit/NSString+UIKitAdditions.mm
@@ -579,158 +579,84 @@ static NSDictionary* _getDefaultUITextAttributes() {
     return fontExtent;
 }
 
+// Private helper for sizeWith... functions
+// Returns the bounding box size this string would occupy when drawn as specified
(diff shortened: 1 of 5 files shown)


ms-jihua:
before: backbutton_ctdw

after: 
backbutton_ctdw_after

Frameworks/UIKit/NSString+UIKitAdditions.mm
```diff
@@ -579,158 +579,84 @@ - (CGSize)drawAtPoint:(CGPoint)pt
...
+static inline NSParagraphStyle* _paragraphStyleWithLineBreakMode(UILineBreakMode lineBreakMode) {
+    NSMutableParagraphStyle* ret = [NSMutableParagraphStyle new];
+    ret.lineBreakMode = lineBreakMode;
     return ret;
 }
 
 /**
- @Status Interoperable
+ @Status Caveat
+ @Notes lineBreakMode is currently not fully supported
```
rajsesh-msft: Is this a regression? If so, we need an issue to track it.

Frameworks/UIKit/NSString+UIKitAdditions.mm
```diff
@@ -579,158 +579,84 @@ - (CGSize)drawAtPoint:(CGPoint)pt
...
-    rct.origin.y = 0;
-    rct.size.width = forWidth;
-    rct.size.height = 0;
-
-    drawString(font, NULL, str, [self length], rct, lineBreak, UITextAlignmentLeft, &ret);
+              forWidth:(float)width
+         lineBreakMode:(UILineBreakMode)lineBreakMode {
     if (actualFontSize) {
-        *actualFontSize = 10.0f;
+        *actualFontSize = [font pointSize];
```
rajsesh-msft: 😮

rajsesh-msft:
:shipit:

Frameworks/UIKit/NSString+UIKitAdditions.mm
ms-jihua: I filed this as #1108, I'll add a comment to the task calling out these functions specifically.

---
In reply to: [82511650](https://github.com/Microsoft/WinObjC/pull/1107#discussion_r82511650) [](ancestors = 82511650)

PR #1117, opened 2016-10-11
Always create solution folders when importing, since they are useful for build telemetry

asimonov-msft:


tools/vsimporter/src/SBProject.cpp
```diff
@@ -300,9 +300,10 @@ void SBProject::constructVCProjects(VSSolution& sln, const StringSet& slnConfigs
   // Get the output format and directory
   String outputDir = m_buildSettings->getValue("VSIMPORTER_OUTPUT_DIR");
 
-  // Create a solution folder for the project iff we're importing multiple targets
+  // Create a solution folder for the project, as long as some Xcode target was imported.
+  // The solution folder is necessary for VS build telemetry to work properly.
   VSSolutionFolderProject* projFolder = NULL;
-  if (m_existingTargets.size() > 1) {
+  if (m_existingTargets.size() > 0) {
```
aballway: Is this check necessary since you already return if it's empty?

asimonov-msft: You are technically correct. The best kind of correct.

---
In reply to: [82833190](https://github.com/Microsoft/WinObjC/pull/1117#discussion_r82833190) [](ancestors = 82833190)

PR #1118, opened 2016-10-11
Tweak ci build script to support CI build targeting any source branch.

### LINKED ISSUE #1089: Enable *any branch* full build for github branches
We have build config that lets us do this with VSO branches, but not with github branches.

rajsesh-msft:
* Removed deprecated build definitions.
* Added support to explicitly specify source branch.
* Note: Source branch could probably be determined from upstream remote,  but it is not very reliable.  Also, explicit parameter allows kicking  off arbitrary builds.

Fix #1089

### DIFF
--- a/tools/build/Submit-CIBuilds.ps1
+++ b/tools/build/Submit-CIBuilds.ps1
@@ -11,6 +11,7 @@ Param (
     [switch]$Reconfigure,
     [switch]$SkipRemoteCheck,
     [string]$BuildFilter,
+	[string]$SourceBranch='develop',
 
     [string]$AuthUsername,
     [string]$AuthToken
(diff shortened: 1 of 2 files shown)


rajsesh-msft:

tools/build/Submit-CIBuilds.ps1
```diff
@@ -11,6 +11,7 @@ Param (
     [switch]$Reconfigure,
     [switch]$SkipRemoteCheck,
     [string]$BuildFilter,
+	[string]$SourceBranch='develop',
```
DHowett-MSFT: >	 [](start = 0, length = 1)

nit: Tab, should be four spaces

DHowett-MSFT:
:shipit:

PR #1131, opened 2016-10-12
Fixed building from source link

antoine-haas:
Updated the readme link to the building from source page of the wiki.


rajsesh-msft:
Bypassing CI build for doc change.

PR #1138, opened 2016-10-13
 Initial implementation of CGPath Using ID2D1Geometries

MSFTFox:
This work provides the foundation for creating a CGPath based on
 ID2D1PathGeometry and GeometrySink. A simple implementation for
 adding a simple line, moving the current location to a new point
 and retrieving some basic information from the Geometry sink has
 been added. Isolated unit tests have also been added to ensure
 these APIs are working as intended for these simple scenarios.


Frameworks/UIKit/NSTextContainer.mm
```diff
@@ -77,8 +77,9 @@ - (CGRect)lineFragmentRectForProposedRect:(CGRect)proposed
 
     if (_exclusionPaths != nil) {
         for (UIBezierPath* path in(NSArray*)_exclusionPaths) {
-            CGRect shapeRect = _CGPathFitRect(path.CGPath, ret, _size, self.lineFragmentPadding);
-            ret = CGRectIntersection(shapeRect, ret);
+            // TODO : Replace with public function or work around.
+            // CGRect shapeRect = _CGPathFitRect(path.CGPath, ret, _size, self.lineFragmentPadding);
```
DHowett-MSFT: Please log a github issue for this!

Frameworks/CoreText/CTFramesetter.mm
```diff
@@ -60,7 +60,9 @@ CTFramesetterRef CTFramesetterCreateWithAttributedString(CFAttributedStringRef s
 */
 CTFrameRef CTFramesetterCreateFrame(CTFramesetterRef framesetter, CFRange stringRange, CGPathRef path, CFDictionaryRef frameAttributes) {
     CGRect frameSize;
-    _CGPathGetBoundingBoxInternal(path, &frameSize);
+
+    // TODO: Replace with public functions or workaround.
```
DHowett-MSFT: >functions [](start = 33, length = 9)

Same as last time; github issue!

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -31,8 +31,92 @@
...
+
+    return true;
+}
+
+static void __CGPathInit(CFTypeRef cf) {
+    CGPathRef path = (CGPathRef)cf;
+    // ComPtr needs to be manually new'd, to avoid calling Create functions to an incomplete out-pointer
+    struct __CGPath* mutablePath = const_cast<struct __CGPath*>(path);
+    new (std::addressof(mutablePath->_pathGeometry)) ComPtr<ID2D1PathGeometry>();
+    new (std::addressof(mutablePath->_geometrySink)) ComPtr<ID2D1GeometrySink>();
```
DHowett-MSFT: Make sure you clear the points on create, otherwise Create+GetPoint will return garbage

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -31,8 +31,92 @@
...
+    ComPtr<ID2D1PathGeometry> _pathGeometry;
+    ComPtr<ID2D1GeometrySink> _geometrySink;
+
+    bool _isPathClosed;
+    CGPoint _currentPoint;
+    CGPoint _startingPoint;
+};
+
+// A private helper function for re-opening a path geometry.
+static void __preparePathForEditing(CGPathRef path) {
```
DHowett-MSFT: Might be good to explain why this is necessary

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -31,8 +31,92 @@
...
+        path->_geometrySink->Close();
+        path->_isPathClosed = true;
+    }
+}
+
+static Boolean __CGPathEqual(CFTypeRef cf1, CFTypeRef cf2) {
+    struct __CGPath* path1 = (struct __CGPath*)cf1;
+    struct __CGPath* path2 = (struct __CGPath*)cf2;
+
+    return true;
```
DHowett-MSFT: >return true; [](start = 4, length = 12)

for now, return FALSE!

DHowett-MSFT: (also, should be a CF boolean, not a C++ one)

---
In reply to: [83128291](https://github.com/Microsoft/WinObjC/pull/1138#discussion_r83128291) [](ancestors = 83128291)

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -65,126 +149,31 @@ void addPoint(float ptX, float ptY) {
...
-    pathref->_applyPath(context);
-}
-
-// Not sure why this is necessary in addition to the public version, need to dig further
-void _CGPathGetBoundingBoxInternal(CGPathRef pathref, CGRect* rectOut) {
-    pathref->_getBoundingBox(rectOut);
-}
-
 void _CGPathAddElement(
     CGPathRef path, CGPathElementType type, CGPoint p0 = CGPointZero, CGPoint p1 = CGPointZero, CGPoint p2 = CGPointZero) {
```
DHowett-MSFT: Empty function; is this for linking compatibility with external consumers?

Frameworks/CoreGraphics/CGPath.mm
DHowett-MSFT: You already did this.
Perhaps it can be done in init for clarity's sake!

---
In reply to: [83128234](https://github.com/Microsoft/WinObjC/pull/1138#discussion_r83128234) [](ancestors = 83128234)

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -65,126 +149,31 @@ void addPoint(float ptX, float ptY) {
...
 }
 
 /**
  @Status Interoperable
 */
 CGMutablePathRef CGPathCreateMutable() {
-    return __CGPath::alloc(nil);
+    size_t memSize = sizeof(struct __CGPath) - sizeof(CFRuntimeBase);
+    CGPathRef ret = ((CGPathRef)(_CFRuntimeCreateInstance(kCFAllocatorDefault, CGPathGetTypeID(), memSize, NULL)));
+    struct __CGPath* mutableRet = const_cast<struct __CGPath*>(ret);
```
DHowett-MSFT: >struct [](start = 4, length = 6)

You can drop the `struct ` in C++ files

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -200,24 +189,40 @@ CGPathRef CGPathCreateCopy(CGPathRef path) {
...
-    return ret;
+
+    __closePath(path);
+    if (FAILED(path->_pathGeometry->Stream(mutableRet->_geometrySink.Get()))) {
+        CFRelease(mutableRet);
+        return nullptr;
+    }
+
+    mutableRet->_currentPoint = { path->_currentPoint.x, path->_currentPoint.y };
+    // Always create an open copy as opening a closed copy would be more expensive.
```
DHowett-MSFT: Don't forget to copy the source point!

Also, you should be able to do `mutableRet->point = path->point` instead of going member by member

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -228,19 +233,20 @@ void CGPathAddLineToPoint(CGMutablePathRef path, const CGAffineTransform* m, flo
...
-    if (m) {
-        CGPoint pt;
-
-        pt.x = x;
-        pt.y = y;
+    __preparePathForEditing(path);
 
+    CGPoint pt;
+    pt.x = x;
+    pt.y = y;
```
DHowett-MSFT: nit: fold into `CGPoint pt{x, y};`

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -228,19 +233,20 @@ void CGPathAddLineToPoint(CGMutablePathRef path, const CGAffineTransform* m, flo
...
+    if (m) {
         pt = CGPointApplyAffineTransform(pt, *m);
-
-        x = pt.x;
-        y = pt.y;
     }
 
-    _CGPathAddElement(path, kCGPathElementAddLineToPoint, { x, y });
+    path->_geometrySink->BeginFigure(D2D1::Point2F(path->_currentPoint.x, path->_currentPoint.y), D2D1_FIGURE_BEGIN_FILLED);
+    path->_geometrySink->AddLine(D2D1::Point2F(pt.x, pt.y));
```
DHowett-MSFT: Only `BeginFigure` if necessary; this might do with some tracking.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -539,41 +503,41 @@ void CGPathAddEllipseInRect(CGMutablePathRef path, const CGAffineTransform* m, C
...
+@Notes
 */
 CGRect CGPathGetBoundingBox(CGPathRef path) {
-    CGRect ret;
+    D2D1_RECT_F bounds;
 
-    if (path == NULL) {
-        return CGRectMake(std::numeric_limits<float>::infinity(), std::numeric_limits<float>::infinity(), 0, 0);
+    __closePath(path);
+
```
DHowett-MSFT: Might be good to comment why this is necessary

Frameworks/CoreGraphics/D2DWrapper.h
```diff
@@ -21,4 +21,4 @@
 #import <D2d1.h>
 #include <COMIncludes_End.h>
 
-Microsoft::WRL::ComPtr<ID2D1Factory> _GetD2DFactoryInstance();
\ No newline at end of file
+Microsoft::WRL::ComPtr<ID2D1Factory> _CreateD2DFactoryInstance();
```
DHowett-MSFT: Did this need to pair with a change to D2DWrapper.mm?

DHowett-MSFT:
:clock1:

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -200,24 +189,40 @@ CGPathRef CGPathCreateCopy(CGPathRef path) {
...
+    }
+    if (FAILED(mutableRet->_pathGeometry->Open(&mutableRet->_geometrySink))) {
+        CFRelease(mutableRet);
+        return nullptr;
     }
-    return ret;
+
+    __closePath(path);
+    if (FAILED(path->_pathGeometry->Stream(mutableRet->_geometrySink.Get()))) {
+        CFRelease(mutableRet);
```
MSFTFox: I think that may create two paths that are both edited when a single operation occurs. I'll verify.

Frameworks/CoreGraphics/D2DWrapper.h
MSFTFox: Looks like this fixed a missed pairing between the mm.

Frameworks/CoreGraphics/CGPath.mm
rajsesh-msft: It seems to destroy the existing path (as both _pathGeometry and _geometrySink are recreated?  Also, we are leaking the old objects.

---
In reply to: [83128267](https://github.com/Microsoft/WinObjC/pull/1138#discussion_r83128267) [](ancestors = 83128267)

rajsesh-msft: >__preparePathForEditing [](start = 12, length = 23)

Nit: these could be member functions of __CGPath.  A struct is a class afterall.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -31,8 +31,92 @@
...
+}
+
+static void __closePath(CGPathRef path) {
+    if (!path->_isPathClosed) {
+        path->_geometrySink->Close();
+        path->_isPathClosed = true;
+    }
+}
+
+static Boolean __CGPathEqual(CFTypeRef cf1, CFTypeRef cf2) {
```
rajsesh-msft: >__CGPathEqual [](start = 15, length = 13)

what about the other members of the struct?  Why not placement new the whole struct and have the default values set for the scalar types?

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -31,8 +31,92 @@
...
+    struct __CGPath* path1 = (struct __CGPath*)cf1;
+    struct __CGPath* path2 = (struct __CGPath*)cf2;
+
+    return true;
+}
+
+static void __CGPathInit(CFTypeRef cf) {
```
rajsesh-msft: >__CGPathInit [](start = 12, length = 12)

what about the other members of the struct?  Why not placement new the whole struct and have the default values set for the scalar types?

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -65,126 +149,31 @@ void addPoint(float ptX, float ptY) {
...
 /**
  @Status Interoperable
 */
 CGMutablePathRef CGPathCreateMutable() {
-    return __CGPath::alloc(nil);
+    size_t memSize = sizeof(struct __CGPath) - sizeof(CFRuntimeBase);
+    CGPathRef ret = ((CGPathRef)(_CFRuntimeCreateInstance(kCFAllocatorDefault, CGPathGetTypeID(), memSize, NULL)));
+    struct __CGPath* mutableRet = const_cast<struct __CGPath*>(ret);
+
+    ComPtr<ID2D1Factory> factory = _CreateD2DFactoryInstance();
```
rajsesh-msft: >_CreateD2DFactoryInstance [](start = 35, length = 25)

I recommend moving the code that follows to the ctor of the struct and do a placement new here.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -228,19 +233,20 @@ void CGPathAddLineToPoint(CGMutablePathRef path, const CGAffineTransform* m, flo
...
-
-        x = pt.x;
-        y = pt.y;
     }
 
-    _CGPathAddElement(path, kCGPathElementAddLineToPoint, { x, y });
+    path->_geometrySink->BeginFigure(D2D1::Point2F(path->_currentPoint.x, path->_currentPoint.y), D2D1_FIGURE_BEGIN_FILLED);
+    path->_geometrySink->AddLine(D2D1::Point2F(pt.x, pt.y));
+    path->_geometrySink->EndFigure(D2D1_FIGURE_END_CLOSED);
+
```
rajsesh-msft: +1.  Also our reopening for edit seems to be broken.

---
In reply to: [83128835](https://github.com/Microsoft/WinObjC/pull/1138#discussion_r83128835) [](ancestors = 83128835)

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -65,126 +149,31 @@ void addPoint(float ptX, float ptY) {
...
 */
 CGMutablePathRef CGPathCreateMutable() {
-    return __CGPath::alloc(nil);
+    size_t memSize = sizeof(struct __CGPath) - sizeof(CFRuntimeBase);
+    CGPathRef ret = ((CGPathRef)(_CFRuntimeCreateInstance(kCFAllocatorDefault, CGPathGetTypeID(), memSize, NULL)));
+    struct __CGPath* mutableRet = const_cast<struct __CGPath*>(ret);
+
+    ComPtr<ID2D1Factory> factory = _CreateD2DFactoryInstance();
+
+    if (FAILED(factory->CreatePathGeometry(&mutableRet->_pathGeometry))) {
```
msft-Jeyaram: >FAILED [](start = 8, length = 6)

this macro should make these easier: RETURN_NULL_IF_FAILED(factory->CreatePathGeometry(&mutableRet->_pathGeometry))

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -228,19 +233,20 @@ void CGPathAddLineToPoint(CGMutablePathRef path, const CGAffineTransform* m, flo
...
         pt = CGPointApplyAffineTransform(pt, *m);
-
-        x = pt.x;
-        y = pt.y;
     }
 
-    _CGPathAddElement(path, kCGPathElementAddLineToPoint, { x, y });
+    path->_geometrySink->BeginFigure(D2D1::Point2F(path->_currentPoint.x, path->_currentPoint.y), D2D1_FIGURE_BEGIN_FILLED);
+    path->_geometrySink->AddLine(D2D1::Point2F(pt.x, pt.y));
+    path->_geometrySink->EndFigure(D2D1_FIGURE_END_CLOSED);
```
rajsesh-msft: >EndFigure [](start = 25, length = 9)

Just nuke this function.  This is used elsewhere in UIKit, where this needs to be re-written using public apis (CGPathApply possibly) to get the exclusion path.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -1021,15 +891,17 @@ bool FindLargestRect(CGPoint point, float rectHeight, CGRect& ret) {
 };
 
 CGRect _CGPathFitRect(CGPathRef pathref, CGRect rect, CGSize maxSize, float padding) {
```
rajsesh-msft: Just nuke this function.  This is used elsewhere in UIKit, where this needs to be re-written using public apis (CGPathApply possibly) to get the exclusion path.

Frameworks/CoreGraphics/D2DWrapper.h
rajsesh-msft: I prefer the _get nomenclature as this is a singleton.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -65,126 +149,31 @@ void addPoint(float ptX, float ptY) {
...
 CGMutablePathRef CGPathCreateMutable() {
-    return __CGPath::alloc(nil);
+    size_t memSize = sizeof(struct __CGPath) - sizeof(CFRuntimeBase);
+    CGPathRef ret = ((CGPathRef)(_CFRuntimeCreateInstance(kCFAllocatorDefault, CGPathGetTypeID(), memSize, NULL)));
+    struct __CGPath* mutableRet = const_cast<struct __CGPath*>(ret);
+
+    ComPtr<ID2D1Factory> factory = _CreateD2DFactoryInstance();
+
+    if (FAILED(factory->CreatePathGeometry(&mutableRet->_pathGeometry))) {
+        return nullptr;
```
msft-Jeyaram: mutableRet should be released.

tests/unittests/CoreGraphics/CGPathTests.mm
```diff
@@ -582,3 +582,72 @@ CGPathRef newPathForRoundRect(CGRect rect, CGFloat radius) {
...
+    EXPECT_EQ(y, pathPoint.y);
+}
+
+void EXPECT_SIZEEQ(CGSize pathSize, CGFloat width, CGFloat height) {
+    EXPECT_EQ(width, pathSize.width);
+    EXPECT_EQ(height, pathSize.height);
+}
+
+// Simple tests for the status of a CGPath during the CGPath rework into D2D.
+TEST(CGPath, CGPathD2DSimpleCreation) {
```
rajsesh-msft: D2D here is an implementation detail.  Since you are only calling public APIs, please drop th D2D prefix.  Or have an issue to go rename them later.

tests/unittests/CoreGraphics/CGPathTests.mm
```diff
@@ -582,3 +582,72 @@ CGPathRef newPathForRoundRect(CGRect rect, CGFloat radius) {
...
+    CGMutablePathRef path = CGPathCreateMutable();
+    EXPECT_NE(nullptr, path);
+
+    CGPathMoveToPoint(path, NULL, 50, 50);
```
rajsesh-msft: >CGPathMoveToPoint [](start = 4, length = 17)

What happens if you don't move to a point? Default at 0,0?

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -601,45 +565,20 @@ CGPathRef CGPathRetain(CGPathRef path) {
 }
 
 /**
- @Status Caveat
- @Notes transform property not supported
+@Status Stub
```
rajsesh-msft: What is our plan to track the apis that go from Interoperable/Caveat to Stubs?  We need to make these get back to those known states before merging to develop.

rajsesh-msft:
:clock1:

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -31,8 +31,92 @@
...
+    CGPoint _startingPoint;
+};
+
+// A private helper function for re-opening a path geometry.
+static void __preparePathForEditing(CGPathRef path) {
+    if (path->_isPathClosed) {
+        // Re-open this geometry.
+        ComPtr<ID2D1Factory> factory = _CreateD2DFactoryInstance();
+        if (FAILED(factory->CreatePathGeometry(&path->_pathGeometry))) {
+            return;
```
aballway: Do we want to just return here or should this FAIL_FAST?

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -31,8 +31,92 @@
...
+
+static void __closePath(CGPathRef path) {
+    if (!path->_isPathClosed) {
+        path->_geometrySink->Close();
+        path->_isPathClosed = true;
+    }
+}
+
+static Boolean __CGPathEqual(CFTypeRef cf1, CFTypeRef cf2) {
+    struct __CGPath* path1 = (struct __CGPath*)cf1;
```
aballway: nit: static/const cast?

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -228,19 +233,20 @@ void CGPathAddLineToPoint(CGMutablePathRef path, const CGAffineTransform* m, flo
...
-        CGPoint pt;
-
-        pt.x = x;
-        pt.y = y;
+    __preparePathForEditing(path);
 
+    CGPoint pt;
+    pt.x = x;
+    pt.y = y;
+    if (m) {
```
aballway: nit: rename m

Frameworks/CoreGraphics/CGPath.mm
MSFTFox: the equal function is just a stand  in at the moment.

---
In reply to: [83136640](https://github.com/Microsoft/WinObjC/pull/1138#discussion_r83136640) [](ancestors = 83136640)

Frameworks/CoreGraphics/CGPath.mm
MSFTFox: A comment on the __closepath function has been added. It would be cumbersome to add a comment to every __closePath call.

---
In reply to: [83128915](https://github.com/Microsoft/WinObjC/pull/1138#discussion_r83128915) [](ancestors = 83128915)

Frameworks/CoreGraphics/CGPath.mm
MSFTFox: Shouldn't return null here, we should fail instead of silently pretending things are okay.

---
In reply to: [83132249](https://github.com/Microsoft/WinObjC/pull/1138#discussion_r83132249) [](ancestors = 83132249)

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -31,8 +31,92 @@
...
+using namespace Microsoft::WRL;
+
+struct __CGPath {
+    CFRuntimeBase _base;
+
+    ComPtr<ID2D1PathGeometry> _pathGeometry;
+    ComPtr<ID2D1GeometrySink> _geometrySink;
+
+    bool _isPathClosed;
+    CGPoint _currentPoint;
```
MSFTFox: The current point? I don't think the sink tells us where the current point is.

---
In reply to: [83128196](https://github.com/Microsoft/WinObjC/pull/1138#discussion_r83128196) [](ancestors = 83128196)

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -31,8 +31,121 @@
...
+    __CGPath* path2 = (__CGPath*)cf2;
+
+    return kCFBooleanFalse;
+}
+
+static void __CGPathInit(CFTypeRef cf) {
+    CGPathRef path = (CGPathRef)cf;
+    // ComPtr needs to be manually new'd, to avoid calling Create functions to an incomplete out-pointer
+    __CGPath* mutablePath = const_cast<__CGPath*>(path);
+    new (std::addressof(mutablePath->_pathGeometry)) ComPtr<ID2D1PathGeometry>();
```
rajsesh-msft: >_pathGeometry [](start = 37, length = 13)

I am not sure this is needed.  Simply setting _pathGeometry = nullptr should be sufficient.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -31,8 +31,121 @@
...
+
+    mutablePath->_currentPoint = { 0, 0 };
+    mutablePath->_startingPoint = { 0, 0 };
+    mutablePath->_isFigureClosed = true;
+}
+
+static void __CGPathDeallocate(CFTypeRef cf) {
+    CGPathRef path = (CGPathRef)cf;
+    // ComPtr needs to be manually destructed, since CFTypes do a memset(0) on dealloc
+    path->_pathGeometry.~ComPtr();
```
rajsesh-msft: >ComPtr [](start = 25, length = 6)

ComPtr doesnt allocate any memory on its own.  You should assign nullptr or Release() instead.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -65,126 +178,26 @@ void addPoint(float ptX, float ptY) {
...
-
-    path->_count++;
-}
 
-/**
- @Status Interoperable
-*/
-CGMutablePathRef CGPathCreateMutable() {
-    return __CGPath::alloc(nil);
+    factory.~ComPtr();
```
rajsesh-msft: >ComPtr [](start = 13, length = 6)

Please no.  When factory goes out of scope, the pointer will be released.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -200,47 +213,46 @@ CGPathRef CGPathCreateCopy(CGPathRef path) {
...
-    ret->_count = path->_count;
-    ret->_elements = (CGPathElementInternal*)IwRealloc(ret->_elements, ret->_max * sizeof(CGPathElementInternal));
-    memcpy(ret->_elements, path->_elements, path->_count * sizeof(CGPathElementInternal));
-    // All of the new elements need to call init
-    // because they were created with alloc + memcpy
-    for (unsigned i = 0; i < path->_count; i++) {
-        ret->_elements[i].init();
+    CGMutablePathRef mutableRet = CGPathCreateMutable();
+
+    path->__closePath();
```
rajsesh-msft: >__closePath [](start = 10, length = 11)

Technically, we shouldn't be mutating the path here (which we are by calling __closepath().

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -539,41 +509,45 @@ void CGPathAddEllipseInRect(CGMutablePathRef path, const CGAffineTransform* m, C
...
+@Status Stub
+@Notes
 */
 CGRect CGPathGetBoundingBox(CGPathRef path) {
-    CGRect ret;
+    D2D1_RECT_F bounds;
 
-    if (path == NULL) {
-        return CGRectMake(std::numeric_limits<float>::infinity(), std::numeric_limits<float>::infinity(), 0, 0);
+    path->__closePath();
```
rajsesh-msft: >__closePath [](start = 10, length = 11)

Another case of side effect mutation.

tests/unittests/CoreGraphics/CGPathTests.mm
rajsesh-msft: Lets add some tests that test that.  The previous iteration of code wasnt initializing the points.

---
In reply to: [83301024](https://github.com/Microsoft/WinObjC/pull/1138#discussion_r83301024) [](ancestors = 83301024,83132608)

Frameworks/CoreGraphics/CGPath.mm
DHowett-MSFT: This one I think is needed, for the same reason we have to placement new the Impl structures: the C++ constructor and destructor for these embedded types need to be called before any member operators.

Frameworks/CoreGraphics/CGPath.mm
DHowett-MSFT: Absolutely. Please don't use destructors for general-purpose management of lifetimes. The compiler will make sure they are called at the right times.

You only need the constructor and destructor calls when you bypassed new and delete, like in CFInit and CFFinalize.

DHowett-MSFT: Additionally, using a value after it has been destructed is totally in the realm of undefined behaviour. It is the proverbial kiss of death.

Frameworks/CoreGraphics/CGPath.mm
DHowett-MSFT: If this does that, why is it safe to do it earlier in this file?

Frameworks/CoreGraphics/CGPath.mm
rajsesh-msft: Synced up offline:
1. Revisit the path re-editing code validity.
2. If the path is not "conceptually" mutated, then this is okay, but we need to be mindful of the side effects.  

---
In reply to: [83480290](https://github.com/Microsoft/WinObjC/pull/1138#discussion_r83480290) [](ancestors = 83480290,83345783)

Frameworks/CoreGraphics/CGPath.mm
MSFTFox: I'm not sure what you mean. We do it earlier to re-open the sink on the same path. Here we're copying the contents of one path into a new sink that is representative of a new path.

---
In reply to: [83363300](https://github.com/Microsoft/WinObjC/pull/1138#discussion_r83363300) [](ancestors = 83363300)

Frameworks/CoreGraphics/CGPath.mm
MSFTFox: There will be a document on OneNote with the status of any APIs that have changed status. It may be helpful to look at the API status snapshots on develop to double check any APIs in question.

---
In reply to: [83132884](https://github.com/Microsoft/WinObjC/pull/1138#discussion_r83132884) [](ancestors = 83132884)

DHowett-MSFT:
> #import "CGPathInternal.h"

duped include

---
Refers to: Frameworks/CoreGraphics/CGPath.mm:18 in 56f89cd. [](commit_id = 56f89cd63317856ea0ff534262ce6c5d1d6e2e85, deletion_comment = False)

DHowett-MSFT:
> #import "CGPathInternal.h"

Can you see if removing this breaks anything?

---
Refers to: Frameworks/CoreGraphics/CGPath.mm:18 in 56f89cd. [](commit_id = 56f89cd63317856ea0ff534262ce6c5d1d6e2e85, deletion_comment = False)

DHowett-MSFT:
> #import <CoreGraphics/CGContext.h>

we've imported so many of these it might just make sense to do CoreGraphics/CoreGraphics.h

---
Refers to: Frameworks/CoreGraphics/CGPath.mm:23 in 56f89cd. [](commit_id = 56f89cd63317856ea0ff534262ce6c5d1d6e2e85, deletion_comment = False)

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -31,8 +31,122 @@
...
+
+            // Without a new figure being created, it's by default closed
+            _impl._isFigureClosed = true;
+        }
+    }
+
+    void closePath() {
+        if (_impl._geometrySink) {
+            endFigure(D2D1_FIGURE_END_OPEN);
+            _impl._geometrySink->Close();
```
DHowett-MSFT: >_ [](start = 18, length = 1)

to echo Raj's earlier comment, these nested member _s can die

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -667,25 +628,28 @@ CGPathRef CGPathCreateWithEllipseInRect(CGRect rect, const CGAffineTransform* tr
...
 }
 
 /**
  @Status Stub
  @Notes
 */
-void CGPathAddRelativeArc(
-    CGMutablePathRef path, const CGAffineTransform* matrix, CGFloat x, CGFloat y, CGFloat radius, CGFloat startAngle, CGFloat delta) {
+void CGPathAddRelativeArc(CGMutablePathRef path,
+                          const CGAffineTransform* transformatrix,
```
DHowett-MSFT: >transformatrix [](start = 51, length = 14)

one of these things is not like the others



it's the transformatrix

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -31,8 +31,122 @@
...
+
+    void endFigure(D2D1_FIGURE_END figureStatus) {
+        if (!_impl._isFigureClosed) {
+            _impl._geometrySink->EndFigure(figureStatus);
+            _impl._isFigureClosed = true;
+        }
+    }
+};
+
+static CFBooleanRef __CGPathEqual(CFTypeRef cf1, CFTypeRef cf2) {
```
DHowett-MSFT: >__CGPathEqual [](start = 20, length = 13)

Not used, but perhaps rename to CGPathEqual(...) and stomp the other one.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -15,176 +15,121 @@
...
 
-class BBox {
-public:
-    float x, y, x1, y1;
-    bool xset, yset, x1set, y1set;
+#include <COMIncludes.h>
+#import <wrl/client.h>
+#import <D2d1.h>
+#include <COMIncludes_End.h>
+#include "CFCPPBase.h"
```
DHowett-MSFT: Since this is in a public folder, use `<>`

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -15,176 +15,121 @@
...
```
DHowett-MSFT: >include [](start = 1, length = 7)

Please use `import` unless strictly necessary

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -15,176 +15,121 @@
...
+    // we must open the path again. This cannot be done normally, so we must
+    // create a new path with the old path information to edit.
+    void preparePathForEditing() {
+        if (!_impl.geometrySink) {
+            // Re-open this geometry.
+            ComPtr<ID2D1Factory> factory = _GetD2DFactoryInstance();
+
+            // Create temp vars for new path/sink
+            Microsoft::WRL::ComPtr<ID2D1PathGeometry> newPath;
+            Microsoft::WRL::ComPtr<ID2D1GeometrySink> newSink;
```
DHowett-MSFT: No need to fully qualify these names.

MSFTFox: Sorry, I had this fix locally, forgot to push last night.

---
In reply to: [83929529](https://github.com/Microsoft/WinObjC/pull/1138#discussion_r83929529) [](ancestors = 83929529)

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -15,176 +15,121 @@
...
-    // because it was created with alloc
-    element->init();
-    element->type = type;
-    element->points[0] = p0;
-    element->points[1] = p1;
-    element->points[2] = p2;
-
-    path->_count++;
+static void __CGPathDeallocate(CFTypeRef cf) {
+    CGPathRef path = (CGPathRef)cf;
```
DHowett-MSFT: Unnecessary now!

Frameworks/CoreGraphics/CGPath.mm
DHowett-MSFT: nit: fold into `CGPoint pt{x, y};` to avoid using the assignment operator or copy constructors.

---
In reply to: [83128604](https://github.com/Microsoft/WinObjC/pull/1138#discussion_r83128604) [](ancestors = 83128604)

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -200,62 +145,64 @@ CGPathRef CGPathCreateCopy(CGPathRef path) {
 }
 
 /**
- @Status Interoperable
+@Status Interoperable
+@Notes
 */
```
DHowett-MSFT: Do we have notes for this?

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -200,62 +145,64 @@ CGPathRef CGPathCreateCopy(CGPathRef path) {
...
-        x = pt.x;
-        y = pt.y;
+    CGPoint pt = { x, y };
+    if (transform) {
+        pt = CGPointApplyAffineTransform(pt, *transform);
     }
 
-    _CGPathAddElement(path, kCGPathElementAddLineToPoint, { x, y });
+    path->beginFigure();
+    path->_impl.geometrySink->AddLine(D2D1::Point2F(pt.x, pt.y));
```
DHowett-MSFT: >D2D1::Point2F [](start = 38, length = 13)

Perhaps use helper

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -404,107 +352,72 @@ void CGPathAddArc(CGMutablePathRef path,
...
-        CGPoint pt;
-
-        pt.x = x;
-        pt.y = y;
-
-        pt = CGPointApplyAffineTransform(pt, *m);
-
-        x = pt.x;
-        y = pt.y;
+void CGPathMoveToPoint(CGMutablePathRef path, const CGAffineTransform* transform, float x, float y) {
```
DHowett-MSFT: >CGPathMoveToPoint [](start = 5, length = 17)

What if you move to a point while you're in the middle of drawing a figure?

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -404,107 +352,72 @@ void CGPathAddArc(CGMutablePathRef path,
...
  @Status Interoperable
 */
-void CGPathAddLines(CGMutablePathRef path, const CGAffineTransform* m, const CGPoint* points, size_t count) {
+void CGPathAddLines(CGMutablePathRef path, const CGAffineTransform* transform, const CGPoint* points, size_t count) {
     if (count == 0 || points == NULL || path == NULL) {
         return;
     }
 
-    CGPathMoveToPoint(path, m, points[0].x, points[0].y);
+    CGPathMoveToPoint(path, transform, points[0].x, points[0].y);
```
DHowett-MSFT: >CGPathMoveToPoint [](start = 4, length = 17)

Is this correct? Move first, then add lines? Or does it add a line to the first point.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -404,107 +352,72 @@ void CGPathAddArc(CGMutablePathRef path,
...
+    toAdd->closePath();
+    path->preparePathForEditing();
+    FAIL_FAST_IF_FAILED(toAdd->_impl.pathGeometry->Stream(path->_impl.geometrySink.Get()));
 }
 
 /**
  @Status Interoperable
 */
-void CGPathAddEllipseInRect(CGMutablePathRef path, const CGAffineTransform* m, CGRect rect) {
+void CGPathAddEllipseInRect(CGMutablePathRef path, const CGAffineTransform* transform, CGRect rect) {
```
DHowett-MSFT: >CGPathAddEllipseInRect [](start = 5, length = 22)

This is not for this review:
I wonder if we can use path combining, or somehow stream an ellipse geometry into the path geometry.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -539,41 +452,59 @@ void CGPathAddEllipseInRect(CGMutablePathRef path, const CGAffineTransform* m, C
  @Status Interoperable
 */
 void CGPathCloseSubpath(CGMutablePathRef path) {
-    if (path == NULL) {
-        return;
+    // We cannot use end figure with the built in way of closing a sub path due to
+    // an issue with reading information
```
DHowett-MSFT: This sentence doesn't into sense

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -539,41 +452,59 @@ void CGPathAddEllipseInRect(CGMutablePathRef path, const CGAffineTransform* m, C
...
     if (path == NULL) {
-        return CGRectMake(std::numeric_limits<float>::infinity(), std::numeric_limits<float>::infinity(), 0, 0);
+        return CGRectMake(INFINITY, INFINITY, 0, 0);
     }
 
-    auto context = CGBitmapContextCreate(0, 1, 1, 1, 1, 0, 0);
-    path->_applyPath(context);
-    ret = CGContextGetPathBoundingBox(context);
-    CGContextRelease(context);
+    D2D1_RECT_F bounds;
```
DHowett-MSFT: >D2D1_RECT_F [](start = 4, length = 11)

This could use an inline helper too!

DHowett-MSFT: D2DRectFToCG

---
In reply to: [83932636](https://github.com/Microsoft/WinObjC/pull/1138#discussion_r83932636) [](ancestors = 83932636)

Frameworks/CoreGraphics/CGPath.mm
MSFTFox: Not particularly no.

---
In reply to: [83931964](https://github.com/Microsoft/WinObjC/pull/1138#discussion_r83931964) [](ancestors = 83931964)

Frameworks/CoreGraphics/CGPath.mm
MSFTFox: Raj had the same comment. Ellipses and curves in general might behave differently than we originally thought. I have to look into it more closely but I'm confident we shouldn't be adding 4 curves to represent an ellipse. Stream seems like the best option to append an ellipse.

---
In reply to: [83932426](https://github.com/Microsoft/WinObjC/pull/1138#discussion_r83932426) [](ancestors = 83932426)

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -539,41 +452,59 @@ void CGPathAddEllipseInRect(CGMutablePathRef path, const CGAffineTransform* m, C
...
- @Status Interoperable
+@Status Stub
+@Notes
 */
 CGRect CGPathGetBoundingBox(CGPathRef path) {
-    CGRect ret;
-
     if (path == NULL) {
-        return CGRectMake(std::numeric_limits<float>::infinity(), std::numeric_limits<float>::infinity(), 0, 0);
+        return CGRectMake(INFINITY, INFINITY, 0, 0);
```
MSFTFox: Yes, I confirmed this matches the reference platform's behavior.

---
In reply to: [83932603](https://github.com/Microsoft/WinObjC/pull/1138#discussion_r83932603) [](ancestors = 83932603)

Frameworks/CoreGraphics/CGPath.mm
MSFTFox: I didn't change the behavior from the previous implementation however it is incorrect. It should draw a line to the first point not just move there. I'll update it. I've made a note to add it in the sample app when we update it for supporting more cases like this.

---
In reply to: [83932270](https://github.com/Microsoft/WinObjC/pull/1138#discussion_r83932270) [](ancestors = 83932270)

Frameworks/CoreGraphics/CGPath.mm
MSFTFox: Good luck? Haha. That seems like something explicitly bad to do that sounds like undefined behavior. Here we wouldn't get any particular change since the current point and starting point are only used in tandem with adding actual geometry figures.

---
In reply to: [83932143](https://github.com/Microsoft/WinObjC/pull/1138#discussion_r83932143) [](ancestors = 83932143)

tests/unittests/CoreGraphics/CGPathTests.mm
```diff
@@ -582,3 +582,115 @@ CGPathRef newPathForRoundRect(CGRect rect, CGFloat radius) {
...
+    EXPECT_SIZEEQ(boundingBox.size, 150, 150);
+
+    // Redundant check for confidence
+    boundingBox = CGPathGetBoundingBox(path);
+    EXPECT_POINTEQ(boundingBox.origin, 50, 50);
+    EXPECT_SIZEEQ(boundingBox.size, 31, 57);
+}
+
+TEST(CGPath, CGPathAddPathTest) {
+    CGMutablePathRef path = CGPathCreateMutable();
```
aballway: do these paths need to be freed?

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -404,107 +346,71 @@ void CGPathAddArc(CGMutablePathRef path,
...
 }
 
 /**
  @Status Interoperable
 */
-void CGPathMoveToPoint(CGMutablePathRef path, const CGAffineTransform* m, float x, float y) {
-    if (path == NULL) {
-        return;
+void CGPathMoveToPoint(CGMutablePathRef path, const CGAffineTransform* transform, float x, float y) {
+    CGPoint pt;
```
aballway: nit: pt{x, y}

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -404,107 +346,71 @@ void CGPathAddArc(CGMutablePathRef path,
...
+    pt.x = x;
+    pt.y = y;
+    if (transform) {
```
aballway: nit: this seems to be a common pattern, maybe a convenience helper?

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -539,41 +445,46 @@ void CGPathAddEllipseInRect(CGMutablePathRef path, const CGAffineTransform* m, C
...
-    if (path == NULL) {
-        return;
+    // Move the current point to the starting point since the line is closed.
+    if (!CGPointEqualToPoint(path->_impl.currentPoint, path->_impl.startingPoint)) {
+        path->_impl.currentPoint = path->_impl.startingPoint;
     }
-
-    _CGPathAddElement(path, kCGPathElementCloseSubpath);
+    path->endFigure(D2D1_FIGURE_END_CLOSED);
 }
```
DHowett-MSFT: So now we are just calling endFigure and setting the current point to the starting point?

No need for the point compare, then!

Frameworks/CoreGraphics/CGPath.mm
MSFTFox: Should I file an issue?

---
In reply to: [83963132](https://github.com/Microsoft/WinObjC/pull/1138#discussion_r83963132) [](ancestors = 83963132,83946485,83932143)

Frameworks/CoreGraphics/CGPath.mm
DHowett-MSFT: and now we _are_ allowing D2D to insert the closure?

---
In reply to: [83963372](https://github.com/Microsoft/WinObjC/pull/1138#discussion_r83963372) [](ancestors = 83963372)

Frameworks/CoreGraphics/CGPath.mm
DHowett-MSFT: Per the haps.

---
In reply to: [83963404](https://github.com/Microsoft/WinObjC/pull/1138#discussion_r83963404) [](ancestors = 83963404,83963132,83946485,83932143)

Frameworks/CoreGraphics/CGPath.mm
MSFTFox: Yes, it's fine to let D2D close the path since we moved it to a function that handles the bad case more easily of the path already being closed.

---
In reply to: [83963413](https://github.com/Microsoft/WinObjC/pull/1138#discussion_r83963413) [](ancestors = 83963413,83963372)

tests/testapps/CGCatalog/CGCatalog/CGPathAddLineToPointViewController.m
```diff
@@ -47,6 +47,10 @@ - (void)loadView {
         CGPathAddLineToPoint(thepath, NULL, 300, 100);
         CGPathAddLineToPoint(thepath, NULL, 235, 100);
 
+        CGPoint points[] = { { 300, 200 }, { 400, 200 }, { 400, 300 }, { 300, 300 }, { 300, 200 } };
+
+        CGPathAddLines(thepath, NULL, points, 5);
```
MSFTFox: Oh woops, didn't mean to check this in. I was verifying the behavior of addlines. Ugly star is sad.

---
In reply to: [83963476](https://github.com/Microsoft/WinObjC/pull/1138#discussion_r83963476) [](ancestors = 83963476)

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -15,176 +15,126 @@
...
-        }
+using namespace std;
+using namespace Microsoft::WRL;
 
-        if (ptY > y1 || !y1set) {
-            y1 = ptY;
-            y1set = true;
-        }
+struct __CGPathImpl {
+    ComPtr<ID2D1PathGeometry> pathGeometry{ nullptr };
```
rajsesh-msft: nit: no need to initialize ComPtr with nullptr, the default constructor will do that.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -15,176 +15,126 @@
...
  @Status Interoperable
 */
 CGMutablePathRef CGPathCreateMutable() {
-    return __CGPath::alloc(nil);
+    __CGPath* mutableRet = __CGPath::CreateInstance();
+
+    ComPtr<ID2D1Factory> factory = _GetD2DFactoryInstance();
+
+    FAIL_FAST_IF_FAILED(factory->CreatePathGeometry(&mutableRet->_impl.pathGeometry));
+    FAIL_FAST_IF_FAILED(mutableRet->_impl.pathGeometry->Open(&mutableRet->_impl.geometrySink));
```
rajsesh-msft: Please add this to a utility method in __CGPath (init for example).

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -404,107 +354,67 @@ void CGPathAddArc(CGMutablePathRef path,
...
+    CGPathAddLineToPoint(path, transform, CGRectGetMaxX(rect), CGRectGetMinY(rect));
+    CGPathAddLineToPoint(path, transform, CGRectGetMaxX(rect), CGRectGetMaxY(rect));
+    CGPathAddLineToPoint(path, transform, CGRectGetMinX(rect), CGRectGetMaxY(rect));
     CGPathCloseSubpath(path);
 }
 
 /**
- @Status Interoperable
+ @Status Caveat
+ @Notes Ignores Affine Transform
```
rajsesh-msft: >Affine [](start = 16, length = 6)

please open an issue to fix this (or are you tracking this elsewhere?)

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -404,107 +354,67 @@ void CGPathAddArc(CGMutablePathRef path,
...
  @Status Interoperable
 */
-void CGPathMoveToPoint(CGMutablePathRef path, const CGAffineTransform* m, float x, float y) {
-    if (path == NULL) {
-        return;
-    }
-
-    if (m) {
-        CGPoint pt;
+void CGPathMoveToPoint(CGMutablePathRef path, const CGAffineTransform* transform, CGFloat x, CGFloat y) {
```
rajsesh-msft: I can't find where you are moving the current point in the geometry.  Ex: AddLine just takes the target point, not the starting point.

Frameworks/CoreGraphics/D2DWrapper.h
```diff
@@ -21,4 +21,20 @@
...
 #include <COMIncludes_End.h>
 
-Microsoft::WRL::ComPtr<ID2D1Factory> _GetD2DFactoryInstance();
\ No newline at end of file
+#import <CoreGraphics/CGGeometry.h>
+#import <CoreGraphics/CGBase.h>
+
+Microsoft::WRL::ComPtr<ID2D1Factory> _GetD2DFactoryInstance();
+
+static inline D2D_POINT_2F _CGPointToD2D_F(CGPoint point) {
```
rajsesh-msft: >static [](start = 0, length = 6)

static inline doesn't make sense in header file.  Change these to inline.

msft-Jeyaram:
:shipit:

PR #1142, opened 2016-10-13
Start the runloop asynchronously rather than synchronously during activation

rajsesh-msft:
Our activation code is starting runloop synchornously in the main thread.

This turned out to be bad for some apps.  This change is to do an aysnc task that sets up the runloop.

### DIFF
--- a/Frameworks/Foundation/MainDispatcher.mm
+++ b/Frameworks/Foundation/MainDispatcher.mm
@@ -107,7 +107,7 @@ extern "C" void ScheduleMainRunLoopAsync() {
         ScheduleMainRunLoop();
         return S_OK;
     });
-    _ensureDispatcher();
+    _EnsureDispatcher();
     ComPtr<IAsyncAction> asyncAction;
     s_dispatcher->RunAsync(CoreDispatcherPriority_Normal, dispatchCallback.Get(), &asyncAction);
 }
@@ -132,8 +132,6 @@ int MainRunLoopTimedMultipleWait(EbrEvent* events, int numEvents, double timeout
                           });
             return 0;
         }
-
-        _EnsureDispatcher();
 
         if (timeout == 0) {
             // Optimization to not schedule a wait on a threadpool when the timeout value passed was 0.
--- a/Frameworks/Foundation/MainDispatcher.mm
+++ b/Frameworks/Foundation/MainDispatcher.mm
@@ -133,7 +133,7 @@ int MainRunLoopTimedMultipleWait(EbrEvent* events, int numEvents, double timeout
             return 0;
         }
 
-        _ensureDispatcher();
+        _EnsureDispatcher();
 
         if (timeout == 0) {
             // Optimization to not schedule a wait on a threadpool when the timeout value passed was 0.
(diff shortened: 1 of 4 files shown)


rajsesh-msft:

Frameworks/UIKit/StarboardXaml/ApplicationMain.mm
```diff
@@ -147,7 +147,7 @@ int ApplicationMainStart(const char* principalName,
     [displayMode _updateDisplaySettings];
 
     UIApplicationMainInit(principalClassName, delegateClassName, defaultOrientation, (int)activationType, activationArgument);
-    ScheduleMainRunLoop();
```
rajsesh-msft: >ScheduleMainRunLoop [](start = 4, length = 19)

This was taking > 20s for internal test app.  With async, we take that cost after app activation so PLM doesn't kill us.

Frameworks/Foundation/MainDispatcher.mm
```diff
@@ -85,6 +85,34 @@ void _DispatchMainRunLoopOnUIThread(int signaledEvent) {
 }
 
 /**
+* Private method to create the dispatcher
+*/
+void _ensureDispatcher() {
```
jaredhms: >_ensureDispatcher [](start = 5, length = 17)

nit: should be _PascalCase

jaredhms:
:shipit:

msft-Jeyaram:
:shipit:

Frameworks/Foundation/MainDispatcher.mm
```diff
@@ -85,6 +85,34 @@ void _DispatchMainRunLoopOnUIThread(int signaledEvent) {
 }
 
 /**
+* Private method to create the dispatcher
+*/
+void _EnsureDispatcher() {
```
DHowett-MSFT: super-nit: should this be static?

Frameworks/Foundation/MainDispatcher.mm
```diff
@@ -85,6 +85,34 @@ void _DispatchMainRunLoopOnUIThread(int signaledEvent) {
...
+/**
+ * Method that schedules the main runloop asynchronously on the UI thread.
+ * Note: This method always schedules the work asynchronously on the UI thread to let the caller stack unwind.
+ */
+extern "C" void ScheduleMainRunLoopAsync() {
+    auto dispatchCallback = WRLHelpers::MakeAgileCallback<IDispatchedHandler>([]() -> HRESULT {
+        ScheduleMainRunLoop();
+        return S_OK;
+    });
+    _ensureDispatcher();
```
DHowett-MSFT: I think one `e` was missed here!

Frameworks/Foundation/MainDispatcher.mm
```diff
@@ -85,6 +85,34 @@ void _DispatchMainRunLoopOnUIThread(int signaledEvent) {
 }
 
 /**
+* Private method to create the dispatcher
+*/
+void _EnsureDispatcher() {
+    if (s_dispatcher == nullptr) {
```
DHowett-MSFT: Can we be sure this is thread-safe?

rajsesh-msft: Good point.  This is always initialized in the main thread, so the _EnsureDispatcher in the other thread is unnecessary, I'll remove that.

---
In reply to: [83293021](https://github.com/Microsoft/WinObjC/pull/1142#discussion_r83293021) [](ancestors = 83293021)

PR #1132, opened 2016-10-12
Implement NSLayoutManager and NSString+UIKit using CoreText

aballway:
NSLayoutManager and NSString drawing and layout are changed to use CoreText rather than manually laying out and drawing glyphs.  Also does small cleanup through CoreText such as changing DWrite to use CFAttributedStringRef rather than a CFTypesetter, as well as fixing our descent which should be inverse of what DWrite gives us.  

Fixes #1017

PR #1145, opened 2016-10-13
Implement CGFont with DWrite

### LINKED ISSUE #1018: Move CGFont to use DirectWrite
This is required to remove FreeType dependency with CG
- [ ] Caveat	CTFontCopyGraphicsFont
- [ ] Caveat	CTFontCreateWithGraphicsFont

### LINKED ISSUE #1065: Implement CTFontCopyTable
Possibly also implement 
* [ ] CTFontCopyAvailableTables

ms-jihua:
- Move majority of DWriteWrapper into CoreGraphics, leave CT-specific functions in CoreText as DWriteWrapper_CoreText
     - Minor edits to handling of keywords "Medium", "Demi", and "Cond" when creating a font with name
 - Strip out use of FreeType in CGContextCairo/Impl
 - Implement CTFont bounding box functions, as they had counterparts in CGFont
 - Add unit tests

Fixes #1018
Fixes #1065


Frameworks/UIKit/NSLayoutManager.mm
```diff
@@ -40,189 +40,7 @@ static float invokeWidthBlock(void* opaque, CFIndex idx, float offset, float hei
 }
 
 - (void)_layoutAllText {
```
ms-jihua: #1132 deals with this, but I needed to get rid of this locally to unbreak some internal test apps.

Frameworks/UIKit/NSString+UIKitAdditions.mm
```diff
@@ -103,274 +49,7 @@ static void drawString(UIFont* font,
                        UITextAlignment alignment,
                        CGSize* sizeOut,
                        CGPoint* lastCharPos = NULL) {
-    if (font == nil) {
```
ms-jihua: #1132 deals with this, but I needed to get rid of this locally to unbreak some internal test apps.

Frameworks/UIKit/UIApplication.mm
```diff
@@ -698,7 +698,8 @@ - (void)setApplicationIconBadgeNumber:(int)num {
     WUNBadgeNotification* notification = [WUNBadgeNotification makeBadgeNotification:doc];
     WUNBadgeUpdater* updater = [WUNBadgeUpdateManager createBadgeUpdaterForApplication];
 
-    [updater update:notification];
+    // TODO #: 0x803e0208 : The notification platform does not have the proper privileges to complete the request.
```
rajsesh-msft: Ugh?? Lets get a bug on this, should be unrelated to what you are doing.

ms-jihua: Yeah I think this repro'd on develop also.

tests/unittests/CoreGraphics/CGFontTests.mm
```diff
@@ -0,0 +1,117 @@
...
+    EXPECT_EQ(3884, box.size.width);
+#endif
+
+    font = CGFontCreateWithFontName(name3);
+    CFAutorelease(font);
+    box = CGFontGetFontBBox(font);
+
+// For Courier New Italic, both horizontal attributes are off by exactly 115, and both vertical attributes are off by exactly 9
+#if TARGET_OS_WIN32
+    EXPECT_EQ(-138, box.origin.x);
```
rajsesh-msft: How different do the font looks visually?

tests/unittests/CoreText/CTFontTests.mm
```diff
@@ -435,4 +435,105 @@ virtual void TearDown() {
...
+    }
+
+    // Null outpointer for boxes should evaluate just the total box
+    totalBox = CTFontGetBoundingRectsForGlyphs(font, kCTFontHorizontalOrientation, glyphs, nullptr, count);
+    EXPECT_NEAR(expectedTotalBox.origin.x, totalBox.origin.x, c_errorMargin);
+    EXPECT_NEAR(expectedTotalBox.origin.y, totalBox.origin.y, c_errorMargin);
+    EXPECT_NEAR(expectedTotalBox.size.width, totalBox.size.width, c_errorMargin);
+    EXPECT_NEAR(expectedTotalBox.size.height, totalBox.size.height, c_errorMargin);
+
+// DWrite seems to not respect orientation
```
rajsesh-msft: Does it do horizontal regardless or fails?

tests/unittests/CoreGraphics/CGFontTests.mm
ms-jihua: Didn't actually check but there's a 0% chance in my mind that it could look different at all. This API returns the total accumulated bounding box of every glyph in the font. As you can see below, we get correct bounding boxes for individual glyphs, so it's most likely that one version of the font has some extra glyphs.

tests/unittests/CoreText/CTFontTests.mm
rajsesh-msft: Would be helpful to follow-up with the DWrite folks offline about GetDesignGlyphMetrics.  Looking at the DWrite code, !sideways should give us the vertical orientation bounding boxes.  

---
In reply to: [83328388](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83328388) [](ancestors = 83328388)

rajsesh-msft: Of course, lets not block this change while we sort that through.

---
In reply to: [83329813](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83329813) [](ancestors = 83329813,83328388)

build/CoreGraphics/dll/CoreGraphics.def
```diff
@@ -276,13 +292,13 @@ LIBRARY CoreGraphics
...
-        CGFontWrap
-        _CGFontLock
-        _CGFontUnlock
+        ; CGFontFitChars
+        ; CGFontGetGlyphs
+        ; CGFontGetGlyphsForUnichars
+        ; CGFontSetFTFontSize
+        ; CGFontWrap
+        ; _CGFontLock
+        ; _CGFontUnlock
```
rajsesh-msft: Assuming these are not gone for good?  Do we have an issue to track putting these back in?

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,572 @@
...
+ *
+ * Note: This function currently uses a cache, meaning that fonts installed during runtime will not be reflected
+ */
+CFStringRef _DWriteGetFamilyNameForFontName(CFStringRef fontName) {
+    static CFDictionaryRef fontToFamilyMap = nullptr;
+
+    CFLocaleRef locale = CFLocaleCopyCurrent();
+    CFAutorelease(locale);
+
+    static dispatch_once_t initOnce = 0;
```
rajsesh-msft: >dispatch [](start = 11, length = 8)

nit: We dont need to take a dependency on dispatch for initializing static variable here, you can assign the return value of the lambda to it.

build/CoreGraphics/dll/CoreGraphics.def
ms-jihua: Oops, these are gone for good. I forgot to actually delete them.

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,572 @@
...
+                static_cast<int32_t>(glyphMetrics[i].advanceWidth) - (glyphMetrics[i].leftSideBearing + glyphMetrics[i].rightSideBearing);
+            boundingRects[i].size.height =
+                static_cast<int32_t>(glyphMetrics[i].advanceHeight) - (glyphMetrics[i].bottomSideBearing + glyphMetrics[i].topSideBearing);
+        }
+
+    } else {
+        TraceError(TAG, L"Unable to get bounding box for glyphs");
+    }
+
+    return ret;
```
bbowman: RETURN_IF_FAILED?

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,572 @@
...
+}
+
+/**
+ * Gets individual bounding boxes for specific glyphs in a DWrite font face, as a CGRect
+ * Returns in terms of design units
+ */
+HRESULT _DWriteFontGetBoundingBoxesForGlyphs(
+    ComPtr<IDWriteFontFace> fontFace, const CGGlyph* glyphs, CGRect* boundingRects, size_t count, bool isSideways) {
+    std::vector<DWRITE_GLYPH_METRICS> glyphMetrics = std::vector<DWRITE_GLYPH_METRICS>(count);
```
bbowman: > = std::vector<DWRITE_GLYPH_METRICS>(count) [](start = 50, length = 43)

why assignment?

Frameworks/CoreGraphics/CGFont.mm
```diff
@@ -16,615 +16,134 @@
...
+    struct __CGFont* font1 = (struct __CGFont*)cf1;
+    struct __CGFont* font2 = (struct __CGFont*)cf2;
 
-#define USE_CAIRO
-
-void CGFontSetFTFontSize(id uiFont, void* ftFont, float pointSize, float scale) {
-    UIFont* fontInfo = uiFont;
-
-    FT_Set_Char_Size((FT_Face)ftFont, 0, (FT_F26Dot6)(pointSize * 64.0f), (FT_UInt)(72.f * scale), (FT_UInt)(72.f * scale));
+    return (font1->_dwriteFontFace == font2->_dwriteFontFace);
```
msft-Jeyaram: are we ignoring _metrics and _cachedMetrics?

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,572 @@
...
+        // Avoid dividing by 0, skip math for dividing with 0
+        return -0.0f;
+    }
+}
+
+/**
+ * Gets the accumulated bounding box of all the glyphs in a DWrite font face, as a CGRect
+ * Returns in terms of design units
+ */
+CGRect _DWriteFontGetBoundingBox(ComPtr<IDWriteFontFace> fontFace) {
```
bbowman: >ComPtr<IDWriteFontFace> fontFace [](start = 33, length = 32)

:-( This is a bad pattern to use. In C++ everything is pass by value by default. This means that you are passing a ComPtr<> by value which means that you making a copy of a ComPtr object and unnecessarily doing an extra AddRef Release on it. 

Please use const ComPtr<>& as I hoped you had gotten used to in previous projects with shared_ptr ...

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,572 @@
...
+    void* tableContext;
+    BOOL exists;
+
+    if (FAILED(fontFace->TryGetFontTable(tag, &tableData, &tableSize, &tableContext, &exists)) || !exists) {
+        TraceError(TAG, L"Unable to get font table for tag %u", tag);
+        return nullptr;
+    }
+
+    // Copy the font table's binary data to a CFDataRef
+    CFDataRef ret = CFDataCreate(kCFAllocatorDefault, reinterpret_cast<const byte*>(tableData), tableSize);
```
bbowman: >CFDataRef ret [](start = 4, length = 13)

consider using a RAII friendly construct to avoid leaking this guy if an exception happens before its handed off. unique_cf is probably what you are looking for ...

Frameworks/CoreGraphics/CGFont.mm
```diff
@@ -16,615 +16,134 @@
...
-
-            FT_Done_Glyph(glyph);
-        }
-        _CGFontUnlock();
-
-        return ret;
+CFStringRef CGFontCopyPostScriptName(CGFontRef font) {
+    if (!font) {
+        return nullptr;
     }
```
msft-Jeyaram: RETURN_NULL_IF

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,572 @@
...
+    return font->CreateFontFace(outFontFace);
+}
+
+/**
+ * Gets an informational string as a CFString from a DWrite font face.
+ */
+CFStringRef _DWriteFontCopyInformationalString(ComPtr<IDWriteFontFace> fontFace, DWRITE_INFORMATIONAL_STRING_ID informationalStringId) {
+    if (fontFace == nullptr) {
+        return nullptr;
+    }
```
bbowman: RETURN_NULL_IF

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,572 @@
...
+        TraceError(TAG, L"Unable to find family for font name \"%hs\"", [static_cast<NSString*>(name) UTF8String]);
+        return E_INVALIDARG;
+    }
+
+    ComPtr<IDWriteFontFamily> fontFamily;
+
+    if (FAILED(_DWriteCreateFontFamilyWithName(familyName, &fontFamily)) || !fontFamily) {
+        TraceError(TAG, L"Unable to find family for font name \"%hs\"", [static_cast<NSString*>(name) UTF8String]);
+        return E_INVALIDARG;
+    }
```
bbowman: RETURN macro here?

in general you want to avoid blindly originating a new HR as that loses context when looking at logs. all of the RETURN_* macros trace and the bottom most one *should* be using some logging to originate the error. blind conversions etc make it confusing when looking at it later.

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,572 @@
...
+
+    ComPtr<IDWriteFont> font;
+    HRESULT hr = fontFamily->GetFirstMatchingFont(weight, stretch, style, &font);
+    if (FAILED(hr)) {
+        TraceError(TAG, L"Unable to create font for name \"%hs\"", [static_cast<NSString*>(name) UTF8String]);
+        return hr;
+    }
+
```
bbowman: RETURN_IF_....

Frameworks/CoreGraphics/CGFont.mm
```diff
@@ -753,46 +323,54 @@ CGFloat CGFontGetStemV(CGFontRef font) {
...
-
-    return ret;
-}
+/**
+ @Status Interoperable
+*/
+bool CGFontGetGlyphAdvances(CGFontRef font, const CGGlyph* glyphs, size_t count, int* advances) {
+    if (!font) {
+        return false;
+    }
```
msft-Jeyaram: RETURN_FALSE_IF

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,572 @@
...
+    BOOL fontFamilyExists;
+
+    THROW_IF_FAILED(systemFontCollection->FindFamilyName(reinterpret_cast<const wchar_t*>(unicharFamilyName.data()),
+                                                         &fontFamilyIndex,
+                                                         &fontFamilyExists));
+
+    if (!fontFamilyExists) {
+        TraceError(TAG, L"Unable to find font family \"%ws\"", unicharFamilyName.data());
+        return E_INVALIDARG;
+    }
```
bbowman: mixing throw and return semantics is not a nice game to play. It makes the callers take on extra responsibility and code paths for your failure cases. Use one or the other. Since you return HRESULT the choice should be RETURN_*

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,572 @@
...
+HRESULT _DWriteCreateFontFamilyWithName(CFStringRef familyName, IDWriteFontFamily** outFontFamily) {
+    ComPtr<IDWriteFactory> dwriteFactory;
+    THROW_IF_FAILED(_DWriteCreateFactoryInstance(&dwriteFactory));
+
+    ComPtr<IDWriteFontCollection> systemFontCollection;
+    THROW_IF_FAILED(dwriteFactory->GetSystemFontCollection(&systemFontCollection));
+
+    CFIndex familyNameLength = CFStringGetLength(familyName);
+    std::vector<UniChar> unicharFamilyName(familyNameLength + 1);
+    CFStringGetCharacters(familyName, CFRangeMake(0, familyNameLength), unicharFamilyName.data());
```
bbowman: There are faster mechanisms here with the GetCharactersPtr functions. Please take a look at those.

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,572 @@
...
+    // Get the direct write factory instance
+    ComPtr<IDWriteFactory> dwriteFactory;
+    HRESULT hr = _DWriteCreateFactoryInstance(&dwriteFactory);
+    if (FAILED(hr)) {
+        TraceError(TAG, L"Unable to create DWrite factory");
+        return hr;
+    }
+
+    return dwriteFactory
+        ->CreateTextFormat(fontFamilyName, NULL, weight, style, stretch, fontSize, __GetUserDefaultLocaleName().data(), outTextFormat);
```
bbowman: awkward indents. clang format?

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,572 @@
...
+    }
+
+    // Store weight, stretch, and style in a single struct to be passed into a CFSetApplierFunction's context
+    struct PropertyContext {
+        DWRITE_FONT_WEIGHT* weight;
+        DWRITE_FONT_STRETCH* stretch;
+        DWRITE_FONT_STYLE* style;
+    };
+
+    struct PropertyContext propertyContext = { weight, stretch, style };
```
bbowman: werid to have a struct definition in a function?

Frameworks/CoreText/CTFont.mm
```diff
@@ -692,13 +704,50 @@ CGGlyph CTFontGetGlyphWithName(CTFontRef font, CFStringRef glyphName) {
...
+        return CGRectNull;
+    }
+
+    if (boundingRects) {
+        if (FAILED(_DWriteFontGetBoundingBoxesForGlyphs(
+                font->_dwriteFontFace, glyphs, boundingRects, count, (orientation == kCTFontVerticalOrientation)))) {
+            TraceError(g_logTag, L"Unable to get glyph bounding boxes");
+            return CGRectNull;
+
+        } else {
```
msft-Jeyaram: >else { [](start = 10, length = 6)

not needed.

Frameworks/CoreText/CTFont.mm
```diff
@@ -692,13 +704,50 @@ CGGlyph CTFontGetGlyphWithName(CTFontRef font, CFStringRef glyphName) {
...
+                // but that seems like overkill for now
+                ret.origin.x = std::min(ret.origin.x, boundingRects[i].origin.x);
+                ret.origin.y = std::min(ret.origin.y, boundingRects[i].origin.y);
+                ret.size.width = std::max(ret.size.width, boundingRects[i].size.width);
+                ret.size.height = std::max(ret.size.height, boundingRects[i].size.height);
+            }
+            return ret;
+        }
+
+    } else {
```
msft-Jeyaram: >else { [](start = 6, length = 6)

can be removed to reduce the if nests

Frameworks/CoreGraphics/CGFont.mm
ms-jihua: they're an optimization to avoid calling getters on _dwriteFontFace every time, they don't matter in terms of comparison.

---
In reply to: [83337659](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83337659) [](ancestors = 83337659)

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,572 @@
...
+    *style = DWRITE_FONT_STYLE_NORMAL;
+
+    *familyName = _DWriteGetFamilyNameForFontName(fontName);
+
+    // Relationship of family name -> font name not always consistent
+    // Usually, properties are added to the end (eg: Arial -> Arial Narrow Bold)
+    // However, this is not always the case (eg: Eras ITC -> Eras Bold ITC)
+    // In addition, some fonts with properties are occasionally placed into their own family (eg: Segoe WP SemiLight -> Segoe WP SemiLight)
+    // Try to be more prudent about these edge cases, by looking only at the difference between the font name and family name
+    CFArrayRef fontNameTokens = CFStringCreateArrayBySeparatingStrings(kCFAllocatorDefault, fontName, CFSTR(" "));
```
bbowman: >fontNameTokens [](start = 15, length = 14)

consider using unique_cf instead of autoreleasing in CF land as CFAutorelease doesn't actually work on any non bridged class / before Foundation is loaded.

Frameworks/CoreText/DWriteWrapper_CoreText.mm
```diff
@@ -971,8 +583,8 @@ HRESULT _DWriteCreateFontFaceWithFontDescriptor(CTFontDescriptorRef fontDescript
         }
 
         // Create a best matching font based on the family name and weight/stretch/style
-        ComPtr<IDWriteFontFamily> fontFamily = __GetDWriteFontFamily(familyName);
-        if (!fontFamily) {
+        ComPtr<IDWriteFontFamily> fontFamily;
+        if (FAILED(_DWriteCreateFontFamilyWithName(familyName, &fontFamily)) || !fontFamily) {
             TraceError(TAG, L"Unable to find font family named \"%hs\"", [static_cast<NSString*>(familyName) UTF8String]);
```
msft-Jeyaram: >UTF8String [](start = 109, length = 10)

UTF8String is a heavy call.
it seems we are over using it.

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,572 @@
...
+ * Note: This function currently uses a cache, meaning that fonts installed during runtime will not be reflected
+ */
+CFStringRef _DWriteGetFamilyNameForFontName(CFStringRef fontName) {
+    static CFDictionaryRef fontToFamilyMap = nullptr;
+
+    CFLocaleRef locale = CFLocaleCopyCurrent();
+    CFAutorelease(locale);
+
+    static dispatch_once_t initOnce = 0;
+    dispatch_once(&initOnce,
```
bbowman: This pattern is rarely needed with C++11 unless you need to control on which thread the one time init happens. I would prefer not taking this dep if you can re-write to just use c++ function local static init.

Frameworks/CoreGraphics/DWriteWrapper.mm
ms-jihua: Tried them, they generally don't work reliably, I'd rather not depend on them.

---
In reply to: [83338650](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83338650) [](ancestors = 83338650)

msft-Jeyaram:
>     _needsLayout = TRUE;

just caught my eye. super init?

---
Refers to: Frameworks/UIKit/NSLayoutManager.mm:60 in 04ba630. [](commit_id = 04ba630750d3ea34edeb9c925d456d0a4904b042, deletion_comment = False)

Frameworks/CoreGraphics/CGFont.mm
```diff
@@ -16,615 +16,134 @@
...
-
-        error = FT_Get_Advance(face, glyphs[i], FT_LOAD_NO_SCALE | FT_LOAD_IGNORE_TRANSFORM, &fixed);
-        assert(error == 0);
-
-        advances[i] = fixed;
+    if (FAILED(_DWriteCreateFontFaceWithName(name, &mutableRet->_dwriteFontFace))) {
+        TraceError(g_logTag, L"Failed to create font.");
+        CFRelease(ret);
+        return nullptr;
     }
```
bbowman: return / throw if failed.

Frameworks/CoreGraphics/CGFont.mm
```diff
@@ -16,615 +16,134 @@
...
-    return (CGFontRef)[[_LazyUIFont fontWithName:(NSString*)name size:1.0f] retain];
-}
-
-/**
- @Status Interoperable
-*/
-bool CGFontGetGlyphAdvances(CGFontRef font, const CGGlyph* glyphs, size_t count, int* advances) {
-    DWORD i;
+    size_t memSize = sizeof(struct __CGFont) - sizeof(CFRuntimeBase);
+    CGFontRef ret = static_cast<CGFontRef>(_CFRuntimeCreateInstance(kCFAllocatorDefault, CGFontGetTypeID(), memSize, NULL));
```
bbowman: >ret [](start = 14, length = 3)

use a RAII construct to avoid ugly release logic that is easy to miss when adding code later.

bbowman: >_CFRuntimeCreateInstance [](start = 43, length = 24)

probably should have a well typed create method

tests/unittests/CoreGraphics/CGFontTests.mm
```diff
@@ -0,0 +1,117 @@
...
+#import <CoreFoundation/CoreFoundation.h>
+#import <Foundation/Foundation.h>
+
+// CTFont functionality is generally a superset of CGFont functionality, and hits the same code path
+// Thus, CTFont unit tests can also be thought of as CGFont test coverage
+
+TEST(CGFont, GetFontBBox) {
+// Font names differ slightly between platforms
+#if TARGET_OS_WIN32
+    CFStringRef name = CFSTR("Arial Bold Italic");
```
msft-Jeyaram: >Arial Bold Italic [](start = 30, length = 17)

I'm sure there would be more tests for fonts at some point in the future.
declare them above as consts. this will reduce the #if/#endif usage every time we want font names.

tests/unittests/CoreText/CTFontTests.mm
```diff
@@ -435,4 +435,105 @@ virtual void TearDown() {
         EXPECT_EQ(0, verAdvances[i].height); // Yes actually!
     }
 #endif
+}
+
+TEST(CTFont, GetBoundingBox) {
+// Font names differ slightly between platforms
+// Windows may have slightly different versions of the same fonts compared to the reference platform
+#if TARGET_OS_WIN32
+    CFStringRef name = CFSTR("Arial Italic");
```
msft-Jeyaram: >Arial Italic [](start = 30, length = 12)

better yet, have a test header file that defines these for OSX and Windows. this will reduce the #if defs and make it easier to maintain.

Frameworks/CoreGraphics/CGFont.mm
bbowman: is a ComPtr comparison really what you want? This is probably a pointer comparison which makes some assumptions on how COM will return pointers to you for the proxy objects. Not super sure that is always what you want ... 

---
In reply to: [83339592](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83339592) [](ancestors = 83339592,83337659)

Frameworks/CoreText/CTFont.mm
```diff
@@ -692,13 +704,50 @@ CGGlyph CTFontGetGlyphWithName(CTFontRef font, CFStringRef glyphName) {
...
+ @Notes  kCTFontVerticalOrientation is unsupported, returns the same as default and horizontal orientations
 */
 CGRect CTFontGetBoundingRectsForGlyphs(
     CTFontRef font, CTFontOrientation orientation, const CGGlyph glyphs[], CGRect* boundingRects, CFIndex count) {
-    UNIMPLEMENTED();
-    return StubReturn();
+    if (!glyphs || count <= 0) {
+        TraceError(g_logTag, L"No glyphs were specified");
+        return CGRectNull;
+    }
```
bbowman: macro use in this function

bbowman:
:clock1:

Frameworks/CoreGraphics/DWriteWrapper.mm
ms-jihua: oops, it's been long enough that I forgot, sorry!

---
In reply to: [83337669](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83337669) [](ancestors = 83337669)

Frameworks/CoreText/CTFont.mm
```diff
@@ -692,13 +704,50 @@ CGGlyph CTFontGetGlyphWithName(CTFontRef font, CFStringRef glyphName) {
...
+            TraceError(g_logTag, L"Unable to get glyph bounding boxes");
+            return CGRectNull;
+
+        } else {
+            CGRect ret = { { std::numeric_limits<CGFloat>::max(), std::numeric_limits<CGFloat>::max() },
+                           { std::numeric_limits<CGFloat>::lowest(), std::numeric_limits<CGFloat>::lowest() } };
+            for (size_t i = 0; i < count; ++i) {
+                // DWrite uses design units, where as CTFont uses point size
+                // Scale all the boundingRects
+                boundingRects[i].origin.x = __CTFontScaleMetric(font, boundingRects[i].origin.x);
```
rajsesh-msft: Nit: Given this is used in many places, may be worth adding __CTFontScaleMetric(font, CGRect*).

Frameworks/CoreText/CTFont.mm
```diff
@@ -838,12 +887,11 @@ CFArrayRef CTFontCopyAvailableTables(CTFontRef font, CTFontTableOptions options)
 }
 
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
```
rajsesh-msft: Caveat then?

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,572 @@
...
+    }
+
+    // Get the default locale for this user.
+    wstring localeName = __GetUserDefaultLocaleName();
+
+    uint32_t index = 0;
+    BOOL exists = false;
+
+    // If the default locale is returned, find that locale name, otherwise use "en-us".
+    THROW_IF_FAILED(localizedString->FindLocaleName(localeName.c_str(), &index, &exists));
```
rajsesh-msft: Since we are not catching any exceptions in CT, any of these THROW_IFs will result in some some CT call throwing an exception when something fails.  Not sure if the caller expects CT apis to throw exceptions.

Frameworks/CoreGraphics/CGFont.mm
```diff
@@ -16,615 +16,134 @@
...
-    //  Did not wrap
-    if (state->lineLen == -1) {
-        state->lineLen = &state->chars[state->count] - state->lineStart;
-        if (sizeOut) {
-            sizeOut->width = ((float)penX) / 64.0f;
-        }
-    }
-
-    return ret;
+static void __CGFontInit(CFTypeRef cf) {
```
rajsesh-msft: nit: maybe at a later point, we should use the new CoreFoundation::CppBase

rajsesh-msft:
:clock1:

Frameworks/CoreGraphics/DWriteWrapper.mm
aballway: nope! if the variable is static the lambda assigning it will only be called once, at least it better be or our activation scheme is terribly flawed...

build/CoreGraphics/dll/CoreGraphics.def
```diff
@@ -36,6 +36,22 @@ LIBRARY CoreGraphics
         _CGBitmapContextCreateWithFormat
         CGBitmapContextGetImage
 
+        ; DWriteWrapper Additions
+        _CFStringFromLocalizedString
+        _DWriteGetFontFamilyNames
+        _DWriteGetFontNamesForFamilyName
+        _DWriteGetFamilyNameForFontName
+        _InitDWriteFontPropertiesFromName
```
msft-Jeyaram: >_InitDWriteFontPropertiesFromName [](start = 8, length = 33)

change of naming convention

msft-Jeyaram:
>         return frame;

needs to be autoreleased.

---
Refers to: Frameworks/CoreText/DWriteWrapper_CoreText.mm:372 in 04ba630. [](commit_id = 04ba630750d3ea34edeb9c925d456d0a4904b042, deletion_comment = False)

msft-Jeyaram:
> }

remove, not needed.

---
Refers to: Frameworks/CoreText/CTFramesetter.mm:30 in 04ba630. [](commit_id = 04ba630750d3ea34edeb9c925d456d0a4904b042, deletion_comment = False)

msft-Jeyaram:
>     }

Macro

---
Refers to: Frameworks/CoreText/CTFramesetter.mm:36 in 04ba630. [](commit_id = 04ba630750d3ea34edeb9c925d456d0a4904b042, deletion_comment = False)

msft-Jeyaram:

update header

---
Refers to: Frameworks/CoreText/CTFramesetter.mm:3 in 04ba630. [](commit_id = 04ba630750d3ea34edeb9c925d456d0a4904b042, deletion_comment = False)

msft-Jeyaram:
> CTLineRef CTTypesetterCreateLineWithOffset(CTTypesetterRef ts, CFRange range, double offset) {

null case is not handled, also _DWriteGetFrame (and the function path where ts ends up) is not handling null cases.

---
Refers to: Frameworks/CoreText/CTTypesetter.mm:87 in 04ba630. [](commit_id = 04ba630750d3ea34edeb9c925d456d0a4904b042, deletion_comment = False)

msft-Jeyaram:
>     THROW_NS_IF_FALSE(E_UNEXPECTED, [frame->_lines count] == 1);

curious, we see the use of return null and throwing exceptions.
what is the standard you are making here?
Also if ts was invalid (null), what is the reference platform behavior? are we trying to match that? I'm curious since I've seen some portion of the code return null for some cases and throw exception for the others.

---
Refers to: Frameworks/CoreText/CTTypesetter.mm:89 in 04ba630. [](commit_id = 04ba630750d3ea34edeb9c925d456d0a4904b042, deletion_comment = False)

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,572 @@
...
+ */
+HRESULT _DWriteFontGetBoundingBoxesForGlyphs(
+    ComPtr<IDWriteFontFace> fontFace, const CGGlyph* glyphs, CGRect* boundingRects, size_t count, bool isSideways) {
+    std::vector<DWRITE_GLYPH_METRICS> glyphMetrics = std::vector<DWRITE_GLYPH_METRICS>(count);
+
+    HRESULT ret = fontFace->GetDesignGlyphMetrics(glyphs, count, glyphMetrics.data(), isSideways);
+
+    if (SUCCEEDED(ret)) {
+        // Convert from DWRITE_GLYPH_METRICS to CGRect
+        for (size_t i = 0; i < count; ++i) {
```
aballway: nit: std::transform with a conversion function

Frameworks/CoreText/DWriteWrapper_CoreText.mm
```diff
@@ -1161,20 +743,18 @@ CFStringRef _DWriteFontCopyName(Microsoft::WRL::ComPtr<IDWriteFontFace> fontFace
...
-
     if (CFEqual(nameKey, kCTFontCopyrightNameKey)) {
         informationalStringId = DWRITE_INFORMATIONAL_STRING_COPYRIGHT_NOTICE;
     } else if (CFEqual(nameKey, kCTFontFamilyNameKey)) {
         informationalStringId = DWRITE_INFORMATIONAL_STRING_WIN32_FAMILY_NAMES;
     } else if (CFEqual(nameKey, kCTFontSubFamilyNameKey)) {
         informationalStringId = DWRITE_INFORMATIONAL_STRING_WIN32_SUBFAMILY_NAMES;
     } else if (CFEqual(nameKey, kCTFontStyleNameKey)) {
+        ComPtr<IDWriteFontFace3> dwriteFontFace3;
+        fontFace.As(&dwriteFontFace3);
```
aballway: THROW_IF_FAILED

tests/unittests/CoreGraphics/CGFontTests.mm
```diff
@@ -0,0 +1,117 @@
...
+    // Best hypothesis: Windows has different versions of the same fonts
+    CGFontRef font = CGFontCreateWithFontName(name);
+    CFAutorelease(font);
+    CGRect box = CGFontGetFontBBox(font);
+
+    // Eg: for Arial Bold Italic, origin is correct, but width and height are slightly off from ref plat (but in the right neighborhood)
+    EXPECT_EQ(-1146, box.origin.x);
+    EXPECT_EQ(-771, box.origin.y);
+#if TARGET_OS_WIN32
+    EXPECT_EQ(3993, box.size.width);
```
aballway: nit: EXPECT_NEAR rather than #if/else

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,572 @@
...
+    fontFace.As(&fontFace1);
+
+    DWRITE_CARET_METRICS caretMetrics;
+    fontFace1->GetCaretMetrics(&caretMetrics);
+    if (caretMetrics.slopeRun && caretMetrics.slopeRise) {
+        CGFloat riseOverRun = static_cast<CGFloat>(caretMetrics.slopeRise) / static_cast<CGFloat>(caretMetrics.slopeRun);
+        return (atan(riseOverRun) * 180 / M_PI) - 90.0f; // Degrees returned by this function must be negative
+    } else {
+        // Avoid dividing by 0, skip math for dividing with 0
+        return -0.0f;
```
ms-jihua: unfortunately not :( the math won't work out

---
In reply to: [83675851](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83675851) [](ancestors = 83675851)

Frameworks/CoreGraphics/CGFont.mm
```diff
@@ -682,69 +213,108 @@ CFDictionaryRef CGFontCopyVariations(CGFontRef font) {
...
+int CGFontGetXHeight(CGFontRef font) {
+    // Value for CGFont metrics are specified in 'glyph space units', which appear to be the same as DWrite's 'design units'
+    return font ? __CGFontGetDWriteMetrics(font).xHeight : 0;
+}
+
+/**
+ @Status Interoperable
+*/
+CGRect CGFontGetFontBBox(CGFontRef font) {
+    if (!font) {
```
ms-jihua: yes but this keeps it more in line with the functions that do have to do work

---
In reply to: [83672784](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83672784) [](ancestors = 83672784)

Frameworks/CoreGraphics/CGFont.mm
```diff
@@ -16,615 +16,134 @@
...
-DWORD CGFontGetFontBBox(CGRect* ret, id font);
-
-static IWLazyClassLookup _LazyUIFont("UIFont");
+using namespace Microsoft::WRL;
 
-#include "LoggingNative.h"
+static const wchar_t* g_logTag = L"CGFont";
 
-static const wchar_t* TAG = L"CGFont";
+struct __CGFont {
```
DHowett-MSFT: >__CGFont [](start = 7, length = 8)

Note: this should move to CoreFoundation::CppBase after develop has merged in

Frameworks/CoreGraphics/CGFont.mm
```diff
@@ -16,615 +16,134 @@
...
+    struct __CGFont* font2 = (struct __CGFont*)cf2;
 
-#define USE_CAIRO
-
-void CGFontSetFTFontSize(id uiFont, void* ftFont, float pointSize, float scale) {
-    UIFont* fontInfo = uiFont;
-
-    FT_Set_Char_Size((FT_Face)ftFont, 0, (FT_F26Dot6)(pointSize * 64.0f), (FT_UInt)(72.f * scale), (FT_UInt)(72.f * scale));
+    return (font1->_dwriteFontFace == font2->_dwriteFontFace);
 }
```
DHowett-MSFT: (What if the CGFonts come from different factories?)

---
In reply to: [84346961](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r84346961) [](ancestors = 84346961)

Frameworks/CoreGraphics/CGFont.mm
```diff
@@ -753,46 +323,54 @@ CGFloat CGFontGetStemV(CGFontRef font) {
...
+ @Notes
+*/
+CFTypeID CGFontGetTypeID() {
+    static dispatch_once_t initOnce = 0;
+    dispatch_once(&initOnce,
+                  ^{
+                      __kCGFontTypeID = _CFRuntimeRegisterClass(&__CGFontClass);
+                  });
+    return __kCGFontTypeID;
+}
```
DHowett-MSFT: magic statics will work a treat here, but there's little wrong with dispatch_once

---
In reply to: [83340826](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83340826) [](ancestors = 83340826)

Frameworks/CoreGraphics/DWriteWrapper.mm
DHowett-MSFT: every class is now bridged, so it does work... but it still is underideal.

---
In reply to: [83339711](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83339711) [](ancestors = 83339711)

Frameworks/CoreGraphics/CGFont.mm
ms-jihua: CFArray doesn't check for it in __CFArrayEqual, so I'll follow its lead. My assumption is that CFequal et al check for null before it gets here.

---
In reply to: [84333067](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r84333067) [](ancestors = 84333067)

include/CoreGraphics/CGFont.h
```diff
@@ -19,6 +19,8 @@
 #import <CoreGraphics/CGGeometry.h>
 #import <CoreGraphics/CGDataProvider.h>
 
+typedef const struct __CGFont* CGFontRef;
+
```
ms-jihua: ?

---
In reply to: [84347629](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r84347629) [](ancestors = 84347629)

tests/unittests/CoreGraphics/CGFontTests.mm
ms-jihua: CGFont values are integer-exact, this isn't a rounding error, and having an error margin of over 100 would be pretty silly.

---
In reply to: [84343783](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r84343783) [](ancestors = 84343783)

Frameworks/UIKit/UIApplication.mm
ms-jihua: #1201

---
In reply to: [83327322](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83327322) [](ancestors = 83327322)

include/CoreGraphics/CGFont.h
```diff
@@ -35,11 +37,11 @@ COREGRAPHICS_EXPORT const CFStringRef kCGFontVariationAxisDefaultValue;
 
 COREGRAPHICS_EXPORT void CGFontRelease(CGFontRef font);
 COREGRAPHICS_EXPORT CGFontRef CGFontRetain(CGFontRef font);
-COREGRAPHICS_EXPORT CGFontRef CGFontCreateWithDataProvider(CGDataProviderRef provider);
+COREGRAPHICS_EXPORT CGFontRef CGFontCreateWithDataProvider(CGDataProviderRef provider) STUB_METHOD;
```
ms-jihua: wrote this before we re-evaluated the line, this is pretty non-blocking, I'm going to file a separate issue for this.

---
In reply to: [83305866](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83305866) [](ancestors = 83305866)

build/CoreGraphics/dll/CoreGraphics.vcxproj
```diff
@@ -77,7 +77,7 @@
     </ClCompile>
     <Link>
       <ModuleDefinitionFile>CoreGraphics.def</ModuleDefinitionFile>
-      <AdditionalDependencies>cairo.lib;pixman.lib;freetype.lib;libpng.lib;libz.lib;libjpeg.lib;%(AdditionalDependencies)</AdditionalDependencies>
+      <AdditionalDependencies>cairo.lib;pixman.lib;freetype.lib;libpng.lib;libz.lib;libjpeg.lib;libdispatch.lib;%(AdditionalDependencies)</AdditionalDependencies>
```
ms-jihua: need it for CGFontGetTypeID, in line with CF patterns

---
In reply to: [83330701](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83330701) [](ancestors = 83330701)

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,572 @@
...
+    ComPtr<IDWriteFactory> dwriteFactory;
+    HRESULT hr = _DWriteCreateFactoryInstance(&dwriteFactory);
+    if (FAILED(hr)) {
+        TraceError(TAG, L"Unable to create DWrite factory");
+        return hr;
+    }
+
+    return dwriteFactory
+        ->CreateTextFormat(fontFamilyName, NULL, weight, style, stretch, fontSize, __GetUserDefaultLocaleName().data(), outTextFormat);
+}
```
ms-jihua: ?

---
In reply to: [83338764](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83338764) [](ancestors = 83338764)

Frameworks/CoreGraphics/DWriteWrapper.mm
ms-jihua: yeah, clang forced this :/

---
In reply to: [83338687](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83338687) [](ancestors = 83338687)

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,572 @@
...
+    };
+
+    CFSetApplyFunction(propertyTokens, initPropertyFromToken, &propertyContext);
+}
+
+/**
+ * Creates a standard DWrite factory object
+ */
+HRESULT _DWriteCreateFactoryInstance(IDWriteFactory** outFactory) {
+    return DWriteCreateFactory(DWRITE_FACTORY_TYPE_SHARED, __uuidof(IDWriteFactory), reinterpret_cast<IUnknown**>(outFactory));
```
ms-jihua: Brian made this recommendation that we should leave such optimizations to DWrite itself - according to him it's likely that DWrite already caches its factories and by taking manual control over the factory we're messing with its memory management.

---
In reply to: [83500804](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83500804) [](ancestors = 83500804)

Frameworks/CoreGraphics/DWriteWrapper.mm
ms-jihua: not much to be gained, then - I'd much rather call Autorelease once than .get() every time I need to use something.

---
In reply to: [84348109](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r84348109) [](ancestors = 84348109,83339711)

Frameworks/CoreGraphics/DWriteWrapper.mm
ms-jihua: I mean, if it's only used in the scope of the function...

---
In reply to: [83338922](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83338922) [](ancestors = 83338922)

Frameworks/CoreText/CTFont.mm
ms-jihua: There's only two options, one of which is 'no options', the other is deprecated and was 'exclude synthetic data' [https://developer.apple.com/reference/coretext/ctfonttableoptions?language=objc](https://developer.apple.com/reference/coretext/ctfonttableoptions?language=objc)
I think it's close enough that we should just consider it interoperable.

---
In reply to: [83499156](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83499156) [](ancestors = 83499156)

Frameworks/CoreGraphics/DWriteWrapper.mm
ms-jihua: yeah idk why I did this

---
In reply to: [83338241](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83338241) [](ancestors = 83338241)

tests/unittests/CoreText/CTFontTests.mm
ms-jihua: too much of a stretch at this point, maybe when the need is more apparent.

---
In reply to: [83340634](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83340634) [](ancestors = 83340634)

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,572 @@
...
+        index = 0;
+    }
+
+    // Get the string length.
+    uint32_t length = 0;
+    THROW_IF_FAILED(localizedString->GetStringLength(index, &length));
+
+    // Get the string.
+    vector<wchar_t> wcharString = std::vector<wchar_t>(length + 1);
+    THROW_IF_FAILED(localizedString->GetString(index, wcharString.data(), length + 1));
```
ms-jihua: getting just length causes a throw

---
In reply to: [84335143](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r84335143) [](ancestors = 84335143)

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,572 @@
...
+        // Get the font family.
+        ComPtr<IDWriteFontFamily> fontFamily;
+        THROW_IF_FAILED(fontCollection->GetFontFamily(i, &fontFamily));
+
+        // Get a list of localized strings for the family name.
+        ComPtr<IDWriteLocalizedStrings> familyNames;
+        THROW_IF_FAILED(fontFamily->GetFamilyNames(&familyNames));
+
+        CFStringRef name = _CFStringFromLocalizedString(familyNames.Get());
+        if (CFStringGetLength(name) == 0) {
```
ms-jihua: sure

---
In reply to: [84339235](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r84339235) [](ancestors = 84339235)

Frameworks/CoreGraphics/DWriteWrapper.mm
ms-jihua: forgot to change that when I changed this function to have an outptr, ty

---
In reply to: [83338560](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83338560) [](ancestors = 83338560)

ms-jihua:
>     _needsLayout = TRUE;

out of scope, would rather not make potential merges with aballway harder

---
In reply to: [253677183](https://github.com/Microsoft/WinObjC/pull/1145#issuecomment-253677183) [](ancestors = 253677183)

---
Refers to: Frameworks/UIKit/NSLayoutManager.mm:60 in 04ba630. [](commit_id = 04ba630750d3ea34edeb9c925d456d0a4904b042, deletion_comment = False)

Frameworks/CoreGraphics/DWriteWrapper.mm
ms-jihua: > don't see it being used after
every param after the first in this function is an outparam

> weird set of values to init with
how so?

> why are these pointers?
they need to be to be outparams?

I don't really understand your issues, but to possibly clarify, I'm stuffing them into this struct because there's no such thing as for (obj in collection) in CF and ApplyFunction with some arbitrary context is the next closest thing.

---
In reply to: [83339511](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83339511) [](ancestors = 83339511)

Frameworks/CoreText/CTFont.mm
ms-jihua: ?

---
In reply to: [83340871](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83340871) [](ancestors = 83340871)

Frameworks/CoreText/CTFont.mm
ms-jihua: it reads cleaner to me with this else, I'd like to keep it

---
In reply to: [83339432](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83339432) [](ancestors = 83339432)

Frameworks/CoreGraphics/CGFont.mm
```diff
@@ -16,615 +16,134 @@
...
-    bool ret = CGFontWrapFunc(state, width, sizeOut, freeTypeGetCharWidth, face);
-
-    _CGFontUnlock();
-
-    return ret;
+static void __CGFontDeallocate(CFTypeRef cf) {
+    CGFontRef font = (CGFontRef)cf;
+    // ComPtr needs to be manually destructed, since CFTypes do a memset(0) on dealloc
+    font->_dwriteFontFace.~ComPtr();
 }
```
ms-jihua: ?

---
In reply to: [83340447](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83340447) [](ancestors = 83340447)

Frameworks/CoreGraphics/CGFont.mm
```diff
@@ -16,615 +16,134 @@
...
-        }
-    }
-
-    return ret;
+static void __CGFontInit(CFTypeRef cf) {
+    CGFontRef font = (CGFontRef)cf;
+    // ComPtr needs to be manually new'd, to avoid calling Create functions to an incomplete out-pointer
+    struct __CGFont* mutableFont = const_cast<struct __CGFont*>(font);
+    new (std::addressof(mutableFont->_dwriteFontFace)) ComPtr<IDWriteFontFace>();
 }
```
ms-jihua: ?

---
In reply to: [83340473](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83340473) [](ancestors = 83340473)

Frameworks/CoreGraphics/CGFont.mm
ms-jihua: some other time

---
In reply to: [83501278](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r83501278) [](ancestors = 83501278)

ms-jihua:
>     THROW_NS_IF_FALSE(E_UNEXPECTED, [frame->_lines count] == 1);

we know it's messy/inconsistent, we're planning a cleanup period after the develop merge. for now this is a wontfix though.

---
In reply to: [254749881](https://github.com/Microsoft/WinObjC/pull/1145#issuecomment-254749881) [](ancestors = 254749881)

---
Refers to: Frameworks/CoreText/CTTypesetter.mm:89 in 04ba630. [](commit_id = 04ba630750d3ea34edeb9c925d456d0a4904b042, deletion_comment = False)

Frameworks/CoreGraphics/CGFont.mm
ms-jihua: I'll just do a postscript name comparison instead, those are pretty unique between fonts.

---
In reply to: [84346993](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r84346993) [](ancestors = 84346993,84346961)

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,559 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+// Do not move this block, it has to come first for some reason
```
bbowman: >/ Do not move this block, it has to come first for some reason [](start = 1, length = 62)

can you figure out the reason?

Frameworks/CoreGraphics/CGFont.mm
bbowman: manually calling dtors seems very shady.

---
In reply to: [84415796](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r84415796) [](ancestors = 84415796,83340447)

Frameworks/CoreGraphics/CGFont.mm
bbowman: manually calling placement new for construction seems very shady

---
In reply to: [84415801](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r84415801) [](ancestors = 84415801,83340473)

Frameworks/CoreGraphics/DWriteWrapper.mm
bbowman: use locals? bridge to something more convenient if you want? Do a normal for loop? I don't see the excuse to have poor code here... 

---
In reply to: [84414262](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r84414262) [](ancestors = 84414262,83339511)

Frameworks/CoreGraphics/DWriteWrapper.mm
bbowman: I kind of don't care what you want. They are different semantics and the difference is important. 

---
In reply to: [84348109](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r84348109) [](ancestors = 84348109,83339711)

Frameworks/CoreGraphics/DWriteWrapper.mm
bbowman: sure seems like a ref counting wrapper like unique_cf would make this super easy so you can't forget / screw it up in the future ... and wouldn't require bizarre comments explaining the ref counting ... 

---
In reply to: [84410817](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r84410817) [](ancestors = 84410817,83337852)

Frameworks/CoreGraphics/CGFont.mm
DHowett-MSFT: This is necessary for integration with CF while having C++ members.

---
In reply to: [84554973](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r84554973) [](ancestors = 84554973,84415796,83340447)

Frameworks/CoreGraphics/CGFont.mm
DHowett-MSFT: This is necessary for integration with CF while having C++ members.

---
In reply to: [84555028](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r84555028) [](ancestors = 84555028,84415801,83340473)

Frameworks/CoreText/CTFont.mm
bbowman: context was lost I think but just more correct use of RETURN_* macros to correctly propagate and log errors I think

---
In reply to: [84415484](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r84415484) [](ancestors = 84415484,83340871)

ms-jihua: this function returns a struct and not an hr though, we don't have a macro for that??

---
In reply to: [84556247](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r84556247) [](ancestors = 84556247,84415484,83340871)

Frameworks/CoreText/CTFont.mm
```diff
@@ -852,9 +903,8 @@ CFDataRef CTFontCopyTable(CTFontRef font, CTFontTableTag table, CTFontTableOptio
...
 CFTypeID CTFontGetTypeID() {
     static dispatch_once_t initOnce = 0;
-    dispatch_once(&initOnce,
-                  ^{
-                      __kCTFontTypeID = _CFRuntimeRegisterClass(&__CTFontClass);
-                  });
+    dispatch_once(&initOnce, ^{
+        __kCTFontTypeID = _CFRuntimeRegisterClass(&__CTFontClass);
+    });
     return __kCTFontTypeID;
```
ms-jihua: ?

---
In reply to: [84556024](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r84556024) [](ancestors = 84556024)

Frameworks/CoreGraphics/DWriteWrapper.mm
bbowman: usually points to a "curious" design decision in the first place. In this case you have a void return function with three out parameters and then a bunch of weird glue code to make that work. How about defining the struct outside the function and using it as a return type? Then you don't need to worry about the modification contracts with the caller's data.

---
In reply to: [84400714](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r84400714) [](ancestors = 84400714,83338922)

ms-jihua: why is it poor? this is the intended usage of CF____ApplyFunction()

CFSet doesn't support a normal for loop as it has nothing like a "get at index". I can bridge to something more convenient, but since an eventual goal is to remove dependencies on Foundation, which would revert the code back to this state, I'd rather just leave it as is?

---
In reply to: [84555482](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r84555482) [](ancestors = 84555482,84414262,83339511)

Frameworks/CoreText/CTFont.mm
bbowman: yeah I don't think it was this function I meant to comment on. Most of my comments seem to be misplaced / recalled / otherwise not as I intended.... 

---
In reply to: [84556614](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r84556614) [](ancestors = 84556614,84556247,84415484,83340871)

bbowman: in general it looks like you cleaned up a lot so probably ok

---
In reply to: [84557133](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r84557133) [](ancestors = 84557133,84556614,84556247,84415484,83340871)

Frameworks/CoreGraphics/DWriteWrapper.mm
ms-jihua: filed #1207 to look into it at another time.

---
In reply to: [84554837](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r84554837) [](ancestors = 84554837)

Frameworks/CoreGraphics/CGFont.mm
```diff
@@ -16,615 +16,129 @@
...
-}
-
-/**
- @Status Interoperable
-*/
-CGFontRef CGFontCreateWithDataProvider(CGDataProviderRef cgDataProvider) {
-    CFRetain(cgDataProvider);
-    return (CGFontRef)[[_LazyUIFont fontWithData:(NSData*)cgDataProvider] retain];
-}
+static CFTypeID __kCGFontTypeID = _CFRuntimeRegisterClass(&__CGFontClass);
```
DHowett-MSFT: I think we should avoid registering these things on library load, _but_:
if you move this entire thing into CGFontGetTypeID, it will automatically only happen once, thread-safely, the first time somebody uses CGFont.

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,525 @@
...
+        CFStringRef name = _CFStringFromLocalizedString(familyNames.Get());
+        if (CFStringGetLength(name) == 0) {
+            TraceError(TAG, L"Failed to convert the localized string to wide string.");
+            continue;
+        }
+
+        CFArrayAppendValue(fontFamilyNames.get(), name);
+    }
+
+    return fontFamilyNames.release();
```
DHowett-MSFT: > fontFamilyNames.release(); [](start = 10, length = 27)

I'm a little worried that a function named with the `Get` convention is returning a +1 object, when `Get` specifies that it should return a +0/unowned reference.

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,525 @@
...
+    std::vector<UniChar> unicharFamilyName(familyNameLength + 1);
+    CFStringGetCharacters(familyName, CFRangeMake(0, familyNameLength), unicharFamilyName.data());
+
+    size_t index = 0;
+    BOOL exists = false;
+
+    RETURN_NULL_IF_FAILED(fontCollection->FindFamilyName(reinterpret_cast<wchar_t*>(unicharFamilyName.data()), &index, &exists));
+    if (!exists) {
+        TraceError(TAG, L"Failed to find the font family name.");
+        return fontNames.release();
```
DHowett-MSFT: >Names.release(); [](start = 19, length = 16)

Same as abov

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,525 @@
...
+    // Copy the font table's binary data to a CFDataRef
+    woc::unique_cf<CFDataRef> ret(CFDataCreate(kCFAllocatorDefault, reinterpret_cast<const byte*>(tableData), tableSize));
+
+    // As part of the contact for TryGetFontTable, tableContext must always be released via ReleaseFontTable
+    if (tableContext) {
+        fontFace->ReleaseFontTable(tableContext);
+    }
+
+    return ret.release();
+}
```
DHowett-MSFT: This one, however, is great. :)

Frameworks/CoreText/CTFont.mm
```diff
@@ -156,6 +156,8 @@ static void __CTFontDeallocate(CFTypeRef cf) {
                                               NULL, //
                                               __CTFontCopyDescription };
 
+static CFTypeID __kCTFontTypeID = _CFRuntimeRegisterClass(&__CTFontClass);
+
```
DHowett-MSFT: Same as in CGFont

DHowett-MSFT:
:clock1:

DHowett-MSFT:
:shipit:

msft-Jeyaram:
>     THROW_NS_IF_FALSE(E_UNEXPECTED, [frame->_lines count] == 1);

A task items to keep track of it?

---
In reply to: [255288503](https://github.com/Microsoft/WinObjC/pull/1145#issuecomment-255288503) [](ancestors = 255288503,254749881)

---
Refers to: Frameworks/CoreText/CTTypesetter.mm:89 in 43b6e0a. [](commit_id = 43b6e0acff20544c2264bfebcf2e94c982bb3ecd, deletion_comment = False)

msft-Jeyaram:
>     THROW_NS_IF_FALSE(E_UNEXPECTED, [frame->_lines count] == 1);

Also given we are adding code, why don't we just set the standard  now to follow through. It seems weird to add code here, then come back to clean it up after we accumulate so much of it throughout the project.

---
In reply to: [255799568](https://github.com/Microsoft/WinObjC/pull/1145#issuecomment-255799568) [](ancestors = 255799568,255288503,254749881)

---
Refers to: Frameworks/CoreText/CTTypesetter.mm:89 in 43b6e0a. [](commit_id = 43b6e0acff20544c2264bfebcf2e94c982bb3ecd, deletion_comment = False)

msft-Jeyaram:
>         ret = (array != nullptr) ? true : false;

can be removed.
Given the statement would produce the value.

---
Refers to: Frameworks/CoreText/DWriteWrapper_CoreText.mm:64 in 43b6e0a. [](commit_id = 43b6e0acff20544c2264bfebcf2e94c982bb3ecd, deletion_comment = False)

msft-Jeyaram:
>         ret = (array != nullptr) ? true : false;

Although it does make it seem much clear.

---
In reply to: [255800842](https://github.com/Microsoft/WinObjC/pull/1145#issuecomment-255800842) [](ancestors = 255800842)

---
Refers to: Frameworks/CoreText/DWriteWrapper.mm:64 in 43b6e0a. [](commit_id = 43b6e0acff20544c2264bfebcf2e94c982bb3ecd, deletion_comment = False)

ms-jihua:
>     THROW_NS_IF_FALSE(E_UNEXPECTED, [frame->_lines count] == 1);

#1212

---
In reply to: [255800252](https://github.com/Microsoft/WinObjC/pull/1145#issuecomment-255800252) [](ancestors = 255800252,255799568,255288503,254749881)

---
Refers to: Frameworks/CoreText/CTTypesetter.mm:89 in 43b6e0a. [](commit_id = 43b6e0acff20544c2264bfebcf2e94c982bb3ecd, deletion_comment = False)

Frameworks/CoreGraphics/DWriteWrapper.mm
bbowman: While true, the "correct" fix here would be to return a +1 item with a different name because CFStyle things should not obey the same idea of +0 items. Those don't really exist in CF land. Unowned returns exist but that semantic is different that what is going on here. CFAutorelease *really* shouldn't be used if possible. since these are private functions I'd prefer a unique_cf return to correctly describe the *transfer* of ownership to the caller which is the pattern that is actually be attempted here with CFAutorelease being used to make the return work with a raw * which is just a silly Objective-C ism due to the bolt on nature of extra meaning to a * without an ability to in language convey that. Relying on naming conventions to express memory management is an error prone system that I'd like to not rely on ... 

---
In reply to: [84563998](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r84563998) [](ancestors = 84563998)

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,525 @@
...
+
+/**
+ * Helper method that maps a font name to the name of its family.
+ *
+ * Note: This function currently uses a cache, meaning that fonts installed during runtime will not be reflected
+ */
+CFStringRef _DWriteGetFamilyNameForFontName(CFStringRef fontName) {
+    woc::unique_cf<CFLocaleRef> locale(CFLocaleCopyCurrent());
+
+    static CFDictionaryRef fontToFamilyMap = [&locale]() {
```
bbowman: >&locale [](start = 46, length = 7)

Seems weird to have a local variable be passed by ref to a one time init static lambda. My expectation would be that the locale would be static too if some static state needed it otherwise it would potentially get out of sync?

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,525 @@
...
+    BOOL fontFamilyExists;
+
+    RETURN_IF_FAILED(systemFontCollection->FindFamilyName(reinterpret_cast<const wchar_t*>(unicharFamilyName.data()),
+                                                          &fontFamilyIndex,
+                                                          &fontFamilyExists));
+
+    if (!fontFamilyExists) {
+        TraceError(TAG, L"Unable to find font family \"%ws\"", unicharFamilyName.data());
+        return E_INVALIDARG;
+    }
```
bbowman: return macro?

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,525 @@
...
+CFDataRef _DWriteFontCopyTable(const ComPtr<IDWriteFontFace>& fontFace, uint32_t tag) {
+    const void* tableData;
+    uint32_t tableSize;
+    void* tableContext;
+    BOOL exists;
+
+    if (FAILED(fontFace->TryGetFontTable(tag, &tableData, &tableSize, &tableContext, &exists)) || !exists) {
+        TraceError(TAG, L"Unable to get font table for tag %u", tag);
+        return nullptr;
+    }
```
bbowman: return macro?

Frameworks/CoreGraphics/DWriteWrapper.mm
ms-jihua: changed the names to Copy and return +1s

---
In reply to: [84745555](https://github.com/Microsoft/WinObjC/pull/1145#discussion_r84745555) [](ancestors = 84745555,84563998)

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -0,0 +1,519 @@
...
+                               DWRITE_FONT_STRETCH_NORMAL,
+                               DWRITE_FONT_STYLE_NORMAL,
+                               _DWriteGetFamilyNameForFontName(fontName) };
+
+    // Relationship of family name -> font name not always consistent
+    // Usually, properties are added to the end (eg: Arial -> Arial Narrow Bold)
+    // However, this is not always the case (eg: Eras ITC -> Eras Bold ITC)
+    // In addition, some fonts with properties are occasionally placed into their own family (eg: Segoe WP SemiLight -> Segoe WP SemiLight)
+    // Try to be more prudent about these edge cases, by looking only at the difference between the font name and family name
+    woc::unique_cf<CFArrayRef> fontNameTokens(CFStringCreateArrayBySeparatingStrings(kCFAllocatorDefault, fontName, CFSTR(" ")));
```
rajsesh-msft: This would break for a font with a space in its name - "Times New Roman", "Courier New"

PR #1147, opened 2016-10-13
Enable additional middleware scenarios

bviglietta:
- Include AudioToolbox.dll in apps using middleware
- Allow UIView animation with spring
- Streamline app protocol activation via Safari view controller

### DIFF
--- a/Frameworks/SafariServices/SFSafariWebViewController.mm
+++ b/Frameworks/SafariServices/SFSafariWebViewController.mm
@@ -28,6 +28,27 @@ static const wchar_t* TAG = L"SFSafariWebViewController";
 static const char* SAFARI_USER_AGENT =
     "Mozilla/5.0 (iPad; CPU OS 6_0 like Mac OS X) AppleWebKit/536.26 (KHTML, like Gecko) Version/6.0 Mobile/10A5355d Safari/8536.25";
 
+static bool appCanOpenURL(NSURL* url) {
+    // Check the app's main bundle to see whether this URL conforms to a scheme
+    // that this app can open
+    NSDictionary* info = [[NSBundle mainBundle] infoDictionary];
+    NSArray<NSDictionary*>* urlTypes = info[@"CFBundleURLTypes"];
+
+    NSString* targetScheme = url.scheme;
+
+    for (NSDictionary* urlType in urlTypes) {
+        NSArray<NSString*>* urlSchemes = urlType[@"CFBundleURLSchemes"];
+
+        for (NSString* scheme in urlSchemes) {
+            if ([scheme isEqualToString:targetScheme]) {
+                return true;
+            }
+        }
+    }
+
+    return false;
+}
+
 // Delegate for the UIWebView that underlies a SFSafariWebViewController
 @interface _SFWebViewDelegate : NSObject <UIWebViewDelegate>
 @property (weak, nonatomic) SFSafariWebViewController* viewController;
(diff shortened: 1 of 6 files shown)


bviglietta:

bbowman:
:shipit:

Frameworks/SafariServices/SFSafariWebViewController.mm
```diff
@@ -177,4 +207,8 @@ - (void)_cancel {
     [self dismissViewControllerAnimated:NO completion:nil];
 }
 
+- (void)dealloc {
+    _webView.delegate = nil;
```
bviglietta: No, if it's null, this will be a no-op.

---
In reply to: [83332466](https://github.com/Microsoft/WinObjC/pull/1147#discussion_r83332466) [](ancestors = 83332466)

DHowett-MSFT: Since we own our webview, doesn't it not matter that its delegate is being nil'd out? It won't have a word to say about it in just a few more seconds.

bviglietta: True, but the UIWebView docs specify very sternly that the delegate field must be reset to nil before deallocating the UIWebView, so maybe I should just comply?

In looking at this question, I did discover that UIWebView has a retain cycle involving its callbacks, so that's something that needs to be fixed.

---
In reply to: [83337136](https://github.com/Microsoft/WinObjC/pull/1147#discussion_r83337136) [](ancestors = 83337136)

Frameworks/SafariServices/SFSafariWebViewController.mm
```diff
@@ -49,13 +70,24 @@ - (BOOL)webView:(UIWebView*)webView
     shouldStartLoadWithRequest:(NSURLRequest*)request
                 navigationType:(UIWebViewNavigationType)navigationType {
     NSURL* url = request.URL;
+    BOOL openWithAppDelegate = NO;
 
     // Check whether this request is for the redirect URL that we're watching out for
     // (modulo the query string and fragment, which are allowed to vary)
-
     if ([url.scheme isEqualToString:_redirectUrl.scheme] && [url.host isEqualToString:_redirectUrl.host] &&
```
aballway: nit: instead of having a flag variable recombine the ifs

bviglietta: I could, but I feel the code would be less clear that way.

---
In reply to: [83669085](https://github.com/Microsoft/WinObjC/pull/1147#discussion_r83669085) [](ancestors = 83669085)

Frameworks/SafariServices/SFSafariWebViewController.mm
bviglietta: [UIWebView issue addressed in second commit.]

---
In reply to: [83342864](https://github.com/Microsoft/WinObjC/pull/1147#discussion_r83342864) [](ancestors = 83342864,83337136)

Frameworks/SafariServices/SFSafariWebViewController.mm
```diff
@@ -49,13 +70,24 @@ - (BOOL)webView:(UIWebView*)webView
...
         [url.path isEqualToString:_redirectUrl.path]) {
-        // Let the client know we received the redirect URL
+        openWithAppDelegate = YES;
+    }
+
+    // Check whether the URL can be opened directly by this application, in which case we'll send it
+    // straight to the application delegate.
+    // Asking the WebView to open such a URL would work, since it would send an activation request
+    // to this app, but that would cause an extraneous confirmation dialog from Windows.
+    if (appCanOpenURL(url)) {
```
bviglietta: First question: no, openWithAppDelegate should be set to YES if either condition is true.

Second question: That's a good point. We can skip the call to appCanOpenURL if openWithAppDelegate is already YES.

---
In reply to: [83742017](https://github.com/Microsoft/WinObjC/pull/1147#discussion_r83742017) [](ancestors = 83742017)

pradipd: Thanks for the clarification.
I'm fine leaving as is or changing it to skip the appCanOpenURL call.  
either way is fine.

---
In reply to: [83745451](https://github.com/Microsoft/WinObjC/pull/1147#discussion_r83745451) [](ancestors = 83745451,83742017)

pradipd:
:shipit:

PR #1151, opened 2016-10-14
Memory leak fix in UIButton

amagams:


Frameworks/UIKit/UIButton.mm
```diff
@@ -182,7 +182,7 @@ - (void)_UIButton_initInternal:(WXFrameworkElement*)xamlElement {
     // Set the XAML element's name so it's easily found in the VS live tree viewer
     [_xamlButton setName:[NSString stringWithUTF8String:object_getClassName(self)]];
 
-    __weak UIButton* weakSelf = self;
+    __block UIButton* weakSelf = self;
```
oliversa-msft: UIButton is not ARC-enabled. File a bug so we can move it to ARC-enabled and then the "weak" decoration will work.

Frameworks/UIKit/UIButton.mm
```diff
@@ -764,6 +762,7 @@ - (void)dealloc {
     }
 
     _xamlButton = nil;
+    [super dealloc];
```
oliversa-msft: We can remove if ARC-enabled?

Frameworks/UIKit/UIButton.mm
amagams: bug filed.

Frameworks/UIKit.Xaml/Button.xaml.cpp
```diff
@@ -161,7 +161,9 @@ UIKIT_XAML_EXPORT void XamlRemoveLayoutEvent(const ComPtr<IInspectable>& inspect
 // This will be used by autolayout to get the intrinsic content size when XAML has done calculating the elements final desired size.
 Windows::Foundation::Size Button::ArrangeOverride(Windows::Foundation::Size finalSize) {
     __super::ArrangeOverride(finalSize);
-    _layoutHook->Invoke(nullptr, nullptr);
+	if (_layoutHook) {
+		_layoutHook->Invoke(nullptr, nullptr);
+	}
```
jaredhms: tabs

PR #1153, opened 2016-10-14
Implement CoreFoundation::CppBase, a C++ helper for CF. Switch CGContext to use it

DHowett-MSFT:
The file should likely be commented.

PR #1156, opened 2016-10-14
Add CFCppBase, a helper class for implementing C++ CF types.

DHowett-MSFT:
This is `CFCppBase` from #1153 (signed off, build status pending), but applied to **develop**. Build status will be tracked in #1153.

This commit adds only the header file, _no consumers_.


rajsesh-msft:
:shipit:

PR #1160, opened 2016-10-17
Do not use telemetry functions in fileutils.cpp, since not all binaries that depend on the file support telemetry

asimonov-msft:


andrejo-msft:
:shipit:

asimonov-msft:
This change does the affect the build, so skipping it.

PR #1163, opened 2016-10-17
Updating cert with expiration date for another year.

pradipd:
1. Updated the cert using VS directions under "Renew a certification" at  https://msdn.microsoft.com/en-us/library/br230260.aspx. The cert is valid for another year.  
2. Updated VSImporter to use  publisher="CN=islandwood" instead of publisher="islandwood".  I think this is due to changes in latest SDK.  I'm not sure how to check if this is backwards compatible as I don't have a machine without 14393 SDK.


pradipd:

msvc/vsimporter-templates/WinStore10-App/Package-native.appxmanifest
```diff
@@ -8,7 +8,7 @@
 
   <Identity
     Name="$guid9$"
-    Publisher="islandwood"
+    Publisher="CN=islandwood"
```
pradipd: >CN=islandwood [](start = 15, length = 13)

I'm nervous about backwards compat.  Does anyone have a machine without 14393?
How can we validate we won't break external developers who haven't installed 14393 SDK.

rajsesh-msft: CN= format has always worked in the past I can see several examples of this in our internal codebase, so this should pose any compat issues.

---
In reply to: [83706840](https://github.com/Microsoft/WinObjC/pull/1163#discussion_r83706840) [](ancestors = 83706840)

rajsesh-msft:
:shipit:

pradipd:
Looks like Oct, 14-17, 2016 was a momentous date.  Since I started this change, many of the samples started failing because the keys for their apps expired.
I went back and update all the keys to use the one from msvc\vsiimporter-templates\WinStore10-App

I also updated all the *.appxmanifest.  I'm not sure why they don't show up in the pull request though.

asimonov-msft:
You should be able to remove the following directories, as they are obsolete:
samples\WOCCatalog\WOCCatalog.vsimporter\WOCCatalog-Headers-WinStore10
samples\XAMLCatalog\XAMLCatalog.vsimporter\XAMLCatalog-Headers-WinStore10

---
In reply to: [254308466](https://github.com/Microsoft/WinObjC/pull/1163#issuecomment-254308466) [](ancestors = 254308466)

asimonov-msft:
:shipit:

edvv:
FYI: I got bit by this expiring cert so had to set the date back two weeks; but then the AppManifest.xml showed with an invalid schema due to the lack of CN= pattern, which is strange because it worked before. I manually edited the pattern to get it to complete the build.

Short story: It would be nice to fold this change into a new release ASAP (release 42).

rajsesh-msft:
@edvv, we will have this in the next release, which would be in early November.

PR #1168, opened 2016-10-18
[UIKit] [Gestures] Add support for cancelsTouchesInView = NO, add Gestures tab to WOCCatalog

davelamb:
Changes of note:

- touchesMoved now only gets called with one touch at a time
- hitTesting only happens on UITouchPhaseBegan
- UITapGestureRecognizer now works on maxTouches, since we never expect more than one UITouch at once
- WOCCatalog/Gestures tab draws to a backing CGImage every time 128 line segments are created. Seems performant enough, but we log a lot of CGLines in the process, slowing everything down. (May remove in a later PR)
- UILabel's background remains clear.

Testing:

Created new WOCCatalog tab, which sets all gesture recognizers' cancelsTouchesInView to NO, and displays trails for all touches and gestures. The rest is ad hoc.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1168)

### DIFF
--- a/Frameworks/UIKit/UILongPressGestureRecognizer.mm
+++ b/Frameworks/UIKit/UILongPressGestureRecognizer.mm
@@ -251,6 +251,26 @@ static TrackedTouch* findTouch(UITouch* touch, std::vector<TrackedTouch>& touche
     [self reset];
 }
 
+/**
+ @Status Interoperable
+*/
+- (CGPoint)locationInView:(UIView*)view {
+    if (_trackedTouches.size() > 0) {
+        CGPoint averagePoint = { 0, 0 };
+        for (size_t i = 0; i < _trackedTouches.size(); i++) {
+            CGPoint touchLocation = [_trackedTouches[i].touch locationInView:view];
+            averagePoint.x += touchLocation.x;
+            averagePoint.y += touchLocation.y;
+        }
+        averagePoint.x /= _trackedTouches.size();
+        averagePoint.y /= _trackedTouches.size();
+        return averagePoint;
+    }
+
+    CGPoint ret = { 0, 0 };
+    return ret;
+}
+
 - (void)_commonInit {
     self->minimumPressDuration = 0.5f;
     self->numberOfTapsRequired = 0;
--- a/Frameworks/UIKit/UILongPressGestureRecognizer.mm
+++ b/Frameworks/UIKit/UILongPressGestureRecognizer.mm
@@ -257,8 +257,8 @@ static TrackedTouch* findTouch(UITouch* touch, std::vector<TrackedTouch>& touche
 - (CGPoint)locationInView:(UIView*)view {
     if (_trackedTouches.size() > 0) {
         CGPoint averagePoint = { 0, 0 };
-        for (size_t i = 0; i < _trackedTouches.size(); i++) {
-            CGPoint touchLocation = [_trackedTouches[i].touch locationInView:view];
+        for (TrackedTouch& trackedTouch : _trackedTouches) {
+            CGPoint touchLocation = [trackedTouch.touch locationInView:view];
             averagePoint.x += touchLocation.x;
             averagePoint.y += touchLocation.y;
         }
@@ -267,8 +267,7 @@ static TrackedTouch* findTouch(UITouch* touch, std::vector<TrackedTouch>& touche
         return averagePoint;
     }
 
-    CGPoint ret = { 0, 0 };
-    return ret;
+    return { 0, 0 };
 }
 
 - (void)_commonInit {
--- a/Frameworks/UIKit/UITapGestureRecognizer.mm
+++ b/Frameworks/UIKit/UITapGestureRecognizer.mm
@@ -256,12 +256,12 @@ static void resetSavedTouches(UITapGestureRecognizer* self) {
     }
 
     id allTouches = [event allTouches];
-    unsigned count = [touches count];
 
-    if (count > _numberOfTouchesRequired || EbrGetMediaTime() - _tapTime > 0.75f) {
-        TraceVerbose(TAG, L"too many touches, numberOfTouchesRequired=%d, count=%d", _numberOfTouchesRequired, count);
+    // We may have to revisit this logic, since the UITouch tap count should drive it.
+    if (_maxTouches > _numberOfTouchesRequired || EbrGetMediaTime() - _tapTime > 0.75f) {
+        TraceVerbose(TAG, L"too many touches, numberOfTouchesRequired=%d, _maxTouches=%d", _numberOfTouchesRequired, _maxTouches);
         _state = UIGestureRecognizerStateFailed;
-    } else if (count == _numberOfTouchesRequired) {
+    } else if (_maxTouches == _numberOfTouchesRequired) {
         bool success = true;
 
         // get max numberofTapsRequired in current Tap gesture list.
--- a/Frameworks/UIKit/UIView.mm
+++ b/Frameworks/UIKit/UIView.mm
@@ -135,7 +135,7 @@ BOOL g_resetAllTrackingGestures = TRUE;
         g_currentlyTrackingGesturesList = [NSMutableArray new];
     }
 
-    bool handled = false;
+    BOOL shouldCancelTouches = false;
 
     UIView* views[128];
     int viewDepth = 0;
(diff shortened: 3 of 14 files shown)


rajsesh-msft:

Frameworks/UIKit/UILongPressGestureRecognizer.mm
```diff
@@ -251,6 +251,26 @@ - (void)touchesCancelled:(NSSet*)touches withEvent:(UIEvent*)event {
     [self reset];
 }
 
+/**
+ @Status Interoperable
+*/
+- (CGPoint)locationInView:(UIView*)view {
+    if (_trackedTouches.size() > 0) {
+        CGPoint averagePoint = { 0, 0 };
+        for (size_t i = 0; i < _trackedTouches.size(); i++) {
```
jaredhms: >size_t i = 0; i < _trackedTouches.size(); i++ [](start = 13, length = 45)

Should use range-based for;

for (TrackedTouch& trackedTouch : _trackedTouches) { ...

Frameworks/UIKit/UILongPressGestureRecognizer.mm
```diff
@@ -251,6 +251,26 @@ - (void)touchesCancelled:(NSSet*)touches withEvent:(UIEvent*)event {
...
+            averagePoint.x += touchLocation.x;
+            averagePoint.y += touchLocation.y;
+        }
+        averagePoint.x /= _trackedTouches.size();
+        averagePoint.y /= _trackedTouches.size();
+        return averagePoint;
+    }
+
+    CGPoint ret = { 0, 0 };
+    return ret;
```
jaredhms: nit: just 'return { 0, 0 };' ?

Frameworks/UIKit/UITapGestureRecognizer.mm
```diff
@@ -256,12 +256,12 @@ - (void)touchesEnded:(NSSet*)touches withEvent:(UIEvent*)event {
     }
 
     id allTouches = [event allTouches];
-    unsigned count = [touches count];
 
-    if (count > _numberOfTouchesRequired || EbrGetMediaTime() - _tapTime > 0.75f) {
-        TraceVerbose(TAG, L"too many touches, numberOfTouchesRequired=%d, count=%d", _numberOfTouchesRequired, count);
+    // We may have to revisit this logic, since the UITouch tap count should drive it.
```
jaredhms: > // We may have to revisit this logic, since the UITouch tap count should drive it. [](start = 3, length = 83)

Why not leverage the tap count now?

jaredhms:
@yiyang-msft is added to the review. #Closed

jaredhms:

jaredhms:
> static const bool DEBUG_TOUCHES_LIGHT = DEBUG_TOUCHES || true;

should we disable this now?

---
Refers to: Frameworks/UIKit/UIView.mm:53 in ec039ec. [](commit_id = ec039ecb4f768c2173c6b568aea8632b734d5c84, deletion_comment = False)

jaredhms:
> static const bool DEBUG_HIT_TESTING_LIGHT = DEBUG_HIT_TESTING || true;

should we disable this now?

---
Refers to: Frameworks/UIKit/UIView.mm:55 in ec039ec. [](commit_id = ec039ecb4f768c2173c6b568aea8632b734d5c84, deletion_comment = False)

jaredhms:
>                 // TODO: This is how it worked before; is that the expected behavior?

nit: we can probably delete this comment now too?

---
Refers to: Frameworks/UIKit/UIView.mm:490 in ec039ec. [](commit_id = ec039ecb4f768c2173c6b568aea8632b734d5c84, deletion_comment = False)

jaredhms:
>                 // TODO: This is how it worked before; is that the expected behavior?

nit: we can probably delete this comment now too?

---
Refers to: Frameworks/UIKit/UIView.mm:524 in ec039ec. [](commit_id = ec039ecb4f768c2173c6b568aea8632b734d5c84, deletion_comment = False)

Frameworks/UIKit/UIView.mm
```diff
@@ -409,6 +408,13 @@ - (UITouchPhase)_processPointerEvent:(WUXIPointerRoutedEventArgs*)pointerEventAr
             // Add this touch to the set of all current touches in the app
             [s_allTouches addObject:touchPoint.touch];
 
+            if (DEBUG_HIT_TESTING_LIGHT) {
+                TraceVerbose(TAG, L"Hit testing view %hs(0x%p).", object_getClassName(self), self);
+            }
+
+            // Update the hit-tested view. (Only one owner per UITouch lifetime)
+            touchPoint.touch->_view = [self _doHitTest:touchPoint.touch allTouches:s_allTouches];
```
jaredhms: we shouldn't have to do another hit test here, as Xaml already chose the hit-test view..  what scenario is this covering?

jaredhms: ah - i see you just moved it up from below.  maybe we can get rid of that call now altogether, or at least wrap it in the if (DEBUG_HIT_TESTING_LIGHT) block and get rid of the 'enable legacy hit-testing' path?

---
In reply to: [84356602](https://github.com/Microsoft/WinObjC/pull/1168#discussion_r84356602) [](ancestors = 84356602)

jaredhms:
@tadam-msft is added to the review. #Closed

davelamb:
Accidentally pushed the wrong branch. Reopen.

Frameworks/UIKit/UITapGestureRecognizer.mm
davelamb: I'd rather not mess with it at this time. I'm not convinced our tapCount implementation is completely interoperable, and I'd rather we address the concerns of the task before investing in a reimplementation of something that may indeed be rewritten with WinRT gestures.

jaredhms:
> static const bool DEBUG_TOUCHES_LIGHT = DEBUG_TOUCHES || true;

?


---
In reply to: [255200660](https://github.com/Microsoft/WinObjC/pull/1168#issuecomment-255200660) [](ancestors = 255200660)

---
Refers to: Frameworks/UIKit/UIView.mm:53 in 6d4531d. [](commit_id = 6d4531d847d7244696c1a347002e33b2dd17e1c4, deletion_comment = False)

jaredhms:
> static const bool DEBUG_HIT_TESTING_LIGHT = DEBUG_HIT_TESTING || true;

?


---
In reply to: [255200684](https://github.com/Microsoft/WinObjC/pull/1168#issuecomment-255200684) [](ancestors = 255200684)

---
Refers to: Frameworks/UIKit/UIView.mm:55 in 6d4531d. [](commit_id = 6d4531d847d7244696c1a347002e33b2dd17e1c4, deletion_comment = False)

Frameworks/UIKit/UIView.mm
jaredhms: ?


---
In reply to: [84356876](https://github.com/Microsoft/WinObjC/pull/1168#discussion_r84356876) [](ancestors = 84356876,84356602)

davelamb:
*[Frameworks/UIKit/UIView.mm, line 416 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1168#-KUxHC769v8LgmXL6aZp-r1-416:-KUyHz3CstKyNYOrZWkV:b-uw0c4m) ([raw file](https://github.com/microsoft/winobjc/blob/ec039ecb4f768c2173c6b568aea8632b734d5c84/Frameworks/UIKit/UIView.mm#L416)):*
<details><summary><i>Previously, jaredhms (Jared) wrote…</i></summary>
> ?
> 
> 
> 
> 
> 
> ---
> 
> In reply to: [84356876](https://github.com/Microsoft/WinObjC/pull/1168#discussion_r84356876) [](ancestors = 84356876,84356602)
</details>
I think it's still a useful tool for debugging. Hit testing isn't always obvious, especially when authoring new controls with xaml behind. I'm not really against removing the useLegacyHitTesting option in display modes, but I have found a few bugs due to the output of hit testing discrepancy. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/winobjc/1168)*

Frameworks/UIKit/UIView.mm
jaredhms: ok, maybe we can track the removal as a separate issue

---
In reply to: [85023181](https://github.com/Microsoft/WinObjC/pull/1168#discussion_r85023181) [](ancestors = 85023181,84356876,84356602)

jaredhms:
:shipit:

PR #1169, opened 2016-10-18
Do not roam NSUserDefaults (or CFPreferences)

### LINKED ISSUE #1167: NSUserDefaults writes to CSIDL_APPDATA folder, which roams on windows
The expectation is that NSUserDefaults does not roam.  "Roaming" settings are usually handled through NSUbiquitousKeyValueStore, which we don't support right now.

A trade off would be to:

1. NSUserDefaults to always go to local data.
2. For roaming the settings, use CFPreferences directly and specify kCFPreferencesAnyHost for hostName.   
All other host names will be local only.

rajsesh-msft:
Save NSUserDefaults to local app data, not roaming app data.
    App preferences writes with kCFPreferencesAnyHost, which means we cannot
    use the host parameter to differentiate roaming vs local.
    
    The correct way to roam would be through ubiquitous store (which would
    have to be covered separately based on need).
    
    Fix #1167

### DIFF
--- a/Frameworks/CoreFoundation/Base.subproj/CFPlatform.c
+++ b/Frameworks/CoreFoundation/Base.subproj/CFPlatform.c
@@ -495,7 +495,7 @@ CF_INLINE CFIndex strlen_UniChar(const UniChar* p) {
  * or if the bundle name cannot be obtained:
  *     <nFolder_path>\Apple Computer\
  * where nFolder_path is obtained by calling SHGetFolderPath with nFolder
- * (CSIDL_LOCAL_APPDATA).
+ * (for example, with CSIDL_APPDATA or CSIDL_LOCAL_APPDATA).
  *
  * The CFMutableStringRef result must be released by the caller.
  *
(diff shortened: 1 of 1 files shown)


rajsesh-msft:

rajsesh-msft:

Frameworks/CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -505,8 +505,8 @@ CF_EXPORT CFMutableStringRef _CFCreateApplicationRepositoryPath(CFAllocatorRef a
     CFMutableStringRef result = NULL;
     CFStringRef str = NULL;
 
-    // WINOBJC: make sure that nFolder is CSIDL_APPDATA or CSIDL_LOCAL_APPDATA and return the app data folder for the app.
-    Wrappers::HString path = GetAppDataPath(nFolder == CSIDL_LOCAL_APPDATA);
+    //WinOBJC- In Reference platform, CFPreferences does not roam, so ignore nFolder
```
DHowett-MSFT: Nit: keep the style of the above comment (`WINOBJC:`)

DHowett-MSFT:
:shipit:

PR #1175, opened 2016-10-18
Consolidate uses of Ebr file functions.

bbowman:
Consolidate uses of Ebr file functions to only where absolutely needed. This helps abstract any weird semantics we have to only a few locations and uses well defined APIs for foundation constructs like NSData and NSStream where possible.

### DIFF
--- a/Frameworks/AudioToolbox/AudioFile.mm
+++ b/Frameworks/AudioToolbox/AudioFile.mm
@@ -20,7 +20,6 @@
 #import <algorithm>
 
 #import "Starboard.h"
-#import "Platform/EbrPlatform.h"
 #import "Etc.h"
 #include "CAFDecoder.h"
 #include "LoggingNative.h"
@@ -74,9 +73,9 @@ public:
 
 class AudioFileOGG : public OpaqueAudioFileID {
     stb_vorbis* _handle;
-    EbrFile* _file;
+    StrongId<NSData> _data;
 
-    AudioFileOGG(stb_vorbis* handle, EbrFile* file) : _handle(handle), _file(file) {
+    AudioFileOGG(stb_vorbis* handle, NSData* data) : _handle(handle), _data(data) {
     }
 
 public:
@@ -85,24 +84,16 @@ public:
             stb_vorbis_close(_handle);
             _handle = 0;
         }
-        if (_file) {
-            EbrFclose(_file);
-            _file = 0;
-        }
     }
 
-    static OpaqueAudioFileID* openURL(char* url) {
-        //  Open the file using EbrFopen so that path mapping is applied
-        //  (since the url will often reference / as the root of the app directory)
-        EbrFile* file = EbrFopen(url, "rb");
-        if (file) {
+    static OpaqueAudioFileID* openURL(NSURL* url) {
+        NSData* data = [NSData dataWithContentsOfURL:url];
+        if ([data length] > 0) {
             int err = 0;
-
-            //  Pass the FILE* handle directly to stb_vorbis
-            stb_vorbis* handle = stb_vorbis_open_file(EbrNativeFILE(file), 0, &err, 0);
+            stb_vorbis* handle = stb_vorbis_open_memory((unsigned char*)[data bytes], [data length], &err, nullptr);
 
             if (handle && err == 0) {
-                return new AudioFileOGG(handle, file);
+                return new AudioFileOGG(handle, data);
             }
 
             TraceError(TAG, L"[OGG] OGG open error: %d", err);
@@ -196,9 +187,10 @@ class AudioFileWAV : public OpaqueAudioFileID {
         u32 dataSize;
     } _dataChunk;
 
-    EbrFile* _file;
+    StrongId<NSInputStream> _inStream;
 
-    AudioFileWAV(Header header, ChunkHeader dataChunk, EbrFile* f) : _header(header), _dataChunk(dataChunk), _file(f) {
+    AudioFileWAV(Header header, ChunkHeader dataChunk, NSInputStream* inStream)
+        : _header(header), _dataChunk(dataChunk), _inStream(inStream) {
         fileFormat.mFormatID = 'lpcm';
         fileFormat.mChannelsPerFrame = _header.channels;
         fileFormat.mSampleRate = _header.samplesPerSec;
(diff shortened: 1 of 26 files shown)


Frameworks/Foundation/NSData.mm
```diff
@@ -396,42 +388,29 @@ - (instancetype)initWithContentsOfFile:(NSString*)filename options:(NSDataReadin
...
+    uint8_t byteBuffer[4096] = {};
+    NSMutableData* tempData = [NSMutableData data];
+
+    while ([inputStream hasBytesAvailable]) {
+        auto result = [inputStream read:byteBuffer maxLength:_countof(byteBuffer)];
+        [tempData appendBytes:byteBuffer length:result];
+    }
+
+    return [self initWithData:tempData];
 }
```
DHowett-MSFT: I'm wary of introducing a copy here.

Frameworks/Foundation/NSData.mm
```diff
@@ -396,42 +388,29 @@ - (instancetype)initWithContentsOfFile:(NSString*)filename options:(NSDataReadin
...
 
-    return self;
+    auto closeFile = wil::ScopeExit([&]() { [inputStream close]; });
+
+    uint8_t byteBuffer[4096] = {};
+    NSMutableData* tempData = [NSMutableData data];
+
+    while ([inputStream hasBytesAvailable]) {
+        auto result = [inputStream read:byteBuffer maxLength:_countof(byteBuffer)];
+        [tempData appendBytes:byteBuffer length:result];
```
DHowett-MSFT: >[tempData appendBytes:byteBuffer length:result]; [](start = 8, length = 48)

Sorry, two copies.

Frameworks/Foundation/NSData.mm
```diff
@@ -303,69 +305,59 @@ - (instancetype)initWithContentsOfFile:(NSString*)filename {
...
-        TraceVerbose(TAG, L"-[NSData::writeToURL]: Only file: URLs are supported. (%hs)", [[url absoluteString] UTF8String]);
-        return NO;
-    }
-    return [self writeToFile:[url path] atomically:atomically];
+    return [self writeToURL:url options:(atomically ? NSDataWritingAtomic : 0) error:nil];
 }
 
 /**
  @Status Caveat
  @Notes options parameter not supported
```
msft-Jeyaram: >options parameter not supported [](start = 8, length = 31)

seems like we are supporting a portion of it?

Frameworks/Foundation/NSData.mm
```diff
@@ -303,69 +305,59 @@ - (instancetype)initWithContentsOfFile:(NSString*)filename {
...
 */
 - (BOOL)writeToURL:(NSURL*)url options:(NSDataWritingOptions)options error:(NSError**)errorp {
-    if (![url isFileURL]) {
-        TraceVerbose(TAG, L"-[NSData::writeToURL]: Only file: URLs are supported. (%hs)", [[url absoluteString] UTF8String]);
+    if (!url) {
+        if (errorp) {
+            *errorp = [NSError errorWithDomain:@"NSData" code:100 userInfo:nil];
```
msft-Jeyaram: >100 [](start = 62, length = 3)

I can't recall If I checked this, but does this match the reference platform?

Frameworks/Foundation/NSData.mm
```diff
@@ -303,69 +305,59 @@ - (instancetype)initWithContentsOfFile:(NSString*)filename {
...
+    [outputStream open];
+
+    if (NSStreamStatusOpen != outputStream.streamStatus) {
+        if (errorp) {
+            *errorp = [NSError errorWithDomain:@"NSData" code:100 userInfo:nil];
+        }
+        return NO;
+    }
+    auto numBytes = [self length];
+    auto result = [outputStream write:(const unsigned char*)[self bytes] maxLength:numBytes];
```
msft-Jeyaram: > result = [outputStream write:(const unsigned char*)[self bytes] maxLength:numBytes]; [](start = 8, length = 85)

we might not write the whole buffer, we need to check if we have any left over and write the rest?
I recall similar code existing already.

Frameworks/Foundation/NSURLProtocol_file.mm
```diff
@@ -46,28 +52,19 @@ + (BOOL)canInitWithRequest:(id)request {
...
-    const char* pFilePath = [_path UTF8String];
-
-    fpIn = EbrFopen(pFilePath, "rb");
-    if (!fpIn) {
-        TraceVerbose(TAG, L"Couldn't open %hs", pFilePath);
-    } else {
-        TraceVerbose(TAG, L"Opened %hs", pFilePath);
-    }
+    // TODO 1162: An NSInputStream should be initialized here and run loop events should be used to schedule
+    // the file operations.
```
DHowett-MSFT: I certainly hope NSURLConnection is up to the task. As well, I wonder if the reference platform short-circuits file URL handling.

It is fully possible to load a file from a file URL without spinning a run loop inbetween.

Frameworks/Foundation/NSData.mm
```diff
@@ -396,42 +388,29 @@ - (instancetype)initWithContentsOfFile:(NSString*)filename options:(NSDataReadin
...
     }
 
-    return self;
+    auto closeFile = wil::ScopeExit([&]() { [inputStream close]; });
+
+    uint8_t byteBuffer[4096] = {};
+    NSMutableData* tempData = [NSMutableData data];
+
+    while ([inputStream hasBytesAvailable]) {
+        auto result = [inputStream read:byteBuffer maxLength:_countof(byteBuffer)];
```
msft-Jeyaram: >byteBuffer [](start = 40, length = 10)

you can also just get the mutableBytes of the NSMutableData and read into it directly.
If you need to grow the capacity, you can do it via increaseLength

bbowman: increaseLength isn't really desirable here. Ideally I'd be able to read the chunks directly in and return the same data. If the length is wrong I can't return that data and need to manually reconstruct it with the right length that I keep track of. appendBytes is morally equivalent of what you are suggesting but it adjusts the capacity instead of length at the expense of the temp buffer (which is only allocated once and is stack space so its "cheap" in the grand scheme of things). Basically this is letting the NSData append impl be smart about growth etc which seems better than me trying to be smart (since I'm not very). I'll look for other ways to avoid extra copies though.

---
In reply to: [83929862](https://github.com/Microsoft/WinObjC/pull/1175#discussion_r83929862) [](ancestors = 83929862)

Frameworks/AudioToolbox/AudioFile.mm
```diff
@@ -580,11 +622,14 @@ DWORD AudioFileOpenWithCallbacks(void* context,
...
-
-    char header[4] = { 0 };
-    EbrFread(&header, sizeof(header), 1, in);
+    NSInputStream* in = [NSCallbackInputStream inputStreamWithCallBacks:context
+                                                               readFunc:readFunc
+                                                              writeFunc:writeFunc
+                                                            getSizeFunc:getSizeFunc
+                                                            setSizeFunc:setSizeFunc];
+    [in open];
+    char header[4] = {};
```
rajsesh-msft: >char [](start = 4, length = 4)

nit: this could be uint8_t to avoid casts.  It is unsigned char, so the other apis should work?

bbowman: hilariously no because of the use of strncmp

---
In reply to: [83959944](https://github.com/Microsoft/WinObjC/pull/1175#discussion_r83959944) [](ancestors = 83959944)

bbowman: a cast is needed somewhere so I put it here.

---
In reply to: [83964371](https://github.com/Microsoft/WinObjC/pull/1175#discussion_r83964371) [](ancestors = 83964371,83959944)

Frameworks/Foundation/NSURLProtocol_file.mm
bbowman: Whoever is looking at 116 2can make those determinations.

---
In reply to: [83928251](https://github.com/Microsoft/WinObjC/pull/1175#discussion_r83928251) [](ancestors = 83928251)

Frameworks/Foundation/NSData.mm
bbowman: copied from the file one now that file forwards to URL. Not the purpose of this pr to validate our error codes.

---
In reply to: [83927171](https://github.com/Microsoft/WinObjC/pull/1175#discussion_r83927171) [](ancestors = 83927171)

Frameworks/Foundation/NSURLProtocol_file.mm
```diff
@@ -46,28 +52,19 @@ + (BOOL)canInitWithRequest:(id)request {
...
-
-    _path = [[url path] copy];
-    // id host = [NSHost hostWithName:hostName];
+        _path.attach([[url path] copy]);
+    }
 
     return self;
 }
 
 - (id)startLoading {
```
rajsesh-msft: >startLoading [](start = 6, length = 12)

This could potentially break something (anyone using this externally).

Frameworks/AudioToolbox/CAFDecoder.mm
```diff
@@ -184,68 +184,45 @@ uint64_t int64Swap(uint64_t val) {
...
-    int cur = EbrFtell(fpIn);
-    EbrFseek(fpIn, 0, SEEK_END);
-    int left = EbrFtell(fpIn) - cur;
-    EbrFseek(fpIn, cur, SEEK_SET);
-
-    return left;
-}
-
-Byte* CAFDecoder::GetBytes(EbrFile* fpIn, int len) {
+Byte* CAFDecoder::GetBytes(NSInputStream* stream, int len) {
```
rajsesh-msft: Changes to this file don't have much collateral, please be sure to validate internal apps.

rajsesh-msft:
:shipit:

bbowman:
@DHowett-MSFT @msft-Jeyaram My full build passed and it seems like CAF audio is not broken. huzzah. Any other comments before I go go?

Frameworks/Foundation/NSURLProtocol_file.mm
bbowman: there is no functional change. 

---
In reply to: [83974374](https://github.com/Microsoft/WinObjC/pull/1175#discussion_r83974374) [](ancestors = 83974374)

DHowett-MSFT:
I'm not 100% stoked about the foundation dep in AudioToolbox, _but_ I understand that we already had one (IIRC) and it's fairly harmless to replace if we do decide to set up a boundary.
LGTM!

---
In reply to: [254903901](https://github.com/Microsoft/WinObjC/pull/1175#issuecomment-254903901) [](ancestors = 254903901)

DHowett-MSFT:
:shipit:

PR #1181, opened 2016-10-19
CGContext improvements - DPI, parameter validation, stroke styles, interpolation mode

DHowett-MSFT:
Featuring:

* Make CGContext display scale-agnostic by propagating the scale into Direct2D's DPI.

    A CGContext rendering to a scaled surface will not apply a scale
    transformation. The scaling logic has been moved into render target
    creation.

* Replace __CGContextDrawGeometry with RenderToCommandList+RenderCommandList.
* Make all CG APIs nullptr-safe.
* Move stroke style caching/computation into drawing state.

### DIFF
--- a/Frameworks/CoreGraphics/CGContext.mm
+++ b/Frameworks/CoreGraphics/CGContext.mm
@@ -588,6 +588,7 @@ static void __CGContextDrawGeometry(CGContextRef context, ID2D1Geometry* geometr
     FAIL_FAST_IF_FAILED(renderTarget.As(&deviceContext));
 
     D2D1_SIZE_F targetSize = renderTarget->GetSize();
+    //D2D1_SIZE_U unscaledTargetSize = renderTarget->GetPixelSize();
     ComPtr<ID2D1Image> originalTarget;
     deviceContext->GetTarget(&originalTarget);
 
@@ -599,6 +600,9 @@ static void __CGContextDrawGeometry(CGContextRef context, ID2D1Geometry* geometr
 
     CGAffineTransform& currentTransform = currentState.transform;
     CGAffineTransform transform = CGAffineTransformScale(currentTransform, 1.0, -1.0);
+    // undo the scale (propagated to DPI)
+    //transform = CGAffineTransformScale(transform, targetSize.width/(double)unscaledTargetSize.width, targetSize.height/(double)unscaledTargetSize.height);
+    // undo the slide (slide was in points, must be done after scale is undone)
     transform = CGAffineTransformTranslate(transform, 0., targetSize.height);
     deviceContext->SetTransform({ transform.a, transform.b, transform.c, transform.d, transform.tx, transform.ty });
 
@@ -1434,12 +1438,14 @@ CGImageRef CGBitmapContextGetImage(CGContextRef context) {
     return ((__CGBitmapContext*)context)->Impl().image.get();
 }
 
-CGContextRef _CGBitmapContextCreateWithTexture(int width, int height, DisplayTexture* texture, DisplayTextureLocking* locking) {
+CGContextRef _CGBitmapContextCreateWithTexture(
+    int width, int height, float scale, DisplayTexture* texture, DisplayTextureLocking* locking) {
     CGImageRef newImage = nullptr;
     __CGSurfaceInfo surfaceInfo = _CGSurfaceInfoInit(width, height, _ColorARGB);
     newImage = new CGGraphicBufferImage(surfaceInfo, texture, locking);
 
     ComPtr<ID2D1RenderTarget> renderTarget = newImage->Backing()->GetRenderTarget();
+    renderTarget->SetDpi(96 * scale, 96 * scale);
 
     __CGBitmapContext* context = __CGBitmapContext::CreateInstance(kCFAllocatorDefault);
     __CGContextInitWithRenderTarget(context, renderTarget.Get());
--- a/Frameworks/CoreGraphics/CGContext.mm
+++ b/Frameworks/CoreGraphics/CGContext.mm
@@ -109,7 +109,7 @@ struct __CGContextImpl {
     }
 };
 
-struct __CGContext: CoreFoundation::CppBase<__CGContext, __CGContextImpl> {
+struct __CGContext : CoreFoundation::CppBase<__CGContext, __CGContextImpl> {
     inline ComPtr<ID2D1RenderTarget>& RenderTarget() {
         return _impl.renderTarget;
     }
@@ -132,6 +132,24 @@ struct __CGContext: CoreFoundation::CppBase<__CGContext, __CGContextImpl> {
         return factory;
     }
 };
+
+#define CGCONTEXT_CHECK_NULL(context)                                         \
+    do {                                                                      \
+        if (!context) {                                                       \
+            TraceError(TAG, L"%hs(...): null context!", __PRETTY_FUNCTION__); \
+            return StubReturn();                                              \
+        }                                                                     \
+    \
+} while (0)
+
+#define CGCONTEXT_CHECK_NULLV(context)                                        \
+    do {                                                                      \
+        if (!context) {                                                       \
+            TraceError(TAG, L"%hs(...): null context!", __PRETTY_FUNCTION__); \
+            return;                                                           \
+        }                                                                     \
+    \
+} while (0)
 
 #pragma region CFRuntimeClass
 /**
@@ -167,6 +185,10 @@ CGContextRef _CGContextCreateWithD2DRenderTarget(ID2D1RenderTarget* renderTarget
  @Status Interoperable
 */
 CGContextRef CGContextRetain(CGContextRef context) {
+    if (!context) {
+        return nullptr;
+    }
+
     CFRetain((CFTypeRef)context);
     return context;
 }
@@ -188,10 +210,12 @@ void CGContextRelease(CGContextRef context) {
  @Status Interoperable
 */
 void CGContextSetBlendMode(CGContextRef context, CGBlendMode mode) {
+    CGCONTEXT_CHECK_NULLV(context);
     UNIMPLEMENTED();
 }
 
 CGBlendMode CGContextGetBlendMode(CGContextRef context) {
+    CGCONTEXT_CHECK_NULL(context);
     UNIMPLEMENTED();
     return StubReturn();
 }
@@ -201,6 +225,7 @@ CGBlendMode CGContextGetBlendMode(CGContextRef context) {
  @Status Interoperable
 */
 void CGContextSetFillPattern(CGContextRef context, CGPatternRef pattern, const CGFloat* components) {
+    CGCONTEXT_CHECK_NULLV(context);
     UNIMPLEMENTED();
 }
 
@@ -208,6 +233,7 @@ void CGContextSetFillPattern(CGContextRef context, CGPatternRef pattern, const C
  @Status Interoperable
 */
 void CGContextSetPatternPhase(CGContextRef context, CGSize phase) {
+    CGCONTEXT_CHECK_NULLV(context);
     UNIMPLEMENTED();
 }
 
@@ -216,6 +242,7 @@ void CGContextSetPatternPhase(CGContextRef context, CGSize phase) {
  @Status Stub
 */
 void CGContextSetCharacterSpacing(CGContextRef context, CGFloat spacing) {
+    CGCONTEXT_CHECK_NULLV(context);
     UNIMPLEMENTED();
 }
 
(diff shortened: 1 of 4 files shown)


Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -84,6 +90,20 @@ static inline D2D_RECT_F __CGRectToD2D_F(CGRect rect) {
...
     // Alpha Blending
     CGFloat alpha = 1.0f;
+
+    inline void ComputeStrokeStyle(ID2D1Factory* factory) {
+        if (strokeStyle) {
+            return;
+        }
+
+        std::vector<CGFloat> adjustedDashes(dashes.size());
+        std::transform(dashes.cbegin(), dashes.cend(), adjustedDashes.begin(), [this](const float& f) -> CGFloat { return f / lineWidth; });
```
DHowett-MSFT: _clang-format_ put this all on one line :(

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -133,6 +153,24 @@ inline void ClearPath() {
     }
 };
 
+#define CGCONTEXT_CHECK_NULL(context)                                         \
+    do {                                                                      \
+        if (!context) {                                                       \
+            TraceError(TAG, L"%hs(...): null context!", __PRETTY_FUNCTION__); \
```
DHowett-MSFT: The reference platform logs for these.

rajsesh-msft:

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -609,10 +609,7 @@ - (void)display {
         }
 
         if (target->Backing()->Height() != 0) {
-            CGContextTranslateCTM(drawContext, 0, float(target->Backing()->Height()));
-        }
-        if (priv->contentsScale != 1.0f) {
-            CGContextScaleCTM(drawContext, priv->contentsScale, priv->contentsScale);
+            CGContextTranslateCTM(drawContext, 0, float(target->Backing()->Height())/priv->contentsScale);
```
rajsesh-msft: This warrants some explanation/comments in the code.

Frameworks/CoreGraphics/CGContext.mm
rajsesh-msft: >lineWidth [](start = 126, length = 9)

would we ever end up with lineWidth ~ 0?

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -133,6 +153,24 @@ inline void ClearPath() {
     }
 };
 
+#define CGCONTEXT_CHECK_NULL(context)                                         \
```
rajsesh-msft: >CGCONTEXT_CHECK_NULL [](start = 8, length = 20)

nit: RETURN_IF_CGCONTEXT_NULL(context, retval) instead?

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -167,6 +205,10 @@ CGContextRef _CGContextCreateWithD2DRenderTarget(ID2D1RenderTarget* renderTarget
  @Status Interoperable
 */
 CGContextRef CGContextRetain(CGContextRef context) {
+    if (!context) {
```
rajsesh-msft: >context [](start = 9, length = 7)

no CHECK_NULL here?

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -574,88 +660,107 @@ void CGContextRestoreGState(CGContextRef context) {
...
     }
 
-    deviceContext->DrawImage(commandList.Get());
+    deviceContext->DrawImage(commandList);
 
     if (layer) {
         renderTarget->PopLayer();
     }
 
     FAIL_FAST_IF_FAILED(deviceContext->EndDraw());
```
rajsesh-msft: >FAIL_FAST_IF_FAILED [](start = 4, length = 19)

[https://msdn.microsoft.com/en-us/library/windows/desktop/dd371766(v=vs.85).aspx](https://msdn.microsoft.com/en-us/library/windows/desktop/dd371766(v=vs.85).aspx), we should not failfast if the EndDraw command results in D2DERR_RECREATE_TARGET.  It can happen if the graphics driver stops responding, and the only way is to recreate the target and replay the commands.

rajsesh-msft: This is only needed right before we move to h/w acceleration codepath (presumable, software codepath will not stop responding).  We still need to track the issue.

---
In reply to: [83979454](https://github.com/Microsoft/WinObjC/pull/1181#discussion_r83979454) [](ancestors = 83979454)

rajsesh-msft: >FAIL_FAST_IF_FAILED [](start = 4, length = 19)

Is this the correct transform *always*?

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -70,9 +74,11 @@ static inline D2D_RECT_F __CGRectToD2D_F(CGRect rect) {
         0.f,
     };
     std::vector<CGFloat> dashes{};
-
     CGFloat lineWidth = 1.0f;
 
+    // Computed from the above at draw time
+    ComPtr<ID2D1StrokeStyle> strokeStyle{ nullptr };
```
MSFTFox: >{ nullptr } [](start = 40, length = 11)

nit: no need for nullptr + other locations.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -992,91 +1145,145 @@ void CGContextSetLineCap(CGContextRef context, CGLineCap lineCap) {
...
+*/
+CGInterpolationQuality CGContextGetInterpolationQuality(CGContextRef context) {
+    CGCONTEXT_CHECK_NULL(context);
+
+    auto& state = context->CurrentGState();
+    switch (state.bitmapInterpolationMode) {
+        case D2D1_INTERPOLATION_MODE_LINEAR:
+            return kCGInterpolationDefault;
+        case D2D1_INTERPOLATION_MODE_NEAREST_NEIGHBOR:
+            return kCGInterpolationNone;
```
rajsesh-msft: This kCGInterpolationModeLow?

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -84,6 +90,20 @@ static inline D2D_RECT_F __CGRectToD2D_F(CGRect rect) {
 
     // Alpha Blending
     CGFloat alpha = 1.0f;
+
+    inline void ComputeStrokeStyle(ID2D1Factory* factory) {
```
rajsesh-msft: >ComputeStrokeStyle [](start = 16, length = 18)

Premature optimization: It might make sense to take the ID2D1DeviceContext here and query the factory only if you already dont have the stroke style.  Will simplify calling code, and save some cycles when you dont need to re-compute.

rajsesh-msft:
:shipit:

Frameworks/CoreGraphics/CGContext.mm
DHowett-MSFT: Brilliant!

Frameworks/CoreGraphics/CGContext.mm
DHowett-MSFT: Nope, these ones are intended to be silent (as per reference platform)

Frameworks/CoreGraphics/CGContext.mm
aballway: nit: no need for -> CGFloat

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -992,91 +1145,145 @@ void CGContextSetLineCap(CGContextRef context, CGLineCap lineCap) {
...
+    CGCONTEXT_CHECK_NULLV(context);
+    static D2D1_INTERPOLATION_MODE d2dModes[] = {
+        /* Default */ D2D1_INTERPOLATION_MODE_LINEAR,
+        /* None    */ D2D1_INTERPOLATION_MODE_NEAREST_NEIGHBOR,
+        /* Low     */ D2D1_INTERPOLATION_MODE_LINEAR,
+        /* Medium  */ D2D1_INTERPOLATION_MODE_MULTI_SAMPLE_LINEAR,
+        /* High    */ D2D1_INTERPOLATION_MODE_CUBIC,
+    };
+
+    if (quality < kCGInterpolationDefault) {
```
aballway: nit: std::min

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -574,88 +660,107 @@ void CGContextRestoreGState(CGContextRef context) {
...
+    // Translate it back onscreen.
+    transform = CGAffineTransformTranslate(transform, 0., targetSize.height);
+    deviceContext->SetTransform(__CGAffineTransformToD2D_F(transform));
 
     deviceContext->BeginDraw();
-    deviceContext->SetTarget(originalTarget.Get());
 
     bool layer = false;
-    if (currentState.alpha != 1.0f || false /* mask, clip, etc. */) {
+    if (state.alpha != 1.0f || false /* mask, clip, etc. */) {
```
msft-Jeyaram: > /* mask, clip, etc. */ [](start = 36, length = 23)

do we need to add a TODO here?

Frameworks/CoreGraphics/CGContext.mm
DHowett-MSFT: actually, there is; but i have the direction backwards.

CGFloat flips between float and double in certain reference platform configurations, but D2D prefers float. Since we're passing a raw buffer, we must type-match.

I should be storing CGFloats internally and converting to float here.

---
In reply to: [84096583](https://github.com/Microsoft/WinObjC/pull/1181#discussion_r84096583) [](ancestors = 84096583)

DHowett-MSFT: 0 doesn't bother it, but i'll bail out here anyway

---
In reply to: [83978267](https://github.com/Microsoft/WinObjC/pull/1181#discussion_r83978267) [](ancestors = 83978267)

Frameworks/CoreGraphics/CGContext.mm
DHowett-MSFT: I'm leaving this one open for now; I know we need to recreate the target, but we might have to do some magic to figure out how to recreate it; perhaps we want a callback that does it for us, since the render target can be specified by an external caller.

---
In reply to: [83979563](https://github.com/Microsoft/WinObjC/pull/1181#discussion_r83979563) [](ancestors = 83979563,83979454)

DHowett-MSFT: (There's no good generic way for us to recreate it, and we'll need to log all the device-specific resources we use to recreate them too.)

---
In reply to: [84196354](https://github.com/Microsoft/WinObjC/pull/1181#discussion_r84196354) [](ancestors = 84196354,83979563,83979454)

DHowett-MSFT: perhaps this code should save the transform and restore it; we need to make certain that we are composing command lists down with the right transformation, but other direct draw functions than this need to know about the transformation too.

Evaluating.

---
In reply to: [83979657](https://github.com/Microsoft/WinObjC/pull/1181#discussion_r83979657) [](ancestors = 83979657)

Frameworks/CoreGraphics/CGContext.mm
DHowett-MSFT: This is covered by upcoming tasks.

---
In reply to: [84129211](https://github.com/Microsoft/WinObjC/pull/1181#discussion_r84129211) [](ancestors = 84129211)

Frameworks/CoreGraphics/CGContext.mm
DHowett-MSFT: Adding a todo for #1194!

---
In reply to: [84196384](https://github.com/Microsoft/WinObjC/pull/1181#discussion_r84196384) [](ancestors = 84196384,84196354,83979563,83979454)

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -133,6 +161,15 @@ inline void ClearPath() {
     }
 };
 
+#define NOISY_RETURN_IF_NULL(param, ...)                                         \
+    do {                                                                         \
+        if (!context) {                                                          \
+            TraceError(TAG, L"%hs(...): null " #param "!", __PRETTY_FUNCTION__); \
```
DHowett-MSFT: >(...) [](start = 33, length = 5)

nit: remove this; __PRETTY_FUNCTION__ includes one ;P

PR #1189, opened 2016-10-19
Fix framesetter test crash on reference platform

### LINKED ISSUE #1150: CTCatalog Framesetter Test crashes on reference platform
CT-DW branch

    CFAttributedStringRef attrString =
        CFAttributedStringCreate(kCFAllocatorDefault, (__bridge CFStringRef)_text, (__bridge CFDictionaryRef)attributesDict);

is the offending line. _text is nil at this point, which we handle more gracefully than the reference platform.

As a quick fix, we can work around this by wrapping the logic in a if (_text) { } block, but for a proper fix, there's probably an invalid assumption about the order of operations that needs to be fixed.

aballway:
Fixes #1150


DHowett-MSFT:
:shipit:

rajsesh-msft:
:clock1:

ms-jihua:
:shipit:

PR #1200, opened 2016-10-20
Fix drawing of line with multiple fonts

aballway:
When we flip the coordinate system to draw for CoreText we need to know the height of the entire line or we will end up with runs with different baselines based upon their respective fonts.

### DIFF
--- a/Frameworks/CoreGraphics/CGContext.mm
+++ b/Frameworks/CoreGraphics/CGContext.mm
@@ -60,7 +60,7 @@ __CGContext::__CGContext(CGImageRef pDest) {
 #ifdef DEBUG_CONTEXT_COUNT
     TraceVerbose(TAG, L"contextCount: %d", contextCount);
 #endif
-    object_setClass((id) this, [CGNSContext class]);
+    object_setClass((id)this, [CGNSContext class]);
     scale = 1.0f;
     _backing = pDest->Backing()->CreateDrawingContext(this);
 }
@@ -1239,8 +1239,8 @@ bool CGContextIsPointInPath(CGContextRef c, bool eoFill, CGFloat x, CGFloat y) {
     return c->Backing()->CGContextIsPointInPath(eoFill, x, y);
 }
 
-void CGContextDrawGlyphRun(CGContextRef ctx, const DWRITE_GLYPH_RUN* glyphRun) {
-    ctx->Backing()->CGContextDrawGlyphRun(glyphRun);
+void CGContextDrawGlyphRun(CGContextRef ctx, const DWRITE_GLYPH_RUN* glyphRun, float lineHeight) {
+    ctx->Backing()->CGContextDrawGlyphRun(glyphRun, lineHeight);
 }
 // TODO 1077:: Remove once D2D render target is implemented
 void _CGContextSetScaleFactor(CGContextRef ctx, float scale) {
--- a/Frameworks/CoreGraphics/CGContext.mm
+++ b/Frameworks/CoreGraphics/CGContext.mm
@@ -1239,8 +1239,8 @@ bool CGContextIsPointInPath(CGContextRef c, bool eoFill, CGFloat x, CGFloat y) {
     return c->Backing()->CGContextIsPointInPath(eoFill, x, y);
 }
 
-void CGContextDrawGlyphRun(CGContextRef ctx, const DWRITE_GLYPH_RUN* glyphRun, float lineHeight) {
-    ctx->Backing()->CGContextDrawGlyphRun(glyphRun, lineHeight);
+void CGContextDrawGlyphRun(CGContextRef ctx, const DWRITE_GLYPH_RUN* glyphRun, float lineAscent) {
+    ctx->Backing()->CGContextDrawGlyphRun(glyphRun, lineAscent);
 }
 // TODO 1077:: Remove once D2D render target is implemented
 void _CGContextSetScaleFactor(CGContextRef ctx, float scale) {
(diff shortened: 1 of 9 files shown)


Frameworks/include/CGContextCairo.h
```diff
@@ -138,7 +138,7 @@ class CGContextCairo : public CGContextImpl {
     virtual bool CGContextIsPointInPath(bool eoFill, float x, float y);
     virtual CGPathRef CGContextCopyPath(void);
 
-    virtual void CGContextDrawGlyphRun(const DWRITE_GLYPH_RUN* glyphRun);
+    virtual void CGContextDrawGlyphRun(const DWRITE_GLYPH_RUN* glyphRun, float lineHeight);
```
ms-jihua: >lineHeight [](start = 79, length = 10)

My only concern is, I don't know if this is a good name. lineHeight to me implies the distance from the bottom of the line to the top of the line, which would be ascent + descent. (ascent - descent)/2 would be the distance between the baseline and the center, but there'd be a weird burden on the caller to know that.

Probably the best way is just to pass both ascent and descent and leave the calculations in CGContextDrawGlyphRun?

ms-jihua: Possibly name them lineAscent and lineDescent so it's clearer that they're metrics of the surrounding line.

---
In reply to: [84549518](https://github.com/Microsoft/WinObjC/pull/1200#discussion_r84549518) [](ancestors = 84549518)

Frameworks/CoreText/CTLine.mm
```diff
@@ -240,7 +240,10 @@ void _CTLineDraw(CTLineRef lineRef, CGContextRef ctx, bool adjustTextPosition) {
             CGContextSetTextPosition(ctx, curTextPos.x + curRun->_relativeXOffset, curTextPos.y);
         }
 
-        _CTRunDraw(static_cast<CTRunRef>(curRun), ctx, CFRange{}, false);
+        // Get height of the line so we draw with the correct baseline for each run
+        CGFloat ascent, descent;
+        CTLineGetTypographicBounds(lineRef, &ascent, &descent, nullptr);
+        _CTRunDraw(static_cast<CTRunRef>(curRun), ctx, CFRange{}, false, ascent - descent);
```
rajsesh-msft: ascent-descent because descent is negative, right?  Would be useful to add a note here for future.

ms-jihua: wait, is descent negative? I didn't think this was the case.

---
In reply to: [84551029](https://github.com/Microsoft/WinObjC/pull/1200#discussion_r84551029) [](ancestors = 84551029)

aballway: descent is negative if it extends below the baseline, positive above

Frameworks/include/CGContextCairo.h
```diff
@@ -138,7 +138,7 @@ class CGContextCairo : public CGContextImpl {
...
```
DHowett-MSFT: I'm concerned that this gives the caller too much control (or, alternatively, _the need_ for too much control)

aballway: Discussed offline, leaving as is

DHowett-MSFT:
:shipit:

ms-jihua:
:shipit:

Frameworks/CoreText/CTLine.mm
ms-jihua: we discussed offline - my mental model of the baseline was wrong, we should just use ascent.

---
In reply to: [84551992](https://github.com/Microsoft/WinObjC/pull/1200#discussion_r84551992) [](ancestors = 84551992)

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -60,7 +60,7 @@ - (void)dealloc {
 #ifdef DEBUG_CONTEXT_COUNT
     TraceVerbose(TAG, L"contextCount: %d", contextCount);
 #endif
-    object_setClass((id) this, [CGNSContext class]);
+    object_setClass((id)this, [CGNSContext class]);
```
msft-Jeyaram: >object_setClass((id)this, [CGNSContext class]); [](start = 4, length = 47)

C++ cast?

Frameworks/CoreGraphics/CGContextCairo.mm
```diff
@@ -2014,22 +2014,19 @@
...
-    DWRITE_FONT_METRICS fontMetrics;
-    glyphRun->fontFace->GetMetrics(&fontMetrics);
-    float baselineCenterDelta = (fontMetrics.ascent - fontMetrics.descent) * glyphRun->fontEmSize / fontMetrics.designUnitsPerEm / 2;
 
     // Transform to text space
     // Technically there should be a horizontal translation to the center as well,
     // but it's to the center of _each individual glyph_, as the reference platform applies the text matrix to each glyph individually
     // Uncertain whether it's ever going to be worth it to support this using DWrite, so just ignore it for now
-    CGAffineTransform transform = CGAffineTransformMake(1, 0, 0, -1, 0, baselineCenterDelta);
+    CGAffineTransform transform = CGAffineTransformMake(1, 0, 0, -1, 0, lineAscent / 2.0f);
```
msft-Jeyaram: >/ 2.0f [](start = 83, length = 6)

what is the significance of this division by 2?

Frameworks/CoreGraphics/CGContext.mm
aballway: I'll let @DHowett-MSFT clean these up once all of this is changed

PR #1205, opened 2016-10-21
Perform major surgery on CGContext.

DHowett-MSFT:
**This pull request contains no functional changes.**

This is part 2 of the CGContext cleanup; I've categorized every method and moved them into reasonable regions. We had two copies of `CGContextShowGlyphsWithAdvances`, one of which has been removed.

Verification method: 
```
$ diff -u <(git show 29e1b23:./CGContext.mm|sort) <(sort CGContext.mm | dos2unix)
...
```

Sorting a source file to diff it is silly, but it shows that the bulk of the changes were in region renames and the deletion of CGContextShowGlyphsWithAdvances.

### DIFF
--- a/Frameworks/CoreGraphics/CGContext.mm
+++ b/Frameworks/CoreGraphics/CGContext.mm
@@ -170,7 +170,9 @@ struct __CGContext : CoreFoundation::CppBase<__CGContext, __CGContextImpl> {
     \
 } while (0)
 
-#pragma region CFRuntimeClass
+static const wchar_t* TAG = L"CGContext";
+
+#pragma region Global State - CFRuntimeClass
 /**
  @Status Interoperable
 */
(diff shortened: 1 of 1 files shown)


Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -419,891 +444,969 @@ void CGContextSetCTM(CGContextRef context, CGAffineTransform transform) {
...
+/**
+ @Status Stub
+ @Notes
+*/
+bool CGContextPathContainsPoint(CGContextRef context, CGPoint point, CGPathDrawingMode mode) {
+    NOISY_RETURN_IF_NULL(context, StubReturn());
+    UNIMPLEMENTED();
+    return StubReturn();
 }
+#pragma endregion
```
rajsesh-msft: @DHowett-MSFT other than arranging the file to be more logical, are there real code changes that need to be reviewed?

ms-jihua:
:shipit:

Frameworks/CoreGraphics/CGContext.mm
rajsesh-msft: I approve this pull request.

msft-Jeyaram:
:shipit:

PR #1215, opened 2016-10-24
NSURL changes for special filepath for Storage Files

bbowman:
Description:
This change adds the ability to create an NSURL with a Storage File and do all the
things one would expect to be able to accomplish with it like reading and writing
NSData and NSStreams. This is accomplished by making use of remnants of a "direct file"
url and the Windows FutureAccessCache to retreive the file at a later point.

EbrFile is refactored, minimized, and then extended to work with Storage Files.

How verified:
New functional test

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1215)


bbowman:
@rajsesh-msft @memontic-ms @DHowett-MSFT  @bviglietta  please take a look

Frameworks/Foundation/NSURL.mm
```diff
@@ -890,4 +908,63 @@ - (BOOL)getPromisedItemResourceValue:(id _Nullable*)value forKey:(NSString*)key
...
+
+    ComPtr<IStorageItemAccessList> accessList;
+    RETURN_NULL_IF_FAILED(storageApplicationPermissionsStatics->get_FutureAccessList(&accessList));
+
+    // See how close we are to the cap.
+    uint32_t maxItems = 0;
+    RETURN_NULL_IF_FAILED(accessList->get_MaximumItemsAllowed(&maxItems));
+
+    // Documentation says 1000 but lets just make sure it was somthing valid.
+    RETURN_NULL_IF_FALSE(maxItems > 0);
```
DHowett-MSFT: >ULL_IF_FALSE(maxItems > 0); [](start = 12, length = 27)

Is there a future world where they're unlimited and managed by the system? We may not want to break on those platforms.

Frameworks/Foundation/NSURL.mm
```diff
@@ -26,9 +26,27 @@
...
+#import "NSStringInternal.h"
 #import "StubReturn.h"
 
 #import <memory>
+#import <mutex>
+
+#include <COMIncludes.h>
+#include "ErrorHandling.h"
+#include <wrl/client.h>
+#include <wrl\wrappers\corewrappers.h>
```
msft-Jeyaram: forward slashes

Frameworks/Starboard/EbrIOFile.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include "PathMapper.h"
+
+#include <sys\stat.h>
+
+
+std::shared_ptr<EbrFile> EbrIOFile::CreateInstance(const char* path, int mode, int share, int pmode) {
+    int fid{};
+    auto result = _sopen_s(&fid, CPathMapper(path), mode, share, pmode);
+
+    if (result != 0) {
```
aballway: nit: ternary return

Frameworks/Starboard/EbrStorageFile.cpp
```diff
@@ -0,0 +1,351 @@
...
+}
+
+int EbrStorageFile::Tell() {
+    return Lseek(0, SEEK_CUR);
+}
+
+int EbrStorageFile::Flush() {
+    std::lock_guard<std::recursive_mutex> lock(m_lock);
+
+    unsigned long long currentPosition{};
```
aballway: super nit: add int to type

Frameworks/Starboard/EbrStorageFile.cpp
```diff
@@ -0,0 +1,351 @@
...
+    if (FAILED(m_randomAccessStream->get_Position(&currentPosition))) {
```
aballway: RETURN_IF_FAILED?

Frameworks/Starboard/EbrStorageFile.cpp
```diff
@@ -0,0 +1,351 @@
...
+}
+
+int EbrStorageFile::Write(const void* src, size_t count) {
+    std::lock_guard<std::recursive_mutex> lock(m_lock);
+
+    if (!m_shouldAllowWrite) {
+        return -1;
+    }
+
+    if (!src) {
```
aballway: nit: make one if

Frameworks/Starboard/EbrStorageFile.cpp
```diff
@@ -0,0 +1,351 @@
...
+    if (FAILED(m_randomAccessStream->GetOutputStreamAt(positionToOpen, &outStream))) {
+        return -1;
+    }
+
+    ComPtr<IBuffer> dataToWrite;
+    if (FAILED(BufferFromRawData(&dataToWrite, (unsigned char*)src, count))) {
+        return -1;
+    }
+
+    UINT32 result;
```
aballway: nit: no shouty

tests/functionaltests/FunctionalTests.cpp
```diff
@@ -643,3 +643,25 @@ class UIApplicationTests {
...
+public:
+    BEGIN_TEST_CLASS(NSURLStorageFileTests)
+    TEST_CLASS_PROPERTY(L"RunAs", L"UAP")
+    TEST_CLASS_PROPERTY(L"UAP:Host", L"Xaml")
+    END_TEST_CLASS()
+
+    TEST_CLASS_SETUP(NSURLStorageFileTestsSetup) {
+        return SUCCEEDED(FrameworkHelper::RunOnUIThread(&UIApplicationDefaultInitialize));
+    }
+
```
aballway: add TEST_METHOD_CLEANUP so state doesn't leak into tests after this one

Frameworks/Foundation/NSURL.mm
```diff
@@ -890,4 +908,63 @@ - (BOOL)getPromisedItemResourceValue:(id _Nullable*)value forKey:(NSString*)key
...
+    RETURN_NULL_IF_FALSE(maxItems > 0);
+
+    uint32_t count = 0;
+    Wrappers::HString token;
+
+    static std::mutex s_lock;
+    {
+        std::lock_guard<std::mutex> guard(s_lock);
+
+        // protect the count check and addint logic with a lock since we don't want to race to add the last guy.
```
bbowman: yep commented myself on that. Looks like some comments are special and don't show up in all PR tools. Hooray innovation

Frameworks/Foundation/NSURL.mm
```diff
@@ -890,4 +908,63 @@ - (BOOL)getPromisedItemResourceValue:(id _Nullable*)value forKey:(NSString*)key
...
+            RETURN_NULL_IF_FAILED(accessList->Remove(lastEntry.Token));
+        }
+
+        RETURN_NULL_IF_FAILED(accessList->AddOverloadDefaultMetadata(comStorageItem.Get(), token.GetAddressOf()));
+    }
+
+    // now that a token has been acquired, form a file path URL similar to a file reference URL based on it.
+    return [self initFileURLWithPath:[NSString stringWithFormat:@"/.file/id=%@", [NSString _stringWithHSTRING:token.Get()]]
+                         isDirectory:NO];
+}
```
bbowman: Yep I totally agree. It would be super grand to have a clear, coherent design here. Unfortunately the public apis for this force our hand a bit. The call pattern in specific that we are seeing for this scenario is:

UIImage *image = ... ;
      if (!image && [imageURL isFileURL]) {
       image = [UIImage imageWithContentsOfFile:[imageURL path]];
      }

Which means the URL needs to return true for isFileURL and needs to interact withimageWithContentsOfFile when using just the raw path. The way to get this to interop easiest and most fully seemed to be to go ahead and use a real file URL with a custom path. Given that /.file/id=... URLs exist on the reference platform, this seemed sort of natural anyway but I definitely agree this isn't a clean design like I normally advocate for. :-(

tests/functionaltests/Tests/NSURLStorageFileTests.mm
```diff
@@ -0,0 +1,85 @@
...
+           storageFile = file;
+           [condition signal];
+         }
+
+         failure:^void(NSError* error) {
+            LOG_ERROR([[error description] UTF8String]);
+            [condition signal];
+         }];
+
+    [condition waitUntilDate:[NSDate dateWithTimeIntervalSinceNow:2]];
```
rajsesh-msft: >2 [](start = 66, length = 1)

This might sporadically fail on the build server.

include/Foundation/NSURL.h
```diff
@@ -230,4 +233,8 @@ FOUNDATION_EXPORT_CLASS
 - (BOOL)checkPromisedItemIsReachableAndReturnError:(NSError* _Nullable*)error STUB_METHOD;
 - (BOOL)getPromisedItemResourceValue:(id _Nullable*)value forKey:(NSString*)key error:(NSError* _Nullable*)error STUB_METHOD;
 - (NSDictionary*)promisedItemResourceValuesForKeys:(NSArray*)keys error:(NSError* _Nullable*)error STUB_METHOD;
+
+// MSFT Additions
```
rajsesh-msft: >MSFT [](start = 3, length = 4)

WinOBJC

tests/functionaltests/Tests/NSURLStorageFileTests.mm
```diff
@@ -0,0 +1,85 @@
...
+    NSURL* storageUrl = [NSURL URLWithStorageFile:storageFile];
+
+    // Lets write some data!!!!
+    unsigned char rawDataToWrite[] = { 'a', 'b', 'c' };
+    NSData* writeData = [NSData dataWithBytesNoCopy:rawDataToWrite length:_countof(rawDataToWrite) freeWhenDone:NO];
+    [writeData writeToURL:storageUrl options:0 error:nil];
+
+    // Ok now lets read it out!!!
+    NSData* readData = [NSData dataWithContentsOfURL:storageUrl];
+    ASSERT_TRUE([readData isEqual:writeData]);
```
DHowett-MSFT: >ASSERT_TRUE [](start = 4, length = 11)

Do we not have the OBJCEQ/OBJCNE macros in here?

tests/functionaltests/Tests/NSURLStorageFileTests.mm
```diff
@@ -0,0 +1,85 @@
...
+    __block WSStorageFile* storageFile = nil;
+    __block NSCondition* condition = [NSCondition new];
+
+    [[[WSApplicationData current] localFolder] createFileAsync:@"tempStorageFile.txt" options:WSCreationCollisionOptionOpenIfExists 
+        success:^void(WSStorageFile* file) {
+           storageFile = file;
+           [condition signal];
+         }
+
+         failure:^void(NSError* error) {
```
DHowett-MSFT: Did clang format this weird?

tests/functionaltests/FunctionalTests.cpp
DHowett-MSFT: agreed; this requires cleanup.

---
In reply to: [84942793](https://github.com/Microsoft/WinObjC/pull/1215#discussion_r84942793) [](ancestors = 84942793)

Frameworks/Starboard/EbrFile.cpp
```diff
@@ -0,0 +1,273 @@
...
+#include <Platform/EbrPlatform.h>
+#include <Starboard.h>
+#include "PathMapper.h"
+#include "LoggingNative.h"
+#include "EbrFile.h"
+#include "EbrIOFile.h"
+#include "EbrDevRandomFile.h"
+#include "EbrStorageFile.h"
+
+static const wchar_t* TAG = L"PlatformSupport";
```
DHowett-MSFT: >PlatformSupport [](start = 30, length = 15)

This should be `EbrFile` now, presumably

Frameworks/Starboard/EbrFile.cpp
```diff
@@ -0,0 +1,273 @@
...
+
+int EbrFstat(int fd, struct stat* ret) {
+    return EbrFile::GetFile(fd)->Stat(ret);
+}
+
+int EbrFstat64i32(int fd, struct _stat64i32* ret) {
+    return EbrFile::GetFile(fd)->Stat64i32(ret);
+}
+
+intptr_t EbrGetOSFHandle(int fd) {
```
DHowett-MSFT: >EbrGetOSFHandle [](start = 9, length = 15)

Does someone use this?

Frameworks/Starboard/EbrFile.cpp
```diff
@@ -0,0 +1,273 @@
...
+        sprintf_s(dirOut, maxLen, "%s\\tmp", g_WritableFolder);
+        mkdir(dirOut);
+        return true;
+    }
+    if (_stricmp(dirName, "shared") == 0) {
+        sprintf_s(dirOut, maxLen, "%s\\shared", g_WritableFolder);
+        mkdir(dirOut);
+        return true;
+    }
+    static std::regex drive("[a-zA-Z]:");
```
DHowett-MSFT: >regex [](start = 16, length = 5)

I know this isn't your code, but we're using a whole regex engine to detect a single letter and a colon??!!!??

Frameworks/Starboard/EbrIOFile.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+#include "EbrIOFile.h"
+#include "PathMapper.h"
+
+#include <sys\stat.h>
+
+
+std::shared_ptr<EbrFile> EbrIOFile::CreateInstance(const char* path, int mode, int share, int pmode) {
+    int fid{};
+    auto result = _sopen_s(&fid, CPathMapper(path), mode, share, pmode);
```
DHowett-MSFT: >auto [](start = 4, length = 4)

Only use auto when the return type cannot be easily expressed or is intuitively obvious from the righthand side.

Frameworks/Starboard/EbrIOFile.cpp
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    return std::make_shared<EbrIOFile>(fid);
+}
+
+EbrIOFile::EbrIOFile(int fid) : m_fid(fid) {
+}
+
+EbrIOFile::~EbrIOFile() {
+    if (m_fid != -1) {
```
DHowett-MSFT: >m_fid != - [](start = 8, length = 10)

How do we get an EbrIOFile with a fid of -1?

Frameworks/Starboard/EbrStorageFile.cpp
```diff
@@ -0,0 +1,351 @@
+//******************************************************************************
+//
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
rajsesh-msft: Drop the corporation here.

Frameworks/Starboard/EbrStorageFile.cpp
```diff
@@ -0,0 +1,351 @@
...
+    if (FAILED(AwaitResult(storageItem, &IStorageItem::GetBasicPropertiesAsync, &basicProperties))) {
+        return -1;
+    }
+
+    unsigned long long size{};
+    if (FAILED(basicProperties->get_Size(&size))) {
+        return -1;
+    }
+
+    // TODO 1099: Add more properites that we can fill in stat.
```
rajsesh-msft: Yes, please remove TODO.

---
In reply to: [84801640](https://github.com/Microsoft/WinObjC/pull/1215#discussion_r84801640) [](ancestors = 84801640)

msft-Jeyaram:
> #define __EBRPLATFORM_H

#pragma once

---
Refers to: Frameworks/include/Platform/EbrPlatform.h:18 in 4e817ad. [](commit_id = 4e817add0220c56ccbfecdf93f5f65a26f10e89b, deletion_comment = False)

Frameworks/Starboard/EbrDevRandomFile.cpp
```diff
@@ -0,0 +1,66 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#include "EbrDevRandomFile.h"
+
+std::shared_ptr<EbrFile> EbrDevRandomFile::CreateInstance(const char* path, int mode, int share, int pmode) {
+    if (strcmp(path, "/dev/urandom") == 0) {
```
msft-Jeyaram: >strcmp [](start = 8, length = 6)

use strncmp

Frameworks/Starboard/EbrStorageFile.cpp
```diff
@@ -0,0 +1,351 @@
...
+        shouldAppend,
+        shouldAllowWrite);
+}
+
+EbrStorageFile::EbrStorageFile(
+        ComPtr<IStorageFile>&& storageFile, 
+        ComPtr<IRandomAccessStream>&& randomAccessStream,
+        bool shouldAppend,
+        bool shouldAllowWrite) : 
+            m_storageFile(storageFile),
```
DHowett-MSFT: >storageFile [](start = 26, length = 11)

Why pass-by-rvalue-reference with std::move on the outside (keeping external consumers aware of our internals) instead of pass-by-value with std::move on the inside? There does not appear to be consensus on this on the internet at large, but it looks like a lot of people lean towards the second.

bbowman:
Review status: 0 of 22 files reviewed at latest revision, 19 unresolved discussions.

---

*[Frameworks/Foundation/NSURL.mm, line 38 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTwiVoZ8D_GXVio-r1-38:-KUwyTHu2Fy49HVKCvYL:b-v3xqc2) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Foundation/NSURL.mm#L38)):*
<details><summary><i>Previously, msft-Jeyaram (Ram) wrote…</i></summary>
> forward slashes
</details>
@msft-Jeyaram you should get a better picture.

---

*[Frameworks/Foundation/NSURL.mm, line 922 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTwiVoZ8D_GXVio-r1-922:-KUwyk9p_eHDE31NLsAm:bhw7as7) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Foundation/NSURL.mm#L922)):*
<details><summary><i>Previously, aballway (AJ Ballway) wrote…</i></summary>
> super nit: storageFile.comObj
</details>
Won't fix. I prefer message send syntax as its more clear what is going on with the objects involved. This is especially true in interop code with C++ classes like ComPtr where . operator means a very different thing.

---

*[Frameworks/Foundation/NSURL.mm, line 940 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTwiVoZ8D_GXVio-r1-940:-KUwz-Lt33RMDMWFZlDb:b-wfx10g) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Foundation/NSURL.mm#L940)):*
<details><summary><i>Previously, DHowett-MSFT (Dustin L. Howett (MSFT)) wrote…</i></summary>
> >ULL_IF_FALSE(maxItems > 0); [](start = 12, length = 27)
> 
> 
> 
> Is there a future world where they're unlimited and managed by the system? We may not want to break on those platforms.
</details>
I agree that we wouldn't want to break in that case but that is ultimately a concern for the api developer not us. I need to guard against maxItems <= 0 for any sort of reasonable for loop / boundary checks so if they want to introduce the notion of an unbounded collection, a new field would be added or something because any sentinel value for this extra logic may mess up existing code (imagine if I blindly did a while(i <= maxItems) loop or something). So won't fix. We will address that if and when new semantics are introduced but not guess at them unless you are thinking of something different that would be more robust?

---

*[Frameworks/Foundation/NSURL.mm, line 949 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTwiVoZ8D_GXVio-r1-949:-KUwzqm0NVZESskQsuto:b-896fix) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Foundation/NSURL.mm#L949)):*
<details><summary><i>Previously, DHowett-MSFT (Dustin L. Howett (MSFT)) wrote…</i></summary>
> >addint [](start = 39, length = 6)
> 
> 
> 
> nit: spelling?
</details>
Done.

---

*[Frameworks/Foundation/NSURL.mm, line 954 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTwiVoZ8D_GXVio-r1-954:-KUx-89-d4WJFTow2nPl:b-9de32g) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Foundation/NSURL.mm#L954)):*
<details><summary><i>Previously, aballway (AJ Ballway) wrote…</i></summary>
> should this remove an item when we have exactly maxItems?
</details>
It should. And does? 

---

*[Frameworks/Foundation/NSURL.mm, line 966 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTwiVoZ8D_GXVio-r1-966:-KUx-em9iOuVV_SLi1mo:b-896fix) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Foundation/NSURL.mm#L966)):*
<details><summary><i>Previously, aballway (AJ Ballway) wrote…</i></summary>
> nit: use Strings::WideToNSString rather than directly using NSString internals
</details>
Done.

---

*[Frameworks/Starboard/EbrFile.cpp, line 95 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTwiVoZ8D_GXVj2-r1-95:-KUx0V3vfwFLzXW3VCd4:b52zto1) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrFile.cpp#L95)):*
<details><summary><i>Previously, aballway (AJ Ballway) wrote…</i></summary>
> nit: flip this if so this only has one return at the end
</details>
OK. not sure I like it better but ok.

---

*[Frameworks/Starboard/EbrFile.cpp, line 109 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTwiVoZ8D_GXVj2-r1-109:-KUx0hm7RSriE8wL4Yx5:bt4wd4y) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrFile.cpp#L109)):*
<details><summary><i>Previously, aballway (AJ Ballway) wrote…</i></summary>
> do we want these to blow up when no file is found?
</details>
Hmm I guess return -1 would be nicer.

---

*[Frameworks/Starboard/EbrFile.cpp, line 168 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTwiVoZ8D_GXVj2-r1-168:-KUx1D4W9L5qa2NGeTxy:b-2g0mpv) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrFile.cpp#L168)):*
<details><summary><i>Previously, aballway (AJ Ballway) wrote…</i></summary>
> this always returns true? make a void/just return at the end if we can't change the signature for some reason
</details>
didn't mess with this guy. I want to expunge / rewrite the mapping logic too at some point but not in this review probably.

---

*[Frameworks/Starboard/EbrFile.cpp, line 253 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTwiVoZ8D_GXVj2-r1-253:-KUx1NN4XLWEgVxL4yy7:b-nywocs) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrFile.cpp#L253)):*
<details><summary><i>Previously, aballway (AJ Ballway) wrote…</i></summary>
> nit: braces
</details>
Ugh you and reviewing code that isn't mine and making me do the right thing and clean up whats around me. Darn you!!!!

---

*[Frameworks/Starboard/EbrIOFile.cpp, line 24 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTwiVoZ8D_GXVj4-r1-24:-KUx1apmI660g2QwIZeX:b-ksxv2h) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrIOFile.cpp#L24)):*
<details><summary><i>Previously, aballway (AJ Ballway) wrote…</i></summary>
> why braces here?
</details>
Hooray initializing variables!!!! If you don't do this or an assignment on release mode, the value of fid will be random stack junk until something more useful happens. As this is an out parameter, it is generally customary to initialize it yourself ahead of time if possible so that any failure conditions don't leave you with junk that you may accidentially use later. 

I generally prefer the {} methodology as it universally works in C++ (i.e. inits structs classes etc too)

---

*[Frameworks/Starboard/EbrIOFile.cpp, line 27 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTwiVoZ8D_GXVj4-r1-27:-KUx3FftGVa2DBmaqp_s:b-lakrv0) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrIOFile.cpp#L27)):*
<details><summary><i>Previously, aballway (AJ Ballway) wrote…</i></summary>
> nit: ternary return
</details>
Eh. looked less clean to me. Going to leave it.

---

*[Frameworks/Starboard/EbrStorageFile.cpp, line 95 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTxdXX6CAMrl51a-r1-95:-KUx3MB7C4pNXgSi7fkM:b-ew21pi) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrStorageFile.cpp#L95)):*
<details><summary><i>Previously, aballway (AJ Ballway) wrote…</i></summary>
> super nit: add int to type
</details>
that isn't the same ... 

---

*[Frameworks/Starboard/EbrStorageFile.cpp, line 96 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTxdXX6CAMrl51a-r1-96:-KUx3PPoir0Mx9ahMAja:bepo9ma) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrStorageFile.cpp#L96)):*
<details><summary><i>Previously, aballway (AJ Ballway) wrote…</i></summary>
> RETURN_IF_FAILED?
</details>
i didn't want to co-opt HRESULT into int returns. It is *generally* true that failure HRs are negative but that is a little different. I could add a RETURN_X_IF_FAILED but that seems maybe overkill as I don't know how many other people will need that sort of translation. 

---

*[Frameworks/Starboard/EbrStorageFile.cpp, line 250 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTxdXX6CAMrl51a-r1-250:-KUx3xpzVp7HSPqHOqmj:b-896fix) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrStorageFile.cpp#L250)):*
<details><summary><i>Previously, aballway (AJ Ballway) wrote…</i></summary>
> nit: make one if
</details>
Done.

---

*[Frameworks/Starboard/EbrStorageFile.cpp, line 279 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTxdXX6CAMrl51a-r1-279:-KUx3z61QfkMUiSVE8qy:b-3vsomx) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrStorageFile.cpp#L279)):*
<details><summary><i>Previously, aballway (AJ Ballway) wrote…</i></summary>
> nit: no shouty
</details>
used for interop with WriteAsync. Won't fix.

---

*[tests/functionaltests/FunctionalTests.cpp, line 649 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTwiVoZ8D_GXVj--r1-649:-KUx43jOUifIwDEJvS6J:b-896fix) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/tests/functionaltests/FunctionalTests.cpp#L649)):*
<details><summary><i>Previously, memontic-ms (Meredith Monticello) wrote…</i></summary>
> >UIApplicationTests [](start = 3, length = 18)
> 
> 
> 
> NSURLStorageFileTests. Also below.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/winobjc/1215)*

Frameworks/Starboard/EbrStorageFile.cpp
DHowett-MSFT: _but why_?

---
In reply to: [84938944](https://github.com/Microsoft/WinObjC/pull/1215#discussion_r84938944) [](ancestors = 84938944)

Frameworks/Starboard/EbrStorageFile.cpp
```diff
@@ -0,0 +1,351 @@
...
+        return -1;
+    }
+
+    ComPtr<IOutputStream> outStream;
+    if (FAILED(m_randomAccessStream->GetOutputStreamAt(currentPosition, &outStream))) {
+        return -1;
+    }
+
+    boolean success = false;
+    if (FAILED(AwaitResult(outStream, &IOutputStream::FlushAsync, &success))) {
```
DHowett-MSFT: >AwaitResult [](start = 15, length = 11)

Will this not break due to the vtable thunk compiler bug?

Frameworks/Starboard/EbrStorageFile.cpp
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    std::vector<unsigned char> buffer(count);
+    ComPtr<IBuffer> rawDataBuffer;
+    if (FAILED(BufferFromRawData(&rawDataBuffer, buffer.data(), buffer.size()))) {
+        return -1;
+    }
+
+    ComPtr<IBuffer> dataRead;
+    if (FAILED(AwaitProgressComplete(inStream, &IInputStream::ReadAsync, rawDataBuffer.Get(), buffer.size(), InputStreamOptions_None, &dataRead))) {
```
DHowett-MSFT: I am very surprised this works.

---
In reply to: [84969356](https://github.com/Microsoft/WinObjC/pull/1215#discussion_r84969356) [](ancestors = 84969356)

Frameworks/Starboard/EbrStorageFile.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+    // move forward the position to account for the read.
+    if (FAILED(m_randomAccessStream->Seek(length + currentPosition))) {
+        return -1;
+    }
+
+    return (int)length;
+}
+
+int EbrStorageFile::Write(const void* src, size_t count) {
```
rajsesh-msft: short circuit for count 0?

Frameworks/Starboard/EbrStorageFile.cpp
```diff
@@ -0,0 +1,351 @@
...
+    if (FAILED(BufferFromRawData(&rawDataBuffer, buffer.data(), buffer.size()))) {
+        return -1;
+    }
+
+    ComPtr<IBuffer> dataRead;
+    if (FAILED(AwaitProgressComplete(inStream, &IInputStream::ReadAsync, rawDataBuffer.Get(), buffer.size(), InputStreamOptions_None, &dataRead))) {
+        return -1;
+    }
+
+    unsigned int length = 0;
```
DHowett-MSFT: >th = 0; [](start = 21, length = 7)

Why `= 0` here and `{}` above?

Frameworks/Starboard/EbrStorageFile.cpp
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    unsigned char* rawBytes = nullptr;
+    if (FAILED(byteAccess->Buffer(&rawBytes))) {
+        return -1;
+    }
+
+    // copy over bytes into the buffer.
+    if (0 != memcpy_s(dest, count, rawBytes, length)) {
+        return -1;
```
DHowett-MSFT: It bothers me so much that ReadAsync can give you a different buffer than the one you provided, effectively eliminating nocopy optimizations. Is there something we can do to fix that?

Frameworks/Starboard/EbrStorageFile.cpp
```diff
@@ -0,0 +1,351 @@
...
+    if (FAILED(m_storageFile.As(&storageItem))) {
+        return -1;
+    }
+
+    ComPtr<IBasicProperties> basicProperties;
+    if (FAILED(AwaitResult(storageItem, &IStorageItem::GetBasicPropertiesAsync, &basicProperties))) {
+        return -1;
+    }
+
+    unsigned long long size{};
```
rajsesh-msft: >{}; [](start = 27, length = 3)

{} are overkill for scalar types in this case..

Frameworks/Starboard/EbrStorageFile.cpp
```diff
@@ -0,0 +1,351 @@
...
+    // move forward the position to account for the read.
+    if (FAILED(m_randomAccessStream->Seek(length + currentPosition))) {
+        return -1;
+    }
+
+    return (int)length;
+}
+
+int EbrStorageFile::Write(const void* src, size_t count) {
+    std::lock_guard<std::recursive_mutex> lock(m_lock);
```
DHowett-MSFT: fwrite returns a size_t, and write returns a ssize_t; I'd prefer that if we're going to pretend to be POSIX we follow their signatures. This gets rid of the cast on 289, too.

Frameworks/Starboard/EbrStorageFile.cpp
```diff
@@ -0,0 +1,351 @@
...
+    *ret = toReturn;
+
+    return 0;
+}
+
+intptr_t EbrStorageFile::GetOSFHandle() {
+    return -1;
+}
+
+int EbrStorageFile::Read(void* dest, size_t count) {
```
DHowett-MSFT: >int [](start = 0, length = 3)

same as Write

Frameworks/Starboard/EbrStorageFile.cpp
```diff
@@ -0,0 +1,351 @@
...
+            m_randomAccessStream(randomAccessStream), 
+            m_shouldAppend(shouldAppend), 
+            m_shouldAllowWrite(shouldAllowWrite) {
+
+    if (shouldAppend) {
+        Lseek(0, SEEK_END);
+    }
+}
+
+int EbrStorageFile::Tell() {
```
DHowett-MSFT: >int [](start = 0, length = 3)

Will this be enough for huge files?

Frameworks/Starboard/EbrFile.cpp
```diff
@@ -0,0 +1,273 @@
...
+int EbrFflush(int fd) {
+    return EbrFile::GetFile(fd)->Flush();
+}
+
+bool EbrRemoveEmptyDir(const char* path) {
+    return RemoveDirectoryA(CPathMapper(path));
+}
+
+bool EbrRename(const char* path1, const char* path2) {
+    return rename(CPathMapper(path1), CPathMapper(path2)) == 0;
```
msft-Jeyaram: >n rename(CPathMapper(path1), CPathMapper(path2)) == 0; [](start = 9, length = 54)

nit: Log errno if failure?

Frameworks/Starboard/EbrFile.cpp
```diff
@@ -0,0 +1,273 @@
...
+bool EbrUnlink(const char* path) {
+    return _unlink(CPathMapper(path)) == 0;
+}
+
+#define FSROOT "."
+#define mkdir _mkdir
+char g_WritableFolder[2048] = ".";
+
+void EbrSetWritableFolder(const char* folder) {
+    strcpy_s(g_WritableFolder, folder);
```
msft-Jeyaram: > strcpy_s [](start = 3, length = 9)

This can fail, log the error?

DHowett-MSFT:
:clock1:

Frameworks/Starboard/EbrFile.cpp
```diff
@@ -0,0 +1,273 @@
...
+    return rename(CPathMapper(path1), CPathMapper(path2)) == 0;
+}
+
+bool EbrUnlink(const char* path) {
+    return _unlink(CPathMapper(path)) == 0;
+}
+
+#define FSROOT "."
+#define mkdir _mkdir
+char g_WritableFolder[2048] = ".";
```
msft-Jeyaram: >2048 [](start = 22, length = 4)

where does this value come from?
MAX_PATH?

Frameworks/Starboard/EbrFile.cpp
```diff
@@ -0,0 +1,273 @@
...
+    return g_WritableFolder;
+}
+
+bool EbrGetRootMapping(const char* dirName, char* dirOut, uint32_t maxLen) {
+    if (dirName == NULL) {
+        strcpy_s(dirOut, maxLen, FSROOT);
+        return true;
+    }
+    if (_stricmp(dirName, "Documents") == 0) {
+        sprintf_s(dirOut, maxLen, "%s\\Documents", g_WritableFolder);
```
msft-Jeyaram: >%s\\Documents" [](start = 35, length = 14)

Put these in an internal header?
These are being repeated in other files, thus is would make sense to do so. Also this would make future maintenance easier

Frameworks/Starboard/EbrFile.cpp
```diff
@@ -0,0 +1,273 @@
...
+    strcpy_s(g_WritableFolder, folder);
+}
+
+const char* EbrGetWritableFolder() {
+    return g_WritableFolder;
+}
+
+bool EbrGetRootMapping(const char* dirName, char* dirOut, uint32_t maxLen) {
+    if (dirName == NULL) {
+        strcpy_s(dirOut, maxLen, FSROOT);
```
msft-Jeyaram: >strcpy_s [](start = 8, length = 8)

what is the copy fails? we end up returning true regardless.

Frameworks/Starboard/EbrFile.cpp
```diff
@@ -0,0 +1,273 @@
...
+            }
+        } else if (s.st_mode & S_IFDIR) {
+            EbrDir* dir = EbrOpenDir(path);
+            if (dir) {
+                EbrDirEnt ent;
+                while (EbrReadDir(dir, &ent)) {
+                    if (strcmp(ent.fileName, ".") == 0 || strcmp(ent.fileName, "..") == 0)
+                        continue;
+
+                    char fullPath[4096]; // max path?
```
msft-Jeyaram: >4096 [](start = 34, length = 4)

MAX_PATH?

Frameworks/Starboard/EbrIOFile.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include <sys\stat.h>
+
+
+std::shared_ptr<EbrFile> EbrIOFile::CreateInstance(const char* path, int mode, int share, int pmode) {
+    int fid{};
+    auto result = _sopen_s(&fid, CPathMapper(path), mode, share, pmode);
+
+    if (result != 0) {
+        return nullptr;
+    }
```
msft-Jeyaram: if(_sopen_s(&fid, CPathMapper(path), mode, share, pmode) != 0 ) {
  return nullptr;
}

Frameworks/Starboard/EbrStorageFile.cpp
```diff
@@ -0,0 +1,351 @@
...
+using namespace ABI::Windows::Storage;
+using namespace ABI::Windows::Storage::AccessCache;
+using namespace ABI::Windows::Storage::FileProperties;
+using namespace ABI::Windows::Storage::Streams;
+
+std::shared_ptr<EbrFile> EbrStorageFile::CreateInstance(const char* path, int mode, int share, int pmode) {
+    // Check to see if this path is a file id path.
+    static const std::string c_fileIdPreamble = "/.file/id=";
+
+    RETURN_NULL_IF_FALSE(0 == strncmp(path, c_fileIdPreamble.c_str() , c_fileIdPreamble.length()));
```
msft-Jeyaram: >path [](start = 38, length = 4)

invalid path pointer?

bbowman:
Review status: 0 of 22 files reviewed at latest revision, 41 unresolved discussions.

---

*[Frameworks/Starboard/EbrDevRandomFile.cpp, line 20 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTwiVoZ8D_GXVj0-r1-20:-KUxBRmQoBs_2fplCCt4:b-896fix) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrDevRandomFile.cpp#L20)):*
<details><summary><i>Previously, msft-Jeyaram (Ram) wrote…</i></summary>
> >strcmp [](start = 8, length = 6)
> 
> 
> 
> use strncmp 
</details>
Done.

---

*[Frameworks/Starboard/EbrDevRandomFile.h, line 19 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTwiVoZ8D_GXVj1-r1-19:-KUxAycUr-ZPU3AA_pux:bh2w7wu) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrDevRandomFile.h#L19)):*
<details><summary><i>Previously, msft-Jeyaram (Ram) wrote…</i></summary>
> >include [](start = 1, length = 7)
> 
> 
> 
> nit: Import
</details>
meh. won't fix. These guys are all straight c++ and just happen to be clang compiled for funsies. No need to introduce objC isms if we don't need it.

---

*[Frameworks/Starboard/EbrFile.cpp, line 36 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTwiVoZ8D_GXVj2-r1-36:-KUxBGEhsRUKgBTo8F5c:b-7j86r0) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrFile.cpp#L36)):*
<details><summary><i>Previously, DHowett-MSFT (Dustin L. Howett (MSFT)) wrote…</i></summary>
> >PlatformSupport [](start = 30, length = 15)
> 
> 
> 
> This should be `EbrFile` now, presumably
</details>
sure.

---

*[Frameworks/Starboard/EbrFile.cpp, line 116 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTwiVoZ8D_GXVj2-r1-116:-KUx6Sg6JiqrshYOew9r:bb7sa0c) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrFile.cpp#L116)):*
<details><summary><i>Previously, DHowett-MSFT (Dustin L. Howett (MSFT)) wrote…</i></summary>
> >EbrGetOSFHandle [](start = 9, length = 15)
> 
> 
> 
> Does someone use this?
</details>
yes.

---

*[Frameworks/Starboard/EbrFile.cpp, line 208 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTwiVoZ8D_GXVj2-r1-208:-KUx6XA8DhpqybUwmRlQ:b6kppj0) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrFile.cpp#L208)):*
<details><summary><i>Previously, DHowett-MSFT (Dustin L. Howett (MSFT)) wrote…</i></summary>
> >regex [](start = 16, length = 5)
> 
> 
> 
> I know this isn't your code, but we're using a whole regex engine to detect a single letter and a colon?!
</details>
apparently yes.

---

*[Frameworks/Starboard/EbrIOFile.cpp, line 25 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTwiVoZ8D_GXVj4-r1-25:-KUx6dfgEerVY_6_U8LV:bb7g23t) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrIOFile.cpp#L25)):*
<details><summary><i>Previously, DHowett-MSFT (Dustin L. Howett (MSFT)) wrote…</i></summary>
> >auto [](start = 4, length = 4)
> 
> 
> 
> Only use auto when the return type cannot be easily expressed or is intuitively obvious from the righthand side.
</details>
fair

---

*[Frameworks/Starboard/EbrIOFile.cpp, line 38 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTwiVoZ8D_GXVj4-r1-38:-KUx6q_Wwbd_DC4XHgk7:b-u3pz1n) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrIOFile.cpp#L38)):*
<details><summary><i>Previously, DHowett-MSFT (Dustin L. Howett (MSFT)) wrote…</i></summary>
> >m_fid != - [](start = 8, length = 10)
> 
> 
> 
> How do we get an EbrIOFile with a fid of -1?
</details>
well its passed in from the ctor so it could presumably happen. Given the new pattern of CreateInstance it shouldn't unless _sopen_s returns 0 but fid is assigned -1.

---

*[Frameworks/Starboard/EbrStorageFile.cpp, line 3 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTxdXX6CAMrl51a-r1-3:-KUxBO0q6lSwY0wpsJLm:b-896fix) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrStorageFile.cpp#L3)):*
<details><summary><i>Previously, rajsesh-msft (Raj Seshasankaran) wrote…</i></summary>
> Drop the corporation here.
</details>
Done.

---

*[Frameworks/Starboard/EbrStorageFile.cpp, line 78 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTxdXX6CAMrl51a-r1-78:-KUx7HhV_WsP_EfC9TN6:b-7e7unr) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrStorageFile.cpp#L78)):*
<details><summary><i>Previously, DHowett-MSFT (Dustin L. Howett (MSFT)) wrote…</i></summary>
> >storageFile [](start = 26, length = 11)
> 
> 
> 
> Why pass-by-rvalue-reference with std::move on the outside (keeping external consumers aware of our internals) instead of pass-by-value with std::move on the inside? There does not appear to be consensus on this on the internet at large, but it looks like a lot of people lean towards the second.
</details>
The reasoning here is to save an addRef Release pair. The consensus I think we should go with is:

1. Use raw * / ** at ABI boundary.
2. use const ComPtr<T>& as a non transferring input argument
3. use ComPtr<T>&& to transfer ownership. 

3 could also be a Detach Attach here but as its internal to this module (and I think imma go ahead and make the ctors private to enforce CreateInstance use) it seems cleaner to keep the jacket and pass it off rather than unwrap and rewrap around the callsite?

---

*[Frameworks/Starboard/EbrStorageFile.cpp, line 106 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTxdXX6CAMrl51a-r1-106:-KUx8Lxlv21AKlY9lwBv:btregyi) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrStorageFile.cpp#L106)):*
<details><summary><i>Previously, DHowett-MSFT (Dustin L. Howett (MSFT)) wrote…</i></summary>
> >AwaitResult [](start = 15, length = 11)
> 
> 
> 
> Will this not break due to the vtable thunk compiler bug?
</details>
No! (Well not anymore). This file is ClCompile. I made that change while trying to figure out what was up. 

---

*[Frameworks/Starboard/EbrStorageFile.cpp, line 137 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTxdXX6CAMrl51a-r1-137:-KUx8d4JDWZWdTv-4U6m:b-896fix) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrStorageFile.cpp#L137)):*
<details><summary><i>Previously, rajsesh-msft (Raj Seshasankaran) wrote…</i></summary>
> Yes, please remove TODO.
> 
> 
> 
> ---
> 
> In reply to: [84801640](https://github.com/Microsoft/WinObjC/pull/1215#discussion_r84801640) [](ancestors = 84801640)
</details>
Done.

---

*[Frameworks/Starboard/EbrStorageFile.cpp, line 149 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTxdXX6CAMrl51a-r1-149:-KUx8cLjIFsswbzHZKno:b-896fix) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrStorageFile.cpp#L149)):*
<details><summary><i>Previously, bbowman wrote…</i></summary>
> consider a templated internal function to remove duplicated stat code.
</details>
Done.

---

*[Frameworks/Starboard/EbrStorageFile.cpp, line 167 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTxdXX6CAMrl51a-r1-167:-KUx8ehyfpGeXgWClb6O:bv1efak) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrStorageFile.cpp#L167)):*
<details><summary><i>Previously, rajsesh-msft (Raj Seshasankaran) wrote…</i></summary>
> >{}; [](start = 27, length = 3)
> 
> 
> 
> {} are overkill for scalar types in this case..
</details>
not really? still needs initialized and this is a universal way to do it.

---

*[Frameworks/Starboard/EbrStorageFile.cpp, line 211 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTxdXX6CAMrl51a-r1-211:-KUx8k36_fWfFhzV05vu:b-6wt90n) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrStorageFile.cpp#L211)):*
<details><summary><i>Previously, DHowett-MSFT (Dustin L. Howett (MSFT)) wrote…</i></summary>
> I am very surprised this works.
> 
> 
> 
> ---
> 
> In reply to: [84969356](https://github.com/Microsoft/WinObjC/pull/1215#discussion_r84969356) [](ancestors = 84969356)
</details>
Magic! (ClCompile as per other comment. Super not obvious due to change being hidden in vcxproj goop)

---

*[Frameworks/Starboard/EbrStorageFile.cpp, line 215 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTxdXX6CAMrl51a-r1-215:-KUx8rMnshDVvRLLdd26:b1ptlf) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrStorageFile.cpp#L215)):*
<details><summary><i>Previously, DHowett-MSFT (Dustin L. Howett (MSFT)) wrote…</i></summary>
> >th = 0; [](start = 21, length = 7)
> 
> 
> 
> nit: Why `= 0` here and `{}` above?
</details>
Cause I'm fickle. (or more likely I copy pasta'ed from elsewhere)

---

*[Frameworks/Starboard/EbrStorageFile.cpp, line 232 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTxdXX6CAMrl51a-r1-232:-KUx94s0yfoQXPTWFmf7:b5xr9ji) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrStorageFile.cpp#L232)):*
<details><summary><i>Previously, DHowett-MSFT (Dustin L. Howett (MSFT)) wrote…</i></summary>
> It bothers me so much that ReadAsync can give you a different buffer than the one you provided, effectively eliminating nocopy optimizations. Is there something we can do to fix that?
</details>
Bitch and moan to the API implementors to make a less shitty interface in a future version.

---

*[Frameworks/Starboard/EbrStorageFile.cpp, line 243 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTxdXX6CAMrl51a-r1-243:-KUx9F0BZvZLMBG5QtZM:b5955g9) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrStorageFile.cpp#L243)):*
<details><summary><i>Previously, rajsesh-msft (Raj Seshasankaran) wrote…</i></summary>
> short circuit for count 0?
</details>
sure why not.

---

*[Frameworks/Starboard/EbrStorageFile.cpp, line 244 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTxdXX6CAMrl51a-r1-244:-KUx9UC45H1pXz8_2r5w:bmomy2m) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrStorageFile.cpp#L244)):*
<details><summary><i>Previously, DHowett-MSFT (Dustin L. Howett (MSFT)) wrote…</i></summary>
> fwrite returns a size_t, and write returns a ssize_t; I'd prefer that if we're going to pretend to be POSIX we follow their signatures. This gets rid of the cast on 289, too.
</details>
yeah there were a few signature weirdnesses that I didn't address here as some will spill out to the public interface. Like int64 seek accepting an int64 arg but not returning an int64 position. Hilarious. I'll be a good dev I guess. You guys are just so demanding :-p

---

*[Frameworks/Starboard/EbrStorageFile.cpp, line 279 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTxdXX6CAMrl51a-r1-279:-KUx9odE_d7-I9uG8krB:b-3cslek) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrStorageFile.cpp#L279)):*
<details><summary><i>Previously, DHowett-MSFT (Dustin L. Howett (MSFT)) wrote…</i></summary>
> >result [](start = 11, length = 6)
> 
> 
> 
> This probably wants to be initialized.
</details>
indeed.

---

*[include/Foundation/NSURL.h, line 237 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTwiVoZ8D_GXViz-r1-237:-KUxBR2aMHkxuwengBJp:b-896fix) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/include/Foundation/NSURL.h#L237)):*
<details><summary><i>Previously, rajsesh-msft (Raj Seshasankaran) wrote…</i></summary>
> >MSFT [](start = 3, length = 4)
> 
> 
> 
> WinOBJC
</details>
Done.

---

*[tests/functionaltests/Tests/NSURLStorageFileTests.mm, line 35 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTxdXX6CAMrl51d-r1-35:-KUxA2IyV6IFCqQ30AkB:b9smfh9) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/tests/functionaltests/Tests/NSURLStorageFileTests.mm#L35)):*
<details><summary><i>Previously, DHowett-MSFT (Dustin L. Howett (MSFT)) wrote…</i></summary>
> Did clang format this weird?
</details>
clang doesn't format for me anymore :-( someone was going to check something in to fix that but I haven't pulled recently.

---

*[tests/functionaltests/Tests/NSURLStorageFileTests.mm, line 40 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTxdXX6CAMrl51d-r1-40:-KUxABMkWR3EKHmI9R3G:b-98x650) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/tests/functionaltests/Tests/NSURLStorageFileTests.mm#L40)):*
<details><summary><i>Previously, rajsesh-msft (Raj Seshasankaran) wrote…</i></summary>
> >2 [](start = 66, length = 1)
> 
> 
> 
> This might sporadically fail on the build server.
</details>
can increase to 10? should never hit timeout but just so tests dont hang.

---

*[tests/functionaltests/Tests/NSURLStorageFileTests.mm, line 65 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTxdXX6CAMrl51d-r1-65:-KUxALiNG9oPin1JU8eV:b-q5d7ur) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/tests/functionaltests/Tests/NSURLStorageFileTests.mm#L65)):*
<details><summary><i>Previously, DHowett-MSFT (Dustin L. Howett (MSFT)) wrote…</i></summary>
> >ASSERT_TRUE [](start = 4, length = 11)
> 
> 
> 
> Do we not have the OBJCEQ/OBJCNE macros in here?
</details>
Hilariously no. Not for TAEF functional test interop with gtest macros.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/winobjc/1215)*

Frameworks/Foundation/NSURL.mm
```diff
@@ -890,4 +908,63 @@ - (BOOL)getPromisedItemResourceValue:(id _Nullable*)value forKey:(NSString*)key
...
+    ComPtr<IStorageApplicationPermissionsStatics> storageApplicationPermissionsStatics;
+    RETURN_NULL_IF_FAILED(
+        GetActivationFactory(Wrappers::HStringReference(RuntimeClass_Windows_Storage_AccessCache_StorageApplicationPermissions).Get(),
+                             &storageApplicationPermissionsStatics));
+
+    ComPtr<IStorageItemAccessList> accessList;
+    RETURN_NULL_IF_FAILED(storageApplicationPermissionsStatics->get_FutureAccessList(&accessList));
+
+    // See how close we are to the cap.
+    uint32_t maxItems = 0;
```
rajsesh-msft: it would help to move this to a private function, so the init method isn't doing too many things..

rajsesh-msft:
:clock1:

Frameworks/Starboard/EbrStorageFile.cpp
```diff
@@ -0,0 +1,351 @@
...
+                             &storageApplicationPermissionsStatics));
+
+    ComPtr<IStorageItemAccessList> accessList;
+    RETURN_NULL_IF_FAILED(storageApplicationPermissionsStatics->get_FutureAccessList(&accessList));
+
+    // Try to grab the item based on id.
+    ComPtr<IStorageFile> storageFile;
+    RETURN_NULL_IF_FAILED(AwaitResult(accessList, &IStorageItemAccessList::GetFileAsync, idHString.Get(), &storageFile));
+
+    bool shouldAllowWrite = ((mode & _O_WRONLY) != 0) ||
```
msft-Jeyaram: >mode [](start = 30, length = 4)

can it be _O_APPEND?

Frameworks/Starboard/EbrStorageFile.cpp
msft-Jeyaram: > if (FAILED(m_randomAccessStream->get_Position(&currentPosition))) { [](start = 3, length = 68)

RETURN_RESULT_IF_FAILED

msft-Jeyaram: I just realized this change is not in the repo. Still doesn't hurt to introduce it.

---
In reply to: [84979111](https://github.com/Microsoft/WinObjC/pull/1215#discussion_r84979111) [](ancestors = 84979111)

Frameworks/Starboard/EbrStorageFile.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+    ComPtr<IBufferByteAccess> byteAccess;
+    if (FAILED(dataRead.As(&byteAccess))) {
+        return -1;
+    }
+
+    unsigned char* rawBytes = nullptr;
+    if (FAILED(byteAccess->Buffer(&rawBytes))) {
+        return -1;
+    }
```
msft-Jeyaram: again, the macro introduction of RETURN_RESULT_IF_FAILED(.......,-1);
would help allot here.

msft-Jeyaram:
Review status: 0 of 22 files reviewed at latest revision, 55 unresolved discussions.

---

*[Frameworks/Starboard/EbrFile.cpp, line 3 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUxF-gxaXyOsE0jrdib:-KUxF-gyK5WuKI562OfP:b-u7so84) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrFile.cpp#L3)):*
> ```cpp
> //******************************************************************************
> //
> // Copyright (c) Microsoft Corporation. All rights reserved.
> ```

drop corporation

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/winobjc/1215)*

include/Foundation/NSURL.h
msft-Jeyaram: >MSFT [](start = 3, length = 4)

WINOBJC?

tests/functionaltests/Tests/NSURLStorageFileTests.mm
```diff
@@ -0,0 +1,85 @@
...
+        [deleterCondition waitUntilDate:[NSDate dateWithTimeIntervalSinceNow:2]];
+    });
+
+    // Ok all the logistics are taken care of. Lets get to work on the meat of the test. mmmmm URLs.
+    NSURL* storageUrl = [NSURL URLWithStorageFile:storageFile];
+
+    // Lets write some data!!!!
+    unsigned char rawDataToWrite[] = { 'a', 'b', 'c' };
+    NSData* writeData = [NSData dataWithBytesNoCopy:rawDataToWrite length:_countof(rawDataToWrite) freeWhenDone:NO];
+    [writeData writeToURL:storageUrl options:0 error:nil];
```
msft-Jeyaram: >  [writeData writeToURL:storageUrl options:0 error:nil]; [](start = 2, length = 56)

ASSERT_TRUE([writeData writeToURL:storageUrl options:0 error:nil]);

aballway:
*[Frameworks/Starboard/EbrStorageFile.cpp, line 279 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/1215#-KUwyPTxdXX6CAMrl51a-r1-279:-KUxJBzdHytoaw7X0aFf:b-2gl8rv) ([raw file](https://github.com/microsoft/winobjc/blob/4e817add0220c56ccbfecdf93f5f65a26f10e89b/Frameworks/Starboard/EbrStorageFile.cpp#L279)):*
<details><summary><i>Previously, bbowman wrote…</i></summary>
> indeed.
</details>
uint32_t then?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/winobjc/1215)*

tests/functionaltests/Tests/NSURLStorageFileTests.mm
```diff
@@ -0,0 +1,85 @@
...
+    unsigned char rawDataToWrite2[] = { 'e', 'f', 'g' };
+    writeData = [NSData dataWithBytesNoCopy:rawDataToWrite2 length:_countof(rawDataToWrite2) freeWhenDone:NO];
+    [writeData writeToURL:storageUrl options:0 error:nil];
+
+    readData = [NSData dataWithContentsOfURL:storageUrl];
+    ASSERT_TRUE([readData isEqual:writeData]);
+
+    // Now with appending!
+    writeData = [NSData dataWithBytesNoCopy:rawDataToWrite length:_countof(rawDataToWrite) freeWhenDone:NO];
+    [writeData writeToURL:storageUrl options:NSDataWritingWithoutOverwriting error:nil];
```
msft-Jeyaram: >writeData writeToURL:storageUrl [](start = 5, length = 31)

check for success, as here and other writes

msft-Jeyaram:
:clock1:

tests/functionaltests/Tests/NSURLSession.mm
```diff
@@ -598,18 +598,21 @@ - (void)URLSession:(NSURLSession*)session downloadTask:(NSURLSessionDownloadTask
...
                                                         }];
     [downloadTask resume];
 
     // Wait for download to complete.
-    if (![condition waitUntilDate:[NSDate dateWithTimeIntervalSinceNow:c_testDownloadTimeoutInSec]]) {
-        // Wait timed out.
-        ASSERT_FALSE_MSG(true, "FAILED: Waiting for download timed out!");
-    }
+    [condition lock];
+    ASSERT_TRUE_MSG( (downloadLocation || downloadResponse || downloadError) ||
```
bbowman: > ASSERT_TRUE_MSG( (downloadLocation || downloadResponse || downloadError) ||  [](start = 3, length = 77)

some of these don't handle spurious wakeups well but its at least a lot better than before. NSCondition needs locking and if you signal before a wait it just goes into the ether so needed another piece of information to check against.

Frameworks/include/Platform/EbrPlatform.h
```diff
@@ -38,17 +38,14 @@ SB_EXPORT int EbrFflush(int fd);
 SB_EXPORT int EbrOpen(const char* file, int mode, int share);
 SB_EXPORT int EbrOpenWithPermission(const char* file, int mode, int share, int pmode);
 SB_EXPORT int EbrClose(int fd);
-SB_EXPORT int EbrFd2Host(int fd);
 SB_EXPORT int EbrFstat(int fd, struct stat* ret);
 SB_EXPORT int EbrFstat64i32(int fd, struct _stat64i32* ret);
 SB_EXPORT int EbrRead(int fd, void* dest, size_t count);
 SB_EXPORT int EbrWrite(int fd, const void* src, size_t count);
 SB_EXPORT int EbrLseek(int fd, __int64 pos, int whence);
```
bbowman: >int [](start = 10, length = 3)

@DHowett-MSFT @msft-Jeyaram  still working on this bit. Got very much sidetracked from test failures due to ridiculously wrong NSCondition use.

Frameworks/Starboard/EbrStorageFile.cpp
bbowman: _read and _write are the msdn ISO c++ compliant methods used and those are int based so I'm going to leave it. 

---
In reply to: [84970060](https://github.com/Microsoft/WinObjC/pull/1215#discussion_r84970060) [](ancestors = 84970060)

PR #1220, opened 2016-10-25
Use surface format to pick the wic bitmap format for render target

### LINKED ISSUE #1216: Internal graphics test app has R and B text color components swapped

rajsesh-msft:
The bitmap render targets we were creating always used same pixel format(GUID_WICPixelFormat32bppPBGRA) regardless of the surface type of the
bitmap context.

This change is to use the surface format as the guide in choosing the
pixel format.  Incidentally, this also fixes other text issues in the
graphics app, where the library was using grayscale context to render
black text.

Fixes #1216

### DIFF
--- a/tests/unittests/CoreGraphics/CGBitmapContextTests.mm
+++ b/tests/unittests/CoreGraphics/CGBitmapContextTests.mm
@@ -66,3 +66,104 @@ TEST(CGBitmapContext, BitmapInfoAPIs_Gray) {
     CGColorSpaceRelease(grayColorSpace);
     CGContextRelease(context);
 }
+
+void _DrawText(CGContextRef context, NSString* text, const CGRect bounds) {
+    // Aligns origin for our frame
+    CGContextTranslateCTM(context, 0.0f, bounds.size.height);
+
+    // Flips y-axis for our frame
+    CGContextScaleCTM(context, 1.0f, -1.0f);
+
+    // Creates path with current rectangle
+    CGMutablePathRef path = CGPathCreateMutable();
+    CGPathAddRect(path, NULL, bounds);
+
+    CTFontRef myCFFont = CTFontCreateWithName((__bridge CFStringRef) @"Arial", 10, NULL);
+    CFAutorelease(myCFFont);
+
+    // Make dictionary for attributed string with font, color, and alignment
+    NSDictionary* attributesDict = [NSDictionary dictionaryWithObjectsAndKeys:(__bridge id)myCFFont,
+                                                                              (id)kCTFontAttributeName,
+                                                                              kCFBooleanTrue,
+                                                                              (id)kCTForegroundColorFromContextAttributeName,
+                                                                              nil];
+
+    CFAttributedStringRef attrString =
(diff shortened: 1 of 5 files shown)


rajsesh-msft:

rajsesh-msft:

ms-jihua:
:shipit:

tests/unittests/CoreGraphics/CGBitmapContextTests.mm
```diff
@@ -65,4 +65,105 @@
...
+void _DrawText(CGContextRef context, NSString* text, const CGRect bounds) {
+    // Aligns origin for our frame
+    CGContextTranslateCTM(context, 0.0f, bounds.size.height);
+
+    // Flips y-axis for our frame
+    CGContextScaleCTM(context, 1.0f, -1.0f);
+
+    // Creates path with current rectangle
+    CGMutablePathRef path = CGPathCreateMutable();
+    CGPathAddRect(path, NULL, bounds);
```
aballway: needs to be released

tests/unittests/CoreGraphics/CGBitmapContextTests.mm
```diff
@@ -65,4 +65,105 @@
...
+    CGColorSpaceRef colorSpace = CGColorSpaceCreateDeviceRGB();
+
+    const size_t width = bounds.size.width;
+    const size_t height = bounds.size.height;
+    const size_t bytesPerPixel = 4;
+    const size_t bitsPerChannel = 8;
+    const CGFloat color[] = { 1.0, 0.5, 0.0, 1.0 };
+    const CGFloat bgColor[] = { 0.0, 0.0, 0.5, 1.0 };
+
+    woc::unique_iw<uint8_t> data(static_cast<uint8_t*>(IwMalloc(width * height * 4)));
```
aballway: why not std::vector here?  and will unique_ptr properly delete the malloced array when created only as a pointer?

tests/unittests/CoreGraphics/CGBitmapContextTests.mm
DHowett-MSFT: or put into a `woc::unique_cf`! :)

---
In reply to: [84944101](https://github.com/Microsoft/WinObjC/pull/1220#discussion_r84944101) [](ancestors = 84944101)

tests/unittests/CoreGraphics/CGBitmapContextTests.mm
DHowett-MSFT: I don't understand the second question, but: unique_iw lacks a `T[]` specialization (projected fix [here](https://github.com/DHowett-MSFT/WinObjC/commit/335cc3e1df4cb276a31f9b54df1de9492c4419bd)), so we need to use the pointer version.

---
In reply to: [84944596](https://github.com/Microsoft/WinObjC/pull/1220#discussion_r84944596) [](ancestors = 84944596)

rajsesh-msft: IwMalloc is similar to malloc and not new, so the array operator on delete is not an issue here (it calls IwFree)

---
In reply to: [84955480](https://github.com/Microsoft/WinObjC/pull/1220#discussion_r84955480) [](ancestors = 84955480,84944596)

DHowett-MSFT:
:shipit:

PR #1223, opened 2016-10-25
Fix adjustment in nsstring uikitadditions

### LINKED ISSUE #1218: Internal UIKit test app broken in CT-DW branch

aballway:
Text position adjustment should be 1/2 of ascent, not including descent.

Fixes #1218


ms-jihua:
:shipit:

DHowett-MSFT:
:shipit:

PR #1243, opened 2016-10-27
Properly get font family name for UILabels

### LINKED ISSUE #1217: Internal graphics test app text layout is not correct

aballway:
Getting the Font Family Name from DWrite would for certain fonts return a different value than Xaml expects, causing Segue to be used as a default.

Partially fixes #1217

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1243)


Frameworks/UIKit/StarboardXaml/CompositorInterface.mm
```diff
@@ -332,7 +333,12 @@ void SetParams(UIFont* font,
         _isBold = (mask & UIFontDescriptorTraitBold) > 0;
         _isItalic = (mask & UIFontDescriptorTraitItalic) > 0;
         std::wstring wideBuffer = Strings::NarrowToWide<std::wstring>(text);
-        ConstructGlyphs([[font fontName] UTF8String], wideBuffer.c_str(), wideBuffer.length());
```
rajsesh-msft: Why not font fontFamilyName?  We should be returning the right family name there.

aballway: fontFamilyName is the function that returns the incompatible value

rajsesh-msft: The question I am really asking is, why aren't we fixing that method?

aballway: The method returns the "correct" value, which is taken directly from the font file as I understand it. The issue is Xaml expects the family name to be a different value than the one on the file in some cases.  A developer may expect familyName to be the actual value from the file, in which case changing it in the CoreText/UIKit calls would be incorrect.

rajsesh-msft: I don't know if it is "correct" if it works with one framework and not the other.  If you really want to keep this separate go with a UIFont private property that rest of UIKit can use where it integrates with xaml.

The correct fix IMO would be for coretext to also handle font family names like xaml does (xaml is also built on dwrite), but that would likely be bigger change than what we want to do at this point.

jaredhms:
> class DisplayTextureXamlGlyphs : public DisplayTexture {

Please don't make any changes to the compositor; this class doesn't even exist anymore in the refactor branch ([https://github.com/jaredhms/WinObjC/tree/compositor_latest2](https://github.com/jaredhms/WinObjC/tree/compositor_latest2)).  Let's discuss how I can best roll this into my changes; it looks like I should be calling _DWriteGetFamilyNameForFontName when I need to grab the font family name?

---
Refers to: Frameworks/UIKit/StarboardXaml/CompositorInterface.h:150 in e193e65. [](commit_id = e193e65752af5c85b868a909848b992943df5f3a, deletion_comment = False)

jaredhms:
:shipit:

jaredhms:
> class DisplayTextureXamlGlyphs : public DisplayTexture {

I talked with Raj offline; we can take this if needed, and I already have a fix for it queued up in my branch.

---
In reply to: [256707554](https://github.com/Microsoft/WinObjC/pull/1243#issuecomment-256707554) [](ancestors = 256707554)

---
Refers to: Frameworks/UIKit/StarboardXaml/CompositorInterface.h:150 in e193e65. [](commit_id = e193e65752af5c85b868a909848b992943df5f3a, deletion_comment = False)

msft-Jeyaram:

update 

---
Refers to: Frameworks/UIKit/StarboardXaml/CompositorInterface.mm:4 in e193e65. [](commit_id = e193e65752af5c85b868a909848b992943df5f3a, deletion_comment = False)

aballway:

this file is going away

---
In reply to: [256725429](https://github.com/Microsoft/WinObjC/pull/1243#issuecomment-256725429) [](ancestors = 256725429)

---
Refers to: Frameworks/UIKit/StarboardXaml/CompositorInterface.mm:4 in e193e65. [](commit_id = e193e65752af5c85b868a909848b992943df5f3a, deletion_comment = False)

tests/unittests/CoreGraphics/DWriteWrapperTests.mm
```diff
@@ -32,4 +32,5 @@
     EXPECT_OBJCEQ(@"Times New Roman", (id)_DWriteGetFamilyNameForFontName(CFSTR("Times New Roman Italic")));
     EXPECT_OBJCEQ(@"Times New Roman", (id)_DWriteGetFamilyNameForFontName(CFSTR("Times New Roman Bold")));
     EXPECT_OBJCEQ(@"Times New Roman", (id)_DWriteGetFamilyNameForFontName(CFSTR("Times New Roman Bold Italic")));
+    EXPECT_OBJCEQ(@"Franklin Gothic", (id)_DWriteGetFamilyNameForFontName(CFSTR("Franklin Gothic Medium Cond")));
```
rajsesh-msft: I second this.

Frameworks/UIKit/StarboardXaml/CompositorInterface.mm
ms-jihua: Let me try to clarify on correctness here. [UIFont familyName] for this on the reference platform would return "Franklin Gothic Medium Cond". This is the same with XAML and DWrite, as well as several websites such as http://fontsgeek.com/fonts/Franklin-Gothic-Medium-Cond-Regular which enumerate family names for fonts, suggesting that it's a property of the ttf file itself.

However, XAML/DWrite only accept "Franklin Gothic" as the _input_ family name, even though they give something else as output. This is fundamentally a mismatch on their end, so I don't know that there's anything "more correct" than this UIFont private function right now.

ms-jihua:
:shipit:

Frameworks/UIKit/StarboardXaml/CompositorInterface.h
```diff
@@ -167,7 +167,7 @@ class DisplayTextureXamlGlyphs : public DisplayTexture {
 
     float _desiredWidth, _desiredHeight;
     void Measure(float width, float height);
-    void ConstructGlyphs(const char* fontName, const wchar_t* str, int len);
+    void ConstructGlyphs(Microsoft::WRL::Wrappers::HString fontFamilyName, const wchar_t* str, int len);
```
bbowman: This needs to be a const & otherwise it's pass by Val which will copy the inner string

Issue #1233, opened 2016-10-26
[UIViewController presentedViewController] will return wrong view controller


memontic-ms:
Current Implementation: will return the view controller that presented it, the view controller that presented its parent, or nil.

Might just be a spelling mistake between presentingViewController and presentedViewController, but its ancestor view controllers also need to be checked.

oliversa-msft:
Looks like a copy-paste bug. A reasonable person would assume the presentedVC property would return the _presentedVC controller.

PR #1247, opened 2016-10-27
Fix UIViewController presentedViewController to return a presented view controller

memontic-ms:
presentedViewController was previously returning a presentingViewController in all cases. Now it returns a presentedViewController.
Fix: #1233.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1247)

### DIFF
--- a/Frameworks/UIKit/UIViewController.mm
+++ b/Frameworks/UIKit/UIViewController.mm
@@ -2194,11 +2194,13 @@ static UIInterfaceOrientation findOrientation(UIViewController* self) {
  @Status Interoperable
 */
 - (UIViewController*)presentedViewController {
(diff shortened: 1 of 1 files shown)


oliversa-msft:
Great stuff! Can we verify with WOCCatalog as well - ControlsViewController references it as well.

oliversa-msft:
:shipit:

jaredhms:
:shipit:

memontic-ms:
Checked functionality of ControlsViewController in WOCCatalog. This change still passes the assert that uses presentedViewController.

---
In reply to: [256765411](https://github.com/Microsoft/WinObjC/pull/1247#issuecomment-256765411) [](ancestors = 256765411)

PR #1252, opened 2016-10-27
Fix Sample App with CAF Decoder

bbowman:
Description:
A recent change regressed the CAF decoder work due to a few small issues.
This change address those as follows:
1. Fixes subscript out of bounds on Debug due to misuse of reserve vs resize
2. Fixes divide by 0 due to incorrect struct size with variable length array.
3. Fixes infinite loop due to issue with setup of number of packets to read. That
code is still incredibly obtuse for whats its doing and should be refactored at a later
point for maintainability.

How verified:
Sample app launches now. There was much rejoicing.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1252)

### DIFF
--- a/Frameworks/AudioToolbox/CAFDecoder.mm
+++ b/Frameworks/AudioToolbox/CAFDecoder.mm
@@ -213,7 +213,7 @@ void CAFDecoder::ProduceOutputPackets(NSInputStream* stream,
     //  clamp the number of packets to produce based on what is available in the
     //  input buffer
     UInt32 inputPacketSize = mInputFormat.mBytesPerPacket;
-    UInt32 numberOfInputPackets = 1;
+    UInt32 numberOfInputPackets = ([stream hasBytesAvailable] ? 1 : 0);
     if (ioNumberPackets < numberOfInputPackets) {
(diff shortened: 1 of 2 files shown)


Frameworks/AudioToolbox/CAFDecoder.mm
```diff
@@ -331,13 +331,16 @@ uint64_t int64Swap(uint64_t val) {
                 break;
 
             default:
-                bytesToSkip.reserve(chunkHeader.mChunkSize);
-                bytesLeftUntilChunkEnd -= [inStream read:(uint8_t*)&bytesToSkip[0] maxLength:chunkHeader.mChunkSize];
+                if (chunkHeader.mChunkSize > 0) {
+                    bytesToSkip.resize(chunkHeader.mChunkSize);
+                    bytesLeftUntilChunkEnd -= [inStream read:(uint8_t*)&bytesToSkip[0] maxLength:chunkHeader.mChunkSize];
```
aballway: nit: bytesToSkip.data()

DHowett-MSFT:
:shipit:

PR #1254, opened 2016-10-27
Fix scaling and translation text drawing issues

aballway:
CALayers were being scaled twice, and NSString draw methods were being translated twice, causing issues in internal test apps

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1254)

### DIFF
--- a/Frameworks/QuartzCore/CALayer.mm
+++ b/Frameworks/QuartzCore/CALayer.mm
@@ -625,6 +625,7 @@ CGContextRef CreateLayerContentsBitmapContext32(int width, int height) {
         if (priv->contentsScale != 1.0f) {
             // TODO 1077:: Remove once D2D render target is implemented
             _CGContextSetScaleFactor(drawContext, priv->contentsScale);
+            CGContextScaleCTM(drawContext, priv->contentsScale, priv->contentsScale);
         }
 
         CGContextScaleCTM(drawContext, 1.0f, -1.0f);
--- a/Frameworks/QuartzCore/CALayer.mm
+++ b/Frameworks/QuartzCore/CALayer.mm
@@ -623,8 +623,6 @@ CGContextRef CreateLayerContentsBitmapContext32(int width, int height) {
             CGContextTranslateCTM(drawContext, 0, float(target->Backing()->Height()));
         }
         if (priv->contentsScale != 1.0f) {
-            CGContextScaleCTM(drawContext, priv->contentsScale, priv->contentsScale);
-
             // TODO 1077:: Remove once D2D render target is implemented
             _CGContextSetScaleFactor(drawContext, priv->contentsScale);
         }
@@ -2133,11 +2131,11 @@ static void doRecursiveAction(CALayer* layer, NSString* actionName) {
 - (void)dealloc {
     if (DEBUG_VERBOSE) {
         TraceVerbose(TAG,
-            L"CALayer dealloc: (%hs - 0x%p, %hs - 0x%p).",
-            object_getClassName(self),
-            self,
-            priv->delegate ? object_getClassName(priv->delegate) : "nil",
-            priv->delegate);
+                     L"CALayer dealloc: (%hs - 0x%p, %hs - 0x%p).",
+                     object_getClassName(self),
+                     self,
+                     priv->delegate ? object_getClassName(priv->delegate) : "nil",
+                     priv->delegate);
     }
 
     [self removeAllAnimations];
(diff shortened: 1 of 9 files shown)


Frameworks/UIKit/NSString+UIKitAdditions.mm
```diff
@@ -41,6 +41,18 @@ void NSStringForceinclude() {
 
 @implementation NSString (UIKitAdditions)
 
+static CTTextAlignment __UITextAlignmentToCTTextAlignment(UITextAlignment alignment) {
```
ms-jihua: not like this - look at what I did in UIKitTypes.h for NS, UI, CT LineBreakMode. Then you can change it back to a static_cast below.

ms-jihua:
:clock1:

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -623,8 +623,6 @@ - (void)display {
             CGContextTranslateCTM(drawContext, 0, float(target->Backing()->Height()));
         }
         if (priv->contentsScale != 1.0f) {
-            CGContextScaleCTM(drawContext, priv->contentsScale, priv->contentsScale);
-
```
DHowett-MSFT: This scale is for Cairo's sake and needs to be done.
Perhaps we should keep it and apply an inverse scale based on the DPI to the D2D context instead of eliminating it here.

DHowett-MSFT:
:clock1:

msft-Jeyaram:
>     int j = 0;

uh I just noticed this, although this is in. i,j might not of been the best choice to denote what they actually represent.

---
Refers to: Frameworks/CoreText/DWriteWrapper_CoreText.mm:397 in 90d1c31. [](commit_id = 90d1c311e482ce908b5b0862b8beda744ea45cba, deletion_comment = False)

include/UIKit/UIKitTypes.h
```diff
@@ -48,7 +48,7 @@ typedef NS_ENUM(NSUInteger, NSWritingDirection) {
 };
 
 enum {
-    UITextAlignmentLeft,
+    UITextAlignmentLeft = 0,
     UITextAlignmentCenter,
```
msft-Jeyaram: make this an NE_ENUM, such as above?

Frameworks/UIKit/NSString+UIKitAdditions.mm
```diff
@@ -88,12 +100,11 @@ static void drawString(UIFont* font,
...
     std::vector<CGPoint> origins([lines count]);
     CTFrameGetLineOrigins(frame, {}, origins.data());
+    CGFloat lastYPosition = rect.origin.y;
     for (size_t i = 0; i < origins.size(); ++i) {
-        // Need to draw each line at given line origin relative to rect origin
-        // Translated vertically by ascent / 2.0 to counteract flip done by CT*Draw methods
-        CGFloat ascent;
-        CTLineGetTypographicBounds(static_cast<CTLineRef>(lines[i]), &ascent, nullptr, nullptr);
-        CGContextSetTextPosition(context, origins[i].x + rect.origin.x, origins[i].y + rect.origin.y - ascent / 2.0);
+        // Need to set text position so each line will be drawn in the correct position relative to eachother
```
msft-Jeyaram: >eachother [](start = 100, length = 9)

each other

include/UIKit/UIKitTypes.h
aballway: This makes UI/NS unable to be coerced, which I don't believe is intended behavior #WontFix

Frameworks/UIKit/NSString+UIKitAdditions.mm
aballway: unfortunately the reference platform has different values for UI/CT right/center alignment, which our internal test app (and presumably outside developer apps) depend on the values of.

aballway: Double checked - the values between CT/UI are different, and we can't change the values of NS/UI because our internal test app (and presumably others) are depending on those values staying the same #ByDesign

ms-jihua: eesh

---
In reply to: [85569934](https://github.com/Microsoft/WinObjC/pull/1254#discussion_r85569934) [](ancestors = 85569934)

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -623,10 +623,9 @@ - (void)display {
...
             // TODO 1077:: Remove once D2D render target is implemented
             _CGContextSetScaleFactor(drawContext, priv->contentsScale);
+            CGContextScaleCTM(drawContext, priv->contentsScale, priv->contentsScale);
```
rajsesh-msft: simply revert all changes in this file?

rajsesh-msft:
:shipit:

include/UIKit/UIKitTypes.h
DHowett-MSFT: Perhaps NS should just be implemented in terms of UI?

---
In reply to: [85553544](https://github.com/Microsoft/WinObjC/pull/1254#discussion_r85553544) [](ancestors = 85553544)

DHowett-MSFT: or the other way around actually.. yeah

---
In reply to: [85629574](https://github.com/Microsoft/WinObjC/pull/1254#discussion_r85629574) [](ancestors = 85629574,85553544)

DHowett-MSFT: We had this as `NS_ENUM` before; it's somewhat silly but it makes it unable to be used itnerchangeably with NSTextAlignment.

---
In reply to: [85448628](https://github.com/Microsoft/WinObjC/pull/1254#discussion_r85448628) [](ancestors = 85448628)

aballway:
New iteration correctly addresses drawing problems on high dpi devices

Frameworks/UIKit/NSLayoutManager.mm
```diff
@@ -245,12 +245,11 @@ - (void)drawGlyphsForGlyphRange:(NSRange)range atPoint:(CGPoint)position {
...
-
-        float ascent, descent, leading;
-        CTLineGetTypographicBounds(line, &ascent, &descent, &leading);
-
         CGContextSaveGState(curCtx);
-        CGContextSetTextPosition(curCtx, _lineOrigins[curLine].x, _lineOrigins[curLine].y);
+
+        CGFloat ascent, leading;
+        CTLineGetTypographicBounds(line, &ascent, nullptr, &leading);
+        CGContextSetTextPosition(curCtx, _lineOrigins[curLine].x, -(_lineOrigins[curLine].y + ascent + leading));
```
rajsesh-msft: descent missing here intentionally?

PR #1255, opened 2016-10-28
Finish Simple line and status CGPath APIs.

### LINKED ISSUE #1071: Implement CGPath Simple Line APIs With D2D Geometry
Implement The following CGPath APIs

CGPathMoveToPoint
CGPathCreateWithRect
CGPathAddLines
CGPathAddLineToPoint
CGPathAddRect

### LINKED ISSUE #1073: Implement CGPath Status APIs Using D2D Geometry
CGPathGetBoundingBox
CGPathGetCurrentPoint
CGPathIsEmpty
CGPathContainsPoint
CGPathEqualToPath

MSFTFox:
Fill out status APIs such as equals and contains path while adding new tests.

Fixes #1071 
Fixes #1073

---

This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1255)

### DIFF
--- a/Frameworks/CoreGraphics/CGPath.mm
+++ b/Frameworks/CoreGraphics/CGPath.mm
@@ -58,6 +58,26 @@ struct __CGPathImpl {
 };
 
 struct __CGPath : CoreFoundation::CppBase<__CGPath, __CGPathImpl> {
+    inline ComPtr<ID2D1PathGeometry>& PathGeometry() {
+        return _impl.pathGeometry;
+    }
+
+    inline ComPtr<ID2D1GeometrySink>& GeometrySink() {
+        return _impl.geometrySink;
+    }
+
+    inline bool& IsFigureClosed() {
+        return _impl.isFigureClosed;
+    }
+
+    inline CGPoint& CurrentPoint() {
+        return _impl.currentPoint;
+    }
+
+    inline CGPoint& StartingPoint() {
+        return _impl.startingPoint;
+    }
+
     // A private helper function for re-opening a path geometry. CGPath does not
     // have a concept of an open and a closed path but D2D relies on it. A
     // path/sink cannot be read from while the path is open thus it must be
@@ -65,7 +85,7 @@ struct __CGPath : CoreFoundation::CppBase<__CGPath, __CGPathImpl> {
     // we must open the path again. This cannot be done normally, so we must
     // create a new path with the old path information to edit.
     void preparePathForEditing() {
-        if (!_impl.geometrySink) {
+        if (!GeometrySink()) {
             // Re-open this geometry.
             ComPtr<ID2D1Factory> factory = _GetD2DFactoryInstance();
 
@@ -78,43 +98,43 @@ struct __CGPath : CoreFoundation::CppBase<__CGPath, __CGPathImpl> {
             // reason so this will force it to do otherwise.
             FAIL_FAST_IF_FAILED(factory->CreatePathGeometry(&newPath));
             FAIL_FAST_IF_FAILED(newPath->Open(&newSink));
-            FAIL_FAST_IF_FAILED(_impl.pathGeometry->Stream(newSink.Get()));
+            FAIL_FAST_IF_FAILED(PathGeometry()->Stream(newSink.Get()));
 
-            _impl.pathGeometry = newPath;
-            _impl.geometrySink = newSink;
+            PathGeometry() = newPath;
+            GeometrySink() = newSink;
 
             // Without a new figure being created, it's by default closed
-            _impl.isFigureClosed = true;
+            IsFigureClosed() = true;
         }
     }
 
     void closePath() {
-        if (_impl.geometrySink) {
+        if (GeometrySink()) {
             endFigure(D2D1_FIGURE_END_OPEN);
-            _impl.geometrySink->Close();
-            _impl.geometrySink = nullptr;
+            GeometrySink()->Close();
+            GeometrySink() = nullptr;
         }
     }
 
     void beginFigure() {
-        if (_impl.isFigureClosed) {
-            _impl.geometrySink->BeginFigure(_CGPointToD2D_F(_impl.currentPoint), D2D1_FIGURE_BEGIN_FILLED);
-            _impl.isFigureClosed = false;
+        if (IsFigureClosed()) {
+            GeometrySink()->BeginFigure(_CGPointToD2D_F(CurrentPoint()), D2D1_FIGURE_BEGIN_FILLED);
+            IsFigureClosed() = false;
         }
     }
 
     void endFigure(D2D1_FIGURE_END figureStatus) {
-        if (!_impl.isFigureClosed) {
-            _impl.geometrySink->EndFigure(figureStatus);
-            _impl.isFigureClosed = true;
+        if (!IsFigureClosed()) {
+            GeometrySink()->EndFigure(figureStatus);
+            IsFigureClosed() = true;
         }
     }
 
     void initializeGeometries() {
         ComPtr<ID2D1Factory> factory = _GetD2DFactoryInstance();
 
-        FAIL_FAST_IF_FAILED(factory->CreatePathGeometry(&_impl.pathGeometry));
-        FAIL_FAST_IF_FAILED(_impl.pathGeometry->Open(&_impl.geometrySink));
+        FAIL_FAST_IF_FAILED(factory->CreatePathGeometry(&PathGeometry()));
+        FAIL_FAST_IF_FAILED(PathGeometry()->Open(&GeometrySink()));
     }
 };
 
@@ -131,18 +151,17 @@ static Boolean __CGPathEqual(CFTypeRef cf1, CFTypeRef cf2) {
 
     // ID2D1 Geometries have no isEquals method. However, for two geometries to be equal they are both reported to contain the other.
     // Thus we must do two comparisons.
-
     D2D1_GEOMETRY_RELATION relation = D2D1_GEOMETRY_RELATION_UNKNOWN;
-    if (SUCCEEDED(path1->_impl.pathGeometry->CompareWithGeometry(path2->_impl.pathGeometry.Get(), D2D1::IdentityMatrix(), &relation))) {
-        // Does path 1 contain path 2?
-        if (relation == D2D1_GEOMETRY_RELATION_IS_CONTAINED) {
-            if (SUCCEEDED(
-                    path2->_impl.pathGeometry->CompareWithGeometry(path1->_impl.pathGeometry.Get(), D2D1::IdentityMatrix(), &relation))) {
-                // Return true if path 2 also contains path 1
-                return (relation == D2D1_GEOMETRY_RELATION_IS_CONTAINED ? true : false);
-            }
-        }
+    RETURN_FALSE_IF_FAILED(path1->PathGeometry()->CompareWithGeometry(path2->PathGeometry().Get(), D2D1::IdentityMatrix(), &relation));
+
+    // Does path 1 contain path 2?
+    if (relation == D2D1_GEOMETRY_RELATION_IS_CONTAINED) {
+        RETURN_FALSE_IF_FAILED(path2->PathGeometry()->CompareWithGeometry(path1->PathGeometry().Get(), D2D1::IdentityMatrix(), &relation));
+
+        // Return true if path 2 also contains path 1
+        return (relation == D2D1_GEOMETRY_RELATION_IS_CONTAINED ? true : false);
     }
+
     return false;
 }
 
@@ -180,13 +199,13 @@ CGMutablePathRef CGPathCreateMutableCopy(CGPathRef path) {
     // Otherwise the D2D calls will return that a bad state has been entered.
     path->closePath();
 
-    if (FAILED(path->_impl.pathGeometry->Stream(mutableRet->_impl.geometrySink.Get()))) {
+    if (FAILED(path->PathGeometry()->Stream(mutableRet->GeometrySink().Get()))) {
         CGPathRelease(mutableRet);
         return NULL;
     }
 
-    mutableRet->_impl.currentPoint = path->_impl.currentPoint;
-    mutableRet->_impl.startingPoint = path->_impl.startingPoint;
+    mutableRet->CurrentPoint() = path->CurrentPoint();
+    mutableRet->StartingPoint() = path->StartingPoint();
 
     return mutableRet;
 }
@@ -204,9 +223,9 @@ void CGPathAddLineToPoint(CGMutablePathRef path, const CGAffineTransform* transf
     CGPoint pt = __CreateCGPointWithTransform(x, y, transform);
 
     path->beginFigure();
-    path->_impl.geometrySink->AddLine(_CGPointToD2D_F(pt));
+    path->GeometrySink()->AddLine(_CGPointToD2D_F(pt));
 
-    path->_impl.currentPoint = pt;
+    path->CurrentPoint() = pt;
 }
 
 CGFloat _CGPathControlPointOffsetMultiplier(CGFloat angle) {
@@ -385,8 +404,8 @@ void CGPathMoveToPoint(CGMutablePathRef path, const CGAffineTransform* transform
     // BeginFigure or anything else on the D2DPath because that would modify it's state in an invalid way.
     CGPoint pt = __CreateCGPointWithTransform(x, y, transform);
 
-    path->_impl.startingPoint = pt;
-    path->_impl.currentPoint = pt;
+    path->StartingPoint() = pt;
+    path->CurrentPoint() = pt;
 
     path->endFigure(D2D1_FIGURE_END_OPEN);
 }
@@ -432,7 +451,7 @@ void CGPathAddPath(CGMutablePathRef path, const CGAffineTransform* transform, CG
     // Close the path we're adding and open the path we need to add to.
     toAdd->closePath();
     path->preparePathForEditing();
-    FAIL_FAST_IF_FAILED(toAdd->_impl.pathGeometry->Stream(path->_impl.geometrySink.Get()));
+    FAIL_FAST_IF_FAILED(toAdd->PathGeometry()->Stream(path->GeometrySink().Get()));
 }
 
 /**
@@ -474,7 +493,7 @@ void CGPathAddEllipseInRect(CGMutablePathRef path, const CGAffineTransform* tran
 */
 void CGPathCloseSubpath(CGMutablePathRef path) {
     // Move the current point to the starting point since the line is closed.
-    path->_impl.currentPoint = path->_impl.startingPoint;
+    path->CurrentPoint() = path->StartingPoint();
     path->endFigure(D2D1_FIGURE_END_CLOSED);
 }
 
@@ -491,7 +510,9 @@ CGRect CGPathGetBoundingBox(CGPathRef path) {
 
     path->closePath();
 
-    FAIL_FAST_IF_FAILED(path->_impl.pathGeometry->GetBounds(D2D1::IdentityMatrix(), &bounds));
+    if (FAILED(path->PathGeometry()->GetBounds(D2D1::IdentityMatrix(), &bounds))) {
+        return CGRectMake(INFINITY, INFINITY, 0, 0);
+    }
 
     return _D2DRectToCGRect(bounds);
 }
@@ -509,7 +530,7 @@ bool CGPathIsEmpty(CGPathRef path) {
 
     path->closePath();
 
-    path->_impl.pathGeometry->GetFigureCount(&count);
+    path->PathGeometry()->GetFigureCount(&count);
     return count == 0;
 }
 
@@ -654,7 +675,7 @@ bool CGPathContainsPoint(CGPathRef path, const CGAffineTransform* transform, CGP
     BOOL containsPoint = FALSE;
 
     path->closePath();
-    FAIL_FAST_IF_FAILED(path->_impl.pathGeometry->FillContainsPoint(_CGPointToD2D_F(point), D2D1::IdentityMatrix(), &containsPoint));
+    RETURN_FALSE_IF_FAILED(path->PathGeometry()->FillContainsPoint(_CGPointToD2D_F(point), D2D1::IdentityMatrix(), &containsPoint));
 
     return (containsPoint ? true : false);
 }
@@ -717,7 +738,7 @@ bool CGPathEqualToPath(CGPathRef path1, CGPathRef path2) {
  @Status Interoperable
 */
 CGPoint CGPathGetCurrentPoint(CGPathRef path) {
-    return path->_impl.currentPoint;
+    return path->CurrentPoint();
 }
 
 /**
--- a/Frameworks/CoreGraphics/CGPath.mm
+++ b/Frameworks/CoreGraphics/CGPath.mm
@@ -58,11 +58,11 @@ struct __CGPathImpl {
 };
 
 struct __CGPath : CoreFoundation::CppBase<__CGPath, __CGPathImpl> {
-    inline ComPtr<ID2D1PathGeometry>& PathGeometry() {
+    inline ComPtr<ID2D1PathGeometry> GetPathGeometry() {
         return _impl.pathGeometry;
     }
 
-    inline ComPtr<ID2D1GeometrySink>& GeometrySink() {
+    inline ComPtr<ID2D1GeometrySink> GetGeometrySink() {
         return _impl.geometrySink;
     }
 
@@ -81,7 +81,7 @@ struct __CGPath : CoreFoundation::CppBase<__CGPath, __CGPathImpl> {
     // we must open the path again. This cannot be done normally, so we must
     // create a new path with the old path information to edit.
     HRESULT PreparePathForEditing() {
-        if (!GeometrySink()) {
+        if (!_impl.geometrySink) {
             // Re-open this geometry.
             ComPtr<ID2D1Factory> factory = _GetD2DFactoryInstance();
 
@@ -94,10 +94,10 @@ struct __CGPath : CoreFoundation::CppBase<__CGPath, __CGPathImpl> {
             // reason so this will force it to do otherwise.
             RETURN_IF_FAILED(factory->CreatePathGeometry(&newPath));
             RETURN_IF_FAILED(newPath->Open(&newSink));
-            RETURN_IF_FAILED(PathGeometry()->Stream(newSink.Get()));
+            RETURN_IF_FAILED(_impl.pathGeometry->Stream(newSink.Get()));
 
-            PathGeometry() = newPath;
-            GeometrySink() = newSink;
+            _impl.pathGeometry = newPath;
+            _impl.geometrySink = newSink;
 
             // Without a new figure being created, it's by default closed
             _impl.figureClosed = true;
@@ -106,33 +106,32 @@ struct __CGPath : CoreFoundation::CppBase<__CGPath, __CGPathImpl> {
     }
 
     void ClosePath() {
-        if (GeometrySink()) {
+        if (_impl.geometrySink) {
             EndFigure(D2D1_FIGURE_END_OPEN);
-            GeometrySink()->Close();
-            GeometrySink() = nullptr;
+            _impl.geometrySink->Close();
+            _impl.geometrySink = nullptr;
         }
     }
 
     void BeginFigure() {
         if (_impl.figureClosed) {
-            GeometrySink()->BeginFigure(_CGPointToD2D_F(CurrentPoint()), D2D1_FIGURE_BEGIN_FILLED);
+            _impl.geometrySink->BeginFigure(_CGPointToD2D_F(CurrentPoint()), D2D1_FIGURE_BEGIN_FILLED);
             _impl.figureClosed = false;
         }
     }
 
     void EndFigure(D2D1_FIGURE_END figureStatus) {
         if (!_impl.figureClosed) {
-            GeometrySink()->EndFigure(figureStatus);
+            _impl.geometrySink->EndFigure(figureStatus);
             _impl.figureClosed = true;
-            ;
         }
     }
 
     HRESULT InitializeGeometries() {
         ComPtr<ID2D1Factory> factory = _GetD2DFactoryInstance();
 
-        RETURN_IF_FAILED(factory->CreatePathGeometry(&PathGeometry()));
-        RETURN_IF_FAILED(PathGeometry()->Open(&GeometrySink()));
+        RETURN_IF_FAILED(factory->CreatePathGeometry(&_impl.pathGeometry));
+        RETURN_IF_FAILED(_impl.pathGeometry->Open(&_impl.geometrySink));
 
         return S_OK;
     }
@@ -143,6 +142,9 @@ CFTypeID CGPathGetTypeID() {
 }
 
 static Boolean __CGPathEqual(CFTypeRef cf1, CFTypeRef cf2) {
+    RETURN_FALSE_IF(!cf1);
+    RETURN_FALSE_IF(!cf2);
+
     __CGPath* path1 = (__CGPath*)cf1;
     __CGPath* path2 = (__CGPath*)cf2;
 
@@ -152,11 +154,13 @@ static Boolean __CGPathEqual(CFTypeRef cf1, CFTypeRef cf2) {
     // ID2D1 Geometries have no isEquals method. However, for two geometries to be equal they are both reported to contain the other.
     // Thus we must do two comparisons.
     D2D1_GEOMETRY_RELATION relation = D2D1_GEOMETRY_RELATION_UNKNOWN;
-    RETURN_FALSE_IF_FAILED(path1->PathGeometry()->CompareWithGeometry(path2->PathGeometry().Get(), D2D1::IdentityMatrix(), &relation));
+    RETURN_FALSE_IF_FAILED(
+        path1->GetPathGeometry()->CompareWithGeometry(path2->GetPathGeometry().Get(), D2D1::IdentityMatrix(), &relation));
 
     // Does path 1 contain path 2?
     if (relation == D2D1_GEOMETRY_RELATION_IS_CONTAINED) {
-        RETURN_FALSE_IF_FAILED(path2->PathGeometry()->CompareWithGeometry(path1->PathGeometry().Get(), D2D1::IdentityMatrix(), &relation));
+        RETURN_FALSE_IF_FAILED(
+            path2->GetPathGeometry()->CompareWithGeometry(path1->GetPathGeometry().Get(), D2D1::IdentityMatrix(), &relation));
 
         // Return true if path 2 also contains path 1
         return (relation == D2D1_GEOMETRY_RELATION_IS_CONTAINED ? true : false);
@@ -199,10 +203,7 @@ CGMutablePathRef CGPathCreateMutableCopy(CGPathRef path) {
     // Otherwise the D2D calls will return that a bad state has been entered.
     path->ClosePath();
 
-    if (FAILED(path->PathGeometry()->Stream(mutableRet->GeometrySink().Get()))) {
-        CGPathRelease(mutableRet);
-        return NULL;
-    }
+    FAIL_FAST_IF_FAILED(path->GetPathGeometry()->Stream(mutableRet->GetGeometrySink().Get()));
 
     mutableRet->CurrentPoint() = path->CurrentPoint();
     mutableRet->StartingPoint() = path->StartingPoint();
@@ -221,7 +222,7 @@ void CGPathAddLineToPoint(CGMutablePathRef path, const CGAffineTransform* transf
     CGPoint pt = __CreateCGPointWithTransform(x, y, transform);
 
     path->BeginFigure();
-    path->GeometrySink()->AddLine(_CGPointToD2D_F(pt));
+    path->GetGeometrySink()->AddLine(_CGPointToD2D_F(pt));
 
     path->CurrentPoint() = pt;
 }
@@ -395,6 +396,8 @@ void CGPathAddArc(CGMutablePathRef path,
  @Status Interoperable
 */
 void CGPathMoveToPoint(CGMutablePathRef path, const CGAffineTransform* transform, CGFloat x, CGFloat y) {
+    RETURN_IF(!path);
+
     // CGPaths do not consider these actions to be segments of the path and are not considered on CGPathApply, thus we should simply end the
     // current figure and move the location of this path to the new point.
     path->EndFigure(D2D1_FIGURE_END_OPEN);
@@ -439,7 +442,7 @@ void CGPathAddPath(CGMutablePathRef path, const CGAffineTransform* transform, CG
     // Close the path we're adding and open the path we need to add to.
     toAdd->ClosePath();
     FAIL_FAST_IF_FAILED(path->PreparePathForEditing());
-    FAIL_FAST_IF_FAILED(toAdd->PathGeometry()->Stream(path->GeometrySink().Get()));
+    FAIL_FAST_IF_FAILED(toAdd->GetPathGeometry()->Stream(path->GetGeometrySink().Get()));
 }
 
 /**
@@ -478,6 +481,8 @@ void CGPathAddEllipseInRect(CGMutablePathRef path, const CGAffineTransform* tran
  @Status Interoperable
 */
 void CGPathCloseSubpath(CGMutablePathRef path) {
+    RETURN_IF(!path);
+
     // Move the current point to the starting point since the line is closed.
     path->CurrentPoint() = path->StartingPoint();
     path->EndFigure(D2D1_FIGURE_END_CLOSED);
@@ -496,7 +501,7 @@ CGRect CGPathGetBoundingBox(CGPathRef path) {
 
     path->ClosePath();
 
-    if (FAILED(path->PathGeometry()->GetBounds(D2D1::IdentityMatrix(), &bounds))) {
+    if (FAILED(path->GetPathGeometry()->GetBounds(D2D1::IdentityMatrix(), &bounds))) {
         return CGRectNull;
     }
 
@@ -516,7 +521,7 @@ bool CGPathIsEmpty(CGPathRef path) {
 
     path->ClosePath();
 
-    path->PathGeometry()->GetFigureCount(&count);
+    RETURN_FALSE_IF_FAILED(path->GetPathGeometry()->GetFigureCount(&count));
     return count == 0;
 }
 
@@ -658,7 +663,7 @@ bool CGPathContainsPoint(CGPathRef path, const CGAffineTransform* transform, CGP
     BOOL containsPoint = FALSE;
 
     path->ClosePath();
-    RETURN_FALSE_IF_FAILED(path->PathGeometry()->FillContainsPoint(_CGPointToD2D_F(point), D2D1::IdentityMatrix(), &containsPoint));
+    RETURN_FALSE_IF_FAILED(path->GetPathGeometry()->FillContainsPoint(_CGPointToD2D_F(point), D2D1::IdentityMatrix(), &containsPoint));
 
     return (containsPoint ? true : false);
 }
@@ -721,6 +726,9 @@ bool CGPathEqualToPath(CGPathRef path1, CGPathRef path2) {
  @Status Interoperable
 */
 CGPoint CGPathGetCurrentPoint(CGPathRef path) {
+    if (!path) {
+        return CGPointZero;
+    }
     return path->CurrentPoint();
 }
 
--- a/Frameworks/CoreGraphics/CGPath.mm
+++ b/Frameworks/CoreGraphics/CGPath.mm
@@ -58,20 +58,28 @@ struct __CGPathImpl {
 };
 
 struct __CGPath : CoreFoundation::CppBase<__CGPath, __CGPathImpl> {
-    inline ComPtr<ID2D1PathGeometry> GetPathGeometry() {
+    ComPtr<ID2D1PathGeometry> GetPathGeometry() {
         return _impl.pathGeometry;
     }
 
-    inline ComPtr<ID2D1GeometrySink> GetGeometrySink() {
+    ComPtr<ID2D1GeometrySink> GetGeometrySink() {
         return _impl.geometrySink;
     }
 
-    inline CGPoint& CurrentPoint() {
+    CGPoint GetCurrentPoint() {
         return _impl.currentPoint;
     }
 
-    inline CGPoint& StartingPoint() {
+    CGPoint GetStartingPoint() {
         return _impl.startingPoint;
+    }
+
+    void SetCurrentPoint(CGPoint newPoint) {
+        _impl.currentPoint = newPoint;
+    }
+
+    void SetStartingPoint(CGPoint newPoint) {
+        _impl.startingPoint = newPoint;
     }
 
     // A private helper function for re-opening a path geometry. CGPath does not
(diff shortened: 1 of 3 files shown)


Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -626,29 +644,20 @@ void CGPathApply(CGPathRef path, void* info, CGPathApplierFunction function) {
 }
 
 /**
- @Status Interoperable
- @Notes
+ @Status Caveat
+ @Notes eoFill ignored. Default fill pattern for ID2D1 Geometry is used.
 */
 bool CGPathContainsPoint(CGPathRef path, const CGAffineTransform* transform, CGPoint point, bool eoFill) {
```
msft-Jeyaram: >path [](start = 35, length = 4)

validity check

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -127,6 +125,24 @@ CFTypeID CGPathGetTypeID() {
...
     __CGPath* path1 = (__CGPath*)cf1;
     __CGPath* path2 = (__CGPath*)cf2;
+
+    path1->closePath();
+    path2->closePath();
+
+    // ID2D1 Geometries have no isEquals method. However, for two geometries to be equal they are both reported to contain the other.
+    // Thus we must do two comparisons.
+
+    D2D1_GEOMETRY_RELATION relation = D2D1_GEOMETRY_RELATION_UNKNOWN;
```
aballway: Will this be needed anywhere else / worth making into a helper?

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -626,29 +644,20 @@ void CGPathApply(CGPathRef path, void* info, CGPathApplierFunction function) {
...
     }
-    // check if the point is outside this box already, if it is, return false
-    CGRect boundingBox = CGPathGetBoundingBox(path);
-    if (!CGRectContainsPoint(boundingBox, point)) {
-        return false;
-    }
 
-    CGContextRef context =
-        CGBitmapContextCreate(0, boundingBox.origin.x + boundingBox.size.width, boundingBox.origin.y + boundingBox.size.height, 1, 1, 0, 0);
+    BOOL containsPoint = false;
```
aballway: nit: bool

tests/unittests/CoreGraphics/CGPathTests.mm
```diff
@@ -696,11 +696,175 @@ void EXPECT_SIZEEQ(CGSize pathSize, CGFloat width, CGFloat height) {
...
+
+TEST(CGPath, CGPathRectanglesTest) {
+    CGRect theRectangle = CGRectMake(50, 50, 100, 100);
+    CGMutablePathRef path = CGPathCreateWithRect(theRectangle, NULL);
+    CGRect boundingBox = CGPathGetBoundingBox(path);
+
+    EXPECT_POINTEQ(boundingBox.origin, theRectangle.origin.x, theRectangle.origin.y);
+    EXPECT_SIZEEQ(boundingBox.size, theRectangle.size.height, theRectangle.size.width);
+
+    path = CGPathCreateMutable();
```
aballway: release first path

tests/unittests/CoreGraphics/CGPathTests.mm
```diff
@@ -696,11 +696,175 @@ void EXPECT_SIZEEQ(CGSize pathSize, CGFloat width, CGFloat height) {
...
+    CGPathMoveToPoint(path2, NULL, 50, 50);
+    CGPathAddLineToPoint(path2, NULL, 75, 75);
+
+    EXPECT_TRUE(testBidirectionalEquals(path1, path2));
+
+    CGPathAddLineToPoint(path2, NULL, 100, 100);
+
+    EXPECT_TRUE(!(testBidirectionalEquals(path1, path2)));
+
+    path1 = CGPathCreateMutable();
```
aballway: same as above

tests/unittests/CoreGraphics/CGPathTests.mm
```diff
@@ -696,11 +696,175 @@ void EXPECT_SIZEEQ(CGSize pathSize, CGFloat width, CGFloat height) {
...
+    CGRect theRectangle = CGRectMake(50, 50, 100, 100);
+    CGMutablePathRef rectPath1 = CGPathCreateWithRect(theRectangle, NULL);
+    CGMutablePathRef rectPath2 = CGPathCreateWithRect(theRectangle, NULL);
+
+    EXPECT_TRUE(testBidirectionalEquals(rectPath1, rectPath2));
+
+    // Add an extra segment coming off of the rectangle.
+    CGPathAddLineToPoint(rectPath2, NULL, 200, 200);
+
+    EXPECT_TRUE(!(testBidirectionalEquals(rectPath1, rectPath2)));
```
aballway: nit: EXPECT_FALSE, that's a super sneaky not

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -127,6 +125,24 @@ CFTypeID CGPathGetTypeID() {
...
+    if (SUCCEEDED(path1->_impl.pathGeometry->CompareWithGeometry(path2->_impl.pathGeometry.Get(), D2D1::IdentityMatrix(), &relation))) {
+        // Does path 1 contain path 2?
+        if (relation == D2D1_GEOMETRY_RELATION_IS_CONTAINED) {
+            if (SUCCEEDED(
+                    path2->_impl.pathGeometry->CompareWithGeometry(path1->_impl.pathGeometry.Get(), D2D1::IdentityMatrix(), &relation))) {
+                // Return true if path 2 also contains path 1
+                return (relation == D2D1_GEOMETRY_RELATION_IS_CONTAINED ? true : false);
+            }
+        }
+    }
```
msft-Jeyaram: This nesting can be uncoupled via the usage of RETURN_FALSE_IF_FAILED.
This would help introduce early fail conditions and make the code much leaner.

msft-Jeyaram:
>     }

switch the validity checks to use the macros in result.h.
RETURN_NULL_IF(!path);

---
Refers to: Frameworks/CoreGraphics/CGPath.mm:167 in ac21011. [](commit_id = ac21011e1d5cf7909c64ecb5cc0cb05ee2f313aa, deletion_comment = False)

msft-Jeyaram:
>     FAIL_FAST_IF_FAILED(path->_impl.pathGeometry->Stream(mutableRet->_impl.geometrySink.Get()));

I feel as this should be returning NULL, rather than failing fast. 
Given how it's a public function that attempts to obtain something, it makes it valid to return NULL. Also the user may/may not have the code to handle the null case.


---
Refers to: Frameworks/CoreGraphics/CGPath.mm:187 in ac21011. [](commit_id = ac21011e1d5cf7909c64ecb5cc0cb05ee2f313aa, deletion_comment = False)

msft-Jeyaram:
>     }

RETURN_IF(!path)

---
Refers to: Frameworks/CoreGraphics/CGPath.mm:201 in ac21011. [](commit_id = ac21011e1d5cf7909c64ecb5cc0cb05ee2f313aa, deletion_comment = False)

msft-Jeyaram:
>  @Notes transform property not supported

is this still true?

---
Refers to: Frameworks/CoreGraphics/CGPath.mm:224 in ac21011. [](commit_id = ac21011e1d5cf7909c64ecb5cc0cb05ee2f313aa, deletion_comment = False)

tests/unittests/CoreGraphics/CGPathTests.mm
```diff
@@ -696,11 +696,175 @@ void EXPECT_SIZEEQ(CGSize pathSize, CGFloat width, CGFloat height) {
...
+    CGPathRelease(path);
+}
+
+TEST(CGPath, CGPathContainsPointTest) {
+    // bool CGPathContainsPoint(CGPathRef path, const CGAffineTransform* transform, CGPoint point, bool eoFill) {
+    CGRect theRectangle = CGRectMake(50, 50, 100, 100);
+    CGMutablePathRef path = CGPathCreateWithRect(theRectangle, NULL);
+
+    EXPECT_TRUE(CGPathContainsPoint(path, NULL, CGPointMake(75, 75), NO));
+    EXPECT_TRUE(!(CGPathContainsPoint(path, NULL, CGPointMake(200, 200), NO)));
```
msft-Jeyaram: >EXPECT_TRUE(! [](start = 4, length = 13)

EXPECT_FALSE

tests/unittests/CoreGraphics/CGPathTests.mm
```diff
@@ -696,11 +696,175 @@ void EXPECT_SIZEEQ(CGSize pathSize, CGFloat width, CGFloat height) {
...
+
+    CGRect smallRectangle = CGRectMake(75, 75, 5, 5);
+    CGMutablePathRef containedRectangle = CGPathCreateWithRect(smallRectangle, NULL);
+
+    EXPECT_TRUE(!(testBidirectionalEquals(rectPath1, containedRectangle)));
+
+    // Two complicated paths that should be equal.
+    theRectangle = CGRectMake(50, 50, 100, 100);
+    rectPath1 = CGPathCreateWithRect(theRectangle, NULL);
+    rectPath2 = CGPathCreateWithRect(theRectangle, NULL);
```
msft-Jeyaram: some of these are not released before being reassigned.

msft-Jeyaram:
:clock1:

Frameworks/CoreGraphics/CGPath.mm
aballway: I'd still prefer bool over SHOUTY/ObjC BOOL types (actually which one is it here?)

MSFTFox:
>     FAIL_FAST_IF_FAILED(path->_impl.pathGeometry->Stream(mutableRet->_impl.geometrySink.Get()));

Actually we should be releasing any resources we made before leaking them when we return NULL. 

---
In reply to: [256970578](https://github.com/Microsoft/WinObjC/pull/1255#issuecomment-256970578) [](ancestors = 256970578)

---
Refers to: Frameworks/CoreGraphics/CGPath.mm:187 in ac21011. [](commit_id = ac21011e1d5cf7909c64ecb5cc0cb05ee2f313aa, deletion_comment = False)

MSFTFox:
>     }

That doesn't exist, even if it did I don't think reducing a null check+return to a macro would be very useful.

---
In reply to: [256970753](https://github.com/Microsoft/WinObjC/pull/1255#issuecomment-256970753) [](ancestors = 256970753)

---
Refers to: Frameworks/CoreGraphics/CGPath.mm:201 in ac21011. [](commit_id = ac21011e1d5cf7909c64ecb5cc0cb05ee2f313aa, deletion_comment = False)

MSFTFox:
>  @Notes transform property not supported

Well, nothing about this function is particularly true... We haven't gotten to arcs yet. ;)

---
In reply to: [256971251](https://github.com/Microsoft/WinObjC/pull/1255#issuecomment-256971251) [](ancestors = 256971251)

---
Refers to: Frameworks/CoreGraphics/CGPath.mm:224 in ac21011. [](commit_id = ac21011e1d5cf7909c64ecb5cc0cb05ee2f313aa, deletion_comment = False)

Frameworks/CoreGraphics/CGPath.mm
MSFTFox: #define BOOL WINDOWS_BOOL

This is correct.

---
In reply to: [85578185](https://github.com/Microsoft/WinObjC/pull/1255#discussion_r85578185) [](ancestors = 85578185)

MSFTFox: Although it should be initialized to FALSE to match the type.

---
In reply to: [85579045](https://github.com/Microsoft/WinObjC/pull/1255#discussion_r85579045) [](ancestors = 85579045,85578185)

msft-Jeyaram:
>  @Notes transform property not supported

gotcha.

---
In reply to: [256985573](https://github.com/Microsoft/WinObjC/pull/1255#issuecomment-256985573) [](ancestors = 256985573,256971251)

---
Refers to: Frameworks/CoreGraphics/CGPath.mm:224 in ac21011. [](commit_id = ac21011e1d5cf7909c64ecb5cc0cb05ee2f313aa, deletion_comment = False)

msft-Jeyaram:
>     }

It exists now after my merge into CGD2D. Agree, but given we are having this defacto standard of using MACROS to deal with these .
You can leave it as is for now.

---
In reply to: [256985291](https://github.com/Microsoft/WinObjC/pull/1255#issuecomment-256985291) [](ancestors = 256985291,256970753)

---
Refers to: Frameworks/CoreGraphics/CGPath.mm:201 in ac21011. [](commit_id = ac21011e1d5cf7909c64ecb5cc0cb05ee2f313aa, deletion_comment = False)

tests/unittests/CoreGraphics/CGPathTests.mm
```diff
@@ -696,11 +696,182 @@ void EXPECT_SIZEEQ(CGSize pathSize, CGFloat width, CGFloat height) {
...
+
+    EXPECT_POINTEQ(boundingBox.origin, theRectangle.origin.x, theRectangle.origin.y);
+    EXPECT_SIZEEQ(boundingBox.size, theRectangle.size.height, theRectangle.size.width);
+
+    CGPathRelease(path);
+    path = CGPathCreateMutable();
+
+    CGPathMoveToPoint(path, NULL, 50, 50);
+    CGPathAddLineToPoint(path, NULL, 100, 100);
+    CGPathAddRect(path, NULL, CGRectMake(100, 100, 200, 100));
```
rajsesh-msft: What does the rest of the code test?

MSFTFox:
>     }

Well I agree that the standard we're using is for the macros to handle them but the macros do a bit more than just returning. I'll change it since that's the standard we're following but I don't think in a case like this it's terribly important.

---
In reply to: [256986759](https://github.com/Microsoft/WinObjC/pull/1255#issuecomment-256986759) [](ancestors = 256986759,256985291,256970753)

---
Refers to: Frameworks/CoreGraphics/CGPath.mm:201 in ac21011. [](commit_id = ac21011e1d5cf7909c64ecb5cc0cb05ee2f313aa, deletion_comment = False)

Frameworks/CoreGraphics/CGPath.mm
bbowman: The nested logic in question is around the if(SUCCEEDED clauses. It is hard to read where the actual conditionals are admist those. A rewritten version is cleaner and looks like:

D2D1_GEOMETRY_RELATION relation = D2D1_GEOMETRY_RELATION_UNKNOWN;
RETURN_FALSE_IF_FAILED(path1->_impl.pathGeometry->CompareWithGeometry(path2->_impl.pathGeometry.Get(), D2D1::IdentityMatrix(), &relation)));

// Does path 1 contain path 2?
if (relation == D2D1_GEOMETRY_RELATION_IS_CONTAINED) {
    RETURN_FALSE_IF_FAILED(path2->_impl.pathGeometry->CompareWithGeometry(path1->_impl.pathGeometry.Get(), D2D1::IdentityMatrix(), &relation)));

    // Return true if path 2 also contains path 1
    return (relation == D2D1_GEOMETRY_RELATION_IS_CONTAINED ? true : false);
}


which makes the actual conditions around relation pop out.

---
In reply to: [85576878](https://github.com/Microsoft/WinObjC/pull/1255#discussion_r85576878) [](ancestors = 85576878,85566823)

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -127,6 +125,24 @@ CFTypeID CGPathGetTypeID() {
...
     __CGPath* path2 = (__CGPath*)cf2;
+
+    path1->closePath();
+    path2->closePath();
+
+    // ID2D1 Geometries have no isEquals method. However, for two geometries to be equal they are both reported to contain the other.
+    // Thus we must do two comparisons.
+
+    D2D1_GEOMETRY_RELATION relation = D2D1_GEOMETRY_RELATION_UNKNOWN;
+    if (SUCCEEDED(path1->_impl.pathGeometry->CompareWithGeometry(path2->_impl.pathGeometry.Get(), D2D1::IdentityMatrix(), &relation))) {
```
bbowman: >path1->_impl.pathGeometry [](start = 18, length = 25)

encapsulation?

bbowman:
:clock1:

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -626,29 +641,22 @@ void CGPathApply(CGPathRef path, void* info, CGPathApplierFunction function) {
...
-        CGBitmapContextCreate(0, boundingBox.origin.x + boundingBox.size.width, boundingBox.origin.y + boundingBox.size.height, 1, 1, 0, 0);
+    BOOL containsPoint = FALSE;
 
-    CGContextAddPath(context, path);
-
-    bool inPath = CGContextIsPointInPath(context, eoFill, point.x, point.y);
-
-    CGContextRelease(context);
+    path->closePath();
+    FAIL_FAST_IF_FAILED(path->_impl.pathGeometry->FillContainsPoint(_CGPointToD2D_F(point), D2D1::IdentityMatrix(), &containsPoint));
```
MSFTFox: The intention I used with fail_fast was that if the geometry APIs ever resulted in an unrecoverable state for the path that this should fail. This one in particular should have been to return false instead of failing.

---
In reply to: [85594909](https://github.com/Microsoft/WinObjC/pull/1255#discussion_r85594909) [](ancestors = 85594909)

bbowman: Talked to @DHowett-MSFT on how to standardize this. *All* internal method that have some failure state should be unified to do the same thing (likely HRESULT return as they call windows-y things). At the outermost layer, the determination is made for that function what style it should be. CG appears to be a generally "no fail" set of apis so likely this means FAIL_FAST for us until we determine to dial it back. If we dial it back we should LOG and early return in any failure cases to preserve the code flow and keep it clean. Sound good?

---
In reply to: [85607025](https://github.com/Microsoft/WinObjC/pull/1255#discussion_r85607025) [](ancestors = 85607025,85594909)

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -60,14 +58,34 @@ static inline CGPoint __CreateCGPointWithTransform(CGFloat x, CGFloat y, const C
...
+        return _impl.startingPoint;
+    }
+
     // A private helper function for re-opening a path geometry. CGPath does not
     // have a concept of an open and a closed path but D2D relies on it. A
     // path/sink cannot be read from while the path is open thus it must be
     // closed. However, a CGPath can be edited again after being read from so
     // we must open the path again. This cannot be done normally, so we must
     // create a new path with the old path information to edit.
     void preparePathForEditing() {
```
bbowman: >preparePathForEditing [](start = 9, length = 21)

For instance this guy would be failure communicating via an HRESULT return and the user of __CGPath (likely a public API function) will do a single FAIL_FAST

DHowett-MSFT: >preparePathForEditing [](start = 9, length = 21)

curious why we have Uppercase and notUppercase functions in the same class

DHowett-MSFT:
>     // CGPathMoveToPoint does not update the geometry in any way. CGPaths do not consider these actions to be

This comment is now a lie, as it closes the geometry's existing figure.

---
Refers to: Frameworks/CoreGraphics/CGPath.mm:402 in 5267579. [](commit_id = 5267579703892ebf3b7ccbd2922aafeaa7782e1d, deletion_comment = False)

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -474,7 +510,9 @@ CGRect CGPathGetBoundingBox(CGPathRef path) {
 
     path->closePath();
 
-    FAIL_FAST_IF_FAILED(path->_impl.pathGeometry->GetBounds(D2D1::IdentityMatrix(), &bounds));
+    if (FAILED(path->PathGeometry()->GetBounds(D2D1::IdentityMatrix(), &bounds))) {
+        return CGRectMake(INFINITY, INFINITY, 0, 0);
+    }
```
DHowett-MSFT: Each INFINITY,INFINITY,0,0 rect here should just be `CGRectNull`

tests/unittests/CoreGraphics/CGPathTests.mm
```diff
@@ -696,11 +696,187 @@ void EXPECT_SIZEEQ(CGSize pathSize, CGFloat width, CGFloat height) {
     EXPECT_POINTEQ(boundingBox.origin, 75, 75);
     EXPECT_SIZEEQ(boundingBox.size, 25, 25);
 
-    CGPathAddPath(path, nullptr, secondPath);
+    CGPathAddPath(path, NULL, secondPath);
     boundingBox = CGPathGetBoundingBox(path);
```
DHowett-MSFT: Why? nullptr is superior in every way.

DHowett-MSFT:
Unrelated to this PR:
It would be nice to see a test similar to

```
Line 0, 0 -> 100, 100
Rect 100,100 +100,+100
```

Except where the rectangle does _not_ touch the line.

I'd like to know whether the reference platform adds a connector, and whether we do with D2D.

tests/unittests/CoreGraphics/CGPathTests.mm
```diff
@@ -696,11 +696,187 @@ void EXPECT_SIZEEQ(CGSize pathSize, CGFloat width, CGFloat height) {
...
+}
+
+TEST(CGPath, CGPathAddLinesTest) {
+    CGMutablePathRef path = CGPathCreateMutable();
+
+    CGPoint points[] = { { 50, 50 }, { 100, 50 }, { 50, 50 }, { 50, 100 } };
+
+    CGPathMoveToPoint(path, NULL, 50, 50);
+
+    CGPathAddLines(path, NULL, points, 4);
```
DHowett-MSFT: >NULL [](start = 25, length = 4)

Every `NULL` to `nullptr`

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -60,14 +58,34 @@ static inline CGPoint __CreateCGPointWithTransform(CGFloat x, CGFloat y, const C
...
+    inline ComPtr<ID2D1PathGeometry>& PathGeometry() {
+        return _impl.pathGeometry;
+    }
+
+    inline ComPtr<ID2D1GeometrySink>& GeometrySink() {
+        return _impl.geometrySink;
+    }
+
+    inline bool& IsFigureClosed() {
+        return _impl.isFigureClosed;
```
aballway: I really dislike returning a reference here.  Traditional setters/getters would be more appropriate

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -127,6 +145,23 @@ CFTypeID CGPathGetTypeID() {
 static Boolean __CGPathEqual(CFTypeRef cf1, CFTypeRef cf2) {
     __CGPath* path1 = (__CGPath*)cf1;
     __CGPath* path2 = (__CGPath*)cf2;
+
+    path1->closePath();
```
aballway: should we be closing the path here, and will doing so change how future operations on the path behave?

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -626,29 +662,22 @@ void CGPathApply(CGPathRef path, void* info, CGPathApplierFunction function) {
...
+    BOOL containsPoint = FALSE;
 
-    bool inPath = CGContextIsPointInPath(context, eoFill, point.x, point.y);
-
-    CGContextRelease(context);
+    path->closePath();
+    RETURN_FALSE_IF_FAILED(path->PathGeometry()->FillContainsPoint(_CGPointToD2D_F(point), D2D1::IdentityMatrix(), &containsPoint));
 
-    return inPath;
+    return (containsPoint ? true : false);
```
aballway: nit: do we need to be explicit about returning a bool here?

tests/unittests/CoreGraphics/CGPathTests.mm
```diff
@@ -696,11 +696,187 @@ void EXPECT_SIZEEQ(CGSize pathSize, CGFloat width, CGFloat height) {
...
+    CGRect theRectangle = CGRectMake(50, 50, 100, 100);
+    CGMutablePathRef path = CGPathCreateWithRect(theRectangle, NULL);
+
+    EXPECT_TRUE(CGPathContainsPoint(path, NULL, CGPointMake(75, 75), NO));
+    EXPECT_FALSE(CGPathContainsPoint(path, NULL, CGPointMake(200, 200), NO));
+
+    CGPathRelease(path);
+}
+
+bool testBidirectionalEquals(CGPathRef path1, CGPathRef path2) {
```
aballway: super nit: best described as "symmetric equivalence"

tests/unittests/CoreGraphics/CGPathTests.mm
MSFTFox: Consistency with the rest of the file. I'll change them to nullptr instead then.

---
In reply to: [85629785](https://github.com/Microsoft/WinObjC/pull/1255#discussion_r85629785) [](ancestors = 85629785)

Frameworks/CoreGraphics/CGPath.mm
MSFTFox: We must close the path here. Any D2D1 operations require closed figures otherwise an error state is reached. This closes the path strictly for D2D1 operations and any future operations will re-open the path as necessary.

---
In reply to: [85667239](https://github.com/Microsoft/WinObjC/pull/1255#discussion_r85667239) [](ancestors = 85667239)

Frameworks/CoreGraphics/CGPath.mm
DHowett-MSFT: I don't recall doing that, but I think naming consistency within a class trumps other concerns. :)

---
In reply to: [85790869](https://github.com/Microsoft/WinObjC/pull/1255#discussion_r85790869) [](ancestors = 85790869,85629662)

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -78,13 +74,21 @@ inline CGPoint __CreateCGPointWithTransform(CGFloat x, CGFloat y, const CGAffine
         return _impl.startingPoint;
     }
 
+    void SetIsFigureClosed(bool closed) {
```
aballway: nit: would make more sense to have a CloseFigure which closes it and sets isFigureClosed

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -78,13 +74,21 @@ inline CGPoint __CreateCGPointWithTransform(CGFloat x, CGFloat y, const CGAffine
...
+        _impl.isFigureClosed = closed;
+    }
+
+    bool GetIsFigureClosed() {
```
aballway: nit: IsFigureClosed would suffice

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -523,12 +516,12 @@ CGRect CGPathGetBoundingBox(CGPathRef path) {
 */
 bool CGPathIsEmpty(CGPathRef path) {
     if (path == NULL) {
-        return kCFBooleanTrue;
+        return true;
     }
 
     UINT32 count;
```
aballway: nit: uint32_t and move after ClosePath

Frameworks/CoreGraphics/CGPath.mm
MSFTFox: Would rather not overload the variable name with a capitalized version of a getter. The naming convention for functions says to use isFigureClosed so I might as well change the entire variable name at this point. However, we're just dancing around semantics at this point.

---
In reply to: [85826920](https://github.com/Microsoft/WinObjC/pull/1255#discussion_r85826920) [](ancestors = 85826920)

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -60,14 +58,38 @@ static inline CGPoint __CreateCGPointWithTransform(CGFloat x, CGFloat y, const C
...
+    inline CGPoint& CurrentPoint() {
+        return _impl.currentPoint;
+    }
+
+    inline CGPoint& StartingPoint() {
+        return _impl.startingPoint;
+    }
+
+    void SetIsFigureClosed(bool closed) {
+        _impl.isFigureClosed = closed;
```
DHowett-MSFT: This is a helper that's only used internally. Why?

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -60,14 +58,38 @@ static inline CGPoint __CreateCGPointWithTransform(CGFloat x, CGFloat y, const C
...
+    }
+
+    bool GetIsFigureClosed() {
+        return _impl.isFigureClosed;
```
DHowett-MSFT: Only used internally?

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -78,45 +100,48 @@ void preparePathForEditing() {
...
     }
 
-    void endFigure(D2D1_FIGURE_END figureStatus) {
-        if (!_impl.isFigureClosed) {
-            _impl.geometrySink->EndFigure(figureStatus);
-            _impl.isFigureClosed = true;
+    void EndFigure(D2D1_FIGURE_END figureStatus) {
+        if (!GetIsFigureClosed()) {
+            GeometrySink()->EndFigure(figureStatus);
+            SetIsFigureClosed(true);
```
DHowett-MSFT: Can this fail?

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -157,21 +197,22 @@ CGPathRef CGPathCreateCopy(CGPathRef path) {
...
 
     // In order to call stream and copy the contents of the original path into the
     // new copy we must close this path.
     // Otherwise the D2D calls will return that a bad state has been entered.
-    path->closePath();
+    path->ClosePath();
 
-    FAIL_FAST_IF_FAILED(path->_impl.pathGeometry->Stream(mutableRet->_impl.geometrySink.Get()));
+    if (FAILED(path->PathGeometry()->Stream(mutableRet->GeometrySink().Get()))) {
+        CGPathRelease(mutableRet);
```
DHowett-MSFT: If you use `woc::unique_cf` here, you don't need to worry about lifetime.

DHowett-MSFT:
>     UNIMPLEMENTED();

A clever implementation would be to make a rect geometry of the path's bounding box, and then do the Contains-Contained relationship!

---
Refers to: Frameworks/CoreGraphics/CGPath.mm:739 in a12387b. [](commit_id = a12387bd4985a53e27223dafba9c515ad06b9b5d, deletion_comment = False)

DHowett-MSFT:
>     UNIMPLEMENTED();

Not for this CR. Sorry. :)

---
In reply to: [257421985](https://github.com/Microsoft/WinObjC/pull/1255#issuecomment-257421985) [](ancestors = 257421985)

---
Refers to: Frameworks/CoreGraphics/CGPath.mm:739 in a12387b. [](commit_id = a12387bd4985a53e27223dafba9c515ad06b9b5d, deletion_comment = False)

DHowett-MSFT:
>     UNIMPLEMENTED();

hmm, perhaps that would not be as good as I'd thought

---
In reply to: [257421985](https://github.com/Microsoft/WinObjC/pull/1255#issuecomment-257421985) [](ancestors = 257421985)

---
Refers to: Frameworks/CoreGraphics/CGPath.mm:739 in a12387b. [](commit_id = a12387bd4985a53e27223dafba9c515ad06b9b5d, deletion_comment = False)

MSFTFox:
>     UNIMPLEMENTED();

I think that would work. return CGPathEqualsPath(path, CGPathCreateWithRect(CGPathGetBoundingBox(path))).

If it's logically equal to the path created by its bounding box.

---
In reply to: [257422149](https://github.com/Microsoft/WinObjC/pull/1255#issuecomment-257422149) [](ancestors = 257422149,257421985)

---
Refers to: Frameworks/CoreGraphics/CGPath.mm:739 in a12387b. [](commit_id = a12387bd4985a53e27223dafba9c515ad06b9b5d, deletion_comment = False)

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -78,45 +92,49 @@ void preparePathForEditing() {
...
-            FAIL_FAST_IF_FAILED(newPath->Open(&newSink));
-            FAIL_FAST_IF_FAILED(_impl.pathGeometry->Stream(newSink.Get()));
+            RETURN_IF_FAILED(factory->CreatePathGeometry(&newPath));
+            RETURN_IF_FAILED(newPath->Open(&newSink));
+            RETURN_IF_FAILED(PathGeometry()->Stream(newSink.Get()));
 
-            _impl.pathGeometry = newPath;
-            _impl.geometrySink = newSink;
+            PathGeometry() = newPath;
+            GeometrySink() = newSink;
```
msft-Jeyaram: This generally seems unconventional. Although it makes sense in code terms. It might be better to have a setter?
Also Why don't you use move this code into the actual implementation and make this a forwarding call, that way you provide better encapsulation and have the majority of the logic attached with the implementation.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -78,45 +92,49 @@ void preparePathForEditing() {
...
 
-    void beginFigure() {
-        if (_impl.isFigureClosed) {
-            _impl.geometrySink->BeginFigure(_CGPointToD2D_F(_impl.currentPoint), D2D1_FIGURE_BEGIN_FILLED);
-            _impl.isFigureClosed = false;
+    void BeginFigure() {
+        if (_impl.figureClosed) {
+            GeometrySink()->BeginFigure(_CGPointToD2D_F(CurrentPoint()), D2D1_FIGURE_BEGIN_FILLED);
+            _impl.figureClosed = false;
         }
```
msft-Jeyaram: If you move this into the Impl, you can avoid the getters and just forward the method call here.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -136,7 +171,7 @@ static Boolean __CGPathEqual(CFTypeRef cf1, CFTypeRef cf2) {
 CGMutablePathRef CGPathCreateMutable() {
     __CGPath* mutableRet = __CGPath::CreateInstance();
 
-    mutableRet->initializeGeometries();
+    FAIL_FAST_IF_FAILED(mutableRet->InitializeGeometries());
```
msft-Jeyaram: >   FAIL_FAST_IF_FAILED(mutableRet->InitializeGeometries()); [](start = 1, length = 59)

Given this is a back and forth discussion we had. This feels more appropriate to return a null, rather than fail fast

DHowett-MSFT: Absolutely not.

The user requested a path. We cannot fulfill that. These APIs cannot "Fail" in the classical sense.

---
In reply to: [85839205](https://github.com/Microsoft/WinObjC/pull/1255#discussion_r85839205) [](ancestors = 85839205)

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -365,22 +395,20 @@ void CGPathAddArc(CGMutablePathRef path,
  @Status Interoperable
 */
 void CGPathMoveToPoint(CGMutablePathRef path, const CGAffineTransform* transform, CGFloat x, CGFloat y) {
```
msft-Jeyaram: >path [](start = 40, length = 4)

validity check here.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -78,45 +92,49 @@ void preparePathForEditing() {
...
 
-    void endFigure(D2D1_FIGURE_END figureStatus) {
-        if (!_impl.isFigureClosed) {
-            _impl.geometrySink->EndFigure(figureStatus);
-            _impl.isFigureClosed = true;
+    void EndFigure(D2D1_FIGURE_END figureStatus) {
+        if (!_impl.figureClosed) {
+            GeometrySink()->EndFigure(figureStatus);
+            _impl.figureClosed = true;
+            ;
```
msft-Jeyaram: >  ; [](start = 10, length = 3)

extra ; remove?

Frameworks/CoreGraphics/CGPath.mm
DHowett-MSFT: This is getting completely out-of-hand.
We are supposed to introduce these sorts of things to *make the code clearer*.

`Impl` is a dumb data container, for all intents and most purposes. This isn't an exposure across a public interface. If every private implementation detail needs a protected-private-public interface, we'll find ourselves drowning in methods with identical names where a basic indirect and field access would work.

`Impl` is **literally** the struct we have to use because we can't put the members on the outer struct. It doesn't need to be treated like a separate entity.

---
In reply to: [85838765](https://github.com/Microsoft/WinObjC/pull/1255#discussion_r85838765) [](ancestors = 85838765)

DHowett-MSFT: In short, this all damages code clarity.

---
In reply to: [85839897](https://github.com/Microsoft/WinObjC/pull/1255#discussion_r85839897) [](ancestors = 85839897,85838765)

msft-Jeyaram: understood, but given we have a mixed use, it does damage the code clarity.
We have function that looks more or less like getters and can be used as setters.
e.g Foo() = x;
Then we have some direct access and some use of encapsulations. 
Given these are private, we may do whatever we wish that's sensible, but we should try to avoid damaging the clarity of the code for the future generation.




---
In reply to: [85839929](https://github.com/Microsoft/WinObjC/pull/1255#discussion_r85839929) [](ancestors = 85839929,85839897,85838765)

msft-Jeyaram: understood, but given we have a mixed use, it does damage the code clarity.
We have function that looks more or less like getters and can be used as setters.
e.g Foo() = x;
Then we have some direct access and some use of encapsulations. 
Given these are private, we may do whatever we wish that's sensible, but we should try to avoid damaging the clarity of the code for the future generation.

---
In reply to: [85839929](https://github.com/Microsoft/WinObjC/pull/1255#discussion_r85839929) [](ancestors = 85839929,85839897,85838765)

msft-Jeyaram: This is one messed up comment sequence. The order is way out of line.

---
In reply to: [85839929](https://github.com/Microsoft/WinObjC/pull/1255#discussion_r85839929) [](ancestors = 85839929,85839897,85838765)

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -457,8 +479,8 @@ void CGPathAddEllipseInRect(CGMutablePathRef path, const CGAffineTransform* tran
 */
 void CGPathCloseSubpath(CGMutablePathRef path) {
```
msft-Jeyaram: >path [](start = 41, length = 4)

validity check here.
Also a validity sweep of this file should be good :)

tests/testapps/CGCatalog/CGCatalog/CGPathApplyViewController.m
```diff
@@ -55,6 +55,9 @@ - (void)loadView {
         // Unnecessary as close subpath will finish this line off but for the sake of consistency, have this here.
         CGPathAddLineToPoint(thepath, NULL, 200, 35);
 
+        CGMutablePathRef nullPath = NULL;
+        CGRect ourRect = CGPathGetBoundingBox(nullPath);
+
```
DHowett-MSFT: wut

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -485,35 +509,30 @@ CGRect CGPathGetBoundingBox(CGPathRef path) {
...
+        return true;
     }
 
     UINT32 count;
 
-    path->closePath();
+    path->ClosePath();
 
-    path->_impl.pathGeometry->GetFigureCount(&count);
+    path->PathGeometry()->GetFigureCount(&count);
```
msft-Jeyaram: >  path->PathGeometry()->GetFigureCount(&count); [](start = 2, length = 47)

This can FAIL

msft-Jeyaram: > path->PathGeometry()->GetFigureCount(&count); [](start = 3, length = 46)

RETURN_FALSE_IF_FAILED

Frameworks/CoreGraphics/CGPath.mm
MSFTFox: Yeah, I'm confused as to what this is actually referencing now. Care to start a new comment thread here?

---
In reply to: [85841234](https://github.com/Microsoft/WinObjC/pull/1255#discussion_r85841234) [](ancestors = 85841234,85839929,85839897,85838765)

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -157,21 +190,22 @@ CGPathRef CGPathCreateCopy(CGPathRef path) {
...
-    FAIL_FAST_IF_FAILED(path->_impl.pathGeometry->Stream(mutableRet->_impl.geometrySink.Get()));
+    if (FAILED(path->PathGeometry()->Stream(mutableRet->GeometrySink().Get()))) {
+        CGPathRelease(mutableRet);
+        return NULL;
+    }
 
-    mutableRet->_impl.currentPoint = path->_impl.currentPoint;
-    mutableRet->_impl.startingPoint = path->_impl.startingPoint;
+    mutableRet->CurrentPoint() = path->CurrentPoint();
+    mutableRet->StartingPoint() = path->StartingPoint();
```
msft-Jeyaram: is figureClosed need to be copied over?

msft-Jeyaram:
>     CGDrawView* cgView = [[CGDrawView alloc] initWithFrame:self.view.bounds drawOptions:self.options];

do we need to release this or is it managed by the View somehow?

---
Refers to: tests/testapps/CGCatalog/CGCatalog/CGPathApplyViewController.m:30 in 3c21e8f. [](commit_id = 3c21e8fa7033a0c736f6e1b9865c261ff987bf3a, deletion_comment = False)

Frameworks/CoreGraphics/CGPath.mm
DHowett-MSFT: @bbowman You're absolutely right in this case: an lvalue reference screams for a refactor. In a lot of other cases, however, we're advocating for identically-named functions two or more levels deep to do simple data access, which is less unequivocally-right than not using lvalue references.

---
In reply to: [85843540](https://github.com/Microsoft/WinObjC/pull/1255#discussion_r85843540) [](ancestors = 85843540,85838588)

Frameworks/CoreGraphics/CGPath.mm
MSFTFox: Nope, the new figure is assumed to be open since it would just be additional work to close/re-open it for editing.

---
In reply to: [85843401](https://github.com/Microsoft/WinObjC/pull/1255#discussion_r85843401) [](ancestors = 85843401)

MSFTFox:
>     CGDrawView* cgView = [[CGDrawView alloc] initWithFrame:self.view.bounds drawOptions:self.options];

Ignore this file, I never meant to edit this, it was merely a quick validation check on null paths and what the return value is for some APIs.

---
In reply to: [257435345](https://github.com/Microsoft/WinObjC/pull/1255#issuecomment-257435345) [](ancestors = 257435345)

---
Refers to: tests/testapps/CGCatalog/CGCatalog/CGPathApplyViewController.m:30 in 3c21e8f. [](commit_id = 3c21e8fa7033a0c736f6e1b9865c261ff987bf3a, deletion_comment = False)

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -157,21 +190,22 @@ CGPathRef CGPathCreateCopy(CGPathRef path) {
...
 
     // In order to call stream and copy the contents of the original path into the
     // new copy we must close this path.
     // Otherwise the D2D calls will return that a bad state has been entered.
-    path->closePath();
+    path->ClosePath();
 
-    FAIL_FAST_IF_FAILED(path->_impl.pathGeometry->Stream(mutableRet->_impl.geometrySink.Get()));
+    if (FAILED(path->PathGeometry()->Stream(mutableRet->GeometrySink().Get()))) {
+        CGPathRelease(mutableRet);
```
bbowman: >CGPathRelease(mutableRet); [](start = 8, length = 26)

seems like an RAII construct may be useful for ensuring resources are properly released?

Frameworks/CoreGraphics/CGPath.mm
DHowett-MSFT: I would expect a new path to have *no* open figure?

---
In reply to: [85844394](https://github.com/Microsoft/WinObjC/pull/1255#discussion_r85844394) [](ancestors = 85844394,85843401)

tests/unittests/CoreGraphics/CGPathTests.mm
```diff
@@ -703,4 +703,215 @@ void EXPECT_SIZEEQ(CGSize pathSize, CGFloat width, CGFloat height) {
...
+    CGRect theRectangle = CGRectMake(50, 50, 100, 100);
+    CGMutablePathRef path = CGPathCreateWithRect(theRectangle, nullptr);
+
+    EXPECT_TRUE(CGPathContainsPoint(path, nullptr, CGPointMake(75, 75), NO));
+    EXPECT_FALSE(CGPathContainsPoint(path, nullptr, CGPointMake(200, 200), NO));
+
+    CGPathRelease(path);
+}
+
+bool testSymmetricEquivalence(CGPathRef path1, CGPathRef path2) {
```
msft-Jeyaram: >ol te [](start = 2, length = 5)

static?

Frameworks/CoreGraphics/CGPath.mm
MSFTFox: figureClosed represents whether or not a CGPath has closed the geometry sink. A newly created geometry sink being closed is just madness.

---
In reply to: [85844653](https://github.com/Microsoft/WinObjC/pull/1255#discussion_r85844653) [](ancestors = 85844653,85844394,85843401)

Frameworks/CoreGraphics/CGPath.mm
rajsesh-msft: I tend to agree with @msft-Jeyaram and @bbowman on this.  We should use setters where applicable in cases such as these.  @DHowett-MSFT What are the cases where this is better?

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -78,45 +92,49 @@ void preparePathForEditing() {
...
+        return S_OK;
     }
 
-    void closePath() {
-        if (_impl.geometrySink) {
-            endFigure(D2D1_FIGURE_END_OPEN);
-            _impl.geometrySink->Close();
-            _impl.geometrySink = nullptr;
+    void ClosePath() {
+        if (GeometrySink()) {
```
rajsesh-msft: The right way to keep _impl encapsulated and still do this would be to do _impl.ClosePath() here and implement the rest of the code in the impl class.  You haven't really helped encapsulation by doing what you have now - just the variable name.

Frameworks/CoreGraphics/CGPath.mm
rajsesh-msft: I agree this is getting out of control because it is not clear what pattern we are using here.  Although the impl struct is to work around inability to add things directly to outer struct, the entire thing fits into the proxy pattern.  From that perspective, forwarding all calls to the impl makes a lot of sense.  

The reason this feels like getting out of hand is because we are not encapsulating Impl the right way.  There is nothing wrong in writing plain old c code and access the struct members directly, but given we are in C++ land, we might as well be better about it.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -125,8 +142,30 @@ CFTypeID CGPathGetTypeID() {
 }
 
 static Boolean __CGPathEqual(CFTypeRef cf1, CFTypeRef cf2) {
+    RETURN_FALSE_IF(!cf1);
```
msft-Jeyaram: >RETURN_FALSE_IF(!cf1); [](start = 4, length = 22)

what if both of them are NULL?

DHowett-MSFT:
:shipit:

PR #1263, opened 2016-10-28
Fix improper retain level on return in [UIFont initWithCoder]

### LINKED ISSUE #1260: Graphics test game crashes when played

ms-jihua:
Fixes #1260

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1263)

### DIFF
--- a/Frameworks/UIKit/UICTFont.mm
+++ b/Frameworks/UIKit/UICTFont.mm
@@ -24,6 +24,19 @@
 
 PROTOTYPE_CLASS_REQUIRED_IMPLS(UICTFont)
 
+// WinObjC-only extension for UINibUnarchiver
+- (instancetype)initWithCoder:(NSCoder*)coder {
+    NSString* name = [coder decodeObjectForKey:@"UIFontName"];
+    if ([name length] < 1) {
+        // fallback to default if could not find a font name
+        name = (__bridge NSString*)kCTFontDefaultFontName;
+    }
+
+    CGFloat size = [coder decodeFloatForKey:@"UIFontPointSize"];
+
+    return reinterpret_cast<UIFontPrototype*>(static_cast<UIFont*>(CTFontCreateWithName((__bridge CFStringRef)name, size, nullptr)));
+}
+
 @end
 #pragma endregion
 
--- a/Frameworks/UIKit/UICTFont.mm
+++ b/Frameworks/UIKit/UICTFont.mm
@@ -27,7 +27,7 @@ PROTOTYPE_CLASS_REQUIRED_IMPLS(UICTFont)
 // WinObjC-only extension for UINibUnarchiver
 - (instancetype)initWithCoder:(NSCoder*)coder {
     NSString* name = [coder decodeObjectForKey:@"UIFontName"];
-    if ([name length] < 1) {
+    if ([name length] == 0) {
         // fallback to default if could not find a font name
         name = (__bridge NSString*)kCTFontDefaultFontName;
     }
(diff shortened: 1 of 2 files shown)


msft-Jeyaram:
>     }

fix this up too?

This looks horribly bad. 

---
Refers to: Frameworks/UIKit/UIFont.mm:311 in 99920d9. [](commit_id = 99920d99ec09a9cabeee28428a65ca16835b7dff, deletion_comment = False)

msft-Jeyaram:
>     }

cleaned up and follow a proper format?
It has a weird pattern of self assignment and pretty much using the objects static method to generate an object within the shell object.
I'm sure it can be cleaned up using CTFontCreateWithName directly or a wrap around it (if we want to keep encapsulation) 

---
In reply to: [257024231](https://github.com/Microsoft/WinObjC/pull/1263#issuecomment-257024231) [](ancestors = 257024231)

---
Refers to: Frameworks/UIKit/UIFont.mm:311 in 99920d9. [](commit_id = 99920d99ec09a9cabeee28428a65ca16835b7dff, deletion_comment = False)

msft-Jeyaram:
:clock1:

DHowett-MSFT:
:shipit:

Frameworks/UIKit/UICTFont.mm
```diff
@@ -24,6 +24,19 @@ @implementation UIFontPrototype
...
+    NSString* name = [coder decodeObjectForKey:@"UIFontName"];
+    if ([name length] < 1) {
+        // fallback to default if could not find a font name
+        name = (__bridge NSString*)kCTFontDefaultFontName;
+    }
+
+    CGFloat size = [coder decodeFloatForKey:@"UIFontPointSize"];
+
+    return reinterpret_cast<UIFontPrototype*>(static_cast<UIFont*>(CTFontCreateWithName((__bridge CFStringRef)name, size, nullptr)));
+}
```
bbowman: Is this not valuable for derived? (if thats a thing) UIFonts. I'm just slightly confused as initWithCoder isn't a typical designated initializer is it?  Is there something we should be providing on the base class that calls a different designated initializer?

Frameworks/UIKit/UICTFont.mm
```diff
@@ -24,6 +24,19 @@ @implementation UIFontPrototype
 
 PROTOTYPE_CLASS_REQUIRED_IMPLS(UICTFont)
 
+// WinObjC-only extension for UINibUnarchiver
+- (instancetype)initWithCoder:(NSCoder*)coder {
+    NSString* name = [coder decodeObjectForKey:@"UIFontName"];
+    if ([name length] < 1) {
```
bbowman: >[name length] < 1 [](start = 8, length = 17)

length < 1 seems like a curious way of writing this.

bbowman:
:shipit:

Frameworks/UIKit/UICTFont.mm
ms-jihua: - re: valuable for derived: this is a WinObjC-only extension, and _we_ don't have any UIFont derived classes, nor should we ever.
- re: designated initializers: UIFont has no initializers we could route to :(

---
In reply to: [85619643](https://github.com/Microsoft/WinObjC/pull/1263#discussion_r85619643) [](ancestors = 85619643)

Frameworks/UIKit/UICTFont.mm
ms-jihua: yeah, now that you mention it (this was copied forward from some really old code).  I'll change it to == 0.

---
In reply to: [85619699](https://github.com/Microsoft/WinObjC/pull/1263#discussion_r85619699) [](ancestors = 85619699)

PR #1289, opened 2016-10-31
Yet again increase the robustness of the NSURLSession tests.

bbowman:
Description:
These tests also used floor instead of round and it turns out the expected size was 11Mb. Oops.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1289)


msft-Jeyaram:
:shipit:

DHowett-MSFT:
`bbowman:bbowman/NSUrlSessionIsTheGiftThatKeepsGiving` 😆 😆

DHowett-MSFT:
:shipit:

PR #1295, opened 2016-11-02
Fixing heap corruption when using core text to draw into opaque layers.

rajsesh-msft:
Fixing heap corruption when using core text to draw into opaque layers.

The heap corruption happened in an internal test app.

The fix is to use BGRX pixel format for opaque layers - this is supported
by WIC render target.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1295)

### DIFF
--- a/Frameworks/QuartzCore/CALayer.mm
+++ b/Frameworks/QuartzCore/CALayer.mm
@@ -574,7 +574,7 @@ CGContextRef CreateLayerContentsBitmapContext32(int width, int height) {
                 if (useVector) {
                     // target = new CGVectorImage(width, height, _ColorBGR);
                 } else {
-                    drawContext = _CGBitmapContextCreateWithFormat(width, height, _ColorBGR);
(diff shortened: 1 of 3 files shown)


rajsesh-msft:

msft-Jeyaram:
:shipit:

ms-jihua:
:shipit:

jaredhms:
:shipit:

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -574,7 +574,7 @@ - (void)display {
                 if (useVector) {
                     // target = new CGVectorImage(width, height, _ColorBGR);
                 } else {
-                    drawContext = _CGBitmapContextCreateWithFormat(width, height, _ColorBGR);
+                    drawContext = _CGBitmapContextCreateWithFormat(width, height, _ColorBGRX);
```
rajsesh-msft: We were allocating 3 bytes per pixel and writing into 4 essentially.

PR #1302, opened 2016-11-02
Properly pass font weight, stretch, style, to XamlCompositor

ms-jihua:
Update from using simplistic mapping using the font's symbolic traits, which ignored stretch,
to getting the weight/stretch/style directly from the font.

UILabel-rendered narrow or condensed fonts were previously clipped at the sides
because the CoreText-calculated width would be for a narrow font, but UILabel would attempt to render a normal font.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1302)

### DIFF
--- a/Frameworks/CoreText/CTFont.mm
+++ b/Frameworks/CoreText/CTFont.mm
@@ -903,3 +903,30 @@ CFTypeID CTFontGetTypeID() {
     static CFTypeID __kCTFontTypeID = _CFRuntimeRegisterClass(&__CTFontClass);
     return __kCTFontTypeID;
 }
+
+// Private function for getting font weight for XAML
+DWRITE_FONT_WEIGHT _CTFontGetDWriteWeight(CTFontRef font) {
+    ComPtr<IDWriteFontFace3> fontFace3;
+    if (font && SUCCEEDED(font->_dwriteFontFace.As(&fontFace3))) {
+        return fontFace3->GetWeight();
+    }
+    return DWRITE_FONT_WEIGHT_NORMAL;
+}
+
+// Private function for getting font stretch for XAML
+DWRITE_FONT_STRETCH _CTFontGetDWriteStretch(CTFontRef font) {
+    ComPtr<IDWriteFontFace3> fontFace3;
+    if (font && SUCCEEDED(font->_dwriteFontFace.As(&fontFace3))) {
+        return fontFace3->GetStretch();
+    }
+    return DWRITE_FONT_STRETCH_NORMAL;
+}
+
+// Private function for getting font style for XAML
+DWRITE_FONT_STYLE _CTFontGetDWriteStyle(CTFontRef font) {
+    ComPtr<IDWriteFontFace3> fontFace3;
+    if (font && SUCCEEDED(font->_dwriteFontFace.As(&fontFace3))) {
+        return fontFace3->GetStyle();
+    }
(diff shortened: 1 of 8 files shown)


Frameworks/UIKit/StarboardXaml/CompositorInterface.mm
```diff
@@ -329,9 +329,10 @@ void SetParams(UIFont* font,
         _centerVertically = centerVertically;
         _lineHeight = [font ascender] - [font descender];
 
-        int mask = [font fontDescriptor].symbolicTraits;
-        _isBold = (mask & UIFontDescriptorTraitBold) > 0;
-        _isItalic = (mask & UIFontDescriptorTraitItalic) > 0;
+        _fontWeight = [font _fontWeight];
+        _fontStretch = [font _fontStretch];
+        _fontStyle = [font _fontStyle];
```
jaredhms: all of the code you're changing here is gone in the CA refactor, but I suppose this can go in and I'll have to forklift it to our UILabel impl.  Seems straightforward enough, but please ping me when this goes in so I can do yet another merge :/

jaredhms:
:shipit:

Frameworks/CoreText/CTFont.mm
```diff
@@ -902,4 +902,31 @@ CFDataRef CTFontCopyTable(CTFontRef font, CTFontTableTag table, CTFontTableOptio
 CFTypeID CTFontGetTypeID() {
     static CFTypeID __kCTFontTypeID = _CFRuntimeRegisterClass(&__CTFontClass);
     return __kCTFontTypeID;
+}
```
rajsesh-msft: since we are adding these private apis for the sake of passing them along to xaml, we should make just one function that gets all this information.

Frameworks/UIKit/StarboardXaml/CompositorInterface.mm
ms-jihua: got it. sorry about this :(

---
In reply to: [86254978](https://github.com/Microsoft/WinObjC/pull/1302#discussion_r86254978) [](ancestors = 86254978)

Frameworks/UIKit/StarboardXaml/XamlCompositor.cpp
```diff
@@ -561,17 +561,9 @@ void DisplayTextureXamlGlyphs::ConstructGlyphs(const Microsoft::WRL::Wrappers::H
...
-    } else {
-        textControl->FontWeight = Windows::UI::Text::FontWeights::Normal;
-    }
-
-    if (_isItalic) {
-        textControl->FontStyle = Windows::UI::Text::FontStyle::Italic;
-    } else {
-        textControl->FontStyle = Windows::UI::Text::FontStyle::Normal;
-    }
+    textControl->FontWeight = Windows::UI::Text::FontWeight{ static_cast<unsigned short>(_fontWeight) };
```
aballway: are these all 1-1 mappings?

Frameworks/CoreText/CTFont.mm
ms-jihua: Discussed offline - I pursued this route earlier, but exporting a struct through this many layers ended up being a much more ugly change. Leaving as is.

---
In reply to: [86256084](https://github.com/Microsoft/WinObjC/pull/1302#discussion_r86256084) [](ancestors = 86256084)

PR #1306, opened 2016-11-02
Fix infinite loop in __CreateDWriteTextLayout

### LINKED ISSUE #1294: Test app hangs while loading page due to infinite loop in __CreateDWriteTextLayout
```
   NSRange attributeRange;
    for (size_t i = 0; i < [subString length]; i += attributeRange.length) { 
```
attributeRange.length is 0 causing the app to hang.

aballway:
Were not handling edge case where frames were being created with ranges not aligned with attributes in attributed strings.

Fixes #1294

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1306)

### DIFF
--- a/Frameworks/CoreText/DWriteWrapper_CoreText.mm
+++ b/Frameworks/CoreText/DWriteWrapper_CoreText.mm
@@ -204,7 +204,7 @@ static ComPtr<IDWriteTextLayout> __CreateDWriteTextLayout(CFAttributedStringRef 
     for (size_t i = 0; i < [subString length]; i += attributeRange.length) {
         NSDictionary* attribs = [static_cast<NSAttributedString*>(string) attributesAtIndex:i + range.location
                                                                       longestEffectiveRange:&attributeRange
-                                                                                    inRange:{ i, [subString length] }];
+                                                                                    inRange:{ i + range.location, [subString length] }];
 
         const DWRITE_TEXT_RANGE dwriteRange = { attributeRange.location, attributeRange.length };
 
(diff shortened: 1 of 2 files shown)


Frameworks/CoreText/DWriteWrapper_CoreText.mm
```diff
@@ -204,7 +204,7 @@ bool _CloneDWriteGlyphRun(_In_ DWRITE_GLYPH_RUN const* src, _Out_ DWRITE_GLYPH_R
     for (size_t i = 0; i < [subString length]; i += attributeRange.length) {
         NSDictionary* attribs = [static_cast<NSAttributedString*>(string) attributesAtIndex:i + range.location
                                                                       longestEffectiveRange:&attributeRange
-                                                                                    inRange:{ i, [subString length] }];
+                                                                                    inRange:{ i + range.location, [subString length] }];
```
ms-jihua: >inRange:{ i + range.location, [subString length] }]; [](start = 84, length = 52)

Mm. This is kind of a big ask, but in retrospect, I would prefer if you relied on [NSAttributedString enumerateAttributesInRange:options:usingBlock:] to do the loop instead. You're a lot less likely to make a mistake that way. (Note that you'd have to declare incompatibleAttributeFlag and textLayout as __block if you do this)

ms-jihua:
:shipit:

Frameworks/CoreText/DWriteWrapper_CoreText.mm
aballway: I'll do this the next time I add an attribute we handle here or on any cleanup pass I make.

rajsesh-msft:
@aballway is this ready to merge?

PR #1307, opened 2016-11-02
Implement drawing CGPaths through CGContext

MSFTFox:
Unify the factory for each context's render target as any geometry object must be created by the same factory instance in order to be rendered there.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1307)

### DIFF
--- a/Frameworks/CoreGraphics/CGBitmapImage.mm
+++ b/Frameworks/CoreGraphics/CGBitmapImage.mm
@@ -453,7 +453,8 @@ ID2D1RenderTarget* CGBitmapImageBacking::GetRenderTarget() {
     if (_renderTarget == nullptr) {
         BYTE* imageData = static_cast<BYTE*>(LockImageData());
         ComPtr<IWICBitmap> wicBitmap = Make<CGIWICBitmap>(this, SurfaceFormat());
-        ComPtr<ID2D1Factory> d2dFactory = _GetD2DFactoryInstance();
+        ComPtr<ID2D1Factory> d2dFactory;
+        FAIL_FAST_IF_FAILED(_CGGetD2DFactory(&d2dFactory));
         ComPtr<ID2D1RenderTarget> renderTarget;
         THROW_IF_FAILED(d2dFactory->CreateWicBitmapRenderTarget(wicBitmap.Get(), D2D1::RenderTargetProperties(), &renderTarget));
         _renderTarget = renderTarget.Detach();
--- a/Frameworks/CoreGraphics/CGBitmapImage.mm
+++ b/Frameworks/CoreGraphics/CGBitmapImage.mm
@@ -26,6 +26,8 @@
 
 #import "LoggingNative.h"
 
+#import "D2DWrapper.h"
+
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wdeprecated-register"
 
@@ -451,8 +453,7 @@ ID2D1RenderTarget* CGBitmapImageBacking::GetRenderTarget() {
     if (_renderTarget == nullptr) {
         BYTE* imageData = static_cast<BYTE*>(LockImageData());
         ComPtr<IWICBitmap> wicBitmap = Make<CGIWICBitmap>(this, SurfaceFormat());
-        ComPtr<ID2D1Factory> d2dFactory;
-        THROW_IF_FAILED(D2D1CreateFactory(D2D1_FACTORY_TYPE_SINGLE_THREADED, __uuidof(ID2D1Factory), &d2dFactory));
+        ComPtr<ID2D1Factory> d2dFactory = _GetD2DFactoryInstance();
         ComPtr<ID2D1RenderTarget> renderTarget;
         THROW_IF_FAILED(d2dFactory->CreateWicBitmapRenderTarget(wicBitmap.Get(), D2D1::RenderTargetProperties(), &renderTarget));
         _renderTarget = renderTarget.Detach();
--- a/Frameworks/CoreGraphics/CGContext.mm
+++ b/Frameworks/CoreGraphics/CGContext.mm
@@ -27,6 +27,7 @@
 #import <CoreGraphics/CGGradient.h>
 #import "CGColorSpaceInternal.h"
 #import "CGContextInternal.h"
+#import "CGPathInternal.h"
 
 #import <CFCppBase.h>
 
@@ -1646,7 +1647,8 @@ void CGContextFillRects(CGContextRef context, const CGRect* rects, size_t count)
 */
 void CGContextDrawPath(CGContextRef context, CGPathDrawingMode mode) {
     NOISY_RETURN_IF_NULL(context);
-    UNIMPLEMENTED();
+
+    __CGContextDrawGeometry(context, _getAndClosePathGeometry(context->Path()), mode);
     context->ClearPath();
 }
 
--- a/Frameworks/CoreGraphics/CGContext.mm
+++ b/Frameworks/CoreGraphics/CGContext.mm
@@ -1647,9 +1647,12 @@ void CGContextFillRects(CGContextRef context, const CGRect* rects, size_t count)
 */
 void CGContextDrawPath(CGContextRef context, CGPathDrawingMode mode) {
     NOISY_RETURN_IF_NULL(context);
-
-    __CGContextDrawGeometry(context, _getAndClosePathGeometry(context->Path()), mode);
-    context->ClearPath();
+    if (context->HasPath()) {
+        ID2D1Geometry* pGeometry;
+        FAIL_FAST_IF_FAILED(_CGPathGetGeometry(context->Path(), &pGeometry));
+        __CGContextDrawGeometry(context, pGeometry, mode);
+        context->ClearPath();
+    }
 }
 
 /**
--- a/Frameworks/CoreGraphics/CGContext.mm
+++ b/Frameworks/CoreGraphics/CGContext.mm
@@ -1648,9 +1648,9 @@ void CGContextFillRects(CGContextRef context, const CGRect* rects, size_t count)
 void CGContextDrawPath(CGContextRef context, CGPathDrawingMode mode) {
     NOISY_RETURN_IF_NULL(context);
     if (context->HasPath()) {
-        ID2D1Geometry* pGeometry;
+        ComPtr<ID2D1Geometry> pGeometry;
         FAIL_FAST_IF_FAILED(_CGPathGetGeometry(context->Path(), &pGeometry));
-        FAIL_FAST_IF_FAILED(__CGContextDrawGeometry(context, pGeometry, mode));
+        FAIL_FAST_IF_FAILED(__CGContextDrawGeometry(context, pGeometry.Get(), mode));
         context->ClearPath();
     }
 }
--- a/Frameworks/CoreGraphics/CGContext.mm
+++ b/Frameworks/CoreGraphics/CGContext.mm
@@ -1650,7 +1650,7 @@ void CGContextDrawPath(CGContextRef context, CGPathDrawingMode mode) {
     if (context->HasPath()) {
         ID2D1Geometry* pGeometry;
         FAIL_FAST_IF_FAILED(_CGPathGetGeometry(context->Path(), &pGeometry));
-        __CGContextDrawGeometry(context, pGeometry, mode);
+        FAIL_FAST_IF_FAILED(__CGContextDrawGeometry(context, pGeometry, mode));
         context->ClearPath();
     }
 }
--- a/Frameworks/CoreGraphics/CGGraphicBufferImage.mm
+++ b/Frameworks/CoreGraphics/CGGraphicBufferImage.mm
@@ -25,6 +25,8 @@
 #import "LoggingNative.h"
 #import <CGGraphicBufferImage.h>
 
+#import "D2DWrapper.h"
+
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wdeprecated-register"
 
@@ -155,8 +157,7 @@ ID2D1RenderTarget* CGGraphicBufferImageBacking::GetRenderTarget() {
     if (_renderTarget == nullptr) {
         BYTE* imageData = static_cast<BYTE*>(LockImageData());
         ComPtr<IWICBitmap> wicBitmap = Make<CGIWICBitmap>(this, SurfaceFormat());
-        ComPtr<ID2D1Factory> d2dFactory;
-        THROW_IF_FAILED(D2D1CreateFactory(D2D1_FACTORY_TYPE_SINGLE_THREADED, __uuidof(ID2D1Factory), &d2dFactory));
+        ComPtr<ID2D1Factory> d2dFactory = _GetD2DFactoryInstance();
         ComPtr<ID2D1RenderTarget> renderTarget;
         THROW_IF_FAILED(d2dFactory->CreateWicBitmapRenderTarget(wicBitmap.Get(), D2D1::RenderTargetProperties(), &renderTarget));
         _renderTarget = renderTarget.Detach();
--- a/Frameworks/CoreGraphics/CGGraphicBufferImage.mm
+++ b/Frameworks/CoreGraphics/CGGraphicBufferImage.mm
@@ -157,7 +157,8 @@ ID2D1RenderTarget* CGGraphicBufferImageBacking::GetRenderTarget() {
     if (_renderTarget == nullptr) {
         BYTE* imageData = static_cast<BYTE*>(LockImageData());
         ComPtr<IWICBitmap> wicBitmap = Make<CGIWICBitmap>(this, SurfaceFormat());
-        ComPtr<ID2D1Factory> d2dFactory = _GetD2DFactoryInstance();
+        ComPtr<ID2D1Factory> d2dFactory;
+        FAIL_FAST_IF_FAILED(_CGGetD2DFactory(&d2dFactory));
         ComPtr<ID2D1RenderTarget> renderTarget;
         THROW_IF_FAILED(d2dFactory->CreateWicBitmapRenderTarget(wicBitmap.Get(), D2D1::RenderTargetProperties(), &renderTarget));
         _renderTarget = renderTarget.Detach();
--- a/Frameworks/CoreGraphics/CGPath.mm
+++ b/Frameworks/CoreGraphics/CGPath.mm
@@ -150,10 +150,10 @@ struct __CGPath : CoreFoundation::CppBase<__CGPath, __CGPathImpl> {
 };
 
 HRESULT _CGPathGetGeometry(CGPathRef path, ID2D1Geometry** pGeometry) {
-    if (path && pGeometry) {
-        RETURN_IF_FAILED(path->ClosePath());
-        path->GetPathGeometry().CopyTo(pGeometry);
-    }
+    RETURN_HR_IF_NULL(E_POINTER, pGeometry);
+    RETURN_HR_IF_NULL(E_POINTER, path);
+    RETURN_IF_FAILED(path->ClosePath());
+    path->GetPathGeometry().CopyTo(pGeometry);
     return S_OK;
 }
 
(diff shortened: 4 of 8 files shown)


Frameworks/include/CGPathInternal.h
```diff
@@ -51,4 +55,6 @@ struct CGPathElementInternal : CGPathElement {
 };
 typedef struct CGPathElementInternal CGPathElementInternal;
 
+ID2D1Geometry* _getAndClosePathGeometry(CGPathRef path);
```
msft-Jeyaram: >_getAndClosePathGeometry [](start = 15, length = 24)

extra nit: _CGPathGetAndClosePathGeometry(...) ?

DHowett-MSFT: please something more like:

```
HRESULT _CGPathGetGeometry(CGPathRef path, ID2D1Geometry** pGeometry);
```

---
In reply to: [86264580](https://github.com/Microsoft/WinObjC/pull/1307#discussion_r86264580) [](ancestors = 86264580)

Frameworks/CoreGraphics/CGGraphicBufferImage.mm
```diff
@@ -155,8 +157,7 @@
     if (_renderTarget == nullptr) {
         BYTE* imageData = static_cast<BYTE*>(LockImageData());
         ComPtr<IWICBitmap> wicBitmap = Make<CGIWICBitmap>(this, SurfaceFormat());
-        ComPtr<ID2D1Factory> d2dFactory;
-        THROW_IF_FAILED(D2D1CreateFactory(D2D1_FACTORY_TYPE_SINGLE_THREADED, __uuidof(ID2D1Factory), &d2dFactory));
+        ComPtr<ID2D1Factory> d2dFactory = _GetD2DFactoryInstance();
```
DHowett-MSFT: >_GetD2DFactoryInstance [](start = 42, length = 22)

This might be going cross-module soon; perhaps it should be named _CGGetD2DFactory(), and have the signature

```
HRESULT _CGGetD2DFactory(ID2D1Factory** pFactory);
```

DHowett-MSFT:
:clock1:

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1646,7 +1647,8 @@ void CGContextFillRects(CGContextRef context, const CGRect* rects, size_t count)
 */
 void CGContextDrawPath(CGContextRef context, CGPathDrawingMode mode) {
     NOISY_RETURN_IF_NULL(context);
-    UNIMPLEMENTED();
+
+    __CGContextDrawGeometry(context, _getAndClosePathGeometry(context->Path()), mode);
```
DHowett-MSFT: >_getAndClosePathGeometry [](start = 37, length = 24)

```
if(!HasPath)
// we can abort
```

Otherwise, `Path()` will create an empty path just to draw it

DHowett-MSFT: >__CGContextDrawGeometry [](start = 4, length = 23)

We may need to invert the user transform on this draw, because otherwise we could apply a double-transformed path.

MSFTFox: I made a note of this. We'll explore this more with the affine transform work item.

---
In reply to: [86265029](https://github.com/Microsoft/WinObjC/pull/1307#discussion_r86265029) [](ancestors = 86265029)

tests/unittests/CoreGraphics/CGPathTests.mm
```diff
@@ -412,7 +412,7 @@ CGPathRef newPathForRoundRect(CGRect rect, CGFloat radius) {
     return path;
 }
 
-TEST(CGPath, CGPathContainsPointOutsideRect) {
+DISABLED_TEST(CGPath, CGPathContainsPointOutsideRect) {
```
MSFTFox: >DISABLED_TEST [](start = 0, length = 13)

These tests should have been disabled initially since they involve work with arcs/curves. These will be re-enabled then.

Frameworks/CoreGraphics/D2DWrapper.mm
```diff
@@ -20,15 +20,21 @@
...
+HRESULT _CGGetD2DFactory(ID2D1Factory** factory) {
+    static ComPtr<ID2D1Factory> sFactory;
+    static HRESULT sHr;
+    static dispatch_once_t dispatchToken;
+
+    dispatch_once(&dispatchToken, ^{
+        sHr = __createD2DFactory(&sFactory);
+    });
+    sFactory.Get()->AddRef();
+    *factory = sFactory.Get();
```
DHowett-MSFT: prefer `sFactory.CopyTo(pFactory)`

Frameworks/CoreGraphics/D2DWrapper.mm
```diff
@@ -20,15 +20,21 @@
 using namespace Microsoft::WRL;
 
 // Private helper for creating a D2DFactory
-static ComPtr<ID2D1Factory> __createD2DFactory() {
-    ComPtr<ID2D1Factory> d2dFactory;
-    FAIL_FAST_IF_FAILED(D2D1CreateFactory(D2D1_FACTORY_TYPE_SINGLE_THREADED, __uuidof(ID2D1Factory), &d2dFactory));
-    return d2dFactory;
+static HRESULT __createD2DFactory(ID2D1Factory** factory) {
+    return D2D1CreateFactory(D2D1_FACTORY_TYPE_SINGLE_THREADED, factory);
```
DHowett-MSFT: If we are not using a magic static, there is no real point in having this be a separate function

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1646,8 +1647,12 @@ void CGContextFillRects(CGContextRef context, const CGRect* rects, size_t count)
 */
 void CGContextDrawPath(CGContextRef context, CGPathDrawingMode mode) {
     NOISY_RETURN_IF_NULL(context);
-    UNIMPLEMENTED();
-    context->ClearPath();
+    if (context->HasPath()) {
+        ID2D1Geometry* pGeometry;
+        FAIL_FAST_IF_FAILED(_CGPathGetGeometry(context->Path(), &pGeometry));
+        __CGContextDrawGeometry(context, pGeometry, mode);
+        context->ClearPath();
```
aballway: should do something with the HRESULT here

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1646,8 +1647,12 @@ void CGContextFillRects(CGContextRef context, const CGRect* rects, size_t count)
 */
 void CGContextDrawPath(CGContextRef context, CGPathDrawingMode mode) {
     NOISY_RETURN_IF_NULL(context);
-    UNIMPLEMENTED();
-    context->ClearPath();
+    if (context->HasPath()) {
+        ID2D1Geometry* pGeometry;
```
DHowett-MSFT: Please make this a `ComPtr<ID2D1Geometry>`. Right now you're getting a +1 reference to a Geometry and then dropping it on the floor.

You should almost never use bare COM interface pointers. :)

DHowett-MSFT: Unless you're at a module boundary, or a boundary that makes sense (like CGContext <-> CGPath), and even then only in function parameters.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -113,12 +115,13 @@ HRESULT PreparePathForEditing() {
         return S_OK;
     }
 
-    void ClosePath() {
+    HRESULT ClosePath() {
         if (_impl.geometrySink) {
             EndFigure(D2D1_FIGURE_END_OPEN);
```
DHowett-MSFT: Can this fail?

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -145,6 +149,14 @@ HRESULT InitializeGeometries() {
     }
 };
 
+HRESULT _CGPathGetGeometry(CGPathRef path, ID2D1Geometry** pGeometry) {
+    if (path && pGeometry) {
+        RETURN_IF_FAILED(path->ClosePath());
+        *pGeometry = path->GetPathGeometry().Get();
```
DHowett-MSFT: You might need to `.CopyTo` this. COM out pointers are expected to be given a new +1.

Frameworks/CoreGraphics/D2DWrapper.mm
```diff
@@ -19,16 +19,18 @@
...
+    static dispatch_once_t dispatchToken;
+
+    dispatch_once(&dispatchToken, ^{
+        sHr = D2D1CreateFactory(D2D1_FACTORY_TYPE_SINGLE_THREADED, __uuidof(ID2D1Factory), &sFactory);
+    });
 
-ComPtr<ID2D1Factory> _GetD2DFactoryInstance() {
-    static ComPtr<ID2D1Factory> factory = __createD2DFactory();
-    return factory;
+    sFactory.Get()->AddRef();
```
DHowett-MSFT: *only CopyTo*
Now you are doing +2.

Frameworks/CoreGraphics/CGBitmapImage.mm
```diff
@@ -452,7 +454,7 @@
         BYTE* imageData = static_cast<BYTE*>(LockImageData());
         ComPtr<IWICBitmap> wicBitmap = Make<CGIWICBitmap>(this, SurfaceFormat());
         ComPtr<ID2D1Factory> d2dFactory;
-        THROW_IF_FAILED(D2D1CreateFactory(D2D1_FACTORY_TYPE_SINGLE_THREADED, __uuidof(ID2D1Factory), &d2dFactory));
+        FAIL_FAST_IF_FAILED(_CGGetD2DFactory(&d2dFactory));
         ComPtr<ID2D1RenderTarget> renderTarget;
         THROW_IF_FAILED(d2dFactory->CreateWicBitmapRenderTarget(wicBitmap.Get(), D2D1::RenderTargetProperties(), &renderTarget));
```
bbowman: :-( super sad at the inconsistencies here. I thought the decision was to be error returning out to the public layer at which point a decision would be made on the correct course of action. This still seems internal as it returns a D2D object.  

Additionally, returning a Com object with a raw * usually speaks to a problematic design pattern. This is because it requires the "unjacketing" of the ComObject to return it to the caller without a clear communication on transfer/sharing of ownership which is by definition occuring by handing it out.

bbowman:
>     ComPtr<ID2D1RenderTarget> renderTarget = newImage->Backing()->GetRenderTarget();

For instance, this line actually causes the Render Target to leak unexpectedly. Because GetRenderTarget returns a raw *, this assignment operator is doing an implicit ComPtr<T> construction around the raw *. 

wrl\client.h has the following (can check in Visual studio sdk):

    ComPtr& operator=(_In_opt_ T *other) throw()
    {
        if (ptr_ != other)
        {
            ComPtr(other).Swap(*this); <-- GREAT SADNESS
        }
        return *this;
    }

Which shows the construction around a raw* that does an additional AddRef. (Was already +1 from the .Detach() )

---
Refers to: Frameworks/CoreGraphics/CGContext.mm:1900 in 5eab616. [](commit_id = 5eab61608a771419a62db861cc1e951982a511da, deletion_comment = False)

Frameworks/CoreGraphics/CGPath.mm
MSFTFox: virtual void EndFigure(
   D2D1_FIGURE_END figureEnd
) = 0;

It cannot. Calling it multiple times can result in a bad state though which makes this API slightly weird since it doesn't report that itself.

---
In reply to: [86611357](https://github.com/Microsoft/WinObjC/pull/1307#discussion_r86611357) [](ancestors = 86611357)

bbowman:
> 

Storing a Com Object as a raw* as a member variable is a problematic pattern. It is removing all the niceness of RAII and makes the consumer have to think at all points about the reference count and its management. For instance, this code is completely not needed if you were to store it as a ComPtr and the other code I commented on is easier to rationalize about (I was actually wrong given how this pattern is. The other code is safe but you *really* need to pay attention to understand that).

---
Refers to: Frameworks/CoreGraphics/CGBitmapImage.mm:404 in 5eab616. [](commit_id = 5eab61608a771419a62db861cc1e951982a511da, deletion_comment = False)

bbowman:
>     }

I don't know what data is but it sure seems like a RAII helper like std::unique / shared ptr would help ref count some memory and delete it for you safely when it is done. This pattern is not thread safe here and the fact that you don't just straight up delete it suggests that the ownership model is more than a simple one so knowing what thread all its uses occur on can be problematic.

---
Refers to: Frameworks/CoreGraphics/CGBitmapImage.mm:409 in 5eab616. [](commit_id = 5eab61608a771419a62db861cc1e951982a511da, deletion_comment = False)

Frameworks/CoreGraphics/CGBitmapImage.mm
DHowett-MSFT: The render target generators in CGImage and friends are going away (and are only being updated as necessary as part of this PR); ideally, yes, we'd fix everything all at once. There's little point when it's gonna be turfed. :)

msft-Jeyaram: This code has gone away, i'll be sending out the PR soon.
Actually this whole file is gone.

---
In reply to: [86612430](https://github.com/Microsoft/WinObjC/pull/1307#discussion_r86612430) [](ancestors = 86612430)

MSFTFox: We should sync up.

---
In reply to: [86614591](https://github.com/Microsoft/WinObjC/pull/1307#discussion_r86614591) [](ancestors = 86614591,86612430)

bbowman:
> 

And actually the non owning return raw * while it doesn't leak now, doesn't necessarily correctly describe the ownership as that * could become invalidated at any point while the caller (one of which jackets immediately and minimizes the room for explosions but not totally removes) still has one laying around. 

Point being, don't use raw * unless you have a real good reason.

---
In reply to: [258521890](https://github.com/Microsoft/WinObjC/pull/1307#issuecomment-258521890) [](ancestors = 258521890)

---
Refers to: Frameworks/CoreGraphics/CGBitmapImage.mm:404 in 5eab616. [](commit_id = 5eab61608a771419a62db861cc1e951982a511da, deletion_comment = False)

Frameworks/CoreGraphics/CGContext.mm
bbowman: Yes, yes, yes.

using raw * means you *really* have to think about it and its easy to forget something / mess up. Com stuff gets a ComPtr almost always.

---
In reply to: [86611265](https://github.com/Microsoft/WinObjC/pull/1307#discussion_r86611265) [](ancestors = 86611265)

Frameworks/CoreGraphics/D2DWrapper.mm
bbowman: >sFactory.Get()->AddRef(); [](start = 4, length = 25)

!!!!!!!!!ATCHUNG!!!!!!!!!

manual AddRef / Release of Com Objects almost always speaks to a design patten issue. Especially when used in conjuction with ComPtr methods in other places. 

What is the goal here? You are AddRefing once and then again in the CopyTo. Hooray extra amazing leakage.

DHowett-MSFT: https://github.com/Microsoft/WinObjC/pull/1307#discussion_r86611665 😉

aballway: achtung? atchung sounds like an old heavy cash register

Frameworks/CoreGraphics/D2DWrapper.mm
```diff
@@ -19,16 +19,18 @@
...
-    ComPtr<ID2D1Factory> d2dFactory;
-    FAIL_FAST_IF_FAILED(D2D1CreateFactory(D2D1_FACTORY_TYPE_SINGLE_THREADED, __uuidof(ID2D1Factory), &d2dFactory));
-    return d2dFactory;
-}
+HRESULT _CGGetD2DFactory(ID2D1Factory** factory) {
+    static ComPtr<ID2D1Factory> sFactory;
+    static HRESULT sHr;
+    static dispatch_once_t dispatchToken;
+
+    dispatch_once(&dispatchToken, ^{
```
bbowman: why did we go back to dispatch once? as mentioned before, dispatch_once doesn't buy you much except make code more dependent on  a third party library and harder to read. I preferred the second helper function / you could do something with a c++ lambda if you are concerned about the "multi-return" / "return transforming" you need to do. That is a language feature that we can be more confident in than our pseudo hacked up libdispatch just to get a static initialized.

bbowman:
:clock1:

Frameworks/CoreGraphics/D2DWrapper.mm
bbowman: I spell real goodly.

---
In reply to: [86616809](https://github.com/Microsoft/WinObjC/pull/1307#discussion_r86616809) [](ancestors = 86616809)

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -145,6 +149,14 @@ HRESULT InitializeGeometries() {
     }
 };
 
+HRESULT _CGPathGetGeometry(CGPathRef path, ID2D1Geometry** pGeometry) {
```
msft-Jeyaram: >pGeometry [](start = 59, length = 9)

nit: if pGeometry is invalid, we end up returning S_OK.
adding a check for pGeoMetry and path should help.

RETURN_HR_IF_NULL(E_POINTER,pGeometry);
RETURN_HR_IF_NULL(E_POINTER,path);

msft-Jeyaram:
:shipit:

Frameworks/CoreGraphics/D2DWrapper.mm
```diff
@@ -19,16 +19,12 @@
...
-}
+HRESULT _CGGetD2DFactory(ID2D1Factory** factory) {
+    static ComPtr<ID2D1Factory> sFactory;
+    static HRESULT sHr = D2D1CreateFactory(D2D1_FACTORY_TYPE_SINGLE_THREADED, __uuidof(ID2D1Factory), &sFactory);
 
-ComPtr<ID2D1Factory> _GetD2DFactoryInstance() {
-    static ComPtr<ID2D1Factory> factory = __createD2DFactory();
-    return factory;
+    sFactory.CopyTo(factory);
+    RETURN_HR(sHr);
```
DHowett-MSFT: >RETURN_HR [](start = 4, length = 9)

return sHR?

DHowett-MSFT: just a nit.

---
In reply to: [86633664](https://github.com/Microsoft/WinObjC/pull/1307#discussion_r86633664) [](ancestors = 86633664)

MSFTFox: Figured this was better from result.h

"// Always returns a known result (HRESULT) - logs failures in pre-release
#define RETURN_HR(hr) __RETURN_HR(wil::verify_hresult(hr), #hr)"

---
In reply to: [86633680](https://github.com/Microsoft/WinObjC/pull/1307#discussion_r86633680) [](ancestors = 86633680,86633664)

Frameworks/CoreGraphics/CGBitmapImage.mm
```diff
@@ -452,7 +454,7 @@
         BYTE* imageData = static_cast<BYTE*>(LockImageData());
         ComPtr<IWICBitmap> wicBitmap = Make<CGIWICBitmap>(this, SurfaceFormat());
         ComPtr<ID2D1Factory> d2dFactory;
-        THROW_IF_FAILED(D2D1CreateFactory(D2D1_FACTORY_TYPE_SINGLE_THREADED, __uuidof(ID2D1Factory), &d2dFactory));
+        FAIL_FAST_IF_FAILED(_CGGetD2DFactory(&d2dFactory));
```
rajsesh-msft: Please no, THROW_IF_FAILED() is marginally better.  FAIL_FAST is essentially abort, and as framework we should seldom be using it.  And in other places too, per our offline discussion.

PR #1313, opened 2016-11-03
Rework CFCppBase to remove Impl classes and the Details namespace.

DHowett-MSFT:
* Adds support for calling non-nullary constructors on the new CF Type.
* Removes the _CFFinalize chain.
* Removes the _CFInitialize chain.
* Uses some dodgy methods to work around the undefined behaviour
  inherent in placement-newing a block of partially-initialized memory

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1313)

### DIFF
--- a/Frameworks/include/CFCppBase.h
+++ b/Frameworks/include/CFCppBase.h
@@ -27,22 +27,15 @@
  * - Inheritance is somewhat fiddly and difficult to pull off.
  *
  * Details:
- * CppBase<NewCFType, NewCFTypeImpl[, parent]>
- * NewCFType: The name of the CF type inheriting CppBase.
- * NewCFTypeImpl: The inner struct or class containing NewCFType's members.
- * parent: Optional parent CF type that also derives CppBase.
- *
- * The Impl's lifetime is managed automatically.
- * It can be accessed by reference via .Impl().
+ * CppBase<NewType[, Parent]>
+ * NewType: The name of the CF type deriving CppBase.
+ * Parent: Optional parent CF type that also derives CppBase.
  *
  * Use:
- * struct CFXImpl {
- *     // implementation details for CFX go in here.
+ *
+ * struct CFX: CoreFoundation::CppBase<CFX> {
  *     int a;
  * };
- *
- * struct CFX: CoreFoundation::CppBase<CFX, CFXImpl> { };
- * // Do not populate CFX with any members. Helper functions are okay.
  *
  * CF_EXPORT CFXGetTypeID() {
  *     return CFX::GetTypeID();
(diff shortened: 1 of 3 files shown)


tests/unittests/CoreFoundation/CFCppBaseTests.cpp
```diff
@@ -0,0 +1,101 @@
...
+        ++ctorCountReference;
+    }
+
+    ~__CFLifetimeCounter() {
+        ++dtorCountReference;
+    }
+};
+
+struct __CFSubLifetimeCounter : CoreFoundation::CppBase<__CFSubLifetimeCounter, __CFLifetimeCounter> {
+    __CFSubLifetimeCounter(int& ctorCounter, int& dtorCounter) : Parent(ctorCounter, dtorCounter) {
```
bbowman: Seems a little silly to be using int references here

bbowman: I guess it makes sense for the dtor count.

Frameworks/include/CFCppBase.h
```diff
@@ -118,15 +103,33 @@ struct CppBase : Details::CppBaseImpl<_Parent> {
...
+        // new (ret) T() is undefined behaviour, since the constructor chain does not change them from their otherwise
+        // "uninitialized" values.
+        //
+        // To work around that, we save the __CFRuntimeBase at the head of ret, call the constructor, and replace/reinitialize
+        // it. As no constructor has been invoked on ret by the time we save temp, this is not ill-defined.
+        __CFRuntimeBase temp = *reinterpret_cast<__CFRuntimeBase*>(ret);
+
+        new (ret) T(std::forward<Args>(args)...);
+
+        // This, however, may be ill-defined.
```
bbowman: can this be a static_cast? Every T here should work out for that and we should maybe static_assert or enable_if with is_base_of to validate that CreateInstance only gets called on __CFRuntimeBase things. (Would be super nifty to know it was the root base (not a multi inheritance weirdness) which is maybe doable with template magic?)

jaredhms:
:shipit:

msft-Jeyaram:
:shipit:

ms-jihua:
:shipit:

tests/unittests/CoreFoundation/CFCppBaseTests.cpp
DHowett-MSFT: This tests a few things: Members expressible only through C++ semantics work properly, constructors and destructors get called properly, and inherited members do not lose their magic.

Frameworks/include/CFCppBase.h
DHowett-MSFT: You shouldn't be able to call `T::CreateInstance` with anything but `T`; and I'd like to use reinterpret to circumvent any chicanery the compiler may pull on layout adjustment. I don't expect it to do any, but the block of memory we get back from `_CFRuntimeCreateInstance` is definitely normally a `__CFRuntimeBase` first and foremost.

Frameworks/include/CFCppBase.h
```diff
@@ -118,15 +103,33 @@ struct CppBase : Details::CppBaseImpl<_Parent> {
...
-    }
 
-    TImpl _impl;
+        // Technically, accessing any members of __CFRuntimeBase (which are set up by _CFRuntimeCreateInstance) after
+        // new (ret) T() is undefined behaviour, since the constructor chain does not change them from their otherwise
+        // "uninitialized" values.
+        //
+        // To work around that, we save the __CFRuntimeBase at the head of ret, call the constructor, and replace/reinitialize
+        // it. As no constructor has been invoked on ret by the time we save temp, this is not ill-defined.
+        __CFRuntimeBase temp = *reinterpret_cast<__CFRuntimeBase*>(ret);
```
DHowett-MSFT: cfisa, cfinfo.

PR #1315, opened 2016-11-03
Ensure CppBase::CreateInstance consumers are __CFRuntimeBase-derived.

DHowett-MSFT:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1315)

### DIFF
--- a/Frameworks/include/CFCppBase.h
+++ b/Frameworks/include/CFCppBase.h
@@ -107,6 +107,8 @@ public:
     static T* CreateInstance(CFAllocatorRef allocator, Args&&... args) {
         static_assert(!std::is_polymorphic<T>::value,
                       "CoreFoundation::CppBase cannot be used with polymorphic types or classes bearing virtual functions.");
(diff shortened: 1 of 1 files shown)


DHowett-MSFT:
Tested manually (since we cannot add compileable tests for it).

```
struct LOL {
};

struct CFEatMyShorts : CoreFoundation::CppBase<CFEatMyShorts, LOL> {
};

TEST(Shorts, EatingWorks) {
    CFEatMyShorts::CreateInstance();
} 

1>E:\src\WinObjC\Frameworks\include\CFCppBase.h(110,9): error : static_assert failed "CoreFoundation::CppBase cannot be used with types that do not derive from __CFRuntimeBase."
1>         E:\src\WinObjC\Frameworks\include\CFCppBase.h(134,16) :  note: in instantiation of function template specialization 'CoreFoundation::CppBase<CFEatMyShorts, LOL>::CreateInstance<>' requested here 
```

PR #1318, opened 2016-11-04
Composition and layout refactor

jaredhms:
This change refactors WinObjC's QuartzCore and UIKit stacks to align more closely with Xaml, as detailed in the docs/CoreAnimation/WinObjc.Composition.and.Layout.docx dev spec.

Some highlights include:

    1.	Streamlines the usage of arbitrary Xaml FrameworkElements by allowing them to be positioned just as any other UIView in the application.
    2.	Supports rendering over Xaml-backed UIKit controls; adding adornments, etc.
    3.	Supports adding subviews and sublayers to Xaml-backed UIKit controls.
    4.	Separates the complex details of our Core Animation composition layer into distinct and manageable subcomponents  improving maintainability.
    5.	Defines a concrete relationship between every UIView, its root CALayer, and its backing Xaml FrameworkElement.
    6.	Streamlines the UIElement tree by removing unnecessary UIElements (for example; this results in a ~30% reduction of UIElements on the initial scene of WOCCatalog).
    7.	Simplifies app rendering by removing our two custom Panel implementations; thus eliminating the need to take part in Arrange and Measure passes.
    8.	Paves the way for eventual Xaml-markup-driven UIElement trees in ported WinObjC apps, because we no longer require re-parenting Xaml FrameworkElements that are positioned by Core Animation.
    9.	Moves the rest of the Core Animation composition layer over to C++, to benefit from modern programming constructs such as RAII, external ref-counting (shared_ptr), etc.

In order to implement these changes, we had to move UILabel over to a fully Xaml-backed implementation (rather than following a more complicated path with fake 'textures' down into the compositor).  As a result, there are some known issues around text layout in one of our internal test apps (MM); we'll address those issues (and anything else that comes up) in a stabilization pass after these changes go in.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1318)


jaredhms:
@msft-Jeyaram is added to the review. #Closed

jaredhms:
@ashvarma is added to the review. #Closed

jaredhms:
@bviglietta is added to the review. #Closed

bbowman:
> //

doesn't seem changed?

---
Refers to: Frameworks/include/MockClass.h:4 in ff95a9a. [](commit_id = ff95a9abc5e318dfa74e9e9dece2664550ace0a3, deletion_comment = False)

jaredhms:
> //

just some clang-format whitespace changes....

---
In reply to: [258527516](https://github.com/Microsoft/WinObjC/pull/1318#issuecomment-258527516) [](ancestors = 258527516)

---
Refers to: Frameworks/include/MockClass.h:4 in ff95a9a. [](commit_id = ff95a9abc5e318dfa74e9e9dece2664550ace0a3, deletion_comment = False)

Frameworks/UIKit/StarboardXaml/LayerTransaction.mm
```diff
@@ -0,0 +1,243 @@
...
+#import "Starboard.h"
+#import "LayerTransaction.h"
+
+#import "LayerAnimation.h"
+#import "LayerProxy.h"
+
+#import "CAAnimationInternal.h"
+#import "CALayerInternal.h"
+
+static const wchar_t* TAG = L"LayerTransaction";
```
jaredhms: >LayerTransaction [](start = 30, length = 16)

A lot of this came from CompositorInterface.mm.

Frameworks/UIKit/StarboardXaml/LayerProxy.mm
```diff
@@ -0,0 +1,179 @@
...
+
+#import "QuartzCore\CATransform3D.h"
+#import "Quaternion.h"
+#import "UIColorInternal.h"
+
+static const wchar_t* TAG = L"LayerProxy";
+
+void LayerProxy::SetTexture(const std::shared_ptr<IDisplayTexture>& texture, float width, float height, float contentScale) {
+    _currentTexture = texture;
+    _SetContents((texture ? texture->GetContent() : nullptr), width, height, contentScale);
```
jaredhms: A lot of this came from CompositorInterface.mm.

Frameworks/UIKit/StarboardXaml/LayerAnimation.mm
```diff
@@ -0,0 +1,581 @@
...
+public:
+    id _animHandler;
+    NSString *_type, *_subType;
+
+    TransitionAnimation(id animHandler, NSString* type, NSString* subType) {
+        _animHandler = [animHandler retain];
+
+        _type = [type retain];
+        _subType = [subType retain];
+        beginTime = 0;
```
jaredhms: A lot of this came from CompositorInterface.mm; sorry it's not easy to diff.

Frameworks/UIKit/StarboardXaml/DisplayTexture.mm
```diff
@@ -0,0 +1,213 @@
...
+#import "CoreGraphics/CGImage.h"
+#import "CGContextInternal.h"
+#import "UIColorInternal.h"
+
+#include <map>
+#include <mutex>
+
+using namespace Microsoft::WRL;
+
+static const wchar_t* TAG = L"DisplayTexture";
```
jaredhms: >DisplayTexture [](start = 30, length = 14)

A lot of this came from CompositorInterface.mm; sorry it's not easy to diff.

Frameworks/UIKit/StarboardXaml/DisplayProperties.mm
```diff
@@ -0,0 +1,163 @@
...
+//******************************************************************************
+#import "Starboard.h"
+#import "DisplayProperties.h"
+#import "XamlCompositor.h"
+
+#import "UIApplicationInternal.h"
+
+#import <algorithm>
+
+static const wchar_t* TAG = L"DisplayProperties";
```
jaredhms: >DisplayProperties [](start = 30, length = 17)

A lot of this came from CompositorInterface.mm; sorry it's not easy to diff.

include/UIKit/UIView.h
```diff
@@ -381,7 +381,9 @@ UIKIT_EXPORT_CLASS
 // WinObjC Xaml Extensions
```
jaredhms: Dustin; I don't think we currently have an annotation for this.

Frameworks/UIKit/StarboardXaml/DisplayTexture.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+using namespace Microsoft::WRL;
+using namespace UIKit::Private::CoreAnimation;
+using namespace Windows::Foundation;
+using namespace Windows::Storage::Streams;
+using namespace Windows::UI::Xaml;
+using namespace Windows::UI::Xaml::Controls;
+using namespace Windows::UI::Xaml::Media;
+
+static const wchar_t* TAG = L"DisplayTexture";
```
jaredhms: >DisplayTexture [](start = 30, length = 14)

A lot of this came from either CALayerXaml.cpp or XamlCompositor.cpp; sorry it's not easy to diff.

Frameworks/UIKit/StarboardXaml/LayerCoordinator.cpp
```diff
@@ -0,0 +1,1335 @@
...
+using namespace UIKit::Private::CoreAnimation;
+using namespace Windows::Foundation;
+using namespace Windows::Storage::Streams;
+using namespace Windows::UI::Xaml;
+using namespace Windows::UI::Xaml::Controls;
+using namespace Windows::UI::Xaml::Media;
+using namespace Windows::UI::Xaml::Media::Animation;
+using namespace Windows::UI::Xaml::Media::Imaging;
+
+static const wchar_t* TAG = L"LayerCoordinator";
```
jaredhms: >LayerCoordinator [](start = 30, length = 16)

A lot of this came from either CALayerXaml.cpp or XamlCompositor.cpp; sorry it's not easy to diff.

Frameworks/UIKit/StarboardXaml/LayerProxy.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+using namespace Microsoft::WRL;
+using namespace UIKit::Private::CoreAnimation;
+using namespace Windows::Foundation;
+using namespace Windows::Storage::Streams;
+using namespace Windows::UI::Xaml;
+using namespace Windows::UI::Xaml::Controls;
+using namespace Windows::UI::Xaml::Media;
+
+static const wchar_t* TAG = L"LayerProxy";
```
jaredhms: >LayerProxy [](start = 30, length = 10)

A lot of this came from either CALayerXaml.cpp or XamlCompositor.cpp; sorry it's not easy to diff.

Frameworks/UIKit/StarboardXaml/StoryboardManager.cpp
```diff
@@ -0,0 +1,286 @@
...
+using namespace UIKit::Private::CoreAnimation;
+using namespace Windows::Foundation;
+using namespace Windows::Storage::Streams;
+using namespace Windows::UI::Xaml;
+using namespace Windows::UI::Xaml::Controls;
+using namespace Windows::UI::Xaml::Media;
+using namespace Windows::UI::Xaml::Media::Animation;
+using namespace Windows::UI::Xaml::Media::Imaging;
+
+static const wchar_t* TAG = L"StoryboardManager";
```
jaredhms: >StoryboardManager [](start = 30, length = 17)

A lot of this came from either CALayerXaml.cpp or XamlCompositor.cpp; sorry it's not easy to diff.

Frameworks/UIKit/StarboardXaml/XamlCompositor.mm
```diff
@@ -0,0 +1,134 @@
...
+#import "LayerAnimation.h"
+#import "LayerProxy.h"
+#import "LayerTransaction.h"
+
+#import "CGContextInternal.h"
+#import "UIApplicationInternal.h"
+
+using namespace Microsoft::WRL;
+
+static const wchar_t* TAG = L"XamlCompositor";
```
jaredhms: >XamlCompositor [](start = 30, length = 14)

This is the remains of CompositorInterface.mm.

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -325,60 +287,45 @@ static void DiscardLayerContents(CALayer* layer) {
...
 */
 - (instancetype)init {
-    assert(priv == NULL);
-    priv = new CAPrivateInfo(self);
+    return [self _initWithXamlElement:nil];
+}
 
+/**
+ Microsoft Extension
+ All CALayers are ultimately backed by a Xaml FrameworkElement.  Passing nil here will
```
jaredhms: that's a good idea; i'll follow up on it.

---
In reply to: [86620194](https://github.com/Microsoft/WinObjC/pull/1318#discussion_r86620194) [](ancestors = 86620194)

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -1652,20 +1468,23 @@ - (void)setAffineTransform:(CGAffineTransform)transform {
...
+    [self setTransform:newTransform];
+}
 
-    memcpy(&priv->transform, &newTransform, sizeof(CATransform3D));
-    priv->_frameIsCached = FALSE;
+/**
+ @Status Interoperable
+*/
+- (CGAffineTransform)affineTransform {
+    CGAffineTransform ret;
```
DHowett-MSFT: >eTransform re [](start = 11, length = 13)

Instead of poking at these directly, you can use `CGAffineTransformMake(a,b,c,d,tx,ty)`

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -528,9 +453,11 @@ - (void)display {
         int height = (int)(ceilf(priv->bounds.size.height) * priv->contentsScale);
 
         if (width <= 0 || height <= 0) {
+            TraceVerbose(TAG, L"Not drawing due to invalid layer dimensions; width=%d, height=%d", width, height);
```
yiyang-msft: >TraceVerbose [](start = 12, length = 12)

if verbose might put in if(verbose) otherwise traceWarning?

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -2495,96 +2290,54 @@ - (void)_setRootLayer:(BOOL)isRootLayer {
...
+    }
 
-        // transform the points in fromLayer to point in toLayer
-        WFPoint* pointInFromLayer = [WXPointHelper fromCoordinates:point.x y:point.y];
-        WFPoint* pointInToLayer = [transform transformPoint:pointInFromLayer];
-        ret = { pointInToLayer.x, pointInToLayer.y };
+    if (!toLayer) {
+        // get the top most layer for the current layer hierarchy
+        toLayer = _findSuperLayerForLayer(fromLayer);
     }
```
yiyang-msft: nit: may consider check if fromLayer == toLayer, if so, return the point directly?

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -2495,96 +2290,54 @@ - (void)_setRootLayer:(BOOL)isRootLayer {
...
 + (CGPoint)convertPoint:(CGPoint)point fromLayer:(CALayer*)fromLayer toLayer:(CALayer*)toLayer {
-    CGPoint ret = { point.x, point.y };
-
-    if (fromLayer && toLayer) {
-        // get the backing xaml UIElement for fromLayer
-        WXUIElement* fromLayerElement = _getBackingXamlElementForCALayer(fromLayer);
+    if (!fromLayer && !toLayer) {
+        if (DEBUG_VERBOSE) {
+            TraceVerbose(TAG, L"convertPoint cannot convert points for two nil layers.");
+        }
```
yiyang-msft: Thanks!

Frameworks/UIKit.Xaml/Layer.xaml.cpp
```diff
@@ -0,0 +1,122 @@
...
+} /* CoreAnimation */
+} /* Private */
+} /* UIKit */
+
+////////////////////////////////////////////////////////////////////////////////////
+// ObjectiveC Interop
+////////////////////////////////////////////////////////////////////////////////////
+
+// Set one or more layer properties for the specified target xaml element
+UIKIT_XAML_EXPORT void XamlSetFrameworkElementLayerProperties(
```
yiyang-msft: >XamlSetFrameworkElementLayerProperties [](start = 23, length = 38)

it is bit inflexible if we need setting new layer properties.  might consider passing a dictionary contains key/value as the second param?

jaredhms: These are the only two properties we support; they map to the attached properties registered above in _RegisterDependencyProperties.  The only other property that we may need to add is Border (if we don't port UIScrollView to Xaml before we implement Border on CALayer).

---
In reply to: [86627070](https://github.com/Microsoft/WinObjC/pull/1318#discussion_r86627070) [](ancestors = 86627070)

Frameworks/include/CACompositor.h
```diff
@@ -13,17 +13,102 @@
...
+};
+
+// Proxy for CG-rendered content
+struct IDisplayTexture {
+public:
+    virtual ~IDisplayTexture() {
+    }
+
+    virtual Microsoft::WRL::ComPtr<IInspectable> GetContent() = 0;
+    virtual void* Lock(int* stride) = 0;
```
rajsesh-msft: Nit: would be helpful to document what the void* is for an innocent future engineer.

Frameworks/include/CACompositor.h
```diff
@@ -13,17 +13,102 @@
...
 #include <memory>
-#include "winobjc\winobjc.h"
+#include <wrl/client.h>
+
+#endif
+
+// Proxy between CALayer and its Xaml representation
+struct ILayerProxy {
+public:
+    virtual ~ILayerProxy() {
```
rajsesh-msft: nit: these virtual destructors could be pure virtual instead.

Frameworks/include/CACompositor.h
```diff
@@ -13,17 +13,102 @@
...
+    }
+
+    // Returns the Xaml element backing this layer
+    virtual Microsoft::WRL::ComPtr<IInspectable> GetXamlElement() = 0;
+
+    // Returns the Xaml element used for hosting this layer's sublayer element
+    virtual Microsoft::WRL::ComPtr<IInspectable> GetSublayerXamlElement() = 0;
+
+    virtual void* GetPropertyValue(const char* propertyName) = 0;
```
rajsesh-msft: nit: some of these methods could be const.

include/UIKit/UIView.h
```diff
@@ -381,7 +381,9 @@ UIKIT_EXPORT_CLASS
...
 @class WXFrameworkElement;
```
ashvarma: >WXFrameworkElement [](start = 7, length = 18)

Nit: All other forward declarations are done in a single pace in this file, except this one.

ashvarma:
> 

All this should stay with the UIKit::Button content... so the declarations for Label and Layer that you've added should move below these lines. 

---
Refers to: Frameworks/UIKit.Xaml/ObjCXamlControls.h:75 in ff95a9a. [](commit_id = ff95a9abc5e318dfa74e9e9dece2664550ace0a3, deletion_comment = False)

ashvarma:
> #define CACompositorRotation90CounterClockwise 270.0f

All these seem to be used only by UIApplication.mm - makes me wonder if they should just be defined in that file. Not sure if they need to be defined as part of "CACompositor" 

---
Refers to: Frameworks/include/CACompositor.h:117 in ff95a9a. [](commit_id = ff95a9abc5e318dfa74e9e9dece2664550ace0a3, deletion_comment = False)

Frameworks/UIKit/XamlUtilities.mm
```diff
@@ -13,17 +13,16 @@
...
+#import "StarboardXaml/DisplayTexture.h"
+
+#include <Windows.UI.Xaml.Media.h>
 
 using namespace Microsoft::WRL;
 using namespace Windows::Foundation;
 
 // cornerRadius when border style is set to round rectangle
 static const int c_borderCornerRadius = 8;
-static const wchar_t* TAG = L"XamlUtilities";
```
ashvarma: unclear to me - why did you remove this?

Frameworks/UIKit/UITextField.mm
```diff
@@ -1248,21 +1267,28 @@ - (void)_initPasswordBox:(WXFrameworkElement*)xamlElement {
...
+    [_subView setAutoresizingMask:UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight];
+    [self addSubview:_subView];
 }
 
 // Main entrance to initialize TextField
-- (void)_initTextField:(WXFrameworkElement*)xamlElement {
+- (void)_initUITextField:(WXFrameworkElement*)xamlElement {
     self->_secureModeLock = [NSRecursiveLock new];
 
     // creating dummy button and hidden view so that it can be used to steal/kill the focus for this UITextField
```
ashvarma: >so that it can be used to steal/kill the focus for this UITextField [](start = 45, length = 67)

unrelated to this change - this comment seems pretty "light" on details... and its not intuitive why we need to create an invisible button here... will follow up offline

build/UIKit.Xaml/dll/UIKit.Xaml.vcxproj
```diff
@@ -161,6 +165,12 @@
     <ClInclude Include="$(MSBuildThisFileDirectory)..\..\..\Frameworks\UIKit.Xaml\Button.xaml.h">
       <DependentUpon>Button.xaml</DependentUpon>
     </ClInclude>
+    <ClInclude Include="$(MSBuildThisFileDirectory)..\..\..\Frameworks\UIKit.Xaml\Label.xaml.h">
+      <DependentUpon>Label.xaml</DependentUpon>
+    </ClInclude>
```
oliversa-msft: >     [](start = 16, length = 4)

Nit - extra spaces

oliversa-msft:
>         [_mediaElement addCurrentStateChangedEvent:^(RTObject* sender, WXRoutedEventArgs* e) {

Nit - Couldn't find the remove/unregister of these events in the dealloc. Might be something we'll need to track with a bug.

---
Refers to: Frameworks/AVFoundation/AVAudioPlayer.mm:78 in ff95a9a. [](commit_id = ff95a9abc5e318dfa74e9e9dece2664550ace0a3, deletion_comment = False)

Frameworks/include/CAAnimationInternal.h
```diff
@@ -29,22 +30,21 @@ struct CAMediaTimingProperties {
...
 - (void)setDelay:(double)delay;
-- (void)_abortAnimation;
 - (void)_removeAnimationsFromLayer;
 - (void)_setFinishedSelector:(SEL)selector;
 @end
 
 @interface NSObject (CAAnimationInternal)
-- (void)_removeAnimationsFromLayer;
 - (BOOL)wasRemoved;
 - (BOOL)wasAborted;
```
oliversa-msft: >(BOOL)wasAborted; [](start = 2, length = 17)

Do we still need this "internal" wasAborted property if we removed the internal _abortAnimation method?

Frameworks/include/CACompositor.h
```diff
@@ -148,4 +160,5 @@ CA_EXPORT CACompositorInterface* GetCACompositor();
 CA_EXPORT void SetCACompositor(CACompositorInterface* compositorInterface);
 
 CA_EXPORT bool CASignalDisplayLink();
```
oliversa-msft: Could we move these global vars/defines to the top of the file? It's really a pain tracking them down when they are scattered within the body of a file.

Frameworks/include/CALayerInternal.h
```diff
@@ -29,28 +29,24 @@ class CADisplayProperties {
...
     float borderWidth, cornerRadius;
     float opacity;
 };
 
-class DisplayNode;
-class DisplayTexture;
+struct ILayerProxy;
+struct IDisplayTexture;
+
 @class CALayer;
```
oliversa-msft: Move these forward declarations with the others on line 25?

Frameworks/UIKit/StarboardXaml/DisplayTexture.h
```diff
@@ -0,0 +1,47 @@
+//******************************************************************************
+//
+// Copyright (c) 2015 Microsoft Corporation. All rights reserved.
```
msft-Jeyaram: >opyright (c) 2015 Microsoft Corporation. All rights reserved. [](start = 4, length = 61)

header updates

Frameworks/include/CALayerInternal.h
```diff
@@ -29,28 +29,24 @@ class CADisplayProperties {
     BOOL hidden;
     CGRect bounds;
     CGRect contentsRect, contentsCenter;
-    CGRect contentsInset; //  Texture insets
     uint32_t contentsOrientation;
```
oliversa-msft: >uint32_t contentsOrientation; [](start = 4, length = 29)

Just for full disclosure - contentsOrientation is not part of the CALayer API reference. It's only here because it's used by UIImage - worth a bug to investigate how to remove it cleanly

Frameworks/UIKit/StarboardXaml/DisplayTexture.mm
```diff
@@ -0,0 +1,213 @@
...
+                    fileData.attach([[NSData alloc] initWithContentsOfFile:[NSString stringWithUTF8String:jpgImg->_fileName]]);
+                    data = [fileData bytes];
+                    len = [fileData length];
+                } else {
+                    data = [(NSData*)jpgImg->_data bytes];
+                    len = [(NSData*)jpgImg->_data length];
+                }
+            } break;
+            default:
+                TraceError(TAG, L"Warning: unrecognized image format sent to DisplayTextureContent!");
```
msft-Jeyaram: >DisplayTextureContent [](start = 77, length = 21)

nit: DisplayTexture

msft-Jeyaram: Never mind, it's getting removed in my change over to use WIC/D2D

---
In reply to: [86869562](https://github.com/Microsoft/WinObjC/pull/1318#discussion_r86869562) [](ancestors = 86869562)

oliversa-msft:
>     CGColorRef _backgroundColor, _borderColor;

I'd really love to know why in the PrivateInfo class we have these underscore properties (e.g. _backgroundColor, _borderColor, _name, _xamlElement) and some properties that are not - just confusing for a new person to the codebase. Shouldn't they all be underscored or not underscored given the internal nature of CAPrivateInfo?

---
Refers to: Frameworks/include/CALayerInternal.h:65 in ff95a9a. [](commit_id = ff95a9abc5e318dfa74e9e9dece2664550ace0a3, deletion_comment = False)

oliversa-msft:
> - (void)exchangeSublayer:(CALayer*)layer1 withLayer:(CALayer*)layer2;

This should be an underscored method - it's not listed in the API reference I could find.

---
Refers to: Frameworks/include/CALayerInternal.h:111 in ff95a9a. [](commit_id = ff95a9abc5e318dfa74e9e9dece2664550ace0a3, deletion_comment = False)

oliversa-msft:
> - (void)bringSublayerToFront:(CALayer*)sublayer;

These two methods are also not listed - move them to internal i.e. _sendSublayerToBack

---
Refers to: Frameworks/include/CALayerInternal.h:114 in ff95a9a. [](commit_id = ff95a9abc5e318dfa74e9e9dece2664550ace0a3, deletion_comment = False)

oliversa-msft:
> - (void)exchangeSubviewAtIndex:(int)index1 withSubviewAtIndex:(int)index2;

This is not needed in CALayer - UIView declares its own interoperable method with the same name

---
Refers to: Frameworks/include/CALayerInternal.h:112 in ff95a9a. [](commit_id = ff95a9abc5e318dfa74e9e9dece2664550ace0a3, deletion_comment = False)

Frameworks/include/CALayerInternal.h
```diff
@@ -115,20 +116,16 @@ class CAPrivateInfo : public CADisplayProperties, public LLTreeNode<CAPrivateInf
...
++ (CGPoint)convertPoint:(CGPoint)point fromLayer:(CALayer*)layer toLayer:(CALayer*)layer;
 
 - (void)_removeAnimation:(CAAnimation*)animation;
-- (DisplayTexture*)_getDisplayTexture;
+- (std::shared_ptr<IDisplayTexture>)_getDisplayTexture;
 
 - (CAPrivateInfo*)_priv;
 - (void)_setRootLayer:(BOOL)isRootLayer;
 
 - (void)_setZIndex:(int)zIndex;
```
oliversa-msft: Another property only placed here because it is used by UIWindow - file a bug so we can clean it up and integrate it more cleanly. We removed all other z-methods (zPosition, anchorPointZ, etc.) from CALayer

oliversa-msft:
> #define _CALAYERPRIVATE_H_

#pragma once

---
Refers to: Frameworks/include/CALayerInternal.h:18 in ff95a9a. [](commit_id = ff95a9abc5e318dfa74e9e9dece2664550ace0a3, deletion_comment = False)

Frameworks/include/CALayerInternal.h
```diff
@@ -115,20 +116,16 @@ class CAPrivateInfo : public CADisplayProperties, public LLTreeNode<CAPrivateInf
...
 
-- (void)_setOrigin:(CGPoint)origin updateContent:(BOOL)updateContent;
-
-@property WXFrameworkElement* contentsElement;
-
 @end
 
 #endif /* _CALAYERPRIVATE_H_ */
```
oliversa-msft: >/* _CALAYERPRIVATE_H_ */ [](start = 7, length = 24)

If you replace with #pragma once, please remove

oliversa-msft:
> CACompositorInterface* _globalCompositor;

Could have defined this in the Compositor itself along with its getter/setter. It's externed there

---
Refers to: Frameworks/QuartzCore/CALayer.mm:101 in ff95a9a. [](commit_id = ff95a9abc5e318dfa74e9e9dece2664550ace0a3, deletion_comment = False)

oliversa-msft:
>     superlayer = 0;

superlayer = nil

---
Refers to: Frameworks/QuartzCore/CALayer.mm:244 in ff95a9a. [](commit_id = ff95a9abc5e318dfa74e9e9dece2664550ace0a3, deletion_comment = False)

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -285,29 +252,24 @@ static void DiscardLayerContents(CALayer* layer) {
     delegate = 0;
```
oliversa-msft: >0 [](start = 15, length = 1)

delegate = nil

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -548,75 +475,50 @@ - (void)display {
...
-        CGImageRef vectorTarget = NULL;
-
-        if ((priv->isOpaque && priv->_backgroundColor == nil) || priv->backgroundColor.a == 1.0) {
-            priv->drewOpaque = TRUE;
-        } else {
-            priv->drewOpaque = FALSE;
-        }
+        // Create the contents
+        CGContextRef drawContext;
+        drawContext = CreateLayerContentsBitmapContext32(width, height);
```
oliversa-msft: Could collapse into a single line

Frameworks/include/CACompositor.h
msft-Jeyaram: >int [](start = 23, length = 3)

size_t?

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -710,26 +585,12 @@ static void doRecursiveAction(CALayer* layer, NSString* actionName) {
...
-    return FALSE;
-}
-
 /**
  @Status Interoperable
 */
 - (void)addSublayer:(CALayer*)subLayerAddr {
     if (subLayerAddr == self) {
-        assert(0);
+        FAIL_FAST();
```
oliversa-msft: >FAIL_FAST(); [](start = 8, length = 12)

We should really use FAIL_FAST_MSG or accompany this state with a Trace* message.

oliversa-msft:
>         assert(0);

We should probably get rid of this assert as well.

---
Refers to: Frameworks/QuartzCore/CALayer.mm:1143 in ff95a9a. [](commit_id = ff95a9abc5e318dfa74e9e9dece2664550ace0a3, deletion_comment = False)

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -1508,21 +1312,33 @@ - (void)setOpaque:(BOOL)isOpaque {
...
+
+/**
+ @Status Stub
+*/
+- (void)setAnchorPointZ:(float)pos {
+    UNIMPLEMENTED();
+}
+
+/**
+ @Status Stub
```
oliversa-msft: >@Status Stub [](start = 1, length = 12)

Nice!

oliversa-msft:
>     newTransform = CATransform3DMakeTranslation(0, 0, 0);

Same comment as Dustins - it's interesting that we didn't use CGAffineTransformMake here

---
Refers to: Frameworks/QuartzCore/CALayer.mm:1463 in ff95a9a. [](commit_id = ff95a9abc5e318dfa74e9e9dece2664550ace0a3, deletion_comment = False)

oliversa-msft:
>     newTransform = CATransform3DMakeTranslation(0, 0, 0);

Actually, it makes sense now - it's using the 3D version below - setTransform() method

---
In reply to: [258990934](https://github.com/Microsoft/WinObjC/pull/1318#issuecomment-258990934) [](ancestors = 258990934)

---
Refers to: Frameworks/QuartzCore/CALayer.mm:1463 in ff95a9a. [](commit_id = ff95a9abc5e318dfa74e9e9dece2664550ace0a3, deletion_comment = False)

oliversa-msft:
>         }

I'm not sure about this code change - I didn't come across it in the Apple's API reference docs if we dealing with a standalone CALayer. Interesting. I guess we could also move it as the first conditional if it is indeed required.

---
Refers to: Frameworks/QuartzCore/CALayer.mm:1803 in ff95a9a. [](commit_id = ff95a9abc5e318dfa74e9e9dece2664550ace0a3, deletion_comment = False)

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -2006,12 +1805,13 @@ - (NSArray*)sublayers {
         if (shouldAnimate) {
             ret = [CATransaction _implicitAnimationForKey:key];
             if (ret != nil) {
-                NSObject* value = GetCACompositor()->getDisplayProperty(priv->_presentationNode, [key UTF8String]);
+                NSObject* value = reinterpret_cast<NSObject*>(priv->_layerProxy->GetPropertyValue([key UTF8String]));
```
oliversa-msft: >priv->_layerProxy->GetPropertyValue [](start = 62, length = 35)

Sweet!

jaredhms:
>         }

what code change are you referring to?  I just deleted a bunch of dead code around setting the 'shouldAnimate' flag.

---
In reply to: [258998954](https://github.com/Microsoft/WinObjC/pull/1318#issuecomment-258998954) [](ancestors = 258998954)

---
Refers to: Frameworks/QuartzCore/CALayer.mm:1803 in ff95a9a. [](commit_id = ff95a9abc5e318dfa74e9e9dece2664550ace0a3, deletion_comment = False)

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -2495,96 +2290,54 @@ - (void)_setRootLayer:(BOOL)isRootLayer {
...
-        WFPoint* pointInToLayer = [transform transformPoint:pointInFromLayer];
-        ret = { pointInToLayer.x, pointInToLayer.y };
+    if (!toLayer) {
+        // get the top most layer for the current layer hierarchy
+        toLayer = _findSuperLayerForLayer(fromLayer);
     }
 
+    // Get the backing sublayer xaml UIElement for fromLayer
+    // We use the sublayer element to support proper point conversion within a scrolled view.
+    WXUIElement* fromLayerElement = [fromLayer _getSublayerXamlElement];
```
oliversa-msft: >WXUIElement* [](start = 4, length = 12)

Could we make this work with WXFrameworkElement instead? Just wondering if we need to keep it consistent for readability or is there something else that is used in the parent class.

jaredhms: transformToVisual is on UIElement.

---
In reply to: [86898072](https://github.com/Microsoft/WinObjC/pull/1318#discussion_r86898072) [](ancestors = 86898072)

yiyang-msft:
:shipit:

oliversa-msft:
> __declspec(thread) CATransaction* _curTransaction, *_rootTransaction;

Really should be prefixed with g_

---
Refers to: Frameworks/QuartzCore/CATransaction.mm:35 in ff95a9a. [](commit_id = ff95a9abc5e318dfa74e9e9dece2664550ace0a3, deletion_comment = False)

Frameworks/QuartzCore/CATransaction.mm
```diff
@@ -36,14 +36,15 @@
```
DHowett-MSFT: Nits about these guys:

* `__declspec(thread)` objective-C things will not be properly ref-counted.
* they're being compared against `NULL` (not even `nullptr`!) instead of `nil` everywhere in this file.

Frameworks/QuartzCore/CATransaction.mm
```diff
@@ -192,72 +196,52 @@ + (CAAnimation*)_implicitAnimationForKey:(NSString*)forKey {
...
+    [self _currentLayerTransaction]->RemoveLayer([sublayer _layerProxy]);
 }
 
 + (void)_moveLayer:(CALayer*)layer beforeLayer:(CALayer*)before afterLayer:(CALayer*)after {
-    GetCACompositor()->moveNode([self _currentTransaction]->_transactionQueue,
-                                [layer _priv]->_presentationNode,
-                                before ? [before _priv]->_presentationNode : NULL,
-                                after ? [after _priv]->_presentationNode : NULL);
+    [self _currentLayerTransaction]->MoveLayer([layer _layerProxy],
+                                               before ? [before _layerProxy] : nullptr,
```
DHowett-MSFT: yeah: if before is nil, `[nil _layerProxy]` will return the right thing. Unless it's a C++ type, but really the ideal here is to do the right thing :P

---
In reply to: [86621323](https://github.com/Microsoft/WinObjC/pull/1318#discussion_r86621323) [](ancestors = 86621323)

jaredhms:
> 

~bad merge;  good catch

---
In reply to: [258638949](https://github.com/Microsoft/WinObjC/pull/1318#issuecomment-258638949) [](ancestors = 258638949)

---
Refers to: Frameworks/UIKit.Xaml/ObjCXamlControls.h:75 in ff95a9a. [](commit_id = ff95a9abc5e318dfa74e9e9dece2664550ace0a3, deletion_comment = False)

jaredhms:
> #define CACompositorRotation90CounterClockwise 270.0f

good point - these should have moved along with the rest of the DisplayProperties -- into DisplayProperties.h -- and switched to an enum.

---
In reply to: [258740082](https://github.com/Microsoft/WinObjC/pull/1318#issuecomment-258740082) [](ancestors = 258740082)

---
Refers to: Frameworks/include/CACompositor.h:117 in ff95a9a. [](commit_id = ff95a9abc5e318dfa74e9e9dece2664550ace0a3, deletion_comment = False)

jaredhms:
>     CGColorRef _backgroundColor, _borderColor;

there's no reason.  they should all be underscored, but i'm not going to do that in this change - I don't want to unnecessarily add to the churn.

---
In reply to: [258966276](https://github.com/Microsoft/WinObjC/pull/1318#issuecomment-258966276) [](ancestors = 258966276)

---
Refers to: Frameworks/include/CALayerInternal.h:65 in ff95a9a. [](commit_id = ff95a9abc5e318dfa74e9e9dece2664550ace0a3, deletion_comment = False)

jaredhms:
>     CGColorRef _backgroundColor, _borderColor;

ideally they would all be private, too....so it's easier to see which state is accessed externally, but I'm not fixing that now, either.

---
In reply to: [259010707](https://github.com/Microsoft/WinObjC/pull/1318#issuecomment-259010707) [](ancestors = 259010707,258966276)

---
Refers to: Frameworks/include/CALayerInternal.h:65 in ff95a9a. [](commit_id = ff95a9abc5e318dfa74e9e9dece2664550ace0a3, deletion_comment = False)

jaredhms:
> - (void)exchangeSublayer:(CALayer*)layer1 withLayer:(CALayer*)layer2;

there are a lot of missing underscores, but I'm not going to address them with this change.

---
In reply to: [258967895](https://github.com/Microsoft/WinObjC/pull/1318#issuecomment-258967895) [](ancestors = 258967895)

---
Refers to: Frameworks/include/CALayerInternal.h:111 in ff95a9a. [](commit_id = ff95a9abc5e318dfa74e9e9dece2664550ace0a3, deletion_comment = False)

jaredhms:
> - (void)exchangeSubviewAtIndex:(int)index1 withSubviewAtIndex:(int)index2;

I'm not going to remove any more code with this change.

---
In reply to: [258968776](https://github.com/Microsoft/WinObjC/pull/1318#issuecomment-258968776) [](ancestors = 258968776)

---
Refers to: Frameworks/include/CALayerInternal.h:112 in ff95a9a. [](commit_id = ff95a9abc5e318dfa74e9e9dece2664550ace0a3, deletion_comment = False)

jaredhms:
> - (void)bringSublayerToFront:(CALayer*)sublayer;

see above; i'm not going to fix the missing underscore issue with this change.

---
In reply to: [258968087](https://github.com/Microsoft/WinObjC/pull/1318#issuecomment-258968087) [](ancestors = 258968087)

---
Refers to: Frameworks/include/CALayerInternal.h:114 in ff95a9a. [](commit_id = ff95a9abc5e318dfa74e9e9dece2664550ace0a3, deletion_comment = False)

jaredhms:
> CACompositorInterface* _globalCompositor;

could have, but didn't bother making that change with this refactor.  it may open up additional cans of worms, and isn't necessary at this time.

---
In reply to: [258982393](https://github.com/Microsoft/WinObjC/pull/1318#issuecomment-258982393) [](ancestors = 258982393)

---
Refers to: Frameworks/QuartzCore/CALayer.mm:101 in ff95a9a. [](commit_id = ff95a9abc5e318dfa74e9e9dece2664550ace0a3, deletion_comment = False)

Frameworks/UIKit/XamlUtilities.mm
jaredhms: it's not used, but I can add it back for the next person who needs to add logging to this file.

---
In reply to: [86710622](https://github.com/Microsoft/WinObjC/pull/1318#discussion_r86710622) [](ancestors = 86710622)

Frameworks/include/CACompositor.h
jaredhms: That would require more churn than I think is worth it at this point, and at the end of the day, the value is an int (*stride = writableBitmap->PixelWidth * 4;)

---
In reply to: [86886920](https://github.com/Microsoft/WinObjC/pull/1318#discussion_r86886920) [](ancestors = 86886920)

Frameworks/UIKit/UIWebView.mm
```diff
@@ -320,7 +331,7 @@ - (void)dealloc {
         [_xamlWebControl removeLoadCompletedEvent:_xamlLoadCompletedEventCookie];
         [_xamlWebControl removeNavigationStartingEvent:_xamlLoadStartedEventCookie];
         [_xamlWebControl removeUnsupportedUriSchemeIdentifiedEvent:_xamlUnsupportedUriSchemeEventCookie];
-        [_xamlWebControl release];
+        _xamlWebControl = nil;
```
jaredhms: not require usually, but here we're freeing the element on the UIThread.

---
In reply to: [86633009](https://github.com/Microsoft/WinObjC/pull/1318#discussion_r86633009) [](ancestors = 86633009)

Frameworks/include/CALayerInternal.h
jaredhms: added a comment

---
In reply to: [86869414](https://github.com/Microsoft/WinObjC/pull/1318#discussion_r86869414) [](ancestors = 86869414)

Frameworks/include/CALayerInternal.h
jaredhms: this is already heavily commented in code (see the impl in CALayer.mm)

---
In reply to: [86874169](https://github.com/Microsoft/WinObjC/pull/1318#discussion_r86874169) [](ancestors = 86874169)

Frameworks/QuartzCore/CATransaction.mm
jaredhms: I don't want to change this now, but I'll add a TODO in the code.

---
In reply to: [86900474](https://github.com/Microsoft/WinObjC/pull/1318#discussion_r86900474) [](ancestors = 86900474)

Frameworks/UIKit/UIView.mm
```diff
@@ -3539,37 +3543,30 @@ - (UIView*)snapshotViewAfterScreenUpdates:(BOOL)afterUpdates {
...
 
-// Sets the backing CALayer's XAML contentsElement (which is null by default)
-// This results in the XAML FrameworkElement being added as a child of this UIView's root CALayer.
-- (void)setXamlElement:(WXFrameworkElement*)xamlElement {
-    [self layer].contentsElement = xamlElement;
+/**
+ Microsoft Extension
+*/
++ (WXFrameworkElement*)createXamlElement {
+    // Default to CALayer's default backing Xaml element
```
oliversa-msft: >// Default to CALayer's default backing Xaml element [](start = 4, length = 52)

Stale comment?

oliversa-msft:
>     return self;

Shouldn't this one be the designated initializer and call [super init]  as well? You can then pipe initWithIdentifier through it for a simple optimization of the autoResizingMask setting.

---
Refers to: Frameworks/UIKit/UIViewController.mm:248 in 6d10e3d. [](commit_id = 6d10e3de033387dfab009ef5214bf95ce31b7fe3, deletion_comment = False)

Frameworks/UIKit/UIView.mm
jaredhms: nope, but i'll change it to '// Use CALayer's default backing Xaml element' since i was overusing 'default' :)

---
In reply to: [87280021](https://github.com/Microsoft/WinObjC/pull/1318#discussion_r87280021) [](ancestors = 87280021)

jaredhms:
>     return self;

maybe? I didn't add/change this code though.

---
In reply to: [259531590](https://github.com/Microsoft/WinObjC/pull/1318#issuecomment-259531590) [](ancestors = 259531590)

---
Refers to: Frameworks/UIKit/UIViewController.mm:248 in 6d10e3d. [](commit_id = 6d10e3de033387dfab009ef5214bf95ce31b7fe3, deletion_comment = False)

Frameworks/UIKit/UIViewController.mm
```diff
@@ -1782,19 +1789,18 @@ - (void)_loadXamlPage {
...
     // Activate an instance of the XAML class and its page
     Microsoft::WRL::ComPtr<IInspectable> pageObj = nullptr;
     auto xamlType = ReturnXamlType(priv->_xamlClassName);
     xamlType->ActivateInstance(pageObj.GetAddressOf());
-
-    priv->_page = XamlControls::_createRtProxy([WXCPage class], pageObj.Get());
     [_pageMappings setObject:self forKey:(id)(void*)pageObj.Get()];
 
+    // Create a template UIView
+    priv->_page = CreateRtProxy([WXCPage class], pageObj.Get());
```
oliversa-msft: >CreateRtProxy [](start = 18, length = 13)

This utility function doesn't have the autorelease support. Is that intentional?

jaredhms: we should file a bug to get rid of CreateRtProxy.  I'll move this one to createWith.

---
In reply to: [87281267](https://github.com/Microsoft/WinObjC/pull/1318#discussion_r87281267) [](ancestors = 87281267)

jaredhms: also - why doesn't CreateRtProxy autorelease? Quite strange.

---
In reply to: [87281492](https://github.com/Microsoft/WinObjC/pull/1318#discussion_r87281492) [](ancestors = 87281492,87281267)

Frameworks/UIKit/UIView.mm
```diff
@@ -724,29 +727,34 @@ - (void)updateAccessibility {
 /**
  @Status Interoperable
 */
-+ (instancetype)allocWithZone:(NSZone*)zone {
-    UIView* ret = [super allocWithZone:zone];
-
+- (instancetype)initWithFrame:(CGRect)frame {
     // Run on the main thread because the underlying XAML objects can only be
     // called from the UI thread
     RunSynchronouslyOnMainThread(^{
```
oliversa-msft: >RunSynchronouslyOnMainThread [](start = 4, length = 28)

Should we have called [super init] preceded this as well? UIView -> UIResponder -> NSObject? I noticed that UIWebView has this pattern.

Frameworks/QuartzCore/CATransaction.mm
jaredhms: _layerProxy returns an std::shared_ptr<ILayerProxy>; not doing the nil check lead to AVs.

---
In reply to: [86900578](https://github.com/Microsoft/WinObjC/pull/1318#discussion_r86900578) [](ancestors = 86900578,86900544,86621323)

Frameworks/UIKit/UIView.mm
jaredhms: Maybe? Although, UIView's init calls initWithFrame, which is the designated initializer for UIView.  Regardless, I'd prefer to get this in before we make any more changes to UIView's init path (assuming this is functionally correct and isn't introducing any regressions).

---
In reply to: [87282018](https://github.com/Microsoft/WinObjC/pull/1318#discussion_r87282018) [](ancestors = 87282018)

Frameworks/UIKit.Xaml/Label.xaml.cpp
```diff
@@ -0,0 +1,81 @@
...
+using namespace Microsoft::WRL;
+using namespace Platform;
+using namespace Windows::UI::Xaml;
+using namespace Windows::UI::Xaml::Controls;
+
+namespace UIKit {
+
+Label::Label() {
+    InitializeComponent();
+    Name = L"UILabel";
```
jaredhms: I'll move it up to UILabel.mm so we can set the name of the actual (potentially derived) class.

---
In reply to: [87285187](https://github.com/Microsoft/WinObjC/pull/1318#discussion_r87285187) [](ancestors = 87285187)

oliversa-msft:
:shipit:

Frameworks/UIKit.Xaml/Label.xaml.cpp
jaredhms: actually, it's already handled in UIView.mm, so I'll delete it from here and also from UIButton.mm.

---
In reply to: [87285928](https://github.com/Microsoft/WinObjC/pull/1318#discussion_r87285928) [](ancestors = 87285928,87285187)

PR #1329, opened 2016-11-07
Suppressing ignore extern-initialize warning in winrt template.

pradipd:
Updating template to ignore extern-initialize warning since the ABI headers generated by MIDL initialize GUID's.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1329)


mukhole:
:shipit:

Issue #1271, opened 2016-10-28
Create a set of test images and investigate a diffing platform to ensure rendering parity


DHowett-MSFT:
1. Write rendering code to run on WinObjC and the reference platform.
2. Render to a bitmap.
3. Visually compare the bitmaps to establish a baseline.
4. Use the new baselined images from WinObjC as input to an image diff engine against future renders.

DHowett-MSFT:
Blocked by #1124

DHowett-MSFT:
Right now, greenlines are being generated manually:

```ps1
gci TestIm* | %{
    magick compare -highlight-color green -lowlight-color white "reference\$($_.Name)" $_.Name -compose src Compared_$($_.Name)
    magick montage -label "Reference" "reference\$($_.Name)" -label "WinObjC" $_.Name -label "Diffs" Compared_$($_.Name) -tile 3x1 -geometry +0+0 "FINAL_$($_.Name)"
}
```

Future work: Generate greenlines automatically.

DHowett-MSFT:
This completed as of b67d620.

Issue #1124, opened 2016-10-11
Implement CGBitmapContext backed by CGImage+IWICBitmap


DHowett-MSFT:
Depends on #1072 and #1070.

msft-Jeyaram:
Closed as per #1371

PR #1348, opened 2016-11-10
Add CoreGraphics.Drawing.UnitTests with 22 image-rendering tests.

DHowett-MSFT:
**WARNING**: Every test here will fail until #1124 is complete.

Right now, this test facade emits images to disk using CGBitmapContext. It works on the reference platform.

I decided to make it a standalone test binary for one reason: I think its ideal interface is multi-modal; it needs to be able to generate a corpus of reference images on disk, and it needs to be able to diff them. For this, it needs a custom EntryPoint (included), and the ability to parse command-like arguments.

The default mode will be one that loads reference images from disk and fails tests if a number of pixels differ. *That work is not yet complete.*

Ideally, the draw tests here will be able to be plugged into another module that can perhaps render them to screen, or display them as part of a UI-driven flow for comparison and demoing purposes.

Refs #1271.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1348)


DHowett:
This change contains a Makefile that was copied from Foundation.UnitTests.

**PROPOSAL**
* Create UnitTests/UT.common.mk, with the following interface:

```make
# Required
UT_NAME = Foundation # Can be automatically derived.
UT_FILES = $(wildcard *.cpp) $(wildcard *.mm)

# Optional, if we want pluggable entry points.
UT_FILES += $(UT_DEFAULT_FILES)

# Optional
UT_CFLAGS = -Wall -Werror

# A space-separated list of all linked frameworks.
# Foundation will be required by default as Logging uses it.
# Therefore, it might make sense to put it in UT.common.mk.
UT_FRAMEWORKS = CoreFoundation Foundation
UT_RESOURCE_DIRS = data/ otherdata/

include ../UT.common.mk
```

* Port Foundation.UnitTests over to it.
* Do the above work on *develop*.
* Consume this after a merge from develop.

@ms-jihua This is pretty quick work; what do you think?

tests/UnitTests/CoreGraphics.drawing/data/.keep
```diff
@@ -0,0 +1 @@
+
```
aballway: what is this for?

tests/UnitTests/CoreGraphics.drawing/DrawingTest.h
```diff
@@ -0,0 +1,75 @@
...
+
+#include <TestFramework.h>
+#include <CoreGraphics/CoreGraphics.h>
+
+#include <Starboard/SmartTypes.h>
+
+namespace testing {
+class DrawTest : public ::testing::Test {
+protected:
+    static woc::unique_cf<CGColorSpaceRef> s_deviceColorSpace;
```
aballway: nit: would prefer these be accessed similar to parameterized test variables than implicitly available (say getContext

tests/UnitTests/CoreGraphics.drawing/DrawingTest.h
```diff
@@ -0,0 +1,75 @@
...
+    CGContextRef context;
```
aballway: should this be in a unique_cf?

DHowett-MSFT:
I forgot to add a few test files. That's been fixed.

tests/UnitTests/CoreGraphics.drawing/data/.keep
DHowett-MSFT: Git does not track directories, it tracks files. This is an empty file intended to keep the data directory around.

tests/UnitTests/CoreGraphics.drawing/DrawingTest.cpp
```diff
@@ -0,0 +1,146 @@
...
+    SetUpContext();
+}
+
+void testing::DrawTest::TearDown() {
+    CGContextRef context = GetDrawingContext();
+
+    woc::unique_cf<CGImageRef> image{ CGBitmapContextCreateImage(context) };
+    ASSERT_NE(nullptr, image);
+
+    woc::unique_cf<CFMutableDataRef> imageData{ CFDataCreateMutable(nullptr, 1048576) };
```
ms-jihua: >1048576 [](start = 77, length = 7)

would be good to comment how this number was derived

tests/UnitTests/CoreGraphics.drawing/DrawingTest.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+void testing::DrawTest::SetUpTestCase() {
+    s_deviceColorSpace.reset(CGColorSpaceCreateDeviceRGB());
+}
+
+void testing::DrawTest::TearDownTestCase() {
+    s_deviceColorSpace.release();
+}
+
+CGSize testing::DrawTest::CanvasSize() {
```
ms-jihua: >CanvasSize() [](start = 26, length = 12)

would we ever care about being able to change this?

tests/UnitTests/CoreGraphics.drawing/DrawingTest.cpp
```diff
@@ -0,0 +1,146 @@
...
+    return g_defaultCanvasSize;
+}
+
+void testing::DrawTest::SetUp() {
+    CGSize size = CanvasSize();
+
+    _context.reset(CGBitmapContextCreate(
+        nullptr, size.width, size.height, 8, size.width * 4, s_deviceColorSpace.get(), kCGImageAlphaPremultipliedFirst));
```
ms-jihua: >size.width * 4 [](start = 45, length = 14)

1) are we ever going to have tests that render to colorspaces with not 4 components?
2) if not, call we call that out in a comment?

tests/UnitTests/CoreGraphics.drawing/DrawingTest.cpp
DHowett-MSFT: Actually, the UIKitMimicTest fixture does exactly that; It presents a 2x-scaled context.

tests/UnitTests/CoreGraphics.drawing/DrawingTest.cpp
DHowett-MSFT:
```c
# Number derived by absolutely guessing at the max size of an image; 1MB
1048576
```

tests/UnitTests/CoreGraphics.drawing/DrawingTest.cpp
DHowett-MSFT: That is worth calling out, absolutely.

rajsesh-msft:
:shipit:

PR #1353, opened 2016-11-10
Create CTFrame without going past string length

aballway:
Our private helper method __CreateFrame was assuming that when given a CFRange with length 0, the location would also be 0, so when given a non-zero location we would try and read past the end of the string.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1353)


DHowett-MSFT:
:shipit:

ms-jihua:
:shipit:

PR #1361, opened 2016-11-11
[UIKit] [UIButton] Address incorrect scaling factor for resizable images used as button

davelamb:
backgrounds by adding a Layer Content image element.

Also address overly aggressive layout cycles in UIButton, and autolayout.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1361)

### DIFF
--- a/Frameworks/UIKit.Xaml/Button.xaml
+++ b/Frameworks/UIKit.Xaml/Button.xaml
@@ -5,21 +5,15 @@
     xmlns:local="using:UIKit.Xaml"
     xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
     xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
-    mc:Ignorable="d"
-    Width="50"
-    Height="50">
+    mc:Ignorable="d">
     <Button.Template>
         <ControlTemplate TargetType="Button" x:Name="UIKitButtonControlTemplate">
-            <Grid x:Name="rootGrid" Background="{TemplateBinding Background}">
-                <!-- TODO: Since NineGrid works on image pixels, to support an image with 2x scaling we need to multiply the Size and NineGrid, 
-                     and inverse render transform it back to the right size. Grid clips its children, so we need to host it in a Canvas first. -->
-                <Image x:Name="backgroundImage" 
-                       Width="{TemplateBinding Width}" Height="{TemplateBinding Height}" IsHitTestVisible="False" Stretch="Fill"/>
+            <Canvas x:Name="rootCanvas" Background="{TemplateBinding Background}">
                 <Canvas x:Name="contentCanvas" Width="{TemplateBinding Width}" Height="{TemplateBinding Height}">
                     <Image x:Name="buttonImage" Stretch="Fill" IsHitTestVisible="False" />
                     <TextBlock x:Name="buttonText" MaxLines="1" TextTrimming="CharacterEllipsis" IsHitTestVisible="False"/>
                 </Canvas>
-            </Grid>
+            </Canvas>
         </ControlTemplate>
     </Button.Template>
 </Button>
--- a/Frameworks/UIKit.Xaml/Button.xaml
+++ b/Frameworks/UIKit.Xaml/Button.xaml
@@ -8,11 +8,9 @@
     mc:Ignorable="d">
     <Button.Template>
         <ControlTemplate TargetType="Button" x:Name="UIKitButtonControlTemplate">
-            <Canvas x:Name="rootCanvas" Background="{TemplateBinding Background}">
-                <Canvas x:Name="contentCanvas" Width="{TemplateBinding Width}" Height="{TemplateBinding Height}">
-                    <Image x:Name="buttonImage" Stretch="Fill" IsHitTestVisible="False" />
-                    <TextBlock x:Name="buttonText" MaxLines="1" TextTrimming="CharacterEllipsis" IsHitTestVisible="False"/>
-                </Canvas>
+            <Canvas x:Name="contentCanvas" Width="{TemplateBinding Width}" Height="{TemplateBinding Height}" Background="{TemplateBinding Background}">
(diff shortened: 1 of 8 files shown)


Frameworks/UIKit.Xaml/Button.xaml.cpp
```diff
@@ -47,9 +45,19 @@ Button::Button() {
 
 // Accessor for our Layer content
 Image^ Button::LayerContent::get() {
-    // TODO: Add support for this.  Perhaps we return the same button's image, or we add another image within the canvas.  It should
-    // go just in front of the other button text/image, but behind any sublayers added by CoreAnimation via the SublayerCanvas property.
-    return nullptr;
+    if (!_content) {
+        Canvas^ rootCanvas = safe_cast<Canvas^>(GetTemplateChild(L"rootCanvas"));
```
jaredhms: >Canvas^ rootCanvas = safe_cast<Canvas^>(GetTemplateChild(L"rootCanvas")); [](start = 8, length = 73)

we should probably cache this guy just like we do the others (in OnApplyTemplate)

Frameworks/UIKit.Xaml/Button.xaml
```diff
@@ -5,21 +5,15 @@
...
+    mc:Ignorable="d">
     <Button.Template>
         <ControlTemplate TargetType="Button" x:Name="UIKitButtonControlTemplate">
-            <Grid x:Name="rootGrid" Background="{TemplateBinding Background}">
-                <!-- TODO: Since NineGrid works on image pixels, to support an image with 2x scaling we need to multiply the Size and NineGrid, 
-                     and inverse render transform it back to the right size. Grid clips its children, so we need to host it in a Canvas first. -->
-                <Image x:Name="backgroundImage" 
-                       Width="{TemplateBinding Width}" Height="{TemplateBinding Height}" IsHitTestVisible="False" Stretch="Fill"/>
+            <Canvas x:Name="rootCanvas" Background="{TemplateBinding Background}">
                 <Canvas x:Name="contentCanvas" Width="{TemplateBinding Width}" Height="{TemplateBinding Height}">
```
jaredhms: >contentCanvas [](start = 32, length = 13)

interesting; do we need both canvases or can we just collapse to a single contentCanvas?

Frameworks/UIKit.Xaml/Button.xaml.cpp
```diff
@@ -47,9 +45,19 @@ Button::Button() {
...
+
+        if (rootCanvas) {
+            _content = ref new Image();
+            _content->Name = "LayerContent";
+
+            // Layer content is behind any sublayers
+            rootCanvas->Children->InsertAt(0, _content);
+        }
```
jaredhms: interesting; i was thinking custom-rendered content would go in front of the Xaml-rendered content -- wouldn't you expect drawRect to draw *above* the Xaml contents?

Frameworks/include/UIViewInternal.h
```diff
@@ -107,7 +108,6 @@ class UIViewPrivateState : public LLTreeNode<UIViewPrivateState, UIView> {
 }
 
 - (UITouchPhase)_processPointerEvent:(WUXIPointerRoutedEventArgs*)pointerEventArgs forTouchPhase:(UITouchPhase)touchPhase;
-- (void)_initPriv;
```
jaredhms: thanks

jaredhms:
>  @Status Caveat

Interoperable

---
Refers to: Frameworks/UIKit/UIButton.mm:510 in b7bf419. [](commit_id = b7bf419708e18d44d492dbc6779d01d907ff683f, deletion_comment = False)

Frameworks/UIKit.Xaml/Button.xaml.cpp
jaredhms: after thinking about this a bit more, I think this makes sense.  we draw the button's contents behind its subviews (including its pre-canned label and image).

---
In reply to: [87646260](https://github.com/Microsoft/WinObjC/pull/1361#discussion_r87646260) [](ancestors = 87646260)

Frameworks/UIKit.Xaml/Button.xaml
jaredhms: any thoughts on this? it definitely seems like we can consolidate this to a single canvas.

---
In reply to: [87646076](https://github.com/Microsoft/WinObjC/pull/1361#discussion_r87646076) [](ancestors = 87646076)

Frameworks/UIKit/UIButton.mm
```diff
@@ -289,6 +284,7 @@ - (void)layoutSubviews {
     CGRect imageFrame = [self imageRectForContentRect:contentFrame];
     self.titleLabel.frame = textFrame;
     self.imageView.frame = imageFrame;
+    UIImageSetLayerContents([self layer], self.currentBackgroundImage);
```
jaredhms: contents behind pre-canned subviews do make sense, but this should still get a comment

---
In reply to: [87647058](https://github.com/Microsoft/WinObjC/pull/1361#discussion_r87647058) [](ancestors = 87647058)

Frameworks/UIKit.Xaml/Button.xaml
davelamb: One is for the background image, the other for subviews. I didn't want the background element polluting the children, possibly causing the background to render on top.

Frameworks/UIKit.Xaml/Button.xaml.cpp
davelamb: I didn't cache it since it's not needed anywhere else, and only used once. I could still move it if you'd like.

Frameworks/UIKit.Xaml/Button.xaml
davelamb: Does that behave correctly with bringSublayerToFront etc?

jaredhms: If not, we should make sure it does.

jaredhms:
:shipit:

PR #1366, opened 2016-11-11
Update CT/CG/UI Font creation functions to support both PostScript an…

### LINKED ISSUE #1221: Font display name and PostScript name issue
Hi, this is related to my issue: https://github.com/Microsoft/WinObjC/issues/1029

I have tested with CT-DW branch and didn’t work. It must be because the SDK does not support kCTFont**DisplayName**Attribute yet.

As I have described in the issue, a font file has different names – real full name and PostScript name. Some of the font file have same name for both but there are many files that shows different names for these. If you look at Typography Insight app and see the list of the fonts in Type Inspector menu, there will be many fonts that are displayed as Segoe UI as fallback since it couldn’t recognize the font name with PostScript name.

As you can see in these pages, kCTFontNameAttribute is PostScript name.
https://developer.apple.com/reference/coretext/kctfontnameattribute

Some of the examples are (typically no space between the words)
ITCAvantGardeStd-MdCn
ITCAvantGardeStd-XLtObl
ITCAvantGardeStd-BkObl
ITCAvantGardeStd-Md
ITCAvantGardeStd-MdObl

kCTFontDisplayNameAttribute is full real name. 
https://developer.apple.com/reference/coretext/kctfontdisplaynameattribute?language=objc 

Same example in DisplayName will be:
ITC Avant Garde Standard Medium Condensed
ITC Avant Garde Standard Extra Large Oblique
ITC Avant Garde Standard Black Oblique
ITC Avant Garde Standard Medium
ITC Avant Garde Standard Medium Oblique

Family name is just a single representative name for the groups of the different styles in a single font.
Family name for the sample would be just:
ITC Avant Garde Standard

Conclusion:
If SDK does not implement kCTFontDisplayNameAttribute, apps that use multiple custom fonts or system fonts will not be able to display many fonts. It will be just shown as default fallback Segoe UI.

ms-jihua:
…d Display names

 - Redesigned _DWriteGetFontPropertiesForName to depend on an internal map,
   rather than string parsing, which was fragile against edge cases
 - Updated kCTFontNameAttribute to return the PostScript name, in accordance with documentation
 - Removed #ifdef in unit tests dividing the font names used on each platform

Fixes #1221

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1366)

### DIFF
--- a/Frameworks/CoreGraphics/DWriteWrapper.mm
+++ b/Frameworks/CoreGraphics/DWriteWrapper.mm
@@ -20,8 +20,8 @@
 #include <COMIncludes_End.h>
 
 #import <CoreGraphics/DWriteWrapper.h>
-#import <Starboard.h>
 
+#import <unordered_map>
 #import <vector>
 
 using namespace std;
@@ -42,6 +42,115 @@ static wstring __GetUserDefaultLocaleName() {
     // If the default locale is returned, find that locale name, otherwise use "en-us".
     return wstring(defaultLocaleSuccess ? localeName : c_defaultUserLanguage);
 }
+
+/**
+ * Private helper that creates an uppercase copy of a CFString
+ */
+CFStringRef __CFStringCreateUppercaseCopy(CFStringRef string, CFLocaleRef locale) {
+    CFMutableStringRef ret = CFStringCreateMutableCopy(nullptr, CFStringGetLength(string), string);
+    CFStringUppercase(ret, locale);
+    return ret;
+}
+
+/**
+ * Private helper that creates the IDWriteFontList for a given font family name
+ */
+HRESULT __DWriteGetFontListForFamilyName(CFStringRef familyName, IDWriteFontList** outFontList) {
+    ComPtr<IDWriteFactory> dwriteFactory;
+    RETURN_IF_FAILED(DWriteCreateFactory(DWRITE_FACTORY_TYPE_SHARED, __uuidof(IDWriteFactory), &dwriteFactory));
+
+    // Get the system font collection.
+    ComPtr<IDWriteFontCollection> fontCollection;
+    RETURN_IF_FAILED(dwriteFactory->GetSystemFontCollection(&fontCollection));
+
+    // Get the font family.
+    CFIndex familyNameLength = CFStringGetLength(familyName);
+    std::vector<UniChar> unicharFamilyName(familyNameLength + 1);
+    CFStringGetCharacters(familyName, CFRangeMake(0, familyNameLength), unicharFamilyName.data());
+
+    size_t index = 0;
+    BOOL exists = false;
+
+    RETURN_IF_FAILED(fontCollection->FindFamilyName(reinterpret_cast<wchar_t*>(unicharFamilyName.data()), &index, &exists));
+    RETURN_HR_IF(E_INVALIDARG, !exists);
+
+    ComPtr<IDWriteFontFamily> fontFamily;
+    RETURN_IF_FAILED(fontCollection->GetFontFamily(index, &fontFamily));
+
+    RETURN_IF_FAILED(
+        fontFamily->GetMatchingFonts(DWRITE_FONT_WEIGHT_NORMAL, DWRITE_FONT_STRETCH_NORMAL, DWRITE_FONT_STYLE_NORMAL, outFontList));
+
+    return S_OK;
+}
+
+/**
+ * Private struct that exposes CFHash and CFEqual to an STL container
+ */
+struct __CFStringHashEqual {
+    std::size_t operator()(const woc::unique_cf<CFStringRef>& item) const {
+        return CFHash(item.get());
+    }
+
+    bool operator()(const woc::unique_cf<CFStringRef>& item1, const woc::unique_cf<CFStringRef>& item2) const {
+        return CFEqual(item1.get(), item2.get());
+    }
+};
+
+/**
+ * Private static map, that maps uppercase display name and postscript name, to a _DWriteFontProperties struct for the corresponding font.
+ */
+static std::unordered_map<woc::unique_cf<CFStringRef>, std::shared_ptr<_DWriteFontProperties>, __CFStringHashEqual, __CFStringHashEqual>
+    c_fontPropertiesMap = []() {
+        decltype(c_fontPropertiesMap) ret;
+
+        woc::unique_cf<CFArrayRef> familyNames(_DWriteCopyFontFamilyNames());
+        woc::unique_cf<CFLocaleRef> locale(CFLocaleCopyCurrent());
+
+        for (size_t i = 0; i < CFArrayGetCount(familyNames.get()); ++i) {
+            woc::unique_cf<CFStringRef> familyName(static_cast<CFStringRef>(CFRetain(CFArrayGetValueAtIndex(familyNames.get(), i))));
+            ComPtr<IDWriteFontList> fontList;
+
+            if (FAILED(__DWriteGetFontListForFamilyName(familyName.get(), &fontList))) {
+                continue;
+            }
+
+            size_t count = fontList->GetFontCount();
+
+            for (size_t j = 0; j < count; j++) {
+                ComPtr<IDWriteFont> font;
+                if (FAILED(fontList->GetFont(j, &font))) {
+                    continue;
+                }
+
+                // For each font in that family, fill out a _DWriteFontProperties
+                auto info = std::make_shared<_DWriteFontProperties>();
+                ComPtr<IDWriteLocalizedStrings> displayName;
+                ComPtr<IDWriteLocalizedStrings> postScriptName;
+                BOOL exist;
+
+                if (SUCCEEDED(font->GetInformationalStrings(DWRITE_INFORMATIONAL_STRING_FULL_NAME, &displayName, &exist)) && exist) {
+                    info->displayName.reset(__CFStringCreateUppercaseCopy(_CFStringFromLocalizedString(displayName.Get()), locale.get()));
+                    woc::unique_cf<CFStringRef> uppercaseNameKey(__CFStringCreateUppercaseCopy(info->displayName.get(), locale.get()));
+                    ret.emplace(std::move(uppercaseNameKey), info);
+                }
+
+                if (SUCCEEDED(font->GetInformationalStrings(DWRITE_INFORMATIONAL_STRING_POSTSCRIPT_NAME, &postScriptName, &exist)) &&
+                    exist) {
+                    info->postScriptName.reset(_CFStringFromLocalizedString(postScriptName.Get()));
+                    woc::unique_cf<CFStringRef> uppercaseNameKey(__CFStringCreateUppercaseCopy(info->postScriptName.get(), locale.get()));
+                    ret.emplace(std::move(uppercaseNameKey), info);
+                }
+
+                info->familyName.reset(CFStringCreateCopy(nullptr, familyName.get()));
+
+                info->weight = font->GetWeight();
+                info->stretch = font->GetStretch();
+                info->style = font->GetStyle();
+            }
+        }
+
+        return ret;
+    }();
 
 /**
  * Helper method to convert IDWriteLocalizedStrings object to CFString object.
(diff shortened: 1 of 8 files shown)


Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -191,151 +275,22 @@ CFArrayRef _DWriteCopyFontNamesForFamilyName(CFStringRef familyName) {
...
+    woc::unique_cf<CFLocaleRef> locale(CFLocaleCopyCurrent());
+    woc::unique_cf<CFStringRef> upperFontName(__CFStringCreateUppercaseCopy(fontName, locale.get()));
 
-    if (ret.familyName) {
-        woc::unique_cf<CFArrayRef> familyNameTokens(
-            CFStringCreateArrayBySeparatingStrings(kCFAllocatorDefault, ret.familyName, CFSTR(" ")));
-        for (size_t i = 0; i < CFArrayGetCount(familyNameTokens.get()); ++i) {
-            CFSetRemoveValue(propertyTokens.get(), CFArrayGetValueAtIndex(familyNameTokens.get(), i));
-        }
+    const auto& info = c_fontPropertiesMap.find(upperFontName);
```
aballway: nit: if c_fontPropertiesMap is only used here, would prefer it scoped within the function

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -44,6 +44,115 @@ static wstring __GetUserDefaultLocaleName() {
...
+
+            if (FAILED(__DWriteGetFontListForFamilyName(familyName.get(), &fontList))) {
+                continue;
+            }
+
+            size_t count = fontList->GetFontCount();
+
+            for (size_t j = 0; j < count; j++) {
+                ComPtr<IDWriteFont> font;
+                if (FAILED(fontList->GetFont(j, &font))) {
```
aballway: is there a performance impact from iterating through and initializing with all fonts? Could this be done more lazily?

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -191,151 +275,22 @@ CFArrayRef _DWriteCopyFontNamesForFamilyName(CFStringRef familyName) {
...
-    // In addition, some fonts with properties are occasionally placed into their own family (eg: Segoe WP SemiLight -> Segoe WP SemiLight)
-    // Try to be more prudent about these edge cases, by looking only at the difference between the font name and family name
-    woc::unique_cf<CFArrayRef> fontNameTokens(CFStringCreateArrayBySeparatingStrings(kCFAllocatorDefault, fontName, CFSTR(" ")));
-    woc::unique_cf<CFMutableSetRef> propertyTokens(CFSetCreateMutable(kCFAllocatorDefault, 0, &kCFTypeSetCallBacks));
-
-    for (size_t i = 0; i < CFArrayGetCount(fontNameTokens.get()); ++i) {
-        CFSetAddValue(propertyTokens.get(), CFArrayGetValueAtIndex(fontNameTokens.get(), i));
-    }
+std::shared_ptr<_DWriteFontProperties> _DWriteGetFontPropertiesFromName(CFStringRef fontName) {
+    woc::unique_cf<CFLocaleRef> locale(CFLocaleCopyCurrent());
```
aballway: nit: can this be static?

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -393,24 +348,24 @@ HRESULT _DWriteCreateFontFamilyWithName(CFStringRef familyName, IDWriteFontFamil
...
     // Eg: Bold, Condensed, Light, Italic
-    _DWriteFontProperties properties = _DWriteGetFontPropertiesFromName(name);
+    std::shared_ptr<_DWriteFontProperties> properties = _DWriteGetFontPropertiesFromName(name);
 
     // TODO: #1250: Need to be able to load fonts from the app's bundle
     // For now return a default font to avoid crashes in case of missing fonts
     // When #1250 is completed, remove this
-    if (!properties.familyName) {
+    if (!properties->familyName.get()) {
         name = CFSTR("Segoe UI");
```
aballway: nit: should pull this into a constant while you're here

Frameworks/include/CoreGraphics/DWriteWrapper.h
```diff
@@ -35,14 +37,17 @@ COREGRAPHICS_EXPORT CFArrayRef _DWriteCopyFontNamesForFamilyName(CFStringRef fam
...
+    DWRITE_FONT_STYLE style = DWRITE_FONT_STYLE_NORMAL;
+
+    woc::unique_cf<CFStringRef> displayName;
+    woc::unique_cf<CFStringRef> postScriptName;
+    woc::unique_cf<CFStringRef> familyName;
 };
 
 // Create DWrite objects
-COREGRAPHICS_EXPORT _DWriteFontProperties _DWriteGetFontPropertiesFromName(CFStringRef fontName);
+extern "C++" std::shared_ptr<_DWriteFontProperties> _DWriteGetFontPropertiesFromName(CFStringRef fontName);
```
aballway: nit: should this be wrapped in #ifdef c++?

Frameworks/CoreGraphics/DWriteWrapper.mm
DHowett-MSFT: I would very much prefer that _not_ to be the case

---
In reply to: [87663736](https://github.com/Microsoft/WinObjC/pull/1366#discussion_r87663736) [](ancestors = 87663736)

DHowett-MSFT:
:shipit:

Frameworks/CoreGraphics/DWriteWrapper.mm
ms-jihua: Agreed, I think the extra nesting would only hurt readability, and this variable seems like we might reuse it for another function in the future anyway.

---
In reply to: [87678020](https://github.com/Microsoft/WinObjC/pull/1366#discussion_r87678020) [](ancestors = 87678020,87663736)

Frameworks/CoreGraphics/DWriteWrapper.mm
ms-jihua: this version is most likely (can't say for sure) more performant than our previous version, as it lacks the string parsing. we can make this a little more lazy, but not by much - we can't avoid iterating through all fonts and families. we can do it later, but it would probably hurt readability (see other comment) for now, leaving as-is. we can pursue optimizations at another point.

---
In reply to: [87664127](https://github.com/Microsoft/WinObjC/pull/1366#discussion_r87664127) [](ancestors = 87664127)

Frameworks/CoreGraphics/DWriteWrapper.mm
ms-jihua: locale can change through the life of an app.

---
In reply to: [87664268](https://github.com/Microsoft/WinObjC/pull/1366#discussion_r87664268) [](ancestors = 87664268)

Frameworks/include/CoreGraphics/DWriteWrapper.h
ms-jihua: yeah probably.

---
In reply to: [87664812](https://github.com/Microsoft/WinObjC/pull/1366#discussion_r87664812) [](ancestors = 87664812)

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -44,6 +44,115 @@ static wstring __GetUserDefaultLocaleName() {
...
+    ComPtr<IDWriteFontCollection> fontCollection;
+    RETURN_IF_FAILED(dwriteFactory->GetSystemFontCollection(&fontCollection));
+
+    // Get the font family.
+    CFIndex familyNameLength = CFStringGetLength(familyName);
+    std::vector<UniChar> unicharFamilyName(familyNameLength + 1);
+    CFStringGetCharacters(familyName, CFRangeMake(0, familyNameLength), unicharFamilyName.data());
+
+    size_t index = 0;
+    BOOL exists = false;
```
rajsesh-msft: >BOOL [](start = 4, length = 4)

this should be bool, or should be assigned a NO.

Frameworks/CoreGraphics/DWriteWrapper.mm
ms-jihua: I think I talked about this before?
- this is code that will eventually be removed.
- this is only used here.
- there are default font constants elsewhere, pulling this into a constant would probably add to confusion.

---
In reply to: [87664357](https://github.com/Microsoft/WinObjC/pull/1366#discussion_r87664357) [](ancestors = 87664357)

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -44,6 +44,115 @@ static wstring __GetUserDefaultLocaleName() {
...
+    CFStringUppercase(ret, locale);
+    return ret;
+}
+
+/**
+ * Private helper that creates the IDWriteFontList for a given font family name
+ */
+HRESULT __DWriteGetFontListForFamilyName(CFStringRef familyName, IDWriteFontList** outFontList) {
+    ComPtr<IDWriteFactory> dwriteFactory;
+    RETURN_IF_FAILED(DWriteCreateFactory(DWRITE_FACTORY_TYPE_SHARED, __uuidof(IDWriteFactory), &dwriteFactory));
```
ms-jihua: I mentioned this on a previous PR - @bbowman stated that factory objects such as this, in the windows api world, often already implement their own reuse on the back end, and that by keeping one as a static, we would be interfering with their reusage patterns.

---
In reply to: [87900429](https://github.com/Microsoft/WinObjC/pull/1366#discussion_r87900429) [](ancestors = 87900429)

Frameworks/CoreGraphics/DWriteWrapper.mm
ms-jihua: this is win32 shouty BOOL, not c bool or objC BOOL. I can make this caps TRUE or FALSE if you want (maybe more correct?).

---
In reply to: [87900568](https://github.com/Microsoft/WinObjC/pull/1366#discussion_r87900568) [](ancestors = 87900568)

Frameworks/CoreGraphics/DWriteWrapper.mm
rajsesh-msft: Font enumeration is still expensive, Lets cover this as part of #1374. [https://github.com/Microsoft/WinObjC/issues/1374](https://github.com/Microsoft/WinObjC/issues/1374) 

---
In reply to: [87900133](https://github.com/Microsoft/WinObjC/pull/1366#discussion_r87900133) [](ancestors = 87900133,87664127)

Frameworks/CoreGraphics/DWriteWrapper.mm
rajsesh-msft: Yes, shouty FALSE then.

---
In reply to: [87901837](https://github.com/Microsoft/WinObjC/pull/1366#discussion_r87901837) [](ancestors = 87901837,87900568)

Frameworks/CoreGraphics/DWriteWrapper.mm
rajsesh-msft: Not DWrite factories.  These aren't same as winrt activation objects. [https://msdn.microsoft.com/en-us/library/windows/desktop/dd368040(v=vs.85).aspx](https://msdn.microsoft.com/en-us/library/windows/desktop/dd368040(v=vs.85).aspx). This function creates a DirectWrite factory object that is used for subsequent creation of individual DirectWrite objects. DirectWrite factory contains internal state data such as font loader registration and cached font data. In most cases it is recommended you use the shared factory object, because it allows multiple components that use DirectWrite to share internal DirectWrite state data, and thereby reduce memory usage. However, there are cases when it is desirable to reduce the impact of a component, such as a plug-in from an untrusted source, on the rest of the process, by sandboxing and isolating it from the rest of the process components. In such cases, it is recommended you use an isolated factory for the sandboxed component. 

---
In reply to: [87901335](https://github.com/Microsoft/WinObjC/pull/1366#discussion_r87901335) [](ancestors = 87901335,87900429)

ms-jihua: "In most cases it is recommended you use the shared factory object, because it allows multiple components that use DirectWrite to share internal DirectWrite state data, and thereby reduce memory usage. "

By this it means, DWRITE_FACTORY_TYPE_SHARED, which we are already doing.

---
In reply to: [87902601](https://github.com/Microsoft/WinObjC/pull/1366#discussion_r87902601) [](ancestors = 87902601,87901335,87900429)

rajsesh-msft:
:shipit:

PR #1367, opened 2016-11-11
NSString+UIKitAdditions treat zero constraints as unlimited

### LINKED ISSUE #1334: -[UISegmentedControl sizeToFit] regression
Dear WinObjC Engineers,

From release 41 to 42 there is a regression in -[UISegmentedControl sizeToFit]. In release 41 it essentially worked. In release 42, the segment widths appear to be adjusted to the width of the first character in the corresponding segment (as opposed to the width of the entire label [text] in that segment).

Sincerely,

Ed

aballway:
When the reference platform gives size constraints of zero to any NSString+UIKitAdditions methods it expects it to be treated as unlimited width and/or height.

Fixes #1334

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1367)

### DIFF
--- a/Frameworks/UIKit/NSString+UIKitAdditions.mm
+++ b/Frameworks/UIKit/NSString+UIKitAdditions.mm
@@ -311,11 +311,11 @@ static NSDictionary* _getDefaultUITextAttributes() {
     woc::unique_cf<CTFramesetterRef> framesetter{ CTFramesetterCreateWithAttributedString((__bridge CFAttributedStringRef)attributedSelf) };
 
     if (size.width == 0.0) {
-        size.width = FLT_MAX;
+        size.width = std::numeric_limits<CGFloat>::max();
     }
 
     if (size.height == 0.0) {
-        size.height = FLT_MAX;
+        size.height = std::numeric_limits<CGFloat>::max();
     }
 
     return CTFramesetterSuggestFrameSizeWithConstraints(framesetter.get(), CFRangeMake(0, self.length), nullptr, size, nullptr);
--- a/Frameworks/UIKit/NSString+UIKitAdditions.mm
+++ b/Frameworks/UIKit/NSString+UIKitAdditions.mm
@@ -308,11 +308,17 @@ static NSDictionary* _getDefaultUITextAttributes() {
     }
 
     NSAttributedString* attributedSelf = [[[NSAttributedString alloc] initWithString:self attributes:attributes] autorelease];
+    woc::unique_cf<CTFramesetterRef> framesetter{ CTFramesetterCreateWithAttributedString((__bridge CFAttributedStringRef)attributedSelf) };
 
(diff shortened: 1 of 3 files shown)


Frameworks/UIKit/NSString+UIKitAdditions.mm
```diff
@@ -308,11 +308,17 @@ - (CGSize)_sizeWithAttributes:(NSDictionary<NSString*, id>*)attributes constrain
...
+    woc::unique_cf<CTFramesetterRef> framesetter{ CTFramesetterCreateWithAttributedString((__bridge CFAttributedStringRef)attributedSelf) };
 
-    CTFramesetterRef framesetter = CTFramesetterCreateWithAttributedString((__bridge CFAttributedStringRef)attributedSelf);
-    CFAutorelease(framesetter);
+    if (size.width == 0.0) {
+        size.width = FLT_MAX;
+    }
+
+    if (size.height == 0.0) {
+        size.height = FLT_MAX;
```
ms-jihua: >FLT_MAX; [](start = 22, length = 8)

nit: std::numeric_limits<CGFloat>::max(), for consistency with the rest of this file?

tests/unittests/UIKit/NSString+UIKitAdditionsTests.mm
```diff
@@ -0,0 +1,25 @@
...
+//
+//******************************************************************************
+
+#import <TestFramework.h>
+#import <Starboard/SmartTypes.h>
+#import <UIKit/UIKit.h>
+
+TEST(NSString_UIKitAdditions, ShouldNotReturnSizeOfZeroWidth) {
+    NSString* string = @"TEST";
+    CGSize size = [string sizeWithFont:[UIFont systemFontOfSize:12.0f] constrainedToSize:{ 0, 0 } lineBreakMode:NSLineBreakByWordWrapping];
```
ms-jihua: >string [](start = 19, length = 6)

nit: could just use a literal

tests/unittests/UIKit/NSString+UIKitAdditionsTests.mm
```diff
@@ -0,0 +1,25 @@
...
+    EXPECT_NE(0, size.width);
```
ms-jihua: >EXPECT_NE(0, size.width); [](start = 4, length = 25)

this is a bit loose, I would probably check specifically that width and height are CGFloat max

aballway: we want the size to be as most FLT_MAX, it should be the size of the text at that font, but this is only testing that we don't say the width is as most 0. #ByDesign

ms-jihua:
:shipit:

DHowett-MSFT:
:shipit:

PR #1368, opened 2016-11-11
Add support for clipping (and EO clipping); add support for non-EO fill.

DHowett-MSFT:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1368)

### DIFF
--- a/Frameworks/CoreGraphics/CGContext.mm
+++ b/Frameworks/CoreGraphics/CGContext.mm
@@ -38,6 +38,7 @@
 #import <wrl/client.h>
 #include <COMIncludes_end.h>
 #import <LoggingNative.h>
+#import <CoreGraphics/D2DWrapper.h>
 
 #import <list>
 #import <vector>
@@ -1479,7 +1480,13 @@ static HRESULT __CGContextRenderToCommandList(CGContextRef context,
 }
 
 static HRESULT __CGContextRenderImage(CGContextRef context, ID2D1Image* image) {
+    if (!image) {
+        // Being asked to draw nothing is valid!
+        return S_OK;
+    }
+
     ComPtr<ID2D1RenderTarget> renderTarget = context->RenderTarget();
+
     auto& state = context->CurrentGState();
 
     ComPtr<ID2D1DeviceContext> deviceContext;
@@ -1517,24 +1524,31 @@ static HRESULT __CGContextRenderImage(CGContextRef context, ID2D1Image* image) {
 
 static HRESULT __CGContextDrawGeometry(CGContextRef context,
                                        _CGCoordinateMode coordinateMode,
-                                       ID2D1Geometry* geometry,
+                                       ID2D1Geometry* pGeometry,
                                        CGPathDrawingMode drawMode) {
     ComPtr<ID2D1CommandList> commandList;
+    ComPtr<ID2D1Geometry> geometry(pGeometry);
     HRESULT hr = __CGContextRenderToCommandList(
         context, coordinateMode, nullptr, &commandList, [geometry, drawMode](CGContextRef context, ID2D1DeviceContext* deviceContext) {
             auto& state = context->CurrentGState();
             if (drawMode & kCGPathFill) {
-                if (drawMode & kCGPathEOFill) {
-                    // TODO(DH): GH#1077 Regenerate geometry in Even/Odd fill mode.
-                }
-                deviceContext->FillGeometry(geometry, state.fillBrush.Get());
+                state.fillBrush->SetOpacity(state.alpha);
+
+                ComPtr<ID2D1Geometry> geometryToFill;
+                D2D1_FILL_MODE d2dFillMode =
+                    (drawMode & kCGPathEOFill) == kCGPathEOFill ? D2D1_FILL_MODE_ALTERNATE : D2D1_FILL_MODE_WINDING;
+                RETURN_IF_FAILED(_CGConvertD2DGeometryToFillMode(geometry.Get(), d2dFillMode, &geometryToFill));
+
+                deviceContext->FillGeometry(geometryToFill.Get(), state.fillBrush.Get());
             }
 
             if (drawMode & kCGPathStroke && std::fpclassify(state.lineWidth) != FP_ZERO) {
                 // This only computes the stroke style if its parameters have changed since the last draw.
                 state.ComputeStrokeStyle(deviceContext);
 
-                deviceContext->DrawGeometry(geometry, state.strokeBrush.Get(), state.lineWidth, state.strokeStyle.Get());
+                state.strokeBrush->SetOpacity(state.alpha);
+
+                deviceContext->DrawGeometry(geometry.Get(), state.strokeBrush.Get(), state.lineWidth, state.strokeStyle.Get());
             }
 
             return S_OK;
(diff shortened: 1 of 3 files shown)


Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1488,11 +1552,14 @@ static HRESULT __CGContextRenderImage(CGContextRef context, ID2D1Image* image) {
     deviceContext->BeginDraw();
 
     bool layer = false;
-    if (state.alpha != 1.0f || false /* mask, clip, etc. */) {
+    if (state.alpha != 1.0f || state.clippingGeometry) {
```
DHowett-MSFT: #1369 removes `alpha` from this check.

DHowett-MSFT:
These will be tested by #1348.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -183,6 +187,49 @@ inline void ClearPath() {
...
+            return S_OK;
+        }
+
+        auto& state = CurrentGState();
+
+        ComPtr<ID2D1Geometry> additionalClippingGeometry;
+        RETURN_IF_FAILED(_CGPathGetGeometry(Path(), &additionalClippingGeometry));
+        ClearPath();
+
+        D2D1_FILL_MODE d2dFillMode = (pathMode & kCGPathEOFill) == kCGPathEOFill ? D2D1_FILL_MODE_ALTERNATE : D2D1_FILL_MODE_WINDING;
```
aballway: nit: why not just (path & kCGPathEOFill) ?

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -183,6 +187,49 @@ inline void ClearPath() {
...
+
+            RETURN_IF_FAILED(state.clippingGeometry->CombineWithGeometry(additionalClippingGeometry.Get(),
+                                                                         D2D1_COMBINE_MODE_INTERSECT,
+                                                                         nullptr,
+                                                                         geometrySink.Get()));
+
+            RETURN_IF_FAILED(geometrySink->Close());
+
+            return newClippingPathGeometry.As(&state.clippingGeometry);
+        } else {
```
msft-Jeyaram: > else { [](start = 9, length = 7)

not needed?

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -787,9 +834,20 @@ void CGContextClipToMask(CGContextRef context, CGRect dest, CGImageRef image) {
...
 */
 CGRect CGContextGetClipBoundingBox(CGContextRef context) {
-    NOISY_RETURN_IF_NULL(context, StubReturn());
-    UNIMPLEMENTED();
-    return StubReturn();
+    NOISY_RETURN_IF_NULL(context, CGRectNull);
+
+    auto& state = context->CurrentGState();
+    if (!state.clippingGeometry) {
+        return CGRectNull;
```
aballway: Does this match reference platform behavior when nothing is being clipped?

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1488,11 +1552,14 @@ static HRESULT __CGContextRenderImage(CGContextRef context, ID2D1Image* image) {
     deviceContext->BeginDraw();
 
     bool layer = false;
```
aballway: nit: rename something more informative

Frameworks/CoreGraphics/D2DWrapper.mm
```diff
@@ -32,4 +32,18 @@ HRESULT _CGGetWICFactory(IWICImagingFactory** factory) {
...
     sWicFactory.CopyTo(factory);
     RETURN_HR(sHr);
+}
+
+HRESULT _CGConvertD2DGeometryToFillMode(ID2D1Geometry* geometry, D2D1_FILL_MODE fillMode, ID2D1Geometry** pNewGeometry) {
+    ComPtr<ID2D1Factory> factory;
+    geometry->GetFactory(&factory);
+
+    ComPtr<ID2D1GeometryGroup> geometryGroup;
+    RETURN_IF_FAILED(factory->CreateGeometryGroup(fillMode, &geometry, 1, &geometryGroup));
```
aballway: nit: magic number

Frameworks/CoreGraphics/CGContext.mm
aballway: better yet invert the if to get the easy branch out of the way

Frameworks/CoreGraphics/D2DWrapper.mm
DHowett-MSFT:
```c
int numberOfGeometriesInGeometries = sizeof(&geometry) / sizeof(geometry);
```

👎

Frameworks/CoreGraphics/CGContext.mm
DHowett-MSFT: you hate early returns.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -183,6 +187,50 @@ inline void ClearPath() {
...
+        ComPtr<ID2D1Geometry> additionalClippingGeometry;
+        RETURN_IF_FAILED(_CGPathGetGeometry(Path(), &additionalClippingGeometry));
+        ClearPath();
+
+        D2D1_FILL_MODE d2dFillMode = (pathMode & kCGPathEOFill) == kCGPathEOFill ? D2D1_FILL_MODE_ALTERNATE : D2D1_FILL_MODE_WINDING;
+
+        if (!state.clippingGeometry) {
+            // If we don't have a clipping geometry, we are free to take this one wholesale (after EO/Winding conversion.)
+            ComPtr<ID2D1Geometry> convertedClippingGeometry;
+            RETURN_IF_FAILED(_CGConvertD2DGeometryToFillMode(additionalClippingGeometry.Get(), d2dFillMode, &convertedClippingGeometry));
```
rajsesh-msft: nit: Could you not pass &state.clippingGeometry here?

Frameworks/CoreGraphics/D2DWrapper.mm
```diff
@@ -32,4 +32,18 @@ HRESULT _CGGetWICFactory(IWICImagingFactory** factory) {
     static HRESULT sHr = CoCreateInstance(CLSID_WICImagingFactory, nullptr, CLSCTX_INPROC_SERVER, IID_PPV_ARGS(&sWicFactory));
     sWicFactory.CopyTo(factory);
     RETURN_HR(sHr);
+}
+
+HRESULT _CGConvertD2DGeometryToFillMode(ID2D1Geometry* geometry, D2D1_FILL_MODE fillMode, ID2D1Geometry** pNewGeometry) {
```
rajsesh-msft: Its a shame that we will always do this conversion even if the geometry is in the right fill mode.  Since we know our defaults, shouldn't we only do the conversion if the mode is different?

Frameworks/CoreGraphics/CGContext.mm
DHowett-MSFT: Oh! Yes. This got muddled in the condition flip AJ suggested.

---
In reply to: [87893114](https://github.com/Microsoft/WinObjC/pull/1368#discussion_r87893114) [](ancestors = 87893114)

Frameworks/CoreGraphics/D2DWrapper.mm
DHowett-MSFT: Indeed. We absolutely should; this should be part of CGPath, but CGPath's default is backwards from CG's default right now.

I'll file an issue to track it and its removal here.

---
In reply to: [87899206](https://github.com/Microsoft/WinObjC/pull/1368#discussion_r87899206) [](ancestors = 87899206)

MSFTFox:
:shipit:

DHowett-MSFT:
@rajsesh-msft Addressed your feedback! I added a comment on the geometry cloning and simplified `Clip()` a little.

PR #1369, opened 2016-11-11
Propagate context alpha into brushes for per-primitive draw, not layer

DHowett-MSFT:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1369)

### DIFF
--- a/Frameworks/CoreGraphics/CGContext.mm
+++ b/Frameworks/CoreGraphics/CGContext.mm
@@ -183,6 +183,10 @@ struct __CGContext : CoreFoundation::CppBase<__CGContext> {
         renderTarget->GetFactory(&factory);
         return factory;
     }
+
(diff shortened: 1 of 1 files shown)


Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1488,7 +1492,7 @@ static HRESULT __CGContextRenderImage(CGContextRef context, ID2D1Image* image) {
     deviceContext->BeginDraw();
 
     bool layer = false;
-    if (state.alpha != 1.0f || false /* mask, clip, etc. */) {
+    if (false /* mask, clip, etc. */) {
```
DHowett-MSFT: I am not removing this because #1368 will just re-add it.

DHowett-MSFT:
This will be tested by #1348.

Issue #1337, opened 2016-11-08
Use the DisplayTexture interface implementation from UX when it's ready.


msft-Jeyaram:
Currently to make code better we've introduced a self contained DisplayTexture class that implements IDisplayTexture.

Until #1318 makes it to our branch. We have implemented IDisplayTextureImpl to be better compatible. Upon the change propagating to our branch, IDisplayTextureImpl can be removed.

PR #1371, opened 2016-11-14
Implementation of CGBitmapContext

msft-Jeyaram:
#1124

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1371)


Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1833,78 +1835,153 @@ inline void SetImage(CGImageRef image) {
...
+
+    if (_CGIsValidRenderTargetPixelFormat(pixelFormat)) {
+        // if data is null, enough memory is allocated via CGIWICBitmap
+        customBitmap = Make<CGIWICBitmap>(data, pixelFormat, height, width);
+    } else {
+        // TODO #<GITHUB-ID>: this will be an issue if we have change in stride, account for that.
+        // Also as per documentation, it's best to leave CGBitmapContext to manage the memory
+        if (data != nullptr) {
+            return nullptr;
+        }
```
msft-Jeyaram: we have an issue in this case where data was provided and we might end up needing more space due to change in format.
e.g 24->32

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1833,78 +1835,153 @@ inline void SetImage(CGImageRef image) {
...
+CGContextRef CGBitmapContextCreateWithData(void* data,
+                                           size_t width,
+                                           size_t height,
+                                           size_t bitsPerComponent,
+                                           size_t bytesPerRow,
+                                           CGColorSpaceRef space,
+                                           uint32_t bitmapInfo,
+                                           CGBitmapContextReleaseDataCallback releaseCallback,
+                                           void* releaseInfo) {
+    RETURN_NULL_IF(!width);
```
aballway: nit: move these into one call

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1833,78 +1835,153 @@ inline void SetImage(CGImageRef image) {
...
+    RETURN_NULL_IF(!height);
+    RETURN_NULL_IF(!space);
+
+    size_t bitsPerPixel = ((bytesPerRow / width) << 3);
```
aballway: nit: no need for bit shifting, do a multiply and let the compiler optimize.  Also, add a comment why multiplying

Frameworks/CoreGraphics/CGContext.mm
msft-Jeyaram: we have the functionality to convert, but we still need a direct access to the backed buffer (WIP)

---
In reply to: [87855777](https://github.com/Microsoft/WinObjC/pull/1371#discussion_r87855777) [](ancestors = 87855777)

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1833,78 +1835,153 @@ inline void SetImage(CGImageRef image) {
...
+    ComPtr<ID2D1RenderTarget> renderTarget;
+    RETURN_NULL_IF_FAILED(factory->CreateWicBitmapRenderTarget(customBitmap.Get(), D2D1::RenderTargetProperties(), &renderTarget));
+    return _CGBitmapContextCreateWithRenderTarget(renderTarget.Get(), image.get());
+}
+
+/**
+ @Status Interoperable
+*/
+CGBitmapInfo CGBitmapContextGetBitmapInfo(CGContextRef context) {
+    NOISY_RETURN_IF_NULL(context, 0);
```
aballway: Is 0 the default bitmap info? Should this be a constant? And do all of these match reference platform?

Frameworks/include/CGIWICBitmap.h
```diff
@@ -106,10 +115,31 @@ class CGIWICBitmap
...
+        : m_dataBuffer(nullptr), m_texture(texture), m_pixelFormat(pixelFormat), m_height(height), m_width(width), m_freeData(false) {
+    }
+
+    // if data is null, then a buffer that is the size of bytes per row is allocated.
+    CGIWICBitmap(_In_ void* data, _In_ WICPixelFormatGUID pixelFormat, _In_ UINT height, _In_ UINT width)
+        : m_texture(nullptr), m_pixelFormat(pixelFormat), m_height(height), m_width(width), m_freeData(false) {
+        // Obtain bytes per row from pixelFormat
+        const __CGImagePixelProperties* properties = _CGGetPixelFormatProperties(m_pixelFormat);
+        FAIL_FAST_IF(properties == nullptr);
+        m_bytesPerRow = (properties->bitsPerPixel >> 3) * m_width;
```
aballway: nit: just divide here

Frameworks/include/CGIWICBitmap.h
```diff
@@ -106,10 +115,31 @@ class CGIWICBitmap
...
+        if (data) {
+            m_dataBuffer = static_cast<BYTE*>(data);
+        } else {
+            m_dataBuffer = static_cast<BYTE*>(IwMalloc(m_height * m_bytesPerRow));
+            m_freeData = true;
+        }
     }
 
     ~CGIWICBitmap() {
+        if (m_freeData) {
```
aballway: this seems very unsafe?  Is there some container which could take care of this?

tests/unittests/CoreGraphics/CGBitmapContextTests.mm
```diff
@@ -44,24 +44,36 @@
     CGContextRelease(context);
 }
 
-DISABLED_TEST(CGBitmapContext, BitmapInfoAPIs_RGB) {
+TEST(CGBitmapContext, BitmapInfoAPIs_RGB) {
```
aballway: feels a bit sparse on tests given how much you've just implemented. I'd prefer a lot more test coverage hitting edge conditions and comparing against reference platform to verify correctness.

Frameworks/CoreGraphics/CGContext.mm
msft-Jeyaram: I think it's fine as it is.

---
In reply to: [87858489](https://github.com/Microsoft/WinObjC/pull/1371#discussion_r87858489) [](ancestors = 87858489)

Frameworks/CoreGraphics/CGContext.mm
msft-Jeyaram: I think we are already there, so not much of an issue.


---
In reply to: [87858799](https://github.com/Microsoft/WinObjC/pull/1371#discussion_r87858799) [](ancestors = 87858799)

build/CoreGraphics/dll/CoreGraphics.def
```diff
@@ -236,6 +236,7 @@ LIBRARY CoreGraphics
         _CGContextSetShadowProjectionTransform
         _CGContextCreateWithD2DRenderTarget
         _CGGetD2DFactory
+        _CGGetPixelFormatProperties
```
DHowett-MSFT: >_CGGetPixelFormatProperties [](start = 8, length = 27)

Does this need to be exported?

Frameworks/include/CGIWICBitmap.h
```diff
@@ -40,6 +40,8 @@
 #import <D2d1.h>
 #include <COMIncludes_End.h>
 
+#import <memory.h>
+
```
DHowett-MSFT: `<memory>` only

tests/unittests/CoreGraphics/CGBitmapContextTests.mm
msft-Jeyaram: Sadly not all of it will work against the reference platform, we already have a caveat regarding that.

---
In reply to: [87862080](https://github.com/Microsoft/WinObjC/pull/1371#discussion_r87862080) [](ancestors = 87862080)

build/CoreGraphics/dll/CoreGraphics.def
msft-Jeyaram: hmmm I don't think so.

---
In reply to: [87862945](https://github.com/Microsoft/WinObjC/pull/1371#discussion_r87862945) [](ancestors = 87862945)

Frameworks/include/CGIWICBitmap.h
```diff
@@ -106,10 +115,33 @@ class CGIWICBitmap
...
+            m_dataBuffer = static_cast<BYTE*>(IwMalloc(m_height * m_bytesPerRow));
+            m_freeData = true;
+        }
     }
 
     ~CGIWICBitmap() {
+        if (m_freeData) {
+            IwFree(m_dataBuffer);
+            m_dataBuffer = nullptr;
+            m_freeData = false;
```
DHowett-MSFT: You don't need to clear these here; the dtor should never get called twice.

Frameworks/include/CGIWICBitmap.h
```diff
@@ -106,10 +115,33 @@ class CGIWICBitmap
...
+        FAIL_FAST_IF(properties == nullptr);
+
+        //bytesPerRow = ((bitsPerPixel) / 8 byte/pixel) * width
+        m_bytesPerRow = (properties->bitsPerPixel >> 3) * m_width;
+
+        if (data) {
+            m_dataBuffer = static_cast<BYTE*>(data);
+        } else {
+            m_dataBuffer = static_cast<BYTE*>(IwMalloc(m_height * m_bytesPerRow));
+            m_freeData = true;
```
DHowett-MSFT: I suppose you could use new[] and delete[] instead of IwMalloc and IwFree!

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1833,78 +1835,155 @@ inline void SetImage(CGImageRef image) {
 
 /**
  @Status Caveat
- @Notes Limited bitmap formats available. Decode, shouldInterpolate, intent parameters
- and some byte orders ignored.
- */
+ @Notes If the image colorspace is not a valid render target format (as per
+ https://msdn.microsoft.com/en-us/library/windows/desktop/dd756766(v=vs.85).aspx#supported_wic_formats), it's converted to 32bpp BGRA
+ format. Does not support conversion if data is already provided.
+*/
```
rajsesh-msft: Could we instead document the CG bitmap formats that we can indeed support, we have already done the mapping and calling it out here will make it easier for anyone using the code.  Also, if we don't support the specified format, please return null instead of silently picking a different format.

Frameworks/include/CGIWICBitmap.h
msft-Jeyaram: I usually get comments to set them to invalidate state. Leaving as it is

Frameworks/CoreGraphics/CGContext.mm
msft-Jeyaram: Sure, I can do that.

Frameworks/include/CGIWICBitmap.h
```diff
@@ -126,22 +160,52 @@ class CGIWICBitmap
             WICRect rect = { 0, 0, m_width, m_height };
             region = &rect;
         } else {
+            // This is for tracking purpose, we do not support sub-regions.
+            // TODO #1379 - should support regional locking.
+            // This check will be removed.
+            if (region->X || region->Y || (region->Height < m_height) || (region->Width < m_width)) {
```
msft-Jeyaram: >>Height < m_height) || (region->Width < m_width) [](start = 50, length = 48)

more so it can be !=

Frameworks/include/CGIWICBitmap.h
```diff
@@ -126,22 +160,52 @@ class CGIWICBitmap
...
+                return E_NOTIMPL;
+            }
+
             RETURN_HR_IF(E_INVALIDARG, (region->Width > m_width) || (region->Height > m_height));
```
msft-Jeyaram: >   RETURN_HR_IF(E_INVALIDARG, (region->Width > m_width) || (region->Height > m_height)); [](start = 9, length = 88)

this might be doing something similar to above, but above code will eventually go away.

tests/unittests/CoreGraphics/CGBitmapContextTests.mm
```diff
@@ -168,3 +181,20 @@ void _TestPixelFormat(const CGRect bounds, const CGBitmapInfo info, DWORD expect
...
+    EXPECT_EQ(CGBitmapContextGetBitsPerPixel(context), 32);
+    EXPECT_EQ(CGBitmapContextGetWidth(context), 120);
+    EXPECT_EQ(CGBitmapContextGetHeight(context), 20);
+
+    EXPECT_EQ(CGBitmapContextGetBytesPerRow(context), 480);
+    EXPECT_NE(CGBitmapContextGetData(context), nullptr);
+
+    CGColorSpaceRelease(rgbColorSpace);
+    CGContextRelease(context);
+}
```
DHowett-MSFT: I might suggest having a test that draws a red rectangle into a 1x1 PRGBA context and checks that the buffer contains `0xff, 0x00, 0x00, 0xff`.

Additionally, you might want to add a test that goes through the image pipeline (Context -> Render -> Create Image -> Read Pixel); that would have caught the GetResolution issue.

Frameworks/include/CGIWICBitmap.h
```diff
@@ -106,10 +115,35 @@ class CGIWICBitmap
...
+            m_dataBuffer = static_cast<BYTE*>(data);
+        } else {
+            m_dataBuffer = new BYTE[m_height * m_bytesPerRow];
+            m_freeData = true;
+        }
     }
 
     ~CGIWICBitmap() {
+        if (m_freeData) {
+            delete m_dataBuffer;
```
msft-Jeyaram: >delete m_dataBuffer; [](start = 12, length = 20)

oh god, delete[] m_dataBuffer

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1833,78 +1835,146 @@ inline void SetImage(CGImageRef image) {
 
 /**
  @Status Caveat
- @Notes Limited bitmap formats available. Decode, shouldInterpolate, intent parameters
- and some byte orders ignored.
- */
+ The following formats are supported GUID_WICPixelFormat32bppPBGRA, GUID_WICPixelFormat32bppPRGBA, GUID_WICPixelFormat32bppBGR and
+ GUID_WICPixelFormat8bppAlpha. Any other format will return a null value.
```
rajsesh-msft: @msft-Jeyaram this is not what I meant.  We should specify what *CG* bitmap formats are supported, not what WicBitmap formats.  Ex: 32bppPBGRA == ARGB and so on.

msft-Jeyaram:
>     RETURN_NULL_IF_FAILED(img->ImageSource()->CopyPixels(nullptr, stride, size, data.get()));

This should be __CGImageGetRawBytes()

---
Refers to: Frameworks/CoreGraphics/CGImage.mm:175 in 81c6ff3. [](commit_id = 81c6ff356933224d1c2e283556357b16dd83a6b3, deletion_comment = False)

bbowman:
> class CGIWICBitmapLock : public RuntimeClass<RuntimeClassFlags<RuntimeClassType::ClassicCom>, IAgileObject, FtmBase, IWICBitmapLock> {

Why are we implementing classes in a header?

---
Refers to: Frameworks/include/CGIWICBitmap.h:56 in 81c6ff3. [](commit_id = 81c6ff356933224d1c2e283556357b16dd83a6b3, deletion_comment = False)

bbowman:
> };

Deserves some big old comments I think.

---
Refers to: Frameworks/include/CGIWICBitmap.h:51 in 81c6ff3. [](commit_id = 81c6ff356933224d1c2e283556357b16dd83a6b3, deletion_comment = False)

Frameworks/include/CGIWICBitmap.h
```diff
@@ -60,8 +62,15 @@ class CGIWICBitmapLock : public RuntimeClass<RuntimeClassFlags<RuntimeClassType:
...
+        : m_dataBuffer(data), m_pixelFormat(pixelFormat), m_locked_rect(region), m_bytesPerRow(bytesPerRow) {
+        m_texture = nullptr;
+    }
+
     ~CGIWICBitmapLock() {
-        m_texture->Unlock();
+        if (m_texture) {
+            m_texture->Unlock();
+        }
         m_texture = nullptr;
```
bbowman: whats going on here? You have a raw* but aren't jacketing in an RAII appropriate construct. Ownership is confusing and certainly not thread safe as is.

Frameworks/include/CGIWICBitmap.h
```diff
@@ -40,6 +40,8 @@
 #import <D2d1.h>
 #include <COMIncludes_End.h>
 
+#import <memory.h>
+
 using namespace Microsoft::WRL;
```
bbowman: this is real bad to do in a header.

msft-Jeyaram: more of the old code. Agreed.

---
In reply to: [88099822](https://github.com/Microsoft/WinObjC/pull/1371#discussion_r88099822) [](ancestors = 88099822)

Frameworks/include/CGIWICBitmap.h
```diff
@@ -161,34 +225,43 @@ class CGIWICBitmap
...
+        m_dpiX = 96;
+        m_dpiY = 96;
+        m_pixelFormat = pixelFormat;
+        m_height = height;
+        m_width = width;
+        m_freeData = false;
+    }
+
     WICPixelFormatGUID m_pixelFormat;
     IDisplayTexture* m_texture;
```
bbowman: >IDisplayTexture [](start = 4, length = 15)

not shown but if this is a COM object (implied by IFoo name style) manual delete is a real, real bad idea.

bbowman:
:clock1:

msft-Jeyaram:
> };

It's pretty straight forward on where it's being used.
non the less, i'll indicate it's used to obtain the texture.

---
In reply to: [260747910](https://github.com/Microsoft/WinObjC/pull/1371#issuecomment-260747910) [](ancestors = 260747910)

---
Refers to: Frameworks/include/CGIWICBitmap.h:51 in 81c6ff3. [](commit_id = 81c6ff356933224d1c2e283556357b16dd83a6b3, deletion_comment = False)

msft-Jeyaram:
> class CGIWICBitmapLock : public RuntimeClass<RuntimeClassFlags<RuntimeClassType::ClassicCom>, IAgileObject, FtmBase, IWICBitmapLock> {

Ideally it should be in a cpp file, given the file manifested as a header,it'll be moved into a cpp file.
as for now it's not the scope of the CR.

---
In reply to: [260747790](https://github.com/Microsoft/WinObjC/pull/1371#issuecomment-260747790) [](ancestors = 260747790)

---
Refers to: Frameworks/include/CGIWICBitmap.h:56 in 81c6ff3. [](commit_id = 81c6ff356933224d1c2e283556357b16dd83a6b3, deletion_comment = False)

Frameworks/include/CGIWICBitmap.h
msft-Jeyaram: This is not supposed to be thread safe, the design of the old DisplayTexture is confusing enough, this has been made much better via Jared's change (hopefully it'll migrate here soon).
until then this is a temporary fix.

---
In reply to: [88099416](https://github.com/Microsoft/WinObjC/pull/1371#discussion_r88099416) [](ancestors = 88099416)

Frameworks/include/CGIWICBitmap.h
bbowman: would still prefer unique_ptr and a release if you don't need to free? 

---
In reply to: [87863221](https://github.com/Microsoft/WinObjC/pull/1371#discussion_r87863221) [](ancestors = 87863221,87861378)

Frameworks/include/CGIWICBitmap.h
msft-Jeyaram: Talked off line, this is ugly enough to stand out. This is a temp fix until jared's changes come here.
We have a task for that #1337

---
In reply to: [88100057](https://github.com/Microsoft/WinObjC/pull/1371#discussion_r88100057) [](ancestors = 88100057)

msft-Jeyaram: hmm technically this can be wrapped in unique_ptr for now.

---
In reply to: [88114838](https://github.com/Microsoft/WinObjC/pull/1371#discussion_r88114838) [](ancestors = 88114838,88100057)

Frameworks/include/CGIWICBitmap.h
msft-Jeyaram: at the end we are still using a flag so it's not 'unsafe' as it sounds given the requirements at hand.
Other than that, i'm wrapping everything that can be wrapped in RAII, including the databuffer.

---
In reply to: [88114811](https://github.com/Microsoft/WinObjC/pull/1371#discussion_r88114811) [](ancestors = 88114811,87863221,87861378)

msft-Jeyaram: The logic seems odd, given that naturally it's suggested that we allocate the data at most times.
Although RAII are great, I feel as if this is seems unconventional. 
If(free) {
 .release();
}

Also it makes the assignments make it look like, it could be free-ed.
I'm leaving it as is for now. 

---
In reply to: [88114811](https://github.com/Microsoft/WinObjC/pull/1371#discussion_r88114811) [](ancestors = 88114811,87863221,87861378)

DHowett-MSFT:
:shipit:

rajsesh-msft:
:shipit:

MSFTFox:
:shipit:

PR #1372, opened 2016-11-14
Fixing regression in our SDK due to recent UIKit.Xaml namespace colli…

jaredhms:
…sion fix.  We were copying the wrong .pri file into the SDK output for UIKit.Xaml.dll.  Validated the fix locally after rebuilding the entire SDK project.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1372)


yiyang-msft:
:shipit:

PR #1376, opened 2016-11-14
Match CT and UI descent to reference platform sign

### LINKED ISSUE #1333: CTFontGetDescent() incorrectly returns negative values
The [current implementation](https://github.com/Microsoft/WinObjC/blob/e7ae8b7b0362c440bc585acfbd06b8bc62264f71/Frameworks/CoreText/CTFont.mm#L582-L583) of CTFontGetDescent() returns a negative value.

While this doesn’t seem to be documented, we could only observe positive values on the reference platform, and using the absolute value on Windows gave us the correct visual result.

aballway:
CoreText descent is expected to be positive when the glyph extends below the baseline, whereas UIFont's descender is expected to be negative.  Now when using CoreText, lineheight := Ascent + Descent + Leading, and in UIKit lineheight := Ascent - Descent + Leading

Fixes #1333

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1376)

### DIFF
--- a/Frameworks/CoreGraphics/CGFont.mm
+++ b/Frameworks/CoreGraphics/CGFont.mm
@@ -235,7 +235,8 @@ int CGFontGetAscent(CGFontRef font) {
 */
 int CGFontGetDescent(CGFontRef font) {
     // Value for CGFont metrics are specified in 'glyph space units', which appear to be the same as DWrite's 'design units'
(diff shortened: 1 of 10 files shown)


ms-jihua:
can you check on the sign for CGFont descent as well? sorry for the trouble

ms-jihua:
:shipit:

PR #1378, opened 2016-11-14
Fixing two newly-found regressions with the compositor refactor.

jaredhms:
[News Test App: Main article title is cropped]
The TextBox within UIKit.Label needs to bind its width/height to that of its parent (the containing Grid).

[News Test App: Contents get obscured when scrolling quickly]
Fixed, along with some other pre-existing flip animation issues that existed on develop (the snapshot layer was often the wrong size - larger than the actual contents we're flipping to).  I also fixed another regression with the flip animation in ux/uxe  the center of the flip rotation has to be between 0 and 1, but we were setting it to width/height of the layer divided by 2 (attempting to rotate at the center).  This accidentally worked on develop due to the hard-coded 1x1 CALayerXaml size, but it was incorrect and led to funny fly-in behavior in ux/uxe.  Fixed by removing the calls altogether since the default PlaneProjection CenterOfRotationX/CenterOfRotationY is .5 (the center).

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1378)

### DIFF
--- a/Frameworks/UIKit/StarboardXaml/LayerAnimation.cpp
+++ b/Frameworks/UIKit/StarboardXaml/LayerAnimation.cpp
@@ -128,7 +128,7 @@ concurrency::task<Layer^> _SnapshotLayer(Layer^ layer) {
     else {
         RenderTargetBitmap^ snapshot = ref new RenderTargetBitmap();
         return concurrency::create_task(
-                snapshot->RenderAsync(layer, static_cast<int>(LayerCoordinator::GetVisualWidth(layer) * DisplayProperties::RawScreenScale()), 0))
+                snapshot->RenderAsync(layer))
             .then([snapshot, layer](concurrency::task<void> result) noexcept {
             try {
                 result.get();
@@ -143,12 +143,14 @@ concurrency::task<Layer^> _SnapshotLayer(Layer^ layer) {
 
             // Return a new 'copy' layer with the rendered content
             Layer^ newLayer = ref new Layer();
+            newLayer->Name = "_Temporary_CoreAnimation_SnapshotLayer";
             LayerCoordinator::InitializeFrameworkElement(newLayer);
 
             // Copy display properties from the old layer to the new layer
             LayerCoordinator::SetValue(newLayer, "opacity", LayerCoordinator::GetValue(layer, "opacity"));
             LayerCoordinator::SetValue(newLayer, "position", LayerCoordinator::GetValue(layer, "position"));
-            LayerCoordinator::SetValue(newLayer, "size", LayerCoordinator::GetValue(layer, "size"));
+            newLayer->Width = layer->Width;
+            newLayer->Height = layer->Height;
             LayerCoordinator::SetValue(newLayer, "anchorPoint", LayerCoordinator::GetValue(layer, "anchorPoint"));
 
             int width = snapshot->PixelWidth;
@@ -159,8 +161,8 @@ concurrency::task<Layer^> _SnapshotLayer(Layer^ layer) {
             LayerCoordinator::SetContent(
                 newLayer, 
                 snapshot,
-                static_cast<float>(width),
-                static_cast<float>(height), 
+                static_cast<float>(layer->Width),
+                static_cast<float>(layer->Height),
                 static_cast<float>(DisplayProperties::RawScreenScale() * dispInfo->RawPixelsPerViewPixel));
 
             // There seems to be a bug in Xaml where Render'd layers get sized to their visible content... sort of.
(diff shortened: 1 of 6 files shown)


yiyang-msft:
>         }

nit:existing. extra line

---
Refers to: Frameworks/UIKit/StarboardXaml/StoryboardManager.cpp:266 in 814a131. [](commit_id = 814a1313da09b6dbe123e54c39ab40193ecd4fde, deletion_comment = False)

Frameworks/UIKit/StarboardXaml/LayerAnimation.cpp
```diff
@@ -143,12 +143,14 @@ concurrency::task<Layer^> _SnapshotLayer(Layer^ layer) {
 
             // Return a new 'copy' layer with the rendered content
             Layer^ newLayer = ref new Layer();
+            newLayer->Name = "_Temporary_CoreAnimation_SnapshotLayer";
             LayerCoordinator::InitializeFrameworkElement(newLayer);
 
             // Copy display properties from the old layer to the new layer
             LayerCoordinator::SetValue(newLayer, "opacity", LayerCoordinator::GetValue(layer, "opacity"));
             LayerCoordinator::SetValue(newLayer, "position", LayerCoordinator::GetValue(layer, "position"));
-            LayerCoordinator::SetValue(newLayer, "size", LayerCoordinator::GetValue(layer, "size"));
```
yiyang-msft: >LayerCoordinator::SetValue(newLayer, "size", LayerCoordinator::GetValue(layer, "size")); [](start = 12, length = 88)

it wasn't obvious to me if this is not working or optimization?

yiyang-msft:
>             int height = snapshot->PixelHeight;

can delete?

---
Refers to: Frameworks/UIKit/StarboardXaml/LayerAnimation.cpp:157 in 814a131. [](commit_id = 814a1313da09b6dbe123e54c39ab40193ecd4fde, deletion_comment = False)

Frameworks/UIKit/StarboardXaml/LayerAnimation.cpp
```diff
@@ -143,12 +143,14 @@ concurrency::task<Layer^> _SnapshotLayer(Layer^ layer) {
...
+            newLayer->Width = layer->Width;
+            newLayer->Height = layer->Height;
```
yiyang-msft: do we even need this given we setContent using layer->Width/Height below?

Frameworks/UIKit/StarboardXaml/LayerAnimation.cpp
jaredhms: switched back to original implementation here

---
In reply to: [87917684](https://github.com/Microsoft/WinObjC/pull/1378#discussion_r87917684) [](ancestors = 87917684)

Frameworks/UIKit/StarboardXaml/StoryboardManager.cpp
```diff
@@ -148,6 +146,16 @@ void StoryboardManager::_CreateFlip(Layer^ layer, bool flipRight, bool invert, b
...
+            } else {
+                // TODO: Track down why we sometimes don't have a parent when the animation completes.
+                TraceWarning(TAG, L"Failed to remove flip animation from scene; didn't have a parent to remove from!");
+                layer->Visibility = Visibility::Collapsed;
+                LayerCoordinator::SetContent(
+                    layer,
+                    nullptr,
+                    static_cast<float>(0),
+                    static_cast<float>(0),
+                    static_cast<float>(0));
```
jaredhms: note: still doesn't seem to catch *all* cases; I think we need to do some cleanup when animations are stopped.  i'll follow up further tomorrow.

Frameworks/UIKit/StarboardXaml/StoryboardManager.cpp
```diff
@@ -137,23 +153,48 @@ void StoryboardManager::_CreateFlip(Layer^ layer, bool flipRight, bool invert, b
...
+                auto strongLayer = weakLayer.Resolve<Layer>();
+                if (strongLayer) {
+                    // Remove the snapshot layer from the scene and from our tracking vector
+                    strongThis->_RemoveSnapshotLayer(strongLayer);
+                    strongThis->_animatingSnapshots.erase(
+                        std::remove_if(
+                            std::begin(strongThis->_animatingSnapshots), 
+                            std::end(strongThis->_animatingSnapshots), 
+                            [strongLayer](Layer^ snapshotLayer) {
+                                return snapshotLayer = strongLayer;
```
yiyang-msft: >= [](start = 53, length = 1)

==?

yiyang-msft:
:shipit:

Issue #1212, opened 2016-10-24
CoreText post-merge cleanup


ms-jihua:
Some cleanup to make code more consistent after CoreText stabilizes and merges back to develop:
- DWriteWrapper/CoreText boundary is inconsistent across CoreText functions. Set a policy for what ownership belongs to CoreText and what ownership belongs to DWriteWrapper, and adjust current DWriteWrapper functions to adhere to that.
- Inconsistent throw/return null policies
  - eg: in CTTypesetterCreateLineWithOffset

PR #1386, opened 2016-11-15
Re-implement CGFontCreateWithDataProvider

### LINKED ISSUE #1202: Re-implement CGFontCreateWithDataProvider
Minor regression from develop, this API creates a font object from arbitrary binary data. Initial investigation suggests that supporting this with DWrite will require a custom (but thin) IDWriteFontFileStream and IDWriteFontFileLoader that we can pass to an IDWriteFactory.

ms-jihua:
- Fix improper retain count in an internal map of font properties

Fixes #1202

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1386)


Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -479,4 +479,164 @@ HRESULT _DWriteFontGetBoundingBoxesForGlyphs(
...
+    ComPtr<IDWriteFontFileLoader> dwriteDataLoader;
+    RETURN_IF_FAILED(MakeAndInitialize<DWriteFontBinaryDataLoader>(&dwriteDataLoader, dataProvider));
+
+    RETURN_IF_FAILED(dwriteFactory->RegisterFontFileLoader(dwriteDataLoader.Get()));
+
+    ComPtr<IDWriteFontFile> fontFile;
+
+    // First two params are not used (see DWriteFontBinaryDataLoader::CreateStreamFromKey())
+    // However, CreateCustomFontFileReference throws E_INVALIDARG if null is passed for the first parameter
+    // Pass in a pointer to an arbitrary, unused key, to get around this.
```
ms-jihua: >// Pass in a pointer to an arbitrary, unused key, to get around this. [](start = 4, length = 69)

Considered passing in the CGDataProviderRef as the key here, instead of at init-time, but decided to favor type-safety instead (passing in as key would require cast-to and back-from const void*)

tests/unittests/CoreGraphics/CGFontTests.mm
```diff
@@ -100,4 +100,51 @@
...
+    CGDataProviderRef dataProvider = CGDataProviderCreateWithURL((__bridge CFURLRef)testFileURL);
+    CFAutorelease(dataProvider);
+
+    CGFontRef font = CGFontCreateWithDataProvider(dataProvider);
+    CFAutorelease(font);
+
+    // Check some very basic properties to make sure that the right font was loaded
+    EXPECT_OBJCEQ((id)CFSTR("Microsoft NeoGothic"), (id)CFAutorelease(CGFontCopyFullName(font)));
+    EXPECT_EQ(2210, CGFontGetAscent(font));
+#if TARGET_OS_WIN32
```
ms-jihua: can get rid of this when #1376 goes in, so planning to merge this after that

DHowett-MSFT:
Is it safe for us to redistribute MSNeoGothic.ttf?

DHowett-MSFT:
Can you add it to LFS?

---
In reply to: [260755904](https://github.com/Microsoft/WinObjC/pull/1386#issuecomment-260755904) [](ancestors = 260755904)

tests/unittests/CoreGraphics/CGFontTests.mm
```diff
@@ -100,4 +100,51 @@
...
         EXPECT_EQ(expectedBoxes[i].size.height, boxes[i].size.height);
     }
+}
+
+TEST(CGFont, CreateWithDataProvider) {
+    // get test startup full path
+    wchar_t fullPath[_MAX_PATH];
+    size_t len = GetModuleFileNameW(NULL, fullPath, _MAX_PATH);
+
+// Window needs extra handling for the drive character, OSX has twice-as-wide wchars
```
rajsesh-msft: >Window [](start = 3, length = 6)

Not sure this comment is clear.  In theory we should be able to use stringByDeletingLastPathComponent on windows just as well.

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -479,4 +479,164 @@ HRESULT _DWriteFontGetBoundingBoxesForGlyphs(
...
+class DWriteFontBinaryDataStream : public RuntimeClass<RuntimeClassFlags<WinRtClassicComMix>, IDWriteFontFileStream> {
+protected:
+    InspectableClass(L"Windows.Bridge.DirectWrite", TrustLevel::BaseTrust);
+
+public:
+    DWriteFontBinaryDataStream() {
+    }
+
+    HRESULT RuntimeClassInitialize(CGDataProviderRef dataProvider) {
+        _data.reset(CGDataProviderCopyData(dataProvider));
```
rajsesh-msft: why copy? You only need the "bytes" so, why not keep a reference to dataProvider and get "bytes"?

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -479,4 +479,164 @@ HRESULT _DWriteFontGetBoundingBoxesForGlyphs(
...
+
+    HRESULT created = dwriteFactory->CreateFontFace(fontFaceType,
+                                                    1, // number of elements in fontFileArray
+                                                    fontFileArray,
+                                                    0, // Just use the first face
+                                                    DWRITE_FONT_SIMULATIONS_NONE,
+                                                    outFontFace);
+
+    // Always try to unregister the font file loader, even if creation failed above
+    RETURN_IF_FAILED(dwriteFactory->UnregisterFontFileLoader(dwriteDataLoader.Get()));
```
rajsesh-msft: nit: you should ignore this failure.

ms-jihua: I guess I sort of see your point

---
In reply to: [88108767](https://github.com/Microsoft/WinObjC/pull/1386#discussion_r88108767) [](ancestors = 88108767)

Frameworks/CoreGraphics/DWriteWrapper.mm
ms-jihua: ? Taking a best guess that you want me to [dataProvider bytes]:
CGDataProvider is only a subclass of NSData on our platform.

We should not [CGDataProvider bytes] as it's taking advantage of an internal implementation detail that's subject to change.
We cannot use CGDataProviderCallbacks for access to underlying buffer, as we have not implemented those APIs.

Therefore we use the public interface, CGDataProviderCopyData(), which copies to a CGDataRef

---
In reply to: [88108446](https://github.com/Microsoft/WinObjC/pull/1386#discussion_r88108446) [](ancestors = 88108446)

ms-jihua:
switched to a custom test font. Thank you @DHowett-MSFT.

---
In reply to: [260755927](https://github.com/Microsoft/WinObjC/pull/1386#issuecomment-260755927) [](ancestors = 260755927,260755904)

Frameworks/CoreGraphics/DWriteWrapper.mm
rajsesh-msft: >RETURN_IF_FAILED [](start = 4, length = 16)

You have to UnRegisterFontFileLoader in cases where CreateCustomFontFileReference or Analyze fail, so you may need to revisit your RegisterFontFileLoader() (instead use if SUCCEEDED()) there to do the other things.

rajsesh-msft:
:clock1:

DHowett-MSFT:
:shipit:

Frameworks/CoreGraphics/CGFont.mm
```diff
@@ -100,11 +100,17 @@ CGFontRef CGFontRetain(CGFontRef font) {
 }
 
 /**
- @Status Stub
+ @Status Interoperable
 */
 CGFontRef CGFontCreateWithDataProvider(CGDataProviderRef cgDataProvider) {
```
msft-Jeyaram: >cgDataProvider [](start = 57, length = 14)

invalid state check? what is the behavior on ref platform.

Frameworks/CoreGraphics/CGFont.mm
```diff
@@ -100,11 +100,17 @@ CGFontRef CGFontRetain(CGFontRef font) {
...
-    UNIMPLEMENTED();
-    return StubReturn();
+    size_t memSize = sizeof(struct __CGFont) - sizeof(CFRuntimeBase);
+    CGFontRef ret = static_cast<CGFontRef>(_CFRuntimeCreateInstance(kCFAllocatorDefault, CGFontGetTypeID(), memSize, NULL));
+    CFAutorelease(ret);
+    struct __CGFont* mutableRet = const_cast<struct __CGFont*>(ret);
```
msft-Jeyaram: at some point it would be a good idea to make this utilize CPPBase

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -479,4 +479,170 @@ HRESULT _DWriteFontGetBoundingBoxesForGlyphs(
...
+ */
+class DWriteFontBinaryDataStream : public RuntimeClass<RuntimeClassFlags<WinRtClassicComMix>, IDWriteFontFileStream> {
+protected:
+    InspectableClass(L"Windows.Bridge.DirectWrite", TrustLevel::BaseTrust);
+
+public:
+    DWriteFontBinaryDataStream() {
+    }
+
+    HRESULT RuntimeClassInitialize(CGDataProviderRef dataProvider) {
```
msft-Jeyaram: given our odd implementation of CGDataProvider, i'm curious about this.

---
In reply to: [88146706](https://github.com/Microsoft/WinObjC/pull/1386#discussion_r88146706) [](ancestors = 88146706)

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -479,4 +479,170 @@ HRESULT _DWriteFontGetBoundingBoxesForGlyphs(
...
+        }
+
+        if (fragmentStart) {
+            const uint8_t* underlyingBuffer = CFDataGetBytePtr(_data.get());
+            *fragmentStart = reinterpret_cast<const void*>(&(underlyingBuffer[fileOffset]));
+        }
+
+        if (fragmentContext) {
+            // Deliberately unused: this is meant to be passed to ReleaseFileFragment() below to free part of the data,
+            // but this stream frees all the underlying CFData at once
```
msft-Jeyaram: would be good to have the comment at the top of this function, so it's obvious for the reader/user

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -479,4 +479,170 @@ HRESULT _DWriteFontGetBoundingBoxesForGlyphs(
...
+        _dataProvider.reset(dataProvider);
+        return S_OK;
+    }
+
+    HRESULT STDMETHODCALLTYPE CreateStreamFromKey(_In_ const void* fontFileReferenceKey,
+                                                  uint32_t fontFileReferenceKeySize,
+                                                  _Out_ IDWriteFontFileStream** fontFileStream) {
+        if (!fontFileStream) {
+            return E_INVALIDARG;
+        }
```
msft-Jeyaram: RETURN_HR_IF_NULL consistent use in file.

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -479,4 +479,170 @@ HRESULT _DWriteFontGetBoundingBoxesForGlyphs(
...
+
+        // Ignore first two params, just return the same kind of stream always
```
msft-Jeyaram: >  // Ignore first two params, just return the same kind of stream always [](start = 6, length = 72)

indicate that at the top of this function, pretty much our Caveats

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -479,4 +479,170 @@ HRESULT _DWriteFontGetBoundingBoxesForGlyphs(
...
+        ComPtr<DWriteFontBinaryDataStream> ret;
+        RETURN_IF_FAILED(MakeAndInitialize<DWriteFontBinaryDataStream>(&ret, _dataProvider.get()));
+
+        *fontFileStream = ret.Detach();
+        return S_OK;
+    }
+
+private:
+    woc::unique_cf<CGDataProviderRef> _dataProvider;
+};
```
msft-Jeyaram: given this is an entity of it's own, It would be best if it's in a separate file, this is not so much of a wrapper.
I fear this file is going to just grow out of proportion and can easily get messy.

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -479,4 +479,170 @@ HRESULT _DWriteFontGetBoundingBoxesForGlyphs(
...
+                                                  _Out_ IDWriteFontFileStream** fontFileStream) {
+        if (!fontFileStream) {
+            return E_INVALIDARG;
+        }
+
+        // Ignore first two params, just return the same kind of stream always
+        ComPtr<DWriteFontBinaryDataStream> ret;
+        RETURN_IF_FAILED(MakeAndInitialize<DWriteFontBinaryDataStream>(&ret, _dataProvider.get()));
+
+        *fontFileStream = ret.Detach();
```
msft-Jeyaram: >*fontFileStream = ret.Detach(); [](start = 8, length = 31)

CopyTo

tests/unittests/CoreGraphics/CGFontTests.mm
```diff
@@ -106,4 +106,38 @@
...
+
+// OSX has twice-as-wide wchars
+#if TARGET_OS_WIN32
+    NSString* testFileFullPath = [NSString stringWithCharacters:(const unichar*)fullPath length:len + 1];
+#else
+    NSString* testFileFullPath = [NSString stringWithBytes:fullPath length:sizeof(wchar_t) * len encoding:WCHAR_ENCODING];
+#endif
+
+    // reconstruct fullpath for test artifact file. e.g., /Volumes/WinObjC/build/Tests/UnitTests/Foundation/OSX/data/NSFileManagerUT.txt
+    testFileFullPath = [[testFileFullPath stringByDeletingLastPathComponent] stringByAppendingString:@"/data/WinObjC-Regular.ttf"];
```
msft-Jeyaram: use getPathToFile() in TestUtils.
getPathToFile("/.....tif")

msft-Jeyaram: If that does not satisfy the usage, make this into a utility function (given it looks like a general sequence) so it can be reused in the future.

---
In reply to: [88152656](https://github.com/Microsoft/WinObjC/pull/1386#discussion_r88152656) [](ancestors = 88152656)

msft-Jeyaram:
:clock1:

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -479,4 +479,170 @@ HRESULT _DWriteFontGetBoundingBoxesForGlyphs(
...
+    HRESULT RuntimeClassInitialize(CGDataProviderRef dataProvider) {
+        _data.reset(CGDataProviderCopyData(dataProvider));
+
+        // Just use current time for _lastWriteTime
+        FILETIME fileTime;
+        SYSTEMTIME systemTime;
+
+        GetSystemTime(&systemTime);
+        // Returns 0 on failure, which is probably good enough for WinObjC purposes?
+        SystemTimeToFileTime(&systemTime, &fileTime);
```
msft-Jeyaram: use GetSystemTimeAsFileTime and pass in fileTime

msft-Jeyaram: >SystemTimeToFileTime [](start = 8, length = 20)

also preferably, if this fails, you can just set the time to 0 and leave, rather than running through the below code.

Frameworks/CoreGraphics/DWriteWrapper.mm
msft-Jeyaram: RETURN_HR_IF_NULL(E_INVALIDARG, fontFileStream);

Frameworks/CoreGraphics/CGFont.mm
ms-jihua: returns null

---
In reply to: [88145504](https://github.com/Microsoft/WinObjC/pull/1386#discussion_r88145504) [](ancestors = 88145504)

Frameworks/CoreGraphics/DWriteWrapper.mm
ms-jihua: I'll take a look at this for #1212. Won't fix for now.

---
In reply to: [88148206](https://github.com/Microsoft/WinObjC/pull/1386#discussion_r88148206) [](ancestors = 88148206)

Frameworks/CoreGraphics/DWriteWrapper.mm
ms-jihua: not sure if GetSystemTimeToFileTime has the same behavior. I realize that 0 will change to 0 in the below code anyway, but I find special-casing it explicitly to be a comparatively minor optimization that, for me, muddies up the code path.

---
In reply to: [88171264](https://github.com/Microsoft/WinObjC/pull/1386#discussion_r88171264) [](ancestors = 88171264)

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -479,4 +479,170 @@ HRESULT _DWriteFontGetBoundingBoxesForGlyphs(
...
-}
\ No newline at end of file
+}
+
+#pragma region CGFontCreateWithDataProvider helpers
+
+/**
+ * Custom implementation of IDWriteFontFileStream that implements Read in terms of an underlying CFDataRef
+ */
+class DWriteFontBinaryDataStream : public RuntimeClass<RuntimeClassFlags<WinRtClassicComMix>, IDWriteFontFileStream> {
```
ms-jihua: InspectableClass is only available to WinRT types

---
In reply to: [88170542](https://github.com/Microsoft/WinObjC/pull/1386#discussion_r88170542) [](ancestors = 88170542)

Frameworks/CoreGraphics/CGFont.mm
ms-jihua: Have concerns about losing Hash, Description, Equal functions from above. Need to talk to @DHowett-MSFT about those wrt CppBase. Noted for #1212

---
In reply to: [88146106](https://github.com/Microsoft/WinObjC/pull/1386#discussion_r88146106) [](ancestors = 88146106)

tests/unittests/CoreGraphics/CGFontTests.mm
ms-jihua: Bit hesitant to move TestUtils out of Foundation. Going to have a chat with everyone first.

---
In reply to: [88152992](https://github.com/Microsoft/WinObjC/pull/1386#discussion_r88152992) [](ancestors = 88152992,88152656)

ms-jihua: done but it ended up requiring a much broader change than I think you intended :(

---
In reply to: [88291272](https://github.com/Microsoft/WinObjC/pull/1386#discussion_r88291272) [](ancestors = 88291272,88152992,88152656)

Frameworks/CoreGraphics/DWriteWrapper.mm
msft-Jeyaram: in all honestly why do we need to wait until later, this is not old code, this code is new code being introduced in this change. Let's do it properly the first time, rather than postponing it.

---
In reply to: [88280909](https://github.com/Microsoft/WinObjC/pull/1386#discussion_r88280909) [](ancestors = 88280909,88148206)

msft-Jeyaram: When we introduce something we should try to do it right as much as possible rather than postponing it to a clean up.

---
In reply to: [88312848](https://github.com/Microsoft/WinObjC/pull/1386#discussion_r88312848) [](ancestors = 88312848,88280909,88148206)

Frameworks/CoreGraphics/DWriteWrapper.mm
msft-Jeyaram: i don't think this got moved :(

---
In reply to: [88147586](https://github.com/Microsoft/WinObjC/pull/1386#discussion_r88147586) [](ancestors = 88147586)

ms-jihua: I misread what you wrote originally, and added an extra comment at the top of the class instead. I think that's still fine.

---
In reply to: [88313379](https://github.com/Microsoft/WinObjC/pull/1386#discussion_r88313379) [](ancestors = 88313379,88147586)

tests/unittests/Foundation/NSFileManagerTests.mm
```diff
@@ -24,31 +24,10 @@
...
-    ASSERT_TRUE(wcscat_s(dir, _countof(dir), L"\\data\\") == 0);
-    ASSERT_TRUE(::_wmakepath_s(fullPath, _countof(fullPath), drive, dir, L"NSFileManagerUT", L".txt") == 0);
-
-    NSString* testFileFullPath = [NSString stringWithCharacters:(const unichar*)fullPath length:len + 1];
-#else
-    NSString* testFileFullPath = [NSString stringWithBytes:fullPath length:sizeof(wchar_t) * len encoding:WCHAR_ENCODING];
-
-    // reconstruct fullpath for test artifact file. e.g., /Volumes/WinObjC/build/Tests/UnitTests/Foundation/OSX/data/NSFileManagerUT.txt
-    testFileFullPath = [[testFileFullPath stringByDeletingLastPathComponent] stringByAppendingString:@"/data/NSFileManagerUT.txt"];
-#endif
```
msft-Jeyaram: :) glad the refactoring helped clean this up.
Given we had this duplicate code.

QQ: I assume this passed in OSX and Windows and not related to the bug you recently opened up?

Frameworks/CoreGraphics/DWriteWrapper.mm
ms-jihua: DWriteWrapper_CoreText is also over 900 lines, and has two classes like this.

I don't want to move just this into a new file right now, and ignore the other two, because that fractures our codebase along conflicting principles.
I also don't want to move all four classes into (a? two? four?) new files right now, because three people have already signed off on this PR, and it'd be easier to get them to look at/provide feedback on large-scale reorganizational changes in a new PR.
This pull request is also rapidly scope-creeping for a change that was supposed to implement one (!) function that was regressed during CoreText bringup.

As a general principle, I prefer more & smaller PRs, to fewer & larger ones. Large changes make it much more difficult to pinpoint what exactly went wrong, when something does. I understand that some things won't be done if they're not attached to larger pull requests. However, this is a case where I have an explicit issue for this. I feel extremely strongly that I would rather move the four classes to new classes (assuming that's what other people want!) in a new PR, rather than try to get them to look once more at a PR that they've already approved.

---
In reply to: [88313117](https://github.com/Microsoft/WinObjC/pull/1386#discussion_r88313117) [](ancestors = 88313117,88312848,88280909,88148206)

tests/unittests/Foundation/NSFileManagerTests.mm
ms-jihua: This did not pass in OSX, did pass in Windows. This did not pass in OSX on develop, so it's not due to a regression on my end. However, the possibility that my change could complicate/build on top of the previous regression is why I'm cautious about 'minor' cleanup work like this, considering it required touching 10+ files just to get it to build everywhere.

---
In reply to: [88314221](https://github.com/Microsoft/WinObjC/pull/1386#discussion_r88314221) [](ancestors = 88314221)

Frameworks/CoreGraphics/DWriteWrapper.mm
msft-Jeyaram: yep exactly, this file is getting out of hand. In the future if you were to introduce any new classes, it would be wise to evaluate the appropriate place, rather than putting them in the default place where everything else is.
Generally it would be appreciated if there were smaller PRs and that would be great. but as I said, if we do this correct the first time, it would be the best.
It would be fine to have it in the post-work PR, as long as that doesn't become a giant bubble.


---
In reply to: [88316782](https://github.com/Microsoft/WinObjC/pull/1386#discussion_r88316782) [](ancestors = 88316782,88313117,88312848,88280909,88148206)

msft-Jeyaram: other than that other changes looks fine.

---
In reply to: [88317726](https://github.com/Microsoft/WinObjC/pull/1386#discussion_r88317726) [](ancestors = 88317726,88316782,88313117,88312848,88280909,88148206)

msft-Jeyaram: I understand your frustration of trying to get this in, but  we all know how much fun it is to refactor code :D (especially if we introduced it). #nohardfeelings.

---
In reply to: [88318143](https://github.com/Microsoft/WinObjC/pull/1386#discussion_r88318143) [](ancestors = 88318143,88317726,88316782,88313117,88312848,88280909,88148206)

PR #1393, opened 2016-11-15
Show modal UI in a XAML popup

bviglietta:
When Islandwood is running in "library" mode (i.e., ObjC code running as a library inside a larger UWA), any UI presented to the top-level UIWindow will now appear inside a XAML Popup object. This improves upon the previous situation of the UI not appearing anywhere at all because there is no top-level UIWindow and Islandwood doesn't own any part of the viewable XAML hierarchy.

To do this, there's a new thing called _UIPopupViewController that is instantiated as a top-level view controller when Islandwood is initialized in library mode. Its job is just to hold on to that Popup object and make it visible whenever it detects that someone's trying to present top-level UI.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1393)


bviglietta:
@jaredh is added to the review. #Closed

bviglietta:

Frameworks/include/_UIPopupViewController.h
```diff
@@ -0,0 +1,22 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#pragma once
+
+#import <UIKit/UIViewController.h>
+
+@interface _UIPopupViewController : UIViewController
```
jaredhms: can you please add some comments to lightly document what this is used for?

Frameworks/UIKit/StarboardXaml/StarboardXaml.h
```diff
@@ -30,6 +38,8 @@ typedef enum {
     ActivationTypeFile = 5,
 } ActivationType;
 
+IInspectable* GetRootXamlElement();
```
jaredhms: >IInspectable* [](start = 0, length = 13)

nit: can we return a ComPtr<IInspectable> here?

jaredhms: Can we move this over to XamlCompositor.h/.cpp, and into the XamlCompositor namespace?

Frameworks/UIKit/_UIPopupViewController.mm
```diff
@@ -0,0 +1,71 @@
...
+}
+
+- (void)presentViewController:(UIViewController*)viewControllerToPresent animated:(BOOL)flag completion:(void (^)())completion {
+    _popup.isOpen = YES;
+    [super presentViewController:viewControllerToPresent animated:flag completion:completion];
+}
+
+- (void)dismissViewControllerAnimated:(BOOL)flag completion:(void (^)())completion {
+    _popup.isOpen = NO;
+    [super dismissViewControllerAnimated:flag completion:completion];
```
jaredhms: nit: we'll probably hide the popup before the animation completes.  should we consider adding a custom completion block that gets called when the animation completes, which hides the popup and also calls the completion that was passed into this function?

Frameworks/UIKit/StarboardXaml/ApplicationMain.mm
```diff
@@ -148,6 +154,15 @@ int ApplicationMainStart(const char* principalName,
 
     UIApplicationMainInit(principalClassName, delegateClassName, defaultOrientation, (int)activationType, activationArgument);
 
+    if (activationType == ActivationTypeLibrary) {
```
jaredhms: nit: a comment here would be nice, too

Frameworks/UIKit/_UIPopupViewController.mm
```diff
@@ -0,0 +1,71 @@
...
+@implementation _UIPopupViewController {
+    WUXCPPopup* _popup;
+    EventRegistrationToken _layoutUpdated;
+}
+
+- (void)loadView {
+    _popup = [WUXCPPopup make];
+    _popup.child = [WXFrameworkElement createWith:GetRootXamlElement()];
+
+    _UIPopupViewController __weak* weakSelf = self;
```
jaredhms: >_UIPopupViewController __weak* [](start = 4, length = 30)

nit: __weak _UIPopupViewController* ?

Frameworks/UIKit/_UIPopupViewController.mm
```diff
@@ -0,0 +1,71 @@
...
+    // Keep the popup's contents sized to fit the window
+    _layoutUpdated = [_popup addLayoutUpdatedEvent:^(RTObject* sender, RTObject* args) {
+        _UIPopupViewController* strongSelf = weakSelf;
+
+        if (strongSelf == nil) {
+            return;
+        }
+
+        WUXCPPopup* popup = strongSelf->_popup;
+        id popupChild = popup.child;
```
jaredhms: >id [](start = 8, length = 2)

nit: UIElement*?

jaredhms: err WUXCUIElement*

---
In reply to: [88145427](https://github.com/Microsoft/WinObjC/pull/1393#discussion_r88145427) [](ancestors = 88145427)

Frameworks/UIKit/StarboardXaml/ApplicationMain.mm
```diff
@@ -94,6 +96,10 @@ int ApplicationMainStart(const char* principalName,
     WOCDisplayMode* displayMode = [UIApplication displayMode];
     [displayMode _setWindowSize:CGSizeMake(windowWidth, windowHeight)];
 
+    if (activationType == ActivationTypeLibrary) {
+        [displayMode setDisplayPreset:WOCDisplayPresetNative];
```
bviglietta: That would require me to understand why it's necessary...

---
In reply to: [88144728](https://github.com/Microsoft/WinObjC/pull/1393#discussion_r88144728) [](ancestors = 88144728)

Frameworks/UIKit/_UIPopupViewController.mm
bviglietta: Sounds reasonable.

---
In reply to: [88144677](https://github.com/Microsoft/WinObjC/pull/1393#discussion_r88144677) [](ancestors = 88144677)

Frameworks/UIKit/StarboardXaml/StarboardXaml.cpp
```diff
@@ -270,12 +270,16 @@ void UIApplicationBackgroundActivated(BackgroundActivatedEventArgs^ args) {
...
     auto uiElem = ref new Xaml::Controls::Grid();
-    auto rootFrame = ref new Xaml::Controls::Frame();
-    rootFrame->Content = uiElem;
 
     XamlCompositor::Initialize(uiElem, activationType);
 
-    if (activationType != ActivationTypeLibrary) {
+    if (activationType == ActivationTypeLibrary) {
+        // In library mode, presented UI should completely cover whatever's behind it
+        uiElem->Background = ref new Xaml::Media::SolidColorBrush(Colors::White);
```
jaredhms: nit: I know I pushed for this change, but now I'm wondering if we want a transparent background to support ViewControllers animating in/out?  We can probably worry about this when/if it becomes an issue, but it's up to you.

jaredhms:
:shipit:

DHowett-MSFT:
:shipit:

Frameworks/UIKit/StarboardXaml/StarboardXaml.cpp
bviglietta: Maybe, but it could also look weird if the controls behind the popup are visible but not interactive. Maybe a dimming effect would be appropriate?

In any event, it's moot for the moment, since all we're using right now is the Safari view controller, which doesn't do animations and is completely opaque. For now I'll leave this as is.

---
In reply to: [88291699](https://github.com/Microsoft/WinObjC/pull/1393#discussion_r88291699) [](ancestors = 88291699)

PR #1398, opened 2016-11-16
Fix edge condition in lineFragmentRectForProposedRect

### LINKED ISSUE #1123: NSTextContainer lineFragmentRectForProposedRect remainingRect
lineFragmentRectForProposedRect: does not handle the edge case of multiple exclusion paths properly when computing the remainingRect.

### LINKED ISSUE #1143: CGD2D - _CGPathFitRect and CTFramesetterCreateFrame Broken
As a part of the work with CGPath some functions have broken since they try to access information inside of the CGPath which is being removed. These functions should be re-enabled or replaced as a part of the CGPath work.

aballway:
Handle case where multiple exclusion zones are on the same line and overlap.  Also removes ScanlineCalculator and _CGPathFitRect.

Fixes #1123
Fixes #1143 

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1398)


Frameworks/UIKit/NSTextContainer.mm
```diff
@@ -75,25 +119,69 @@ - (CGRect)lineFragmentRectForProposedRect:(CGRect)proposed
...
+
+            // TODO 1394: Convert to CGPathContainsPoint for CGD2D
+            // Find the leftmost and rightmost points of the exclusion zone intersecting the same horizontal region as the proposed rect
+            lineIntersection.origin.x = __GetXPositionIntersectingZone(lineIntersection, _exclusionContexts[i].get(), padding);
+            lineIntersection.size.width =
+                __GetXPositionIntersectingZone({ { lineIntersection.origin.x + lineIntersection.size.width, lineIntersection.origin.y },
+                                                 lineIntersection.size },
+                                               _exclusionContexts[i].get(),
+                                               -padding) -
+                lineIntersection.origin.x;
```
memontic-ms: Please help clang

rajsesh-msft:
How are you testing any of this works?

aballway:
Currently just through the text layout in WOCCatalog.  I'll add some unit tests :D

---
In reply to: [261048109](https://github.com/Microsoft/WinObjC/pull/1398#issuecomment-261048109) [](ancestors = 261048109)

build/CoreGraphics/dll/CoreGraphics.def
```diff
@@ -233,6 +233,7 @@ LIBRARY CoreGraphics
         CGContextSetDirty
         CGContextGetBacking
         CGContextDrawGlyphRun
+        CGContextIsPointInPath
```
aballway: It's an existing function, and will be removed with the CGD2D changes

---
In reply to: [88314147](https://github.com/Microsoft/WinObjC/pull/1398#discussion_r88314147) [](ancestors = 88314147)

Frameworks/UIKit/NSLayoutManager.mm
```diff
@@ -128,17 +127,16 @@ - (void)__layoutAllText {
...
+            woc::unique_cf<CGPathRef> path{ CGPathCreateWithRect(CGRectMake(0, 0, rect.size.width, lineHeight), nullptr) };
+            woc::unique_cf<CTFrameRef> frame{
+                CTFramesetterCreateFrame(framesetter.get(), { stringIndex, lastIndex - stringIndex }, path.get(), nullptr)
+            };
 
             // Create line to fit as much text as possible in given rect
-            CTLineRef fitLine = static_cast<CTLineRef>([[[static_cast<NSArray*>(CTFrameGetLines(frame)) firstObject] retain] autorelease]);
-            CFRelease(frame);
+            woc::unique_cf<CTLineRef> fitLine{ static_cast<CTLineRef>(
+                [[static_cast<NSArray*>(CTFrameGetLines(frame.get())) firstObject] retain]) };
```
msft-Jeyaram: in all honestly this looks gruesome to read. uh with all those static_casts :(

Frameworks/UIKit/NSTextContainer.mm
```diff
@@ -75,25 +119,69 @@ - (CGRect)lineFragmentRectForProposedRect:(CGRect)proposed
...
-    if (_exclusionPaths != nil) {
-        for (UIBezierPath* path in(NSArray*)_exclusionPaths) {
-            CGRect shapeRect = _CGPathFitRect(path.CGPath, ret, _size, self.lineFragmentPadding);
-            ret = CGRectIntersection(shapeRect, ret);
+    std::vector<CGRect> exclusionRectsForHorizontal{};
+    for (size_t i = 0; i < _exclusionPathBoundingRects.size(); ++i) {
+        // If our proposed area doesn't intersect the bounding box of the exclusion zone at all, no need to compare against it
+        if (proposed.origin.y < _exclusionPathBoundingRects[i].origin.y + _exclusionPathBoundingRects[i].size.height &&
+            proposed.origin.y + proposed.size.height > _exclusionPathBoundingRects[i].origin.y &&
+            proposed.origin.x < _exclusionPathBoundingRects[i].origin.x + _exclusionPathBoundingRects[i].size.width) {
```
msft-Jeyaram: parenthesis (which would make this so much easier to read)

Frameworks/UIKit/NSTextContainer.mm
```diff
@@ -75,25 +119,69 @@ - (CGRect)lineFragmentRectForProposedRect:(CGRect)proposed
...
+            CGRect lineIntersection{ { _exclusionPathBoundingRects[i].origin.x, proposed.origin.y },
+                                     { _exclusionPathBoundingRects[i].size.width, proposed.size.height } };
```
msft-Jeyaram: new line between, hard to read with everything clumped.

Frameworks/UIKit/NSTextContainer.mm
```diff
@@ -75,25 +119,69 @@ - (CGRect)lineFragmentRectForProposedRect:(CGRect)proposed
                             remainingRect:(CGRect*)remainingRect {
     CGRect totalRect = CGRectMake(0, 0, _size.width, _size.height);
     CGRect ret = CGRectIntersection(proposed, totalRect);
+    if ([_exclusionPaths count] == 0L || ret.size.width <= 0) {
+        return ret;
+    }
+
+    CGFloat padding = self.lineFragmentPadding;
```
msft-Jeyaram: > self.lineFragmentPadding; [](start = 21, length = 26)

you have a mix usage of accessing though self and direct access via _xxxxx
standardize.

msft-Jeyaram:
:clock1:

Frameworks/UIKit/NSTextContainer.mm
```diff
@@ -75,25 +119,69 @@ - (CGRect)lineFragmentRectForProposedRect:(CGRect)proposed
...
+    std::sort(exclusionRectsForHorizontal.begin(), exclusionRectsForHorizontal.end(), [](CGRect a, CGRect b) {
+        return a.origin.x < b.origin.x;
+    });
+    ret = __FirstPossibleRectForProposed(ret, exclusionRectsForHorizontal, padding, _size.width);
+
+    if (remainingRect) {
+        CGFloat newXPosition = ret.origin.x + ret.size.width + 2 * padding;
+        if (newXPosition < _size.width) {
+            *remainingRect =
+                __FirstPossibleRectForProposed(CGRectMake(newXPosition, ret.origin.y, _size.width - newXPosition, ret.size.height),
```
aballway: It's in UIKit, and is a private (should be static whoops) helper

Frameworks/UIKit/NSTextContainer.mm
```diff
@@ -75,25 +119,69 @@ - (CGRect)lineFragmentRectForProposedRect:(CGRect)proposed
...
+                                               exclusionRectsForHorizontal,
+                                               padding,
+                                               _size.width);
         }
     }
```
aballway: It's just two ifs... and I prefer this to early returns anyways

Frameworks/UIKit/NSTextContainer.mm
```diff
@@ -75,25 +119,69 @@ - (CGRect)lineFragmentRectForProposedRect:(CGRect)proposed
...
+    // Sort by X origin to prevent skipping possible areas not covered by an exclusion zone
+    std::sort(exclusionRectsForHorizontal.begin(), exclusionRectsForHorizontal.end(), [](CGRect a, CGRect b) {
+        return a.origin.x < b.origin.x;
+    });
+    ret = __FirstPossibleRectForProposed(ret, exclusionRectsForHorizontal, padding, _size.width);
+
+    if (remainingRect) {
+        CGFloat newXPosition = ret.origin.x + ret.size.width + 2 * padding;
+        if (newXPosition < _size.width) {
+            *remainingRect =
```
aballway: remainingRect is zeroed out in the beginning of the method, but looking at it again it makes more sense to do so here

Frameworks/UIKit/NSTextContainer.mm
msft-Jeyaram: yes it's two ifs and rather than going with preference, we should look at code readability.
Most of this code is already heavy and it would make sense reduce the clutter. 

---
In reply to: [88341543](https://github.com/Microsoft/WinObjC/pull/1398#discussion_r88341543) [](ancestors = 88341543)

aballway: I think that adding another return here would make it harder to parse, I prefer it in a sense that I find it easier to read and understand what the code is doing.

Frameworks/UIKit/NSTextContainer.mm
msft-Jeyaram: >__FirstPossibleRectForProposed [](start = 16, length = 30)

reading this makes me think that it's defined in here. given __ usually means very restricted internal usage.
More appropriate would be _xxFirstPoxxx

Frameworks/UIKit/NSTextContainer.mm
aballway: unfortunately clang format is really opinionated about there not being a newline here :/ Added a comment for readability

Frameworks/UIKit/NSTextContainer.mm
aballway: It is defined in this file

Frameworks/UIKit/NSLayoutManager.mm
```diff
@@ -128,15 +127,13 @@ - (void)__layoutAllText {
...
-            CGPathRelease(path);
+            woc::unique_cf<CGPathRef> path{ CGPathCreateWithRect(CGRectMake(0, 0, rect.size.width, lineHeight), nullptr) };
+            woc::unique_cf<CTFrameRef> frame{
+                CTFramesetterCreateFrame(framesetter.get(), { stringIndex, lastIndex - stringIndex }, path.get(), nullptr)
+            };
 
             // Create line to fit as much text as possible in given rect
-            CTLineRef fitLine = static_cast<CTLineRef>([[[static_cast<NSArray*>(CTFrameGetLines(frame)) firstObject] retain] autorelease]);
-            CFRelease(frame);
+            CTLineRef fitLine = (CTLineRef)CFArrayGetValueAtIndex(CTFrameGetLines(frame.get()), 0);
```
ms-jihua: >(CTLineRef) [](start = 32, length = 11)

would keep this static_cast still

Frameworks/UIKit/NSTextContainer.mm
```diff
@@ -62,38 +78,120 @@ - (NSArray*)exclusionPaths {
     return _exclusionPaths;
 }
 
+// Finds the first viable region for the proposed rect given the exclusion areas based upon x positions
+// Assumes writing direction is left to right, exclusionLines is sorted and preprocessed for y positions, and all inputs are valid
```
ms-jihua: >Assumes writing direction is left to right [](start = 3, length = 42)

What's the plan for eventually dealing with this? It's not clear to me atm whether this section will be removed/fixed once CGD2D comes in

aballway: This would not be resolved with CGD2D changes, it would need to be a separate task.

rajsesh-msft: Please create an issue for this.

Frameworks/UIKit/NSTextContainer.mm
```diff
@@ -62,38 +78,121 @@ - (NSArray*)exclusionPaths {
...
+        if (path.origin.x + path.size.width > proposed.origin.x) {
+            proposed.origin.x = (path.origin.x + path.size.width) + padding;
+            proposed.size.width = maxWidth - proposed.origin.x;
+        }
+    }
+
+    return proposed;
+}
+
+// Finds the first point at which the given rect intersects the given exclusionContext, moving by delta each iteration
```
ms-jihua: >moving by delta each iteration [](start = 88, length = 30)

is there a more efficient way to do this? this current implementation seems incredibly expensive.

Frameworks/UIKit/NSTextContainer.mm
```diff
@@ -17,12 +17,19 @@
...
 
+#include <vector>
+#include <algorithm>
+
+static constexpr CGFloat c_defaultPadding = 5.0f;
 @implementation NSTextContainer {
     CGSize _size;
     NSLayoutManager* _layoutManager;
     StrongId<NSArray> _exclusionPaths;
+    // TODO 1394: Remove for CGD2D
```
ms-jihua: nit: doesn't matter too much since these'll eventually be removed, but if the counts of these two vectors are expected to stay the same, it's good to put in an enforcement mechanism for that (ie: use one vector, or even an std::map)

I would just add a comment noting that assumption in this case, since we'll eventually remove this

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -888,182 +888,4 @@ CFTypeID CGPathGetTypeID() {
 bool CGPathIsRect(CGPathRef path, CGRect* rect) {
     UNIMPLEMENTED();
     return StubReturn();
-}
-
-class ScanlineCalculator {
```
aballway: Yea this was from the old implementation

Frameworks/UIKit/NSTextContainer.mm
aballway: Not that I'm aware, and this gives similar performance to previous implementation and reference platform

tests/unittests/UIKit/NSTextContainerTests.mm
```diff
@@ -0,0 +1,65 @@
...
+
+TEST(NSTextContainer, LineFragmentRectForProposedRect) {
+    NSTextContainer* textContainer = [[[NSTextContainer alloc] initWithSize:CGSizeMake(1000, FLT_MAX)] autorelease];
+    textContainer.exclusionPaths = @[
+        [UIBezierPath bezierPathWithRect:CGRectMake(100, 0, 100, 100)],
+        [UIBezierPath bezierPathWithRect:CGRectMake(300, 0, 200, 150)],
+        [UIBezierPath bezierPathWithRect:CGRectMake(700, 0, 150, 300)]
+    ];
+
+    CGRect remaining = CGRectInfinite;
```
ms-jihua: can you add some comments about your intentions with each of these test cases?

Frameworks/UIKit/NSTextContainer.mm
```diff
@@ -62,38 +78,121 @@ - (NSArray*)exclusionPaths {
...
-        for (UIBezierPath* path in(NSArray*)_exclusionPaths) {
-            CGRect shapeRect = _CGPathFitRect(path.CGPath, ret, _size, self.lineFragmentPadding);
-            ret = CGRectIntersection(shapeRect, ret);
+    std::vector<CGRect> exclusionRectsForHorizontal{};
+    for (size_t i = 0; i < _exclusionPathBoundingRects.size(); ++i) {
+        CGRect boundingRect = _exclusionPathBoundingRects[i];
+        // If our proposed area doesn't intersect the bounding box of the exclusion zone at all, no need to compare against it
+        if ((proposed.origin.y < boundingRect.origin.y + boundingRect.size.height) &&
+            (proposed.origin.y + proposed.size.height > boundingRect.origin.y) &&
+            (proposed.origin.x < boundingRect.origin.x + boundingRect.size.width)) {
```
ms-jihua: >proposed.origin.x < boundingRect.origin.x + boundingRect.size.width [](start = 13, length = 67)

what if proposed.origin.x + proposed.size.width > boundingRect.origin.x? (just wondering why there are two lines for y and only one for x)

Frameworks/UIKit/NSTextContainer.mm
ms-jihua: I mean like, even if you're not wanting to do line intersections or whatever (would probably be a lot faster though), you can, off the top of my head:
- not calculate rect.origin.x + delta three times
- do a binary search instead of a linear search

how long do you expect this function to stick around? is it worth investing into a faster implementation at this/a later time?

---
In reply to: [90086723](https://github.com/Microsoft/WinObjC/pull/1398#discussion_r90086723) [](ancestors = 90086723)

Frameworks/UIKit/NSTextContainer.mm
ms-jihua: CGRectIntersectsRect?

---
In reply to: [90091646](https://github.com/Microsoft/WinObjC/pull/1398#discussion_r90091646) [](ancestors = 90091646)

DHowett-MSFT:
:shipit:

PR #1401, opened 2016-11-16
Removing and old unused version of nuget.exe

pradipd:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1401)


pradipd:
A newer version is already checked in at build\nuget\nuget.exe.

pradipd:

pradipd:
I also checked to make sure were not using this in the build definitions.

---
In reply to: [261065025](https://github.com/Microsoft/WinObjC/pull/1401#issuecomment-261065025) [](ancestors = 261065025)

PR #1403, opened 2016-11-16
Updated objc2winmd binaries: Support marshalling NSError as PropertySet

mukhole:
This binary update supports marshalling NSError as PropertySet.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1403)


pradipd:
:shipit:

PR #1407, opened 2016-11-17
Fix UIButton layout issues when using titleLabel and imageView

davelamb:
lazy intrinsic content size.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1407)


oliversa-msft:
Don't forget to kick off the sanity CI builds.

oliversa-msft:
:shipit:

PR #1408, opened 2016-11-17
[XAMLCatalog] Custom Text Control sample update

davelamb:
Addresses out of bounds crash and fills out the textBounds to give CoreTextInputContext a hint
about where our text is rendered.

Also offsets the first character in the spiral so it's not clipped.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1408)


oliversa-msft:
:shipit:

PR #1409, opened 2016-11-17
Return false for tab stops in CTParagraphStyle

### LINKED ISSUE #1404: [CoreText] AV Crash in __CTParagraphStyle::SetValue with kCTParagraphStyleSpecifierTabStops specifier
WOCCatalog crashes with an AV on startup trying to set kCTParagraphStyleSpecifierTabStops. (CTParagraphStyle.mm@ln76)

Thanks AJ!

aballway:
CTParagraphStyleCreate would cause an issue when trying to set tab stops when none are present.  Since they are not supported just ignore any when creating for now.

Fixes #1404 

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1409)

### DIFF
--- a/Frameworks/CoreText/CTParagraphStyle.mm
+++ b/Frameworks/CoreText/CTParagraphStyle.mm
@@ -74,8 +74,8 @@ struct __CTParagraphStyle : CoreFoundation::CppBase<__CTParagraphStyle> {
             woc::unique_cf<CFTypeRef> newValue{};
             switch (specifier) {
                 case kCTParagraphStyleSpecifierTabStops:
-                    newValue.reset(CFArrayCreateCopy(nullptr, static_cast<CFArrayRef>(value)));
-                    break;
+                    // Tab Stops not supported
+                    return false;
                 case kCTParagraphStyleSpecifierAlignment:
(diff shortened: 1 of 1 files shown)


Frameworks/CoreText/CTParagraphStyle.mm
```diff
@@ -106,8 +106,8 @@ bool GetValue(CTParagraphStyleSpecifier specifier, size_t size, void* outValue)
         if (specifier <= kCTParagraphStyleSpecifierCount && outValue != nullptr && size == __CTParagraphStyleSizeTable[specifier]) {
             switch (specifier) {
                 case kCTParagraphStyleSpecifierTabStops:
```
aballway: I didn't think we supported tab stops before.

rajsesh-msft: okay.  If we have sample code that is using it, we should remove it.

---
In reply to: [88520384](https://github.com/Microsoft/WinObjC/pull/1409#discussion_r88520384) [](ancestors = 88520384)

rajsesh-msft:
:shipit:

PR #1411, opened 2016-11-17
Update objc2winmd: Support delegate callbacks as properties

mukhole:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1411)


pradipd:
:shipit:

mukhole:
Checking in this change without waiting for the builds as this binary does not affect the builds in any way.

PR #1415, opened 2016-11-18
This change fix two issues on UIActivityIndicatorView and one issue on WocCatalog

yiyang-msft:
UIActivityIndicatorView should not be UserInteractionable - which blocks events to be dispatched to UI beneath it and also it should stop animation when asked to.

    In sample app Woccolog - the size of one of cells is way too small for a large textview which causes the text rendering to be jumpy when that cell is scrolled into/out of the view.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1415)


Frameworks/UIKit/UIActivityIndicatorView.mm
```diff
@@ -221,9 +221,9 @@ - (void)stopAnimating {
         _isAnimating = NO;
     }
 
-    [_progressRing setIsActive:YES];
+    [_progressRing setIsActive:NO];
```
oliversa-msft: Is it possible that calling stopAnimating displays the spinner without the animation. Wouldn't your change hide it now?

samples/WOCCatalog/WOCCatalog/TextDisplayViewController.m~RF39dcea7.TMP
```diff
@@ -0,0 +1,730 @@
...
+// This code is licensed under the MIT License (MIT).
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
```
oliversa-msft: I am certain you don't want to check in this temp file :-)

oliversa-msft:
:shipit:

PR #1426, opened 2016-11-18
Updating our version to use semver1.0.0 scheme.

pradipd:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1426)

### DIFF
--- a/build/nuget/build.ps1
+++ b/build/nuget/build.ps1
@@ -21,7 +21,7 @@ if(!$Repo) {
 
 $version = "0.0.1"
 $timeStamp = get-date -format yyyyMMddHHmmss
-$fullVersion = "$version-$timeStamp.pr"
+$fullVersion = "$version-$timeStamp" + "pr"
 
 $nuget = [System.IO.Path]::Combine($scriptPath, "nuget.exe")
 


pradipd:

build/nuget/build.ps1
```diff
@@ -21,7 +21,7 @@ if(!$Repo) {
 
 $version = "0.0.1"
```
rajsesh-msft: This is not new change, but we are currently 0.0.2 according to build\winobjc.version.

bbowman: That is a different version number that I don't really understand. It is what is baked into vsimporter and xib2nib only at the moment. It is independent of the nuget version but we could later bring them in line. Remember that the package @pradipd  is working on here is strictly prototype for now. We will start over the versioning with the new packages I am working on.

pradipd: Like Brian mentioned.  this is prototyping.  
We are testing out the workflow using PM's as guinea pigs.

---
In reply to: [88757560](https://github.com/Microsoft/WinObjC/pull/1426#discussion_r88757560) [](ancestors = 88757560)

PR #1433, opened 2016-11-21
Fix forward declarations for NSArray and NSDictionary in UIPasteboard.h

asimonov-msft:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1433)

### DIFF
--- a/include/UIKit/UIPasteboard.h
+++ b/include/UIKit/UIPasteboard.h
@@ -20,7 +20,8 @@
 #import <Foundation/NSObject.h>
 
 @class NSString;
-@class NSArray;
+@class NSArray<ObjectType>;
+@class NSDictionary<KeyType, ObjectType>;
 @class NSData;
 @class NSIndexSet;
 @class UIImage;


DHowett-MSFT:
:shipit:

DHowett-MSFT:
I do wonder if it is better to simply include the necessary headers. After all, you _cannot actualy use_ UIPasteboard without having imported the rest of the foundation headers, no matter how hard you try.

pradipd:
:shipit:

oliversa-msft:
Agreed

---
In reply to: [262001903](https://github.com/Microsoft/WinObjC/pull/1433#issuecomment-262001903) [](ancestors = 262001903)

include/UIKit/UIPasteboard.h
```diff
@@ -20,7 +20,8 @@
 #import <Foundation/NSObject.h>
 
 @class NSString;
-@class NSArray;
+@class NSArray<ObjectType>;
```
yiyang-msft: >ObjectType [](start = 15, length = 10)

what is the difference between these two usage?

oliversa-msft:
:shipit:

yiyang-msft:
:shipit:

include/UIKit/UIPasteboard.h
asimonov-msft: NSArray<ObjectType>is how you forward-declare a generic type.

---
In reply to: [88955275](https://github.com/Microsoft/WinObjC/pull/1433#discussion_r88955275) [](ancestors = 88955275)

PR #1436, opened 2016-11-21
Fix stackoverflow from regression in NSNumberFormatter nilSymbol property implementation

ehren:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1436)


Frameworks/Foundation/NSNumberFormatter.mm
```diff
@@ -278,25 +279,17 @@ - (NSString*)currencySymbol {
  @Status Interoperable
 */
 - (void)setNilSymbol:(NSString*)symbol {
-    @synchronized(self) {
-        if (self.nilSymbol != symbol) {
-            [self.nilSymbol autorelease];
-            self.nilSymbol = [symbol copy];
-        }
-    }
+    _nilSymbol = symbol;
```
DHowett-MSFT: This should probably be `_nilSymbol.attach([symbol copy])` to uphold the property contract.

ehren: fixed that in rebase

Issue #1450, opened 2016-11-23
CGDataProvider is, but should not be, an NS/CFData.


DHowett-MSFT:
* CGDataProviderCopyData returns another CGDataProvider masquerading as a CFData.
* CGDataProvider is an Objective-C class in an otherwise C framework.
* CGDataProvider confounds the debugger, which knows it is an NSData (but not really, so it shows `0 bytes`)
* CGDataProvider cannot hash like a normal NSCFData, making it useless for operations requiring hashing.
* CGDataProvider does not support streams.
* CGDataProvider doesn't have a CFTypeID since it's not a CFType.

DHowett-MSFT:
Adding TODO in the merge from develop into CGD2D; UIPasteboard.mm:397.
No SHA1 to reference right now.

PR #1454, opened 2016-11-23
Add support for registering and unregistering fonts

### LINKED ISSUE #1022: Support CTFontManagerRegisterFontsForURL and CTFontManagerRegisterGraphicsFont with DWrite
This is a regression after our move to DWrite. Investigate on how to support this API
- [ ] Interoperable CTFontManagerRegisterFontsForURL
- [ ] Interoperable CTFontManagerRegisterGraphicsFont

### LINKED ISSUE #1250: TTF fonts in app bundles should be added to list of available fonts for the app

aballway:
Adds support to register and unregister fonts in CTFontManager.  This allows apps to use custom fonts by either registering them programmatically or by including them in the app's info.plist

Fixes #1022
Fixes #1250

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1454)

### DIFF
--- a/Frameworks/CoreGraphics/CGDataProvider.mm
+++ b/Frameworks/CoreGraphics/CGDataProvider.mm
@@ -75,11 +75,9 @@ CGDataProviderRef CGDataProviderCreateWithURL(CFURLRef url) {
  @Status Interoperable
 */
 CFDataRef CGDataProviderCopyData(CGDataProviderRef provider) {
-    void* data = (void*)[(NSData*)provider bytes];
-    DWORD size = [(NSData*)provider length];
-    id ret = [[CGDataProvider alloc] initWithBytes:data length:size];
-
-    return (CFDataRef)ret;
+    uint8_t* bytes = (uint8_t*)[(NSData*)provider bytes];
+    size_t length = [(NSData*)provider length];
+    return CFDataCreate(nullptr, bytes, length);
 }
 
 /**
--- a/Frameworks/CoreGraphics/CGFont.mm
+++ b/Frameworks/CoreGraphics/CGFont.mm
@@ -19,6 +19,7 @@
 
 #import <CFRuntime.h>
 #import <CoreGraphics/DWriteWrapper.h>
+#import <CoreGraphics/CGFontInternal.h>
 
 #include <COMIncludes.h>
 #import <DWrite_3.h>
@@ -39,6 +40,9 @@ static const wchar_t* g_logTag = L"CGFont";
 struct __CGFont {
     CFRuntimeBase _base;
     ComPtr<IDWriteFontFace> _dwriteFontFace;
+
+    // Contains a value when created via Data Provider, null o/w
+    woc::unique_cf<CFDataRef> _data;
 
     struct DWRITE_FONT_METRICS _metrics;
     bool _cachedMetrics; // Set to true when _metrics is init'd
@@ -110,7 +114,8 @@ CGFontRef CGFontCreateWithDataProvider(CGDataProviderRef cgDataProvider) {
     CFAutorelease(ret);
     struct __CGFont* mutableRet = const_cast<struct __CGFont*>(ret);
 
-    RETURN_NULL_IF_FAILED(_DWriteCreateFontFaceWithDataProvider(cgDataProvider, &mutableRet->_dwriteFontFace));
+    mutableRet->_data.reset(CGDataProviderCopyData(cgDataProvider));
+    RETURN_NULL_IF_FAILED(_DWriteCreateFontFaceWithData(mutableRet->_data.get(), &mutableRet->_dwriteFontFace));
 
     return static_cast<CGFontRef>(CFRetain(ret));
 }
@@ -367,4 +372,8 @@ int CGFontGetUnitsPerEm(CGFontRef font) {
 CFTypeID CGFontGetTypeID() {
     static CFTypeID __kCGFontTypeID = _CFRuntimeRegisterClass(&__CGFontClass);
     return __kCGFontTypeID;
+}
+
+CFDataRef _CGFontGetData(CGFontRef font) {
+    return font ? font->_data.get() : nullptr;
 }
--- a/Frameworks/CoreGraphics/CGFont.mm
+++ b/Frameworks/CoreGraphics/CGFont.mm
@@ -374,6 +374,9 @@ CFTypeID CGFontGetTypeID() {
     return __kCGFontTypeID;
 }
 
+// TODO 1450: Convert this and all references to CGDataProviderRef
+// Currently CGDataProviderRef is broken for the needs of CTFontManager
+// So to prevent potentially copying multiple times, save a reference to the data
 CFDataRef _CGFontGetData(CGFontRef font) {
     return font ? font->_data.get() : nullptr;
 }
--- a/Frameworks/CoreGraphics/DWriteWrapper.mm
+++ b/Frameworks/CoreGraphics/DWriteWrapper.mm
@@ -725,7 +725,7 @@ public:
 
         CFIndex count = CFArrayGetCount(fontDatas);
         for (CFIndex i = 0; i < count; ++i) {
-            CFDataRef data = (CFDataRef)CFArrayGetValueAtIndex(fontDatas, i);
+            CFDataRef data = static_cast<CFDataRef>(CFArrayGetValueAtIndex(fontDatas, i));
             if (data) {
                 if (CFSetContainsValue(m_fontDatasSet.get(), data)) {
                     __AppendErrorIfExists(outErrors, kCTFontManagerErrorAlreadyRegistered);
@@ -749,7 +749,7 @@ public:
 
         CFIndex count = CFArrayGetCount(fontDatas);
         for (CFIndex i = 0; i < count; ++i) {
-            CFDataRef data = (CFDataRef)CFArrayGetValueAtIndex(fontDatas, i);
+            CFDataRef data = static_cast<CFDataRef>(CFArrayGetValueAtIndex(fontDatas, i));
             if (data) {
                 if (CFSetContainsValue(m_fontDatasSet.get(), data)) {
                     CFSetRemoveValue(m_fontDatasSet.get(), data);
--- a/Frameworks/CoreGraphics/DWriteWrapper.mm
+++ b/Frameworks/CoreGraphics/DWriteWrapper.mm
@@ -540,8 +540,6 @@ HRESULT _DWriteFontGetBoundingBoxesForGlyphs(
     return ret;
 }
 
-#pragma region CGFontCreateWithDataProvider helpers
-
 /**
  * Custom implementation of IDWriteFontFileStream that implements Read in terms of an underlying CFDataRef
  *
@@ -650,7 +648,6 @@ private:
     woc::unique_cf<CFDataRef> _data;
 };
 
-#pragma region Font File Registration
 // Implementation of IDwriteFontFileEnumerator, which enumerates over internal data and returns a font file
 class DWriteFontFileEnumerator : public RuntimeClass<RuntimeClassFlags<WinRtClassicComMix>, IDWriteFontFileEnumerator> {
 protected:
@@ -668,31 +665,34 @@ public:
     }
 
     HRESULT STDMETHODCALLTYPE GetCurrentFontFile(_Out_ IDWriteFontFile** fontFile) {
-        if (location < m_previouslyCreatedFiles->size()) {
-            *fontFile = m_previouslyCreatedFiles->at(location).Get();
+        RETURN_HR_IF(E_ILLEGAL_METHOD_CALL, m_location < 0 || m_location > CFArrayGetCount(m_fontDatas.get()));
+
+        if (0 <= m_location && m_location < m_previouslyCreatedFiles->size()) {
+            *fontFile = m_previouslyCreatedFiles->at(m_location).Get();
         } else {
             ComPtr<IDWriteFactory> dwriteFactory;
             RETURN_IF_FAILED(DWriteCreateFactory(DWRITE_FACTORY_TYPE_SHARED, __uuidof(IDWriteFactory), &dwriteFactory));
 
             ComPtr<DWriteFontBinaryDataLoader> loader;
             RETURN_IF_FAILED(
-                MakeAndInitialize<DWriteFontBinaryDataLoader>(&loader, (CFDataRef)CFArrayGetValueAtIndex(m_fontDatas.get(), location)));
+                MakeAndInitialize<DWriteFontBinaryDataLoader>(&loader, (CFDataRef)CFArrayGetValueAtIndex(m_fontDatas.get(), m_location)));
             RETURN_IF_FAILED(dwriteFactory->RegisterFontFileLoader(loader.Get()));
 
             int unused;
-            RETURN_IF_FAILED(dwriteFactory->CreateCustomFontFileReference(&unused, sizeof(int), loader.Get(), fontFile));
+            RETURN_IF_FAILED(dwriteFactory->CreateCustomFontFileReference(&unused, sizeof(unused), loader.Get(), fontFile));
             m_previouslyCreatedFiles->emplace_back(*fontFile);
         }
+
         return S_OK;
     }
 
     HRESULT STDMETHODCALLTYPE MoveNext(_Out_ BOOL* hasCurrentFile) {
-        *hasCurrentFile = ++location < CFArrayGetCount(m_fontDatas.get());
+        *hasCurrentFile = ++m_location < CFArrayGetCount(m_fontDatas.get());
         return S_OK;
     }
 
 private:
-    int location = -1;
+    int m_location = -1;
 
     woc::unique_cf<CFArrayRef> m_fontDatas;
     std::vector<ComPtr<IDWriteFontFile>>* m_previouslyCreatedFiles;
@@ -857,8 +857,6 @@ HRESULT _DWriteUnregisterFontsWithDatas(CFArrayRef datas, CFArrayRef* errors) {
     return __DWriteUpdateUserCreatedFontCollection(datas, errors, std::mem_fn(&DWriteFontCollectionLoader::RemoveDatas));
 }
 
-#pragma endregion
-
 /**
  * Creates an IDWriteFontFace by attempting to use a CGDataProviderRef as a font file
  */
@@ -912,5 +910,3 @@ HRESULT _DWriteCreateFontFaceWithData(CFDataRef data, IDWriteFontFace** outFontF
 
     return S_OK;
 }
-
-#pragma endregion
--- a/Frameworks/CoreGraphics/DWriteWrapper.mm
+++ b/Frameworks/CoreGraphics/DWriteWrapper.mm
@@ -24,6 +24,8 @@
 #import <unordered_map>
 #import <vector>
 #import <functional>
+#import <mutex>
+#import <type_traits>
 
 using namespace std;
 using namespace Microsoft::WRL;
@@ -32,7 +34,8 @@ static const wchar_t* TAG = L"_DWriteWrapper";
 static const wchar_t* c_defaultUserLanguage = L"en-us";
 
 // Static collection which holds all fonts registered by user programatically, through plist, etc.
-static ComPtr<IDWriteFontCollection> __UserCreatedFontCollection;
+static ComPtr<IDWriteFontCollection> _userCreatedFontCollection;
+static std::mutex _userCreatedFontCollectionMutex;
 
 /**
  * Private helper method to return the user set default locale string.
@@ -57,15 +60,39 @@ static CFStringRef __CFStringCreateUppercaseCopy(CFStringRef string, CFLocaleRef
 }
 
 /**
+ * Private helper that creates an array of CFString family names from a font collection
+ */
+CFArrayRef __DWriteCopyFontFamilyNamesFromCollection(IDWriteFontCollection* fontCollection) {
+    // Get the number of font families in the collection.
+    uint32_t count = fontCollection->GetFontFamilyCount();
+    woc::unique_cf<CFMutableArrayRef> fontFamilyNames(CFArrayCreateMutable(kCFAllocatorSystemDefault, count, &kCFTypeArrayCallBacks));
+    for (uint32_t i = 0; i < count; ++i) {
+        // Get the font family.
+        ComPtr<IDWriteFontFamily> fontFamily;
+        RETURN_NULL_IF_FAILED(fontCollection->GetFontFamily(i, &fontFamily));
+
+        // Get a list of localized strings for the family name.
+        ComPtr<IDWriteLocalizedStrings> familyNames;
+        RETURN_NULL_IF_FAILED(fontFamily->GetFamilyNames(&familyNames));
+
+        CFStringRef name = _CFStringFromLocalizedString(familyNames.Get());
+        if (CFStringGetLength(name) == 0) {
+            TraceError(TAG, L"Failed to convert the localized string to wide string.");
+            continue;
+        }
+
+        CFArrayAppendValue(fontFamilyNames.get(), name);
+    }
+
+    return fontFamilyNames.release();
+}
+
+/**
  * Private helper that creates the IDWriteFontList for a given font family name
  */
-HRESULT __DWriteGetFontListForFamilyName(CFStringRef familyName, IDWriteFontList** outFontList) {
+HRESULT __DWriteGetFontListForFamilyName(CFStringRef familyName, IDWriteFontCollection* fontCollection, IDWriteFontList** outFontList) {
     ComPtr<IDWriteFactory> dwriteFactory;
     RETURN_IF_FAILED(DWriteCreateFactory(DWRITE_FACTORY_TYPE_SHARED, __uuidof(IDWriteFactory), &dwriteFactory));
-
-    // Get the system font collection.
-    ComPtr<IDWriteFontCollection> fontCollection;
-    RETURN_IF_FAILED(dwriteFactory->GetSystemFontCollection(&fontCollection));
 
     // Get the font family.
     CFIndex familyNameLength = CFStringGetLength(familyName);
@@ -76,23 +103,12 @@ HRESULT __DWriteGetFontListForFamilyName(CFStringRef familyName, IDWriteFontList
     BOOL exists = FALSE;
 
     RETURN_IF_FAILED(fontCollection->FindFamilyName(reinterpret_cast<wchar_t*>(unicharFamilyName.data()), &index, &exists));
-    if (exists) {
-        ComPtr<IDWriteFontFamily> fontFamily;
-        RETURN_IF_FAILED(fontCollection->GetFontFamily(index, &fontFamily));
 
-        RETURN_IF_FAILED(
-            fontFamily->GetMatchingFonts(DWRITE_FONT_WEIGHT_NORMAL, DWRITE_FONT_STRETCH_NORMAL, DWRITE_FONT_STYLE_NORMAL, outFontList));
-
-        return S_OK;
-    }
-
-    RETURN_HR_IF(E_INVALIDARG, !__UserCreatedFontCollection);
-
-    RETURN_IF_FAILED(__UserCreatedFontCollection->FindFamilyName(reinterpret_cast<wchar_t*>(unicharFamilyName.data()), &index, &exists));
+    // Trying to create with a nonexistent font
     RETURN_HR_IF(E_INVALIDARG, !exists);
 
     ComPtr<IDWriteFontFamily> fontFamily;
-    RETURN_IF_FAILED(__UserCreatedFontCollection->GetFontFamily(index, &fontFamily));
+    RETURN_IF_FAILED(fontCollection->GetFontFamily(index, &fontFamily));
 
     RETURN_IF_FAILED(
         fontFamily->GetMatchingFonts(DWRITE_FONT_WEIGHT_NORMAL, DWRITE_FONT_STRETCH_NORMAL, DWRITE_FONT_STYLE_NORMAL, outFontList));
@@ -113,18 +129,25 @@ struct __CFStringHashEqual {
     }
 };
 
-typedef std::unordered_map<woc::unique_cf<CFStringRef>, std::shared_ptr<_DWriteFontProperties>, __CFStringHashEqual, __CFStringHashEqual>
-    __DWritePropertiesMap;
+/**
+ * Local typedef for map from font name to font properties
+ */
+using __DWritePropertiesMap =
+    std::unordered_map<woc::unique_cf<CFStringRef>, std::shared_ptr<_DWriteFontProperties>, __CFStringHashEqual, __CFStringHashEqual>;
 
-static __DWritePropertiesMap __CreatePropertiesMapForFamilyNames(CFArrayRef familyNames) {
+/**
+ * Private helper to create a map of font names to properties for all fonts in a collection
+ */
+static __DWritePropertiesMap __CreatePropertiesMapForFontCollection(IDWriteFontCollection* fontCollection) {
     __DWritePropertiesMap ret;
     woc::unique_cf<CFLocaleRef> locale(CFLocaleCopyCurrent());
-    CFIndex familyCount = CFArrayGetCount(familyNames);
+    woc::unique_cf<CFArrayRef> familyNames{ __DWriteCopyFontFamilyNamesFromCollection(fontCollection) };
+    CFIndex familyCount = CFArrayGetCount(familyNames.get());
     for (size_t i = 0; i < familyCount; ++i) {
-        CFStringRef familyName = static_cast<CFStringRef>(CFArrayGetValueAtIndex(familyNames, i));
+        CFStringRef familyName = static_cast<CFStringRef>(CFArrayGetValueAtIndex(familyNames.get(), i));
         ComPtr<IDWriteFontList> fontList;
 
-        if (FAILED(__DWriteGetFontListForFamilyName(familyName, &fontList))) {
+        if (FAILED(__DWriteGetFontListForFamilyName(familyName, fontCollection, &fontList))) {
             continue;
         }
 
@@ -166,20 +189,10 @@ static __DWritePropertiesMap __CreatePropertiesMapForFamilyNames(CFArrayRef fami
 }
 
 /**
- * Private static map, that maps uppercase display name and postscript name, to a _DWriteFontProperties struct for the corresponding system
- * font.
+ * Private static map, that maps display name and postscript name to a _DWriteFontProperties struct for the corresponding user defined font.
+ * Kept separate from c_systemFontPropertiesMap because this needs to be edited whenever a font is (un)registered
  */
-static const __DWritePropertiesMap c_systemFontPropertiesMap = []() {
-    woc::unique_cf<CFArrayRef> familyNames(_DWriteCopyFontFamilyNames());
-    return __CreatePropertiesMapForFamilyNames(familyNames.get());
-}();
-
-/**
- * Private static map, that maps uppercase display name and postscript name, to a _DWriteFontProperties struct for the corresponding user
- * defined font.
- * Kept separate from c_systemFontPropertiesMap because this cannot be constant
- */
-static __DWritePropertiesMap __UserFontPropertiesMap;
+static __DWritePropertiesMap _userFontPropertiesMap;
 
 /**
  * Helper method to convert IDWriteLocalizedStrings object to CFString object.
@@ -224,33 +237,6 @@ CFStringRef _CFStringFromLocalizedString(IDWriteLocalizedStrings* localizedStrin
         CFStringCreateWithCharacters(kCFAllocatorSystemDefault, reinterpret_cast<UniChar*>(wcharString.data()), length));
 }
 
-CFArrayRef __DWriteCopyFontFamilyNamesFromCollection(const ComPtr<IDWriteFontCollection>& fontCollection) {
-    woc::unique_cf<CFMutableArrayRef> fontFamilyNames(CFArrayCreateMutable(kCFAllocatorSystemDefault, 0, &kCFTypeArrayCallBacks));
-    // Get the number of font families in the collection.
-    uint32_t count = 0;
-    count = fontCollection->GetFontFamilyCount();
-
-    for (uint32_t i = 0; i < count; ++i) {
-        // Get the font family.
-        ComPtr<IDWriteFontFamily> fontFamily;
-        RETURN_NULL_IF_FAILED(fontCollection->GetFontFamily(i, &fontFamily));
-
-        // Get a list of localized strings for the family name.
-        ComPtr<IDWriteLocalizedStrings> familyNames;
-        RETURN_NULL_IF_FAILED(fontFamily->GetFamilyNames(&familyNames));
-
-        CFStringRef name = _CFStringFromLocalizedString(familyNames.Get());
-        if (CFStringGetLength(name) == 0) {
-            TraceError(TAG, L"Failed to convert the localized string to wide string.");
-            continue;
-        }
-
-        CFArrayAppendValue(fontFamilyNames.get(), name);
-    }
-
-    return fontFamilyNames.release();
-}
-
 /**
  * Helper method to retrieve font family names installed in the system.
  *
@@ -264,9 +250,11 @@ CFArrayRef _DWriteCopyFontFamilyNames() {
     // Get the system font collection.
     ComPtr<IDWriteFontCollection> fontCollection;
     RETURN_NULL_IF_FAILED(dwriteFactory->GetSystemFontCollection(&fontCollection));
-    woc::unique_cf<CFMutableArrayRef> systemFamilyNames{ (CFMutableArrayRef)__DWriteCopyFontFamilyNamesFromCollection(fontCollection) };
-    if (__UserCreatedFontCollection) {
-        woc::unique_cf<CFArrayRef> userFamilyNames{ __DWriteCopyFontFamilyNamesFromCollection(__UserCreatedFontCollection) };
+    woc::unique_cf<CFMutableArrayRef> systemFamilyNames{ (CFMutableArrayRef)__DWriteCopyFontFamilyNamesFromCollection(
+        fontCollection.Get()) };
+    std::lock_guard<std::mutex> guard(_userCreatedFontCollectionMutex);
+    if (_userCreatedFontCollection) {
+        woc::unique_cf<CFArrayRef> userFamilyNames{ __DWriteCopyFontFamilyNamesFromCollection(_userCreatedFontCollection.Get()) };
         CFArrayAppendArray(systemFamilyNames.get(), userFamilyNames.get(), { 0, CFArrayGetCount(userFamilyNames.get()) });
     }
 
@@ -279,8 +267,16 @@ CFArrayRef _DWriteCopyFontFamilyNames() {
 CFArrayRef _DWriteCopyFontNamesForFamilyName(CFStringRef familyName) {
     woc::unique_cf<CFMutableArrayRef> fontNames(CFArrayCreateMutable(kCFAllocatorSystemDefault, 0, &kCFTypeArrayCallBacks));
 
+    ComPtr<IDWriteFactory> dwriteFactory;
+    RETURN_NULL_IF_FAILED(DWriteCreateFactory(DWRITE_FACTORY_TYPE_SHARED, __uuidof(IDWriteFactory), &dwriteFactory));
+    ComPtr<IDWriteFontCollection> systemFontCollection;
+    RETURN_NULL_IF_FAILED(dwriteFactory->GetSystemFontCollection(&systemFontCollection));
+
     ComPtr<IDWriteFontList> fontList;
-    RETURN_NULL_IF_FAILED(__DWriteGetFontListForFamilyName(familyName, &fontList));
+    if (FAILED(__DWriteGetFontListForFamilyName(familyName, systemFontCollection.Get(), &fontList))) {
+        std::lock_guard<std::mutex> guard(_userCreatedFontCollectionMutex);
+        RETURN_NULL_IF_FAILED(__DWriteGetFontListForFamilyName(familyName, _userCreatedFontCollection.Get(), &fontList));
+    }
 
     size_t count = fontList->GetFontCount();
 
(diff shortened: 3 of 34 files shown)


Frameworks/CoreGraphics/CGDataProvider.mm
```diff
@@ -75,11 +75,9 @@ CGDataProviderRef CGDataProviderCreateWithURL(CFURLRef url) {
  @Status Interoperable
 */
 CFDataRef CGDataProviderCopyData(CGDataProviderRef provider) {
-    void* data = (void*)[(NSData*)provider bytes];
-    DWORD size = [(NSData*)provider length];
-    id ret = [[CGDataProvider alloc] initWithBytes:data length:size];
-
-    return (CFDataRef)ret;
+    uint8_t* bytes = (uint8_t*)[(NSData*)provider bytes];
```
msft-Jeyaram: static_cast?

build/Tests/UnitTests/CoreText/CoreText.UnitTests.vcxproj
```diff
@@ -242,7 +242,14 @@
     <ClangCompile Include="$(StarboardBasePath)\tests\unittests\CoreText\CTRunTests.mm" />
     <ClangCompile Include="$(StarboardBasePath)\tests\unittests\CoreText\CTTypesetterTests.mm" />
     <ClangCompile Include="$(StarboardBasePath)\tests\unittests\CoreText\CTFramesetterTests.mm" />
+    <ClangCompile Include="$(StarboardBasePath)\tests\unittests\CoreText\CTFontManagerTests.mm" />
   </ItemGroup>
+  <Target Name="CopyTestResourcesToOutput" AfterTargets="AfterBuild">
+    <ItemGroup>
+      <TestResourceFile Include="$(StarboardBasePath)\tests\unittests\CoreGraphics\data\*" />
```
DHowett-MSFT: >CoreGraphics [](start = 70, length = 12)

???

build/Headers/Headers.Shared/Internal.Shared.vcxitems
```diff
@@ -11,9 +11,6 @@
     </ClCompile>
   </ItemDefinitionGroup>
   <ItemGroup>
```
rajsesh-msft: why is this removed?

Frameworks/CoreText/CTFontManager.mm
```diff
@@ -34,53 +34,103 @@ CFArrayRef CTFontManagerCreateFontDescriptorsFromURL(CFURLRef fileURL) {
     return StubReturn();
 }
 
+// Helper functions to reduce code duplication, as (un)registering fonts have almost identical path
+// (Un)Registers font by creating temporary CFArray and calling into function for multiple fonts
+template <typename TLambda>
+static bool __CTFontManagerUpdateWithFont(CFURLRef fontURL, CTFontManagerScope scope, CFErrorRef _Nullable* error, const TLambda&& func) {
```
DHowett-MSFT: >const [](start = 115, length = 5)

curious choice; const rvalue?

Frameworks/CoreText/CTFontManager.mm
```diff
@@ -34,53 +34,103 @@ CFArrayRef CTFontManagerCreateFontDescriptorsFromURL(CFURLRef fileURL) {
...
+
+        CFRelease(errors);
+        return ret;
+    }
+
+    return func(fontURLs.get(), scope, nullptr);
+}
+
+// Converts CFURLs to CFDatas which are passed into DWriteWrapper methods
+template <typename TLambda>
```
DHowett-MSFT: >TLambda [](start = 19, length = 7)

Please add a comment documenting the typical parameter set and return type expected of a lambda

Frameworks/CoreText/CTFontManager.mm
```diff
@@ -34,53 +34,103 @@ CFArrayRef CTFontManagerCreateFontDescriptorsFromURL(CFURLRef fileURL) {
...
+static bool __CTFontManagerUpdateWithFonts(CFArrayRef fontURLs,
+                                           CTFontManagerScope scope,
+                                           CFArrayRef _Nullable* errors,
+                                           const TLambda&& func) {
+    CFIndex count = CFArrayGetCount(fontURLs);
+    woc::unique_cf<CFMutableArrayRef> fontDatas{ CFArrayCreateMutable(nullptr, 0, &kCFTypeArrayCallBacks) };
```
DHowett-MSFT: >0 [](start = 79, length = 1)

nit: save some array growths and reallocs by passing count as capacity.

Frameworks/UIKit/UIApplicationMain.mm
```diff
@@ -166,18 +166,15 @@ int UIApplicationMainInit(NSString* principalClassName,
...
-                NSString* path = [[NSBundle mainBundle] pathForResource:curFontName ofType:nil];
-                if (path != nil) {
-                    NSData* data = [NSData dataWithContentsOfFile:path];
-                    if (data != nil) {
-                        UIFont* font = [UIFont fontWithData:data];
-                        if (font != nil) {
-                            CTFontManagerRegisterGraphicsFont((CGFontRef)font, NULL);
-                        }
-                    }
+                NSURL* url = [[NSBundle mainBundle] URLForResource:curFontName withExtension:nil];
```
memontic-ms: >withExtension:nil [](start = 79, length = 17)

This should specify extension(s) that we support for fonts.

tests/testapps/CTCatalog/CTCatalog.vsimporter/CTCatalog-WinStore10/CTCatalog.vcxproj
```diff
@@ -233,7 +233,11 @@
       <PrecompiledHeader>Create</PrecompiledHeader>
     </ClCompile>
   </ItemGroup>
-  <ItemGroup />
+  <ItemGroup>
+    <Font Include="..\..\CTCatalog\Fonts\WinObjC-Bold.ttf" />
+    <Font Include="..\..\CTCatalog\Fonts\WinObjC-Italic.ttf" />
+    <Font Include="..\..\CTCatalog\Fonts\WinObjC.ttf" />
```
DHowett-MSFT: You seem to have added the fonts, but not committed to any testing strategy.

Can you add a visual/UI-driven test that renders text in the three winobjc fonts?

Do these fonts propagate to XAML controls? If so, can you set up a UIButton or UITextField that uses the winobjc font? That'd be cool.

Frameworks/UIKit/UIApplicationMain.mm
DHowett-MSFT: I don't think it can; curFontName is pulled from info.plist with its extension fully formed.

`... URLForResource:@"x.ttf" withExtension:@"ttf"` will look for a file x.ttf.ttf.

---
In reply to: [89406546](https://github.com/Microsoft/WinObjC/pull/1454#discussion_r89406546) [](ancestors = 89406546)

tests/unittests/CoreText/CTFontManagerTests.mm
```diff
@@ -0,0 +1,85 @@
...
+    EXPECT_OBJCEQ(@"WinObjC", familyName);
+
+    EXPECT_TRUE(CTFontManagerUnregisterFontsForURL((__bridge CFURLRef)testFileURL, kCTFontManagerScopeSession, &error));
+    EXPECT_EQ(nullptr, error);
+
+    woc::unique_cf<CTFontRef> failedFont{ CTFontCreateWithName(CFSTR("WinObjC"), 20, nullptr) };
+    EXPECT_NE(nullptr, failedFont);
+
+    familyName = (NSString*)CTFontCopyFullName(failedFont.get());
+    EXPECT_OBJCEQ(@"Segoe UI", familyName);
```
DHowett-MSFT: Does this hold on the reference platform?

Not Segoe exactly, but font failure leading to default font.

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -23,13 +23,17 @@
...
+#import <functional>
 
 using namespace std;
 using namespace Microsoft::WRL;
 
 static const wchar_t* TAG = L"_DWriteWrapper";
 static const wchar_t* c_defaultUserLanguage = L"en-us";
 
+// Static collection which holds all fonts registered by user programatically, through plist, etc.
+static ComPtr<IDWriteFontCollection> __UserCreatedFontCollection;
```
DHowett-MSFT: >__UserCreatedFontCollection [](start = 37, length = 27)

nit: `u`

DHowett-MSFT: or nit: `s_`

---
In reply to: [89407071](https://github.com/Microsoft/WinObjC/pull/1454#discussion_r89407071) [](ancestors = 89407071)

DHowett-MSFT: This is absolutely not created in a threadsafe manner. Magic statics?

---
In reply to: [89407080](https://github.com/Microsoft/WinObjC/pull/1454#discussion_r89407080) [](ancestors = 89407080,89407071)

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -72,10 +76,23 @@ HRESULT __DWriteGetFontListForFamilyName(CFStringRef familyName, IDWriteFontList
...
     RETURN_IF_FAILED(fontCollection->FindFamilyName(reinterpret_cast<wchar_t*>(unicharFamilyName.data()), &index, &exists));
+    if (exists) {
+        ComPtr<IDWriteFontFamily> fontFamily;
+        RETURN_IF_FAILED(fontCollection->GetFontFamily(index, &fontFamily));
+
+        RETURN_IF_FAILED(
+            fontFamily->GetMatchingFonts(DWRITE_FONT_WEIGHT_NORMAL, DWRITE_FONT_STRETCH_NORMAL, DWRITE_FONT_STYLE_NORMAL, outFontList));
+
+        return S_OK;
+    }
```
DHowett-MSFT: nit: early return

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -96,61 +113,73 @@ bool operator()(const woc::unique_cf<CFStringRef>& item1, const woc::unique_cf<C
     }
 };
 
-/**
- * Private static map, that maps uppercase display name and postscript name, to a _DWriteFontProperties struct for the corresponding font.
- */
-static std::unordered_map<woc::unique_cf<CFStringRef>, std::shared_ptr<_DWriteFontProperties>, __CFStringHashEqual, __CFStringHashEqual>
-    c_fontPropertiesMap = []() {
-        decltype(c_fontPropertiesMap) ret;
+typedef std::unordered_map<woc::unique_cf<CFStringRef>, std::shared_ptr<_DWriteFontProperties>, __CFStringHashEqual, __CFStringHashEqual>
```
DHowett-MSFT: >typedef [](start = 0, length = 7)

prefer `using x = y;` over `typedef y x;`

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -195,22 +224,8 @@ CFStringRef _CFStringFromLocalizedString(IDWriteLocalizedStrings* localizedStrin
...
 }
 
-/**
- * Helper method to retrieve font family names installed in the system.
- *
- * @return Immutable array of font family name strings that are installed in the system.
- */
-CFArrayRef _DWriteCopyFontFamilyNames() {
+CFArrayRef __DWriteCopyFontFamilyNamesFromCollection(const ComPtr<IDWriteFontCollection>& fontCollection) {
     woc::unique_cf<CFMutableArrayRef> fontFamilyNames(CFArrayCreateMutable(kCFAllocatorSystemDefault, 0, &kCFTypeArrayCallBacks));
```
DHowett-MSFT: Using a `const ComPtr<T>&` confers no benefit over using a `T*`, and disallows the use of this function as a C API facet.

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -566,9 +617,9 @@ void STDMETHODCALLTYPE ReleaseFileFragment(void* fragmentContext){
     DWriteFontBinaryDataLoader() {
     }
 
-    HRESULT RuntimeClassInitialize(CGDataProviderRef dataProvider) {
-        CFRetain(dataProvider);
-        _dataProvider.reset(dataProvider);
+    HRESULT RuntimeClassInitialize(CFDataRef data) {
+        CFRetain(data);
+        _data.reset(data);
```
DHowett-MSFT: fold?

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -580,29 +631,232 @@ HRESULT STDMETHODCALLTYPE CreateStreamFromKey(_In_ const void* fontFileReference
...
+    woc::unique_cf<CFArrayRef> userFamilyNames{ __DWriteCopyFontFamilyNamesFromCollection(__UserCreatedFontCollection) };
+    __UserFontPropertiesMap = __CreatePropertiesMapForFamilyNames(userFamilyNames.get());
+
+    return S_OK;
+}
+
+// Registers user defined fonts to a collection so they can be created later
+// Expects datas to be array of CFDataRefs, errors to be out pointer to array of CFErrorRef
+HRESULT _DWriteRegisterFontsWithDatas(CFArrayRef datas, CFArrayRef* errors) {
+    return __DWriteUpdateUserCreatedFontCollection(datas, errors, std::mem_fn(&DWriteFontCollectionLoader::AddDatas));
```
DHowett-MSFT: >DWriteFontCollectionLoader [](start = 79, length = 26)

This scares me a lot: You're implicitly calling a pointer-to-member-function across a function boundary with an instance created elsewhere.

Is there a better way to do this? Perhaps a lambda that consumes a `DWriteFontCollectionLoader*`?

DHowett-MSFT:
:clock1:

tests/testapps/CTCatalog/CTCatalog.vsimporter/CTCatalog-WinStore10/CTCatalog.vcxproj
aballway: These are tested by being in the list of available fonts in the font and alignment pages, and I'm unsure if these propogate to Xaml, will require investigation

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -96,61 +113,73 @@ bool operator()(const woc::unique_cf<CFStringRef>& item1, const woc::unique_cf<C
...
-                info->style = font->GetStyle();
-            }
-        }
+/**
+ * Private static map, that maps uppercase display name and postscript name, to a _DWriteFontProperties struct for the corresponding system
+ * font.
+ */
+static const __DWritePropertiesMap c_systemFontPropertiesMap = []() {
+    woc::unique_cf<CFArrayRef> familyNames(_DWriteCopyFontFamilyNames());
+    return __CreatePropertiesMapForFamilyNames(familyNames.get());
```
rajsesh-msft: Yes.  This can be a static inside a method and that wont be invoked during dll load time.  There are very limited things you can do in static initialization due to the loader locks. [https://msdn.microsoft.com/en-us/library/windows/desktop/dn633971(v=vs.85).aspx#general_best_practices](https://msdn.microsoft.com/en-us/library/windows/desktop/dn633971(v=vs.85).aspx#general_best_practices)

---
In reply to: [89407585](https://github.com/Microsoft/WinObjC/pull/1454#discussion_r89407585) [](ancestors = 89407585)

Frameworks/CoreGraphics/DWriteWrapper.mm
aballway: I normally dislike them but find it makes this codepath cleaner #WontFix

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -498,8 +548,9 @@ HRESULT _DWriteFontGetBoundingBoxesForGlyphs(
     DWriteFontBinaryDataStream() {
     }
 
-    HRESULT RuntimeClassInitialize(CGDataProviderRef dataProvider) {
-        _data.reset(CGDataProviderCopyData(dataProvider));
+    HRESULT RuntimeClassInitialize(CFDataRef data) {
+        CFRetain(data);
+        _data.reset(data);
```
aballway: would have to cast to CFDataRef to do so, which feels really silly #WontFix

Frameworks/CoreGraphics/DWriteWrapper.mm
DHowett-MSFT: Oh, sorry: I think I misinterpreted your response. Carry on!

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -72,10 +76,23 @@ HRESULT __DWriteGetFontListForFamilyName(CFStringRef familyName, IDWriteFontList
...
+        ComPtr<IDWriteFontFamily> fontFamily;
+        RETURN_IF_FAILED(fontCollection->GetFontFamily(index, &fontFamily));
+
+        RETURN_IF_FAILED(
+            fontFamily->GetMatchingFonts(DWRITE_FONT_WEIGHT_NORMAL, DWRITE_FONT_STRETCH_NORMAL, DWRITE_FONT_STYLE_NORMAL, outFontList));
+
+        return S_OK;
+    }
+
+    RETURN_HR_IF(E_INVALIDARG, !__UserCreatedFontCollection);
```
aballway: This failure represents trying to create a font with an invalid font family name, and will fail the same whether or not the user registered any custom fonts #ByDesign

Frameworks/CoreGraphics/DWriteWrapper.mm
DHowett-MSFT: The propagation of the loader into the functor will be explicit, rather than implicit. Right now there is no express contract between the caller and callee that a bound member function is in use.

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -580,29 +631,232 @@ HRESULT STDMETHODCALLTYPE CreateStreamFromKey(_In_ const void* fontFileReference
...
+    }(&loader);
+    RETURN_IF_FAILED(createdLoader);
+
+    // Call member function to update loader's font data
+    func(loader.Get(), datas, errors);
+
+    // DWrite won't create a new font collection unless we provide a new key each time
+    static size_t collectionKey = 0;
+    RETURN_IF_FAILED(
+        dwriteFactory->CreateCustomFontCollection(loader.Get(), &(++collectionKey), sizeof(collectionKey), &__UserCreatedFontCollection));
```
aballway: It's out of a ComPtr so that should magic the retain count no?

Frameworks/CoreGraphics/CGDataProvider.mm
aballway: I expect this will all be replaced in #1450 #WontFix

tests/unittests/CoreText/CTFontManagerTests.mm
aballway: Yep!

tests/testapps/CTCatalog/CTCatalog.vsimporter/CTCatalog-WinStore10/CTCatalog.vcxproj
aballway: Update: I cannot find a way to let Xaml use our custom font collection, will need further investigation to figure out if it is possible to do so

Frameworks/CoreGraphics/DWriteWrapper.mm
aballway: I see your point, but it feels silly to create a lambda just to do this, and given it's internal and documented I don't think it will pose an issue #WontFix

Frameworks/CoreGraphics/CGFont.mm
```diff
@@ -110,7 +114,8 @@ CGFontRef CGFontCreateWithDataProvider(CGDataProviderRef cgDataProvider) {
     CFAutorelease(ret);
     struct __CGFont* mutableRet = const_cast<struct __CGFont*>(ret);
 
-    RETURN_NULL_IF_FAILED(_DWriteCreateFontFaceWithDataProvider(cgDataProvider, &mutableRet->_dwriteFontFace));
+    mutableRet->_data.reset(CGDataProviderCopyData(cgDataProvider));
+    RETURN_NULL_IF_FAILED(_DWriteCreateFontFaceWithData(mutableRet->_data.get(), &mutableRet->_dwriteFontFace));
```
MSFTFox: Seems like this should be FAIL_FAST? Ignoring a dwrite error seems to go against the pattern of failing at the public api level that we've been following.

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -580,29 +640,238 @@ HRESULT STDMETHODCALLTYPE CreateStreamFromKey(_In_ const void* fontFileReference
...
+        CFMutableArrayRef outErrors = nil;
+        if (errors) {
+            outErrors = CFArrayCreateMutable(nullptr, 0, &kCFTypeArrayCallBacks);
+            *errors = outErrors;
+        }
+
+        CFIndex count = CFArrayGetCount(fontDatas);
+        for (CFIndex i = 0; i < count; ++i) {
+            CFDataRef data = (CFDataRef)CFArrayGetValueAtIndex(fontDatas, i);
+            if (data) {
```
msft-Jeyaram: Both if can be joined. If(data && CFSetContains...)
To be fair, the nested ifs are growing in length throughout. It would be good to see if things can be factored or structured differently

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -580,29 +640,238 @@ HRESULT STDMETHODCALLTYPE CreateStreamFromKey(_In_ const void* fontFileReference
...
+        CFRetain(fontDatas);
+        m_fontDatas.reset(fontDatas);
+        m_previouslyCreatedFiles = previouslyCreatedFiles;
+        return S_OK;
+    }
+
+    HRESULT STDMETHODCALLTYPE GetCurrentFontFile(_Out_ IDWriteFontFile** fontFile) {
+        if (location < m_previouslyCreatedFiles->size()) {
+            *fontFile = m_previouslyCreatedFiles->at(location).Get();
+        } else {
```
msft-Jeyaram: Don't need else, do an early return.

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -580,29 +640,238 @@ HRESULT STDMETHODCALLTYPE CreateStreamFromKey(_In_ const void* fontFileReference
...
+    }
+
+    HRESULT RuntimeClassInitialize(CFArrayRef fontDatas, std::vector<ComPtr<IDWriteFontFile>>* previouslyCreatedFiles) {
+        CFRetain(fontDatas);
+        m_fontDatas.reset(fontDatas);
+        m_previouslyCreatedFiles = previouslyCreatedFiles;
+        return S_OK;
+    }
+
+    HRESULT STDMETHODCALLTYPE GetCurrentFontFile(_Out_ IDWriteFontFile** fontFile) {
```
msft-Jeyaram: The default implementation has validity checks. Add them to be compatible

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -580,29 +640,238 @@ HRESULT STDMETHODCALLTYPE CreateStreamFromKey(_In_ const void* fontFileReference
...
+        return S_OK;
+    }
+
+    HRESULT STDMETHODCALLTYPE MoveNext(_Out_ BOOL* hasCurrentFile) {
+        *hasCurrentFile = ++location < CFArrayGetCount(m_fontDatas.get());
+        return S_OK;
+    }
+
+private:
+    int location = -1;
```
msft-Jeyaram: MLocation

msft-Jeyaram: Unsigned?

tests/unittests/CoreText/CTFontManagerTests.mm
```diff
@@ -0,0 +1,85 @@
...
+}
+
+TEST(CTFontManager, ShouldBeAbleToRegisterFontsForGraphicsFont) {
+    char fullPath[_MAX_PATH];
```
msft-Jeyaram: Re-computation on every test. Make it a utility

tests/unittests/CoreText/CTFontManagerTests.mm
```diff
@@ -0,0 +1,85 @@
...
+        fileURLWithPath:[[@(fullPath) stringByDeletingLastPathComponent] stringByAppendingPathComponent:@"/data/WinObjC-Italic.ttf"]];
+    woc::unique_cf<CGDataProviderRef> dataProvider{ CGDataProviderCreateWithURL((__bridge CFURLRef)testFileURL) };
+    woc::unique_cf<CGFontRef> graphicsFont{ CGFontCreateWithDataProvider(dataProvider.get()) };
+    EXPECT_NE(nullptr, graphicsFont);
+
+    CFErrorRef error = nullptr;
+    EXPECT_TRUE(CTFontManagerRegisterGraphicsFont(graphicsFont.get(), &error));
+    EXPECT_EQ(nullptr, error);
+
+    woc::unique_cf<CTFontRef> font{ CTFontCreateWithName(CFSTR("WinObjC-Italic"), 20, nullptr) };
```
msft-Jeyaram: Validate size

tests/unittests/CoreText/CTFontManagerTests.mm
```diff
@@ -0,0 +1,85 @@
...
+    EXPECT_NE(nullptr, font);
+
+    StrongId<NSString> familyName = (NSString*)CTFontCopyFullName(font.get());
+    EXPECT_OBJCEQ(@"WinObjC Italic", familyName);
+
+    EXPECT_TRUE(CTFontManagerUnregisterFontsForURL((__bridge CFURLRef)testFileURL, kCTFontManagerScopeSession, &error));
+    EXPECT_EQ(nullptr, error);
+
+    woc::unique_cf<CTFontRef> failedFont{ CTFontCreateWithName(CFSTR("WinObjC-Italic"), 20, nullptr) };
```
msft-Jeyaram: Check if the default value is returned for size of 0

tests/unittests/CoreText/CTFontManagerTests.mm
```diff
@@ -0,0 +1,85 @@
...
+    EXPECT_NE(nullptr, failedFont);
+
+    StrongId<NSString> failedFamilyName = (NSString*)CTFontCopyFullName(failedFont.get());
+    EXPECT_OBJCEQ(@"Segoe UI", failedFamilyName);
+}
+
+TEST(CTFontManager, ShouldFailToUnregisterNonregisteredFonts) {
+    char fullPath[_MAX_PATH];
+    GetModuleFileNameA(NULL, fullPath, _MAX_PATH);
```
msft-Jeyaram: As noted earlier, you are computing the same base path again.

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -580,29 +640,238 @@ HRESULT STDMETHODCALLTYPE CreateStreamFromKey(_In_ const void* fontFileReference
...
+                        woc::unique_cf<CFErrorRef> error{
+                            CFErrorCreate(nullptr, kCFErrorDomainCocoa, kCTFontManagerErrorAlreadyRegistered, nullptr)
+                        };
+                        CFArrayAppendValue(outErrors, error.get());
+                    }
+                } else {
+                    CFArrayAppendValue(m_fontDatas.get(), data);
+                    CFSetAddValue(m_fontDatasSet.get(), data);
+                    if (errors) {
+                        CFArrayAppendValue(outErrors, nullptr);
```
msft-Jeyaram: Any specific error

Frameworks/CoreGraphics/DWriteWrapper.mm
aballway: it's supposed to start before the position of the first element, which would be 0 #WontFix

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -580,29 +640,238 @@ HRESULT STDMETHODCALLTYPE CreateStreamFromKey(_In_ const void* fontFileReference
...
+    return __DWriteUpdateUserCreatedFontCollection(datas, errors, std::mem_fn(&DWriteFontCollectionLoader::AddDatas));
+}
+
+// Unregisters user defined fonts to a collection so they can be created later
+// Expects datas to be array of CFDataRefs, errors to be out pointer to array of CFErrorRef
+HRESULT _DWriteUnregisterFontsWithDatas(CFArrayRef datas, CFArrayRef* errors) {
+    return __DWriteUpdateUserCreatedFontCollection(datas, errors, std::mem_fn(&DWriteFontCollectionLoader::RemoveDatas));
+}
+
+#pragma endregion
```
msft-Jeyaram: Nit: end region name

Frameworks/CoreGraphics/DWriteWrapper.mm
aballway: They can't be joined because there are different errors if (data && !contains) vs (!data) #ByDesign

tests/unittests/CoreText/CTFontManagerTests.mm
aballway: Out of scope of PR #WontFix

tests/unittests/CoreText/CTFontManagerTests.mm
aballway: Out of scope of PR #WontFix

Frameworks/CoreText/CTFontManager.mm
```diff
@@ -34,53 +34,102 @@ CFArrayRef CTFontManagerCreateFontDescriptorsFromURL(CFURLRef fileURL) {
     return StubReturn();
 }
 
+// Helper functions to reduce code duplication, as (un)registering fonts have almost identical path
+// (Un)Registers font by creating temporary CFArray and calling into function for multiple fonts
+// TLambda :: (CFArrayRef -> CTFontManagerScope -> CFArrayRef*) -> bool
+template <typename TLambda>
+static bool __CTFontManagerUpdateWithFont(CFURLRef fontURL, CTFontManagerScope scope, CFErrorRef _Nullable* error, TLambda&& func) {
+    woc::unique_cf<CFArrayRef> fontURLs{ CFArrayCreate(nullptr, (const void**)&fontURL, 1, &kCFTypeArrayCallBacks) };
+    if (error) {
```
msft-Jeyaram: You don't need to double the logic, just call with the errors, log it, if the caller passed in error set it and return

tests/unittests/CoreText/CTFontManagerTests.mm
```diff
@@ -0,0 +1,85 @@
...
+
+    CFErrorRef error = nullptr;
+    EXPECT_TRUE(CTFontManagerRegisterGraphicsFont(graphicsFont.get(), &error));
+    EXPECT_EQ(nullptr, error);
+
+    woc::unique_cf<CTFontRef> font{ CTFontCreateWithName(CFSTR("WinObjC-Italic"), 20, nullptr) };
+    EXPECT_NE(nullptr, font);
+
+    StrongId<NSString> familyName = (NSString*)CTFontCopyFullName(font.get());
+    EXPECT_OBJCEQ(@"WinObjC Italic", familyName);
```
msft-Jeyaram: Validate some properties to wrap up the loading tests

tests/unittests/CoreText/CTFontManagerTests.mm
msft-Jeyaram: This isn't a hard test to add and will make sure we work as expected

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -53,16 +60,40 @@ static CFStringRef __CFStringCreateUppercaseCopy(CFStringRef string, CFLocaleRef
...
+ * Private helper that creates an array of CFString family names from a font collection
+ */
+CFArrayRef __DWriteCopyFontFamilyNamesFromCollection(IDWriteFontCollection* fontCollection) {
+    // Get the number of font families in the collection.
+    uint32_t count = fontCollection->GetFontFamilyCount();
+    woc::unique_cf<CFMutableArrayRef> fontFamilyNames(CFArrayCreateMutable(kCFAllocatorSystemDefault, count, &kCFTypeArrayCallBacks));
+    for (uint32_t i = 0; i < count; ++i) {
+        // Get the font family.
+        ComPtr<IDWriteFontFamily> fontFamily;
+        RETURN_NULL_IF_FAILED(fontCollection->GetFontFamily(i, &fontFamily));
```
aballway: This is only being used by internal methods, which themselves don't return an HRESULT, so I don't think it's worth changing #WontFix

Frameworks/CoreGraphics/DWriteWrapper.mm
msft-Jeyaram: We should try to match the rest of the code. Given overall framework usage is early returns, it would be best to go with the pattern. Although you might dislike it :)

Frameworks/CoreGraphics/DWriteWrapper.mm
msft-Jeyaram: Add the member convention *m_

tests/unittests/CoreText/CTFontManagerTests.mm
aballway: It's also irrelevant to what this test is validating

tests/unittests/CoreText/CTFontManagerTests.mm
aballway: The font doesn't have properties beyond the name afaik #WontFix

tests/unittests/CoreText/CTFontManagerTests.mm
msft-Jeyaram: Given we register the font. We should be able to obtain information about it (which we supplied). Thus we should make sure what we requested is what we got.
 Again pretty straightforward test to add and will make sure what you added work as intended.

aballway: This is also irrelevant to what this test is validating.

Frameworks/CoreGraphics/DWriteWrapper.mm
aballway: ?

Frameworks/CoreGraphics/DWriteWrapper.mm
aballway: I don't think doing an early return here will gain us any readability #WontFix

tests/unittests/CoreText/CTFontManagerTests.mm
msft-Jeyaram: Please elaborate how it is irrelevant?

Frameworks/include/CoreGraphics/CGFontInternal.h
```diff
@@ -0,0 +1,18 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+#pragma once
+
+COREGRAPHICS_EXPORT CFDataRef _CGFontGetData(CGFontRef font);
```
msft-Jeyaram: nit: explicit imports

tests/unittests/CoreText/CTFontManagerTests.mm
aballway: This is testing that we are able to register fonts from files, not what size we get when we pass in 0 to CTFontCreateWithName.  We should definitely be testing CTFontCreateWithName somewhere, but that is certainly neither this test nor part of this PR.

tests/unittests/CoreText/CTFontManagerTests.mm
```diff
@@ -0,0 +1,82 @@
...
+//
+//******************************************************************************
+
+#include <TestFramework.h>
+#import <UIKit/UIKit.h>
+#import <Foundation/Foundation.h>
+#import <CoreText/CoreText.h>
+#import <Starboard/SmartTypes.h>
+
+NSURL* __GetRelativeURL(NSString* relativePath) {
```
aballway: whoops good catch

tests/unittests/CoreText/CTFontManagerTests.mm
msft-Jeyaram: Another questions is: What is the default value for registered fonts? Does the answer match with reference platform?

Frameworks/CoreGraphics/CGFont.mm
ms-jihua: _have_ we been following that? on develop, the CoreText standard seems to be null return...

---
In reply to: [89845644](https://github.com/Microsoft/WinObjC/pull/1454#discussion_r89845644) [](ancestors = 89845644)

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -53,16 +60,40 @@ static CFStringRef __CFStringCreateUppercaseCopy(CFStringRef string, CFLocaleRef
...
+
+        // Get a list of localized strings for the family name.
+        ComPtr<IDWriteLocalizedStrings> familyNames;
+        RETURN_NULL_IF_FAILED(fontFamily->GetFamilyNames(&familyNames));
+
+        CFStringRef name = _CFStringFromLocalizedString(familyNames.Get());
+        if (CFStringGetLength(name) == 0) {
```
ms-jihua: should you check   if (name)   here?

tests/unittests/CoreGraphics/CGFontTests.mm
```diff
@@ -18,6 +18,7 @@
 #import <CoreGraphics/CoreGraphics.h>
 #import <CoreFoundation/CoreFoundation.h>
 #import <Foundation/Foundation.h>
+#import <CoreGraphics/DWriteWrapper.h>
```
ms-jihua: doesn't seem to be needed?

ms-jihua:
>  @Status Stub

how come we're leaving this out?

---
Refers to: Frameworks/CoreText/CTFontManager.mm:128 in f49940a. [](commit_id = f49940a0f302d667fae27df18c680b174b3fb430, deletion_comment = False)

aballway:
>  @Status Stub

It wasn't one of the functions in issue #1022.  I only added the other unregistering ones because they were needed for testing.

---
In reply to: [263722645](https://github.com/Microsoft/WinObjC/pull/1454#issuecomment-263722645) [](ancestors = 263722645)

---
Refers to: Frameworks/CoreText/CTFontManager.mm:128 in f49940a. [](commit_id = f49940a0f302d667fae27df18c680b174b3fb430, deletion_comment = False)

Frameworks/UIKit/UIApplicationMain.mm
ms-jihua: should probably add a comment about this

tests/unittests/CoreText/CTFontManagerTests.mm
```diff
@@ -0,0 +1,82 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#include <TestFramework.h>
```
ms-jihua: standardize on #import?

tests/unittests/CoreText/CTFontManagerTests.mm
```diff
@@ -0,0 +1,82 @@
...
+#import <UIKit/UIKit.h>
+#import <Foundation/Foundation.h>
+#import <CoreText/CoreText.h>
+#import <Starboard/SmartTypes.h>
+
+static NSURL* __GetRelativeURL(NSString* relativePath) {
```
ms-jihua: nit: mm, don't really like this name - "relative" seems too vague and could mean any number of things

ms-jihua:
possibly excessive but might be good to add tests that validate that the registered fonts are accessible from CG-side as well

tests/unittests/CoreText/CTFontManagerTests.mm
ms-jihua: We have very little reason to believe that this is an interesting combination of features to test.
- font files don't have any associated 'default' size
- no reason to believe that default size of registered fonts is different from default size of non-registered fonts

We have tests for the 0 -> default font size feature in UIFontTests. We can add a slightly redundant test to CTFontTests if we want, but that's different from testing the combination of "registered font" and "default font size".

ms-jihua:
>  @Status Stub

I realize we didn't explicitly call it out in #1022, but it'd be good to take care of this now, just to finish up the feature set.

---
In reply to: [263724102](https://github.com/Microsoft/WinObjC/pull/1454#issuecomment-263724102) [](ancestors = 263724102,263722645)

---
Refers to: Frameworks/CoreText/CTFontManager.mm:128 in f49940a. [](commit_id = f49940a0f302d667fae27df18c680b174b3fb430, deletion_comment = False)

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -97,60 +130,69 @@ bool operator()(const woc::unique_cf<CFStringRef>& item1, const woc::unique_cf<C
...
-        return ret;
-    }();
+    return ret;
+}
+
+/**
+ * Private static map, that maps display name and postscript name to a _DWriteFontProperties struct for the corresponding user defined font.
+ * Kept separate from c_systemFontPropertiesMap because this needs to be edited whenever a font is (un)registered
+ */
+static __DWritePropertiesMap _userFontPropertiesMap;
```
ms-jihua: s_userFontPropertiesMap, then?

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -201,39 +243,22 @@ CFStringRef _CFStringFromLocalizedString(IDWriteLocalizedStrings* localizedStrin
...
-        if (CFStringGetLength(name) == 0) {
-            TraceError(TAG, L"Failed to convert the localized string to wide string.");
-            continue;
-        }
-
-        CFArrayAppendValue(fontFamilyNames.get(), name);
+    woc::unique_cf<CFMutableArrayRef> systemFamilyNames{ (CFMutableArrayRef)__DWriteCopyFontFamilyNamesFromCollection(
+        fontCollection.Get()) };
+    std::lock_guard<std::mutex> guard(_userCreatedFontCollectionMutex);
+    if (_userCreatedFontCollection) {
```
ms-jihua: does registering a font actually influence [UIFont familyNames] on the reference platform? (if so, can we add a test for this?)

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -97,60 +130,69 @@ bool operator()(const woc::unique_cf<CFStringRef>& item1, const woc::unique_cf<C
...
+    }
 
-        return ret;
-    }();
+    return ret;
+}
+
+/**
+ * Private static map, that maps display name and postscript name to a _DWriteFontProperties struct for the corresponding user defined font.
+ * Kept separate from c_systemFontPropertiesMap because this needs to be edited whenever a font is (un)registered
```
ms-jihua: (personally i'm against it being function-local, it feels like these two maps ought to exist at the same scopes?)

aballway: Since it's not used anywhere else I'd like to keep it function local, and as @DHowett-MSFT and @rajsesh-msft learned me we run into fun DLL/Com loading stuff if it's file static

ms-jihua:
>  @Status Stub

fix annotation!

---
In reply to: [263735377](https://github.com/Microsoft/WinObjC/pull/1454#issuecomment-263735377) [](ancestors = 263735377,263724102,263722645)

---
Refers to: Frameworks/CoreText/CTFontManager.mm:128 in f49940a. [](commit_id = f49940a0f302d667fae27df18c680b174b3fb430, deletion_comment = False)

tests/unittests/CoreText/CTFontManagerTests.mm
```diff
@@ -0,0 +1,110 @@
...
+TEST(CTFontManager, ShouldFailToUnregisterNonregisteredFonts) {
+    NSURL* testFileURL = __GetURLFromPathRelativeToCurrentDirectory(@"/data/WinObjC.ttf");
+    CFErrorRef error = nullptr;
+    EXPECT_TRUE(CTFontManagerUnregisterFontsForURL((__bridge CFURLRef)testFileURL, kCTFontManagerScopeSession, &error));
+    EXPECT_NE(nullptr, error);
+    EXPECT_EQ(kCTFontManagerErrorNotRegistered, CFErrorGetCode(error));
+}
+
+TEST(CTFontManager, UIFontFamilyNamesShouldContainRegisteredFonts) {
+    NSArray* familyNames = [UIFont familyNames];
```
ms-jihua: Wait do we actually have access to UIKit functions in CoreText tests? That's kind of scary. Can we remove the UIKit import and move this test to UIKit tests instead?

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -580,29 +637,238 @@ HRESULT STDMETHODCALLTYPE CreateStreamFromKey(_In_ const void* fontFileReference
...
+    void AddDatas(CFArrayRef fontDatas, CFArrayRef* errors) {
+        CFMutableArrayRef outErrors = nil;
+        if (errors) {
+            outErrors = CFArrayCreateMutable(nullptr, 0, &kCFTypeArrayCallBacks);
+            *errors = outErrors;
+        }
+
+        CFIndex count = CFArrayGetCount(fontDatas);
+        for (CFIndex i = 0; i < count; ++i) {
+            CFDataRef data = (CFDataRef)CFArrayGetValueAtIndex(fontDatas, i);
```
ms-jihua: >(CFDataRef) [](start = 29, length = 11)

static_cast

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -580,29 +637,238 @@ HRESULT STDMETHODCALLTYPE CreateStreamFromKey(_In_ const void* fontFileReference
...
+            if (data) {
+                if (CFSetContainsValue(m_fontDatasSet.get(), data)) {
+                    if (errors) {
```
ms-jihua: this amount of nesting is pretty hard to read - consider making a helper that creates & appends an error, if !outErrors, and an overload for appending nullptr

tests/unittests/CoreText/CTFontManagerTests.mm
ms-jihua: Philosophical component isolation issue - UIKit is built on top of CT, so it's okay for UIKit tests to have CT, but not really vice-versa.
Additionally, for cross-compiling purposes, UIKit doesn't exist on OSX (only iOS), so we'd have to move this test out anyway if we were to make CoreText tests cross-compile.

Frameworks/include/CoreGraphics/CGFontInternal.h
```diff
@@ -0,0 +1,20 @@
+//******************************************************************************
```
ms-jihua: nit: move this header to just be in Frameworks/CoreGraphics - it's not needed cross-module

aballway: It's used in CTFontManager to (un)register graphics fonts

ms-jihua: nvm i dumb

Frameworks/CoreGraphics/CGFont.mm
```diff
@@ -40,6 +41,9 @@
     CFRuntimeBase _base;
     ComPtr<IDWriteFontFace> _dwriteFontFace;
 
+    // Contains a value when created via Data Provider, null o/w
+    woc::unique_cf<CFDataRef> _data;
```
ms-jihua: hmm, wondering if it's worth it to cache the data provider instead here. our current data provider impl is just a data, but we could save some memory in the future once data provider is properly implemented and defers loading its binary data

aballway: That's the plan, did this to keep from having to copy the data each time this would be used.

ms-jihua: makes sense, can you jot a comment down about the tradeoff then? i agree that this is preferable for the time being, but keeping a ttf file entirely in memory is still less than desirable.

PR #1458, opened 2016-11-24
remove export of internal function  _CGImageGetWICImageSource

msft-Jeyaram:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1458)


DHowett-MSFT:
Since this pull request touches the WIC image source function, you could also put in a fix for the COM over-release!

PR #1459, opened 2016-11-24
Always call locationManager:didUpdateLocations: when location changes…

### LINKED ISSUE #1431: CoreLocation/CLLocationManager doesn't work
Detecting location using CLLocationManager from CoreLocation does not work: delegate methods
```
locationManager:didUpdateLocations:
locationManager:didUpdateHeading:
```
never called!

Specifying 'location' in capability requirements as mentioned here https://github.com/Microsoft/WinObjC/issues/1321  does not help.  Permission to use location by app granted in Settings. Location Services do not work neither in WOCCatalog example nor in our project.
Could you please give any advice what could be the problem and how to fix it?

Thank you in advance.

rajsesh-msft:
Always call locationManager:didUpdateLocations: when location changes. Fixes #1431.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1459)

### DIFF
--- a/Frameworks/CoreLocation/CLLocationManager.mm
+++ b/Frameworks/CoreLocation/CLLocationManager.mm
@@ -103,7 +103,7 @@ static void _requestExtendedExecutionSession() {
  */
 static void _removeExtendedExecutionSession() {
     std::lock_guard<std::mutex> lock(s_extendedExecutionMutex);
-    
+
     if (s_extendedExecutionRequestCount == 0) {
         return;
     }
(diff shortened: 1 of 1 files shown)


rajsesh-msft:
@DHowett-MSFT, the deprecated delegate is the one that gets called if the position did changed.  The delegate that is outside the if is the new one and is always called when the location device says something changed (even if it didn't change).

DHowett-MSFT:
Whoops, I had it backwards! Carry on :+1:

PR #1464, opened 2016-11-28
Fix various view controller, popover, and presentation controller memory management problems (crashes and leaks)

ehren:
Commit fixes a number of problems (many of them my doing from earlier popover work). The fact that WOCCatalog doesn't crash prior to this commit is essentially due to multiple bugs cancelling each other out. 

- leaks from failing to .attach a StrongId to a +1 init
- bad aliasing and general memory management of priv->_presentationController and priv->_popoverPresentationController
- bad existing implementation of UIPresentationController
- storyboard setter = crash
- Had to be more careful with the block based cleanup on popover dismiss (self retain/release a little gross but it's better than poking at the UIViewController private ivars from UIPopoverPresentationController)

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1464)


Frameworks/UIKit/UIPopoverPresentationController.mm
```diff
@@ -124,6 +122,17 @@ - (void)_dismissAnimated:(BOOL)animated completion:(dispatch_block_t)dismissComp
     [_popoverControllerInternal _dismissPopoverAnimated:animated completion:dismissCompletion];
 }
 
+- (void)_handleDismissCleanup {
+    // stay alive while UIViewController forgets about us
```
oliversa-msft: >stay [](start = 7, length = 4)

Super nitpicky - "Stay" - I'm sorry just my nature. ;-)

Frameworks/UIKit/UIViewController.mm
```diff
@@ -2245,7 +2253,6 @@ - (void)dealloc {
 
     priv->_childViewControllers = nil;
     priv->_searchDisplayController = nil;
-    priv->_storyboard = nil;
```
jaredhms: ...although I see that it's now a StrongId and therefore assignment to nil is unnecessary.

---
In reply to: [89896831](https://github.com/Microsoft/WinObjC/pull/1464#discussion_r89896831) [](ancestors = 89896831)

Frameworks/UIKit/UIPopoverPresentationController.mm
```diff
@@ -124,6 +122,17 @@ - (void)_dismissAnimated:(BOOL)animated completion:(dispatch_block_t)dismissComp
...
+    [self retain];
```
jaredhms: >[self retain]; [](start = 4, length = 14)

Perhaps cleaner to just do:

StrongId<UIPopoverPresentationController> strongSelf = self; 

...so we don't have to manually call retain/release.

Frameworks/include/UIPopoverPresentationControllerInternal.h
```diff
@@ -19,7 +19,7 @@
 @interface UIPopoverPresentationController () {
 @public
     BOOL _managesPresentation;
-    dispatch_block_t _willDismissCompletion;
+    dispatch_block_t _onDismissCleanupHandler;
```
jaredhms: >dispatch_block_t [](start = 4, length = 16)

Can we use a StrongId here, too?

Frameworks/UIKit/UIPopoverPresentationController.mm
```diff
@@ -316,7 +325,7 @@ - (void)setPopoverBackgroundViewClass:(Class)popoverBackgroundViewClass {
 }
 
 - (void)dealloc {
-    [_willDismissCompletion release];
+    [_onDismissCleanupHandler release];
```
oliversa-msft: >[_onDismissCleanupHandler release]; [](start = 4, length = 35)

Do we need to set to nil as well? It's possible this has already happened if _handleDismissCleanup was called.

Frameworks/UIKit/UIPresentationController.mm
```diff
@@ -19,24 +19,51 @@
...
  @Status Interoperable
 */
 - (instancetype)initWithPresentedViewController:(UIViewController*)presentedViewController
              presentingViewController:(UIViewController*)presentingViewController {
-    _presentingViewController = presentingViewController;
-    _presentedViewController = presentedViewController;
-    _presentationStyle = _presentedViewController.modalPresentationStyle;
-    _containerView = _presentingViewController.view;
+    if (self = [super init]) {
+        _presentingViewController = presentingViewController;
```
oliversa-msft: >_presentingViewController = presentingViewController [](start = 8, length = 52)

Is this right? Do we now managing the passed-in presentingViewController lifetime? If UIPresentationController is dealloc'ed, can we nil out the presentingViewController because we own it? Do we need to make a copy or something?

oliversa-msft: I guess this is not a problem since both objects exist on UIViewControllerInternal for the lifetime of the the UIViewControllerPriv.

---
In reply to: [89899882](https://github.com/Microsoft/WinObjC/pull/1464#discussion_r89899882) [](ancestors = 89899882)

Frameworks/UIKit/UIViewController.mm
```diff
@@ -1336,28 +1349,23 @@ - (void)_addToTop:(NSNumber*)animatedValue {
...
-
-            priv->_presentationController = priv->_popoverPresentationController;
-
             [priv->_popoverPresentationController _prepareForPresentation];
 
             displayPopover = ![self _hidesParent];
 
             if (displayPopover) {
-                priv->_popoverPresentationController->_managesPresentation = TRUE;
+                priv->_popoverPresentationController->_managesPresentation = YES;
```
jaredhms: >_popoverPresentationController->_managesPresentation [](start = 22, length = 52)

it's unfortunate that we have to reach into _popoverPresentationController's internal state in so many places.  Is there any way to further encapsulate this block of logic?  Most of it seems like it should be handled by the UIPopoverPresentationController.

ehren: This is slightly improved by the followup commit. I've still opted to keep the teardown of the UIViewController priv variables inside UIViewController.mm (instead of poking at these from UIPopoverPresentationController.mm). The _prepareForPresentation call needs to stay in UIViewController so that the UIPopoverPresentationDelegate is still informed even when there's no actual popover (non-tablet scenarios).

Frameworks/UIKit/UIPopoverPresentationController.mm
ehren: Handled by strong id change. The dismissCompletion won't be invoked prior to dealloc though (this is ensured by grabbing the strong ref in _handleDismiss)

Frameworks/UIKit/UIPopoverPresentationController.mm
ehren: Fixed and added more detail to comment

Frameworks/UIKit/UIPopoverPresentationController.mm
ehren: fixed in new commit

Frameworks/UIKit/UIViewController.mm
ehren: I see that the UIViewController.mm `setStoryboard:` method isn't exposed publicly so technically there's no crash in the current code (but the ivar should still be changed to a strong id)

Frameworks/UIKit/UIPopoverPresentationController.mm
ehren: that should be won't be invoked *after* dealloc

Frameworks/UIKit/UIViewController.mm
jaredhms: Thanks for simplifying this; it seems quite a bit cleaner.  Can you elaborate on why we need to support non-tablet scenarios in WinObjC though?  Would this be a spot where we could simplify the implementation by always showing popups (even on phone)?

---
In reply to: [90111314](https://github.com/Microsoft/WinObjC/pull/1464#discussion_r90111314) [](ancestors = 90111314)

Frameworks/include/UIPopoverPresentationControllerInternal.h
```diff
@@ -16,14 +16,11 @@
 
 #pragma once
 
-@interface UIPopoverPresentationController () {
-@public
-    BOOL _managesPresentation;
-    dispatch_block_t _willDismissCompletion;
-}
+@interface UIPopoverPresentationController()
```
oliversa-msft: The file is named UIPopoverPresentationControllerInternal - generally the interfaces match that name as well.

Frameworks/UIKit/UIPopoverPresentationController.mm
```diff
@@ -81,6 +79,8 @@ @implementation UIPopoverPresentationController {
     StrongId<UIBarButtonItem*> _barButtonItem;
     StrongId<UIView*> _sourceView;
     CGRect _sourceRect;
+    StrongId<dispatch_block_t> _dismissCompletion;
+    BOOL _isManagingPresentation;
```
oliversa-msft: >BOOL _isManagingPresentation; [](start = 4, length = 29)

This BOOL exist on the interface as well - is that intentional?

Frameworks/include/UIPopoverPresentationControllerInternal.h
ehren: I need this header for the class extension / anonymous category exposing methods for UIViewController. I did fix a whitespace error with the "()" in the last commit.

Frameworks/UIKit/UIViewController.mm
```diff
@@ -1222,9 +1230,9 @@ - (void)dismissViewControllerAnimated:(BOOL)animated completion:(void (^)(void))
     [curController retain];
     [curController autorelease];
 
-    BOOL popoverPresented = [curController _popoverManagesPresentation];
+    BOOL realPopoverPresented = [[curController popoverPresentationController] _isManagingPresentation];
```
ehren: @jaredhms I've added a comment. Regarding your message (which I think was deleted) about whether to simplify the implementation by always showing a popover on phones/non-tablets, I'm initially supportive though would prefer a separate pull. The behaviour has changed in iOS 9 anyway to allow popovers on phones in "non-horizontally compact" environments. There would also be an easy workaround for users that don't want the popover - just use a different modal presentation style. Simplification wise, this would allow the popover presentation controller setup to be moved out of _addToTop and into presentViewController:. I'm also not sure whether the UIPopoverPresentationControllerDelegate is being handled correctly for non actual popover presentations right now (I'm not sure if calling dismissViewController: should fire the didDismiss delegate - would have to be checked) - so it probably would be an improvement to just always show a popover.

We'd have to consider the behaviour of UIModalPresentationFormSheet on non-tablets as well. There I'm thinking that the current behaviour to display as full screen might still be best (since the default hardcoded dimensions of the FormSheet modal are inappropriate for phones) - though on the other hand these are resizable via preferredContentSize.

Note however, there are a few framework issues I need to file that have come up regarding popovers though the problems are unrelated to this pull (just a heads up).

jaredhms: This sounds great, @ehren.  Can you please file an issue/suggestion tracking the simplification to always show a popup (in addition to the rest of the issues you've been hitting)?  Thanks!

---
In reply to: [90317212](https://github.com/Microsoft/WinObjC/pull/1464#discussion_r90317212) [](ancestors = 90317212)

jaredhms:
:shipit:

jaredhms:
In CI build.

oliversa-msft:
:shipit:

PR #1466, opened 2016-11-28
Make tgmath header C compliant

memontic-ms:
Previously, .m files that imported tgmath.h could not be compiled as Objective C. Since tgmath.h included the C++ header cmath, .m files had to be compiled as Objective C++.

Fix: #735

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1466)


msvc/LLVM-3.6.0/lib/clang/3.7.0/include/tgmath.h
```diff
@@ -29,7 +29,11 @@
 #define __TGMATH_H
 
 /* C99 7.22 Type-generic math <tgmath.h>. */
+#if defined(__cplusplus)
```
aballway: should the copyright banner be updated?

memontic-ms: It's not a Microsoft copyright banner. What would need updating?

DHowett-MSFT:
I'm concerned that when we get another compiler header/lib drop we'll stomp this change sight unseen.

memontic-ms:
@DHowett-MSFT I have included ObjCImportTest.m within the Clang unit tests that will catch a regression on this issue for future compiler drops.
Related, I have renamed the ClangModules unit tests project to just Clang since the added test file is not dependent on modules support.

PR #1469, opened 2016-11-29
Remove an internal header from drawing tests.

DHowett-MSFT:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1469)


ms-jihua:
:shipit:

DHowett-MSFT:
The build failed because the test actually _is_ relying on an internal feature. @msft-Jeyaram  :(

DHowett-MSFT:
Abandoning PR.

msft-Jeyaram:
The build failed since _CGContextDrawImageRectwas in CGContextInternal.h.
We still need that import (needs to be ifdefined for Mac testing)

PR #1470, opened 2016-11-29
Fix the drawing tests to not use Objective-C and build on OS X.

### LINKED ISSUE #1453: [CGD2D] CGImage.mm:580 causes over-release of COM object
The `.Get()` [here](https://github.com/Microsoft/WinObjC/pull/1414/commits/9fe3638ca7157d72d9063fb4f65e7503d624aa72#diff-16e01a72b686518a7ed4ff3668919f2bR580) does not store a +1 reference into the destination but the destination ([here](https://github.com/Microsoft/WinObjC/pull/1414/commits/9fe3638ca7157d72d9063fb4f65e7503d624aa72#diff-71f2997808c3d09d7d9ae3411de51489R1794)) is a ComPtr and will release it.

The preferred method for getting a +1 reference from a ComPtr **a** into an out pointer **b** is `a.CopyTo(b)`.

DHowett-MSFT:
This fixes #1453 and #1448.

I thought that #1453 was the cause of the teardown crash I was seeing, so it got fixed here; it turns out that CGImageDestination was overreleasing a CGImage's CGDataProvider.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1470)

### DIFF
--- a/Frameworks/CoreGraphics/CGImage.mm
+++ b/Frameworks/CoreGraphics/CGImage.mm
@@ -576,8 +576,7 @@ const __CGImagePixelProperties* _CGGetPixelFormatProperties(WICPixelFormatGUID p
 HRESULT _CGImageGetWICImageSource(CGImageRef image, IWICBitmap** source) {
     RETURN_HR_IF_NULL(E_INVALIDARG, image);
     RETURN_HR_IF_NULL(E_POINTER, source);
-    *source = image->ImageSource().Get();
-    return S_OK;
+    return image->ImageSource().CopyTo(source);
 }
 
 DisplayTexture* _CGImageGetDisplayTexture(CGImageRef image) {
--- a/tests/UnitTests/CoreGraphics.drawing/CGContextDrawingTests.cpp
+++ b/tests/UnitTests/CoreGraphics.drawing/CGContextDrawingTests.cpp
@@ -14,30 +14,20 @@
 //
 //******************************************************************************
 
-#import "DrawingTest.h"
-#import <windows.h>
-#import "CGContextInternal.h"
+#include "DrawingTest.h"
+#include "DrawingTestConfig.h"
+#include "ImageHelpers.h"
+#include <windows.h>
 
-// TODO:#1448 Remove the use of OBJC code
 #ifdef WINOBJC
-
-static NSString* getModulePath() {
-    char fullPath[_MAX_PATH];
-    GetModuleFileNameA(NULL, fullPath, _MAX_PATH);
-    return [@(fullPath) stringByDeletingLastPathComponent];
-}
-
-static NSString* getPathToFile(NSString* fileName) {
-    static StrongId<NSString*> refPath = getModulePath();
-    return [refPath stringByAppendingPathComponent:fileName];
-}
+#include "CGContextInternal.h"
 
 DISABLED_DRAW_TEST_F(CGContext, DrawIntoRect, UIKitMimicTest) {
     // Draw a portion of an image into a different region.
-    CFDataRef data = (CFDataRef)[NSData dataWithContentsOfFile:getPathToFile(@"png3.9.png")];
-    woc::unique_cf<CGDataProviderRef> dataProvider(CGDataProviderCreateWithCFData(data));
+    auto drawingConfig = DrawingTestConfig::Get();
 
-    woc::unique_cf<CGImageRef> image(CGImageCreateWithPNGDataProvider(dataProvider.get(), NULL, NO, kCGRenderingIntentDefault));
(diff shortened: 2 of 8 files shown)


tests/UnitTests/CoreGraphics.drawing/EntryPoint.cpp
```diff
@@ -82,6 +82,11 @@ class CommandLineDrawingTestConfigImpl : public DrawingTestConfigImpl {
     virtual std::string GetComparisonPath() override {
         return _comparisonPath;
     }
+
+    virtual std::string GetResourcePath(const std::string& resource) override {
+        // Using / as a path separator is valid on both Windows and OS X.
+        return __ModuleDirectory() + "/data/" + resource;
```
msft-Jeyaram: > "/data/"  [](start = 36, length = 10)

Curious: In windows the files aren't copied into data. in OSX it is?
Also it would be better not to hard code /data/ in this function and pass it in?
(make it more generic? maybe?)

msft-Jeyaram:
> /* static */ DrawingTestConfigImpl* DrawingTestConfig::Get() {

static?

---
Refers to: tests/UnitTests/CoreGraphics.drawing/EntryPoint.cpp:93 in f000645. [](commit_id = f000645256b7d268d71b39deb9765050ea357d3e, deletion_comment = False)

tests/UnitTests/CoreGraphics.drawing/EntryPoint.cpp
DHowett-MSFT: The Drawing test vcxproj absolutely copies a data/ directory.

If we standardize on `data/`, we don't have to propagate that decision into every test file consumer (and centralizing it here lets us change it in a single place if we do decide to.)

DHowett-MSFT:
> /* static */ DrawingTestConfigImpl* DrawingTestConfig::Get() {

Yes.

This is a static member function on DrawingTestConfig.

Since we cannot annotate it as such via classical means, this comment serves to illustrate that it is static.

---
In reply to: [263743436](https://github.com/Microsoft/WinObjC/pull/1470#issuecomment-263743436) [](ancestors = 263743436)

---
Refers to: tests/UnitTests/CoreGraphics.drawing/EntryPoint.cpp:93 in f000645. [](commit_id = f000645256b7d268d71b39deb9765050ea357d3e, deletion_comment = False)

tests/UnitTests/CoreGraphics.drawing/EntryPoint.cpp
msft-Jeyaram: for some reason I was thinking it wasn't. Double checked and it is.
gotcha.

---
In reply to: [90145289](https://github.com/Microsoft/WinObjC/pull/1470#discussion_r90145289) [](ancestors = 90145289)

Frameworks/CoreGraphics/CGImage.mm
```diff
@@ -576,8 +576,7 @@ bool _CGIsValidRenderTargetPixelFormat(WICPixelFormatGUID pixelFormat) {
 HRESULT _CGImageGetWICImageSource(CGImageRef image, IWICBitmap** source) {
     RETURN_HR_IF_NULL(E_INVALIDARG, image);
     RETURN_HR_IF_NULL(E_POINTER, source);
-    *source = image->ImageSource().Get();
-    return S_OK;
+    return image->ImageSource().CopyTo(source);
```
rajsesh-msft: 😮

tests/UnitTests/CoreGraphics.drawing/CGContextDrawingTests.cpp
```diff
@@ -14,30 +14,20 @@
...
-    static StrongId<NSString*> refPath = getModulePath();
-    return [refPath stringByAppendingPathComponent:fileName];
-}
+#include "CGContextInternal.h"
 
 DISABLED_DRAW_TEST_F(CGContext, DrawIntoRect, UIKitMimicTest) {
     // Draw a portion of an image into a different region.
-    CFDataRef data = (CFDataRef)[NSData dataWithContentsOfFile:getPathToFile(@"png3.9.png")];
-    woc::unique_cf<CGDataProviderRef> dataProvider(CGDataProviderCreateWithCFData(data));
+    auto drawingConfig = DrawingTestConfig::Get();
```
rajsesh-msft: nit: not a good use of auto when the return type is known.

DHowett-MSFT: Very true.

---
In reply to: [90145740](https://github.com/Microsoft/WinObjC/pull/1470#discussion_r90145740) [](ancestors = 90145740)

Issue #1451, opened 2016-11-23
Canabalt [ARM-x86]: blocked at splash screen 


MichelLopez:
Desktop running 14975
Run the app (debug/release)

Expected: App runs, shows buttons after splash screen appears.
Actual: App runs, blocked at splash screen both with debug and release

ArnOmsft:
This bug has been fixed in the latest release.

PR #1471, opened 2016-11-30
Fix [GitHub Issue #1451] Canabalt [ARM-x86]: blocked at splash screen

yiyang-msft:
The issue here is that when a sublayer is removed from super, we need call setNeedsLayout which will flush out corresponding transactions to update the view hierarchy.  

This is based on the observation from reference platform when removing layer. The following call sequence is happening. 

[subLayer removeFromSuper] triggers the following call sequence
[CALayer setNeedsLayout] -> [CALayer drawInContext] ->[UIView drawRect]

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1471)

### DIFF
--- a/Frameworks/QuartzCore/CALayer.mm
+++ b/Frameworks/QuartzCore/CALayer.mm
@@ -802,8 +802,6 @@ static void doRecursiveAction(CALayer* layer, NSString* actionName) {
         return;
     }
 
-    CALayer* oursuper = priv->superlayer;
-
     //  If our layer is visible, order all subviews out
     bool isVisible = false;
 
@@ -833,6 +831,8 @@ static void doRecursiveAction(CALayer* layer, NSString* actionName) {
     [CATransaction _removeLayer:self];
 
     pSuper->priv->removeChild(self);
+
+    [pSuper setNeedsLayout];
     [self release];
 }
 
--- a/Frameworks/QuartzCore/CALayer.mm
+++ b/Frameworks/QuartzCore/CALayer.mm
@@ -806,7 +806,7 @@ static void doRecursiveAction(CALayer* layer, NSString* actionName) {
     bool isVisible = false;
 
     CALayer* curLayer = self;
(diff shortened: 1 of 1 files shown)


Frameworks/QuartzCore/CALayer.mm
```diff
@@ -833,6 +831,8 @@ - (void)removeFromSuperlayer {
     [CATransaction _removeLayer:self];
 
     pSuper->priv->removeChild(self);
+
+    [pSuper setNeedsLayout];
     [self release];
```
tadam-msft: Looks good, can we get a XAMLCatalog sample that would demonstrate this issue 100%--maybe on a new "Bugs" section

jaredhms: >[self release]; [](start = 4, length = 15)

wow - that's kind of odd.  you'd think that the superlayer would be holding the reference....?

jaredhms:
>     CALayer* pSuper = (CALayer*)priv->superlayer;

nit: can we remove the hungarian while we're here?

---
Refers to: Frameworks/QuartzCore/CALayer.mm:809 in 8cdcd27. [](commit_id = 8cdcd27fe340aa879f54db9b7e36d67169927cb2, deletion_comment = False)

jaredhms:
:shipit:

Frameworks/QuartzCore/CALayer.mm
yiyang-msft: indeed strange. we will re-visit its usage pattern later though.

---
In reply to: [90150584](https://github.com/Microsoft/WinObjC/pull/1471#discussion_r90150584) [](ancestors = 90150584)

yiyang-msft: I have a repro and a test case in progress. will check it in separately.

---
In reply to: [90150570](https://github.com/Microsoft/WinObjC/pull/1471#discussion_r90150570) [](ancestors = 90150570)

Issue #1472, opened 2016-11-30
[CGD2D] CGBitmapContext does not clean the buffer it has allocated.


DHowett-MSFT:
There is garbage data left in the backing store for a CGBitmapContext's bitmap. On debug builds, it is filled with `0xCD`, and on release builds it is `<uninitialized>`.

It can be presumed that a bitmap context is _fully empty_; in fact, on the reference platform, creating a bitmap context and immediately saving its image to disk results in an alpha-0 image with the provided dimensions.

DHowett-MSFT:
Fallout addressed [here](https://github.com/Microsoft/WinObjC/pull/1470/commits/1319fdca5d75a3fecc3c566c732f637c27257ca3).

PR #1475, opened 2016-11-30
Zero out memory for backing buffer of Context

msft-Jeyaram:
#1472

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1475)


DHowett-MSFT:
Today I learned that this is valid! According to *ISO C++03 5.3.4[expr.new]/15*, even!

DHowett-MSFT:
:shipit:

msft-Jeyaram:
yep not so known

PR #1480, opened 2016-11-30
objc2winmd binary update: Added support for marshaling NSURL

mukhole:
NSURL will now be marshalled as StorageFile or Uri or both depending on the type of annotations provided.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1480)


pradipd:
:shipit:

PR #1489, opened 2016-12-01
Copying the entire test directory to the device.

pradipd:
ARM Unit tests that are trying to load a font are failing because the run-unittests.ps1 script doesn't copy the fonts to the device.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1489)


ms-jihua:
:shipit:

PR #1493, opened 2016-12-02
Fix issue in UIButton that we are not fallback to value of normal state when setting value of other states to be nil

yiyang-msft:
UIButton can set different title/color/image for different state. if the value of a state, e.g., title for highlighted state is set to nil, then when button is lightlighted state, its value should fallback to normal state value. 

Currently when we set nil to title/color/image value for a state (e.g., highlighted), we are creating a non-nil inspectable object which in turns contains null comObj, so when button is in that state (e.g., highlighted state), we use this inspectable to retrieve their contained value (null) to update UIButton, which causes button text/image display incorrectly.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1493)

### DIFF
--- a/Frameworks/UIKit/UIButton.mm
+++ b/Frameworks/UIKit/UIButton.mm
@@ -262,13 +262,19 @@ Microsoft Extension
 */
 - (void)setImage:(UIImage*)image forState:(UIControlState)state {
     _states[state].image = image;
+
+    // NOTE: check if image is nil before creating inspetableImage
+    // ConvertUIImageToWUXMImageBrush:nil creates a valid imageBrush with null comObj
+    // which isn't what we want
     if (image) {
         WUXMImageBrush* imageBrush = ConvertUIImageToWUXMImageBrush(image);
         if (imageBrush) {
             _states[state].inspectableImage = [imageBrush comObj];
         }
     } else {
-        _states[state].inspectableImage = nil;
+        // this enforces the fallback of using Image of normalState
+        // when a image for other states does not exis
+        _states[state].inspectableImage = nullptr;
     }
 
     // Update the Xaml elements immediately, so the proxies reflect reality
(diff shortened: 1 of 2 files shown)


Frameworks/UIKit/UIButton.mm
```diff
@@ -481,9 +485,19 @@ - (UIImage*)currentBackgroundImage {
 */
 - (void)setTitle:(NSString*)title forState:(UIControlState)state {
     _states[state].title.attach([title copy]);
-    RTObject* rtString = [WFPropertyValue createString:title];
-    if (rtString) {
-        _states[state].inspectableTitle = [rtString comObj];
+
+    // NOTE: check if title is nil before creating inspetableTitle
```
jaredhms: >inspetableTitle [](start = 51, length = 15)

nit: typo

Frameworks/UIKit/UIButton.mm
```diff
@@ -262,9 +262,13 @@ - (void)initAccessibility {
...
-    WUXMImageBrush* imageBrush = ConvertUIImageToWUXMImageBrush(image);
-    if (imageBrush) {
-        _states[state].inspectableImage = [imageBrush comObj];
+    if (image) {
+        WUXMImageBrush* imageBrush = ConvertUIImageToWUXMImageBrush(image);
+        if (imageBrush) {
+            _states[state].inspectableImage = [imageBrush comObj];
+        }
+    } else {
+        _states[state].inspectableImage = nil;
```
jaredhms: >nil [](start = 42, length = 3)

nit: should be nullptr for all of these; you're assigning to a ComPtr.

Frameworks/UIKit/UIButton.mm
```diff
@@ -262,9 +262,13 @@ - (void)initAccessibility {
 */
 - (void)setImage:(UIImage*)image forState:(UIControlState)state {
     _states[state].image = image;
-    WUXMImageBrush* imageBrush = ConvertUIImageToWUXMImageBrush(image);
-    if (imageBrush) {
-        _states[state].inspectableImage = [imageBrush comObj];
+    if (image) {
```
jaredhms: >   if (image) { [](start = 1, length = 15)

nit: worth adding the same comment here as you did in the other two portions of the fix?

jaredhms:
:shipit:

tadam-msft:
Thanks

Sent from Windows Mail

Issue #1374, opened 2016-11-14
Investigate core text performance with DWrite stack


rajsesh-msft:
We have a few inefficiencies:

1. RenderTarget Begin/EndDraw() is called for each rendering block (glpyh runs), which will cause performance issues because EndDraw() is expensive.  We should batch these calls and known points.

2. Creating of numerous dwrite factory instances.

3. Static initializations.

rajsesh-msft:
One demo scenario is the non-standard edit control in xamlcatalog.

rajsesh-msft:
There was one thing that stood out during perf analysis - we spent a lot of time in DWriteWrapper.mm _DWriteGetFontPropertiesFromName(), specifically in __CFLocaleCopyCurrent.  While this is a result of #1490, the font properties don't need to be localized, and there is no need for any usage of cflocale in this file.  CFLocaleCopyCurrent() can hence go away and nullptr can be used where locale is needed.



2.

ms-jihua:
Summary of current CoreText performance issues after #1558 and assorted recommendations:

UIKit functions to watch:
- [NSString drawAtPoint:]()/NSString+UIKitAdditions.mm::drawString()
- [NSString sizeWithAttributes:]
- [NSLayoutManager __layoutAllText]

Bottlenecks:
- CFRelease()
    - About 20% of [NSString drawAtPoint:] is spent in CFRelease(). We discussed this offline before, but our CFRelease is very slow compared to our objective-C versions. We have some code cooking to address this, but the engineering hurdle to complete the effort seems high.

- _DWriteGetFrame()/CTFramesetterCreateFrame()/CTFramesetterSuggestFrameSizeWithConstraints()
    - Latter two functions call _DWriteGetFrame()

    - Very expensive. We spend 20% of [NSString drawAtPoint:], more than 60% of [NSString sizeWithAttributes:], and more than 70% of [NSLayoutManager __layoutAllText] (depending on number of line segments) here.

    - About 50% of _DWriteGetFrame() time is spent in IDWriteTextLayout::Draw(), which creates a TextLayout by analyzing our input and laying our the text. We should follow up with DWrite team as to whether we can make gains by structuring our input differently.

[...]

Issue #1207, opened 2016-10-21
DWriteWrapper, DWriteWrapper_CoreText require specific include ordering


ms-jihua:
# include <COMIncludes.h>
# import <wrl/implements.h>
# include <COMIncludes_End.h>

needs to come first for some reason. Investigate why this is the case, whether followup action is needed.

rajsesh-msft:
Since this is internal only, it is not worth spending time on this right now.

PR #1504, opened 2016-12-02
Refactor some DWriteWrapper code

### LINKED ISSUE #1212: CoreText post-merge cleanup

ms-jihua:
- Split DWriteWrapper, DWriteWrapper_CoreText among more files
  - DWriteFontBinaryDataLoader is now in its own .h + .mm along with BinaryDataStream
  - DWriteFontCollectionLoader is now in its own .h + .mm along with FileEnumerator
    - Renamed to DWriteFontBinaryDataCollectionLoader to get the name farther from the interface name (IDWriteFontCollectionLoader)
  - Implementation for _DWriteFontCreatePathForGlyph(), and the __CGPathGeometrySink class,
    which was only used in this function, are now in a separate .mm file
  - Implementation for DWriteFont_CoreText functions relating to CTFont, CTFontDescriptor are now in a separate .mm file
  - Refactored _userFontCollection, s_userFontPropertiesMap, systemFontPropertiesMap to be behind an abstraction layer,
    DWriteFontCollectionHelper, which wraps around either a user or system IDWriteFontCollection and provides
    font name -> other name/family name, properties mappings.
    - This is in its own .h + .mm file
    - As a result, many _DWrite functions now reduce cleanly to
      __GetSystemFontCollectionHelper()->Foo(), lock, __GetUserFontCollectionHelper()->Foo()

- Refactored __CreateDWriteTextFormat(), __CreateDWriteTextLayout()
  - Removed export of _DWriteCreateTextFormat(), this is now done directly in DWriteWrapper_CoreText.mm
  - Now return HRESULT instead of ComPtr, now create through out-pointers, changed naming convention
    - As a result, __CreateDWriteTextFormat no longer always returns a nullptr if an error occurs -
      this should not impact anything, as this was an internal function, and return values were never checked for nullptr anyway
  - Use CFAttributedStringGetAttribute directly instead of copying to a dictionary first,
    this should save some time on redundant copies/copies of unused attributes
  - Parts of these functions were moved into separate inline functions for easier reading

- Reversed order of i, j in _DWriteGetFrame(), such that i is on the outside and j on the inside
  - Replaced the inner two while loops, with a single for loop

- Exported _GetDefaultUserLocaleName()
- Broke parts of _DWriteCreateFontFaceWithFontDescriptor() into separate functions for easier readability (less nesting)
- Moved CustomDWriteTextRenderer member functions bodies to inside the class body
- _CFStringCreateUppercaseCopy now uses a nullptr CFLocale (faster, font names are not localized anyway)
  - Removed a redundant call to CFStringCreateUppercaseCopy() in the initialization of a font property map

- Changed most remaining instances where CFAutorelease was used to use woc::unique_cf instead
- Standardized naming convention of functions that translate a DWrite constant to CT/CG, or vice-versa
- Changed members of most C++ classes to use the m_ naming convention
- Standardized most file-private functions to be static
- Standardized most instances of loops that call functions that return HRESULTs to call continue on failure
- Changed CTTypesetterCreateLineWithOffset() to log and return null instead of throw
- Added a bunch of SAL annotations
- Added a string helper for copying a CFString to a vector, this was a pattern that occurred multiple times
- Added comments in various places

Fixes #1212

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1504)

### DIFF
--- a/Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.h
+++ b/Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.h
@@ -35,12 +35,10 @@ class DWriteFontBinaryDataCollectionLoader
     : public Microsoft::WRL::RuntimeClass<Microsoft::WRL::RuntimeClassFlags<Microsoft::WRL::WinRtClassicComMix>,
                                           IDWriteFontCollectionLoader> {
 protected:
-    InspectableClass(L"Windows.Bridge.DirectWrite", TrustLevel::BaseTrust);
+    InspectableClass(L"Windows.Bridge.DirectWrite.DWriteFontBinaryDataCollectionLoader", TrustLevel::BaseTrust);
 
 public:
-    DWriteFontBinaryDataCollectionLoader() {
-    }
-
+    DWriteFontBinaryDataCollectionLoader();
     HRESULT RuntimeClassInitialize();
 
     HRESULT STDMETHODCALLTYPE CreateEnumeratorFromKey(_In_ IDWriteFactory* factory,
@@ -52,13 +50,17 @@ public:
     HRESULT AddDatas(CFArrayRef fontDatas, CFArrayRef* errors);
     HRESULT RemoveDatas(CFArrayRef fontDatas, CFArrayRef* errors);
 
-private:
-    // Array of CFDataRef containing data of fonts, in order of being added
-    woc::unique_cf<CFMutableArrayRef> m_fontDatas;
+    // Returns a DWriteFontFile corresponding to the index-th CFDataRef in internal storage
+    HRESULT GetFontFileAt(int index, IDWriteFontFile** outFontFile);
 
+    // Returns number of files currently in internal storage
+    size_t GetFontFileCount();
+
+private:
     // Set of CFDataRef containing data of fonts, used to simplify checking if font has been added
     woc::unique_cf<CFMutableSetRef> m_fontDatasSet;
 
-    // Array of previously created font files, which saves us from having to read a file multiple times
-    std::vector<Microsoft::WRL::ComPtr<IDWriteFontFile>> m_previouslyCreatedFiles;
+    // List of CFDataRefs registered to this font collection loader through AddDatas()
+    // As an optimization, a font data may cache an IDWriteFontFile that was created using it, for fast access if the font is needed again
+    std::vector<std::pair<woc::unique_cf<CFDataRef>, Microsoft::WRL::ComPtr<IDWriteFontFile>>> m_fontDatas;
 };
--- a/Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.h
+++ b/Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.h
@@ -25,7 +25,7 @@
 #import <CoreFoundation/CFArray.h>
 #import <CoreFoundation/CFData.h>
 
-#import <vector>
+#import <unordered_map>
 
 /**
  * Custom implementation of IDWriteFontCollectionLoader, which creates an IDWriteFontFileEnumerator to create font files
@@ -39,7 +39,6 @@ protected:
 
 public:
     DWriteFontBinaryDataCollectionLoader();
-    HRESULT RuntimeClassInitialize();
 
     HRESULT STDMETHODCALLTYPE CreateEnumeratorFromKey(_In_ IDWriteFactory* factory,
                                                       _In_ const void* collectionKey,
@@ -50,17 +49,25 @@ public:
     HRESULT AddDatas(CFArrayRef fontDatas, CFArrayRef* errors);
     HRESULT RemoveDatas(CFArrayRef fontDatas, CFArrayRef* errors);
 
-    // Returns a DWriteFontFile corresponding to the index-th CFDataRef in internal storage
-    HRESULT GetFontFileAt(int index, IDWriteFontFile** outFontFile);
-
-    // Returns number of files currently in internal storage
-    size_t GetFontFileCount();
-
 private:
-    // Set of CFDataRef containing data of fonts, used to simplify checking if font has been added
-    woc::unique_cf<CFMutableSetRef> m_fontDatasSet;
+    struct __CFDataHashEqual {
+        std::size_t operator()(const woc::unique_cf<CFDataRef>& data) const {
+            return CFHash(data.get());
+        }
 
-    // List of CFDataRefs registered to this font collection loader through AddDatas()
-    // As an optimization, a font data may cache an IDWriteFontFile that was created using it, for fast access if the font is needed again
-    std::vector<std::pair<woc::unique_cf<CFDataRef>, Microsoft::WRL::ComPtr<IDWriteFontFile>>> m_fontDatas;
+        bool operator()(const woc::unique_cf<CFDataRef>& data1, const woc::unique_cf<CFDataRef>& data2) const {
+            return CFEqual(data1.get(), data2.get());
+        }
+    };
+
+    // Hash set of CFDataRefs registered to this font collection loader through AddDatas(),
+    // mapped to a IDWriteFontFile that was created using it
+    // This latter element is lazily initialized when it is first requested, then cached in the map
+    std::unordered_map<woc::unique_cf<CFDataRef>, Microsoft::WRL::ComPtr<IDWriteFontFile>, __CFDataHashEqual, __CFDataHashEqual>
+        m_fontDatas;
+
+public:
+    decltype(m_fontDatas)::iterator GetDataBegin();
+    decltype(m_fontDatas)::iterator GetDataEnd();
+    HRESULT GetFontFileAt(decltype(m_fontDatas)::iterator it, IDWriteFontFile** outFontFile);
 };
--- a/Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.h
+++ b/Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.h
@@ -14,6 +14,8 @@
 //
 //******************************************************************************
 
+#pragma once
+
 #include <COMIncludes.h>
 #import <wrl/implements.h>
 #include <COMIncludes_End.h>
--- a/Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.h
+++ b/Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.h
@@ -1,0 +1,62 @@
+//******************************************************************************
+//
+// Copyright (c) Microsoft. All rights reserved.
+//
+// This code is licensed under the MIT License (MIT).
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#include <COMIncludes.h>
+#import <wrl/implements.h>
+#include <COMIncludes_End.h>
+
+#import <Starboard.h>
+#import <CoreGraphics/DWriteWrapper.h>
+#import <CoreFoundation/CFArray.h>
+#import <CoreFoundation/CFData.h>
+
+#import <vector>
+
+/**
+ * Custom implementation of IDWriteFontCollectionLoader, which creates an IDWriteFontFileEnumerator to create font files
+ * Which are used to create a font collection.  This is expected to be a singleton.
+ */
+class DWriteFontBinaryDataCollectionLoader
+    : public Microsoft::WRL::RuntimeClass<Microsoft::WRL::RuntimeClassFlags<Microsoft::WRL::WinRtClassicComMix>,
+                                          IDWriteFontCollectionLoader> {
+protected:
+    InspectableClass(L"Windows.Bridge.DirectWrite", TrustLevel::BaseTrust);
+
+public:
+    DWriteFontBinaryDataCollectionLoader() {
+    }
+
+    HRESULT RuntimeClassInitialize();
+
+    HRESULT STDMETHODCALLTYPE CreateEnumeratorFromKey(_In_ IDWriteFactory* factory,
+                                                      _In_ const void* collectionKey,
+                                                      _In_ UINT32 collectionKeySize,
+                                                      _Out_ IDWriteFontFileEnumerator** enumerator);
+
+    // Adds/removes an array of font data from this collection loader's internal storage
+    HRESULT AddDatas(CFArrayRef fontDatas, CFArrayRef* errors);
+    HRESULT RemoveDatas(CFArrayRef fontDatas, CFArrayRef* errors);
+
+private:
+    // Array of CFDataRef containing data of fonts, in order of being added
+    woc::unique_cf<CFMutableArrayRef> m_fontDatas;
+
+    // Set of CFDataRef containing data of fonts, used to simplify checking if font has been added
+    woc::unique_cf<CFMutableSetRef> m_fontDatasSet;
+
+    // Array of previously created font files, which saves us from having to read a file multiple times
+    std::vector<Microsoft::WRL::ComPtr<IDWriteFontFile>> m_previouslyCreatedFiles;
+};
--- a/Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.mm
+++ b/Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.mm
@@ -43,61 +43,40 @@ static inline void __AppendNullptrIfExists(CFMutableArrayRef errors) {
  */
 class DWriteFontFileEnumerator : public RuntimeClass<RuntimeClassFlags<WinRtClassicComMix>, IDWriteFontFileEnumerator> {
 protected:
-    InspectableClass(L"Windows.Bridge.DirectWrite", TrustLevel::BaseTrust);
+    InspectableClass(L"Windows.Bridge.DirectWrite.DWriteFontFileEnumerator", TrustLevel::BaseTrust);
 
 public:
-    DWriteFontFileEnumerator() {
+    DWriteFontFileEnumerator(DWriteFontBinaryDataCollectionLoader* parent) {
+        m_parent = parent;
     }
 
-    HRESULT RuntimeClassInitialize(CFArrayRef fontDatas, std::vector<ComPtr<IDWriteFontFile>>* previouslyCreatedFiles) {
-        CFRetain(fontDatas);
-        m_fontDatas.reset(fontDatas);
-        m_previouslyCreatedFiles = previouslyCreatedFiles;
+    HRESULT RuntimeClassInitialize() {
         return S_OK;
     }
 
-    HRESULT STDMETHODCALLTYPE GetCurrentFontFile(_Out_ IDWriteFontFile** fontFile) {
-        RETURN_HR_IF_NULL(E_POINTER, fontFile);
-        RETURN_HR_IF(E_ILLEGAL_METHOD_CALL, m_location < 0 || m_location > CFArrayGetCount(m_fontDatas.get()));
-
-        if (0 <= m_location && m_location < m_previouslyCreatedFiles->size()) {
-            m_previouslyCreatedFiles->at(m_location).CopyTo(fontFile);
-        } else {
-            ComPtr<IDWriteFactory> dwriteFactory;
-            RETURN_IF_FAILED(DWriteCreateFactory(DWRITE_FACTORY_TYPE_SHARED, __uuidof(IDWriteFactory), &dwriteFactory));
-
-            ComPtr<DWriteFontBinaryDataLoader> loader;
-            RETURN_IF_FAILED(MakeAndInitialize<DWriteFontBinaryDataLoader>(&loader,
-                                                                           static_cast<CFDataRef>(
-                                                                               CFArrayGetValueAtIndex(m_fontDatas.get(), m_location))));
-            RETURN_IF_FAILED(dwriteFactory->RegisterFontFileLoader(loader.Get()));
-
-            int unused;
-            RETURN_IF_FAILED(dwriteFactory->CreateCustomFontFileReference(&unused, sizeof(unused), loader.Get(), fontFile));
-            m_previouslyCreatedFiles->emplace_back(*fontFile);
-        }
-
-        return S_OK;
+    HRESULT STDMETHODCALLTYPE GetCurrentFontFile(_Out_ IDWriteFontFile** outFontFile) {
+        return m_parent->GetFontFileAt(m_location, outFontFile);
     }
 
     HRESULT STDMETHODCALLTYPE MoveNext(_Out_ BOOL* hasCurrentFile) {
         RETURN_HR_IF_NULL(E_POINTER, hasCurrentFile);
-        *hasCurrentFile = ++m_location < CFArrayGetCount(m_fontDatas.get());
+        *hasCurrentFile = ++m_location < m_parent->GetFontFileCount();
         return S_OK;
     }
 
 private:
     int m_location = -1;
 
-    woc::unique_cf<CFArrayRef> m_fontDatas;
-    std::vector<ComPtr<IDWriteFontFile>>* m_previouslyCreatedFiles;
+    ComPtr<DWriteFontBinaryDataCollectionLoader> m_parent;
 };
 
 // DWriteFontBinaryDataCollectionLoader member functions
 
-HRESULT DWriteFontBinaryDataCollectionLoader::DWriteFontBinaryDataCollectionLoader::RuntimeClassInitialize() {
-    m_fontDatas.reset(CFArrayCreateMutable(kCFAllocatorDefault, 0, &kCFTypeArrayCallBacks));
+DWriteFontBinaryDataCollectionLoader::DWriteFontBinaryDataCollectionLoader() {
     m_fontDatasSet.reset(CFSetCreateMutable(nullptr, 0, &kCFTypeSetCallBacks));
+}
+
+HRESULT DWriteFontBinaryDataCollectionLoader::DWriteFontBinaryDataCollectionLoader::RuntimeClassInitialize() {
     return S_OK;
 }
 
@@ -105,17 +84,14 @@ HRESULT STDMETHODCALLTYPE DWriteFontBinaryDataCollectionLoader::CreateEnumerator
                                                                                         _In_ const void* collectionKey,
                                                                                         _In_ UINT32 collectionKeySize,
                                                                                         _Out_ IDWriteFontFileEnumerator** enumerator) {
-    MakeAndInitialize<DWriteFontFileEnumerator>(enumerator, m_fontDatas.get(), &m_previouslyCreatedFiles);
+    *enumerator = Make<DWriteFontFileEnumerator>(this).Detach();
     return S_OK;
 }
 
 HRESULT DWriteFontBinaryDataCollectionLoader::AddDatas(CFArrayRef fontDatas, CFArrayRef* errors) {
-    CFMutableArrayRef outErrors = nil;
+    // Don't bother creating this array if errors is null
+    CFMutableArrayRef outErrors = errors ? CFArrayCreateMutable(nullptr, 0, &kCFTypeArrayCallBacks) : nil;
     HRESULT ret = S_OK;
-    if (errors) {
-        outErrors = CFArrayCreateMutable(nullptr, 0, &kCFTypeArrayCallBacks);
-        *errors = outErrors;
-    }
 
     CFIndex count = CFArrayGetCount(fontDatas);
     for (CFIndex i = 0; i < count; ++i) {
@@ -125,7 +101,11 @@ HRESULT DWriteFontBinaryDataCollectionLoader::AddDatas(CFArrayRef fontDatas, CFA
                 __AppendErrorIfExists(outErrors, kCTFontManagerErrorAlreadyRegistered);
                 ret = S_FALSE;
             } else {
-                CFArrayAppendValue(m_fontDatas.get(), data);
+                m_fontDatas.push_back({});
+
+                CFRetain(data);
+                m_fontDatas.back().first.reset(data);
+
                 CFSetAddValue(m_fontDatasSet.get(), data);
                 __AppendNullptrIfExists(outErrors);
             }
@@ -135,16 +115,17 @@ HRESULT DWriteFontBinaryDataCollectionLoader::AddDatas(CFArrayRef fontDatas, CFA
         }
     }
 
+    if (errors) {
+        *errors = outErrors;
+    }
+
     return ret;
 }
 
 HRESULT DWriteFontBinaryDataCollectionLoader::RemoveDatas(CFArrayRef fontDatas, CFArrayRef* errors) {
-    CFMutableArrayRef outErrors = nil;
+    // Don't bother creating this array if errors is null
+    CFMutableArrayRef outErrors = errors ? CFArrayCreateMutable(nullptr, 0, &kCFTypeArrayCallBacks) : nil;
     HRESULT ret = S_OK;
-    if (errors) {
-        outErrors = CFArrayCreateMutable(nullptr, 0, &kCFTypeArrayCallBacks);
-        *errors = outErrors;
-    }
 
     CFIndex count = CFArrayGetCount(fontDatas);
     for (CFIndex i = 0; i < count; ++i) {
@@ -152,9 +133,14 @@ HRESULT DWriteFontBinaryDataCollectionLoader::RemoveDatas(CFArrayRef fontDatas, 
         if (data) {
             if (CFSetContainsValue(m_fontDatasSet.get(), data)) {
                 CFSetRemoveValue(m_fontDatasSet.get(), data);
-                CFIndex index = CFArrayGetFirstIndexOfValue(m_fontDatas.get(), { 0, CFArrayGetCount(m_fontDatas.get()) }, data);
-                CFArrayRemoveValueAtIndex(m_fontDatas.get(), index);
-                m_previouslyCreatedFiles.erase(m_previouslyCreatedFiles.begin() + index);
+
+                for (auto it = m_fontDatas.begin(); it != m_fontDatas.end(); ++it) {
+                    if (CFEqual(it->first.get(), data)) {
+                        m_fontDatas.erase(it);
+                        break;
+                    }
+                }
+
                 __AppendNullptrIfExists(outErrors);
             } else {
                 __AppendErrorIfExists(outErrors, kCTFontManagerErrorNotRegistered);
@@ -166,5 +152,35 @@ HRESULT DWriteFontBinaryDataCollectionLoader::RemoveDatas(CFArrayRef fontDatas, 
         }
     }
 
+    if (errors) {
+        *errors = outErrors;
+    }
+
     return ret;
+}
+
+HRESULT DWriteFontBinaryDataCollectionLoader::GetFontFileAt(int index, IDWriteFontFile** outFontFile) {
+    RETURN_HR_IF_NULL(E_POINTER, outFontFile);
+    RETURN_HR_IF(E_ILLEGAL_METHOD_CALL, index < 0 || index > m_fontDatas.size());
+
+    auto& it = m_fontDatas.at(index);
+
+    // Create the IDWriteFontFile in the cache, if it's not already cached
+    if (!it.second) {
+        ComPtr<IDWriteFactory> dwriteFactory;
+        RETURN_IF_FAILED(DWriteCreateFactory(DWRITE_FACTORY_TYPE_SHARED, __uuidof(IDWriteFactory), &dwriteFactory));
+
+        ComPtr<DWriteFontBinaryDataLoader> loader;
+        RETURN_IF_FAILED(MakeAndInitialize<DWriteFontBinaryDataLoader>(&loader, it.first.get()));
+        RETURN_IF_FAILED(dwriteFactory->RegisterFontFileLoader(loader.Get()));
+
+        int unused;
+        RETURN_IF_FAILED(dwriteFactory->CreateCustomFontFileReference(&unused, sizeof(unused), loader.Get(), &it.second));
+    }
+
+    return it.second.CopyTo(outFontFile);
+}
+
+size_t DWriteFontBinaryDataCollectionLoader::GetFontFileCount() {
+    return m_fontDatas.size();
 }
--- a/Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.mm
+++ b/Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.mm
@@ -1,0 +1,168 @@
+//******************************************************************************
+//
+// Copyright (c) Microsoft. All rights reserved.
+//
+// This code is licensed under the MIT License (MIT).
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#import "DWriteFontBinaryDataCollectionLoader.h"
+#import "DWriteFontBinaryDataLoader.h"
+
+#import <CoreFoundation/CFError.h>
+
+using namespace Microsoft::WRL;
+
+/**
+ * Private helpers to append a CFErrorRef or nullptr to the end of a CFMutableArray if it exists
+ */
+static inline void __AppendErrorIfExists(CFMutableArrayRef errors, CFIndex errorCode) {
+    if (errors) {
+        woc::unique_cf<CFErrorRef> error{ CFErrorCreate(nullptr, kCFErrorDomainCocoa, errorCode, nullptr) };
+        CFArrayAppendValue(errors, error.get());
+    }
+}
+
+static inline void __AppendNullptrIfExists(CFMutableArrayRef errors) {
+    if (errors) {
+        CFArrayAppendValue(errors, nullptr);
+    }
+}
+
+/**
+ * Private class.
+ * Custom implementation of IDWriteFontFileEnumerator, which enumerates over internal data and returns a font file
+ */
+class DWriteFontFileEnumerator : public RuntimeClass<RuntimeClassFlags<WinRtClassicComMix>, IDWriteFontFileEnumerator> {
+protected:
+    InspectableClass(L"Windows.Bridge.DirectWrite", TrustLevel::BaseTrust);
+
+public:
+    DWriteFontFileEnumerator() {
+    }
+
+    HRESULT RuntimeClassInitialize(CFArrayRef fontDatas, std::vector<ComPtr<IDWriteFontFile>>* previouslyCreatedFiles) {
+        CFRetain(fontDatas);
+        m_fontDatas.reset(fontDatas);
+        m_previouslyCreatedFiles = previouslyCreatedFiles;
+        return S_OK;
+    }
+
+    HRESULT STDMETHODCALLTYPE GetCurrentFontFile(_Out_ IDWriteFontFile** fontFile) {
+        RETURN_HR_IF(E_ILLEGAL_METHOD_CALL, m_location < 0 || m_location > CFArrayGetCount(m_fontDatas.get()));
+
+        if (0 <= m_location && m_location < m_previouslyCreatedFiles->size()) {
+            m_previouslyCreatedFiles->at(m_location).CopyTo(fontFile);
+        } else {
+            ComPtr<IDWriteFactory> dwriteFactory;
+            RETURN_IF_FAILED(DWriteCreateFactory(DWRITE_FACTORY_TYPE_SHARED, __uuidof(IDWriteFactory), &dwriteFactory));
+
+            ComPtr<DWriteFontBinaryDataLoader> loader;
+            RETURN_IF_FAILED(MakeAndInitialize<DWriteFontBinaryDataLoader>(&loader,
+                                                                           static_cast<CFDataRef>(
+                                                                               CFArrayGetValueAtIndex(m_fontDatas.get(), m_location))));
+            RETURN_IF_FAILED(dwriteFactory->RegisterFontFileLoader(loader.Get()));
+
+            int unused;
+            RETURN_IF_FAILED(dwriteFactory->CreateCustomFontFileReference(&unused, sizeof(unused), loader.Get(), fontFile));
+            m_previouslyCreatedFiles->emplace_back(*fontFile);
+        }
+
+        return S_OK;
+    }
+
+    HRESULT STDMETHODCALLTYPE MoveNext(_Out_ BOOL* hasCurrentFile) {
+        *hasCurrentFile = ++m_location < CFArrayGetCount(m_fontDatas.get());
+        return S_OK;
+    }
+
+private:
+    int m_location = -1;
+
+    woc::unique_cf<CFArrayRef> m_fontDatas;
+    std::vector<ComPtr<IDWriteFontFile>>* m_previouslyCreatedFiles;
+};
+
+// DWriteFontBinaryDataCollectionLoader member functions
+
+HRESULT DWriteFontBinaryDataCollectionLoader::DWriteFontBinaryDataCollectionLoader::RuntimeClassInitialize() {
+    m_fontDatas.reset(CFArrayCreateMutable(kCFAllocatorDefault, 0, &kCFTypeArrayCallBacks));
+    m_fontDatasSet.reset(CFSetCreateMutable(nullptr, 0, &kCFTypeSetCallBacks));
+    return S_OK;
+}
+
+HRESULT STDMETHODCALLTYPE DWriteFontBinaryDataCollectionLoader::CreateEnumeratorFromKey(_In_ IDWriteFactory* factory,
+                                                                                        _In_ const void* collectionKey,
+                                                                                        _In_ UINT32 collectionKeySize,
+                                                                                        _Out_ IDWriteFontFileEnumerator** enumerator) {
+    MakeAndInitialize<DWriteFontFileEnumerator>(enumerator, m_fontDatas.get(), &m_previouslyCreatedFiles);
+    return S_OK;
+}
+
+HRESULT DWriteFontBinaryDataCollectionLoader::AddDatas(CFArrayRef fontDatas, CFArrayRef* errors) {
+    CFMutableArrayRef outErrors = nil;
+    HRESULT ret = S_OK;
+    if (errors) {
+        outErrors = CFArrayCreateMutable(nullptr, 0, &kCFTypeArrayCallBacks);
+        *errors = outErrors;
+    }
+
+    CFIndex count = CFArrayGetCount(fontDatas);
+    for (CFIndex i = 0; i < count; ++i) {
+        CFDataRef data = static_cast<CFDataRef>(CFArrayGetValueAtIndex(fontDatas, i));
+        if (data) {
+            if (CFSetContainsValue(m_fontDatasSet.get(), data)) {
+                __AppendErrorIfExists(outErrors, kCTFontManagerErrorAlreadyRegistered);
+                ret = S_FALSE;
+            } else {
+                CFArrayAppendValue(m_fontDatas.get(), data);
+                CFSetAddValue(m_fontDatasSet.get(), data);
+                __AppendNullptrIfExists(outErrors);
+            }
+        } else {
+            __AppendErrorIfExists(outErrors, kCTFontManagerErrorInvalidFontData);
+            ret = S_FALSE;
+        }
+    }
+
+    return ret;
+}
+
+HRESULT DWriteFontBinaryDataCollectionLoader::RemoveDatas(CFArrayRef fontDatas, CFArrayRef* errors) {
+    CFMutableArrayRef outErrors = nil;
+    HRESULT ret = S_OK;
+    if (errors) {
+        outErrors = CFArrayCreateMutable(nullptr, 0, &kCFTypeArrayCallBacks);
+        *errors = outErrors;
+    }
+
+    CFIndex count = CFArrayGetCount(fontDatas);
+    for (CFIndex i = 0; i < count; ++i) {
+        CFDataRef data = static_cast<CFDataRef>(CFArrayGetValueAtIndex(fontDatas, i));
+        if (data) {
+            if (CFSetContainsValue(m_fontDatasSet.get(), data)) {
+                CFSetRemoveValue(m_fontDatasSet.get(), data);
+                CFIndex index = CFArrayGetFirstIndexOfValue(m_fontDatas.get(), { 0, CFArrayGetCount(m_fontDatas.get()) }, data);
+                CFArrayRemoveValueAtIndex(m_fontDatas.get(), index);
+                m_previouslyCreatedFiles.erase(m_previouslyCreatedFiles.begin() + index);
+                __AppendNullptrIfExists(outErrors);
+            } else {
+                __AppendErrorIfExists(outErrors, kCTFontManagerErrorNotRegistered);
+                ret = S_FALSE;
+            }
+        } else {
+            __AppendErrorIfExists(outErrors, kCTFontManagerErrorInvalidFontData);
+            ret = S_FALSE;
+        }
+    }
+
+    return ret;
+}
--- a/Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.mm
+++ b/Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.mm
@@ -57,6 +57,7 @@ public:
     }
 
     HRESULT STDMETHODCALLTYPE GetCurrentFontFile(_Out_ IDWriteFontFile** fontFile) {
+        RETURN_HR_IF_NULL(E_POINTER, fontFile);
         RETURN_HR_IF(E_ILLEGAL_METHOD_CALL, m_location < 0 || m_location > CFArrayGetCount(m_fontDatas.get()));
 
         if (0 <= m_location && m_location < m_previouslyCreatedFiles->size()) {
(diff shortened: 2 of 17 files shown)


msft-Jeyaram:
:D

Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.mm
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    HRESULT RuntimeClassInitialize(CFArrayRef fontDatas, std::vector<ComPtr<IDWriteFontFile>>* previouslyCreatedFiles) {
+        CFRetain(fontDatas);
+        m_fontDatas.reset(fontDatas);
+        m_previouslyCreatedFiles = previouslyCreatedFiles;
+        return S_OK;
+    }
+
+    HRESULT STDMETHODCALLTYPE GetCurrentFontFile(_Out_ IDWriteFontFile** fontFile) {
```
msft-Jeyaram: >fontFile [](start = 73, length = 8)

validity check.

msft-Jeyaram:
++ Also thank you for draining the nestlings :)

Frameworks/CoreGraphics/DWriteFontBinaryDataLoader.mm
```diff
@@ -0,0 +1,119 @@
...
+    uint64_t m_lastWriteTime;
+};
+
+// DWriteFontBinaryDataLoader member functions
+
+DWriteFontBinaryDataLoader::DWriteFontBinaryDataLoader() {
+    // Deliberate no-op
+}
+
+HRESULT DWriteFontBinaryDataLoader::RuntimeClassInitialize(CFDataRef data) {
```
msft-Jeyaram: i'm being too picky here, but what if data was null.
Is this a valid success?

Frameworks/CoreGraphics/DWriteFontBinaryDataLoader.mm
```diff
@@ -0,0 +1,119 @@
...
+ */
+class DWriteFontBinaryDataStream : public RuntimeClass<RuntimeClassFlags<WinRtClassicComMix>, IDWriteFontFileStream> {
+protected:
+    InspectableClass(L"Windows.Bridge.DirectWrite", TrustLevel::BaseTrust);
+
+public:
+    DWriteFontBinaryDataStream() {
+    }
+
+    HRESULT RuntimeClassInitialize(CFDataRef data) {
```
msft-Jeyaram: same question as below.
I assume null data is not valid?

Frameworks/CoreGraphics/DWriteFontBinaryDataLoader.mm
```diff
@@ -0,0 +1,119 @@
...
+    m_data.reset(data);
+    return S_OK;
+}
+
+// Ignores first two params, just return the same kind of stream always
+// Stream returned is dictated by dataProvider passed in at initialization time
+HRESULT STDMETHODCALLTYPE DWriteFontBinaryDataLoader::CreateStreamFromKey(_In_ const void* fontFileReferenceKey,
+                                                                          uint32_t fontFileReferenceKeySize,
+                                                                          _Out_ IDWriteFontFileStream** fontFileStream) {
+    RETURN_HR_IF_NULL(E_INVALIDARG, fontFileStream);
```
ms-jihua: >E_INVALIDARG [](start = 22, length = 12)

debating whether this should be E_POINTER instead...

msft-Jeyaram: Usually __OUT__ invalid args are pretty much: E_POINTER
that's used  prominently  in windows.

---
In reply to: [90736872](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90736872) [](ancestors = 90736872)

Frameworks/CoreGraphics/DWriteFontBinaryDataLoader.mm
ms-jihua: yeah, but since this is only used from the below class, it's sufficient that we validate below.

---
In reply to: [90736864](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90736864) [](ancestors = 90736864)

Frameworks/CoreText/DWriteWrapper_CoreText.mm
```diff
@@ -214,77 +261,58 @@ bool _CloneDWriteGlyphRun(_In_ DWRITE_GLYPH_RUN const* src, _Out_ DWRITE_GLYPH_R
...
         // Forces run breaks without interfering with any layout features
         // Necessary for attributes which DWrite does not support during layout (e.g. Color)
-        RETURN_NULL_IF_FAILED(typography->AddFontFeature({ DWRITE_FONT_FEATURE_TAG_DEFAULT, ++incompatibleAttributeFlag }));
-        RETURN_NULL_IF_FAILED(textLayout->SetTypography(typography.Get(), dwriteRange));
+        RETURN_IF_FAILED(typography->AddFontFeature({ DWRITE_FONT_FEATURE_TAG_DEFAULT, ++incompatibleAttributeFlag }));
+        RETURN_IF_FAILED(textLayout->SetTypography(typography.Get(), dwriteRange));
     }
 
-    return textLayout;
+    *outTextLayout = textLayout.Detach();
```
msft-Jeyaram: >xtLayout.Detach(); [](start = 23, length = 18)

use CopyTo

Frameworks/CoreText/DWriteWrapper_CoreText.mm
```diff
@@ -196,15 +240,18 @@ bool _CloneDWriteGlyphRun(_In_ DWRITE_GLYPH_RUN const* src, _Out_ DWRITE_GLYPH_R
...
- * @return the created IDWriteTextLayout object.
+ * @return whether the creation was successful
  */
-static ComPtr<IDWriteTextLayout> __CreateDWriteTextLayout(CFAttributedStringRef string, CFRange range, CGRect frameSize) {
-    ComPtr<IDWriteTextFormat> textFormat = __CreateDWriteTextFormat(string, range);
+static HRESULT __DWriteTextLayoutCreate(CFAttributedStringRef string,
+                                        CFRange range,
+                                        CGRect frameSize,
+                                        _Out_ IDWriteTextLayout** outTextLayout) {
+    RETURN_HR_IF(E_INVALIDARG, !outTextLayout);
```
msft-Jeyaram: >E_INVALIDARG [](start = 17, length = 12)

invalid pointer.

Frameworks/CoreText/DWriteWrapper_CoreText.mm
```diff
@@ -97,97 +108,130 @@ bool _CloneDWriteGlyphRun(_In_ DWRITE_GLYPH_RUN const* src, _Out_ DWRITE_GLYPH_R
...
-                    break;
-            }
-
-            THROW_IF_FAILED(textFormat->SetWordWrapping(wrapping));
-        }
+        RETURN_IF_FAILED(__DWriteTextFormatApplyParagraphStyle(textFormat, settings));
     }
 
-    return textFormat;
+    *outTextFormat = textFormat.Detach();
```
msft-Jeyaram: >Detach [](start = 32, length = 6)

use CopyTo(...)

Frameworks/CoreGraphics/DWriteFontBinaryDataLoader.mm
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        if (fragmentStart) {
+            const uint8_t* underlyingBuffer = CFDataGetBytePtr(m_data.get());
+            *fragmentStart = reinterpret_cast<const void*>(&(underlyingBuffer[fileOffset]));
+        }
+
+        if (fragmentContext) {
+            // Deliberately unused: this is meant to be passed to ReleaseFileFragment() below to free part of the data,
+            // but this stream frees all the underlying CFData at once
```
msft-Jeyaram: Given it's an internal class.. meh i'll leave it up to you.

---
In reply to: [90737774](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90737774) [](ancestors = 90737774)

Frameworks/CoreGraphics/DWriteFontBinaryDataLoader.h
```diff
@@ -0,0 +1,43 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
```
ms-jihua: oh yeah! i knew i forgot something...

---
In reply to: [90737886](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90737886) [](ancestors = 90737886)

Frameworks/CoreGraphics/DWriteFontCollectionHelper.h
```diff
@@ -0,0 +1,60 @@
...
+ */
+class DWriteFontCollectionHelper {
+public:
+    DWriteFontCollectionHelper() {
+    }
+
+    CFMutableArrayRef CopyFontFamilyNames();
+    CFMutableArrayRef CopyFontNamesForFamilyName(CFStringRef familyName);
+    std::shared_ptr<_DWriteFontProperties> GetFontPropertiesFromUppercaseFontName(const woc::unique_cf<CFStringRef>& upperFontName);
+    HRESULT CreateFontFamilyWithName(_In_ const wchar_t* unicharFamilyName, _Out_ IDWriteFontFamily** outFontFamily);
```
msft-Jeyaram: >_Out_ [](start = 76, length = 5)

as per SAL _Outptr_ is the most appropriate here.
use _Outptr_

msft-Jeyaram:
> // #1207: Do not move this block, it has to come first for some reason

given the task is closed, doesn't make sense to have it here anymore.
this mystery might be solved one day.

---
Refers to: Frameworks/CoreGraphics/DWriteWrapper.mm:17 in 4fad654. [](commit_id = 4fad65482028cad182459816aceab43dcd4ab981, deletion_comment = False)

Frameworks/CoreGraphics/DWriteFontBinaryDataLoader.mm
```diff
@@ -0,0 +1,119 @@
...
+// Stream returned is dictated by dataProvider passed in at initialization time
+HRESULT STDMETHODCALLTYPE DWriteFontBinaryDataLoader::CreateStreamFromKey(_In_ const void* fontFileReferenceKey,
+                                                                          uint32_t fontFileReferenceKeySize,
+                                                                          _Out_ IDWriteFontFileStream** fontFileStream) {
+    RETURN_HR_IF_NULL(E_INVALIDARG, fontFileStream);
+
+    ComPtr<DWriteFontBinaryDataStream> ret;
+    RETURN_IF_FAILED(MakeAndInitialize<DWriteFontBinaryDataStream>(&ret, m_data.get()));
+
+    *fontFileStream = ret.Detach();
```
DHowett-MSFT: That's likely acceptable :)

---
In reply to: [90742277](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90742277) [](ancestors = 90742277,90737178)

Frameworks/include/CoreGraphics/DWriteWrapper.h
```diff
@@ -29,12 +29,17 @@
 #import <memory>
 
 // General DWrite helpers
-COREGRAPHICS_EXPORT CFStringRef _CFStringFromLocalizedString(IDWriteLocalizedStrings* localizedString);
+#ifdef __cplusplus
+extern "C++" std::wstring _GetUserDefaultLocaleName();
+#endif
```
bbowman: not sure how I feel about this. Why are we cross module exporting c++ things?

Frameworks/include/CoreGraphics/DWriteWrapper.h
```diff
@@ -46,22 +51,18 @@ struct _DWriteFontProperties {
     woc::unique_cf<CFStringRef> familyName;
 };
 
-// Create DWrite objects
+// Font name <-> font family, font properties
 #ifdef __cplusplus
 extern "C++" std::shared_ptr<_DWriteFontProperties> _DWriteGetFontPropertiesFromName(CFStringRef fontName);
 #endif
```
bbowman: also here.

Frameworks/include/StringHelpers.h
```diff
@@ -465,9 +466,22 @@ inline TSource ContainerFromDelimitedString(_In_ const std::wstring& delimitedSt
     return result;
 }
 
-// Returns a a vector from a delmited string
+// Returns a vector from a delmited string
 inline std::vector<std::wstring> VectorFromDelimitedString(_In_ const std::wstring& delimitedString, wchar_t delimiter) {
     return ContainerFromDelimitedString<std::vector<std::wstring>>(delimitedString, delimiter);
 }
 
+#ifdef __OBJC__
```
bbowman: >__OBJC__ [](start = 7, length = 8)

is this actually needed? CF should be useable from straight c / c++ right?

Frameworks/CoreText/DWriteWrapper_CoreText.mm
```diff
@@ -52,10 +46,26 @@ static DWRITE_TEXT_ALIGNMENT __CoreTextAlignmentToDwrite(CTTextAlignment alignme
     }
 }
 
+static DWRITE_WORD_WRAPPING __CTLineBreakModeToDWrite(CTLineBreakMode lineBreakMode) {
```
bbowman: >__CTLineBreakModeToDWrite [](start = 28, length = 25)

whats the coding convention on _ vs __ as a prefix?

Frameworks/CoreGraphics/DWriteFontCollectionHelper.mm
```diff
@@ -0,0 +1,207 @@
...
+HRESULT DWriteFontCollectionHelper::CreateFontFamilyWithName(_In_ const wchar_t* unicharFamilyName,
+                                                             _Outptr_ IDWriteFontFamily** outFontFamily) {
+    ComPtr<IDWriteFontCollection> fontCollection = _GetFontCollection();
+
+    // If no/empty font collection, say that the font family couldn't be found
+    // Use S_FALSE to represent 'not found'
+    RETURN_HR_IF_NULL(S_FALSE, fontCollection);
+
+    size_t fontFamilyIndex;
+    BOOL fontFamilyExists;
```
bbowman: uninitialized

Frameworks/CoreGraphics/DWriteFontCollectionHelper.h
bbowman: just curious why we decided to use SAL here. Do we have uniform conventions about this?

---
In reply to: [90738161](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90738161) [](ancestors = 90738161)

Frameworks/CoreGraphics/DWriteFontBinaryDataLoader.mm
```diff
@@ -0,0 +1,120 @@
...
+protected:
+    InspectableClass(L"Windows.Bridge.DirectWrite", TrustLevel::BaseTrust);
+
+public:
+    DWriteFontBinaryDataStream() {
+    }
+
+    HRESULT RuntimeClassInitialize(CFDataRef data) {
+        CFRetain(data);
+        m_data.reset(data);
```
bbowman: this pattern suggests an issue in the design here. Should we create a shared_cf? or maybe more correctly just auto_cf?

Frameworks/CoreGraphics/DWriteFontBinaryDataLoader.mm
```diff
@@ -0,0 +1,120 @@
...
+        // Just use current time for m_lastWriteTime
+        FILETIME fileTime;
+        GetSystemTimeAsFileTime(&fileTime);
+
+        // Concat filetime into a single uint64_t
+        m_lastWriteTime = 0;
+        m_lastWriteTime |= static_cast<uint64_t>(fileTime.dwLowDateTime);
+        m_lastWriteTime |= static_cast<uint64_t>(fileTime.dwHighDateTime) << 32;
+        return S_OK;
+    }
```
bbowman: RuntimeClassInitialize is useful if failure is possible. That doesn't seem to be the case here. Why not just make a ctor with args?

Frameworks/CoreGraphics/DWriteFontBinaryDataLoader.mm
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        if (fragmentContext) {
+            *fragmentContext = nullptr;
+        }
+
+        return S_OK;
+    };
+
+    void STDMETHODCALLTYPE ReleaseFileFragment(void* fragmentContext){
```
bbowman: >ReleaseFileFragment [](start = 27, length = 19)

I see what you are doing but should you reset m_data here? What is the contract supposed to be?

Frameworks/CoreGraphics/DWriteFontBinaryDataLoader.mm
```diff
@@ -0,0 +1,120 @@
...
+// Stream returned is dictated by dataProvider passed in at initialization time
+HRESULT STDMETHODCALLTYPE DWriteFontBinaryDataLoader::CreateStreamFromKey(_In_ const void* fontFileReferenceKey,
+                                                                          uint32_t fontFileReferenceKeySize,
+                                                                          _Out_ IDWriteFontFileStream** fontFileStream) {
+    RETURN_HR_IF_NULL(E_POINTER, fontFileStream);
+
+    ComPtr<DWriteFontBinaryDataStream> ret;
+    RETURN_IF_FAILED(MakeAndInitialize<DWriteFontBinaryDataStream>(&ret, m_data.get()));
+
+    return ret.CopyTo(fontFileStream);
```
bbowman: >CopyTo [](start = 15, length = 6)

prefer detach here if you aren't saving the reference internally for later. CopyTo adds an extra AddRef that technically isn't needed since you are *transferring* ownership.

Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.mm
```diff
@@ -0,0 +1,170 @@
...
+        RETURN_HR_IF_NULL(E_POINTER, hasCurrentFile);
+        *hasCurrentFile = ++m_location < CFArrayGetCount(m_fontDatas.get());
+        return S_OK;
+    }
+
+private:
+    int m_location = -1;
+
+    woc::unique_cf<CFArrayRef> m_fontDatas;
+    std::vector<ComPtr<IDWriteFontFile>>* m_previouslyCreatedFiles;
```
bbowman: >* [](start = 40, length = 1)

interesting decision to use a raw * here. What is the ownership contract of this vector?

Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.mm
```diff
@@ -0,0 +1,170 @@
...
+    MakeAndInitialize<DWriteFontFileEnumerator>(enumerator, m_fontDatas.get(), &m_previouslyCreatedFiles);
+    return S_OK;
+}
+
+HRESULT DWriteFontBinaryDataCollectionLoader::AddDatas(CFArrayRef fontDatas, CFArrayRef* errors) {
+    CFMutableArrayRef outErrors = nil;
+    HRESULT ret = S_OK;
+    if (errors) {
+        outErrors = CFArrayCreateMutable(nullptr, 0, &kCFTypeArrayCallBacks);
+        *errors = outErrors;
```
bbowman: >*errors = outErrors; [](start = 8, length = 20)

strange pattern. Typically you want to wait until the end to assign out parameters as returning a "half baked" structure in the case of failure can create confusion on the part of the caller.

Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.mm
```diff
@@ -0,0 +1,170 @@
...
+                __AppendErrorIfExists(outErrors, kCTFontManagerErrorAlreadyRegistered);
+                ret = S_FALSE;
+            } else {
+                CFArrayAppendValue(m_fontDatas.get(), data);
+                CFSetAddValue(m_fontDatasSet.get(), data);
+                __AppendNullptrIfExists(outErrors);
+            }
+        } else {
+            __AppendErrorIfExists(outErrors, kCTFontManagerErrorInvalidFontData);
+            ret = S_FALSE;
```
bbowman: >S_FALSE [](start = 18, length = 7)

This is my least favorite return ever. It is still a success value (S_* is success) so all the macros like RETURN_IF_FAILED won't trip. I'd recommend against a pattern that relies on this.

bbowman:
:clock1:

Frameworks/CoreGraphics/DWriteFontCollectionHelper.h
ms-jihua: It seemed more preferable than not using SAL, and we're dealing with COM out pointers which typically uses SAL in function signatures. I don't have a strong opinion on this, I can remove them if you feel it's better that way.

---
In reply to: [90743039](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90743039) [](ancestors = 90743039,90738161)

Frameworks/CoreText/DWriteWrapper_CoreText.mm
ms-jihua: __ is file-private, _ is framework-"private" (not intended to be used by general consumers but exported nonetheless)

---
In reply to: [90742638](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90742638) [](ancestors = 90742638)

Frameworks/CoreGraphics/DWriteFontCollectionHelper.mm
```diff
@@ -0,0 +1,207 @@
...
+            ComPtr<IDWriteFont> font;
+            if (FAILED(fontList->GetFont(j, &font))) {
+                continue;
+            }
+
+            // For each font, fill out a _DWriteFontProperties
+            auto info = std::make_shared<_DWriteFontProperties>();
+            ComPtr<IDWriteLocalizedStrings> displayName;
+            ComPtr<IDWriteLocalizedStrings> postScriptName;
+            BOOL exist;
```
ms-jihua: does this actually matter? I would expect GetInformationalStrings() to always populate this.

---
In reply to: [90742971](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90742971) [](ancestors = 90742971)

Frameworks/CoreGraphics/DWriteFontBinaryDataLoader.mm
ms-jihua: The IDWriteFontFileStream is that the stream is intended to be releasable little-by-little as it's consumed. ReadFileFragment() is expected to return information about the data that was read, through fragmentContext. ReleaseFileFragment() is supposed to then take this, to release part of the stream.

CFDataRef doesn't have a clean mapping to this. We could engineer one using CFMutableData but I don't think we care enough. It's much easier to just go through normal CF-lifetime patterns and release the data all at once.

---
In reply to: [90743300](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90743300) [](ancestors = 90743300)

Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.mm
aballway: This is owned by the FontCollectionLoader which creates this FontFileEnumerator.  FileEnumerator will never exist without the CollectionLoader.

Frameworks/CoreGraphics/DWriteFontCollectionHelper.mm
```diff
@@ -0,0 +1,207 @@
...
+            CFArrayAppendValue(ret, name);
+        }
+    }
+
+    return ret;
+}
+
+std::shared_ptr<_DWriteFontProperties> DWriteFontCollectionHelper::GetFontPropertiesFromUppercaseFontName(
+    const woc::unique_cf<CFStringRef>& upperFontName) {
+    if (!m_propertiesMap) {
```
aballway: should this be thread-safe?

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -0,0 +1,368 @@
...
+ * derived from that traits dict.
+ *
+ * Note that the name fields in the _DWriteFontProperties are left as blank
+ */
+static _DWriteFontProperties __DWriteFontPropertiesFromTraits(CFDictionaryRef traits) {
+    if (!traits) {
+        return {};
+    }
+
+    _DWriteFontProperties ret = {};
```
aballway: super nit: move this above and return if !traits

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -237,173 +156,92 @@ CFStringRef _CFStringFromLocalizedString(IDWriteLocalizedStrings* localizedStrin
...
-    RETURN_IF_FAILED(systemFontCollection->FindFamilyName(reinterpret_cast<const wchar_t*>(unicharFamilyName.data()),
-                                                          &fontFamilyIndex,
-                                                          &fontFamilyExists));
-
-    if (fontFamilyExists) {
-        return systemFontCollection->GetFontFamily(fontFamilyIndex, outFontFamily);
+    HRESULT result =
+        __GetSystemFontCollectionHelper()->CreateFontFamilyWithName(reinterpret_cast<const wchar_t*>(familyNameUnichars.data()),
+                                                                    outFontFamily);
+    if (result != S_FALSE) {
```
aballway: nit: == S_OK?

Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.mm
bbowman: The fact that this class, DWriteFontFileEnumerator, is a COM class makes this a dubious assumption. COM describes a shared ownership of objects. Presumably this guy is being handed out somewhere and the caller is free to hang on to it for as long as he/she wants. This could be longer than DWriteFontBinaryDataCollectionLoader is held onto. And in fact there is nothing else linking the Enumerator to the Loader which raises serious concerns about the assumptions you made in that comment. Can you explain more about how you are ensuring the validity of this pointer?

---
In reply to: [90908183](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90908183) [](ancestors = 90908183)

Frameworks/CoreGraphics/DWriteFontBinaryDataLoader.mm
bbowman: would CFStream afford you anything nicer / different / more efficient?

---
In reply to: [90747118](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90747118) [](ancestors = 90747118,90743300)

Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.mm
aballway: Our FontCollectionLoader is a static singleton, so it will always exist

Frameworks/CoreGraphics/DWriteFontCollectionHelper.h
bbowman: It fine to use SAL I guess (we don't get the benefit of actually doing anything useful with it except for reading clarity) but it seems to be inconsistent about when you are using it. Take the below _GetFontLIstForFamilyName method that doesn't annotate its in parameter. If we are going to do it, lets make it consistent so that COM style HRESULT returning functions all use SAL or whatever ...

---
In reply to: [90746813](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90746813) [](ancestors = 90746813,90743039,90738161)

Frameworks/CoreText/DWriteWrapper_CoreText.mm
bbowman: you used the word export there and I want to make sure we actually know what the relationships are there. Did you mean exported, as in a dll export, or extern, as in visible outside the current translation unit (.mm or .cpp)

---
In reply to: [90746929](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90746929) [](ancestors = 90746929,90742638)

Frameworks/CoreGraphics/DWriteFontCollectionHelper.h
ms-jihua: ? That function _does_ annotate in its parameter..., or do you just mean that it doesn't annotate the CFStringRef?

---
In reply to: [90917892](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90917892) [](ancestors = 90917892,90746813,90743039,90738161)

bbowman: well seeing as I mentioned an *in* parameter, which one do you think I meant? If you don't think a CF type should be annotated, it begs the question why and what value you are seeing from this process ... 

---
In reply to: [90928031](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90928031) [](ancestors = 90928031,90917892,90746813,90743039,90738161)

Frameworks/CoreGraphics/DWriteFontCollectionHelper.mm
bbowman: does it? Is that a safe assumption from the caller's point of view? 

---
In reply to: [90747010](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90747010) [](ancestors = 90747010,90742971)

Frameworks/CoreGraphics/DWriteFontCollectionHelper.h
ms-jihua: ah, I missed a word. I think I'll just get rid of the SAL then. it was a nice idea but I think the maintenance cost on it going forward won't be worth it for what little we gain in terms of readability.

---
In reply to: [90931524](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90931524) [](ancestors = 90931524,90928031,90917892,90746813,90743039,90738161)

Frameworks/CoreGraphics/DWriteWrapper.mm
ms-jihua: no, this is also intended to be an early return if the call failed above.

---
In reply to: [90911452](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90911452) [](ancestors = 90911452)

Frameworks/CoreGraphics/DWriteFontCollectionHelper.mm
ms-jihua: was thinking about this over the weekend and probably yes

---
In reply to: [90909514](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90909514) [](ancestors = 90909514)

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -818,11 +384,11 @@ static HRESULT __DWriteUpdateUserCreatedFontCollection(CFArrayRef datas, CFArray
     RETURN_IF_FAILED(DWriteCreateFactory(DWRITE_FACTORY_TYPE_SHARED, __uuidof(IDWriteFactory), &dwriteFactory));
 
     // Create singleton Font Collection loader, which will be shared for registering/unregistering fonts
-    static ComPtr<DWriteFontCollectionLoader> loader;
-    static HRESULT createdLoader = [](DWriteFontCollectionLoader** collectionLoader) {
+    static ComPtr<DWriteFontBinaryDataCollectionLoader> loader;
```
bbowman: >loader [](start = 56, length = 6)

s_loader?

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -832,35 +398,29 @@ static HRESULT __DWriteUpdateUserCreatedFontCollection(CFArrayRef datas, CFArray
...
-    // So every time we modify the font collection increment the key
-    static size_t collectionKey = 0;
-    std::lock_guard<std::mutex> guard(_userCreatedFontCollectionMutex);
-    RETURN_IF_FAILED(
-        dwriteFactory->CreateCustomFontCollection(loader.Get(), &(++collectionKey), sizeof(collectionKey), &_userCreatedFontCollection));
-
-    // Update s_userFontPropertiesMap with new values
-    s_userFontPropertiesMap = __CreatePropertiesMapForFontCollection(_userCreatedFontCollection.Get());
+    // Update the user font collection
+    std::lock_guard<std::recursive_mutex> guard(s_userFontCollectionLock);
```
bbowman: >s_userFontCollectionLock [](start = 48, length = 24)

what is this protecting? Seems like the encapsulation is potentially wrong if the FontCollectionHelper is the one actually manipulating the data that needs protecting. Why not lock in that UpdateCollection function?

Frameworks/CoreGraphics/DWriteFontCollectionHelper.mm
ms-jihua: [https://msdn.microsoft.com/en-us/library/windows/desktop/dd371147(v=vs.85).aspx](https://msdn.microsoft.com/en-us/library/windows/desktop/dd371147(v=vs.85).aspx)
exists [out]
Type: BOOL*
When this method returns, TRUE if the font contains the specified string ID; otherwise, FALSE.

I think it's probably safe to interpret that as saying 'this will always be populated'?

---
In reply to: [90931716](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90931716) [](ancestors = 90931716,90747010,90742971)

Frameworks/CoreGraphics/DWriteFontCollectionHelper.mm
ms-jihua: same as the other uninitialized comment ([https://msdn.microsoft.com/en-us/library/windows/desktop/dd368217(v=vs.85).aspx](https://msdn.microsoft.com/en-us/library/windows/desktop/dd368217(v=vs.85).aspx))

---
In reply to: [90742919](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90742919) [](ancestors = 90742919)

Frameworks/CoreText/DWriteWrapper_CoreText.mm
ms-jihua: both, I think? we have unit tests that test _ functions

---
In reply to: [90918451](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90918451) [](ancestors = 90918451,90746929,90742638)

Frameworks/CoreGraphics/DWriteFontCollectionHelper.mm
bbowman: probably. I was trying to get you to think about a little more defensive programming but whatever.

---
In reply to: [90936195](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90936195) [](ancestors = 90936195,90931716,90747010,90742971)

Frameworks/CoreGraphics/DWriteFontCollectionHelper.mm
```diff
@@ -0,0 +1,207 @@
...
+    // If no/empty font collection, fail to return a font list
+    // Use of E_FAIL rather than a more nuanced HRESULT is permissable due to this being a private function only used in
+    // _InitializePropertiesMap()
+    RETURN_HR_IF_NULL(E_FAIL, fontCollection);
+
+    ComPtr<IDWriteFactory> dwriteFactory;
+    RETURN_IF_FAILED(DWriteCreateFactory(DWRITE_FACTORY_TYPE_SHARED, __uuidof(IDWriteFactory), &dwriteFactory));
+
+    size_t index = 0;
+    BOOL exists = FALSE;
```
bbowman: LOL consistency.

Frameworks/CoreGraphics/DWriteFontBinaryDataLoader.mm
bbowman: it offers little by little reading which allows you to build releasable chunks no? This would improve memory consumption right?

---
In reply to: [90935815](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90935815) [](ancestors = 90935815,90916971,90747118,90743300)

Frameworks/CoreGraphics/DWriteWrapper.mm
ms-jihua: read-write access to the whole font collection helper. if we locked in UpdateCollection(), we could still call ->CopyFontFamilyNames() from it, for instance. (we _could_ alleviate that by overriding CopyFontFamilyNames() and the other read functions to also lock, but I don't find that to be preferable to the current pattern of a lock around the whole data structure)

---
In reply to: [90936157](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90936157) [](ancestors = 90936157)

Frameworks/CoreGraphics/DWriteFontCollectionHelper.mm
ms-jihua: >o>

---
In reply to: [90937198](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90937198) [](ancestors = 90937198)

Frameworks/CoreText/DWriteWrapper_CoreText.mm
bbowman: hmmm. Seems like we should be more confident with this. How do the unit tests build? Are these items in a .def file? Would a private dll export have any other potential pitfalls with regards to encapsulation and versioning? 

---
In reply to: [90936774](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90936774) [](ancestors = 90936774,90918451,90746929,90742638)

Frameworks/CoreGraphics/DWriteWrapper.mm
ms-jihua: do we do this naming convention for function-local statics?

---
In reply to: [90935689](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90935689) [](ancestors = 90935689)

Frameworks/CoreGraphics/DWriteWrapper.mm
bbowman: The key thing you said is "lock *around* the whole data structure"

This points to a problem where you don't have the ownership quite correct. The data structure should be able to manage itself (hooray object oriented programming) and therefore the responsibility of locking should be encapsulated to the owning class / object. The pattern as is presents maintainability issues as every new callsite needs to remember to lock otherwise the whole lock is useless. Presumably the problem space for the class itself is much smaller and isolated which makes this a simpler maintainability problem.

---
In reply to: [90937664](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90937664) [](ancestors = 90937664,90936157)

Frameworks/include/StringHelpers.h
ms-jihua: it's actually needed. do you want to grab the build error for you?

---
In reply to: [90742443](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90742443) [](ancestors = 90742443)

Frameworks/CoreGraphics/DWriteWrapper.mm
ms-jihua: That makes sense. I'll move the locks to inside the class.

---
In reply to: [90938541](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90938541) [](ancestors = 90938541,90937664,90936157)

Frameworks/CoreText/DWriteWrapper_CoreText.mm
ms-jihua: _How do the unit tests build?_
not sure what this question means, or what answer you're fishing for. probably something like "CoreGraphics.UnitTests links agains CoreGraphics dll"?

_Are these items in a .def file?_
yes

_Would a private dll export have any other potential pitfalls with regards to encapsulation and versioning?_
i mean, we deliberately broke encapsulation here. versioning risks... _ is intended to communicate the same level of privacy as the stuff in CFFoundationInternal.h, so I figured we'd just continue whatever risk we accepted already, but yeah because of this we'd have to make sure unit tests, coregraphics, coretext, and foundation, all get a version update if the contract of these functions changes.

---
In reply to: [90937900](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90937900) [](ancestors = 90937900,90936774,90918451,90746929,90742638)

Frameworks/CoreGraphics/DWriteWrapper.mm
bbowman: hence the question mark, lol. I personally think its more clear but don't care too much

---
In reply to: [90938094](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90938094) [](ancestors = 90938094,90935689)

Frameworks/include/StringHelpers.h
bbowman: I want you to find the error and think about it more. I have a sneaking suspicion that the error is ore related to our #defs etc for clang than for objective c as a language. I bet if you built for c++ in clang it would work fine. This mostly points to a failing in our sbclang.props / .targets in that its not symmetric building clang and cl but i'd like us to now overindex on this difference as it makes fixing it even harder as each of these things needs to be re-evaluated for what they really are trying to do.  

perhaps an ifdef for clang? or something else related to having CoreFoundation includes available?

---
In reply to: [90938607](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90938607) [](ancestors = 90938607,90742443)

Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.mm
ms-jihua: I'll look into making this a shared_ptr or something similar, I think...

---
In reply to: [90943527](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90943527) [](ancestors = 90943527,90917725)

Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.mm
```diff
@@ -0,0 +1,170 @@
...
+    int m_location = -1;
+
+    woc::unique_cf<CFArrayRef> m_fontDatas;
+    std::vector<ComPtr<IDWriteFontFile>>* m_previouslyCreatedFiles;
+};
+
+// DWriteFontBinaryDataCollectionLoader member functions
+
+HRESULT DWriteFontBinaryDataCollectionLoader::DWriteFontBinaryDataCollectionLoader::RuntimeClassInitialize() {
+    m_fontDatas.reset(CFArrayCreateMutable(kCFAllocatorDefault, 0, &kCFTypeArrayCallBacks));
```
msft-Jeyaram: >kCFAllocatorDefault [](start = 43, length = 19)

consistent usage.

Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.mm
aballway: Documentation for the interface recommends they be singletons, and it seems silly to move it to a shared_ptr when it shouldn't be owning it at all.  Maybe a comment would be helpful but I think wrapping it in a shared_ptr would suggest the opposite of what we want.

---
In reply to: [90943752](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90943752) [](ancestors = 90943752,90943527,90917725)

Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.h
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Custom implementation of IDWriteFontCollectionLoader, which creates an IDWriteFontFileEnumerator to create font files
+ * Which are used to create a font collection.  This is expected to be a singleton.
+ */
+class DWriteFontBinaryDataCollectionLoader
+    : public Microsoft::WRL::RuntimeClass<Microsoft::WRL::RuntimeClassFlags<Microsoft::WRL::WinRtClassicComMix>,
+                                          IDWriteFontCollectionLoader> {
+protected:
+    InspectableClass(L"Windows.Bridge.DirectWrite", TrustLevel::BaseTrust);
+
```
bbowman: very true (its not ever going to be used since we aren't registering our runtime classes to be activatable in a RoActivateInstance type scenario) but yes would be great to do.

---
In reply to: [90944936](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90944936) [](ancestors = 90944936)

Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.mm
bbowman: "shouldn't be owning at all" ... the class blindly uses the pointer which implies something about the lifetime of that memory. This is pretty much the definition of shared ownership. It is saying, "I didn't create the thing, but I have a reasonable expectation that whenever I want, I can use it and therefore store it among my *member* variables, the set of objects I own and manage". To correctly make that work, you need something that enforces that. Right now its a tenuous contract with the singleton parent object that requires a future code maintainer to read  and understand all of the dwrite loader classes and uses as there is nothing else linking the two. The "child" class is not a child in any traditional sense. Its lifetime is not tied to the parent, its definition is not private to the parent, etc. 

This makes the contract hard to continue in the future as someone may invalidate the baseline assumption and then be real confused. One of the *amazing* things about shared_ptr and friends is that the type itself implies the ownership so its much harder to do the wrong thing and at all times the use is clear and communicated. 

My 2 cents is that code should be correct first, maintainable second, and fast/clever/whatever third. If you think that the shared_ptr is overly cumbersome, please take performance measurements to show its impact and we can evaluate from there. Until then, lets make it as easy as possible for the next person.

---
In reply to: [90944931](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90944931) [](ancestors = 90944931,90943752,90943527,90917725)

Frameworks/CoreGraphics/DWriteFontBinaryDataLoader.mm
ms-jihua: @bbowman I'm not sure what you mean by 'build releasable chunks'. Are you suggesting that we write an extension/subclass to CFReadStream that releases bytes as it goes? That seems like a bit heavy in terms of engineering cost for this.

@DHowett-MSFT I'm not certain if CGDataProvider would fare better either, its release callback appears to want to free all the data at once as well [https://developer.apple.com/reference/coregraphics/cgdataproviderreleasedatacallback?language=objc](https://developer.apple.com/reference/coregraphics/cgdataproviderreleasedatacallback?language=objc)

---
In reply to: [90945909](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90945909) [](ancestors = 90945909,90937378,90935815,90916971,90747118,90743300)

bbowman: Talked offline but the basic idea was that CFReadStream can read into arbitrary buffers which we could manage and release when they tell us to. The real kicker is that we are stealing the whole data from CGDataProvider at the beginning nullifying this benefit. A later change that actually implements CGDataProvider correctly should do better.

---
In reply to: [90956518](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90956518) [](ancestors = 90956518,90945909,90937378,90935815,90916971,90747118,90743300)

Frameworks/CoreGraphics/DWriteFontBinaryDataLoader.mm
ms-jihua: mm, I don't think you need to go that far. this is still one-lineable, I just picked this way of stating it since CFRetain returns CFTypeRef, and I find casting back rather ugly. If we wanted, we could add like a 'retain_and_reset' to unique_cf and it'd make this neater.

---
In reply to: [90743195](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90743195) [](ancestors = 90743195)

Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.mm
ms-jihua: settled on a pattern where the enumerator delegates to the loader, which owns the data.

---
In reply to: [90948737](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90948737) [](ancestors = 90948737,90944931,90943752,90943527,90917725)

Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.mm
ms-jihua: Mm, pushing back on this. I'm not convinced that a single iteration failing here should fail the whole function, and I don't think that RETURN_IF_FAILED should be tripped in such a case. S_FALSE has a connotation of 'partial success', and we have the errors out-array if we need more specific error information anyway.

---
In reply to: [90743736](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90743736) [](ancestors = 90743736)

Frameworks/include/CoreGraphics/DWriteWrapper.h
ms-jihua: I guess I don't have a better answer for you than 'convenience'. The alternative would be to make the caller allocate a char buffer for us to dump into. Same below, it's more convenient to just pass a shared_ptr around than to define (iirc) a copy constructor for that struct. What are the downsides of cross module exporting c++?

---
In reply to: [90742313](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90742313) [](ancestors = 90742313)

Frameworks/include/CoreGraphics/DWriteWrapper.h
ms-jihua: continuing discussion above

---
In reply to: [90742335](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90742335) [](ancestors = 90742335)

ms-jihua:
Ping!

Frameworks/CoreGraphics/DWriteFontCollectionHelper.mm
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    return ret;
+}
+
+std::shared_ptr<_DWriteFontProperties> DWriteFontCollectionHelper::GetFontPropertiesFromUppercaseFontName(
+    const woc::unique_cf<CFStringRef>& upperFontName) {
+    std::lock_guard<std::recursive_mutex> lock(m_lock);
+
+    if (!m_propertiesMap) {
```
aballway: nit: maybe a dispatch once here rather than flag and a mutex

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -14,184 +14,101 @@
...
+    ComPtr<IDWriteFontCollection> _GetFontCollection() {
+        // Seems okay to fail-fast in this case - hard to continue without any system font collection
+        ComPtr<IDWriteFactory> dwriteFactory;
+        THROW_IF_FAILED(DWriteCreateFactory(DWRITE_FACTORY_TYPE_SHARED, __uuidof(IDWriteFactory), &dwriteFactory));
 
-    // If the default locale is returned, find that locale name, otherwise use "en-us".
-    return wstring(defaultLocaleSuccess ? localeName : c_defaultUserLanguage);
-}
+        ComPtr<IDWriteFontCollection> fontCollection;
+        THROW_IF_FAILED(dwriteFactory->GetSystemFontCollection(&fontCollection));
```
aballway: nit: Would prefer to keep this static to only pull once rather than getting every time this is called

Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.mm
```diff
@@ -0,0 +1,186 @@
...
+
+    CFIndex count = CFArrayGetCount(fontDatas);
+    for (CFIndex i = 0; i < count; ++i) {
+        CFDataRef data = static_cast<CFDataRef>(CFArrayGetValueAtIndex(fontDatas, i));
+        if (data) {
+            if (CFSetContainsValue(m_fontDatasSet.get(), data)) {
+                CFSetRemoveValue(m_fontDatasSet.get(), data);
+
+                for (auto it = m_fontDatas.begin(); it != m_fontDatas.end(); ++it) {
+                    if (CFEqual(it->first.get(), data)) {
```
DHowett-MSFT: `O(N*M)`! While this is fairly low-bounded for the common case, I'm worried about not using something more hashtabley here. CFData supports hashing, and we could make do with an adapter like CFStringRef's?

Frameworks/CoreText/CTTypesetter.mm
```diff
@@ -84,6 +84,15 @@ CTLineRef CTTypesetterCreateLineWithOffset(CTTypesetterRef ts, CFRange range, do
     RETURN_NULL_IF(!frame);
     THROW_NS_IF_FALSE(E_UNEXPECTED, [frame->_lines count] == 1);
 
+    if ([frame->_lines count] != 1) {
```
aballway: remove the throw above

Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.h
bbowman: be wary of classic com as the runtime flags. We should make ours mix because it implies the implementation is Agile (can be called on any thread). Unless we want to cause sta marshalling of calls, we probably want to stay with IInspectable and be agile.

---
In reply to: [91176606](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r91176606) [](ancestors = 91176606,90946927,90944936)

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -0,0 +1,354 @@
...
+ * then parses them into a dictionary suitable for kCTFontTraitsAttribute
+ */
+static CFDictionaryRef _DWriteFontCreateTraitsDict(const ComPtr<IDWriteFontFace>& fontFace) {
+    // Get pointers for the additional FontFace interfaces
+    ComPtr<IDWriteFontFace1> fontFace1;
+    RETURN_NULL_IF_FAILED(fontFace.As(&fontFace1));
+    ComPtr<IDWriteFontFace2> fontFace2;
+    RETURN_NULL_IF_FAILED(fontFace.As(&fontFace2));
+    ComPtr<IDWriteFontFace3> fontFace3;
+    RETURN_NULL_IF_FAILED(fontFace.As(&fontFace3));
```
aballway: IDWriteFontFace3 inherits from IDWriteFontFace2, so just use that one

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -0,0 +1,354 @@
...
+
+    if (fontFace1->HasVerticalGlyphVariants()) {
+        symbolicTraits |= kCTFontVerticalTrait;
+    }
+
+    if (fontFace2->IsColorFont()) {
+        symbolicTraits |= kCTFontColorGlyphsTrait;
+    }
+
+    // TODO: The symbolic traits below are poorly documented/have no clear DWrite mapping
```
aballway: No commented out code

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -0,0 +1,354 @@
...
+ * Gets a name/informational string from a DWrite font face corresponding to a CTFont constant
+ */
+CFStringRef _DWriteFontCopyName(const ComPtr<IDWriteFontFace>& fontFace, CFStringRef nameKey) {
+    if (nameKey == nullptr || fontFace == nullptr) {
+        return nullptr;
+    }
+
+    DWRITE_INFORMATIONAL_STRING_ID informationalStringId;
+
+    if (CFEqual(nameKey, kCTFontCopyrightNameKey)) {
```
aballway: is this making enough of an impact to convert to dictionary-lookup?

Frameworks/CoreGraphics/DWriteFontBinaryDataLoader.mm
```diff
@@ -0,0 +1,116 @@
...
+        FILETIME fileTime;
+        GetSystemTimeAsFileTime(&fileTime);
+
+        // Concat filetime into a single uint64_t
+        m_lastWriteTime = 0;
+        m_lastWriteTime |= static_cast<uint64_t>(fileTime.dwLowDateTime);
+        m_lastWriteTime |= static_cast<uint64_t>(fileTime.dwHighDateTime) << 32;
+    }
+
+    HRESULT RuntimeClassInitialize() {
```
DHowett-MSFT: >RuntimeClassInitialize [](start = 12, length = 22)

If we use `Make` I don't think we even need `RuntimeClassInitialize` as a stub!

Frameworks/CoreText/DWriteWrapper_CoreText.mm
```diff
@@ -14,30 +14,24 @@
...
 #import <iterator>
-#import <numeric>
 
 using namespace std;
 using namespace Microsoft::WRL;
 
 static const wchar_t* TAG = L"_DWriteWrapper_CoreText";
 
-static DWRITE_TEXT_ALIGNMENT __CoreTextAlignmentToDwrite(CTTextAlignment alignment) {
+static DWRITE_TEXT_ALIGNMENT __CTAlignmentToDWrite(CTTextAlignment alignment) {
```
aballway: nit: should make these inline too :)

Frameworks/CoreText/DWriteWrapper_CoreText.mm
aballway: still not CopyTo?

Frameworks/CoreText/DWriteWrapper_CoreText.mm
aballway: same as above

Frameworks/CoreText/DWriteWrapper_CoreText.mm
DHowett-MSFT: @aballway `CopyTo` adds an unnecessary +1 on the pointed-to COM object that's -1'd when the method returns immediately after. `*x = y.Detach()` is the idiomatic way to commute a +1 reference to a caller.

Frameworks/CoreText/CTTypesetter.mm
ms-jihua: oops, I thought I did that...

---
In reply to: [91176623](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r91176623) [](ancestors = 91176623)

Frameworks/CoreGraphics/DWriteWrapper.mm
ms-jihua: Mm, but then we have to keep the whole font collection in memory, which I think I hate more than getting it every time. I'll possibly revisit this in #1374 but I don't think it'll be big enough of an issue to come up.

---
In reply to: [91175727](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r91175727) [](ancestors = 91175727)

Frameworks/CoreGraphics/DWriteFontBinaryDataCollectionLoader.mm
ms-jihua: Oh, thanks for noticing this (little bit glad that the changes made the O(N*M) more apparent). Thinking about how to change m_fontDatasSet to be something better for this purpose...

---
In reply to: [91176040](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r91176040) [](ancestors = 91176040)

ms-jihua: Probably rather than the current setup (CFSet and vector<pair>), we should just use an std::unordered_map that uses CFData's hash/equal

---
In reply to: [91189304](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r91189304) [](ancestors = 91189304,91176040)

Frameworks/include/CoreGraphics/DWriteWrapper.h
ms-jihua: Discussed offline. It's a little sketch still but we're letting it pass in this case.

---
In reply to: [90993784](https://github.com/Microsoft/WinObjC/pull/1504#discussion_r90993784) [](ancestors = 90993784,90742313)

ms-jihua:
CI builds both completed for 4th commit, skipping for 5th commit as it is very minor.

PR #1506, opened 2016-12-05
Fix for issue 1499: UIView: [UIView invalidateIntrinsicContentSize] does far more work than necessary

jaredhms:
Moving to a lazy model for constraint recalculation allows IslandwoodNews to launch more reliably on phone, but additional perf investigation is ongoing.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1506)

### DIFF
--- a/Frameworks/AutoLayout/AutoLayout.mm
+++ b/Frameworks/AutoLayout/AutoLayout.mm
@@ -102,6 +102,7 @@ public:
     ClLinearInequality _contentCompressionResistanceConstraint[NumDirections];
 
     NSMutableArray* _associatedConstraints;
+    bool _isContentSizeValid = false;
 };
 
 // Since categories can't have ivars, we bundle up a couple C++ classes in NSObjects and associate them to the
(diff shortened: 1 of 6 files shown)


Frameworks/AutoLayout/AutoLayout.mm
```diff
@@ -517,6 +496,58 @@ - (void)autoLayoutInvalidateContentSize {
...
 }
 
+- (void)autoLayoutLayoutSubviews {
+    UIView* topView = [self autolayoutRoot];
+
+    // Make sure our constraints are up to date
+    [self _validateContentSize];
+    [topView _validateContentSize];
+
+    for (int i = 0; i < [self.subviews count]; i++) {
```
aballway: nit: for in loop

Frameworks/UIKit/StarboardXaml/ApplicationCompositor.cpp
```diff
@@ -59,19 +59,24 @@ extern "C" void RunApplicationMain(Platform::String^ principalClassName,
                                    float windowHeight,
                                    ActivationType activationType,
                                    Platform::Object^ activationArg) {
+    // Return from activation immediatly; push all subsequent app launch work to
```
aballway: This seems unrelated to AutoLayout issues, and should be tested really thoroughly to make sure we don't accidentally regress anywhere with app launching

Frameworks/AutoLayout/AutoLayout.mm
jaredhms: going to leave as-is for now; not new code.

Frameworks/UIKit/StarboardXaml/ApplicationCompositor.cpp
jaredhms: Agreed; it's an additional 'fix' for the app launch timeouts.  Raj had intended to do this with a previous change, but it didn't push all work off of the initial activation.  I agree that we need to test adequately.

yiyang-msft:
:shipit:

jaredhms:
I have a much larger change for this; we'll roll it out in a separate PR.

PR #1511, opened 2016-12-06
presentViewController/dismissViewController fixups

ehren:
-fix too early popover deallocation when programmatically dismissing via dismissViewController
-stop requiring tablet mode to display popovers, allowing simplifications to popover presentation
-fix/implement proper handling of multiple children dismissal via dismissViewController (example added to WOCCatalog)
-kill _modalViewController
-more sane handling of _presentedViewController memory management (existing approach with __unsafe_unretained ivar and retain/autorelease gymnastics on dismiss not proper)
-general cleanups

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1511)

note: the crash on dismiss of popover via dismissViewController existed prior to #1464


ehren:
These changes (as several commits) have been incorporated into pull #1601

PR #1514, opened 2016-12-06
Update objc2winmd binary: 

mukhole:
Changed the annotation used for heterogeneous containers.
Heterogeneous containers are now annotated as:
ContainerType<type1, type2, type3[,...]>
This change maintains backward compatibility, so the previous annotation:
ContainerType<type1>:ContainerType<type2>:ContainerType<type3> 
is still valid.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1514)


pradipd:
:shipit:

PR #1519, opened 2016-12-08
Control should not update its state value if its value didn't change

yiyang-msft:
If an app setting this value during layout pass, it can trigger layout dead loop.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1519)


jaredhms:
:shipit:

PR #1523, opened 2016-12-12
Move functional tests to use TAEF directly. This removes the need for…

bbowman:
… trampoline functions and paves the way for other ObjC tests to use the RunAs:UAP functionality that the test framework provies.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1523)

### DIFF
--- a/build/Tests/FunctionalTests/FunctionalTests.vcxproj
+++ b/build/Tests/FunctionalTests/FunctionalTests.vcxproj
@@ -281,6 +281,7 @@
     <ClangCompile Include="$(MSBuildThisFileDirectory)..\..\..\tests\functionaltests\Tests\NSURLStorageFileTests.mm">
       <ObjectiveCARC>true</ObjectiveCARC>
     </ClangCompile>
+    <ClangCompile Include="$(WINOBJC_SDK_ROOT)\tests\unittests\CoreText\CTRunTests.mm" />
   </ItemGroup>
   <ItemGroup>
     <Xml Include="$(MSBuildThisFileDirectory)..\..\..\tests\functionaltests\Tests\Default.AppxManifest.xml" />
--- a/deps/prebuilt/include/TAEF/WexTestClass.h
+++ b/deps/prebuilt/include/TAEF/WexTestClass.h
@@ -3,6 +3,7 @@
 /// <summary>Test Authoring and Execution Framework native macro definitions</summary>
 // Copyright (c) Microsoft Corporation.  All Rights Reserved.
 //----------------------------------------------------------------------------------------------------------------------
+// clang-format off
 #pragma once
 
 // We currently require the C++ compiler from VS 2012 or later.
@@ -13,9 +14,11 @@
 #pragma warning(push)
 #pragma warning(disable:4481)
 
+#if !defined(__clang__)
 #pragma comment(lib, "TE.Common.lib")
 #pragma comment(lib, "Wex.Common.lib")
 #pragma comment(lib, "Wex.Logger.lib")
+#endif
 
 // Allow anyone who has defined an Assert macro to compile with this header file included
 #pragma push_macro("Assert")
@@ -35,6 +38,11 @@
 #include "WexException.h"
 
 #include <new>
+#endif
+
+#if defined(__clang__)
+#include <array>
+#include <utility>
 #endif
 
 #pragma push_macro("_Outptr_")
@@ -122,6 +130,29 @@ namespace WEX
 {
     namespace Private
     {
+#if defined(__clang__)
+        // Helper function to index into arrays to build up big std::array with concatenated characters of two. 
+        // Uses template function arg deduction to pull out NumChars from constant c strings and parameter pack expansion with
+        // std::index_sequence to rewrite into a new constant "string like object" that resembles a C++ qualified member name.
+        //
+        // TL;DR:
+        //    "ClassName", "MethodName" -> "ClassName::MethodName"
+        //    all at compile time because of constexpr and template magic.
+        template<unsigned NumChar1, unsigned NumChar2, std::size_t... Indices1, std::size_t... Indices2>
+        constexpr std::array<wchar_t const, NumChar1 + NumChar2 + 1> BuildQualifiedName(
+            wchar_t const (&str1)[NumChar1], wchar_t const (&str2)[NumChar2], std::index_sequence<Indices1...>, std::index_sequence<Indices2...>) {
+            static_assert(sizeof...(Indices1) < NumChar1, "Range of characters to use from first string must not include null terminator.");
+            return {{ str1[Indices1]..., L':', L':', str2[Indices2]...}};
+        }
+
+        // Helper to make calling the other one a little nicer using template argument type deduction.
+        template<unsigned NumChar1, unsigned NumChar2>
+        constexpr std::array<wchar_t const, NumChar1 + NumChar2 + 1> BuildQualifiedName(
+            wchar_t const (&str1)[NumChar1], wchar_t const (&str2)[NumChar2]) {
+            return BuildQualifiedName(str1, str2, std::make_index_sequence<NumChar1-1>(), std::make_index_sequence<NumChar2>());
+        }
+#endif
+
         /// \internal
         /// <summary>
         /// A SEH filter.
@@ -587,13 +618,40 @@ namespace WEX
 #pragma section("testdata$g_TCM", read)
 #pragma section("testdata$h_TMM", read)
 
+#define TAEF_ATTRIBUTE_SECTION(sectionName)                 __declspec(allocate(sectionName))
+
 // The IntelliSense compiler reports errors when trying to parse 'TAEF_PIN_FUNCTION_SYMBOL', so it should be skipped.
 #if defined(__INTELLISENSE__)
 # define TAEF_PIN_FUNCTION_SYMBOL
 #else
+
 /// \internal
 /// <summary> Instruct the compiler not to remove the function from the obj file during the optimization phase </summary>
+#if defined(__clang__)
+# define TAEF_PIN_FUNCTION_SYMBOL                 __attribute__((used)) 
+#else
 # define TAEF_PIN_FUNCTION_SYMBOL                 __pragma (comment(linker, "/include:" __FUNCDNAME__))
+#endif
+    
+#endif
+
+#if defined (__clang__)
+
+#define TAEF_CLASS_NAME_SYMBOL c_taefClassNameArray
+#define TAEF_CLASS_NAME &TAEF_CLASS_NAME_SYMBOL[0]
+#define TAEF_DEFINE_CLASS_NAME(className) \
+    static constexpr const wchar_t TAEF_CLASS_NAME_SYMBOL[] = L# className
+
+#define TAEF_QUALIFIED_METHOD_NAME_SYMBOL s_qualifiedMethodName
+#define TAEF_QUALIFIED_METHOD_NAME &TAEF_QUALIFIED_METHOD_NAME_SYMBOL[0]
+#define TAEF_DEFINE_QUALIFIED_METHOD_NAME(methodName) \
+    static constexpr auto TAEF_QUALIFIED_METHOD_NAME_SYMBOL = WEX::Private::BuildQualifiedName(TAEF_CLASS_NAME_SYMBOL, methodName)
+
+#else
+
+#define TAEF_DEFINE_QUALIFIED_METHOD_NAME(methodName)
+#define TAEF_DEFINE_CLASS_NAME(className)
+
 #endif
 
 /// \internal
@@ -610,6 +668,16 @@ namespace WEX
 #define TAEF_REGISTER_FIXTURE_METHOD(methodName, methodType)
 #define TAEF_REGISTER_TEST_METHOD(methodName)
 #else
+
+#if defined (__clang__)
+#define TAEF_FIXTURE_METHOD_INFO(methodType, methodName) \
+    {methodType, L#methodName, TAEF_QUALIFIED_METHOD_NAME, &TAEF_FixtureInvoker::TAEF_Invoke}
+#else
+#define TAEF_FIXTURE_METHOD_INFO(methodType, methodName) \
+    {methodType, L#methodName, TAEF__WFUNCTION__, &TAEF_FixtureInvoker::TAEF_Invoke}
+#endif
+
+
 #define TAEF_REGISTER_FIXTURE_METHOD(methodName, methodType) \
     struct TAEF_FixtureInvoker \
     { \
@@ -617,13 +685,22 @@ namespace WEX
         { \
             /* TAEF_TestClassType was typedef'd in the TEST_CLASS macro so we can safely cast from \
                void* to TAEF_TestClassType* and invoke the correct derived type's member functions */ \
-            return WEX::SafeInvoke(WEX::FixtureInvokeFunctor<TAEF_TestClassType>(*reinterpret_cast<TAEF_TestClassType*>(pTestClass), &methodName)); \
+            return WEX::SafeInvoke(WEX::FixtureInvokeFunctor<TAEF_TestClassType>(*reinterpret_cast<TAEF_TestClassType*>(pTestClass), &TAEF_TestClassType::methodName)); \
         } \
     }; \
+    TAEF_DEFINE_QUALIFIED_METHOD_NAME(TAEF__WFUNCTION__); \
     TAEF_PUSH_IGNORE_WARNINGS \
-    TAEF_PIN_FUNCTION_SYMBOL __declspec(allocate("testdata$e_FMI")) \
-    static const WEX::FixtureMethodInfo s_testInfo = {methodType, L#methodName, TAEF__WFUNCTION__, &TAEF_FixtureInvoker::TAEF_Invoke}; \
+    TAEF_PIN_FUNCTION_SYMBOL TAEF_ATTRIBUTE_SECTION("testdata$e_FMI") \
+    static const WEX::FixtureMethodInfo s_testInfo = TAEF_FIXTURE_METHOD_INFO(methodType, methodName); \
     return &s_testInfo; /* Return a pointer to s_testInfo in order to pin the struct into the dll so it does not get stripped out by the linker. */
+
+#if defined (__clang__)
+#define TAEF_TEST_METHOD_INFO(methodName) \
+    {WEX::TAEF_Identifier::TestMethodInfo, L#methodName, TAEF_QUALIFIED_METHOD_NAME, __COUNTER__ - TAEF_TestMethodIndexOffset, &TAEF_TestInvoker::TAEF_Invoke}
+#else
+#define TAEF_TEST_METHOD_INFO(methodName) \
+    {WEX::TAEF_Identifier::TestMethodInfo, L#methodName, TAEF__WFUNCTION__, __COUNTER__ - TAEF_TestMethodIndexOffset, &TAEF_TestInvoker::TAEF_Invoke}
+#endif
 
 #define TAEF_REGISTER_TEST_METHOD(methodName) \
     struct TAEF_TestInvoker \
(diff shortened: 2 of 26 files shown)


yiyang-msft:
:shipit:

pradipd:
:shipit:

tests/functionaltests/Tests/AssetsLibraryTests/AssetsLibraryTests.mm
```diff
@@ -23,85 +23,208 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
```
pradipd: we can remove this.

tests/functionaltests/functionaltest-api.h
```diff
@@ -25,7 +52,7 @@
 #ifdef TEST
 #undef TEST
 #endif
-#define TEST(__Name1, __Name2) void __Name1##__Name2()
+#define TEST(__Name1, __Name2) TEST_METHOD(__Name1##__Name2)
```
jaredhms: can we just get rid of this now?

tests/functionaltests/Tests/NSBundleTests.mm
```diff
@@ -19,39 +19,54 @@
...
-    ASSERT_TRUE (signaled || [condition waitUntilDate:[NSDate dateWithTimeIntervalSinceNow:2]]);
-    [condition unlock];
-
-    ASSERT_EQ(YES, success);
-}
\ No newline at end of file
+class NSBundleTests {
+public:
+    BEGIN_TEST_CLASS(NSBundleTests)
+    END_TEST_CLASS()
```
jaredhms: nit: can just be TEST_CLASS(NSBundleTests) - no need for begin/end since we don't add any class-specific test metadata here.

jaredhms: ...in several files...

---
In reply to: [92248298](https://github.com/Microsoft/WinObjC/pull/1523#discussion_r92248298) [](ancestors = 92248298)

tests/functionaltests/Tests/NSBundleTests.mm
```diff
@@ -19,39 +19,54 @@
...
+            failure:^void(NSError* error) {
+                LOG_ERROR([[error description] UTF8String]);
+                [condition lock];
+                signaled = YES;
+                [condition signal];
+                [condition unlock];
+            }];
+
+        [condition lock];
+        ASSERT_TRUE(signaled || [condition waitUntilDate:[NSDate dateWithTimeIntervalSinceNow:2]]);
```
jaredhms: >ASSERT_TRUE [](start = 8, length = 11)

can we switch over to the TAEF macros now?

tests/functionaltests/Tests/UIKitTests/UITextFieldTests.mm
```diff
@@ -30,17 +30,36 @@
...
+        return SUCCEEDED(FrameworkHelper::RunOnUIThread(&UIApplicationDefaultInitialize));
+    }
+
+    TEST_METHOD_CLEANUP(UIKitTestsCleanup) {
+        FunctionalTestCleanupUIApplication();
+        return true;
+    }
+
+    TEST(UITextField, CreateXamlElement) {
+        FrameworkHelper::RunOnUIThread([]() {
```
jaredhms: >rameworkHelper::RunOnUIThread [](start = 9, length = 29)

should we delete this and switch to TAEF's API ([https://osgwiki.com/wiki/TAEF_RunAs_UAP#Native_API](https://osgwiki.com/wiki/TAEF_RunAs_UAP#Native_API))? Maybe just track as a separate issue?

deps/prebuilt/include/TAEF/WexTestClass.h
```diff
@@ -115,13 +119,44 @@ extern "C" __declspec(noreturn) void __fastfail(unsigned int);
...
 /// The WEX namespace contains classes used for marking up C++ Tests.
 /// </summary>
 namespace WEX
 {
     namespace Private
     {
+        #if defined(__clang__)
+        // tempalated struct using variadic template args to represent a sequence of characters
+        template <unsigned... CharacterIndices> 
+        struct index_sequence{};
```
jaredhms: >index_sequence [](start = 15, length = 14)

nit: we should probably stick to TAEF's coding conventions for all of these (all structs would be PascalCase

deps/prebuilt/include/TAEF/WexTestClass.h
```diff
@@ -610,6 +653,23 @@ namespace WEX
...
+
+#if defined (__clang__)
+#define TAEF_REGISTER_FIXTURE_METHOD(methodName, methodType) \
+    struct TAEF_FixtureInvoker \
+    { \
+        static HRESULT __cdecl TAEF_Invoke(void* pTestClass) \
+        { \
+            /* TAEF_TestClassType was typedef'd in the TEST_CLASS macro so we can safely cast from \
+               void* to TAEF_TestClassType* and invoke the correct derived type's member functions */ \
+            return WEX::SafeInvoke(WEX::FixtureInvokeFunctor<TAEF_TestClassType>(*reinterpret_cast<TAEF_TestClassType*>(pTestClass), &TAEF_TestClassType::methodName)); \
```
jaredhms: >&TAEF_TestClassType::methodName [](start = 133, length = 31)

wouldn't this work for TAEF, too?  Can/should we - instead of forking these entire macros - only fork the portions that need to be different?

deps/prebuilt/include/TAEF/WexTestClass.h
```diff
@@ -661,10 +741,10 @@ namespace WEX
 ///    }
 /// \endcode
 #define TEST_METHOD(methodName) \
-    TAEF_TEST_METHOD(methodName) TAEF_TERMINATOR
+    TAEF_TEST_METHOD(methodName)
 
 #define TAEF_TEST_METHOD_METADATA_START \
-    TAEF_PIN_FUNCTION_SYMBOL __declspec(allocate("testdata$h_TMM")) \
+    TAEF_PIN_FUNCTION_SYMBOL TAEF_ATTRIBUTE_SECTION("testdata$h_TMM") \
     static WEX::TestPropertyMetadata s_Metadata[] = { {WEX::TAEF_Identifier::TestMethodMetadata, L"Name", TAEF__WFUNCTION__}, \
```
jaredhms: >L"Name", TAEF__WFUNCTION__}, \ [](start = 97, length = 30)

do we need to also fix this up for method metadata?

deps/prebuilt/include/TAEF/WexTestClass.h
```diff
@@ -115,13 +119,44 @@ extern "C" __declspec(noreturn) void __fastfail(unsigned int);
 #define TAEF_FAST_FAIL System::Environment::FailFast(nullptr)
 #endif // #if !defined(__cplusplus_cli)
 
+#if defined(__clang__)
```
mschofie: Can you move this #include further up? Even though it's clang-specific it should be with the rest (somewhere near the 'except.h' include).

deps/prebuilt/include/TAEF/WexTestClass.h
```diff
@@ -80,6 +83,7 @@
 /// <summary>Wide string version of __FUNCTION__ macro </summary>
 #define TAEF__WFUNCTION__ TAEF_WIDEN(__FUNCTION__)
 
+
```
mschofie: Can you pull this extra blank line, just to minimize churn?

deps/prebuilt/include/TAEF/WexTestClass.h
bbowman: yes agreed. Was just putting it close to the use as I was trying things out.

---
In reply to: [92518736](https://github.com/Microsoft/WinObjC/pull/1523#discussion_r92518736) [](ancestors = 92518736)

deps/prebuilt/include/TAEF/WexTestClass.h
mschofie: Brian's copying the standard names, so I'd be fine with just a comment. std::index_sequence should be in 'utility' - should you use that? TBH, I'm torn as to whether you should #include utility or not. It's always been painful to remove #includes, since downstream consumers get grumpy if they were getting them for free; so absolutely minimizing the #includes is typically a priority. Anyone else have comments?

deps/prebuilt/include/TAEF/WexTestClass.h
```diff
@@ -587,17 +622,25 @@ namespace WEX
...
+#define TAEF_PIN_FUNCTION_SYMBOL                 __pragma (comment(linker, "/include:" __FUNCDNAME__))
+#else
+#define TAEF_PIN_FUNCTION_SYMBOL                 __attribute__((used)) 
+#endif
+    
 #endif
 
 /// \internal
-/// <summary> Instruct oacr to ignore "static initializer causes copy on write pages due to member function pointers" warning
+/// <summary> Instruct oa cr to ignore "static initializer causes copy on write pages due to member function pointers" warning
```
mschofie: Revert the 'oacr' --> 'oa cr' change, please.

deps/prebuilt/include/TAEF/WexTestClass.h
mschofie: +1. It looks like macroizing the 'class name' piece to be MSVC or clang specific /should/ be sufficient.

deps/prebuilt/include/TAEF/WexTestClass.h
```diff
@@ -587,17 +622,25 @@ namespace WEX
 #pragma section("testdata$g_TCM", read)
 #pragma section("testdata$h_TMM", read)
 
+#define TAEF_ATTRIBUTE_SECTION(sectionName)                 __declspec(allocate(sectionName))
+
 // The IntelliSense compiler reports errors when trying to parse 'TAEF_PIN_FUNCTION_SYMBOL', so it should be skipped.
 #if defined(__INTELLISENSE__)
-# define TAEF_PIN_FUNCTION_SYMBOL
+#define TAEF_PIN_FUNCTION_SYMBOL
```
bbowman: So I thought that my code formatter was going wacky here. I'm fine putting it back but I wasn't aware of any "scoping" convention here. # define looks like a syntax error at first glance. Would it be more clear to do 

#if (...)
    #define ...

as a scoping mechanism? 

I'm also fine punting this just to minimize churn though.

---
In reply to: [92519227](https://github.com/Microsoft/WinObjC/pull/1523#discussion_r92519227) [](ancestors = 92519227)

deps/prebuilt/include/TAEF/WexTestClass.h
bbowman: oops

---
In reply to: [92519294](https://github.com/Microsoft/WinObjC/pull/1523#discussion_r92519294) [](ancestors = 92519294)

deps/prebuilt/include/TAEF/WexTestClass.h
```diff
@@ -784,14 +864,54 @@ namespace WEX
 ///          TEST_CLASS(TestFeatureClass)
 ///      }
 /// \endcode
+#if defined (__clang__)
 #define TEST_CLASS(className) \
+protected: \
+    static constexpr auto c_taefClassName = L# className; \
```
mschofie: You shouldn't need both c_taefClassName and c_taefClassNameArray.

c_taefClassNameArray should be the one to use - it's a C++ array, but can be used to initialize a const wchar_t* (which is what it looks like you were using c_taefClassName for).

deps/prebuilt/include/TAEF/WexTestClass.h
bbowman: Yeah I agree. I was overly cautious in the first go to not touch MSVC stuff as much as possible but I think adding new macros for the deltas should make it more clear and still be safe. I think here specifically we could add the class type qualifier and be happy in both case but I'm fine if we want a macro that bifuractes here.

---
In reply to: [92519390](https://github.com/Microsoft/WinObjC/pull/1523#discussion_r92519390) [](ancestors = 92519390)

deps/prebuilt/include/TAEF/WexTestClass.h
bbowman: I agree. I was having a few issues using the Array version to initialize wchar_t* types in the structs but I didn't put much time in. Probably just need a &array[0] or to muck with the consts to get it to be happy.

---
In reply to: [92519559](https://github.com/Microsoft/WinObjC/pull/1523#discussion_r92519559) [](ancestors = 92519559)

deps/prebuilt/include/TAEF/WexTestClass.h
```diff
@@ -115,13 +119,44 @@ extern "C" __declspec(noreturn) void __fastfail(unsigned int);
...
 namespace WEX
 {
     namespace Private
     {
+        #if defined(__clang__)
+        // tempalated struct using variadic template args to represent a sequence of characters
+        template <unsigned... CharacterIndices> 
+        struct index_sequence{};
+
+        // helper to convert N -> N, N-1, N-2, .. , 1 , 0 as template args
```
bbowman: > N -> N, N-1, N-2, .. , 1 , 0 [](start = 28, length = 29)

comment is actually backwards in terms of template arg order (correct in recursion order though). Probably just remove in favor of std::integer_sequence

tests/functionaltests/functionaltest-api.h
bbowman: yeah. Talked offline about the UT plan but lets make functional tests nice. Will leave the verification macros as is for now and file a task to clean those up later.

---
In reply to: [92248037](https://github.com/Microsoft/WinObjC/pull/1523#discussion_r92248037) [](ancestors = 92248037)

tests/functionaltests/Tests/NSBundleTests.mm
bbowman: just moved over from functionaltests.cpp but yes that would be cleaner.

---
In reply to: [92248343](https://github.com/Microsoft/WinObjC/pull/1523#discussion_r92248343) [](ancestors = 92248343,92248298)

deps/prebuilt/include/TAEF/WexTestClass.h
mschofie: Using std::index_sequence with the #include <utility> wrapped in a clang #if def sounds good to me.

build/Tests/FunctionalTests/FunctionalTests.vcxproj
```diff
@@ -281,6 +281,7 @@
     <ClangCompile Include="$(MSBuildThisFileDirectory)..\..\..\tests\functionaltests\Tests\NSURLStorageFileTests.mm">
       <ObjectiveCARC>true</ObjectiveCARC>
     </ClangCompile>
+    <ClangCompile Include="$(WINOBJC_SDK_ROOT)\tests\unittests\CoreText\CTRunTests.mm" />
```
bbowman: ><ClangCompile Include="$(WINOBJC_SDK_ROOT)\tests\unittests\CoreText\CTRunTests.mm" /> [](start = 4, length = 85)

Shows a "fancy" unit test working. Was added here since FunctionalTests rely on a custom logger for TAEF that will need injected into other builds via .props / .targets but then we can move UTs to use TAEF.

tests/frameworks/gtest/include/gtest/gtest.h
```diff
@@ -478,6 +480,23 @@ class GTEST_API_ Test {
   GTEST_DISALLOW_COPY_AND_ASSIGN_(Test);
 };
 
+// Helper TestRunner class used to indirect from a TAEF test class which is named with a Prefix to a single
+// class that TEST_P can friend. TODO: move to a impl file to not violate ODR (doesn't matter since all defs are the same though)
```
bbowman: >impl file [](start = 49, length = 9)

gtest.cc seems like a good candidate.

tests/functionaltests/functionaltest-api.h
```diff
@@ -385,4 +402,88 @@
...
+        TEST_CLASS(className##_##methodName) \
+    \
+        TEST_METHOD(Test); \
+    };  \
+    \
+    void className##_##methodName::Test()
+
+#pragma push_macro("TEST_F")
+#define TEST_F(className, methodName)                  \
+/* This assumes that there already exists a className type with SetUp and TearDown methods to call */ \
```
bbowman: this is the same assumption that gtest makes.

tests/functionaltests/functionaltest-api.h
```diff
@@ -385,4 +402,88 @@
...
+    \
+    class className##_##methodName : className { \
```
bbowman: >className##_##methodName [](start = 10, length = 24)

assumed to be unique. Also means that we don't get nice grouping in the output but I'll take what we can get here. An alternative would be to add new macros that do nothing in gtest land but make a new test class in taef land. the normal TEST macros could then just become TEST_METHOD. This would let us do fancier things with test class properties but would require modifying every test file and firmly put us in a franken test framework state that is neither GTEST or TAEF.

tests/functionaltests/functionaltest-api.h
```diff
@@ -385,4 +402,88 @@
...
+    }; \
+    \
+    bool className##_##methodName::_Setup() { \
+        SetUp(); \
+        return true; \
+    } \
+    bool className##_##methodName::_Teardown() { \
+        TearDown(); \
+        return true; \
+    } \
```
bbowman: needed for bool conversion. Maybe should guard differently like with a try/catch to indicate success?

tests/functionaltests/functionaltest-api.h
```diff
@@ -385,4 +402,88 @@
...
+        TEST_CLASS(className##_##methodName) \
+    \
+        TEST_METHOD(Test); \
+        TEST_METHOD_SETUP(_Setup); \
+        TEST_METHOD_CLEANUP(_Teardown); \
+        \
+        /* TODO: fix up inheritance to not use GTestClasses if not needed. */ \
+        virtual void TestBody() { \
+            Test(); \
+        } \
```
bbowman: overly cautious here. Test could be TestBody so this funciton isn't needed but I didn't want to mess with trying to figure out if the attributes etc on the functions matched.

tests/functionaltests/functionaltest-api.h
```diff
@@ -385,4 +402,88 @@
...
+#define  INSTANTIATE_TEST_CASE_P(prefix, className, generator, ...)                  \
+/* This assumes that there already exists a className type with SetUp and TearDown methods to call */ \
+    \
+    class className##_##prefix { \
+        TEST_CLASS(className##_##prefix) \
+    \
+        TEST_METHOD(Test) { \
+            WEX::Logging::Log::Comment(L"\nStarting parameterized subtests for " L#className L"_" L#prefix L":"); \
+            ::testing::internal::ParamGenerator<className::ParamType> paramGenerator(generator); \
+            for (const auto& metaFactory : ::testing::UnitTest::GetInstance()->parameterized_test_registry(). \
```
bbowman: >for (const auto& metaFactory : ::testing::UnitTest::GetInstance()->parameterized_test_registry() [](start = 12, length = 96)

This is unfortunately super extra double complicated because TEST_P does (className, testName) and INSTANTIATE_TEST_P does (prefix, className) but expects all values to be run over all TEST_P for a matching className. This means we need to know about all classes that derive from className here to iterate over. Luckily the GTEST framework handles this but it feels super icky to reach in like this.

bbowman:
No longer needed. Was rolled into different PR.

PR #1530, opened 2016-12-13
Properly calculate CALayer's ContentsCenter by translating them to co…

jaredhms:
…rresponding Xaml Image NineGrid values.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1530)

### DIFF
--- a/Frameworks/UIKit/StarboardXaml/LayerCoordinator.cpp
+++ b/Frameworks/UIKit/StarboardXaml/LayerCoordinator.cpp
@@ -604,6 +604,7 @@ DependencyProperty^ LayerCoordinator::s_visualHeightProperty = nullptr;
 DependencyProperty^ LayerCoordinator::s_contentGravityProperty = nullptr;
 DependencyProperty^ LayerCoordinator::s_contentCenterProperty = nullptr;
 DependencyProperty^ LayerCoordinator::s_contentSizeProperty = nullptr;
+DependencyProperty^ LayerCoordinator::s_contentScaleProperty = nullptr;
 
 void LayerCoordinator::InitializeFrameworkElement(FrameworkElement^ element) {
     // No-op if already registered
@@ -701,6 +702,9 @@ void LayerCoordinator::SetContent(FrameworkElement^ element, ImageSource^ source
     }
     _SetContentSize(element, Size(width / scale, height / scale));
 
+    // Store the content scale
+    _SetContentScale(element, scale);
+
     // Refresh any content center settings
     _ApplyContentCenter(element, _GetContentCenter(element));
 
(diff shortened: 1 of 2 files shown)


yiyang-msft:
:shipit:

Frameworks/UIKit/StarboardXaml/LayerCoordinator.cpp
```diff
@@ -752,10 +756,15 @@ void LayerCoordinator::_ApplyContentCenter(FrameworkElement^ element, Rect conte
     if (contentCenter.Equals(Rect(0, 0, 1.0, 1.0))) {
         image->NineGrid = Thickness(0, 0, 0, 0);
     } else {
-        int left = static_cast<int>(contentCenter.X * contentSize.Width);
-        int top = static_cast<int>(contentCenter.Y * contentSize.Height);
-        int right = (static_cast<int>(contentSize.Width) - (left + (static_cast<int>(contentCenter.Width * contentSize.Width))));
-        int bottom = (static_cast<int>(contentSize.Height) - (top + (static_cast<int>(contentCenter.Height * contentSize.Height))));
+        // When creating the NineGrid, we need to take the scaled image dimensions into account
+        const double contentScale = _GetContentScale(element);
+        const double imageWidth = contentSize.Width * contentScale;
```
tadam-msft: >ontentSize.Width * contentScale [](start = 35, length = 31)

Should we save the original width and height from SetContent--this is undoing the divide by scale done in that function.

Frameworks/UIKit/StarboardXaml/LayerCoordinator.cpp
```diff
@@ -765,12 +774,12 @@ void LayerCoordinator::_ApplyContentCenter(FrameworkElement^ element, Rect conte
         bottom = std::max<int>(0, bottom);
 
         // Cap the left/right to the maximum width
-        int maxWidth = static_cast<int>(contentSize.Width / 3);
+        int maxWidth = static_cast<int>(imageWidth / 3);
```
jaredhms: This caps left/right/top/bottom portions of the NineGrid to not exceed the size of the center - otherwise the Xaml Image won't render at all.

---
In reply to: [92279685](https://github.com/Microsoft/WinObjC/pull/1530#discussion_r92279685) [](ancestors = 92279685)

Frameworks/UIKit/StarboardXaml/LayerCoordinator.cpp
jaredhms: I think I prefer this approach.  This is the only place we have to 'undo' the scale, whereas the contentSize is used in more places and it was more convenient to divide the scale before setting the contentSize.  For example, the _ApplyContentGravity function would be a bit messier.  Thoughts?

---
In reply to: [92278738](https://github.com/Microsoft/WinObjC/pull/1530#discussion_r92278738) [](ancestors = 92278738)

PR #1533, opened 2016-12-13
Update objc2winmd binary: Fixed UI element rendered at wrong place

mukhole:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1533)


pradipd:
:shipit:

PR #1539, opened 2016-12-14
Fixing a UIButton backgroundimage-stretching regression found in IslandwoodHouse

jaredhms:
Our UIButton implementation renders its background via the backing CALayer's content image (similar to how this was done in our legacy UIButton implementation).

IslandwoodHouse loads its UIButtons from nibs; some of which were setting their contentMode values to UIViewContentModeScaleAspectFit.  We internally translate UIViewContentModeScaleAspectFit on a UIView to kCAGravityResizeAspect on its backing CALayer.  With kCAGravityResizeAspect set on the CALayer, we no longer stretch the button's background image (by design for kCAGravityResizeAspect).

The fix is to *always* use kCAGravityResize for UIButton's CALayer's contentsGravity, so we're sure to always stretch the background image (again, this is how it was done in the legacy UIButton implementation).

Also modifying LayerCoordinator to make it possible to view each layer's ContentsGravity value in the VS debugger.

Filing a separate issue to add samples that exercise this scenario.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1539)

### DIFF
--- a/Frameworks/UIKit/StarboardXaml/LayerCoordinator.cpp
+++ b/Frameworks/UIKit/StarboardXaml/LayerCoordinator.cpp
@@ -824,11 +824,12 @@ void LayerCoordinator::_RegisterDependencyProperties() {
             ref new PropertyMetadata(0.0,
             ref new PropertyChangedCallback(&LayerCoordinator::_VisualHeightChangedCallback)));
 
+        // Store as an int (rather than as a ContentGravity enum value) so we can read the values in the VS debugger at runtime
         s_contentGravityProperty = DependencyProperty::RegisterAttached(
             "ContentGravity",
-            ContentGravity::typeid,
+            int::typeid,
             FrameworkElement::typeid,
-            ref new PropertyMetadata(ContentGravity::Resize, nullptr));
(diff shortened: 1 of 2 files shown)


yiyang-msft:
:shipit:

Frameworks/UIKit/StarboardXaml/LayerCoordinator.cpp
```diff
@@ -1060,11 +1061,13 @@ void LayerCoordinator::_VisualHeightChangedCallback(DependencyObject^ sender, De
 
 // ContentGravity
 ContentGravity LayerCoordinator::GetContentGravity(FrameworkElement^ element) {
-    return static_cast<ContentGravity>(element->GetValue(s_contentGravityProperty));
+    // We store as an int (rather than as a ContentGravity enum value) so we can read the values in the VS debugger at runtime
+    return static_cast<ContentGravity>(static_cast<int>(element->GetValue(s_contentGravityProperty)));
```
jaredhms: I think the difficulty in debugging this without the change is worth the extra boxing cost (I didn't notice any perf issues with the change).  If this ends up being a hot path, we can revisit though.

PR #1545, opened 2016-12-15
Fix '1544: Use-after-free AV possible in CAAnimation'

### LINKED ISSUE #1544: Use-after-free AV possible in CAAnimation
It's possible to hit this while playing through IslandwoodHouse on recent builds; [CAAnimation _removeAnimationsFromLayer] can call into _attachedLayer to remove the **final** reference to the CAAnimation instance, resulting in an AV when it then attempts to clear out self's _attachedLayer and set self's _wasRemoved to TRUE.  Thanks @DHowett-MSFT for assistance with root causing the AV.

jaredhms:
Fix #1544 : Use-after-free AV possible in CAAnimation - grab a strong reference to self before clearing from _attachedLayer

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1545)

### DIFF
--- a/Frameworks/UIKit/StarboardXaml/LayerCoordinator.cpp
+++ b/Frameworks/UIKit/StarboardXaml/LayerCoordinator.cpp
@@ -824,11 +824,12 @@ void LayerCoordinator::_RegisterDependencyProperties() {
             ref new PropertyMetadata(0.0,
             ref new PropertyChangedCallback(&LayerCoordinator::_VisualHeightChangedCallback)));
 
+        // Store as an int (rather than as a ContentGravity enum value) so we can read the values in the VS debugger at runtime
         s_contentGravityProperty = DependencyProperty::RegisterAttached(
             "ContentGravity",
-            ContentGravity::typeid,
+            int::typeid,
             FrameworkElement::typeid,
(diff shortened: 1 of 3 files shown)


Frameworks/UIKit/StarboardXaml/LayerCoordinator.cpp
```diff
@@ -1060,11 +1061,13 @@ void LayerCoordinator::_VisualHeightChangedCallback(DependencyObject^ sender, De
 
 // ContentGravity
 ContentGravity LayerCoordinator::GetContentGravity(FrameworkElement^ element) {
-    return static_cast<ContentGravity>(element->GetValue(s_contentGravityProperty));
+    // We store as an int (rather than as a ContentGravity enum value) so we can read the values in the VS debugger at runtime
```
yiyang-msft: sorry, changes in UIButton/LayerCoordnator

---
In reply to: [92522383](https://github.com/Microsoft/WinObjC/pull/1545#discussion_r92522383) [](ancestors = 92522383)

jaredhms: doh - let me fix this up

PR #1546, opened 2016-12-15
Show back and forward button on WebView only if it is possible to moe back or forward

mukhole:
We use the properties on Windows.Xaml.Control.WebView, canGoForward and canGoBack to determine when to show these buttons.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1546)

### DIFF
--- a/Frameworks/SafariServices/SFSafariWebViewController.mm
+++ b/Frameworks/SafariServices/SFSafariWebViewController.mm
@@ -49,6 +49,10 @@ static bool appCanOpenURL(NSURL* url) {
     return false;
 }
 
+@interface SFSafariWebViewController ()
+- (void)_updateUIToolbarButtons;
+@end
+
 // Delegate for the UIWebView that underlies a SFSafariWebViewController
 @interface _SFWebViewDelegate : NSObject <UIWebViewDelegate>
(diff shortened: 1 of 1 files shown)


Frameworks/SafariServices/SFSafariWebViewController.mm
```diff
@@ -177,24 +177,46 @@ - (void)loadView {
...
+
+    if ([_webView canGoBack] == YES) {
+        // <-
+        UIBarButtonItem* backButton =
+            [[UIBarButtonItem alloc] initWithImage:backArrow style:UIBarButtonItemStylePlain target:_webView action:@selector(goBack)];
+        [toolbarItemsArray addObject:backButton];
+    } else {
+        // This is to ensure that the other buttons are rendered at the correct position.
+        [toolbarItemsArray addObject:space];
+    }
```
jaredhms: This only works when we load the view; what happens after we navigate in the WebView?  Don't we want to update our buttons as we navigate around?

mukhole: Yes, I misunderstood the way webview loads.
Will update and send another PR.

pradipd:
:shipit:

PR #1547, opened 2016-12-15
Fix '#1544: Use-after-free AV possible in CAAnimation' - grab a strong…

jaredhms:
… reference to self before clearing from _attachedLayer

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1547)

### DIFF
--- a/Frameworks/QuartzCore/CAAnimation.mm
+++ b/Frameworks/QuartzCore/CAAnimation.mm
@@ -295,11 +295,14 @@ static const wchar_t* TAG = L"CAAnimation";
         [layer _displayChanged];
     }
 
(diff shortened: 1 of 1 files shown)


yiyang-msft:
:shipit:

DHowett-MSFT:
:shipit:

DHowett-MSFT:
This does solve the issue? Neat.

jaredhms:
Yep; no repro with the fix.  Thanks again for the help!

---
In reply to: [267209718](https://github.com/Microsoft/WinObjC/pull/1547#issuecomment-267209718) [](ancestors = 267209718)

PR #1550, opened 2016-12-15
Fix Affine Transform in CGPath

### LINKED ISSUE #1179: Fix Affine Transform in CGPath
Affine Transform is broken for a few methods that require access to the individual figures of a D2DGeometry. This includes the function CGPathAddPath.

MSFTFox:
Implement the CGPathCreateMutableCopyByTransformingPath API which allows CGContext to properly transform any paths in conjunction with the User-To-Device transform. Unit tests show identical output.

Fixes #1179

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1550)

### DIFF
--- a/Frameworks/CoreGraphics/CGPath.mm
+++ b/Frameworks/CoreGraphics/CGPath.mm
@@ -754,21 +754,34 @@ CGPathRef CGPathCreateCopyByStrokingPath(
 }
 
 /**
- @Status Stub
- @Notes
+ @Status Caveat
+ @Notes Creates a mutable copy
 */
 CGPathRef CGPathCreateCopyByTransformingPath(CGPathRef path, const CGAffineTransform* transform) {
-    UNIMPLEMENTED();
-    return StubReturn();
+    return CGPathCreateMutableCopyByTransformingPath(path, transform);
 }
 
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 CGMutablePathRef CGPathCreateMutableCopyByTransformingPath(CGPathRef path, const CGAffineTransform* transform) {
-    UNIMPLEMENTED();
-    return StubReturn();
+    RETURN_NULL_IF(!path);
+
+    if (transform) {
+        if (CGAffineTransformEqualToTransform(*transform, CGAffineTransformIdentity)) {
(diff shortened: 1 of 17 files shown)


Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -754,21 +754,34 @@ CGPathRef CGPathCreateCopyByStrokingPath(
...
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 CGMutablePathRef CGPathCreateMutableCopyByTransformingPath(CGPathRef path, const CGAffineTransform* transform) {
-    UNIMPLEMENTED();
-    return StubReturn();
+    RETURN_NULL_IF(!path);
+
+    if (transform) {
```
DHowett-MSFT: nit: fold these together? `if (transform && !CGAffine...`

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -754,21 +754,34 @@ CGPathRef CGPathCreateCopyByStrokingPath(
...
+        if (CGAffineTransformEqualToTransform(*transform, CGAffineTransformIdentity)) {
+            return CGPathCreateMutableCopy(path);
+        }
+
+        CGMutablePathRef transformedPath = CGPathCreateMutable();
+        path->ClosePath();
+
+        FAIL_FAST_IF_FAILED(transformedPath->AddGeometryToPathWithTransformation(path->GetPathGeometry().Get(), transform));
```
DHowett-MSFT: It bothers me a bit that GetPathGeometry returns a `ComPtr<ID2D1Geometry>`; it's an unnecessary +1/-1/+1 for this case (and an unnecessary +1/-1 for every other case where you call it (!)

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -754,21 +754,34 @@ CGPathRef CGPathCreateCopyByStrokingPath(
 }
 
 /**
- @Status Stub
- @Notes
+ @Status Caveat
```
rajsesh-msft: I think this should be interoperable.  It would be a caveat if ref plat created mutable copy and we didn't.

PR #1555, opened 2016-12-15
Add support for CGImageCreateWithMask.

### LINKED ISSUE #1425: Implement CGImageCreateWithMask

DHowett-MSFT:
Fixes #1425.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1555)

### DIFF
--- a/Frameworks/CoreGraphics/CGImage.mm
+++ b/Frameworks/CoreGraphics/CGImage.mm
@@ -525,13 +525,34 @@ CGImageRef CGImageCreateCopyWithColorSpace(CGImageRef ref, CGColorSpaceRef color
 }
 
 /**
- @Status Stub
+ @Status Interoperable
+ @Notes This function does not defer the composition of its mask.
+        This function supports non-grayscale alpha masks, unlike the reference platform.
  */
 CGImageRef CGImageCreateWithMask(CGImageRef image, CGImageRef mask) {
-    // TODO  #1425 - masks are applied during rendering via D2D.
     RETURN_NULL_IF(!image);
-    UNIMPLEMENTED();
-    return StubReturn();
+    RETURN_NULL_IF(!mask);
+    RETURN_NULL_IF(CGImageIsMask(image));
+
+    size_t width = CGImageGetWidth(image);
+    size_t height = CGImageGetHeight(image);
+
+    woc::unique_cf<CGContextRef> context{ CGBitmapContextCreate(nullptr,
+                                                                width,
+                                                                height,
+                                                                8,
+                                                                width * 4,
+                                                                CGImageGetColorSpace(image),
+                                                                kCGImageAlphaPremultipliedFirst | kCGBitmapByteOrder32Big) };
+    RETURN_NULL_IF(!context);
+
+    CGRect rect{
+        CGPointZero, { width, height },
+    };
+    CGContextClipToMask(context.get(), rect, mask);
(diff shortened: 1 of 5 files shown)


Frameworks/CoreGraphics/CGImage.mm
```diff
@@ -525,13 +525,33 @@ CGImageRef CGImageCreateCopyWithColorSpace(CGImageRef ref, CGColorSpaceRef color
...
+ @Notes This function does not defer the composition of its mask.
+        This function supports non-grayscale alpha masks, unlike the reference platform.
  */
 CGImageRef CGImageCreateWithMask(CGImageRef image, CGImageRef mask) {
-    // TODO  #1425 - masks are applied during rendering via D2D.
     RETURN_NULL_IF(!image);
-    UNIMPLEMENTED();
-    return StubReturn();
+    RETURN_NULL_IF(!mask);
+    RETURN_NULL_IF(CGImageIsMask(image));
```
msft-Jeyaram: >CGImageIsMask [](start = 19, length = 13)

return null if it's not a mask?

DHowett-MSFT: `image` is the input image, it is contractually disallowed from being a mask.

Frameworks/CoreGraphics/CGImage.mm
```diff
@@ -525,13 +525,33 @@ CGImageRef CGImageCreateCopyWithColorSpace(CGImageRef ref, CGColorSpaceRef color
...
+
+    size_t width = CGImageGetWidth(image);
+    size_t height = CGImageGetHeight(image);
+
+    woc::unique_cf<CGContextRef> context{ CGBitmapContextCreate(nullptr,
```
DHowett-MSFT: good idea. this will return a cascading null, but it's better to be safe.

msft-Jeyaram: Yep, I'm thinking if someone decides to change the bitmap info (without realizing we don't support all of it due to D2D restrictions), we would want to catch that failure fast.

---
In reply to: [92708281](https://github.com/Microsoft/WinObjC/pull/1555#discussion_r92708281) [](ancestors = 92708281)

Frameworks/CoreGraphics/CGImage.mm
msft-Jeyaram: Ops, LOL I was thinking it was mask.

Can you add a check to see if the mask is really as mask?

---
In reply to: [92708221](https://github.com/Microsoft/WinObjC/pull/1555#discussion_r92708221) [](ancestors = 92708221)

msft-Jeyaram: some how my comment won't appear:

Ops, LOL I was thinking it was mask.

Can you add a check to see if the mask is really as mask?

DHowett-MSFT: See notes above; the reference platform requires a _mask_ (black/white image). We add support for an alpha clip (alpha image), but an alpha image _is not a mask_.

Frameworks/CoreGraphics/CGImage.mm
DHowett-MSFT: They can't change the bitmap info. Do you mean some internal consumer?

MSFTFox:
:shipit:

PR #1558, opened 2016-12-15
Fix some CoreText perf low-hanging fruit

ms-jihua:
- Remove NSLayoutManager __lineHasGlyphsAfterIndex(), which is called multiple times per line
    - Instead, directly search for the index of the last visible glyph once, compare against this

 - DWriteWrapper_CoreText
    - Skip first range of attributes in __DWriteTextLayoutCreate()
        - was redundant due to the underlaying Format already taking it into account
    - reserve() ahead of time for glyphRunDescriptionInfo._clusterMap, CTRun->_glyphOrigins, ->_glyphAdvances
        - A lot of time was being spent in reallocation, resizing

 - Remove DWriteWrapper _GetUserDefaultLocaleName(), don't wrap in a wstring, directly use a wchar_t buffer
 - Reduce the number of character buffer copies in _CFStringFromLocalizedString() by 1
 - Remove unused/not useful _characters member from CTTypesetter

 - Misc

Related to #1374

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1558)

### DIFF
--- a/Frameworks/CoreGraphics/DWriteWrapper.mm
+++ b/Frameworks/CoreGraphics/DWriteWrapper.mm
@@ -96,17 +96,10 @@ static std::shared_ptr<UserFontCollectionHelper> __GetUserFontCollectionHelper()
     return s_userFontCollection;
 }
 
-/**
- * Helper method to return the user set default locale string.
- *
- * @return use set locale string as wstring.
- */
-std::wstring _GetUserDefaultLocaleName() {
-    wchar_t localeName[LOCALE_NAME_MAX_LENGTH];
-    int defaultLocaleSuccess = GetUserDefaultLocaleName(localeName, LOCALE_NAME_MAX_LENGTH);
-
-    // If the default locale is returned, find that locale name, otherwise use "en-us".
-    return std::wstring(defaultLocaleSuccess ? localeName : c_defaultUserLanguage);
+// Private helper, wraps around GetUserDefaultLocaleName() and returns a default of "en-us" if it fails
+static inline const wchar_t* __GetUserDefaultLocaleName(wchar_t* buf, size_t bufferSize) {
+    int defaultLocaleSuccess = GetUserDefaultLocaleName(buf, bufferSize);
+    return defaultLocaleSuccess ? buf : c_defaultUserLanguage;
 }
 
 /**
@@ -122,14 +115,13 @@ CFStringRef _CFStringFromLocalizedString(IDWriteLocalizedStrings* localizedStrin
         return nil;
     }
 
-    // Get the default locale for this user.
-    std::wstring localeName = _GetUserDefaultLocaleName();
+    wchar_t localeNameBuf[LOCALE_NAME_MAX_LENGTH];
+    const wchar_t* localeName = __GetUserDefaultLocaleName(localeNameBuf, LOCALE_NAME_MAX_LENGTH);
 
     uint32_t index = 0;
     BOOL exists = FALSE;
 
-    // If the default locale is returned, find that locale name, otherwise use "en-us".
-    RETURN_NULL_IF_FAILED(localizedString->FindLocaleName(localeName.c_str(), &index, &exists));
+    RETURN_NULL_IF_FAILED(localizedString->FindLocaleName(localeName, &index, &exists));
     if (!exists) {
         RETURN_NULL_IF_FAILED(localizedString->FindLocaleName(c_defaultUserLanguage, &index, &exists));
     }
@@ -143,13 +135,15 @@ CFStringRef _CFStringFromLocalizedString(IDWriteLocalizedStrings* localizedStrin
     uint32_t length = 0;
     RETURN_NULL_IF_FAILED(localizedString->GetStringLength(index, &length));
 
-    // Get the string.
-    std::vector<wchar_t> wcharString = std::vector<wchar_t>(length + 1, 0);
-    RETURN_NULL_IF_FAILED(localizedString->GetString(index, wcharString.data(), length + 1));
+    // Get the string. Use length + 1 here as GetString requires room for the null terminator.
+    woc::unique_iw<wchar_t> wcharString(static_cast<wchar_t*>(IwMalloc(sizeof(wchar_t) * (length + 1))));
+    RETURN_NULL_IF_FAILED(localizedString->GetString(index, wcharString.get(), length + 1));
 
     // Strip out unnecessary null terminator
-    return (CFStringRef)CFAutorelease(
-        CFStringCreateWithCharacters(kCFAllocatorDefault, reinterpret_cast<UniChar*>(wcharString.data()), length));
+    return (CFStringRef)CFAutorelease(CFStringCreateWithCharactersNoCopy(kCFAllocatorDefault,
+                                                                         reinterpret_cast<UniChar*>(wcharString.release()),
+                                                                         length,
+                                                                         kCFAllocatorDefault));
 }
 
 /**
@@ -280,6 +274,9 @@ HRESULT _DWriteCreateTextFormatWithFontNameAndSize(CFStringRef optionalFontName,
         userFontInfo = __GetUserFontCollectionHelper()->GetFontPropertiesFromUppercaseFontName(upperFontName);
     }
 
+    wchar_t localeNameBuf[LOCALE_NAME_MAX_LENGTH];
+    const wchar_t* localeName = __GetUserDefaultLocaleName(localeNameBuf, LOCALE_NAME_MAX_LENGTH);
+
     if (info) {
         RETURN_IF_FAILED(
             dwriteFactory->CreateTextFormat(reinterpret_cast<wchar_t*>(Strings::VectorFromCFString(info->familyName.get()).data()),
@@ -288,7 +285,7 @@ HRESULT _DWriteCreateTextFormatWithFontNameAndSize(CFStringRef optionalFontName,
                                             info->style,
                                             info->stretch,
                                             fontSize,
-                                            _GetUserDefaultLocaleName().data(),
+                                            localeName,
                                             &textFormat));
     }
 
@@ -301,7 +298,7 @@ HRESULT _DWriteCreateTextFormatWithFontNameAndSize(CFStringRef optionalFontName,
                                             userFontInfo->style,
                                             userFontInfo->stretch,
                                             fontSize,
-                                            _GetUserDefaultLocaleName().data(),
+                                            localeName,
                                             &textFormat));
     }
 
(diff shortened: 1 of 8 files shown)


Frameworks/UIKit/NSLayoutManager.mm
```diff
@@ -72,105 +85,107 @@ - (void)__layoutAllText {
...
                 }
 
                 continue;
             }
 
-            woc::unique_cf<CGPathRef> path{ CGPathCreateWithRect(CGRectMake(0, 0, rect.size.width, lineHeight), nullptr) };
+            frameDrawRect.size.width = rect.size.width;
+            woc::unique_cf<CGPathRef> path{ CGPathCreateWithRect(frameDrawRect, nullptr) };
             woc::unique_cf<CTFrameRef> frame{
                 CTFramesetterCreateFrame(framesetter.get(), { stringIndex, lastIndex - stringIndex }, path.get(), nullptr)
```
ms-jihua: I'm throwing in the towel on trying to batch these for the time being. There's a general chicken-and-egg problem re:batching here where some information required for the next line isn't known until the current one is drawn. We can estimate these like lastIndex above, but the more we estimate before grounding ourselves, the more likely it is that our estimates will diverge. Not really sure how to tackle this for the time being.

Frameworks/UIKit/NSLayoutManager.mm
```diff
@@ -72,105 +85,107 @@ - (void)__layoutAllText {
...
             CTLineRef fitLine = static_cast<CTLineRef>(CFArrayGetValueAtIndex(CTFrameGetLines(frame.get()), 0));
 
-            CFIndex fitLength = CTLineGetStringRange(fitLine).length;
+            const CFIndex fitLength = CTLineGetStringRange(fitLine).length;
             if (fitLength == 0L) {
                 // Failed to fit any text in the current rect, continue to next
                 if (remainingRect.size.width > 0.0f && stringIndex < lineEnd) {
                     origin = { remainingRect.origin.x, origin.y };
                 } else {
                     origin = { 0.0f, origin.y + lineHeight };
```
ms-jihua: >+ lineHeight [](start = 46, length = 12)

note that this is not always correct, but I'm not sure how best to deal with it, and this is a good approximation for the time being.

a single character with a particularly large font+fontsize in the _paragraph_ affects this lineHeight metric, which is computed from the height across the whole paragraph, even if the chars in the current line are a smaller font size. we can't just take the height of the line we actually drew, either, because if it's just a period or an underscore, that'll throw off the height measurement. what we actually need here is the height of an 'H' in the largest font+fontsize in this current line, but this is messy to determine :s

ms-jihua:
>             RETURN_IF_FAILED(typography->AddFontFeature({ DWRITE_FONT_FEATURE_TAG_KERNING, ++incompatibleAttributeFlag }));

@aballway, to be sure, this should still fulfill the intended functionality? if i'm reading correctly, we previously had this as 1, 2, 3... and now we should have 0, 1, 2...

---
Refers to: Frameworks/CoreText/DWriteWrapper_CoreText.mm:296 in b491db6. [](commit_id = b491db621b778f6ef04f07f80f57a383dff29bcf, deletion_comment = False)

ms-jihua:
>             RETURN_IF_FAILED(typography->AddFontFeature({ DWRITE_FONT_FEATURE_TAG_KERNING, ++incompatibleAttributeFlag }));

oh hmm was I overeager? @aballway did I mess this up actually?

---
Refers to: Frameworks/CoreText/DWriteWrapper_CoreText.mm:296 in b491db6. [](commit_id = b491db621b778f6ef04f07f80f57a383dff29bcf, deletion_comment = False)

aballway:
>             RETURN_IF_FAILED(typography->AddFontFeature({ DWRITE_FONT_FEATURE_TAG_KERNING, ++incompatibleAttributeFlag }));

no, this is being used to break apart the different attribute ranges.  Since this would be done on the second range, it shouldn't cause a problem.  I believe there are unit tests to verify this works.

---
In reply to: [267478579](https://github.com/Microsoft/WinObjC/pull/1558#issuecomment-267478579) [](ancestors = 267478579)

---
Refers to: Frameworks/CoreText/DWriteWrapper_CoreText.mm:296 in b491db6. [](commit_id = b491db621b778f6ef04f07f80f57a383dff29bcf, deletion_comment = False)

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -123,13 +110,15 @@ CFStringRef _CFStringFromLocalizedString(IDWriteLocalizedStrings* localizedStrin
     }
 
     // Get the default locale for this user.
-    std::wstring localeName = _GetUserDefaultLocaleName();
+    // If the default locale is returned, find that locale name, otherwise use "en-us".
+    wchar_t localeNameBuf[LOCALE_NAME_MAX_LENGTH];
+    int defaultLocaleSuccess = GetUserDefaultLocaleName(localeNameBuf, LOCALE_NAME_MAX_LENGTH);
+    const wchar_t* localeName = defaultLocaleSuccess ? localeNameBuf : c_defaultUserLanguage;
```
rajsesh-msft: could be factored into a private method. This could be a static as well, if we don't concern ourselves with locale changing when app is running.

rajsesh-msft: More importantly, is this really that big a perf impact (constructing the std::wstring object? 

---
In reply to: [92735069](https://github.com/Microsoft/WinObjC/pull/1558#discussion_r92735069) [](ancestors = 92735069)

ms-jihua: it _was_ a private method. i unrolled it because it would be clunky without the std::wstring return (caller would need to declare buffer/know size required, private function would only end up being two lines anyway, not saving much).

---
In reply to: [92735069](https://github.com/Microsoft/WinObjC/pull/1558#discussion_r92735069) [](ancestors = 92735069)

Frameworks/UIKit/NSLayoutManager.mm
```diff
@@ -34,22 +34,38 @@ @implementation NSLayoutManager {
...
+// Given a CTLineRef, returns the index of the character in the original string, that maps to the last visible glyph in the line
+static CFIndex __CTLineGetStringIndexOfLastVisibleGlyph(CTLineRef line) {
+    CFArrayRef runs = CTLineGetGlyphRuns(line);
+
+    CFIndex ret = -1;
+    for (CFIndex i = CFArrayGetCount(runs) - 1; i >= 0; --i) {
+        CTRunRef run = static_cast<CTRunRef>(CFArrayGetValueAtIndex(runs, i));
+        CFIndex glyphCount = CTRunGetGlyphCount(run);
+
+        // Open question: Can runs ever be non-monotonic with respect to the original string?
```
rajsesh-msft: >question [](start = 16, length = 8)

For one thing, this would fall apart for RTL languages?

Frameworks/CoreText/CTFramesetter.mm
```diff
@@ -32,7 +32,7 @@ @implementation _CTFramesetter : NSObject
     // Call _DWriteWrapper to get _CTLine object list that makes up this frame
     _CTTypesetter* typesetter = static_cast<_CTTypesetter*>(framesetter->_typesetter);
     if (range.length == 0L) {
-        range.length = typesetter->_characters.size() - range.location;
+        range.length = [typesetter->_string length] - range.location;
```
DHowett-MSFT: Perhaps prefer CFStringGetLength if it is provably a CFString.

ms-jihua: it's stored as an NSString :(

---
In reply to: [92743322](https://github.com/Microsoft/WinObjC/pull/1558#discussion_r92743322) [](ancestors = 92743322)

Frameworks/UIKit/NSLayoutManager.mm
ms-jihua: I can't imagine RTL languages would store either their glyphs, runs, or characters back-to-front in their respective arrays?
Also, note that this comment is meant to convey a future optimization that could be made, if it turns out that the assumption _is_ safe. I think the current logic should handle RTL languages just fine,

---
In reply to: [92738146](https://github.com/Microsoft/WinObjC/pull/1558#discussion_r92738146) [](ancestors = 92738146)

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -143,13 +132,15 @@ CFStringRef _CFStringFromLocalizedString(IDWriteLocalizedStrings* localizedStrin
     uint32_t length = 0;
     RETURN_NULL_IF_FAILED(localizedString->GetStringLength(index, &length));
 
-    // Get the string.
-    std::vector<wchar_t> wcharString = std::vector<wchar_t>(length + 1, 0);
-    RETURN_NULL_IF_FAILED(localizedString->GetString(index, wcharString.data(), length + 1));
+    // Get the string. Use length + 1 here as GetString requires room for the null terminator.
+    woc::unique_iw<wchar_t> wcharString(static_cast<wchar_t*>(IwMalloc(sizeof(wchar_t) * (length + 1))));
```
aballway: does this zero out the memory?  I faintly recall running into an issue because GetString wasn't copying over the null terminator.

Frameworks/UIKit/NSLayoutManager.mm
```diff
@@ -72,105 +85,107 @@ - (void)__layoutAllText {
         _frame.reset(CTFramesetterCreateFrame(framesetter.get(), {}, path.get(), nullptr));
     }
 
+    NSTextContainer* container = (NSTextContainer*)_textContainers[0];
+    const CGSize containerSize = container.size;
+    CGPoint origin{ 0, 0 };
+
     for (id lineRef in static_cast<NSArray*>(CTFrameGetLines(_frame.get()))) {
```
aballway: If you're just using the WOCCatalog stuff to test this, may want to mix up the text by tossing in some newlines to make sure those are handled well.  Everything here looks like it will work better, but I'm not sure if there are any edge cases snuck in.

Frameworks/CoreGraphics/DWriteWrapper.mm
ms-jihua: unit tests still pass, and we have a number of tests that would exercise this, through CT/CGFont

---
In reply to: [92845292](https://github.com/Microsoft/WinObjC/pull/1558#discussion_r92845292) [](ancestors = 92845292)

Frameworks/UIKit/NSLayoutManager.mm
ms-jihua: it seems to be fine

---
In reply to: [92847146](https://github.com/Microsoft/WinObjC/pull/1558#discussion_r92847146) [](ancestors = 92847146)

PR #1570, opened 2016-12-16
Getting more detailed error messages across ABI boundary.

pradipd:
3 changes:
1. Updated libobjc2.  Libobjc2 will call RoOriginateError to set IRestrictedErrorInfo on the thread.  This will provide more details for exceptions, that are converted to HRESULTSs as they cross the ABI boundary.

2. Updated objc2winmd to add an exception guard around  calls into winrt component.

3. Modified NSException _hresult to convert some NSException's to HRESULTs.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1570)


pradipd:

pradipd:

pradipd:

Frameworks/Foundation/NSException.mm
```diff
@@ -202,6 +202,29 @@ case __HRESULT_FROM_WIN32(ERROR_NOT_READY):
     return WinObjCException;
 }
 
+// Returns a HRESULT for an exception name
+// Note: The table is not the exact inverse of _exceptionNameForHRESULT. 
+// E.g. E_ACCESSDENIED and RPC_E_WRONG_THREAD both map to NSObjectInaccessibleException.
+// However, NSObjectInaccessibleException only maps to E_ACCESSDENIED
++ (int)_HRESULTForExceptionName:(NSString *)exceptionName {
```
DHowett-MSFT: nit: this should return `HRESULT`

Frameworks/include/FoundationErrorHandling.h
```diff
@@ -21,6 +21,7 @@
 
 @interface NSException (errorHandlingInternal)
 + (NSString*)_exceptionNameForHRESULT:(int)errorCode;
++ (int)_HRESULTForExceptionName:(NSString *)exceptionName;
```
DHowett-MSFT: nit: as above

Frameworks/Foundation/NSException.mm
```diff
@@ -216,7 +244,18 @@ - (HRESULT)_hresult {
         return [hresultValue unsignedIntValue];
     }
 
-    TraceWarning(L"NSException", L"Called -hresult on an NSException without an HRESULT!");
-    return E_UNEXPECTED;
+    return [[self class] _HRESULTForExceptionName:self.name];
+}
+
+- (void)_processException {
+    HRESULT hr = [self _hresult];
```
bbowman: >_hresult [](start = 23, length = 8)

responds to selector?

Frameworks/Foundation/NSException.mm
```diff
@@ -216,7 +244,18 @@ - (HRESULT)_hresult {
...
+    HSTRING_HEADER header;
+    HSTRING hsReason = nullptr;
+    
+    if(self.reason != nil) {
+        LPCWSTR reason = (LPCWSTR)([[self reason] cStringUsingEncoding:NSUnicodeStringEncoding]);
+        WindowsCreateStringReference(reason, wcslen(reason), &header, &hsReason);
+    }
+    RoOriginateError(hr, hsReason);
```
bbowman: >hr [](start = 21, length = 2)

what if this is S_OK?

Frameworks/Foundation/NSException.mm
```diff
@@ -216,7 +244,18 @@ - (HRESULT)_hresult {
...
+}
+
+- (void)_processException {
+    HRESULT hr = [self _hresult];
+    HSTRING_HEADER header;
+    HSTRING hsReason = nullptr;
+    
+    if(self.reason != nil) {
+        LPCWSTR reason = (LPCWSTR)([[self reason] cStringUsingEncoding:NSUnicodeStringEncoding]);
+        WindowsCreateStringReference(reason, wcslen(reason), &header, &hsReason);
```
bbowman: You leak this. :-( 

Try using [https://msdn.microsoft.com/en-us/library/hh771147.aspx](https://msdn.microsoft.com/en-us/library/hh771147.aspx) as mentioned in previous iteration.

bbowman:
:clock1:

Frameworks/Foundation/NSException.mm
pradipd: Hmm. That link wasn't all that helpful.  Or maybe I didn't get the jist of it.
Did I leak reason?  or hsReason?
From the link, as I read it, it sounds like I'm doing this correctly.


---
In reply to: [93125423](https://github.com/Microsoft/WinObjC/pull/1570#discussion_r93125423) [](ancestors = 93125423)

pradipd: reason is the backing store for hsReason and gets freed up when the autorelase pool is drained.  That is my understanding.


---
In reply to: [93297284](https://github.com/Microsoft/WinObjC/pull/1570#discussion_r93297284) [](ancestors = 93297284,93125423)

Frameworks/Foundation/NSException.mm
bbowman: seems like a weird internal implementation detail to rely on here. 

---
In reply to: [93300439](https://github.com/Microsoft/WinObjC/pull/1570#discussion_r93300439) [](ancestors = 93300439,93125265)

Frameworks/Foundation/NSException.mm
pradipd: Why? Do we normally check to this?  I don't see us doing this consistently in other files.

---
In reply to: [93125224](https://github.com/Microsoft/WinObjC/pull/1570#discussion_r93125224) [](ancestors = 93125224)

Frameworks/Foundation/NSException.mm
bbowman: The "jist" of it is that you should be using **Wrappers::HStringReference** instead of manually creating one and having to worry about its lifetime. 

I totally understand that "reason" is the backing store but that then begs the question of why don't need to delete the HSTRING that is presumably pointing to it. The interesting bit is the HEADER that is on the stack and that under the covers that header is being cast to the output HSTRING hsReason. This means that the storage for the HSTRING itself is just the header on the stack. 

The fact though that you had to read the remarks, and still weren't 100% certain what was going on here suggests that a lifetime managed object like the wrappers one will make the code more maintainable, less verbose, and all around more clear what you are intending. 

---
In reply to: [93299684](https://github.com/Microsoft/WinObjC/pull/1570#discussion_r93299684) [](ancestors = 93299684,93297284,93125423)

Frameworks/Foundation/NSException.mm
DHowett-MSFT: You can't get into `_processException` with a nil self to then call `_hresult` on a nil self.

It is completely safe to assume that a method declared on a class can access an adjacent method declared on the same class.

bbowman: Good point. Also be on vacation!

---
In reply to: [93306728](https://github.com/Microsoft/WinObjC/pull/1570#discussion_r93306728) [](ancestors = 93306728)

Frameworks/Foundation/NSException.mm
bbowman: The "jist" of it is that you should be using **Wrappers::HStringReference** instead of manually creating one and having to worry about its lifetime. 

I totally understand that "reason" is the backing store but that then begs the question of why don't need to delete the HSTRING that is presumably pointing to it. The interesting bit is the HEADER that is on the stack and that under the covers that header is being cast to the output HSTRING hsReason. This means that the storage for the HSTRING itself is just the header on the stack. 

The fact though that you had to read the remarks, and still weren't 100% certain what was going on here suggests that a lifetime managed object like the wrappers one will make the code more maintainable, less verbose, and all around more clear what you are intending.

Frameworks/Foundation/NSException.mm
pradipd: How is that "weird"?  It is clearly documented.


---
In reply to: [93300590](https://github.com/Microsoft/WinObjC/pull/1570#discussion_r93300590) [](ancestors = 93300590,93300439,93125265)

Frameworks/Foundation/NSException.mm
pradipd: Ah, now, I see (or at least I hope).  You *prefer* me to use Wrappers::HStringReference ?
If so, that is fine with me.

The original remark (from the original CR) was:
:-( Wrappers::HString*Reference*???
I did not know what you meant.    It is obvious to me now.
And then the original comment here:
You leak this. :-( (technically you don't....
Confused me even more because there was no leak.

For future reference, something like "prefer using Wrappers::HstringReference" would be more useful than ":-( Wrappers::HString*Reference*???"

I have no problems using HStringReference.  In retrospect, (again *painfully* obviously now) I don't know why I didn't use it in the first place.  Especially since I was pointed to it.  (Yeah, I'm feeling a little embarrassed.)


---
In reply to: [93301103](https://github.com/Microsoft/WinObjC/pull/1570#discussion_r93301103) [](ancestors = 93301103,93299684,93297284,93125423)

pradipd: And it is less code....
Again, feeling stupid I missed the 

"Try ***using**** [https://msdn.microsoft.com/en-us/library/hh771147.aspx](https://msdn.microsoft.com/en-us/library/hh771147.aspx) as mentioned in previous iteration."


---
In reply to: [93377905](https://github.com/Microsoft/WinObjC/pull/1570#discussion_r93377905) [](ancestors = 93377905,93301103,93299684,93297284,93125423)

Frameworks/Foundation/NSException.mm
bbowman: assert sounds good enough probably. Just seems like an obvious place to be a little defensive in our programming. 

---
In reply to: [93384408](https://github.com/Microsoft/WinObjC/pull/1570#discussion_r93384408) [](ancestors = 93384408,93375861,93375624,93300590,93300439,93125265)

Frameworks/Foundation/NSException.mm
pradipd: I believe Dustin answered this question.

---
In reply to: [93302629](https://github.com/Microsoft/WinObjC/pull/1570#discussion_r93302629) [](ancestors = 93302629,93301074,93125224)

Frameworks/Foundation/NSException.mm
pradipd: added assert

---
In reply to: [93495346](https://github.com/Microsoft/WinObjC/pull/1570#discussion_r93495346) [](ancestors = 93495346,93384408,93375861,93375624,93300590,93300439,93125265)

Frameworks/Foundation/NSException.mm
pradipd: Used MS::WRL::Wrappers::HstringReference

---
In reply to: [93378401](https://github.com/Microsoft/WinObjC/pull/1570#discussion_r93378401) [](ancestors = 93378401,93377905,93301103,93299684,93297284,93125423)

pradipd: Used HStringReference.

Also, I understood how this worked.  I just missed the  " you should be using **Wrappers::HStringReference**" from previous comments.  
I'm not sure why/how I missed that.

---
In reply to: [93327581](https://github.com/Microsoft/WinObjC/pull/1570#discussion_r93327581) [](ancestors = 93327581)

bbowman:
:shipit:

PR #1579, opened 2016-12-16
Rewrite CGDataProvider and change DWrite to use it directly

### LINKED ISSUE #1450: CGDataProvider is, but should not be, an NS/CFData.

aballway:
Rewrites CGDataProviderRef to subclass from CppBase rather than being NSData.  Changes the DWrite classes to (un)register fonts to use CGDataProviderRef directly.

Fixes #1450

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1579)


Frameworks/CoreGraphics/CGDataProvider.mm
```diff
@@ -16,130 +16,395 @@
...
+    ~__DirectDataProvider() {
+        if (m_data != nullptr) {
+            m_callbacks->releaseBytePointer(m_info, m_data);
+        }
+
+        m_callbacks->releaseInfo(m_info);
+    }
+};
+
+class __DirectAccessDataProvider : public __CGDataProviderInternal {
```
ms-jihua: nit: this class and the class above are the same except for the type of callback. seems ripe for a template?

Frameworks/include/CGDataProviderInternal.h
```diff
@@ -13,22 +13,8 @@
 // THE SOFTWARE.
 //
 //******************************************************************************
+#pragma once
```
DHowett-MSFT: nit: \n before

Frameworks/CoreGraphics/CGDataProvider.mm
```diff
@@ -16,130 +16,395 @@
...
+class __UserDataProvider : public __CGDataProviderInternal {
+    mutable void* m_info;
+    const CGDataProviderCallbacks* m_callbacks;
+    mutable std::vector<unsigned char> m_data;
 
-- (NSUInteger)length {
-    return [_data length];
-}
+public:
+    __UserDataProvider(void* info, const CGDataProviderCallbacks* callbacks) : m_info(info), m_callbacks(callbacks) {
```
DHowett-MSFT: >acks(callba [](start = 100, length = 11)

callbacks isn't guaranteed to outlive the caller's scope (it could be on the stack); take a copy

Frameworks/CoreGraphics/CGDataProvider.mm
```diff
@@ -16,130 +16,395 @@
...
+#import <vector>
+
+// Determines what amount of bytes to try and read at a time for indirect access data providers
+static constexpr size_t c_streamSize = 2048;
+
+#pragma region __CGDataProviderInternal
+namespace __CGDataProviderInternal {
+class __CGDataProviderInternal {
+    mutable std::unique_ptr<CFHashCode> m_hashValue;
+
```
DHowett-MSFT: are you using unique_ptr for std::optional?
Having a 4-byte indirect allocation is a minor waste :)

Frameworks/CoreGraphics/CGDataProvider.mm
```diff
@@ -16,130 +16,395 @@
...
-}
+        return memcmp(GetData(), rhs.GetData(), GetSize()) == 0;
+    }
 
-- (instancetype)init {
-    return [self initWithBytes:"" length:0];
-}
+    CFHashCode Hash() const {
+        if (!m_hashValue) {
+            m_hashValue.reset(new CFHashCode(CFHashBytes((UInt8*)GetData(), GetSize())));
```
DHowett-MSFT: threadsafe?

Frameworks/CoreGraphics/CGDataProvider.mm
```diff
@@ -16,130 +16,395 @@
...
+    bool operator==(const __CGDataProviderInternal& rhs) const {
+        if (GetSize() != rhs.GetSize()) {
+            return false;
+        }
 
-- (NSString*)_fileName {
-    return filename;
-}
+        return memcmp(GetData(), rhs.GetData(), GetSize()) == 0;
+    }
```
DHowett-MSFT: perhaps short-circuit via hash

Frameworks/CoreGraphics/CGDataProvider.mm
```diff
@@ -16,130 +16,395 @@
...
+
+    ~__UserDataProvider() {
+        m_callbacks->releaseProvider(m_info);
+    }
+};
+
+class __RawDataProvider : public __CGDataProviderInternal {
+    void* m_info;
+    const void* m_data;
+    size_t m_size;
```
DHowett-MSFT: do we have to take a copy?

---
In reply to: [92906947](https://github.com/Microsoft/WinObjC/pull/1579#discussion_r92906947) [](ancestors = 92906947)

Frameworks/CoreGraphics/CGDataProvider.mm
```diff
@@ -16,130 +16,395 @@
...
+    size_t GetSize() const override {
+        return m_size;
+    }
+
+    ~__RawDataProvider() {
+        m_releaseData(m_info, m_data, m_size);
+    }
+};
+
+class __DirectDataProvider : public __CGDataProviderInternal {
```
DHowett-MSFT: >__DirectDataProvider [](start = 6, length = 20)

DirectData and Raw look like the same thing; is there a chance of merging them?

Frameworks/CoreGraphics/CGDataProvider.mm
```diff
@@ -16,130 +16,395 @@
...
+    ~__DirectAccessDataProvider() {
+        if (m_data != nullptr) {
+            m_callbacks->releaseBytePointer(m_info, m_data);
+        }
+
+        m_callbacks->releaseProvider(m_info);
+    }
+};
+
+class __SequentialDataProvider : public __CGDataProviderInternal {
```
DHowett-MSFT: >__SequentialDataProvider [](start = 6, length = 24)

is this a user data provider with skipForward instead of skipBytes? :(

Frameworks/CoreGraphics/CGDataProvider.mm
```diff
@@ -16,130 +16,395 @@
...
+
+        return m_data.size();
+    }
+
+    ~__SequentialDataProvider() {
+        m_callbacks->releaseInfo(m_info);
+    }
+};
+
+class __CFDataProvider : public __CGDataProviderInternal {
```
DHowett-MSFT: >__CFDataProvider [](start = 6, length = 16)

nit: _CFDataDataProvider

Frameworks/CoreGraphics/CGDataProvider.mm
```diff
@@ -16,130 +16,395 @@
...
+    size_t GetSize() const {
+        return m_internal->GetSize();
+    }
+
+    bool operator==(const __CGDataProvider& other) {
+        return *m_internal == *(other.m_internal);
+    }
+
+    static CFTypeID GetTypeID() {
+        static __CFRuntimeClass _cls{
```
DHowett-MSFT: please note why this is overridden

Frameworks/CoreGraphics/CGDataProvider.mm
msft-Jeyaram: >UInt8 [](start = 58, length = 5)

nit:static_cast

Frameworks/CoreGraphics/CGDataProvider.mm
```diff
@@ -16,130 +16,395 @@
...
-    return [_data length];
-}
+public:
+    __UserDataProvider(void* info, const CGDataProviderCallbacks* callbacks) : m_info(info), m_callbacks(callbacks) {
+    }
 
-@end
+    const void* GetData() const override {
+        if (m_data.empty()) {
+            size_t count = 0, delta = 0;
```
msft-Jeyaram: >size_t count = 0, delta = 0; [](start = 12, length = 28)

declare them separately. This makes it less readable and more prone to error as one line is doing multiple things.

Frameworks/CoreGraphics/CGDataProvider.mm
```diff
@@ -16,130 +16,395 @@
...
+            do {
+                m_data.resize(m_data.size() + c_streamSize);
+                delta = m_callbacks->getBytes(m_info, m_data.data() + count, c_streamSize);
```
msft-Jeyaram: >m_callbacks [](start = 24, length = 11)

are you assuming the callback can never be invalid?

Frameworks/CoreGraphics/CGDataProvider.mm
```diff
@@ -16,130 +16,395 @@
...
+                count += delta;
+                m_callbacks->skipForward(m_info, delta);
+            } while (delta == c_streamSize);
+            m_data.resize(count);
```
msft-Jeyaram: template + function pointer
given most of these classes are pretty much the same with minor difference, it doesn't make since to duplicate code.

Frameworks/CoreGraphics/CGDataProvider.mm
```diff
@@ -156,16 +421,28 @@ void CGDataProviderRelease(CGDataProviderRef data) {
...
+#pragma endregion // CGDataProvider Functions
+
+#pragma region Internal CGDataProvider Functions
+
+const void* _CGDataProviderGetData(CGDataProviderRef provider) {
+    return provider ? provider->GetData() : nullptr;
+}
+
+size_t _CGDataProviderGetSize(CGDataProviderRef provider) {
+    return provider ? provider->GetSize() : 0;
```
msft-Jeyaram: > return provider ? provider->GetSize() : 0; [](start = 3, length = 43)

lets not do ternary statements.
It complicates stuff more than it needs to and makes it easier to lose during merges.

Frameworks/CoreGraphics/CGDataProvider.mm
```diff
@@ -156,16 +421,28 @@ void CGDataProviderRelease(CGDataProviderRef data) {
...
\ No newline at end of file
+    return __CGDataProvider::GetTypeID();
+}
+
+#pragma endregion // CGDataProvider Functions
+
+#pragma region Internal CGDataProvider Functions
+
+const void* _CGDataProviderGetData(CGDataProviderRef provider) {
+    return provider ? provider->GetData() : nullptr;
```
msft-Jeyaram: >vider ? provider->GetData() : nullpt [](start = 14, length = 36)

same as below

Frameworks/CoreGraphics/CGDataProvider.mm
```diff
@@ -16,130 +16,395 @@
...
-    if (!ret) {
-        return 0;
+CFDataRef CGDataProviderCopyData(CGDataProviderRef provider) {
+    CFDataRef ret = nullptr;
+    if (provider != nullptr) {
+        ret = CFDataCreate(nullptr, (const UInt8*)provider->GetData(), provider->GetSize());
     }
-    ret->filename = [NSString stringWithCString:filename];
 
     return ret;
```
msft-Jeyaram: RETURN_NULL_IF(!provider);
return CFDataCreate(nullptr, (const UInt8*)provider->GetData(), provider->GetSize());

Frameworks/CoreGraphics/CGDataProvider.mm
```diff
@@ -16,130 +16,395 @@
...
+CGDataProviderRef CGDataProviderCreateWithURL(CFURLRef url) {
+    return __CGDataProvider::CreateInstance(nullptr, url);
 }
 
 /**
  @Status Interoperable
 */
 CGDataProviderRef CGDataProviderCreateWithCFData(CFDataRef data) {
-    return (CGDataProviderRef)CFRetain(data);
+    return __CGDataProvider::CreateInstance(nullptr, data);
```
msft-Jeyaram: you aren't checking for invalid state here nor within the internal implementation.
please do so here and elsewhere

Frameworks/CoreGraphics/CGFont.mm
```diff
@@ -116,7 +116,7 @@ CGFontRef CGFontCreateWithDataProvider(CGDataProviderRef cgDataProvider) {
     CFAutorelease(ret);
     struct __CGFont* mutableRet = const_cast<struct __CGFont*>(ret);
```
msft-Jeyaram: duplicate code, make it into a helper

msft-Jeyaram:
>     size_t memSize = sizeof(struct __CGFont) - sizeof(CFRuntimeBase);

it would be nice to have it as cpp based at some point 

---
Refers to: Frameworks/CoreGraphics/CGFont.mm:129 in 9a3d204. [](commit_id = 9a3d204398ea7c113b82eacf4299d2f9c5cbd10a, deletion_comment = False)

msft-Jeyaram:
:clock1:

Frameworks/CoreGraphics/CGDataProvider.mm
aballway: Nope, reference platform uses the data directly, no copy taken

Frameworks/CoreGraphics/CGDataProvider.mm
aballway: Unfortunately the types and names of the callbacks in the struct are different, making templatizing this more trouble than it's worth.  Ends up making the code a lot more complex imo.

Frameworks/CoreGraphics/CGDataProvider.mm
msft-Jeyaram: curios why you don't want to fix this?
we seen errors introduced easily via this and not a good practice.

---
In reply to: [92931361](https://github.com/Microsoft/WinObjC/pull/1579#discussion_r92931361) [](ancestors = 92931361)

Frameworks/CoreGraphics/CGDataProvider.mm
msft-Jeyaram: curious why you don't want to fix this?

---
In reply to: [92931590](https://github.com/Microsoft/WinObjC/pull/1579#discussion_r92931590) [](ancestors = 92931590)

Frameworks/CoreGraphics/CGDataProvider.mm
aballway: I don't think there are significant gains in readability by splitting this into two lines.  If anything I find it more readable in one line.

---
In reply to: [93115225](https://github.com/Microsoft/WinObjC/pull/1579#discussion_r93115225) [](ancestors = 93115225,92931361)

Frameworks/CoreGraphics/CGDataProvider.mm
aballway: I think ternary statements are plenty reasonable, and shouldn't pose a challenge for any merges

---
In reply to: [93115443](https://github.com/Microsoft/WinObjC/pull/1579#discussion_r93115443) [](ancestors = 93115443,92931590)

Frameworks/CoreGraphics/CGDataProvider.mm
msft-Jeyaram: there is a significant gain, which is reducing errors that can be accident;y introduced by a future author.
Also this is a standard no no situation within C++

---
In reply to: [93115701](https://github.com/Microsoft/WinObjC/pull/1579#discussion_r93115701) [](ancestors = 93115701,93115225,92931361)

Frameworks/CoreGraphics/CGDataProvider.mm
msft-Jeyaram: it's reasonable but not ideal. we seen issues with merges where ternary operations for conditional checks end up being more work than it needs to be.
usually these conditions do not have tests and regressions come back.

---
In reply to: [93115787](https://github.com/Microsoft/WinObjC/pull/1579#discussion_r93115787) [](ancestors = 93115787,93115443,92931590)

msft-Jeyaram: let's do the right thing :)

---
In reply to: [93116128](https://github.com/Microsoft/WinObjC/pull/1579#discussion_r93116128) [](ancestors = 93116128,93115787,93115443,92931590)

Frameworks/CoreGraphics/CGFont.mm
msft-Jeyaram: uh this what happens when we push off things to be fixed later :(

---
In reply to: [93115850](https://github.com/Microsoft/WinObjC/pull/1579#discussion_r93115850) [](ancestors = 93115850,93115537,92931648)

PR #1580, opened 2016-12-16
Store CGContext's text position in its text matrix.

DHowett-MSFT:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1580)


Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1112,16 +1112,19 @@ CGAffineTransform CGContextGetTextMatrix(CGContextRef context) {
 */
 void CGContextSetTextPosition(CGContextRef context, CGFloat x, CGFloat y) {
     NOISY_RETURN_IF_NULL(context);
-    UNIMPLEMENTED();
+    context->textMatrix.tx = x;
```
msft-Jeyaram: nit: can we use a inline function to take both of them in and set.

context->setTextMatrix(x,y);

msft-Jeyaram: hmmm given getting it back separately, maybe not

PR #1581, opened 2016-12-17
Fix last one for too small textfield in displayMode table in Woccatalog

yiyang-msft:
The height for most of other textField has already been fixed by change 88975cab8f6346ca3b318e2847b2ba6dae4aa767. 

This is only one left behind.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1581)


bbowman:
:shipit:

PR #1582, opened 2016-12-17
Implementation of planned AudioConverter Services APIs from AudioToolbox

vkvenkat:
Implementation of the following three AudioConverter Services APIs from AudioToolbox Framework and the corresponding WOCCatalog sample.

AudioConverterDispose
AudioConverterNew
AudioConverterConvertBuffer

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1582)

### DIFF
--- a/Frameworks/AudioToolbox/AudioConverter.mm
+++ b/Frameworks/AudioToolbox/AudioConverter.mm
@@ -1,5 +1,6 @@
 //******************************************************************************
 //
+// Copyright (c) 2016 Intel Corporation. All rights reserved.
 // Copyright (c) 2016 Microsoft Corporation. All rights reserved.
 //
 // This code is licensed under the MIT License (MIT).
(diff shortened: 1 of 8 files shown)


Frameworks/AudioToolbox/AudioConverter.mm
```diff
@@ -16,28 +17,186 @@
...
+#include <mfidl.h>
+#include <mferror.h>
+#include <wmcodecdsp.h>
+#include <COMIncludes_End.h>
+
+using namespace Microsoft::WRL;
+using namespace Microsoft::WRL::Wrappers;
+
+struct OpaqueAudioConverter {
+    IMFTransform* transform;
```
bbowman: >IMFTransform [](start = 4, length = 12)

comptr?

Frameworks/AudioToolbox/AudioConverter.mm
```diff
@@ -16,28 +17,186 @@
...
+    HRESULT status;
+
+    status = MFShutdown();
+    if (!SUCCEEDED(status)) {
+        NSTraceInfo(TAG, @"MFShutdown Failed");
+        return kAudioConverterErr_UnspecifiedError;
+    }
+
+    CoUninitialize();
+    delete inAudioConverter;
```
bbowman: >inAudioConverter [](start = 11, length = 16)

Calling delete on passed in memory is not a scalable memory management strategy. This creates CRT / heap dependencies between the allocation call and this call. Take for instance if someone made a derived AudioCOnverterRef etc. Given that the type is called Foo*Ref* there is a certain expectation about its memory management as well that a straight delete is likely violating. Consider instead calling release on this object and making it a real CF or ObjC object under the convers.

Frameworks/AudioToolbox/AudioConverter.mm
```diff
@@ -16,28 +17,186 @@
...
  @Notes
 */
 OSStatus AudioConverterDispose(AudioConverterRef inAudioConverter) {
-    UNIMPLEMENTED();
-    return StubReturn();
+    HRESULT status;
+
+    status = MFShutdown();
+    if (!SUCCEEDED(status)) {
+        NSTraceInfo(TAG, @"MFShutdown Failed");
```
bbowman: We nice macros for this sort of thing.

Frameworks/AudioToolbox/AudioConverter.mm
```diff
@@ -16,28 +17,186 @@
...
+    }
+
+    CoUninitialize();
+    delete inAudioConverter;
+
+    return noErr;
+}
+
+void _setMFProperties(const AudioStreamBasicDescription* format, IMFMediaType** mediaType) {
+    IMFMediaType* tempMediaType = nullptr;
```
bbowman: >IMFMediaType [](start = 4, length = 12)

ComPtr

Frameworks/AudioToolbox/AudioConverter.mm
```diff
@@ -16,28 +17,186 @@
...
+    }
+
+    if (inSourceFormat->mFramesPerPacket != 1 || inSourceFormat->mFramesPerPacket != inDestinationFormat->mFramesPerPacket) {
+        UNIMPLEMENTED();
+        return StubReturn();
+    }
+
+    HRESULT status;
+
+    status = CoInitializeEx(nullptr, COINIT_APARTMENTTHREADED | COINIT_DISABLE_OLE1DDE);
```
bbowman: >CoInitializeEx [](start = 13, length = 14)

is this needed? I think we CoInit on our threads already and I'm not very comfortable with the flag you have here. Can you explain the logic behind this?

Frameworks/AudioToolbox/AudioConverter.mm
```diff
@@ -16,28 +17,186 @@
...
+        return kAudioConverterErr_UnspecifiedError;
+    }
+
+    status = MFStartup(MF_VERSION, MFSTARTUP_NOSOCKET);
+    if (!SUCCEEDED(status)) {
+        NSTraceInfo(TAG, @"Media Foundation Initialization Failed");
+        return kAudioConverterErr_UnspecifiedError;
+    }
+
+    IUnknown* spTransformUnk;
```
bbowman: >IUnknown* spTransformUnk; [](start = 4, length = 25)

ComPtr

Frameworks/AudioToolbox/AudioConverter.mm
```diff
@@ -16,28 +17,186 @@
...
+
+    status = CoCreateInstance(CLSID_AudioResamplerMediaObject, nullptr, CLSCTX_INPROC_SERVER, IID_IUnknown, (void**)&spTransformUnk);
+    if (!SUCCEEDED(status)) {
+        NSTraceInfo(TAG, @"Audio Resampler Initialization Failed");
+        spTransformUnk->Release();
```
bbowman: >spTransformUnk->Release(); [](start = 8, length = 26)

Is this a safe call to be making?  What is the failure means it wasn't assigned a value??

Frameworks/AudioToolbox/AudioConverter.mm
```diff
@@ -16,28 +17,186 @@
...
+        return kAudioConverterErr_UnspecifiedError;
+    }
+
+    AudioConverterRef outConverter = new OpaqueAudioConverter;
+    outConverter->transform = nullptr;
```
bbowman: this seems awfully reminiscent of a constructor that would encapsulate what you want better. ALong with other comments about the lifetime and ownership of this object, please consider an alternate approach.

Frameworks/AudioToolbox/AudioConverter.mm
```diff
@@ -16,28 +17,186 @@
...
+
+    status = spTransformUnk->QueryInterface(IID_PPV_ARGS(&outConverter->transform));
+    if (!SUCCEEDED(status)) {
+        NSTraceInfo(TAG, @"Transform Initialization Failed");
+        delete outConverter;
```
bbowman: who controls the lifetime of transform then??

samples/WOCCatalog/WOCCatalog/AudioToolboxViewController.mm
```diff
@@ -64,31 +64,65 @@ - (void)viewDidLoad {
     [callbackFunction setBackgroundColor:[UIColor whiteColor]];
     [callbackFunction addTarget:self action:@selector(callbackFunctionChanged:) forControlEvents:UIControlEventValueChanged];
     [scrollView addSubview:callbackFunction];
-    
+
+    // Test Run Code For Audio Converter
```
bbowman: Unit tests?

bbowman:
:clock1:

Frameworks/AudioToolbox/AudioConverter.mm
vkvenkat: Fixed

Frameworks/AudioToolbox/AudioConverter.mm
```diff
@@ -16,28 +17,186 @@
...
-    return StubReturn();
+    HRESULT status;
+
+    status = MFShutdown();
+    if (!SUCCEEDED(status)) {
+        NSTraceInfo(TAG, @"MFShutdown Failed");
+        return kAudioConverterErr_UnspecifiedError;
+    }
+
+    CoUninitialize();
```
vkvenkat: Removed

Frameworks/AudioToolbox/AudioConverter.mm
vkvenkat: Now defined AudioConverter as an ObjC interface (typedef'ed as a struct in header). ARC is enabled on this repo. Now calling CFBridgingRelease here to transfer ownership back to ARC after prior CFBridgingRetain. Also, ComPtrs used elsewhere have self-cleanup.

Frameworks/AudioToolbox/AudioConverter.mm
vkvenkat: Now using the RETURN_AUDIOERR_IF_FAILED_WITH_MSG macro

Frameworks/AudioToolbox/AudioConverter.mm
vkvenkat: Fixed

Frameworks/AudioToolbox/AudioConverter.mm
```diff
@@ -16,28 +17,186 @@
...
+}
+
+void _setMFProperties(const AudioStreamBasicDescription* format, IMFMediaType** mediaType) {
+    IMFMediaType* tempMediaType = nullptr;
+    MFCreateMediaType(&tempMediaType);
+
+    tempMediaType->SetGUID(MF_MT_MAJOR_TYPE, MFMediaType_Audio);
+
+    if (format->mFormatFlags & kAudioFormatFlagIsFloat) {
+        tempMediaType->SetGUID(MF_MT_SUBTYPE, MFAudioFormat_Float);
```
vkvenkat: Now encapsulated in status-check macro

Frameworks/AudioToolbox/AudioConverter.mm
vkvenkat: Removed

Frameworks/AudioToolbox/AudioConverter.mm
vkvenkat: Fixed

Frameworks/AudioToolbox/AudioConverter.mm
vkvenkat: Is now a ComPtr with self-cleanup

Frameworks/AudioToolbox/AudioConverter.mm
vkvenkat: Replaced with new ObjC interface-based AudioConverter

Frameworks/AudioToolbox/AudioConverter.mm
vkvenkat: Removed from here. Also, transform is ComPtr. Now transferring ownership of AudioConverter to ARC in AudioConverterDispose.

samples/WOCCatalog/WOCCatalog/AudioToolboxViewController.mm
afaruqui: Moved all of the tests from here to AudioToolbox unit test.

afaruqui:
CR changes complete.

tests/UnitTests/AudioToolbox/AudioConverterTest.mm
```diff
@@ -0,0 +1,148 @@
...
+#import <Foundation/Foundation.h>
+#import <AudioToolbox/AudioToolbox.h>
+#import <StubReturn.h>
+#import <NSBundleInternal.h>
+#import <NSLogging.h>
+
+// This is needed for getting mfapi.h to build
+#ifndef OUT
+#define OUT
+#endif
```
bbowman: Should we have a common header (possibly internal) that does this and the com includes for the mf stuff?

Frameworks/AudioToolbox/AudioConverter.mm
```diff
@@ -16,28 +17,160 @@
...
+        return kAudioConverterErr_UnspecifiedError; \
+    }
+
+@implementation AudioConverter
+- (instancetype)initWithSourceFormat:(const AudioStreamBasicDescription*)srcFormat
+                   destinationFormat:(const AudioStreamBasicDescription*)destFormat {
+    if (self = [super init]) {
+        ComPtr<IUnknown> spTransformUnk;
+
+        RETURN_NULL_IF_FAILED(
```
bbowman: >RETURN_NULL_IF_FAILED [](start = 8, length = 21)

I like the use of macro here but I'm a little concerned. Should you do a [self release] here? Alloc init ends up at a +1 ref and you are returning null which would mean that the instance is leaked right?

Frameworks/AudioToolbox/AudioConverter.mm
```diff
@@ -16,28 +17,160 @@
...
 */
 OSStatus AudioConverterNew(const AudioStreamBasicDescription* inSourceFormat,
                            const AudioStreamBasicDescription* inDestinationFormat,
-                           DWORD* outAudioConverter) {
-    // TODO: outAudioConverter ought to be made AudioConverterRef _Nullable * when this is unstubbed
-    UNIMPLEMENTED();
-    if (outAudioConverter) {
-        *outAudioConverter = 1234;
+                           AudioConverterRef _Nullable* outAudioConverter) {
+    if (inSourceFormat->mFormatID != kAudioFormatLinearPCM || inDestinationFormat->mFormatID != kAudioFormatLinearPCM) {
```
bbowman: >(inSourceFormat-> [](start = 7, length = 17)

can inSourceFormat be null / do we care?

Frameworks/AudioToolbox/AudioConverter.mm
```diff
@@ -16,28 +17,160 @@
...
+        return StubReturn();
+    }
+
+    RETURN_AUDIOERR_IF_FAILED_WITH_MSG(MFStartup(MF_VERSION, MFSTARTUP_NOSOCKET), @"MFStartup Failed");
+
+    AudioConverter* outConverter = [[AudioConverter alloc] initWithSourceFormat:inSourceFormat destinationFormat:inDestinationFormat];
+
+    ComPtr<IMFTransform> mediaTransform = [outConverter getTransform];
+
+    ComPtr<IMFMediaType> sourceMediaType = nullptr;
```
bbowman: > = nullptr; [](start = 40, length = 11)

nit: = nullptr not really needed for ComPtr<> init

Frameworks/AudioToolbox/AudioConverter.mm
```diff
@@ -99,13 +232,104 @@ OSStatus AudioConverterSetProperty(AudioConverterRef inAudioConverter,
...
+    ComPtr<IMFTransform> mediaTransform = [(__bridge AudioConverter*)inAudioConverter getTransform];
+
+    RETURN_AUDIOERR_IF_FAILED_WITH_MSG(mediaTransform->ProcessInput(0, sampleIn.Get(), 0), @"Input processing failed");
+
+    ComPtr<IMFSample> sampleOut = NULL;
+    // Create the output sample
+    RETURN_AUDIOERR_IF_FAILED_WITH_MSG(MFCreateSample(&sampleOut), @"Output sample creation failed");
+
+    ComPtr<IMFMediaBuffer> bufferOut = NULL;
+    // create a buffer for the output sample
```
bbowman: super nit: I'd put the comment above the ComPtr decl and try to be consistent with capitalization.

tests/UnitTests/AudioToolbox/AudioConverterTest.mm
afaruqui: Moved all of the common stuff to AudioConverterInternal.h

Frameworks/AudioToolbox/AudioConverter.mm
afaruqui: I believe ARC should handle it as we are returning NULL. Also went ahead and explicitly declared self = nil for the failure case

Frameworks/AudioToolbox/AudioConverter.mm
afaruqui: Added checks for it

Frameworks/AudioToolbox/AudioConverter.mm
afaruqui: Removed

Frameworks/AudioToolbox/AudioConverter.mm
afaruqui: Moved them above the statements

bbowman:
@vkvenkat @afaruqui 

I can't merge this PR because the CI build failed with: 

`2017-01-12T06:43:46.6538180Z          LINK : fatal error LNK1104: cannot open file 'AudioToolbox.lib' [D:\Build\Agent1\_work\2\s\build\Tests\UnitTests\AudioToolbox\AudioToolbox.UnitTests.vcxproj]`

Which usually means that the project producing the import lib failed to build, or a build order dependency meant that the unit tests were built before the implementation was. 

As a general rule, the UTs link against the static lib not the dll (in this case that would be AudioToolbox**Lib**.lib ) so I'd double check the .vcxproj file for the project references, /msvc/ for any sbclang.props/.targets link goo and try a clean build locally to see what is going on.

build/Tests/UnitTests/AudioToolbox/AudioToolbox.UnitTests.vcxproj
```diff
@@ -135,7 +135,7 @@
     <Link>
       <SubSystem>Console</SubSystem>
       <GenerateDebugInformation>true</GenerateDebugInformation>
-      <AdditionalDependencies>;libdispatch.lib;mincore.lib;%(AdditionalDependencies)</AdditionalDependencies>
+      <AdditionalDependencies>AudioToolbox.lib;mfuuid.lib;mfplat.lib;libdispatch.lib;mincore.lib;%(AdditionalDependencies)</AdditionalDependencies>
```
bbowman: >AudioToolbox.lib [](start = 30, length = 16)

Looks like this is the culprit. Please remove. The project reference above for the static lib should be sufficient. If it is not, please comment with the build error you get and lets investigate.

afaruqui: I wasnt getting any build errors and confirmed that again now with a new clone and build. 
Without the dependency on AudioToolbox.lib (or with dependency on AudioToolboxLib.lib) I am getting an unresolved external symbol error for the AudioConverterNew and AudioConvertDispose calls

afaruqui:
@winobjc-bot please test Debug

afaruqui:
@winobjc-bot please test

build/Tests/UnitTests/AudioToolbox/AudioToolbox.UnitTests.vcxproj
bbowman: @afaruqui I cannot in good faith just re-test the current build as I see the issue and agree with the build error. Any successful builds are likely due to chance and the binary in question just happening to be built in time but without a strong ordering enforced by the build machine. If I were to accept as-is, even with testing again, I would be knowingly making our build results unstable going forward.

All of that being said, it should be pretty easy to make the correct fix here. You mention unresolved external symbols for  AudioConverterNew  and AudioConverterDispose. Those symbols are found in **AudioConverter.mm** and declared in **AudioConverter.h**.  You can see that AudioToolboxLib.vcxproj is responsible for building this .mm file which means that the project reference should be sufficient to pull in the lib that has the implementations in question. 

This then begs the question why is the linker complaining about unresovled **external** symbols if they are in a lib being built with the unit tests. Well the key thing there is that they shouldn't be external. They are located simply in another TU in that module. It appears that changing the preprocessorDefines as mentioned in a comment I"m about to make should do the trick.

bbowman:
>       <PreprocessorDefinitions>NO_STUBS;_CRT_SECURE_NO_WARNINGS;%(PreprocessorDefinitions)</PreprocessorDefinitions>

All of the PreprocessorDefinitions lines should read:

<PreprocessorDefinitions>NO_STUBS;_CRT_SECURE_NO_WARNINGS;**AUDIOTOOLBOX_IMPEXP;**%(PreprocessorDefinitions)</PreprocessorDefinitions>

---
Refers to: build/Tests/UnitTests/AudioToolbox/AudioToolbox.UnitTests.vcxproj:192 in a36d92c. [](commit_id = a36d92ca61b61fcfae31f3de43117f51413bcd24, deletion_comment = False)

afaruqui:
Added "-DAUDIOTOOLBOX_IMPEXP= " under additional options.

bbowman:
@afaruqui Kicking off new builds. Thanks for the changes!

PR #1601, opened 2016-12-21
Fix crash dismissing popover via dismissViewControllerAnimated: plus fixes/cleanups for presentViewController/dismissViewController

ehren:
Prior to #1464 the popover presentation controller leaked which was
fixed by making the content view controller hold an explicit owning
reference to the popover presentation controller via a StrongId ivar.
However, dismissViewControllerAnimated: tore down references to the
popover presentation controller _prior_ to dismiss of the popover,
resulting in too early deallocation (fixed by this commit).

---------

Note: I'm going to rebase #1511 on top of this pull. This only handles the crash on press of the Done button in WOCCatalog. The only change to the non-popover code path is that we're killing the _parent/_presenting/_popover controller refs after setting the view hidden (instead of before). Even in the non-popover case this is too early - we should be killing these refs only after the dismissal is complete (in the current implementation, e.g. clients will notice that ``parentViewController`` is incorrectly nil during viewWillDisappear) - these issues will be fixed by a rebased commit in #1511.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1601)


yiyang-msft:
>     [curController autorelease];

can we use strong<Id> here? 

---
Refers to: Frameworks/UIKit/UIViewController.mm:1237 in 2ca7f9d. [](commit_id = 2ca7f9d2d106fd2f719bbb31f58d4186f4dfffe0, deletion_comment = False)

Frameworks/UIKit/UIViewController.mm
```diff
@@ -1250,18 +1256,29 @@ - (void)dismissViewControllerAnimated:(BOOL)animated completion:(void (^)(void))
...
-    curController->priv->_parentViewController = nil;
-    curController->priv->_presentingViewController = nil;
-    curController->priv->_popoverPresentationController = nil;
-
     UIView* curView = [curController view];
 
     UIView* myView = [self view];
     [myView setHidden:FALSE];
 
     if (realPopoverPresented) {
```
yiyang-msft: >realPopoverPresented [](start = 8, length = 20)

would be great if we have some comments here to clarify why we have different code path for dismiss

Frameworks/UIKit/UIViewController.mm
```diff
@@ -1250,18 +1256,29 @@ - (void)dismissViewControllerAnimated:(BOOL)animated completion:(void (^)(void))
...
+
+            if (completion) {
+                completion();
+            }
+        }];
+
+        return;
+    }
+
+    [curController _onDismissCleanup];
```
yiyang-msft: >[curController _onDismissCleanup]; [](start = 4, length = 34)

should this be moved to end of the function? because we still see curController being used below.

yiyang-msft:
>         if ([self parentViewController] != nil) {

we could combine, but need comments here why we early return. 

---
Refers to: Frameworks/UIKit/UIViewController.mm:1225 in 2ca7f9d. [](commit_id = 2ca7f9d2d106fd2f719bbb31f58d4186f4dfffe0, deletion_comment = False)

yiyang-msft:
>         [curController dismissViewControllerAnimated:animated completion:completion];

do we need need earlier return?

---
Refers to: Frameworks/UIKit/UIViewController.mm:1246 in 2ca7f9d. [](commit_id = 2ca7f9d2d106fd2f719bbb31f58d4186f4dfffe0, deletion_comment = False)

ehren:
Re the comment (can't reply directly via github interface...) 

```
        [curController dismissViewControllerAnimated:animated completion:completion];
```

>> do we need need earlier return? 

This is the existing logic for dismissing multiple view controllers. An early return would improve the situation somewhat but a more extensive fix is needed (one has to walk the parent hierarchy dismissing view controllers - this is handled by the more extensive changes in #1511).

Frameworks/UIKit/UIViewController.mm
ehren: Yes it should. This is handled by #1511 but I'll add the proper late cleanup of curController to this pull (in a coming commit).

ehren:
```
        if ([self parentViewController] != nil) {
```
>>> we could combine, but need comments here why we early return.

I'll add the comments from #1511 regarding this (this is the existing code - which works - handling `dismissViewController:` when it's called on the modal instead of the parent).

Frameworks/UIKit/UIViewController.mm
```diff
@@ -1250,18 +1256,29 @@ - (void)dismissViewControllerAnimated:(BOOL)animated completion:(void (^)(void))
         curController->priv->_parentViewController->priv->_modalViewController = nil;
     }
```
ehren: I have some changes in #1511 making the _presentedViewController a StrongId and some related changes that better handle the cleanup of the modal you've pointed to. I can put those changes into this commit as well.

ehren:
Hi Yi, To address those comments I did have to pull in more changes from the other pull. With both commits here this is essentially the same change as #1511 except I've avoided making the changes (with corresponding cleanups) to always present a popover even in non-tablet mode. In particular, to address the comment around the bad early teardown (and autorelease dance) of the presented controller (formerly called 'curController'), I've gone ahead with the StrongId change (and also removal of the unnecessary alias ivar _modalViewController). 

Also, the change adding proper multiple child dismiss logic arguably should be in a separate pull - however as noted above the current implementation of multiple child dismiss is so broken that I think it makes sense to go forward with the changes in this pull.

ehren:
I've forced pushed a change making the subsequent fixup commit more simple (we can address more robust handling of multiple child dismiss in a subsequent pull). 

The current followup commit 477ac1f addresses these comments:

**Comment 1**

```
> @@ -1250,18 +1256,29 @@ - (void)dismissViewControllerAnimated:(BOOL)animated completion:(void (^)(void))
         curController->priv->_parentViewController->priv->_modalViewController = nil;
     }

it is strange these clean up have to be done here.
```

all parent/child cleanup has been refactored into a common method (and it's called at the appropriate time - just before calling the dismiss completion block).

**Comment 2**

```
        [curController dismissViewControllerAnimated:animated completion:completion];

do we need need earlier return?
``` 

I ended up doing some testing last night regarding multiple child dismissal (using my WOCCatalog testcase). 

The current recursive implementation doesn't happen to crash (with my WOCCatalog testcase) - but it is unsafe to call dismissViewControllerAnimated:YES on a parent before the animation of a dismissed child has completed. I've just added a comment to clarify that this needs work but have avoided more extensive modifications (and I've left the multi child dismiss example in WOCCatalog out of this pull)

** Comment 3 **

```
    [curController autorelease];

can we use strong here?
```

We could just change curController to a StrongId there. However, the better change (which also addresses the comment "it is strange these clean up have to be done here") is to make the _presentedViewController ivar a StrongId which I have done

 ---------

The fixup commit also includes various comment improvements in general.

Note:

Sorry this is down to the wire! 

It's still safe to take just the first patch  2ca7f9d which only fixes the crash on popover Done button press in WOCCatalog. However, I would opt to also take the subsequent patch 477ac1f as well (it is a definite improvement to handle the child/parent cleanup just before calling the user completion block). I've manually tested to ensure both patches do not introduce new leaks or crashes.

Frameworks/UIKit/UIViewController.mm
```diff
@@ -1139,9 +1144,7 @@ - (void)presentViewController:(UIViewController*)controller animated:(BOOL)anima
         return;
     }
 
-    UIViewController* oldViewController = self;
-    if (priv->_modalViewController != nil) {
-        oldViewController = priv->_modalViewController;
+    if ([self presentedViewController]) {
```
jaredhms: >[self presentedViewController] [](start = 8, length = 30)

nit: priv->_presentedViewController would avoid a message send

jaredhms: although this file does a lot of both, so feel free to disregard.

---
In reply to: [95265157](https://github.com/Microsoft/WinObjC/pull/1601#discussion_r95265157) [](ancestors = 95265157)

Frameworks/UIKit/UIViewController.mm
```diff
@@ -1211,57 +1211,71 @@ - (void)_dismissTransitionStopped:(id)anim finished:(BOOL)finished {
...
+        // will fail on the stale presentingViewController/presentedViewController relationship.
+        if (completion) {
+            completion();
+        }
+    };
 
     // We maintain a popoverPresentationController instance whenever UIModalPresentationPopover is specified.
     // However, in non-tablet operation mode, UIModalPresentationPopover should result in the presentation of
     // a full screen modal instead. We check here if an actual popover has been presented to handle the
     // dismiss (of a genuine popover or alternately a full screen modal) appropriately.
```
jaredhms: is now a good time to delete the non-tablet version of this codepath?  do any of your apps rely on it?

Frameworks/UIKit/UIViewController.mm
```diff
@@ -1211,57 +1211,71 @@ - (void)_dismissTransitionStopped:(id)anim finished:(BOOL)finished {
...
-    BOOL realPopoverPresented = [[curController popoverPresentationController] _isManagingPresentation];
+    UIPopoverPresentationController* popover = [presented popoverPresentationController];
+    BOOL realPopoverPresented = [popover _isManagingPresentation];
 
-    if (!realPopoverPresented && curController->priv->_modalViewController) {
-        [curController dismissViewControllerAnimated:animated completion:completion];
+    if (realPopoverPresented) {
```
jaredhms: does this mean it's tablet mode? or is this the non-tablet path?

Frameworks/UIKit/UIViewController.mm
```diff
@@ -1211,57 +1211,71 @@ - (void)_dismissTransitionStopped:(id)anim finished:(BOOL)finished {
...
         return;
     }
 
-    UIViewController* curController = priv->_modalViewController;
+    UIViewController* presented = [self presentedViewController];
+
+    if ([presented presentedViewController]) {
+        // Recursively handle multiple child dismissal.
+        // TODO: This is unsafe because we end up dispatching the dimiss animation of a
+        // parent before the child dispatch animation completes.
```
jaredhms: do you have a proposed fix?  would it work to move this into the cleanupCompletion, after the inner completion is called?

Frameworks/UIKit/UIViewController.mm
```diff
@@ -1211,57 +1211,71 @@ - (void)_dismissTransitionStopped:(id)anim finished:(BOOL)finished {
...
-        curController->priv->_parentViewController->priv->_presentedViewController = nil;
-        curController->priv->_parentViewController->priv->_modalViewController = nil;
-    }
-
-    curController->priv->_parentViewController = nil;
-    curController->priv->_presentingViewController = nil;
-    curController->priv->_popoverPresentationController = nil;
-
-    UIView* curView = [curController view];
+    UIView* curView = [presented view];
```
jaredhms: >curView [](start = 12, length = 7)

nit: can we call this presentedView for clarity?

Frameworks/UIKit/UIViewController.mm
ehren: I think it makes sense to handle removing the tablet vs nontablet logic now since it affects presentViewController. The pull in #1511 includes this but I think would be better to bring in the additional changes as a third commit in this pull for clarity, then close #1511.

Frameworks/UIKit/UIViewController.mm
ehren: I've got some changes for handling the dismissal in the required order that I can handle in a third commit to this pull.

Frameworks/UIKit/UIViewController.mm
ehren: will add this

Frameworks/UIKit/UIViewController.mm
ehren: I was thinking for cases where the property just returns the ivar (at least in UIViewController) it's better to access via the property for the benefit of subclassing (not that subclassing presentedViewController is likely to work or be useful).

Frameworks/UIKit/UIViewController.mm
jaredhms: ok, sounds good.

---
In reply to: [95274680](https://github.com/Microsoft/WinObjC/pull/1601#discussion_r95274680) [](ancestors = 95274680)

ehren:
One note re multiple child dismissal: the latest commit uses a do while for walking the inheritance chain. This could also be done (safely) with recursion in fewer lines (see similar changes here: https://github.com/ehren/WinObjC/blob/8f02a5d00fb02dc829ea70d5215a878562cfed6f/Frameworks/UIKit/UIViewController.mm#L1225 against an older WinObjC) however I opted to avoid unnecessary use of non-tail recursion.

Frameworks/UIKit/UIViewController.mm
```diff
@@ -491,9 +491,15 @@ - (CGRect)_modalPresentationFormSheetFrame {
 
 - (BOOL)_hidesParent {
     UIModalPresentationStyle style = [self modalPresentationStyle];
-    if ((style == UIModalPresentationFormSheet || style == UIModalPresentationPopover) && DisplayProperties::IsTablet()) {
+
+    if (style == UIModalPresentationPopover) {
+        // Popovers (which don't hide their parent) are always shown regardless of tablet mode.
```
jaredhms: >regardless of tablet mode. [](start = 69, length = 26)

I thought you were removing support for non-tablet mode?

jaredhms:
I can hit this assert in ControlsViewController.m when trying to show a popup while a popup is already shown:

    assert(viewController.popoverPresentationController != nil);

Are these overzealous asserts?

Frameworks/UIKit/UIViewController.mm
ehren: Presenting a popover still requires setting presentationStyle = UIModalPresentationPopover. Before this would result in a full screen modal.

ehren:
That assert is due to the 'overlay doesn't block touches issue' ;) It's because you're trying to launch a new popover by pressing a button that's normally blocked by the popover overlay (resulting in WOCCatalog trying to present a popover on a view controller that's already presenting a popover).

Frameworks/UIKit/UIViewController.mm
ehren: To clarify, specifying presentationStyle = UIModalPresentationPopover under a non-tablet UIInterfaceIdiom resulted in presentation of a full screen modal previously.

jaredhms:
doh - of course! I'm hoping to have time to look at that tomorrow.

jaredhms:
Thanks for the fixes; in CI build...

Frameworks/UIKit/UIViewController.mm
```diff
@@ -1168,27 +1164,72 @@ - (void)presentViewController:(UIViewController*)controller animated:(BOOL)anima
...
+    if ([controller modalPresentationStyle] == UIModalPresentationPopover) {
+        controller->priv->_popoverPresentationController.attach([[UIPopoverPresentationController alloc] initWithPresentedViewController:controller presentingViewController:self]);
+
+        dispatch_block_t popoverPresent = ^{
+            [[UIApplication sharedApplication] endIgnoringInteractionEvents];
 
-        [controller performSelectorOnMainThread:@selector(_addToTop:) withObject:[NSNumber numberWithInt:animated] waitUntilDone:NO];
+            if (completion) {
+                completion();
+            }
```
jaredhms: Don't we need to copy this block so it's guaranteed to exist beyond the lifetime of - (void)presentViewController:(UIViewController*)controller animated:(BOOL)animated completion:(void (^)(void))completion ?

ehren: I _think_ even with non-ARC an implicit copy takes place as a result of the block capture (We don't have to worry about leaks because the capturing block is guaranteed to be nilled out.... so actually there are a few cases where I'm unnecessarily grabbing a strongSelf - just for safety's sake)

ehren: In particular, using a reference to weakSelf with the cleanupCompletion block on line 1323 is unnecessary (but doesn't hurt).

just fyi, have a small unrelated fixup push coming in next 5 mins

jaredhms: right - the weakSelf pattern is to avoid cycles - probably ok to leave as is.  i hadn't realized that block capture copied block variables - thanks for clarifiying.  i'll keep my eye out for your last change, but that means i'll have to kick off another CI build - not a problem though.

---
In reply to: [95633496](https://github.com/Microsoft/WinObjC/pull/1601#discussion_r95633496) [](ancestors = 95633496)

jaredhms:
:shipit:

ehren:
re that last fixup commit, I read some docs incorrectly wrt presentedViewController property.

On the one hand they say :

"When you present a view controller modally (either explicitly or implicitly) using the present(_:animated:completion:) method, the view controller that called the method has this property set to the view controller that it presented. If the current view controller did not present another view controller modally, the value in this property is nil. "

Lies. The right behaviour is the sentence above "The view controller that is presented by this view controller, or one of its ancestors in the view controller hierarchy."

Apparently this is the correct behaviour: http://stackoverflow.com/questions/31316238/what-does-presentedviewcontroller-in-uiviewcontroller-mean/31329884#31329884

(commit just undoes my changes to presentedViewController property and replaces my use of the property with direct use of ivar - I still think it makes sense that _presentedViewController ivar actually holds the modal)

jaredhms:
Thanks for the clarification.  Can you please heavily comment the code as such? 

---
In reply to: [271945385](https://github.com/Microsoft/WinObjC/pull/1601#issuecomment-271945385) [](ancestors = 271945385)

PR #1603, opened 2016-12-21
Improve perf for CTFramesetterSuggestFrameSizeWithConstraints

aballway:
CTFramesetterSuggestFrameSizeWithConstraints was going through the entire process of creating a frame to get the size, which was very wasteful.  Now only does the minimum amount of layout to determine the suggested size.  Cleans up NSString+UIKitAdditions for better perf analysis.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1603)

### DIFF
--- a/Frameworks/CoreText/CTFramesetter.mm
+++ b/Frameworks/CoreText/CTFramesetter.mm
@@ -123,23 +123,18 @@ CTTypesetterRef CTFramesetterGetTypesetter(CTFramesetterRef framesetter) {
 }
 
 /**
- @Status Interoperable
+ @Status Caveat
+ @Notes frameAttributes parameter ignored
  @Notes
 */
 CGSize CTFramesetterSuggestFrameSizeWithConstraints(
     CTFramesetterRef framesetter, CFRange stringRange, CFDictionaryRef frameAttributes, CGSize constraints, CFRange* fitRange) {
-    CGRect frameSize = CGRectZero;
-    frameSize.size = constraints;
-
-    _CTFrame* frame = __CreateFrame(static_cast<_CTFramesetter*>(framesetter), frameSize, stringRange);
-    CGSize ret = frame ? frame->_frameRect.size : CGSizeZero;
-
-    if (fitRange) {
-        *fitRange = CTFrameGetVisibleStringRange(static_cast<CTFrameRef>(frame));
-    }
-
-    [frame release];
-    return ret;
+    return framesetter ?
+               _DWriteGetSize((CFAttributedStringRef) static_cast<_CTFramesetter*>(framesetter)->_typesetter->_attributedString.get(),
+                              stringRange,
+                              constraints,
+                              fitRange) :
+               CGSizeZero;
 }
 
 /**
--- a/Frameworks/CoreText/CTFramesetter.mm
+++ b/Frameworks/CoreText/CTFramesetter.mm
@@ -129,12 +129,14 @@ CTTypesetterRef CTFramesetterGetTypesetter(CTFramesetterRef framesetter) {
 */
 CGSize CTFramesetterSuggestFrameSizeWithConstraints(
     CTFramesetterRef framesetter, CFRange stringRange, CFDictionaryRef frameAttributes, CGSize constraints, CFRange* fitRange) {
-    return framesetter ?
-               _DWriteGetSize((CFAttributedStringRef) static_cast<_CTFramesetter*>(framesetter)->_typesetter->_attributedString.get(),
-                              stringRange,
-                              constraints,
-                              fitRange) :
-               CGSizeZero;
+    if (framesetter == nil) {
+        return CGSizeZero;
+    }
+
+    CFAttributedStringRef string =
+        static_cast<CFAttributedStringRef>(static_cast<_CTFramesetter*>(framesetter)->_typesetter->_attributedString.get());
+
+    _DWriteGetSize(string, stringRange, constraints, fitRange);
 }
 
 /**
--- a/Frameworks/CoreText/CTFramesetter.mm
+++ b/Frameworks/CoreText/CTFramesetter.mm
@@ -136,7 +136,7 @@ CGSize CTFramesetterSuggestFrameSizeWithConstraints(
     CFAttributedStringRef string =
         static_cast<CFAttributedStringRef>(static_cast<_CTFramesetter*>(framesetter)->_typesetter->_attributedString.get());
 
-    _DWriteGetSize(string, stringRange, constraints, fitRange);
+    return _DWriteGetSize(string, stringRange, constraints, fitRange);
 }
 
 /**
--- a/Frameworks/CoreText/CTFramesetter.mm
+++ b/Frameworks/CoreText/CTFramesetter.mm
@@ -128,7 +128,7 @@ CGSize CTFramesetterSuggestFrameSizeWithConstraints(
     CFAttributedStringRef string =
         static_cast<CFAttributedStringRef>(static_cast<_CTFramesetter*>(framesetter)->_typesetter->_attributedString.get());
 
-    return _DWriteGetSize(string, stringRange, constraints, fitRange);
+    return _DWriteGetFrameSize(string, stringRange, constraints, fitRange);
 }
 
 /**
--- a/Frameworks/CoreText/CTFramesetter.mm
+++ b/Frameworks/CoreText/CTFramesetter.mm
@@ -21,13 +21,27 @@
 #import "CGPathInternal.h"
 #import "DWriteWrapper_CoreText.h"
 
-using namespace std;
-
 @implementation _CTFramesetter : NSObject
 @end
 
-static _CTFrame* __CreateFrame(_CTFramesetter* framesetter, CGRect frameRect, CFRange range) {
-    RETURN_NULL_IF(framesetter == nil);
+/**
+ @Status Interoperable
+*/
+CTFramesetterRef CTFramesetterCreateWithAttributedString(CFAttributedStringRef string) {
+    _CTFramesetter* ret = [_CTFramesetter alloc];
+    ret->_typesetter = static_cast<_CTTypesetter*>(CTTypesetterCreateWithAttributedString(string));
+    return static_cast<CTFramesetterRef>(ret);
+}
+
+/**
+ @Status Caveat
+ @Notes frameAttributes parameter ignored
+*/
+CTFrameRef CTFramesetterCreateFrame(CTFramesetterRef framesetterRef, CFRange range, CGPathRef path, CFDictionaryRef frameAttributes) {
+    RETURN_NULL_IF(framesetterRef == nil || path == nullptr);
+    CGRect frameRect;
+    _CGPathGetBoundingBoxInternal(path, &frameRect);
+    _CTFramesetter* framesetter = static_cast<_CTFramesetter*>(framesetterRef);
 
     // Call _DWriteWrapper to get _CTLine object list that makes up this frame
     _CTTypesetter* typesetter = static_cast<_CTTypesetter*>(framesetter->_typesetter);
@@ -36,10 +50,12 @@ static _CTFrame* __CreateFrame(_CTFramesetter* framesetter, CGRect frameRect, CF
     }
 
     StrongId<_CTFrame> ret = _DWriteGetFrame(static_cast<CFAttributedStringRef>(typesetter->_attributedString.get()), range, frameRect);
+    ret->_path.reset(CGPathRetain(path));
+    ret->_frameRect.origin = frameRect.origin;
 
     // Trying to access attributes without any text will throw an error
     if (range.length <= 0L) {
-        return ret.detach();
+        return static_cast<CTFrameRef>(ret.detach());
     }
 
     CTParagraphStyleRef settings =
@@ -48,7 +64,7 @@ static _CTFrame* __CreateFrame(_CTFramesetter* framesetter, CGRect frameRect, CF
                                                                    effectiveRange:nullptr]);
 
     if (settings == nullptr) {
-        return ret.detach();
+        return static_cast<CTFrameRef>(ret.detach());
     }
 
     // DWrite only gives manual control of lineheight when it is constant through a frame
@@ -87,31 +103,7 @@ static _CTFrame* __CreateFrame(_CTFramesetter* framesetter, CGRect frameRect, CF
         }
     }
 
-    return ret.detach();
-}
-
-/**
- @Status Interoperable
-*/
-CTFramesetterRef CTFramesetterCreateWithAttributedString(CFAttributedStringRef string) {
-    _CTFramesetter* ret = [_CTFramesetter alloc];
-    ret->_typesetter = static_cast<_CTTypesetter*>(CTTypesetterCreateWithAttributedString(string));
-    return static_cast<CTFramesetterRef>(ret);
-}
-
-/**
- @Status Caveat
- @Notes frameAttributes parameter ignored
-*/
-CTFrameRef CTFramesetterCreateFrame(CTFramesetterRef framesetter, CFRange stringRange, CGPathRef path, CFDictionaryRef frameAttributes) {
-    CGRect frameSize;
-    _CGPathGetBoundingBoxInternal(path, &frameSize);
-
-    _CTFrame* ret = __CreateFrame(static_cast<_CTFramesetter*>(framesetter), frameSize, stringRange);
-    ret->_path.reset(CGPathRetain(path));
-    ret->_frameRect.origin = frameSize.origin;
-
-    return static_cast<CTFrameRef>(ret);
+    return static_cast<CTFrameRef>(ret.detach());
 }
 
 /**
--- a/Frameworks/CoreText/DWriteWrapper_CoreText.h
+++ b/Frameworks/CoreText/DWriteWrapper_CoreText.h
@@ -51,7 +51,7 @@ struct _DWriteGlyphRunDetails {
 
 bool _CloneDWriteGlyphRun(_In_ DWRITE_GLYPH_RUN const* src, _Outptr_ DWRITE_GLYPH_RUN* dest);
 
-CGSize _DWriteGetSize(CFAttributedStringRef string, CFRange range, CGSize maxSize, CFRange* fitRange);
+CGSize _DWriteGetFrameSize(CFAttributedStringRef string, CFRange range, CGSize maxSize, CFRange* fitRange);
 _CTFrame* _DWriteGetFrame(CFAttributedStringRef string, CFRange range, CGRect frameSize);
 _CTLine* _DWriteGetLine(CFAttributedStringRef string);
 
--- a/Frameworks/CoreText/DWriteWrapper_CoreText.h
+++ b/Frameworks/CoreText/DWriteWrapper_CoreText.h
@@ -51,6 +51,7 @@ struct _DWriteGlyphRunDetails {
 
 bool _CloneDWriteGlyphRun(_In_ DWRITE_GLYPH_RUN const* src, _Outptr_ DWRITE_GLYPH_RUN* dest);
 
+CGSize _DWriteGetSize(CFAttributedStringRef string, CFRange range, CGSize maxSize, CFRange* fitRange);
 _CTFrame* _DWriteGetFrame(CFAttributedStringRef string, CFRange range, CGRect frameSize);
 _CTLine* _DWriteGetLine(CFAttributedStringRef string);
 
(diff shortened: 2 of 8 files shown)


Frameworks/CoreText/CTFramesetter.mm
```diff
@@ -123,23 +123,18 @@ CTTypesetterRef CTFramesetterGetTypesetter(CTFramesetterRef framesetter) {
...
-    _CTFrame* frame = __CreateFrame(static_cast<_CTFramesetter*>(framesetter), frameSize, stringRange);
-    CGSize ret = frame ? frame->_frameRect.size : CGSizeZero;
-
-    if (fitRange) {
-        *fitRange = CTFrameGetVisibleStringRange(static_cast<CTFrameRef>(frame));
-    }
-
-    [frame release];
-    return ret;
+    return framesetter ?
```
msft-Jeyaram: >ramesetter ? [](start = 12, length = 12)

LOL

Frameworks/CoreText/CTFramesetter.mm
```diff
@@ -123,23 +123,18 @@ CTTypesetterRef CTFramesetterGetTypesetter(CTFramesetterRef framesetter) {
...
+               _DWriteGetSize((CFAttributedStringRef) static_cast<_CTFramesetter*>(framesetter)->_typesetter->_attributedString.get(),
```
msft-Jeyaram: >ingRef) static_cast<_CTFra [](start = 46, length = 26)

mix of type casting conventions.

Frameworks/CoreText/DWriteWrapper_CoreText.mm
```diff
@@ -522,4 +523,56 @@ HRESULT STDMETHODCALLTYPE GetPixelsPerDip(_In_opt_ void* clientDrawingContext, _
     }
 
     return frame;
+}
+
+static CGSize _DWriteGetSize(CFAttributedStringRef string, CFRange range, CGSize maxSize, CFRange* fitRange) {
```
msft-Jeyaram: >CFRange [](start = 59, length = 7)

github id

Frameworks/CoreText/DWriteWrapper_CoreText.mm
```diff
@@ -522,4 +523,56 @@ HRESULT STDMETHODCALLTYPE GetPixelsPerDip(_In_opt_ void* clientDrawingContext, _
...
+    // TODO:: find more precise value than 1.0 to increase width by to fully enclose frame
+    ret.width = std::min(maxSize.width, textMetrics.widthIncludingTrailingWhitespace + 1.0f);
+    ret.height = std::min(maxSize.height, textMetrics.height);
+
+    if (fitRange) {
+        *fitRange = {};
+        uint32_t lineCount = 0;
+
+        // Ignore return value, we only want the lineCount for now
+        textLayout->GetLineMetrics(nullptr, 0, &lineCount);
```
msft-Jeyaram: > textLayout->GetLineMetrics(nullptr, 0, &lineCount); [](start = 7, length = 52)

is this guaranteed to return the correct line count if the function fails?

Frameworks/CoreText/DWriteWrapper_CoreText.mm
```diff
@@ -522,4 +523,56 @@ HRESULT STDMETHODCALLTYPE GetPixelsPerDip(_In_opt_ void* clientDrawingContext, _
...
+
+        std::vector<DWRITE_LINE_METRICS> metrics(lineCount);
```
msft-Jeyaram: >::vector<DWRITE_LINE_METRICS> metrics(lineCount); [](start = 11, length = 49)

since this is about performance, use array rather than a vector here (you will have a slight gain)

msft-Jeyaram: hmm I guess you might not know the size? nvm

---
In reply to: [93508769](https://github.com/Microsoft/WinObjC/pull/1603#discussion_r93508769) [](ancestors = 93508769)

Frameworks/UIKit/NSString+UIKitAdditions.mm
```diff
@@ -161,48 +88,78 @@ - (void)drawInRect:(CGRect)rect withAttributes:(NSDictionary*)attrs {
...
+- (CGSize)drawInRect:(CGRect)rect withFont:(UIFont*)font lineBreakMode:(UILineBreakMode)lineBreakMode {
+    return [self drawInRect:rect withFont:font lineBreakMode:lineBreakMode alignment:UITextAlignmentLeft];
 }
 
 /**
  @Status Interoperable
 */
-- (CGSize)drawAtPoint:(CGPoint)pt withFont:(UIFont*)font {
-    CGSize fontExtent;
+- (CGSize)drawInRect:(CGRect)rect withFont:(UIFont*)font lineBreakMode:(UILineBreakMode)lineBreakMode alignment:(UITextAlignment)alignment {
```
aballway: Drawing with nil font just uses default value.  Actually didn't mean to get rid of that in this change, but I think it makes sense to do.  @ms-jihua any thoughts?

Frameworks/UIKit/NSString+UIKitAdditions.mm
```diff
@@ -161,48 +88,78 @@ - (void)drawInRect:(CGRect)rect withAttributes:(NSDictionary*)attrs {
...
+    CGContextRef context = UIGraphicsGetCurrentContext();
+    if (rect.size.width == 0) {
+        rect.size.width = std::numeric_limits<CGFloat>::max();
+    }
```
msft-Jeyaram: is the intended behavior is to extend the region, specifically if one of the constraints are 0?

Frameworks/CoreText/DWriteWrapper_CoreText.mm
msft-Jeyaram: I saw that, but that is only if E_NOT_SUFFICIENT_BUFFER is returned as the HRESULT, it doesn't specify for other errors.

---
In reply to: [93509971](https://github.com/Microsoft/WinObjC/pull/1603#discussion_r93509971) [](ancestors = 93509971,93508458,93508293)

msft-Jeyaram: safer to check for that exact case (also good for unwanted behavior)
Given the implementation and documentation only mention it for the specific error above.
e.g someone decides to change the LINE_METRICS to and invalid state, this would help catch the error faster.

---
In reply to: [93510430](https://github.com/Microsoft/WinObjC/pull/1603#discussion_r93510430) [](ancestors = 93510430,93509971,93508458,93508293)

Frameworks/CoreText/CTFramesetter.mm
```diff
@@ -123,23 +123,20 @@ CTTypesetterRef CTFramesetterGetTypesetter(CTFramesetterRef framesetter) {
...
+    if (framesetter == nil) {
+        return CGSizeZero;
     }
 
-    [frame release];
-    return ret;
+    CFAttributedStringRef string =
+        static_cast<CFAttributedStringRef>(static_cast<_CTFramesetter*>(framesetter)->_typesetter->_attributedString.get());
+
+    _DWriteGetSize(string, stringRange, constraints, fitRange);
```
msft-Jeyaram: should be returning this value :)

Frameworks/UIKit/NSString+UIKitAdditions.mm
```diff
@@ -161,48 +88,78 @@ - (void)drawInRect:(CGRect)rect withAttributes:(NSDictionary*)attrs {
...
-    CGRect rct;
+    CTParagraphStyleSetting styles[2];
+    CTTextAlignment align = _NSTextAlignmentToCTTextAlignment(alignment);
+    styles[0] = { kCTParagraphStyleSpecifierAlignment, sizeof(CTTextAlignment), &align };
 
-    rct.origin.x = pt.x;
-    rct.origin.y = pt.y;
-    rct.size.width = 0;
-    rct.size.height = 0;
+    CTLineBreakMode breakMode = static_cast<CTLineBreakMode>(lineBreakMode);
```
msft-Jeyaram: >lineBreakMode [](start = 61, length = 13)

some of the below functions indicate this is not fully supported, update?

Frameworks/UIKit/NSString+UIKitAdditions.mm
```diff
@@ -161,48 +88,78 @@ - (void)drawInRect:(CGRect)rect withAttributes:(NSDictionary*)attrs {
...
+    styles[1] = { kCTParagraphStyleSpecifierLineBreakMode, sizeof(CTLineBreakMode), &breakMode };
+
+    NSAttributedString* attr =
+        [[[NSAttributedString alloc] initWithString:self
+                                         attributes:@{
+                                             (NSString*)kCTForegroundColorFromContextAttributeName : (id)kCFBooleanTrue,
```
msft-Jeyaram: >(NSString*) [](start = 45, length = 11)

switching to static cast is going to really obscure readability.
I guess we can leave it as be

Frameworks/UIKit/NSString+UIKitAdditions.mm
```diff
@@ -161,48 +88,78 @@ - (void)drawInRect:(CGRect)rect withAttributes:(NSDictionary*)attrs {
...
+                                             (NSString*)kCTParagraphStyleAttributeName : (id)CTParagraphStyleCreate(styles, 2),
```
msft-Jeyaram: hmmm given it's a compile time allocated array, can deduce the count from the array?

Frameworks/UIKit/NSString+UIKitAdditions.mm
```diff
@@ -161,48 +88,78 @@ - (void)drawInRect:(CGRect)rect withAttributes:(NSDictionary*)attrs {
...
+    woc::unique_cf<CGPathRef> path{ CGPathCreateWithRect(rect, nullptr) };
+    woc::unique_cf<CTFrameRef> frame{ CTFramesetterCreateFrame(framesetter.get(), {}, path.get(), nullptr) };
+
+    // Invert text matrix so glyphs are drawn with correct orientation
+    CGContextSetTextMatrix(context, CGAffineTransformMakeScale(1.0f, -1.0f));
+
+    // Can't draw the entire frame because it assumes our space has been flipped and translated
+    NSArray* lines = static_cast<NSArray*>(CTFrameGetLines(frame.get()));
+    std::vector<CGPoint> origins([lines count]);
+    CTFrameGetLineOrigins(frame.get(), {}, origins.data());
```
msft-Jeyaram: >  CTFrameGetLineOrigins(frame.get(), {}, origins.data()); [](start = 2, length = 57)

this is going to cause an nullptr access, if CTFramesetterCreateFrame fails.

Frameworks/UIKit/NSString+UIKitAdditions.mm
```diff
@@ -161,48 +88,78 @@ - (void)drawInRect:(CGRect)rect withAttributes:(NSDictionary*)attrs {
...
+                                         attributes:@{
+                                             (NSString*)kCTForegroundColorFromContextAttributeName : (id)kCFBooleanTrue,
+                                             (NSString*)kCTParagraphStyleAttributeName : (id)CTParagraphStyleCreate(styles, 2),
+                                             (NSString*)kCTFontAttributeName : font
+                                         }] autorelease];
+
+    woc::unique_cf<CTFramesetterRef> framesetter{ CTFramesetterCreateWithAttributedString(static_cast<CFAttributedStringRef>(attr)) };
+
+    woc::unique_cf<CGPathRef> path{ CGPathCreateWithRect(rect, nullptr) };
+    woc::unique_cf<CTFrameRef> frame{ CTFramesetterCreateFrame(framesetter.get(), {}, path.get(), nullptr) };
```
msft-Jeyaram: I would perfer you catch the failures of creation here rather than adding more functions calls, attempt to allocate a vector of size 0 and then returning a the size.

msft-Jeyaram: also this ends up propagating the errors and makes it harder to catch later on.

---
In reply to: [93514860](https://github.com/Microsoft/WinObjC/pull/1603#discussion_r93514860) [](ancestors = 93514860)

Frameworks/UIKit/NSString+UIKitAdditions.mm
msft-Jeyaram: how are these being verified?

Frameworks/UIKit/NSString+UIKitAdditions.mm
```diff
@@ -161,48 +88,78 @@ - (void)drawInRect:(CGRect)rect withAttributes:(NSDictionary*)attrs {
...
+        [[[NSAttributedString alloc] initWithString:self
+                                         attributes:@{
+                                             (NSString*)kCTForegroundColorFromContextAttributeName : (id)kCFBooleanTrue,
+                                             (NSString*)kCTParagraphStyleAttributeName : (id)CTParagraphStyleCreate(styles, 2),
+                                             (NSString*)kCTFontAttributeName : font
+                                         }] autorelease];
+
+    woc::unique_cf<CTFramesetterRef> framesetter{ CTFramesetterCreateWithAttributedString(static_cast<CFAttributedStringRef>(attr)) };
+
+    woc::unique_cf<CGPathRef> path{ CGPathCreateWithRect(rect, nullptr) };
```
msft-Jeyaram: >{ [](start = 34, length = 1)

RETURN_RESULT_IF_NULL(path,CGSizeZero);

Frameworks/UIKit/NSString+UIKitAdditions.mm
msft-Jeyaram: >frame [](start = 31, length = 5)

RETURN_RESULT_IF_NULL(frame,CGSizeZero);
:)

Frameworks/CoreText/DWriteWrapper_CoreText.mm
```diff
@@ -522,4 +523,56 @@ HRESULT STDMETHODCALLTYPE GetPixelsPerDip(_In_opt_ void* clientDrawingContext, _
...
+
+static CGSize _DWriteGetSize(CFAttributedStringRef string, CFRange range, CGSize maxSize, CFRange* fitRange) {
+    CGSize ret = CGSizeZero;
+    if (range.length == 0L) {
+        range.length = CFAttributedStringGetLength(string);
+    }
+
+    if (!string || range.length <= 0L) {
+        return ret;
+    }
```
aballway: wut?

ms-jihua:
> #endif

unrelated, but can you see if we can ditch these? UIKIT_EXPORT already has 'extern "C"' when __cplusplus is defined

---
Refers to: include/UIKit/UIGraphics.h:40 in c3bbc8f. [](commit_id = c3bbc8f3f3f29e775f13f584a2c3ebcd558d7a4b, deletion_comment = False)

Frameworks/CoreText/DWriteWrapper_CoreText.mm
```diff
@@ -522,4 +523,58 @@ HRESULT STDMETHODCALLTYPE GetPixelsPerDip(_In_opt_ void* clientDrawingContext, _
     }
 
     return frame;
+}
+
+static CGSize _DWriteGetSize(CFAttributedStringRef string, CFRange range, CGSize maxSize, CFRange* fitRange) {
+    CGSize ret = CGSizeZero;
+    if (range.length == 0L) {
+        range.length = CFAttributedStringGetLength(string);
```
ms-jihua: I'm guessing @msft-Jeyaram missed this part up here wrt the comment below? (thinking that probably at least one of these two deserves a callout in a comment)

ms-jihua:
any idea of the savings? rough measurements are fine

Frameworks/CoreText/DWriteWrapper_CoreText.mm
ms-jihua: I'm guessing @msft-Jeyaram missed this part up here wrt the comment below? (thinking that probably at least one of these two deserves a callout in a comment)

Frameworks/CoreText/DWriteWrapper_CoreText.h
```diff
@@ -51,6 +51,7 @@ struct _DWriteGlyphRunDetails {
 
 bool _CloneDWriteGlyphRun(_In_ DWRITE_GLYPH_RUN const* src, _Outptr_ DWRITE_GLYPH_RUN* dest);
 
+CGSize _DWriteGetSize(CFAttributedStringRef string, CFRange range, CGSize maxSize, CFRange* fitRange);
```
ms-jihua: >_DWriteGetSize [](start = 7, length = 14)

nit: would prefer more descriptive name

Frameworks/CoreText/DWriteWrapper_CoreText.mm
```diff
@@ -522,4 +523,58 @@ HRESULT STDMETHODCALLTYPE GetPixelsPerDip(_In_opt_ void* clientDrawingContext, _
...
+        }
+
+        std::vector<DWRITE_LINE_METRICS> metrics(lineCount);
+        if (FAILED(textLayout->GetLineMetrics(metrics.data(), lineCount, &lineCount))) {
+            return ret;
+        }
+
+        float totalHeight = 0;
+        CFIndex endPos = range.location;
+        for (size_t i = 0; i < lineCount; ++i) {
```
ms-jihua: nit: prefer enhanced for

Frameworks/UIKit/NSString+UIKitAdditions.mm
ms-jihua: uses default value as in, WinObjC currently uses a default value if a nil font is passed to DWrite, or the reference platform uses a default value? if the former, how?

my recommendation is to verify whether the reference platform uses a default font if nil is passed - if so, set font = [UIFont defaultFont] instead of putting nil in a dict

---
In reply to: [93509601](https://github.com/Microsoft/WinObjC/pull/1603#discussion_r93509601) [](ancestors = 93509601)

Frameworks/UIKit/NSString+UIKitAdditions.mm
```diff
@@ -161,48 +88,86 @@ - (void)drawInRect:(CGRect)rect withAttributes:(NSDictionary*)attrs {
...
 
 /**
- @Status Interoperable
+ @Status Caveat
+ @Notes Clipping line break modes unsupported
 */
-- (CGSize)drawAtPoint:(CGPoint)pt withFont:(UIFont*)font {
-    CGSize fontExtent;
+- (CGSize)drawInRect:(CGRect)rect withFont:(UIFont*)font lineBreakMode:(UILineBreakMode)lineBreakMode alignment:(UITextAlignment)alignment {
+    CGContextRef context = UIGraphicsGetCurrentContext();
```
ms-jihua: create this lower, this isn't needed for a while

Frameworks/UIKit/NSString+UIKitAdditions.mm
ms-jihua: +1

---
In reply to: [93514919](https://github.com/Microsoft/WinObjC/pull/1603#discussion_r93514919) [](ancestors = 93514919,93514860)

ms-jihua:
:clock1:

rajsesh-msft:
> #endif

Or include the new definition before the ending #ifdef to be consistent.

---
In reply to: [269369006](https://github.com/Microsoft/WinObjC/pull/1603#issuecomment-269369006) [](ancestors = 269369006)

---
Refers to: include/UIKit/UIGraphics.h:40 in c3bbc8f. [](commit_id = c3bbc8f3f3f29e775f13f584a2c3ebcd558d7a4b, deletion_comment = False)

aballway:
It cuts down frame measurement time a little more than half, saving about 25% of CPU time drawing in the XAML custom text editor.

---
In reply to: [269371016](https://github.com/Microsoft/WinObjC/pull/1603#issuecomment-269371016) [](ancestors = 269371016)

Frameworks/CoreText/DWriteWrapper_CoreText.mm
ms-jihua: extra nit: const auto&

---
In reply to: [93963412](https://github.com/Microsoft/WinObjC/pull/1603#discussion_r93963412) [](ancestors = 93963412)

Frameworks/UIKit/NSString+UIKitAdditions.mm
aballway: Just visually using test apps, we don't have any unit/drawing tests for UI kit

aballway:
Ping! Please update reviews

Issue #1561, opened 2016-12-16
Typing mid-string in a UIKit/XAML text field forces cursor to end of string


bebauer-msft:
_Edited by @DHowett-MSFT._

There may be a regression in UITextField/UITextView's caret handling. It appears that moving the caret mid-string and typing a character (or deleting a character) causes the caret to move to the end of the text field.

This is reproduceable in CTCatalog, CGCatalog, WOCCatalog, and any application with an unadorned text field.

rajsesh-msft:
#1577

DHowett-MSFT:
Reopening as per discussion with @rajsesh-msft; issue has been repurposed to track UI bug.

rajsesh-msft:
Thanks @DHowett-MSFT

edvv:
@bebauer-msft Please also see: https://github.com/Microsoft/WinObjC/issues/1609 using release 0.2.161130. You stated: "There may be a regression in []UITextView's caret handling". If editing of a UITextView instance is in the works then that would be very timely.

yiyang-msft:
fixed.

PR #1604, opened 2016-12-21
Fix [GitHub Issue #1561] Typing mid-string in a UIKit/XAML text field…

yiyang-msft:
When we got notification from textbox/password that string value changed, we should use the new value to set it on textbox/password themselves again because the values are the same.  This also fix the issue that we always setting the cursor position to end of the string when user typing.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1604)

### DIFF
--- a/Frameworks/UIKit/UITextField.mm
+++ b/Frameworks/UIKit/UITextField.mm
@@ -129,13 +129,17 @@ void SetTextControlContentVerticalAlignment(WXCControl* control, WXVerticalAlign
 
         [_secureModeLock lock];
         if (_secureTextMode) {
-            _passwordBox.password = _text;
+            if (![_passwordBox.password isEqualToString:_text]) {
+                _passwordBox.password = _text;
+            }
         } else {
-            _textBox.text = _text;
-            // Ensure caret at end of field in case we programmatically
-            // gain focus (becomeFirstResponder) after the text is set:
-            _textBox.selectionStart = [_text length];
-            _textBox.selectionLength = 0;
+            if (![_textBox.text isEqualToString:_text]) {
+                _textBox.text = _text;
+                // Ensure caret at end of field in case we programmatically
+                // gain focus (becomeFirstResponder) after the text is set:
+                _textBox.selectionStart = [_text length];
+                _textBox.selectionLength = 0;
+            }
         }
         [_secureModeLock unlock];
 
@@ -240,8 +244,8 @@ void SetTextControlContentVerticalAlignment(WXCControl* control, WXVerticalAlign
     _passwordBox.fontFamily = _textBox.fontFamily = [WUXMFontFamily makeInstanceWithName:[font _compatibleFamilyName]];
 
     // The following enums map from DWrite directly to Xaml
-    _passwordBox.fontStretch = _textBox.fontStretch = (WUTFontStretch)[font _fontStretch]; 
-    _passwordBox.fontStyle = _textBox.fontStyle = (WUTFontStyle)[font _fontStyle]; 
+    _passwordBox.fontStretch = _textBox.fontStretch = (WUTFontStretch)[font _fontStretch];
+    _passwordBox.fontStyle = _textBox.fontStyle = (WUTFontStyle)[font _fontStyle];
     _passwordBox.fontSize = _textBox.fontSize = font.pointSize;
 
     WUTFontWeight* weight = nil;
@@ -250,7 +254,7 @@ void SetTextControlContentVerticalAlignment(WXCControl* control, WXVerticalAlign
         case DWRITE_FONT_WEIGHT_THIN:
             weight = [WUTFontWeights thin];
             break;
-        //case DWRITE_FONT_WEIGHT_EXTRA_LIGHT:
+        // case DWRITE_FONT_WEIGHT_EXTRA_LIGHT:
         case DWRITE_FONT_WEIGHT_ULTRA_LIGHT:
             weight = [WUTFontWeights extraLight];
             break;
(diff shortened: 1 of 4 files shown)


Frameworks/UIKit/UITextField.mm
```diff
@@ -129,13 +129,17 @@ - (void)setText:(NSString*)text {
...
-            // Ensure caret at end of field in case we programmatically
-            // gain focus (becomeFirstResponder) after the text is set:
-            _textBox.selectionStart = [_text length];
-            _textBox.selectionLength = 0;
+            if (![_textBox.text isEqualToString:_text]) {
+                _textBox.text = _text;
+                // Ensure caret at end of field in case we programmatically
+                // gain focus (becomeFirstResponder) after the text is set:
+                _textBox.selectionStart = [_text length];
+                _textBox.selectionLength = 0;
```
DHowett-MSFT: This doesn't seem like it should fix the issue; when you delete a character it will still jump, because `_text != _textBox.text`?

yiyang-msft: both insertion/deletion case will be guarded by this newly added check, the code right above updated the cached value when insertion/deletion happens via text/password changed listener. we so when it comes down to the newly added check, the evaluation will be true and thus no-op.  

---
In reply to: [94637079](https://github.com/Microsoft/WinObjC/pull/1604#discussion_r94637079) [](ancestors = 94637079)

oliversa-msft:
:shipit:

Frameworks/UIKit/UITextField.mm
```diff
@@ -129,13 +129,17 @@ - (void)setText:(NSString*)text {
...
+            if (![_passwordBox.password isEqualToString:_text]) {
+                _passwordBox.password = _text;
+            }
         } else {
-            _textBox.text = _text;
-            // Ensure caret at end of field in case we programmatically
-            // gain focus (becomeFirstResponder) after the text is set:
-            _textBox.selectionStart = [_text length];
-            _textBox.selectionLength = 0;
+            if (![_textBox.text isEqualToString:_text]) {
```
jaredhms: > if (![_textBox.text isEqualToString:_text]) { [](start = 11, length = 46)

It seems like the problem is due to us calling setText in the textchanged handler.  If we just apply the Xaml controls' templates, can we get rid of all of these cached values?  That would set us up for an easier transition to a Xaml control implementation, too.

Frameworks/UIKit/XamlUtilities.mm
```diff
@@ -235,7 +239,6 @@ void SetControlBorderStyle(WXCControl* control, UITextBorderStyle style) {
             break;
 
         case UITextBorderStyleRoundedRect:
-            [control applyTemplate];
```
jaredhms: >          [control applyTemplate]; [](start = 2, length = 34)

Can you file a bug to scrub UIButton's use of applyTemplate?

Frameworks/UIKit/UITextField.mm
```diff
@@ -120,34 +113,30 @@ @implementation UITextField {
...
     }
+    [_secureModeLock unlock];
+
+    [self _adjustFontSizeToFitWidthOrApplyCurrentFont];
 }
 
 /**
  @Status Interoperable
 */
 - (NSString*)text {
```
jaredhms: @tadam-msft, should we add some UITextFiled tests that set/retrieve some of these values? or are we still waiting to finalize a pattern?  I think the pattern's relatively clear at this point.

yiyang-msft: These are still going through manual verification through XAMLCatalog. we can add more functional test once the pattern is fixed. I believe Oliver is still working out the detail how functional test/XAMLCatalog can share viewcontroller code. Once the contract is defined. we should be able to start to author more automated tests. 

---
In reply to: [95400508](https://github.com/Microsoft/WinObjC/pull/1604#discussion_r95400508) [](ancestors = 95400508)

Frameworks/UIKit/UITextField.mm
```diff
@@ -323,20 +306,33 @@ - (void)setTextColor:(UIColor*)color {
...
+    WUXMSolidColorBrush* colorBrush = nil;
+    if (_secureTextMode) {
+        colorBrush = rt_dynamic_cast<WUXMSolidColorBrush>(_passwordBox.background);
+        if (colorBrush != nil) {
+            return ConvertWUColorToUIColor(colorBrush.color);
+        } else {
+            return nil;
+        }
+    } else {
+        colorBrush = rt_dynamic_cast<WUXMSolidColorBrush>(_textBox.background);
```
jaredhms: shouldn't this be foreground?

Frameworks/UIKit/UITextField.mm
```diff
@@ -323,20 +306,33 @@ - (void)setTextColor:(UIColor*)color {
...
 - (UIColor*)textColor {
-    return _textColor;
+    WUXMSolidColorBrush* colorBrush = nil;
+    if (_secureTextMode) {
+        colorBrush = rt_dynamic_cast<WUXMSolidColorBrush>(_passwordBox.background);
+        if (colorBrush != nil) {
+            return ConvertWUColorToUIColor(colorBrush.color);
+        } else {
+            return nil;
+        }
```
jaredhms: nit: this could be pulled out of the if/else cases to avoid code duplication

Frameworks/UIKit/UITextField.mm
```diff
@@ -323,20 +306,33 @@ - (void)setTextColor:(UIColor*)color {
  @Status Interoperable
 */
 - (UIColor*)textColor {
-    return _textColor;
+    WUXMSolidColorBrush* colorBrush = nil;
+    if (_secureTextMode) {
+        colorBrush = rt_dynamic_cast<WUXMSolidColorBrush>(_passwordBox.background);
```
jaredhms: shouldn't this be foreground?

Frameworks/UIKit/UITextField.mm
```diff
@@ -352,19 +348,32 @@ - (void)setBackgroundColor:(UIColor*)color {
...
 - (UIColor*)backgroundColor {
-    return _backgroundColor;
+    WUXMSolidColorBrush* colorBrush = nil;
+    if (_secureTextMode) {
+        colorBrush = rt_dynamic_cast<WUXMSolidColorBrush>(_passwordBox.background);
+        if (colorBrush != nil) {
+            return ConvertWUColorToUIColor(colorBrush.color);
+        } else {
+            return nil;
+        }
```
jaredhms: nit: same here regarding pulling this out to avoid code duplication

Frameworks/UIKit/UITextField.mm
jaredhms: @yiyang-msft; does XamlCatalog let you toggle between secure/unsecure?

---
In reply to: [95400508](https://github.com/Microsoft/WinObjC/pull/1604#discussion_r95400508) [](ancestors = 95400508)

Frameworks/UIKit/UITextField.mm
```diff
@@ -352,19 +348,32 @@ - (void)setBackgroundColor:(UIColor*)color {
...
 */
 - (void)setTextAlignment:(UITextAlignment)alignment {
-    _alignment = alignment;
-
     [_secureModeLock lock];
     if (!_secureTextMode) {
         // passwordBox does not support text alignment
-        _textBox.textAlignment = ConvertUITextAlignmentToWXTextAlignment(_alignment);
+        _textBox.textAlignment = ConvertUITextAlignmentToWXTextAlignment(alignment);
     }
```
jaredhms: nit: else, should we log something like UNIMPLEMENTED_WITH_MSG("SecureTextEntry mode does not support UITextAlignment")?

Frameworks/UIKit/UITextField.mm
yiyang-msft: yes ;/

---
In reply to: [95470285](https://github.com/Microsoft/WinObjC/pull/1604#discussion_r95470285) [](ancestors = 95470285)

Frameworks/UIKit/UITextField.mm
yiyang-msft: good idea

---
In reply to: [95474383](https://github.com/Microsoft/WinObjC/pull/1604#discussion_r95474383) [](ancestors = 95474383)

jaredhms:
:shipit:

PR #1608, opened 2016-12-22
Fix issue [ARM] [uxe] WOCCatalog: Search bar - cancel button doesn’t do anything

yiyang-msft:
setting nil to background will disable hit-testing for cancel button which is the root cause of this issue. In the mean time, we hook up the searchBarCancelButtonClicked delegate and add test to it - by clearing the text in search bar to make it more obvious whether cancelbutton has worked or not.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1608)


oliversa-msft:
:shipit:

jaredhms:
:shipit:

PR #1623, opened 2017-01-04
Impromptu perf fix: NSString _fastestEncoding

ms-jihua:
Found while checking CTCatalog affine transform test.
In cases of longer strings, _DWriteGetFrame() spends an inordinate amount of time in:
_DWriteGetFrame (~30%) -> [NSString substring...] (~27%) -> [NSString _fastestEncodingWith...] (~25%)  (sample trace)

Changed this function to not call [self length] every iteration, fetch characters in chunks.
After this change, the same test looks like:
_DWriteGetFrame (~5%) -> [NSString substring...] (~1.2%) -> [NSString _fastestEncodingWith...] (~0.5%)  (sample trace)

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1623)

### DIFF
--- a/Frameworks/Foundation/NSString.mm
+++ b/Frameworks/Foundation/NSString.mm
@@ -493,12 +493,12 @@ BASE_CLASS_REQUIRED_IMPLS(NSString, NSStringPrototype, CFStringGetTypeID);
  @Notes
 */
 - (BOOL)getBytes:(void*)buffer
-       maxLength:(NSUInteger)maxBuf
-      usedLength:(NSUInteger*)usedLength
-        encoding:(NSStringEncoding)encoding
-         options:(NSStringEncodingConversionOptions)options
-           range:(NSRange)range
-  remainingRange:(NSRangePointer)left {
+         maxLength:(NSUInteger)maxBuf
+        usedLength:(NSUInteger*)usedLength
+          encoding:(NSStringEncoding)encoding
+           options:(NSStringEncodingConversionOptions)options
+             range:(NSRange)range
+    remainingRange:(NSRangePointer)left {
     CFIndex totalBytesWritten = 0;
     unsigned int numCharsProcessed = 0;
     CFStringEncoding cfStringEncoding = CFStringConvertNSStringEncodingToEncoding(encoding);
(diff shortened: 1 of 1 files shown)


Frameworks/Foundation/NSString.mm
```diff
@@ -493,12 +493,12 @@ - (const char*)UTF8String {
  @Notes
 */
 - (BOOL)getBytes:(void*)buffer
-       maxLength:(NSUInteger)maxBuf
-      usedLength:(NSUInteger*)usedLength
-        encoding:(NSStringEncoding)encoding
-         options:(NSStringEncodingConversionOptions)options
-           range:(NSRange)range
-  remainingRange:(NSRangePointer)left {
+         maxLength:(NSUInteger)maxBuf
```
ms-jihua: looks like our clang-format got updated (10/31) since the last change to this file (early sep)

Frameworks/Foundation/NSString.mm
```diff
@@ -2138,8 +2138,18 @@ - (CFStringEncoding)_smallestEncodingInCFStringEncoding {
...
     // Return Unicode encoding as soon as a single non-ASCII character is found. Otherwise, return ASCII encoding.
-    for (int32_t i = 0; i < [self length]; i++) {
-        if ([self characterAtIndex:i] > 0x7F) {
+    // Check characters in chunks for perf reasons
+    NSUInteger length = self.length;
+    const size_t bufferSize = 128;
+    NSRange range;
+
+    for (range.location = 0; range.location < length; range.location += range.length) {
+        UniChar buf[bufferSize];
```
rajsesh-msft: >UniChar buf[bufferSize] [](start = 8, length = 23)

move this buffer outside the for loop.

Frameworks/Foundation/NSString.mm
```diff
@@ -2138,8 +2138,19 @@ - (CFStringEncoding)_smallestEncodingInCFStringEncoding {
 
 - (CFStringEncoding)_fastestEncodingInCFStringEncoding {
     // Return Unicode encoding as soon as a single non-ASCII character is found. Otherwise, return ASCII encoding.
-    for (int32_t i = 0; i < [self length]; i++) {
-        if ([self characterAtIndex:i] > 0x7F) {
+    // Check characters in chunks for perf reasons
+    const size_t bufferSize = 128;
```
rajsesh-msft: >128 [](start = 30, length = 3)

consider making this bigger (512?).

rajsesh-msft:
:shipit:

Frameworks/Foundation/NSString.mm
ms-jihua: sure

PR #1627, opened 2017-01-04
Fix CoreAnimation functional tests after recent changes to CATransactions

oliversa-msft:
CATransactions are now executed synchronously which affected our CoreAnimation tests that validate the ObjC state is in sync with its backing XAML element state.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1627)

### DIFF
--- a/Frameworks/QuartzCore/CALayer.mm
+++ b/Frameworks/QuartzCore/CALayer.mm
@@ -1532,6 +1532,10 @@ static void doRecursiveAction(CALayer* layer, NSString* actionName) {
     priv->_backgroundColor = CGColorRetain(color);
     CGColorRelease(old);
 
+    // The CATransaction is now synchronous and in this scenario, the backgroundColor
+    // is updated almost immediately.By setting the priv->background first,
+    // we can ensure both states are kept in sync which is useful when performing
+    // automated testing.
     [CATransaction _setPropertyForLayer:self name:@"backgroundColor" value:(NSObject*)color];
 
     [self setNeedsDisplay];
--- a/Frameworks/QuartzCore/CALayer.mm
+++ b/Frameworks/QuartzCore/CALayer.mm
@@ -1528,10 +1528,11 @@ static void doRecursiveAction(CALayer* layer, NSString* actionName) {
         _ClearColorQuad(priv->backgroundColor);
     }
 
-    [CATransaction _setPropertyForLayer:self name:@"backgroundColor" value:(NSObject*)color];
     CGColorRef old = priv->_backgroundColor;
     priv->_backgroundColor = CGColorRetain(color);
     CGColorRelease(old);
+
+    [CATransaction _setPropertyForLayer:self name:@"backgroundColor" value:(NSObject*)color];
 
     [self setNeedsDisplay];
 }
--- a/tests/functionaltests/FunctionalTests.cpp
+++ b/tests/functionaltests/FunctionalTests.cpp
@@ -185,7 +185,7 @@ public:
     }
 
     TEST_METHOD_CLEANUP(NSURLCleanup) {
-        FunctionalTestCleanupUIApplication();
+        (void)FrameworkHelper::RunOnUIThread(&FunctionalTestCleanupUIApplication);
         return true;
     }
 
@@ -274,7 +274,7 @@ public:
     }
 
     TEST_METHOD_CLEANUP(NSUserDefaultsCleanup) {
-        FunctionalTestCleanupUIApplication();
+        (void)FrameworkHelper::RunOnUIThread(&FunctionalTestCleanupUIApplication);
         return true;
     }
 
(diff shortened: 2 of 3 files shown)


tests/functionaltests/FunctionalTests.cpp
```diff
@@ -691,7 +697,7 @@ class ProjectionTest {
     }
 
     TEST_METHOD_CLEANUP(ProjectionTestCleanup) {
-        FunctionalTestCleanupUIApplication();
+        (void)FrameworkHelper::RunOnUIThread(&FunctionalTestCleanupUIApplication);
```
jaredhms: >(void) [](start = 8, length = 6)

please remove

jaredhms: in all places

---
In reply to: [94668735](https://github.com/Microsoft/WinObjC/pull/1627#discussion_r94668735) [](ancestors = 94668735)

jaredhms: also - if RunOnUIThread returns an HRESULT, we should return SUCCEEDED(FrameworkHelper::RunOnUIThread(...

---
In reply to: [94668773](https://github.com/Microsoft/WinObjC/pull/1627#discussion_r94668773) [](ancestors = 94668773,94668735)

tests/functionaltests/Tests/CoreAnimationTests/CALayerAppearanceTests.mm
```diff
@@ -104,7 +106,7 @@ - (void)viewDidLoad {
     });
 
     [condition lock];
-    ASSERT_TRUE_MSG([condition waitUntilDate:[NSDate dateWithTimeIntervalSinceNow:c_testTimeoutInSec]],
+    ASSERT_TRUE_MSG(signaled || [condition waitUntilDate:[NSDate dateWithTimeIntervalSinceNow:c_testTimeoutInSec]],
```
jaredhms: >ASSERT_TRUE_MSG(signaled || [condi [](start = 4, length = 34)

Can you explain why we need the 'signaled' variable?

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -1528,11 +1528,12 @@ - (void)setBackgroundColor:(CGColorRef)color {
         _ClearColorQuad(priv->backgroundColor);
     }
 
-    [CATransaction _setPropertyForLayer:self name:@"backgroundColor" value:(NSObject*)color];
     CGColorRef old = priv->_backgroundColor;
     priv->_backgroundColor = CGColorRetain(color);
     CGColorRelease(old);
 
+    [CATransaction _setPropertyForLayer:self name:@"backgroundColor" value:(NSObject*)color];
```
yiyang-msft: >_setPropertyForLayer [](start = 19, length = 20)

mind adding  a comments why moving it down this is necessary?

tests/functionaltests/Tests/CoreAnimationTests/CALayerAppearanceTests.mm
oliversa-msft: It's the proper design when using NSCondition as per Apple doc ([https://developer.apple.com/reference/foundation/nscondition](https://developer.apple.com/reference/foundation/nscondition)). In this example, if the property change callback is fired before we get to the wait state, we can conclude the test at that point using the signaled predicate.

---
In reply to: [94669086](https://github.com/Microsoft/WinObjC/pull/1627#discussion_r94669086) [](ancestors = 94669086)

tests/functionaltests/FunctionalTests.cpp
oliversa-msft: I assumed that the return result may have affected the outcome of the test. However, it doesn't so I will move to test the HRESULT as well.

---
In reply to: [94668846](https://github.com/Microsoft/WinObjC/pull/1627#discussion_r94668846) [](ancestors = 94668846,94668773,94668735)

tests/functionaltests/Tests/CoreAnimationTests/CALayerAppearanceTests.mm
jaredhms: Ah, right.  I think Brian already had to go fix up a bunch of other tests regarding this same issue.  Can we use a Win32 or C++ construct for this?  That's a lot of code for what should be a simple wait.

---
In reply to: [94687381](https://github.com/Microsoft/WinObjC/pull/1627#discussion_r94687381) [](ancestors = 94687381,94669086)

jaredhms:
:shipit:

yiyang-msft:
:shipit:

oliversa-msft:
CATransactions are now executed synchronously which affected our CoreAnimation tests that validate the ObjC state is in sync with its backing XAML element state. Moved clean-up of several tests to run on the UI thread.

PR #1633, opened 2017-01-05
Fix our busted CFKeyedArchiverUID support. Fixes #649.

### LINKED ISSUE #649: Binary property lists and keyed archives from the reference platform may behave badly

DHowett-MSFT:
For some reason, we had a separate binary plist writer too. It was only being used for NSDictionary and NSArray `writeToFile:`.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1633)


Frameworks/Foundation/NSKeyedArchiver.mm
```diff
@@ -65,12 +66,7 @@ static inline void _checkKey(NSString* aKey, NSMutableDictionary* _enc) {
 * Make a dictionary referring to the object at ref in the array of all objects.
 */
 static id makeReference(unsigned ref) {
```
rajsesh-msft: >ref [](start = 33, length = 3)

nit: ref is a special word in some languages and can confuse some of us, could we rename this?

bbowman:
:shipit:

rajsesh-msft:
:shipit:

rajsesh-msft:
It would be a good idea to test islandwoodnews with this change.

msft-Jeyaram:
:shipit:

DHowett-MSFT:
@rajsesh-msft IWNews launches and loads its config properly. It doesn't look like it archives (via keyed archival) any data at runtime, however!

Issue #988, opened 2016-09-09
UIWebView method - (NSString*)stringByEvaluatingJavaScriptFromString:(NSString*)string throws exception


Juju4ka1:
Hi.
If you create UIWebView and load simple html page with simple javascript in it and then try to invoke this script from native using 'stringByEvaluatingJavaScriptFromString:' method, you will get 'UNSException' exception.

For instance: (simple html page was taken from example https://code.msdn.microsoft.com/windowsapps/XAML-WebView-control-sample-58ad63f7)

```
 self.webView = [[UIWebView alloc] initWithFrame:self.view.frame];
    self.webView.delegate = self;
NSString* appHtml = @"<html><head><script type='text/javascript'>function doSomething(){  document.getElementById('myDiv').innerText = 'GoodBye';return 'Hello World!'; }</script></head><body><div id='myDiv'>Hello</div></body></html>";   
[self.webView loadHTMLString:appHtml baseURL:nil];
[self.view addSubview:_webView];

.....
//Invoke script when webView finished loading
- (void)webViewDidFinishLoad:(UIWebView *)webView {
[_webView stringByEvaluatingJavaScriptFromString:@"doSomething"]; //You will get exception here
}
```

jboich:
Hi, there.

As a hack around solution to get this feature working, we added method 
```
- (void)evaluateJavaScript:(NSString *)javaScriptString 
         completionHandler:(void (^)(id, NSError *error))completionHandler;
```
to UIWebView class and use [_xamlWebControl.InvokeScriptAsync](https://msdn.microsoft.com/en-us/library/windows.ui.xaml.controls.webview.invokescriptasync.aspx) in implementation.

DHowett-MSFT:
@jboich Awesome. Have you run into any compatibility issues?

Juju4ka1:
Nope, because we wrapped code using preprocessor directive. So objective-c code is:
```
#if !WINOBJC
    NSString *result = [self.webView stringByEvaluatingJavaScriptFromString:javascriptCommand];
    if (completionHandler) {
        completionHandler(result,nil);
    }
#else
    [self.webView evaluateJavaScript:javascriptCommand completionHandler:completionHandler];
#endif
```
where webView is an instance of UIWebView class.

jaredhms:
Looks like a dupe of #582.  

[...]

PR #1638, opened 2017-01-05
Workaround to evaluate javascript using UIWebView

jboich:
Workaround for issue [#988 ](https://github.com/Microsoft/WinObjC/issues/988)

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1638)

### DIFF
--- a/Frameworks/UIKit/UIWebView.mm
+++ b/Frameworks/UIKit/UIWebView.mm
@@ -15,6 +15,7 @@
 //******************************************************************************
 
 #include "Starboard.h"
+#include "StubReturn.h"
 #include "UIKit/UIKit.h"
 #include "UIKit/UIView.h"
 #include "CoreGraphics/CGContext.h"
--- a/Frameworks/UIKit/UIWebView.mm
+++ b/Frameworks/UIKit/UIWebView.mm
@@ -251,7 +251,8 @@ static void _initUIWebView(UIWebView* self) {
 }
 
 /**
- @Status Interoperable
+ @Status Caveat
+  @Notes This is a workaround. Original UIWebView does not have this method
 */
 - (void)evaluateJavaScript:(NSString*)javaScriptString completionHandler:(void (^)(id, NSError*))completionHandler {
   
--- a/Frameworks/UIKit/UIWebView.mm
+++ b/Frameworks/UIKit/UIWebView.mm
@@ -255,16 +255,17 @@ static void _initUIWebView(UIWebView* self) {
   @Notes This is a workaround. Original UIWebView does not have this method
 */
 - (void)evaluateJavaScript:(NSString*)javaScriptString completionHandler:(void (^)(id, NSError*))completionHandler {
-  
-    [_xamlWebControl invokeScriptAsync:@"eval" arguments:[NSArray arrayWithObject:javaScriptString] success:^void(NSString* success) {
-        if(completionHandler != nil) {
-            completionHandler(success, nil);
+    [_xamlWebControl invokeScriptAsync:@"eval"
+        arguments:[NSArray arrayWithObject:javaScriptString]
+        success:^void(NSString* success) {
+            if (completionHandler != nil) {
+                completionHandler(success, nil);
+            }
         }
-    } 
         failure:^void(NSError* failure) {
-                if(completionHandler != nil) {
-                    completionHandler(nil, failure);
-                }
+            if (completionHandler != nil) {
+                completionHandler(nil, failure);
+            }
         }];
 }
 
--- a/Frameworks/UIKit/UIWebView.mm
+++ b/Frameworks/UIKit/UIWebView.mm
@@ -243,11 +243,12 @@ static void _initUIWebView(UIWebView* self) {
 }
 
 /**
- @Status Interoperable
+ @Status Gap
 */
 - (NSString*)stringByEvaluatingJavaScriptFromString:(NSString*)string {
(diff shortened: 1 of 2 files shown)


Frameworks/UIKit/UIWebView.mm
```diff
@@ -251,6 +251,25 @@ - (NSString*)stringByEvaluatingJavaScriptFromString:(NSString*)string {
 }
 
 /**
+ @Status Caveat
+  @Notes This is a workaround. Original UIWebView does not have this method
+*/
+- (void)evaluateJavaScript:(NSString*)javaScriptString completionHandler:(void (^)(id, NSError*))completionHandler {
```
rajsesh-msft: Rather than adding a new function, you can do this in stringByEvaluatingJavaScriptFromString.

jaredhms: stringByEvaluatingJavaScriptFromString isn't async though.

---
In reply to: [95190774](https://github.com/Microsoft/WinObjC/pull/1638#discussion_r95190774) [](ancestors = 95190774)

rajsesh-msft: ah, yes.  Inexplicably, I thought this function was "waiting" for the eval.  It isn't. Apologies for the false flag.

---
In reply to: [95208519](https://github.com/Microsoft/WinObjC/pull/1638#discussion_r95208519) [](ancestors = 95208519,95190774)

jaredhms:
>     return ret;

Since we know this will never work, maybe we should change this status to Stub and replace this code with:

UNIMPLEMENTED_WITH_MSG("stringByEvaluatingJavaScriptFromString is not supported on our platform.  Call evaluateJavaScript:completionHandler: instead.");
return StubReturn();

?

We'd also need to add STUB_METHOD to the declaration in the header file, possibly with the same comment saying why it's stubbed out:

// stringByEvaluatingJavaScriptFromString is not supported on our platform.  Call evaluateJavaScript:completionHandler: instead.
- (NSString*)stringByEvaluatingJavaScriptFromString:(NSString*)script STUB_METHOD;

---
Refers to: Frameworks/UIKit/UIWebView.mm:250 in 6c9c08e. [](commit_id = 6c9c08ec9a2365ebc0304b80a1fcf8d5f1e2d2b1, deletion_comment = False)

jaredhms:
>     return ret;

@Raj Seshasankaran - this is another place where an @Status Unsupported and a UNSUPPORTED_WITH_MSG would be helpful; agree?  

---
In reply to: [271356945](https://github.com/Microsoft/WinObjC/pull/1638#issuecomment-271356945) [](ancestors = 271356945)

---
Refers to: Frameworks/UIKit/UIWebView.mm:250 in 6c9c08e. [](commit_id = 6c9c08ec9a2365ebc0304b80a1fcf8d5f1e2d2b1, deletion_comment = False)

rajsesh-msft:
>     return ret;

We don't (yet) have a status unsupported.  At the very least, this should be status Gap.

---
In reply to: [271357323](https://github.com/Microsoft/WinObjC/pull/1638#issuecomment-271357323) [](ancestors = 271357323,271356945)

---
Refers to: Frameworks/UIKit/UIWebView.mm:250 in 6c9c08e. [](commit_id = 6c9c08ec9a2365ebc0304b80a1fcf8d5f1e2d2b1, deletion_comment = False)

jaredhms:
>     return ret;

@jboich, can you please use status Gap for now rather than status Stub?

/**
 @Status Gap
*/


---
In reply to: [271357323](https://github.com/Microsoft/WinObjC/pull/1638#issuecomment-271357323) [](ancestors = 271357323,271356945)

---
Refers to: Frameworks/UIKit/UIWebView.mm:250 in 6c9c08e. [](commit_id = 6c9c08ec9a2365ebc0304b80a1fcf8d5f1e2d2b1, deletion_comment = False)

jaredhms:
:shipit:

jaredhms:
Thanks for the fix and updates; kicking off a CI build.

Frameworks/UIKit/UIWebView.mm
```diff
@@ -243,11 +243,31 @@ - (void)setDetectsPhoneNumbers:(BOOL)detect {
...
 /**
- @Status Interoperable
+ @Status Gap
 */
 - (NSString*)stringByEvaluatingJavaScriptFromString:(NSString*)string {
-    NSString* ret = [_xamlWebControl invokeScript:@"eval" arguments:[NSArray arrayWithObject:string]];
-    return ret;
+    UNIMPLEMENTED_WITH_MSG(
+        "stringByEvaluatingJavaScriptFromString is not supported on our platform. Call evaluateJavaScript:completionHandler: instead.");
+    return StubReturn();
```
jaredhms: This is hitting a build break;

D:\Build\Agent1\_work\2\s\Frameworks\UIKit\UIWebView.mm(251,12): error : use of undeclared identifier 'StubReturn' [D:\Build\Agent1\_work\2\s\build\UIKit\lib\UIKitLib.vcxproj]

Can you please add #include "StubReturn.h" just below the #include "Starboard.h"?

jboich: Sure, I will add. Sorry, I missed this. Why the build did not break in Debug mode?

jaredhms: Thanks, no problem.  I'm not sure why it didn't fail for you; it failed on both debug and release CI builds.

---
In reply to: [95416396](https://github.com/Microsoft/WinObjC/pull/1638#discussion_r95416396) [](ancestors = 95416396)

Issue #1571, opened 2016-12-16
Canabalt, Climbers [x86|Debug, Release] Minimizing and reopening the app disables interaction


nicholasgerard-msft:
x86 running 14975
https://github.com/Microsoft/WinObjC/commit/4804ed9bf1b16225116ff7bb06477ce3c8c300d7

Run Canabalt or Climbers, minimize the game window and reopen it. Graphics will be frozen in the game window but music continues. Maximizing the window restarts graphics.

rajsesh-msft:
@nicholasgerard-msft do you have any regression info?

nicholasgerard-msft:
I've never seen this before (I've tried the same sequence on previous builds) and didn't find any bugs matching it.

________________________________

ashvarma:
I tried this on several previous releases and the behavior is the same. The sound pauses as does the graphics and activating/maximizing the app doesn't resume the app - it basically stays frozen (the graphics doesn't "restart"). 
Here are the builds I tried - 
20161213.2
20161130.1
20160908.1

rajsesh-msft:
We broke this once before.  See #1026.  At this point, at least some apps worked.

rajsesh-msft:
Fixed with #1640.

PR #1640, opened 2017-01-05
Reverting:  "Moving the bulk of app initialization off of the app activation path.  We should let the app activate before calling into app code to finish starting up."

jaredhms:
Reverting: "Moving the bulk of app initialization off of the app activation path.  We should let the app activate before calling into app code to finish starting up."

It doesn't seem to be needed for perf improvements, and it broke some functional tests.  We'll resurrect the change if needed, along with fixed tests.

This reverts commit b3e269ee1ddf7bb48f9f4b0f61611d5d381fff10.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1640)


oliversa-msft:
:shipit:

rajsesh-msft:
The original change also caused #1571, and the revert has fixed that as well.

Issue #1595, opened 2016-12-20
Islandwood News app - unable to get focus on search text box


ashvarma:
master build: 20161213.2
flavor: x86

With touch you have to tap multiple times before it gets focus 
Tried repeatedly with mouse - didn't work

PR #1641, opened 2017-01-05
Fix[GitHub Issue #1595] Islandwood News app - unable to get focus on search text box

yiyang-msft:
The issue here is that app sets an empty text to prompt text in UISearchbar which meant to clear or remove the prompt label if it exists. Our implementation,  currently does not handle setting prompt to nil/empty prompt text case ( thus quick return or removing prompt label), it creates the prompt label blindly, which app layout code didn't expect this label thus didn't increase the Searchbar height, and this label (without any prompt text) blocks part of User input to the searchbox in UISearchbar.  

this change allows UISearchbar handles setting prompt text to nil/empty correctly.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1641)


oliversa-msft:
:shipit:

Frameworks/UIKit/UISearchBar.mm
```diff
@@ -72,17 +72,33 @@ - (BOOL)resignFirstResponder {
...
 - (void)setPrompt:(NSString*)prompt {
-    if (_promptLabel == nil) {
-        CGRect promptFrame = CGRectMake(0, c_marginTopForPrompt, self.frame.size.width, c_defaultTextFieldHeight);
-
-        self->_promptLabel.attach([[UILabel alloc] initWithFrame:promptFrame]);
-        [self->_promptLabel setTextAlignment:UITextAlignmentCenter];
-        [self->_promptLabel setBackgroundColor:nil];
-        [self->_promptLabel setTextColor:[UIColor blackColor]];
-        [self addSubview:self->_promptLabel];
+    if (_prompt == prompt || (([_prompt length] == 0) && ([prompt length] == 0))) {
```
jaredhms: >_prompt == prompt [](start = 8, length = 17)

don't we want to do a string comparison here?  in which case, wouldn't it also work for the check against an empty string?

jaredhms: [_prompt isEqualToString:prompt] ?

---
In reply to: [94880288](https://github.com/Microsoft/WinObjC/pull/1641#discussion_r94880288) [](ancestors = 94880288)

Frameworks/UIKit/UISearchBar.mm
```diff
@@ -72,17 +72,33 @@ - (BOOL)resignFirstResponder {
...
+        // if old prompt text is the same as the new prompt text, quick return
+        // also, treat empty string @"" prompt as nil because setting empty string to prompt does not
+        // show prompt label on reference platform.
+        return;
     }
 
+    // otherwise, old prompt text is different from new prompt text, need either add or remove prompt label
     _prompt = prompt;
```
jaredhms: >prompt; [](start = 14, length = 7)

don't we need to copy?

Frameworks/UIKit/UISearchBar.mm
yiyang-msft: _prompt == prompt will do the comparison for string . length check if for nil and empty

---
In reply to: [94880288](https://github.com/Microsoft/WinObjC/pull/1641#discussion_r94880288) [](ancestors = 94880288)

Frameworks/UIKit/UISearchBar.mm
```diff
@@ -72,17 +72,33 @@ - (BOOL)resignFirstResponder {
...
+    if ([_prompt length] == 0) {
+        // remove prompt label since prompt text is either nil or empty
+        [_promptLabel removeFromSuperview];
+        _promptLabel = nil;
+    } else {
+        // need to add prompt label if it does not exist since new prompt text is not empty
+        if (_promptLabel == nil) {
+            CGRect promptFrame = CGRectMake(0, c_marginTopForPrompt, self.frame.size.width, c_defaultTextFieldHeight);
+
+            self->_promptLabel.attach([[UILabel alloc] initWithFrame:promptFrame]);
```
jaredhms: >self-> [](start = 12, length = 6)

nit: probably don't need the 'self' on these?

Frameworks/UIKit/UISearchBar.mm
```diff
@@ -72,17 +72,33 @@ - (BOOL)resignFirstResponder {
...
+            [self->_promptLabel setTextAlignment:UITextAlignmentCenter];
+            [self->_promptLabel setBackgroundColor:nil];
+            [self->_promptLabel setTextColor:[UIColor blackColor]];
+            [self addSubview:self->_promptLabel];
+        }
+    }
+
     [self setNeedsDisplay];
```
jaredhms: >  [self setNeedsDisplay]; [](start = 2, length = 25)

not new code, but since we assign the text in layoutSubviews, shouldn't this be setNeedsLayout?

Frameworks/UIKit/UISearchBar.mm
DHowett-MSFT: I don't think `==` is what you want here.

jaredhms: Yi agrees; we discussed offline :)

Frameworks/UIKit/UISearchBar.mm
```diff
@@ -72,18 +72,32 @@ - (BOOL)resignFirstResponder {
...
-    if (_promptLabel == nil) {
-        CGRect promptFrame = CGRectMake(0, c_marginTopForPrompt, self.frame.size.width, c_defaultTextFieldHeight);
-
-        self->_promptLabel.attach([[UILabel alloc] initWithFrame:promptFrame]);
-        [self->_promptLabel setTextAlignment:UITextAlignmentCenter];
-        [self->_promptLabel setBackgroundColor:nil];
-        [self->_promptLabel setTextColor:[UIColor blackColor]];
-        [self addSubview:self->_promptLabel];
+    if ((([_prompt length] == 0) && ([prompt length] == 0)) || [_prompt isEqualToString:prompt]) {
+        // if old prompt text is "equlal" to the new prompt text (nil and emptyString are considered equal), quick return
```
jaredhms: >equlal [](start = 34, length = 6)

nit: typo

Frameworks/UIKit/UISearchBar.mm
```diff
@@ -72,18 +72,32 @@ - (BOOL)resignFirstResponder {
...
+        return;
+    }
+
+    // otherwise, old prompt text is different from new prompt text, need either add or remove prompt label
+    _prompt = [prompt copy];
```
jaredhms: > _prompt = [prompt copy]; [](start = 3, length = 25)

how will this get freed?

Frameworks/UIKit/UISearchBar.mm
```diff
@@ -72,18 +73,39 @@ - (BOOL)resignFirstResponder {
...
-        [self->_promptLabel setBackgroundColor:nil];
-        [self->_promptLabel setTextColor:[UIColor blackColor]];
-        [self addSubview:self->_promptLabel];
+    if ((([_prompt length] == 0) && ([prompt length] == 0)) || [_prompt isEqualToString:prompt]) {
+        // if old prompt text is "equal" to the new prompt text (nil and emptyString are considered equal), quick return
+        return;
+    }
+
+    // otherwise, old prompt text is different from new prompt text, need either add or remove prompt label
+    _prompt = [prompt copy];
```
jaredhms: IIRC copy will return an object with a retain count of 1, so this similar to assigning a StrongId<NSString> = [[NSString alloc] init] (which ends up bumping the retain count to 2), thus still leaking the pointer.  I'd confirm with Dustin or step through the debugger, but I *think* you want to .attach here to avoid the extra refcount bump.

jaredhms:
:shipit:

PR #1651, opened 2017-01-06
Reintroduce the legacy UID parser. Fixes #1650.

### LINKED ISSUE #1650: Regression from #1633: Legacy archives *do* in fact need a compatibility shim.
#1633 introduced a silly regression.

NSKeyedArchiver was generating CFKeyedArchiverUIDs manually, by way of their legacy XML-encoded dictionary format.
NSPropertyListWriter_binary knew about that hack and wrote a real UID (tag `0x80`) to the binary plist.
When we switched to CoreFoundation in early 2016, we switched to its property list writer. It did not know about that hack.

As tested in [the new tests added by #1633](https://github.com/Microsoft/WinObjC/blob/develop/tests/unittests/Foundation/NSPropertyListSerializationTests.mm#L109), a dictionary encoding a fake `CF$UID` will round-trip as a dictionary.

Because of that, all archives created since that switch are *invalid*, and will not parse on the reference platform. Since the merge of #1633 brings us into compliance with the reference platform and the CFKeyedArchiverUID format, we fail as well.

Found by @pradipd.

DHowett-MSFT:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1651)


pradipd:
:shipit:

PR #1652, opened 2017-01-06
Change CT/CG text rendering to default on DWRITE_RENDERING/MEASURING_…

### LINKED ISSUE #1594: Canabalt - about screen text is blurry
master build: 20161213.2 
flavor: x86

I compared it with an image from build 20160923.1 and we've regressed this. 
Images attached. 

canabalt_aboutscreen_20160923 1drop
canabalt_aboutscreen_20161213 2drop

ms-jihua:
…MODE_GDI_NATURAL

Fixes #1594

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1652)


ms-jihua:
I tried a few different rendering modes. The GDI modes were the only ones that actually fixed the problem while not introducing noticeable aliasing to other test apps like iLabyrinth. Between GDI_CLASSIC and GDI_NATURAL, there was not very much noticeable difference, but GDI_NATURAL seemed to look slightly better? (very scientific)

ms-jihua:
We decided that this solution was not satisfactory - going to investigate another one.

PR #1657, opened 2017-01-09
NSRegularExpression firstMatchInString should return nil if there is no match

rajsesh-msft:
NSRegularExpression firstMatchInString should return nil if there is no match.  Currently we are returning a non-nil value with range (NSNotFound, 0), which is not correct.  The documentation is vague on
this, but the examples on the documentation page show that the match
could be nil.  This fixes an internal test app that was used to parse
xml.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1657)

### DIFF
--- a/Frameworks/Foundation/NSRegularExpression.mm
+++ b/Frameworks/Foundation/NSRegularExpression.mm
@@ -129,7 +129,6 @@ static void _matchCallBack(void* context, CFRange* ranges, CFIndex count, _CFReg
  @Status Interoperable
 */
 - (NSTextCheckingResult*)firstMatchInString:(NSString*)string options:(NSMatchingOptions)options range:(NSRange)range {
-    NSRange tempRange = NSMakeRange(NSNotFound, 0);
     __block NSTextCheckingResult* result = nil;
 
     [self enumerateMatchesInString:string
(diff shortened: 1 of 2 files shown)


Frameworks/Foundation/NSRegularExpression.mm
```diff
@@ -130,8 +130,7 @@ - (NSUInteger)numberOfMatchesInString:(NSString*)string options:(NSMatchingOptio
 */
 - (NSTextCheckingResult*)firstMatchInString:(NSString*)string options:(NSMatchingOptions)options range:(NSRange)range {
     NSRange tempRange = NSMakeRange(NSNotFound, 0);
```
bviglietta: >NSRange tempRange = NSMakeRange(NSNotFound, 0); [](start = 4, length = 47)

This line can be removed.

bviglietta:
:shipit:

MSFTFox:
:shipit:

msft-Jeyaram:
:shipit:

tests/unittests/Foundation/NSRegularExpressionTests.m
```diff
@@ -174,6 +174,11 @@ NSUInteger checkMatches(NSRegularExpression* regex, NSString* testString) {
 
     ASSERT_EQ(theRange.location, NSNotFound);
     ASSERT_EQ(theRange.length, 0);
+
+    StrongId<NSTextCheckingResult> result =
```
DHowett-MSFT: There's not much point in sticking an autoreleased thing into a StrongId ;P

rajsesh-msft: The rest of the code in the file does exactly that.  Went with when in Rome rule here :(.  

---
In reply to: [95231289](https://github.com/Microsoft/WinObjC/pull/1657#discussion_r95231289) [](ancestors = 95231289)

DHowett-MSFT:
(I signed off, just cheekily used the clock emoji 😉)

rajsesh-msft:
Got me good :)

---
In reply to: [271388626](https://github.com/Microsoft/WinObjC/pull/1657#issuecomment-271388626) [](ancestors = 271388626)

PR #1660, opened 2017-01-09
Waking the NSRunLoop upon app resume.  

jaredhms:
Without this change, we stay blocked in MainDispatcher::MainRunLoopTimedMultipleWait, and we never make it back into _DispatchMainRunLoopOnUIThread to continue processing messages until the app's window is resized.

This fixes the VSO bug around Climbers not resuming from PLM suspension.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1660)


yiyang-msft:
:shipit:

rajsesh-msft:
:shipit:

Issue #832, opened 2016-08-05
NSHTTPURLResponse does not respect MIME type when generating filenames.


DHowett-MSFT:
Potential fix:

Get the extension from MobileCoreServices,

```
CFStringRef uti = UTTypeCreatePreferredIdentifierForTag(kUTTagClassMIMEType, mimeType, nullptr);
CFStringRef preferredExtension = UTTypeCopyPreferredTagWithClass(uti, kUTTagClassFilenameExtension);
```

and make sure the filename ends with it.

Not sure yet if Foundation can take a dependency on MobileCoreServices.

DHowett-MSFT:
Per the [documentation](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSURLResponse_Class/index.html#//apple_ref/occ/instp/NSURLResponse/suggestedFilename),
>  In most cases, this property appends the proper file extension based on the MIME type. Accessing this property always returns a valid filename regardless of whether the resource is saved to disk.

rajsesh-msft:
Our mobilecoreservices implementation is quite limited and has a static map.  Also depends on foundation, so we have a cyclic dependency.

DHowett-MSFT:
@rajsesh-msft It looks like the dependency on Foundation is only for NSString message dispatch, and it's only in logging codepaths. We might be able to get away without linking it, or stripping out the dependency completely.

It uses CoreFoundation everywhere else internally. It's awesome.

DHowett-MSFT:
@mnithish might have some more insights here!

mnithish:
Mobilecoreservices should only depend on CF and we should be able to do that now with the new CF we have. Foundation can then depend on mobilecoreservices

PR #1662, opened 2017-01-09
Implementation of  proper filename[extension] based on MIME type. #832

msft-Jeyaram:
Changes:
- MobileCoreServices has the map to look up mime extensions
- Remove the dependency MobileCoreServices has on Foundation 
- Foundation links MobileCoreServices
- Viola

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1662)


Frameworks/Foundation/NSURLResponse.mm
```diff
@@ -89,8 +94,22 @@ - (id)_initWithURL:(NSURL*)url
...
+                                                       static_cast<CFStringRef>(mimeType),
+                                                       nullptr)
+            };
+            StrongId<NSString> extension {
+                static_cast<NSString*>(
+                    _UTTypeCopyPreferredTagWithClass(identifier.get(), static_cast<CFStringRef>(@"public.filename-extension")))
+            };
+
+            // If the filename already has the extension do not append
+            _suggestedFilename.attach((extension && ![extension isEqualToString:[fileName pathExtension]]) ?
```
rajsesh-msft: Not sure that we want a situation such as "filename.pdf.txt", that is adding a new extension if one already exists.  Is this expected?

Frameworks/Foundation/_UTTypeInternal.mm
```diff
@@ -0,0 +1,585 @@
...
+    }
+
+    return NULL;
+}
+
+#pragma endregion internalMapHelpers
+
+#pragma region directFunctionalInteropMethods
+/**
+ @Status Caveat
```
rajsesh-msft: Annotations don't make sense here, these are private functions.

msft-Jeyaram: removing it and just leaving the notes

---
In reply to: [95273116](https://github.com/Microsoft/WinObjC/pull/1662#discussion_r95273116) [](ancestors = 95273116)

Frameworks/Foundation/NSURLResponse.mm
msft-Jeyaram: this case ensures that if something has foo.pdf and the designated extension is .txt , it will get appended,
but if the designated extension is .pdf the it won't

---
In reply to: [95272900](https://github.com/Microsoft/WinObjC/pull/1662#discussion_r95272900) [](ancestors = 95272900)

Frameworks/Foundation/NSURLResponse.mm
```diff
@@ -89,8 +94,22 @@ - (id)_initWithURL:(NSURL*)url
...
+                _UTTypeCreatePreferredIdentifierForTag(static_cast<CFStringRef>(@"public.mime-type"),
+                                                       static_cast<CFStringRef>(mimeType),
+                                                       nullptr)
+            };
+            StrongId<NSString> extension {
+                static_cast<NSString*>(
+                    _UTTypeCopyPreferredTagWithClass(identifier.get(), static_cast<CFStringRef>(@"public.filename-extension")))
+            };
+
+            // If the filename already has the extension do not append
```
msft-Jeyaram: >// If the filename already has the extension do not append [](start = 12, length = 58)

more like: // If the filename already has the expected extension (based on MIME type_ do not append

build/Foundation/dll/Foundation.def
```diff
@@ -874,4 +874,11 @@ LIBRARY Foundation
         ; MainDispatcher
         ScheduleMainRunLoop
         ScheduleMainRunLoopAsync
-        SetupMainRunLoopTimedMultipleWaiter
\ No newline at end of file
+        SetupMainRunLoopTimedMultipleWaiter
+
+        ;UTTypeHelpers (Internal)
+        _UTTypeCreatePreferredIdentifierForTag
+        _UTTypeCreateAllIdentifiersForTag
```
DHowett-MSFT: !! Absolutely not!

DHowett-MSFT: The solution to the mixed up dependency chain is to remove the use of foundation (which, iirc, is *only logging-related*) from MobileCoreServices, not to merge MobileCoreServices into Foundation

---
In reply to: [95276171](https://github.com/Microsoft/WinObjC/pull/1662#discussion_r95276171) [](ancestors = 95276171)

DHowett-MSFT:
:clock1:

tests/unittests/Foundation/ReferenceFoundation/TestNSURLResponse.mm
```diff
@@ -34,6 +34,7 @@
     NSURLResponse* res =
         [[[NSURLResponse alloc] initWithURL:url MIMEType:@"text/plain" expectedContentLength:0 textEncodingName:nil] autorelease];
     ASSERT_OBJCEQ_MSG(url, res.URL, @"should be the expected url");
+    ASSERT_OBJCEQ(@"path.txt", res.suggestedFilename);
 }
```
DHowett-MSFT: These are the reference tests. We must make certain that they work on the reference platform, and we must make sure we document every change we make to them.

Where possible, do tests for this sort of stuff outside of the `Reference*` directories.

msft-Jeyaram: These were run on reference platform and they pass.
Yep we should keep the separate, but it was soo tempting to add an extra check here.

---
In reply to: [95276334](https://github.com/Microsoft/WinObjC/pull/1662#discussion_r95276334) [](ancestors = 95276334)

tests/unittests/Foundation/ReferenceFoundation/TestNSURLResponse.mm
```diff
@@ -74,22 +77,22 @@
...
+    ASSERT_OBJCEQ(@"test", res2.suggestedFilename);
 }
 
-// Disabled pending GH#832; we do not respect the incoming MIME type when generating filenames. OS X does.
-OSX_DISABLED_TEST(NSURLResponse, SuggestedFilename) {
+TEST(NSURLResponse, SuggestedFilename) {
     NSURL* url = [NSURL URLWithString:@"a/test/name.extension"];
     NSURLResponse* res =
         [[[NSURLResponse alloc] initWithURL:url MIMEType:@"text/plain" expectedContentLength:0 textEncodingName:nil] autorelease];
-    ASSERT_OBJCEQ(@"name.extension", res.suggestedFilename);
```
msft-Jeyaram: >name.extension [](start = 20, length = 14)

the old value was hardcoded to work with the improper implementation.
the actual values were ran on OSX and they pass.

Frameworks/Foundation/NSURLResponse.mm
```diff
@@ -89,8 +94,22 @@ - (id)_initWithURL:(NSURL*)url
...
+        if (fileName && ![fileName isEqualToString:s_unknownFileName]) {
+            woc::unique_cf<CFStringRef> identifier {
+                _UTTypeCreatePreferredIdentifierForTag(static_cast<CFStringRef>(@"public.mime-type"),
+                                                       static_cast<CFStringRef>(mimeType),
+                                                       nullptr)
+            };
+            StrongId<NSString> extension {
+                static_cast<NSString*>(
+                    _UTTypeCopyPreferredTagWithClass(identifier.get(), static_cast<CFStringRef>(@"public.filename-extension")))
+            };
```
bbowman: well that way to initialize variables looks absurdly confusing to me. Why did you choose multiline brace init that makes it look like a function definition?

tests/unittests/Foundation/ReferenceFoundation/TestNSURLResponse.mm
DHowett-MSFT: These three tests should not even test filenames; nor should the ones below. Only the filename tests should test them!

---
In reply to: [95277050](https://github.com/Microsoft/WinObjC/pull/1662#discussion_r95277050) [](ancestors = 95277050,95276334)

Frameworks/Foundation/NSURLResponse.mm
DHowett-MSFT: this also masks the fact that it will be adding another +1, leaking this extension.

---
In reply to: [95277280](https://github.com/Microsoft/WinObjC/pull/1662#discussion_r95277280) [](ancestors = 95277280)

tests/unittests/Foundation/ReferenceFoundation/TestNSURLResponse.mm
msft-Jeyaram: I'll remove the extra checks, but given how our tests are the name of the test case + extra checking is usual.
We can be strict and keep it to only such that.

---
In reply to: [95277356](https://github.com/Microsoft/WinObjC/pull/1662#discussion_r95277356) [](ancestors = 95277356,95277050,95276334)

build/Foundation/dll/Foundation.def
bbowman: Talked to dustin offline a bit. I think he is correct (per usual) and the "Core" set of frameworks are below foundation. It means I will need to adjust my packaging to include them in the"Core" package but it is almost certainly more correct and cleaner overall than adding these privates here. Sorry for giving you bad advice.

---
In reply to: [95276230](https://github.com/Microsoft/WinObjC/pull/1662#discussion_r95276230) [](ancestors = 95276230,95276171)

bbowman:
:clock1:

Frameworks/Foundation/NSURLResponse.mm
msft-Jeyaram: uh too much time spent with unique_ptrs :(

---
In reply to: [95277449](https://github.com/Microsoft/WinObjC/pull/1662#discussion_r95277449) [](ancestors = 95277449,95277280)

Frameworks/Foundation/_UTTypeInternal.mm
```diff
@@ -0,0 +1,585 @@
...
+ * @return UTI
+ */
+CFArrayRef _UTGetUTIsForFileNameExtension(CFStringRef fileNameExtension) {
+    if (fileNameExtension == NULL) {
+        return NULL;
+    }
+
+    CFArrayRef UTIs =
+        static_cast<CFArrayRef>(const_cast<void*>(CFDictionaryGetValue(s_fileExtensionToSystemUTIMap, (const void*)fileNameExtension)));
+    return UTIs ? CFArrayCreateCopy(kCFAllocatorDefault, UTIs) : NULL;
```
bviglietta: >CFArrayCreateCopy [](start = 18, length = 17)

If the caller owns a ref, shouldn't the name of this function be changed to indicate that?

It doesn't look like the caller in _UTTypeCreatePreferredIdentifierForTag is aware of this.

msft-Jeyaram: this is old code that was moved, clearly this is confusing, names need an update.

---
In reply to: [95278794](https://github.com/Microsoft/WinObjC/pull/1662#discussion_r95278794) [](ancestors = 95278794)

build/Foundation/dll/Foundation.def
msft-Jeyaram: and much more offline talks ended up in just compiling it as C++ to get over the error linking issues.

---
In reply to: [95277632](https://github.com/Microsoft/WinObjC/pull/1662#discussion_r95277632) [](ancestors = 95277632,95276230,95276171)

build/MobileCoreServices/lib/MobileCoreServicesLib.vcxproj
```diff
@@ -19,8 +19,12 @@
     </ProjectConfiguration>
   </ItemGroup>
   <ItemGroup>
-    <ClangCompile Include="..\..\..\Frameworks\MobileCoreServices\UTType.mm" />
-    <ClangCompile Include="..\..\..\Frameworks\MobileCoreServices\UTTypeInternalMap.mm" />
+    <ClangCompile Include="..\..\..\Frameworks\MobileCoreServices\UTType.mm">
+      <CompileAs Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">CompileAsCpp</CompileAs>
```
DHowett: Right now, these files are only being compiled as C++ for Debug+Win32!

Two suggestions:
1. Make this apply to all configurations.
2. Rename those files to .cpp to clear up the naming confusion.

Frameworks/MobileCoreServices/UTType.mm
```diff
@@ -14,134 +14,137 @@
...
-                               [static_cast<NSString*>(inConformingToUTI) UTF8String]);
+                               CFStringGetCStringPtr(inConformingToUTI, kCFStringEncodingUTF8));
     }
 
     CFComparisonResult result;
 
     result = CFStringCompare(inTagClass, kUTTagClassFilenameExtension, 0);
     if (result == kCFCompareEqualTo) {
-        CFArrayRef fileNameExtensions = _UTGetUTIsForFileNameExtension(inTag);
+        CFArrayRef fileNameExtensions = _UTCreateUTIsForFileNameExtension(inTag);
```
DHowett: Very worried about the memory management here! These Create functions return +1 references, and those references are immediately discarded and leaked.

Frameworks/MobileCoreServices/UTType.mm
```diff
@@ -14,134 +14,137 @@
...
         return fileNameExtensions ? static_cast<CFStringRef>(CFArrayGetValueAtIndex(fileNameExtensions, 0)) : NULL;
     }
 
     result = CFStringCompare(inTagClass, kUTTagClassMIMEType, 0);
     if (result == kCFCompareEqualTo) {
-        CFArrayRef mimeTypes = _UTGetUTIsForMIMEType(inTag);
+        CFArrayRef mimeTypes = _UTCreateUTIsForMIMEType(inTag);
```
DHowett: As above.

Frameworks/MobileCoreServices/UTType.mm
DHowett: Prefer `woc::unique_cf` for the array, and make sure that you return `CFRetain(...get value...)`. This function is supposed to return a +1, but it's returning a +0 owned by a leaked +1.

msft-Jeyaram: on point. I had to fix these since as you mentioned most of the memory management here was not well written before.

---
In reply to: [95506147](https://github.com/Microsoft/WinObjC/pull/1662#discussion_r95506147) [](ancestors = 95506147)

Issue #1613, opened 2016-12-30
Starboard DNSServiceRegister in failure log


jboich:
Hi.
Some of the customers have a problem with our app: the app crashes soon after launch. Unfortunately, I ca not reproduce this issue.
The following failure log indicates that it looks like the problem is inside WINOBJC bridge:

Frame	Image	Function	Offset
0	Starboard 	DNSServiceRegister	 0x5CE	
1	Starboard 	EbrMkdir	0x1B	
2	CoreFoundation	NS_mkdir 	0xC	
3	CoreFoundation	CFCopyExtensionForAbstractType	0x2B	
4	CoreFoundation	CFPreferencesDomainExists	0x533	
5	CoreFoundation	CFPreferencesDomainExists	0x3CA	
6	CoreFoundation	CFPreferencesStandardDomain	0xC6	
7	CoreFoundation	CFApplicationPreferencesSetStandardSearchList	0x3B	
8	CoreFoundation	CFStandardApplicationPreferences	0x99	
9	CoreFoundation	CFPreferencesCopyAppValue	0xB	
10	Foundation	NSSwapHostShortToBig	0x195B5	
11	MyAppName	-[CustomViewController setup:]	0x82	
12	MyAppName	-[CustomViewController setup]	0x12	
13	MyAppName	-[ViewController	0xDF	
14	MyAppName	-[ViewController setup]	0x30E	
15	MyAppName	-[MainViewController viewDidAppear:]	0x99	
16	UIKIT	UIOrientationFromString	0x29998	
17	UIKIT	UIImageJPEGRepresentation	0x6CCD	
18	UIKIT	UIImageJPEGRepresentation	0x7166	
19	MyAppName	-[AppDelegate application:didFinishLaunchingWithOptions:]	0x28A	
20	UIKIT	UIOrientationFromString	0x6CC	
21	Foundation	NSDeallocateObject	0xAFC	
22	Foundation	WinObjC_SetZombiesEnabled	0x3600	
23	UIKIT	UIViewControllerFirePageEvent	0xD4648	
24	SHCore	Microsoft::WRL::Details::RuntimeClass_Microsoft::WRL::Details::InterfaceList_CScalingInfoBase,Microsoft::WRL::Details::InterfaceList_IWindowMonitorChangeListener,Microsoft::WRL::Details::InterfaceList_ICurrentWindowChangeListener,Microsoft::WRL::Details::	0x84	
[...]

PR #1668, opened 2017-01-10
Jb/issue 1613

jboich:
I haven't write unit tests so far: I'm not sure what data to use to mock symbols in different languages as input. 
Using CFURLGetFileSystemRepresentation seems right for this issue

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1668)

### DIFF
--- a/Frameworks/CoreFoundation/Preferences.subproj/CFPreferences.c
+++ b/Frameworks/CoreFoundation/Preferences.subproj/CFPreferences.c
@@ -188,7 +188,40 @@ if (completePath) {
     // append "Preferences\" and make the CFURL
     CFStringAppend(completePath, CFSTR("Preferences\\"));
     url = CFURLCreateWithFileSystemPath(alloc, completePath, kCFURLWindowsPathStyle, true);
-    _CFCreateDirectory(CFStringGetCStringPtr(completePath, kCFStringEncodingUTF8)); // WINOBJC: ensure directory exists
+	
+	// Converting mutable string to characters array
+	// When CFStringGetCStringPtr() fails,
+	// it tries CFStringGetCString().
+	char *fullPath;
+	char outPath[512];
+
+	Boolean conversionResult;
+	CFStringEncoding encodingMethod = CFStringGetSystemEncoding();
+
+	// 1st try for system encoding
+	fullPath = (char*)CFStringGetCStringPtr(completePath, encodingMethod);
+	if (fullPath == NULL) {
+		// 2nd try for Japanese system
+		encodingMethod = kCFStringEncodingUTF8;
+		fullPath = (char*)CFStringGetCStringPtr(completePath, encodingMethod);
+	}
+
+	// for safer operation.
+	if (fullPath == NULL) {
+		CFIndex length = CFStringGetLength(completePath);
+		fullPath = (char *)malloc(length + 1);
+		conversionResult = CFStringGetCString(completePath, fullPath, length, kCFStringEncodingUTF8);
+
+		strcpy(outPath, fullPath);
+
+		free(fullPath);
+	}
+	else {
+		strcpy(outPath, fullPath);
+	}
+
+	_CFCreateDirectory(outPath);
+
     CFRelease(completePath);
 }
 
--- a/Frameworks/CoreFoundation/Preferences.subproj/CFPreferences.c
+++ b/Frameworks/CoreFoundation/Preferences.subproj/CFPreferences.c
@@ -182,52 +182,24 @@ static CFURLRef _preferencesDirectoryForUserHostSafetyLevel(CFStringRef userName
 #if DEPLOYMENT_TARGET_WINDOWS
 
 CFURLRef url = NULL;
+Boolean success = false;
 
 CFMutableStringRef completePath = _CFCreateApplicationRepositoryPath(alloc, CSIDL_APPDATA);
 if (completePath) {
     // append "Preferences\" and make the CFURL
     CFStringAppend(completePath, CFSTR("Preferences\\"));
     url = CFURLCreateWithFileSystemPath(alloc, completePath, kCFURLWindowsPathStyle, true);
-	
-	// Converting mutable string to characters array
-	// When CFStringGetCStringPtr() fails,
-	// it tries CFStringGetCString().
-	char *fullPath;
-	char outPath[512];
 
-	Boolean conversionResult;
-	CFStringEncoding encodingMethod = CFStringGetSystemEncoding();
-
-	// 1st try for system encoding
-	fullPath = (char*)CFStringGetCStringPtr(completePath, encodingMethod);
-	if (fullPath == NULL) {
-		// 2nd try for Japanese system
-		encodingMethod = kCFStringEncodingUTF8;
-		fullPath = (char*)CFStringGetCStringPtr(completePath, encodingMethod);
+	char cPath[CFMaxPathSize];
+	if (CFURLGetFileSystemRepresentation(url, true, (unsigned char *)cPath, CFMaxPathSize)) {
+		success = _CFCreateDirectory(cPath);
 	}
-
-	// for safer operation.
-	if (fullPath == NULL) {
-		CFIndex length = CFStringGetLength(completePath);
-		fullPath = (char *)malloc(length + 1);
-		conversionResult = CFStringGetCString(completePath, fullPath, length, kCFStringEncodingUTF8);
-
-		strcpy(outPath, fullPath);
-
-		free(fullPath);
-	}
-	else {
-		strcpy(outPath, fullPath);
-	}
-
-	_CFCreateDirectory(outPath);
 
     CFRelease(completePath);
 }
 
-
 // Can't find a better place?  Home directory then?
-if (url == NULL)
+if (!success)
     url = CFCopyHomeDirectoryURLForUser((userName == kCFPreferencesCurrentUser) ? NULL : userName);
 
 return url;


Frameworks/CoreFoundation/Preferences.subproj/CFPreferences.c
```diff
@@ -182,19 +182,24 @@ static CFURLRef _preferencesDirectoryForUserHostSafetyLevel(CFStringRef userName
...
 CFMutableStringRef completePath = _CFCreateApplicationRepositoryPath(alloc, CSIDL_APPDATA);
 if (completePath) {
     // append "Preferences\" and make the CFURL
     CFStringAppend(completePath, CFSTR("Preferences\\"));
     url = CFURLCreateWithFileSystemPath(alloc, completePath, kCFURLWindowsPathStyle, true);
-    _CFCreateDirectory(CFStringGetCStringPtr(completePath, kCFStringEncodingUTF8)); // WINOBJC: ensure directory exists
+
+	char cPath[CFMaxPathSize];
+	if (CFURLGetFileSystemRepresentation(url, true, (unsigned char *)cPath, CFMaxPathSize)) {
+		success = _CFCreateDirectory(cPath);
```
bbowman: >		 [](start = 0, length = 2)

Please make sure to run clang-format before submitting any code. In this case you have tabs that should be spaces. Coding conventions are a sticking point for a lot of people so we make it pretty easy to ensure the code is conformant ([https://github.com/Microsoft/WinObjC/wiki/ClangFormat](https://github.com/Microsoft/WinObjC/wiki/ClangFormat)) for more info.

Frameworks/CoreFoundation/Preferences.subproj/CFPreferences.c
```diff
@@ -182,19 +182,24 @@ static CFURLRef _preferencesDirectoryForUserHostSafetyLevel(CFStringRef userName
...
 
 CFMutableStringRef completePath = _CFCreateApplicationRepositoryPath(alloc, CSIDL_APPDATA);
 if (completePath) {
     // append "Preferences\" and make the CFURL
     CFStringAppend(completePath, CFSTR("Preferences\\"));
     url = CFURLCreateWithFileSystemPath(alloc, completePath, kCFURLWindowsPathStyle, true);
-    _CFCreateDirectory(CFStringGetCStringPtr(completePath, kCFStringEncodingUTF8)); // WINOBJC: ensure directory exists
+
+	char cPath[CFMaxPathSize];
+	if (CFURLGetFileSystemRepresentation(url, true, (unsigned char *)cPath, CFMaxPathSize)) {
```
bbowman: Please make sure any and all changes to CoreFoundation are annotated with a // WINOBJC:  
comment. This helps more easily pinpoint differences from the reference platform implementation.

bbowman:
You have misspelled words in your commit message.

I also don't quite understand the issue and what the fix is here. Can you elaborate?

Frameworks/CoreFoundation/Preferences.subproj/CFPreferences.c
```diff
@@ -182,19 +182,24 @@ static CFURLRef _preferencesDirectoryForUserHostSafetyLevel(CFStringRef userName
 #if DEPLOYMENT_TARGET_WINDOWS
 
 CFURLRef url = NULL;
+Boolean success = false;
 
 CFMutableStringRef completePath = _CFCreateApplicationRepositoryPath(alloc, CSIDL_APPDATA);
 if (completePath) {
     // append "Preferences\" and make the CFURL
     CFStringAppend(completePath, CFSTR("Preferences\\"));
     url = CFURLCreateWithFileSystemPath(alloc, completePath, kCFURLWindowsPathStyle, true);
```
bbowman: >completePath [](start = 47, length = 12)

what is this used for now?

Frameworks/CoreFoundation/Preferences.subproj/CFPreferences.c
```diff
@@ -182,19 +182,24 @@ static CFURLRef _preferencesDirectoryForUserHostSafetyLevel(CFStringRef userName
...
-    _CFCreateDirectory(CFStringGetCStringPtr(completePath, kCFStringEncodingUTF8)); // WINOBJC: ensure directory exists
+
+	char cPath[CFMaxPathSize];
```
bbowman: >CFMaxPathSize [](start = 12, length = 13)

what is this value and is it sufficient for all cases?

bbowman:
Please add unit tests. I'm not sure what the confusion is for testing? 
My usual strategy for bug fixes:
1. Write a test that encapsulates the current bug condition. This should fail.
2. Fix the product code.
3. Verify test now passes.

In terms of mocking, we have support for using MOCK_CLASS and similar macros that allow for lambda replacement of function bodies at runtime but that is usually used with an interface pattern to substitute in a mock class in place of the production layers above and below the class under test.

In this case, it seems like you want to verify the URL returned by the function you modified? The key consideration here would be the file system differences on other people's machines and the access rights of the test code. A generally safe assumption is that you can access resources in the current test directory and can access / mess with paths in well known locations like TEMP, etc. Please make sure that the tests use good RAII / similar patterns to ensure consistent cleanup of any machine state that is modified.

bbowman:
:clock1:

DHowett-MSFT:
Note: When we merge, we should update the commit message / take one from the commits in this PR instead of the PR title.

Frameworks/CoreFoundation/Preferences.subproj/CFPreferences.c
jboich: > Can you just manually change your tabs to spaces? 

sure, will do

Frameworks/CoreFoundation/Preferences.subproj/CFPreferences.c
jboich: > Please make sure any and all changes to CoreFoundation are annotated with a // WINOBJC:

will do, thank you

Frameworks/CoreFoundation/Preferences.subproj/CFPreferences.c
jboich: Hi, @bbowman .

> I also don't quite understand the issue and what the fix is here. Can you elaborate?

The changes I made are fix for the issue [#1613 ](https://github.com/Microsoft/WinObjC/issues/1613). In my last comment to the issue 1613 I mentioned that I found the [solution](https://jongampark.wordpress.com/2009/06/24/be-careful-with-cfstringgetcstringptr-and-its-companion-calls/), which I applied in "_preferencesDirectoryForUserHostSafetyLevel" in CFPreferences.c  . But then I changed it to the implementation, which I submitted as PR, because I thought it was more relevant.

jboich:
Hi, @bbowman .

> Please add unit tests. I'm not sure what the confusion is for testing?

I think that I am not familiar with the bridge project as good as you are, so I was asking for advice :)

> it seems like you want to verify the URL returned by the function you modified?

Because changes, that I made, are related to issue #1613 I'd like to verify that if completePath contains different set of characters in different languages, then 
1. CFURLGetFileSystemRepresentation returns true and
2. directory creates successfully by _CFCreateDirectory using character buffer, which filled in by CFURLGetFileSystemRepresentation function.

I will provide you an example of data I mean for 'completePath' later (I don't have them at hand right now)

Thanks for all your feedback!

Frameworks/CoreFoundation/Preferences.subproj/CFPreferences.c
jboich: Hi, @bbowman .
CFMaxPathSize defined in CFInternal.h as following:
`/* Buffer size for file pathname */
#if DEPLOYMENT_TARGET_WINDOWS
#define CFMaxPathSize ((CFIndex)262)
#define CFMaxPathLength ((CFIndex)260)`

I do not know why such value was chosen for Windows target. CFMaxPathSize definition used in many places of CoreFoundation.

jboich:
Hi, @bbowman .
So an example of data I get in completePath looks like this
`"C:/Users/<name of user>/AppData/Local/Packages/<package name>/LocalState/Preferences"`

<name of user> could be in different languages and may contain space. 
I point to different languages, because if user name is in English and contain space, issue #1613 did not take place.

<package name> is from dev center and it is in specified format and should not cause any problems when url converts to char array.

Frameworks/CoreFoundation/Preferences.subproj/CFPreferences.c
bbowman: >CFURLGetFileSystemRepresentation [](start = 5, length = 32)

looking at the implementation for this, it appears that it can return false if the buffer is too small as well. Not a lot we can do here about that though. I think the MaxPath thing is fine for now but it does worry me a for very long names.

bbowman:
The likely case here is simply a UTF encoding issue where we are getting a different CFstring encoding resulting in a nullptr return from CFStringGetPtr or whatever that call was. 
This is probably the interesting bit to test. In terms of testing, its a bit annoying because its a file scoped static function that we want to test. It does look like though that the inner most failing function we could test would be:

_CFPreferencesCreateDomainList

So i'd suggest making a new test in WinObjC\tests\unittests\CoreFoundation\ for CFPreferences that calls _CFPreferencesCreateDomainList and checks the result when the username field is something weird (foreign language and with a space). It would be good to write this test and try it without your fix to see that it fails and then make sure it passes with it.

Thanks!
-Brian


---
In reply to: [271939958](https://github.com/Microsoft/WinObjC/pull/1668#issuecomment-271939958) [](ancestors = 271939958)

jboich:
Hi @bbowman. Thanks for feedback!

>  that calls _CFPreferencesCreateDomainList and checks the result when the username field is something weird 

Assigning different values to username of **CFPreferencesCreateDomainList**  method won't allow us to test what we want, because value of username field is not used to create a path, which will be used then to create url and directory.

Method `CF_PRIVATE CFArrayRef  _CFPreferencesCreateDomainList(CFStringRef  userName, CFStringRef  hostName)`  calls  `CFURLRef prefDir = _preferencesDirectoryForUserHost(userName, hostName);`


_`preferencesDirectoryForUserHost `calls `_preferencesDirectoryForUserHostSafetyLevel`


And if you take a look at the implementation of **_preferencesDirectoryForUserHostSafetyLevel** you will see that username parameter is not taking any part in **completePath** :
`CFMutableStringRef completePath = _CFCreateApplicationRepositoryPath(alloc, CSIDL_APPDATA);`


And this guy ` _CFCreateApplicationRepositoryPath(alloc, CSIDL_APPDATA);` returns application data local folder (`GetAppDataPath(true)`)

As far as I understand I need to simulate data returned by **GetAppDataPath** . Could you give any advice of how to do this?

Kind regards,
Julia

bbowman:
@jboich Sorry you are correct. It does appear that the user name (which we believe is the issue we are fixing) comes in via that **GetAppDataPath**  call.

This makes testing even a bit more difficult which is unfortunate. We have the capability to mock out the WinRT calls that are being made there but we will need to modify the CoreFoundation library and tests to make it more testable. This is nontrivial. Please file a separate issue for that and we can write that test at a later point. 

Otherwise I think we are ok accepting your change pending CI build and sign off (I'll go ahead and schedule a build).

Thanks!

jboich:
Hi, @bbowman . Thanks for getting back to me with response.
I did many manual tests and noticed that CFURLGetFileSystemRepresentation sometimes returns odd results. I did look close in the code and noticed that `_CFCreateApplicationRepositoryPath` returns path in UTF16 encoding and `CFURLGetFileSystemRepresentation` uses `CFStringFileSystemEncoding()`.  
Can I make another PR with the suggestion of the fix?

bbowman:
@jboich That would be great. You can submit a new commit on this one if you feel like it is still part of the same issue seen here.

Issue #1672, opened 2017-01-10
UIView's dealloc isn't resilient to subclasses stomping 'self' in their custom init implementations


jaredhms:
This code in a custom UIView subclass causes UIView to AV in dealloc:

```
- (instancetype)init
{
    NSArray *nib = [[NSBundle mainBundle] loadNibNamed:@"MyViewCell" owner:self options:nil];
    self = [nib objectAtIndex:0];
    return self;
}
```

The fix is to not do any work in UIView's dealloc if we were never init'ed.  Thanks, @DHowett-MSFT your help in tracking this down.

jaredhms:
Fixed by #1674.

PR #1674, opened 2017-01-10
Fix for #1672.  Don't uninitialize in UIView's dealloc if we never initialized

jaredhms:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1674)


yiyang-msft:
:shipit:

Issue #869, opened 2016-08-10
NSOperationQueue is suboptimal (but it does work)


DHowett-MSFT:
- Busy wait if an operation with dependencies is added, but the dependencies are not _yet_ added.
- Queue does not subscribe to notifications for operation status.
  - `isReady` could help avoid the busy wait.
- We only support three priority tiers out of the five which are defined.
- The magic inside `RunOperationFromLists` is arcane and black.
  - Takes the source list. Drops it down, flips it and reverses it.
  - That new list is the "pending" list. Operations are run preferentially from the pending list.
  - The pending list is unceremoniously destroyed (via `memset`) at the start of the next queue thread iteration.

PR #1676, opened 2017-01-10
Redo locking and ref counting in NSOperationQueue to be much safer. A…

bbowman:
…dditionally document code to try to understand the intention. NOTE: A total rewrite is still needed due to a number of deficienies in the code.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1676)

### DIFF
--- a/Frameworks/Foundation/NSOperationQueue.mm
+++ b/Frameworks/Foundation/NSOperationQueue.mm
@@ -38,6 +38,11 @@ OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 
 static const wchar_t* TAG = L"NSOperationQueue";
 
+// NOTE: NSOperationQueuePriv is totally not needed. This class is just acting as the logic encapsulation of state and data ownership for
+// the
+// operation queue class. The ObjectiveC class itself is perfectly capable of storing and using all this information. This is simply a
+// vestige
+// of the previous implementation.
 class NSOperationQueuePriv {
 public:
     NSOperationQueuePriv(NSOperationQueue* parent) : _thread(nullptr), _maxConcurrentOperationCount(1), _isSuspended(NO) {
@@ -74,7 +79,7 @@ public:
         }
 
         // Now that a candidate has been determined, drop the lock so that it can potientially run.
-        // NOTE: a [operation start] call will potienitally run the operation on that this thread.
+        // NOTE: a [operation start] call will potienitally run the operation on this thread.
 
         if (candidateOperation != nil) {
             // Valid candidate to start. Make sure its actually ready for liftoff.
@@ -142,6 +147,11 @@ public:
         if (!toReturn) {
             // No work left to do. Clear out the thread here. This signals that a new work thread is needed to be created and
             // current one can end its execution.
+            //
+            // NOTE: its awfully weird to have a "getter" have a side effect like nil'ing out _thread. This is needed because the
+            // decision to stop the work thread is made based on not having more work and this means that the thread needs restarted
+            // on the next go around. A real fix is to not manage the thread manually and use a threadpool or other mechanism to run the
+            // operations.
             _thread = nil;
         }
 
@@ -153,7 +163,7 @@ public:
             std::lock_guard<std::mutex> lock(_lock);
 
             StrongId<NSOperation> strongOperation = operation;
-            _operationsHoldingArea[priority].emplace_back(strongOperation);
+            _operationsHoldingArea[priority].emplace_back(std::move(strongOperation));
             if (_thread == nil || _thread.get().finished) {
                 _thread.attach([[NSThread alloc] initWithTarget:queue selector:selector object:nil]);
                 [_thread start];
@@ -285,7 +295,7 @@ private:
         // NOTE: This may be insufficient / racy. This only checks the holding area and not "on deck" items.
         // Its possible that the check is made right after a transfer to on deck which would incorrectly report
         // all done when some items may go back into holding and certainly all of them aren't completed. This particularly
-        // would affect WaitUntilAllOperationsAreFinished and
+        // would affect WaitUntilAllOperationsAreFinished.
         for (int i = 0; i < NSOperationQueuePriority_Count; i++) {
             if (_operationsHoldingArea[i].empty()) {
                 return TRUE;
@@ -315,10 +325,6 @@ private:
     // special care is needed to ensure that no re-entrant functions are used while under the lock.
     std::mutex _lock;
 };
-
-extern pthread_key_t g_currentDispatchQueue;
-
-static id _mainQueue;
 
 @interface NSOperationQueue () {
     std::unique_ptr<NSOperationQueuePriv> _priv;
--- a/Frameworks/Foundation/NSOperationQueue.mm
+++ b/Frameworks/Foundation/NSOperationQueue.mm
@@ -30,167 +30,320 @@ OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 #import "LoggingNative.h"
 #import "NSOperationInternal.h"
 
+#include <deque>
+#include <mutex>
+#include <condition_variable>
+#include <algorithm>
+#include <memory>
+
 static const wchar_t* TAG = L"NSOperationQueue";
 
-typedef void* gpointer;
+class NSOperationQueuePriv {
+public:
+    NSOperationQueuePriv(NSOperationQueue* parent) : _thread(nullptr), _maxConcurrentOperationCount(1), _isSuspended(NO) {
+        _name.attach([[NSString alloc] initWithFormat:@"NSOperationQueue %08x", parent]);
+    }
 
-struct NSAtomicListNode {
-    struct NSAtomicListNode* next;
-    StrongId<NSObject> elt;
-};
-typedef struct NSAtomicListNode* NSAtomicListRef;
+    NSString* Name() {
+        std::lock_guard<std::mutex> lock(_lock);
+        return _name;
+    }
 
-struct NSOperationQueuePriv {
-    NSAtomicListRef myQueues[NSOperationQueuePriority_Count];
-    NSAtomicListRef queues[NSOperationQueuePriority_Count];
+    void SetName(NSString* name) {
+        std::lock_guard<std::mutex> lock(_lock);
+        _name.attach([name copy]);
+    }
 
-    NSThread* _thread;
-    pthread_mutex_t _threadRunningLock;
+    BOOL RunOperationFromLists(unsigned int priority) {
+        StrongId<NSOperation> candidateOperation;
+
+        {
+            std::lock_guard<std::mutex> lock(_lock);
+            if (_candidateOperationsToRunNext[priority].empty()) {
+                // No operations "on deck". Go pull some more in from the holding area.
+                // Holding area lists are in LIFO order, but we want to execute operations in the order they were enqueued
+                // so we reverse the list before we do anything with it.
+                _candidateOperationsToRunNext[priority].swap(_operationsHoldingArea[priority]);
+                std::reverse(_candidateOperationsToRunNext[priority].begin(), _candidateOperationsToRunNext[priority].end());
+            }
+
+            if (!_candidateOperationsToRunNext[priority].empty()) {
+                candidateOperation = _candidateOperationsToRunNext[priority].back();
+                _candidateOperationsToRunNext[priority].pop_back();
+            }
+        }
+
+        // Now that a candidate has been determined, drop the lock so that it can potientially run.
+        // NOTE: a [operation start] call will potienitally run the operation on that this thread.
+
+        if (candidateOperation != nil) {
+            // Valid candidate to start. Make sure its actually ready for liftoff.
+            if ([candidateOperation isReady]) {
+                {
+                    std::lock_guard<std::mutex> lock(_lock);
+
+                    // Start it and set to current operation so that accounting methods (operationCount and friends)
+                    // can still include this guy even though it is no longer in any list.
+                    // NOTE: this (incorrectly) assumes that only one operation is ever running at a time. This is easily
+                    // seen to be false whe considering asynchronous operations and any values of _maxConcurrentOperationCount
+                    // other than 1 (currently ignored but should be added in the future).
+                    _curOperation = candidateOperation;
+                }
+
+                [candidateOperation start];
+
+                {
+                    std::lock_guard<std::mutex> lock(_lock);
+
+                    // Clear the current operation now that start is finished.
+                    // NOTE: the assumption (inccorect) here is that the operation is now finished and
+                    // we no longer need any accounting of the current operation.
+                    // This is wrong because an operation could be a new fangled asynchronous
+                    // one and it could finish muc later (not synchronously executed with [operation start]).
+                    _curOperation = nil;
+                }
+            } else {
+                {
+                    std::lock_guard<std::mutex> lock(_lock);
+                    // Not yet ready. Demote it to the holding area. Will pick it up again later.
+                    _operationsHoldingArea[priority].emplace_back(std::move(candidateOperation));
+                }
+
+                // We claim that we have done some work here. This keeps the queue pumping and
+                // will eventually bring this item back out for consideration and will continue
+                // through this list for anyother ready operations.
+                return TRUE;
+            }
+        }
+
+        return candidateOperation != nil;
+    }
+
+    void WaitUntilNotSuspended() {
+        std::unique_lock<std::mutex> lock(_lock);
+        while (_isSuspended) {
+            _suspendedCondition.wait(lock);
+        }
+        lock.unlock();
+    }
+
+    void ClearCandidateLists() {
+        std::lock_guard<std::mutex> lock(_lock);
+        for (int i = 0; i < NSOperationQueuePriority_Count; i++) {
+            _candidateOperationsToRunNext[i].clear();
+        }
+    }
+
+    BOOL HasMoreWork() {
+        std::lock_guard<std::mutex> lock(_lock);
+
+        BOOL toReturn = _HasMoreWorkNoLock();
+
+        if (!toReturn) {
+            // No work left to do. Clear out the thread here. This signals that a new work thread is needed to be created and
+            // current one can end its execution.
+            _thread = nil;
+        }
+
+        return toReturn;
+    }
+
+    void AddOperation(NSOperation* operation, unsigned int priority, NSOperationQueue* queue, SEL selector) {
+        {
+            std::lock_guard<std::mutex> lock(_lock);
+
+            StrongId<NSOperation> strongOperation = operation;
+            _operationsHoldingArea[priority].emplace_back(strongOperation);
+            if (_thread == nil || _thread.get().finished) {
+                _thread.attach([[NSThread alloc] initWithTarget:queue selector:selector object:nil]);
+                [_thread start];
+            }
+        }
+
+        _workAvailableCondition.notify_one();
+    }
+
+    void SignalAllWorkDone() {
+        _allWorkDoneCondition.notify_one();
+    }
+
+    void WaitUntilAllOperationsAreFinished() {
+        std::unique_lock<std::mutex> lock(_lock);
+
+        while (_HasMoreWorkNoLock()) {
+            _allWorkDoneCondition.wait(lock);
+        }
+    }
+
+    void SetMaxConcurrentOperationCount(NSInteger count) {
+        std::lock_guard<std::mutex> lock(_lock);
+        _maxConcurrentOperationCount = count;
+    }
+
+    NSInteger MaxConcurrentOperationCount() {
+        std::lock_guard<std::mutex> lock(_lock);
+        return _maxConcurrentOperationCount;
+    }
+
+    NSArray<NSOperation*>* GetOperations() {
+        std::lock_guard<std::mutex> lock(_lock);
+
+        id ret = [NSMutableArray array];
+
+        if (_curOperation != nil) {
+            [ret addObject:_curOperation];
+        }
+
+        for (int i = 0; i < NSOperationQueuePriority_Count; i++) {
+            if (!_operationsHoldingArea[i].empty()) {
+                for (const auto& op : _operationsHoldingArea[i]) {
+                    [ret addObject:op];
+                }
+            }
+
+            if (!_candidateOperationsToRunNext[i].empty()) {
+                for (const auto& op : _candidateOperationsToRunNext[i]) {
+                    [ret addObject:op];
+                }
+            }
+        }
+
+        return ret;
+    }
+
+    unsigned int GetOperationCount() {
+        std::lock_guard<std::mutex> lock(_lock);
+
+        unsigned int ret = 0;
+
+        if (_curOperation != nil) {
+            ret++;
+        }
+
+        for (int i = 0; i < NSOperationQueuePriority_Count; i++) {
+            ret += _operationsHoldingArea[i].size();
+            ret += _candidateOperationsToRunNext[i].size();
+        }
+
+        return ret;
+    }
+
+    void CancelAllOperations() {
+        std::lock_guard<std::mutex> lock(_lock);
+
+        if (_curOperation != nil) {
+            [_curOperation cancel];
+        }
+
+        for (int i = 0; i < NSOperationQueuePriority_Count; i++) {
+            if (!_operationsHoldingArea[i].empty()) {
+                for (const auto& op : _operationsHoldingArea[i]) {
+                    [op cancel];
+                }
+            }
+
+            if (!_candidateOperationsToRunNext[i].empty()) {
+                for (const auto& op : _candidateOperationsToRunNext[i]) {
+                    [op cancel];
+                }
+            }
+        }
+    }
+
+    void Resume() {
+        BOOL shouldBroadcast = FALSE;
+        {
+            std::lock_guard<std::mutex> lock(_lock);
+            if (_isSuspended) {
+                _isSuspended = FALSE;
+                shouldBroadcast = TRUE;
+            }
+        }
+
+        if (shouldBroadcast) {
+            _suspendedCondition.notify_all();
+        }
+    }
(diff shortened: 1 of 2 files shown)


Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -472,109 +477,63 @@ - (void)setSuspended:(BOOL)suspend {
...
- @Status Interoperable
-*/
 + (id)currentQueue {
     if ([NSThread isMainThread]) {
         return [self mainQueue];
     }
 
+    // NOTE: this is obviously wrong as self is a META class not an instance of the class.
+    // Likely there needs to be thread local storage queue or perhaps just only return the mainQueue
     assert(0);
```
rajsesh-msft: >assert [](start = 4, length = 6)

should nuke assert() here?

bbowman: probably but I was just going to let whoever is redo'ing the whole thing do that.

---
In reply to: [95446765](https://github.com/Microsoft/WinObjC/pull/1676#discussion_r95446765) [](ancestors = 95446765)

bbowman: its still wrong without the assert and I'm not confident in a quick fix for a queue per thread or anything like that.

---
In reply to: [95447067](https://github.com/Microsoft/WinObjC/pull/1676#discussion_r95447067) [](ancestors = 95447067,95446765)

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -30,167 +30,320 @@
...
+
+        if (candidateOperation != nil) {
+            // Valid candidate to start. Make sure its actually ready for liftoff.
+            if ([candidateOperation isReady]) {
+                {
+                    std::lock_guard<std::mutex> lock(_lock);
+
+                    // Start it and set to current operation so that accounting methods (operationCount and friends)
+                    // can still include this guy even though it is no longer in any list.
+                    // NOTE: this (incorrectly) assumes that only one operation is ever running at a time. This is easily
```
aballway: Should make a task w/ list of all caveats on this

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -30,167 +30,320 @@
...
+        {
+            std::lock_guard<std::mutex> lock(_lock);
+            if (_isSuspended) {
+                _isSuspended = FALSE;
+                shouldBroadcast = TRUE;
+            }
+        }
+
+        if (shouldBroadcast) {
+            _suspendedCondition.notify_all();
```
aballway: don't want to be locked to notify_all?

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -330,132 +424,43 @@ - (void)addOperations:(id)operations waitUntilFinished:(BOOL)wait {
 */
 - (void)setMaxConcurrentOperationCount:(NSInteger)count {
     UNIMPLEMENTED();
-    priv->_maxConcurrentOperationCount = count;
+    _priv->SetMaxConcurrentOperationCount(count);
```
aballway: these are implemented?

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -30,167 +30,320 @@
...
+#include <algorithm>
+#include <memory>
+
 static const wchar_t* TAG = L"NSOperationQueue";
 
-typedef void* gpointer;
+class NSOperationQueuePriv {
+public:
+    NSOperationQueuePriv(NSOperationQueue* parent) : _thread(nullptr), _maxConcurrentOperationCount(1), _isSuspended(NO) {
+        _name.attach([[NSString alloc] initWithFormat:@"NSOperationQueue %08x", parent]);
```
DHowett-MSFT: nit: could move this to the initializer list and use stringWithFormat.

include/Foundation/NSOperationQueue.h
```diff
@@ -28,6 +28,8 @@
 
 enum { NSOperationQueueDefaultMaxConcurrentOperationCount = -1 };
 
+// NOTE: This value appears to be a private detail that leaked into the public header. Unfortunately it is
+// difficult to know if it is used externally. At any rate it is wrong and should be 5.
```
aballway: this is amazing #Closed

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -30,167 +30,320 @@
...
-    BOOL isSuspended;
+            if (!_candidateOperationsToRunNext[priority].empty()) {
+                candidateOperation = _candidateOperationsToRunNext[priority].back();
+                _candidateOperationsToRunNext[priority].pop_back();
+            }
+        }
 
-    id _name;
+        // Now that a candidate has been determined, drop the lock so that it can potientially run.
+        // NOTE: a [operation start] call will potienitally run the operation on that this thread.
```
DHowett-MSFT: nit: "on that this"

DHowett-MSFT:
The other issue on this is #869.

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -30,167 +30,320 @@
...
+    BOOL HasMoreWork() {
+        std::lock_guard<std::mutex> lock(_lock);
 
-void NSAtomicListReverse(NSAtomicListRef* listPtr) {
-    struct NSAtomicListNode* cur = *listPtr;
-    struct NSAtomicListNode* prev = NULL;
-    struct NSAtomicListNode* next = NULL;
+        BOOL toReturn = _HasMoreWorkNoLock();
+
+        if (!toReturn) {
```
DHowett-MSFT: worry: this "getter" has a side effect.

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -30,167 +30,320 @@
...
-        cur->next = prev;
+    void AddOperation(NSOperation* operation, unsigned int priority, NSOperationQueue* queue, SEL selector) {
+        {
+            std::lock_guard<std::mutex> lock(_lock);
 
-        if (next) {
-            prev = cur;
-            cur = next;
+            StrongId<NSOperation> strongOperation = operation;
+            _operationsHoldingArea[priority].emplace_back(strongOperation);
```
DHowett-MSFT: Could this be done in one fell swoop? Or should you std::move into the operationsHoldingArea?

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -30,167 +30,320 @@
...
-    *listPtr = node->next;
+    void WaitUntilAllOperationsAreFinished() {
+        std::unique_lock<std::mutex> lock(_lock);
 
-    NSObject* elt = [[node->elt retain] autorelease];
-    delete node;
-    return elt;
-}
+        while (_HasMoreWorkNoLock()) {
+            _allWorkDoneCondition.wait(lock);
```
DHowett-MSFT: nit: `condition.wait(lock, [](){ predicate! })`

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -30,167 +30,320 @@
...
+#include <condition_variable>
+#include <algorithm>
+#include <memory>
+
 static const wchar_t* TAG = L"NSOperationQueue";
 
-typedef void* gpointer;
+class NSOperationQueuePriv {
+public:
+    NSOperationQueuePriv(NSOperationQueue* parent) : _thread(nullptr), _maxConcurrentOperationCount(1), _isSuspended(NO) {
```
DHowett-MSFT: Note for eventual future implementer: The C++ class is providing no benefit. Objective-C++ classes can contain C++ things, and these aren't being custom-initialized.

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -30,167 +30,320 @@
...
+                    [ret addObject:op];
+                }
+            }
 
-@implementation NSOperationQueue
-static void ClearList(NSAtomicListRef* listPtr) {
-    for (int i = 0; i < NSOperationQueuePriority_Count; i++) {
-        while (NSAtomicListPop(&listPtr[i])) {
-            ;
+            if (!_candidateOperationsToRunNext[i].empty()) {
```
DHowett-MSFT: nit: Unnecessary? It should be just as many checks as iterating over an empty list.

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -30,167 +30,320 @@
...
+            [ret addObject:_curOperation];
+        }
 
-@interface NSOperationQueue () {
-    struct NSOperationQueuePriv* priv;
-    dispatch_queue_t _completionQueue;
-}
-@end
+        for (int i = 0; i < NSOperationQueuePriority_Count; i++) {
+            if (!_operationsHoldingArea[i].empty()) {
```
DHowett-MSFT: As below

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -200,15 +353,11 @@ - (NSOperationQueue*)_doMainWork {
...
-            [priv->suspendedCondition unlock];
-
-            if (RunOperationFromLists(&priv->myQueues[i], &priv->queues[i], &priv->curOperation)) {
-                didWork = TRUE;
-            }
+            // Wait until the operation queue is no longer suspended.
+            // NOTE: if an operation queue becomes suspended after this point, the "in flight" operation
+            // will not be suspended, it will continue as if the queue is still running.
+            _priv->WaitUntilNotSuspended();
+            didWork |= _priv->RunOperationFromLists(i);
```
DHowett-MSFT: I'm not certain a binary OR here is valid

DHowett-MSFT: prefer `= x || ` for clarity/operator integrity?

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -220,65 +369,19 @@ - (NSOperationQueue*)_doMainWork {
...
-        priv = new NSOperationQueuePriv();
-
-        priv->_maxConcurrentOperationCount = 1;
-        priv->workAvailable = [[NSCondition alloc] init];
-        priv->suspendedCondition = [[NSCondition alloc] init];
-        priv->allWorkDone = [[NSCondition alloc] init];
-        priv->isSuspended = 0;
-
         _completionQueue = dispatch_queue_create("NSOperation finish queue", nullptr);
+        _priv = std::make_unique<NSOperationQueuePriv>(self);
```
DHowett-MSFT: If the no-argument constructor on NSOperationQueuePriv does what we need, we don't need to indirect through a  unique_ptr and can embed it directly.

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -472,109 +477,63 @@ - (void)setSuspended:(BOOL)suspend {
...
-/**
- @Status Interoperable
-*/
 + (id)currentQueue {
     if ([NSThread isMainThread]) {
         return [self mainQueue];
     }
 
+    // NOTE: this is obviously wrong as self is a META class not an instance of the class.
+    // Likely there needs to be thread local storage queue or perhaps just only return the mainQueue
```
DHowett-MSFT: LOL. likely -> absolutely!

Frameworks/Foundation/NSOperationQueue.mm
DHowett-MSFT: ah, i see. name should perhaps be injected to make it a 0-arg ctor.

Frameworks/Foundation/NSOperationQueue.mm
DHowett-MSFT: implemented but unused.

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -30,167 +30,320 @@
...
+    BOOL RunOperationFromLists(unsigned int priority) {
+        StrongId<NSOperation> candidateOperation;
+
+        {
+            std::lock_guard<std::mutex> lock(_lock);
+            if (_candidateOperationsToRunNext[priority].empty()) {
+                // No operations "on deck". Go pull some more in from the holding area.
+                // Holding area lists are in LIFO order, but we want to execute operations in the order they were enqueued
+                // so we reverse the list before we do anything with it.
+                _candidateOperationsToRunNext[priority].swap(_operationsHoldingArea[priority]);
```
rajsesh-msft: ugh.  why can't this list be FIFO instead?

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -30,167 +30,320 @@
...
+    // NOTE: a normal (as opposed to recursive) mutex is used here to interop with condition variable.
+    // This helps simplify the locking in this class in that fewer locks are used so the overall ownership
+    // of the state this class owns is encapsulated under the single lock. However, this also means that
+    // special care is needed to ensure that no re-entrant functions are used while under the lock.
+    std::mutex _lock;
+};
+
+extern pthread_key_t g_currentDispatchQueue;
+
+static id _mainQueue;
```
bviglietta: This is unused and shadowed in the mainQueue function.

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -472,109 +477,63 @@ - (void)setSuspended:(BOOL)suspend {
...
 
 /**
  @Status Interoperable
 */
 + (id)mainQueue {
-    if (_mainQueue == nil) {
-        _mainQueue = [[self alloc] _initMainThread];
-    }
-
+    static StrongId<NSOperationQueue> _mainQueue = [[[self alloc] init] autorelease];
```
bviglietta: >autorelease [](start = 72, length = 11)

Should this really be released?

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -30,167 +30,320 @@
...
+    // NOTE: this method is intended to be run under the lock but does not take the lock itself.
+    BOOL _HasMoreWorkNoLock() {
+        if (_curOperation != nil) {
             return TRUE;
         }
+
+        // NOTE: This may be insufficient / racy. This only checks the holding area and not "on deck" items.
+        // Its possible that the check is made right after a transfer to on deck which would incorrectly report
+        // all done when some items may go back into holding and certainly all of them aren't completed. This particularly
+        // would affect WaitUntilAllOperationsAreFinished and
```
bviglietta: >and [](start = 58, length = 3)

I'm on the edge of my seat.

Frameworks/Foundation/NSOperationQueue.mm
DHowett-MSFT: It's autoreleased, but stored in a StrongId (conferring +1). This is, unfortunately, idiomatic use of StrongId's constructor.

Frameworks/Foundation/NSOperationQueue.mm
bviglietta: Although that could change the behavior because of short circuiting.

---
In reply to: [95450897](https://github.com/Microsoft/WinObjC/pull/1676#discussion_r95450897) [](ancestors = 95450897)

Frameworks/Foundation/NSOperationQueue.mm
bviglietta: You're right, I overlooked the StrongId. Never mind...

---
In reply to: [95456936](https://github.com/Microsoft/WinObjC/pull/1676#discussion_r95456936) [](ancestors = 95456936)

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -30,167 +30,320 @@
...
+    StrongId<NSOperation> _curOperation;
+    BOOL _isSuspended;
+
+    StrongId<NSString> _name;
+
+    // NOTE: a normal (as opposed to recursive) mutex is used here to interop with condition variable.
+    // This helps simplify the locking in this class in that fewer locks are used so the overall ownership
+    // of the state this class owns is encapsulated under the single lock. However, this also means that
+    // special care is needed to ensure that no re-entrant functions are used while under the lock.
+    std::mutex _lock;
```
rajsesh-msft: suggest moving these variables to the top of this class, would make it easier to read the rest of the code.

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -30,167 +30,320 @@
...
-}
+    NSInteger MaxConcurrentOperationCount() {
+        std::lock_guard<std::mutex> lock(_lock);
+        return _maxConcurrentOperationCount;
+    }
 
-void NSAtomicListInsert(NSAtomicListRef* listPtr, NSObject* elt) {
-    struct NSAtomicListNode* node = new NSAtomicListNode();
-    node->elt = elt;
+    NSArray<NSOperation*>* GetOperations() {
```
rajsesh-msft: nit: could be const method.

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -30,167 +30,320 @@
...
-    if (!node) {
-        return NULL;
+    void SetMaxConcurrentOperationCount(NSInteger count) {
+        std::lock_guard<std::mutex> lock(_lock);
+        _maxConcurrentOperationCount = count;
     }
 
-    return node->elt;
-}
+    NSInteger MaxConcurrentOperationCount() {
```
rajsesh-msft: >MaxConcurrentOperationCount() [](start = 14, length = 29)

nit: could be const method.

Frameworks/Foundation/NSOperationQueue.mm
bbowman: yes that is the intent. There is already an issue Dustin filed awhile ago. I will update it to mention that all the NOTE: comments here should be addressed.

---
In reply to: [95448320](https://github.com/Microsoft/WinObjC/pull/1676#discussion_r95448320) [](ancestors = 95448320)

Frameworks/Foundation/NSOperationQueue.mm
bbowman: documentation states that notify calls are most efficient outside of the lock. Otherwise it causes a "hurry up and wait" scenario where the notified thread immediately waits again to re-acquire the mutex. Some implementations optimize for this but to be most portable, the recommendation is to notify outside of the lock.

---
In reply to: [95448692](https://github.com/Microsoft/WinObjC/pull/1676#discussion_r95448692) [](ancestors = 95448692)

Frameworks/Foundation/NSOperationQueue.mm
bbowman: not in spirit. They are implemented in that they don't break things but they do not actually influence concurrency. I decided to leave as is. The re-write will actually do something useful there presumably.

---
In reply to: [95449162](https://github.com/Microsoft/WinObjC/pull/1676#discussion_r95449162) [](ancestors = 95449162)

Frameworks/Foundation/NSOperationQueue.mm
bbowman: I could but I don't think I will. Its a pretty long line already and I think it would look weird in the init list and it uses the weirder autorelease pattern that I don't super enjoy.

---
In reply to: [95449339](https://github.com/Microsoft/WinObjC/pull/1676#discussion_r95449339) [](ancestors = 95449339)

Frameworks/Foundation/NSOperationQueue.mm
bbowman: this.

---
In reply to: [95449581](https://github.com/Microsoft/WinObjC/pull/1676#discussion_r95449581) [](ancestors = 95449581)

Frameworks/Foundation/NSOperationQueue.mm
bbowman: I'll add a note. I agree that its weird. I didn't have it like that initially but that is most in the spirit of the previous code. I very much dislike nil'ing out the thread to indicate state but some flag is needed under the lock to indicate that things are done and a new thread is needed next go around. 

A much preferable solution long term is to use a threadpool or some other mechanism where we aren't manually trying to optimize thread lifetime.

---
In reply to: [95450052](https://github.com/Microsoft/WinObjC/pull/1676#discussion_r95450052) [](ancestors = 95450052)

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -30,167 +30,320 @@
...
-void NSAtomicListReverse(NSAtomicListRef* listPtr) {
-    struct NSAtomicListNode* cur = *listPtr;
-    struct NSAtomicListNode* prev = NULL;
-    struct NSAtomicListNode* next = NULL;
+        BOOL toReturn = _HasMoreWorkNoLock();
+
+        if (!toReturn) {
+            // No work left to do. Clear out the thread here. This signals that a new work thread is needed to be created and
+            // current one can end its execution.
+            _thread = nil;
```
rajsesh-msft: >_thread [](start = 12, length = 7)

why isn't this set to nil when _workThread is done?  that seems more appropriate.

Frameworks/Foundation/NSOperationQueue.mm
bbowman: lol oops. I meant to std::move. It doesn't really matter though too much.

---
In reply to: [95450127](https://github.com/Microsoft/WinObjC/pull/1676#discussion_r95450127) [](ancestors = 95450127)

Frameworks/Foundation/NSOperationQueue.mm
bbowman: totally. The issue is that an add could happen after the decision to stop the thread has been made but before the stop actually occured. Right now its the add that determines the new thread needs created. This means that the decision point of stopping the current thread needs to also indicate the new one needs constructed. Its weird and stupid but it is the existing logic that I didn't want to touch without ripping out the entire thread management.

---
In reply to: [95459642](https://github.com/Microsoft/WinObjC/pull/1676#discussion_r95459642) [](ancestors = 95459642)

Frameworks/Foundation/NSOperationQueue.mm
bbowman: yeah I thought about it. I'm going to wont fix just because all our logic is currently written in the positive and it seemed more confusing to me to ! it some places. A full rewrite could make use of it better.

---
In reply to: [95450281](https://github.com/Microsoft/WinObjC/pull/1676#discussion_r95450281) [](ancestors = 95450281)

Frameworks/Foundation/NSOperationQueue.mm
bbowman: yep I agree. Was keeping existing logic as much as possible. 

---
In reply to: [95450533](https://github.com/Microsoft/WinObjC/pull/1676#discussion_r95450533) [](ancestors = 95450533)

Frameworks/Foundation/NSOperationQueue.mm
bbowman: why not?

---
In reply to: [95450845](https://github.com/Microsoft/WinObjC/pull/1676#discussion_r95450845) [](ancestors = 95450845)

bbowman: short circuit here would not be advisable. The logic here is pretty dastardly in that it for loops without a didWork check in the while loop. This means that even in the cases where work already occured, we should be trying the other priorities. It also means we can't just set it because we don't want to invalidate a previous priority that did in fact do work.

---
In reply to: [95457228](https://github.com/Microsoft/WinObjC/pull/1676#discussion_r95457228) [](ancestors = 95457228,95450897)

Frameworks/Foundation/NSOperationQueue.mm
bbowman: a setName call could be made but I don't really like that. I think unique_ptr is fine here.

---
In reply to: [95451424](https://github.com/Microsoft/WinObjC/pull/1676#discussion_r95451424) [](ancestors = 95451424)

Frameworks/Foundation/NSOperationQueue.mm
bbowman: swell.

---
In reply to: [95453130](https://github.com/Microsoft/WinObjC/pull/1676#discussion_r95453130) [](ancestors = 95453130)

Frameworks/Foundation/NSOperationQueue.mm
bbowman: kept the logic but yes its rather silly all around. This is why a deque was used so that reversing it was fast but totally not needed in reality. The real redesign won't use this weird holding area with demotion business, instead relying on KVO, the ready state, and a std::priority_queue to manage the pool.

---
In reply to: [95453430](https://github.com/Microsoft/WinObjC/pull/1676#discussion_r95453430) [](ancestors = 95453430)

Frameworks/Foundation/NSOperationQueue.mm
bbowman: good catch. forgot to remove when I moved to safe static init.

---
In reply to: [95456671](https://github.com/Microsoft/WinObjC/pull/1676#discussion_r95456671) [](ancestors = 95456671)

Frameworks/Foundation/NSOperationQueue.mm
bbowman: should be removed. was going to mention the thread loop but it doesn't actually impact that case.

---
In reply to: [95456923](https://github.com/Microsoft/WinObjC/pull/1676#discussion_r95456923) [](ancestors = 95456923)

Frameworks/Foundation/NSOperationQueue.mm
bbowman: True. I just kept the priv around as the encapsulating doodad but that could have just as easily been the objc class itself.

---
In reply to: [95450433](https://github.com/Microsoft/WinObjC/pull/1676#discussion_r95450433) [](ancestors = 95450433)

Frameworks/Foundation/NSOperationQueue.mm
bbowman: I thought the general convention was at the end for C++ classes? 

---
In reply to: [95458070](https://github.com/Microsoft/WinObjC/pull/1676#discussion_r95458070) [](ancestors = 95458070)

PR #1677, opened 2017-01-10
Fix exact match behavior after updating a unit.

### LINKED ISSUE #1629: NSCalendar enumerateDatesStartingAfterDate ExactMatch is incorrect.
The exact match parameter being passed to the enumeration block may have an incorrect value in some cases. For example:

    NSCalendar* calendar = [NSCalendar calendarWithIdentifier:NSCalendarIdentifierGregorian];
    NSDate* start = [calendar dateWithEra:1 year:2010 month:3 day:1 hour:1 minute:2 second:3 nanosecond:0];

    NSDateComponents* components = [[NSDateComponents new] autorelease];
    // Should find any date with day == 30. Since this is the only component to match against, any
    // date found should be an exact match.
    components.day = 30;

    [calendar enumerateDatesStartingAfterDate:start
                           matchingComponents:components
                                      options:NSCalendarMatchPreviousTimePreservingSmallerUnits | NSCalendarSearchBackwards
                                   usingBlock:^(NSDate* date, BOOL exactMatch, BOOL* stop) {
                                       NSInteger month = [calendar component:NSCalendarUnitMonth fromDate:date];

                                       if (month != 2) {
                                           // This check fails as exactMatch should be true in every case.
                                           ASSERT_TRUE(exactMatch);
                                       }
                                   }];

MSFTFox:
After updating the largest allowed component of the date components for enumerating matches, attempt again to match the exact date using the newly updated date components.

Prior to this, the matching algorithm was incorrect such that iterating backwards to find the 30th day of a month for example would instead first find the 28th day of January due to February restricting the day component to 28.

Fixes #1629

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1677)

### DIFF
--- a/Frameworks/Foundation/NSCFCalendar.mm
+++ b/Frameworks/Foundation/NSCFCalendar.mm
@@ -558,6 +558,22 @@ static NSDate* _validDateNormalUnit(NSCalendar* cal,
         workingDate = [cal dateByAddingComponents:updateUnit toDate:workingDate options:opts];
     }
 
+    NSDateComponents* directMatchingComps = [cal components:s_NSDateComponentsAllFlagOptions fromDate:workingDate];
+    for (int i = 0; i < _countof(s_NSDateComponentsIndividualFlags); i++) {
+        NSInteger expectedValue = [matchComps valueForComponent:s_NSDateComponentsIndividualFlags[i]];
+        NSInteger workingValue = [workingComps valueForComponent:s_NSDateComponentsIndividualFlags[i]];
+        if (expectedValue != NSUndefinedDateComponent) {
+            [directMatchingComps setValue:expectedValue forComponent:s_NSDateComponentsIndividualFlags[i]];
+        }
+    }
+
+    // Attempt to directly create the matching date using the expected matching components now that the largest
+    // possible unit has been updated.
+    if (_isValidDate(cal, directMatchingComps)) {
+        NSDate* exactDate = [cal dateFromComponents:directMatchingComps];
+        return exactDate;
+    }
+
     return workingDate;
 }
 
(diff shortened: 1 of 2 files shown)


Frameworks/Foundation/NSCFCalendar.mm
```diff
@@ -558,6 +558,22 @@ static void _initAllFlagsComponents() {
         workingDate = [cal dateByAddingComponents:updateUnit toDate:workingDate options:opts];
     }
 
+    NSDateComponents* directMatchingComps = [cal components:s_NSDateComponentsAllFlagOptions fromDate:workingDate];
+    for (int i = 0; i < _countof(s_NSDateComponentsIndividualFlags); i++) {
```
aballway: nit: use std::extent

Frameworks/Foundation/NSCFCalendar.mm
```diff
@@ -558,6 +558,22 @@ static void _initAllFlagsComponents() {
...
+        NSInteger expectedValue = [matchComps valueForComponent:s_NSDateComponentsIndividualFlags[i]];
+        NSInteger workingValue = [workingComps valueForComponent:s_NSDateComponentsIndividualFlags[i]];
```
aballway: not used?

tests/unittests/Foundation/NSCalendarTests.mm
```diff
@@ -534,4 +534,29 @@
...
+    __block int iterationsToTest = 0;
+
+    [calendar enumerateDatesStartingAfterDate:start
+                           matchingComponents:components
+                                      options:NSCalendarMatchPreviousTimePreservingSmallerUnits | NSCalendarSearchBackwards
+                                   usingBlock:^(NSDate* date, BOOL exactMatch, BOOL* stop) {
+                                       NSInteger month = [calendar component:NSCalendarUnitMonth fromDate:date];
+
+                                       if (month != 2) {
+                                           ASSERT_TRUE(exactMatch);
```
DHowett-MSFT: Might like to also assert that if month == 2, !exactMatch. s

PR #1678, opened 2017-01-10
Re-enable setApplicationBadgeIconNumber.  Fixes #1201.  Works very well.

### LINKED ISSUE #1201: UIApplication setApplicationIconBadgeNumber doesn't have proper privileges for notification
UIApplication.mm::700

[updater update:notification];
throws 0x803e0208 : The notification platform does not have the proper privileges to complete the request.

may have been a change in the underlying windows APIs. Temporarily commenting this out.

rajsesh-msft:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1678)

### DIFF
--- a/Frameworks/UIKit/UIApplication.mm
+++ b/Frameworks/UIKit/UIApplication.mm
@@ -714,7 +714,6 @@ static int __EbrSortViewPriorities(id val1, id val2, void* context) {
     WUNBadgeNotification* notification = [WUNBadgeNotification makeBadgeNotification:doc];
     WUNBadgeUpdater* updater = [WUNBadgeUpdateManager createBadgeUpdaterForApplication];
 
-    // TODO #1201: 0x803e0208 : The notification platform does not have the proper privileges to complete the request.
     [updater update:notification];
 }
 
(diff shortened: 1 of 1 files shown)


Frameworks/UIKit/UIApplication.mm
```diff
@@ -715,7 +715,7 @@ - (void)setApplicationIconBadgeNumber:(int)num {
     WUNBadgeUpdater* updater = [WUNBadgeUpdateManager createBadgeUpdaterForApplication];
 
     // TODO #1201: 0x803e0208 : The notification platform does not have the proper privileges to complete the request.
```
rajsesh-msft: forgot to remove this TODO.

ms-jihua:
Do we know what caused this to no longer be an issue?

bbowman:
"Works very well."

Should we not make subjective assertions about the code in the commit message? Seems like either you are trying to convince someone of it which seems wrong in a commit message or that you are extra proud of yourself which is fine but also probably not appropriate for the commit message. I'd prefer if we had a concrete metric+comparison point to back up any such claims.

rajsesh-msft:
No.  Quite likely resolved by new projections drop with RS1 SDK?  

---
In reply to: [271708686](https://github.com/Microsoft/WinObjC/pull/1678#issuecomment-271708686) [](ancestors = 271708686)

rajsesh-msft:
I'll drop that fragment when i merge.

---
In reply to: [271709239](https://github.com/Microsoft/WinObjC/pull/1678#issuecomment-271709239) [](ancestors = 271709239)

PR #1679, opened 2017-01-11
Fix -classForCoder on bridged classes to respect mutability.

### LINKED ISSUE #902: Bridged class implementation of classForCoder always returns immutable class
This can be an issue if an app tries to encode/decode mutable types

DHowett-MSFT:
* Add tests for all of the cases we support.
* We do not support archival of NSAttributedString (!)

Fixes #902.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1679)

### DIFF
--- a/Frameworks/Foundation/NSString.mm
+++ b/Frameworks/Foundation/NSString.mm
@@ -2021,7 +2021,13 @@ static std::vector<NSStringEncoding> _getNSStringEncodings() {
  @Status Interoperable
 */
 - (void)encodeWithCoder:(NSCoder*)coder {
-    [coder encodeObject:self forKey:@"NS.string"];
+    if ([coder isKindOfClass:[NSKeyedArchiver class]]) {
+        [coder _encodeRawObject:self forKey:@"NS.string"];
+    } else {
+        // This can cause custom string subclasses to act up.
+        // Ideally, a developer would override encodeWithCoder:.
+        [coder encodeObject:[NSString stringWithString:self] forKey:@"NS.string"];
+    }
 }
 
 /**
--- a/tests/unittests/Foundation/NSAttributedStringTests.mm
+++ b/tests/unittests/Foundation/NSAttributedStringTests.mm
@@ -773,3 +773,18 @@ TEST(NSAttributedString, Subclass) {
     ASSERT_EQ(2, CFAttributedStringGetLength((__bridge CFAttributedStringRef)didOverride));
     ASSERT_EQ(YES, [didOverride methodCalled]);
 }
+
+DISABLED_TEST(NSAttributedString, MutableInstanceArchivesAsMutable) {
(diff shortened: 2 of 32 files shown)


tests/unittests/Foundation/NSAttributedStringTests.mm
```diff
@@ -772,4 +772,19 @@ - (NSUInteger)length {
...
+    ASSERT_OBJCNE(nil, data);
+
+    NSMutableAttributedString* output = [NSKeyedUnarchiver unarchiveObjectWithData:data];
+    ASSERT_OBJCNE(nil, output);
+
+    EXPECT_NO_THROW([output appendAttributedString:[[[NSMutableAttributedString alloc] initWithString:@" world"] autorelease]]);
+    EXPECT_OBJCEQ(@"hello world", [output string]);
+
+    EXPECT_OBJCNE(input, output);
+}
```
DHowett-MSFT: This file has CRLF line endings.

tests/unittests/Foundation/NSCharacterSetTests.m
```diff
@@ -222,4 +222,21 @@ void testCharacter(NSCharacterSet* charSet, NSString* testString) {
...
+    NSMutableCharacterSet* output = [NSKeyedUnarchiver unarchiveObjectWithData:data];
+    ASSERT_OBJCNE(nil, output);
+
+    EXPECT_NO_THROW([output addCharactersInString:@"b"]);
+
+    // We cannot compare these for equality. Despite having different ranges (!) they are considered equal
+    // on the reference platform.
+    EXPECT_FALSE([input characterIsMember:'b']);
+    EXPECT_TRUE([output characterIsMember:'b']);
+}
```
DHowett-MSFT: This file has CRLF line endings.

Frameworks/Foundation/NSString.mm
```diff
@@ -2021,7 +2021,13 @@ - (id)initWithCoder:(NSCoder*)coder {
  @Status Interoperable
 */
 - (void)encodeWithCoder:(NSCoder*)coder {
-    [coder encodeObject:self forKey:@"NS.string"];
+    if ([coder isKindOfClass:[NSKeyedArchiver class]]) {
```
DHowett-MSFT: I am not happy about having to do this.
However.

Without this escape hatch, our `NSKeyedArchiver` will encode `[NSMutableString stringWithString:@"x"]` as:

```
0: $null
1: {
    class: <2>
    NS.string: <1>
}
2: {
    $className: NSMutableString
    ...
}
```

(where the string is object **1** and `<x>` indicates a reference to object *x*)

DHowett-MSFT: That is to say: NSKeyedArchiver will encode a string that refers to itself as its string.

rajsesh-msft:
:shipit:

PR #1685, opened 2017-01-11
Fixing hit testing for CALayers and UIViews that have clear backgrounds

### LINKED ISSUE #1479: Popover related touch/view-hierarchy problem
(this is not an urgent issue for me as there's a workaround noted below and my app development is currently against a stable branch)

Some of the recent view/layer work has broken popover dismisses (via touch on a clear overlay view outside the bounds of the popover content view controller). 

Steps to reproduce:

1) Open WOCCatalog app
2) Enable tablet mode in display mode view controller, press back
3) Load Controls View Controller
4) Touch a popover button
5) Dismiss popover via touch outside the popover content

Expected:
The popover is dismissed

Actual:
The touch passes through to the table below; popover is not dismissed

I've been able to bisect the commits and I believe the problem was introduced by ec804795a924609eb975a6518475a6371c871b14

The last good version (where popover is dismissed) is b52a219eba2093a3c273ecee10e0a81242072979

The relevant WYPopoverController code is here:
https://github.com/Microsoft/WinObjC/blob/812d6636036a8d2eaeaaf7b92aa0c7b76252e190/Frameworks/UIKit/WYPopoverController.m#L1937

```
        overlayView = [[WYPopoverOverlayView alloc] initWithFrame:inView.window.bounds];
        overlayView.autoresizesSubviews = NO;
        overlayView.isAccessibilityElement = YES;
        overlayView.accessibilityTraits = UIAccessibilityTraitNone;
        overlayView.delegate = self;
        overlayView.passthroughViews = passthroughViews;
        
        backgroundView = [[WYPopoverBackgroundView alloc] initWithContentSize:contentViewSize];
        backgroundView.appearing = YES;
        backgroundView.isAccessibilityElement = YES;
        backgroundView.accessibilityTraits = UIAccessibilityTraitNone;
        
        backgroundView.delegate = self;
        backgroundView.hidden = YES;
        
        [inView.window addSubview:backgroundView];
        [inView.window insertSubview:overlayView belowSubview:backgroundView];
```

Something is going wrong with `overlayView` here.

A few other notes:

- This problem can be bypassed by selecting the legacy hit test operation mode.
- However I don't believe this is a "legacy hit test" issue - though overlayView does implement hitTest it doesn't request to participate in hit testing outside it's bounds.
- I suspect something is funny with the view hierarchy at the point the background and overlay views are added to the window of the inView (which corresponds to the UIPopoverPresentationController sourceView)
- I have not yet attempted to dump a description of the view hierarchy at the time the above code is executed. This could be enlightening.

### LINKED ISSUE #1574: WOCCatalog [x86|Debug] Controls: Popovers cause app to crash
x86 running 14975
https://github.com/Microsoft/WinObjC/commit/4804ed9bf1b16225116ff7bb06477ce3c8c300d7

Launch app. Navigate to Display Mode and enable Tablet Mode. Navigate to Controls. Launch a popover via the "Popover (left arrow direction)" button. Close it by hitting the done button. App will crash.

jaredhms:
; they should receive pointer/touch input.  Note that this is different than setting the UIView's alpha value to 0.0 (in which case they should not be hit-testable).

The legacy version of the compositor had an 'optimization' for transparent layer backgrounds; if the set background was transparent, it would set the background brush on the corresponding Xaml element to null.  This is incorrect, because Xaml elements with null backgrounds aren't able to receive pointer input, so this breaks hit testing.  I think this used to work pre-compositor-refactor because any layer that didn't have a background would instead have a transparent rectangle for receiving touch input.  Post-refactor, the extra rectangles are no longer needed, so this legacy 'optimization' is incorrect.

Fixes #1479

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1685)


jaredhms:
@ehren - this fixes the issue in WoCCatalog.

ehren:
@jaredhms will give it a full test tomorrow (it's very likely this fixes a number of issues in my app against the current develop). thanks!

tadam-msft:
:shipit:

tadam-msft:
Would be great to have a XAML Catalog test case that illustrates this (or CALayer test case).

tadam-msft:
:shipit:

samples/XAMLCatalog/XAMLCatalog/UIViewViewController.m
```diff
@@ -0,0 +1,158 @@
...
+        label.text = @"Alpha=0.0; yellow background SHOULD NOT be visible and view SHOULD NOT respond to touch.";
+        testView.alpha = 0.0;
+        testView.shouldRespondToTouch = NO;
+    } else if (indexPath.row == 1) {
+        // Setup test
+        label.text = @"Yellow background SHOULD be visible and view SHOULD respond to touch.";
+    } else if (indexPath.row == 2) {
+        // Setup test
+        label.text = @"Clear background SHOULD NOT be visible, but view SHOULD respond to touch.";
+        testView.backgroundColor = [UIColor clearColor];
```
yiyang-msft: > testView.backgroundColor  [](start = 7, length = 26)

nit: can we add a test for nil color?

jaredhms: It should be have the same as clear, but good idea.

---
In reply to: [95709193](https://github.com/Microsoft/WinObjC/pull/1685#discussion_r95709193) [](ancestors = 95709193)

yiyang-msft:
:shipit:

PR #1688, opened 2017-01-11
Register bundled fonts earlier in UIApplicationMainInit

### LINKED ISSUE #1661: Islandwood News: Missing 'video' and 'slideshow' images - they show up as small grey squares
**Repro Steps**
Images sometimes get obscured with a dark grey surface with a black rectangle (bottom right hand corner)
attachfilehandler

**Comments**
Jingtao Huang 
Seems like a bug. When you say ‘slideshow’ image, you mean something like an icon that looks like this , right? (This is from the Photos app, I imagine the reference platform button would look a little different) If so,  small squares are a ‘default’/character-not-found glyph in most fonts, they represent that there was no glyph data stored at that index in the font.
I notice that the button to the right of the search bar is also character-not-found, so it’s probably the same root cause? (I tried two articles in hunt of a slideshow icon, but they both ended up crashing)
In October I made a commit c79a2b6372931389b5456b36eb1c310c3676b29f that switched some fonts used in IWNews – when we merged CT-DW, we regressed the ability to load fonts from ttf files/the app bundle. (We have since reimplemented this functionality) One of the changes made was from a non-system default font (ICEAppleGlyphs) to Lucida Sans Regular, and we never changed it back. I also notice that if I break on your code snippet, the first hit uses the font Lucida Sans Regular. I’m guessing that the root cause is that Lucida Sans Regular doesn’t actually have this glyph that we’re supposed to be able to get from ICEAppleGlyphs.

Jared Henderson (3 days ago)
I think I’ve narrowed it down to a font/glyph issue.  It looks like this code:  
 
    NSString *text = [NSString stringWithCharacters:&glyph length:1]; // ‘glyph’ = 57747 in this case
    // set the font type and size
    UIFont *font = [axpApplication.appConfig fontFromFontReference:@"AXPGlyphFont"];
    CGFloat fontSize = MIN(size.height, size.width);
    font = [font fontWithSize:fontSize];
    
    NSDictionary *attrDict = [NSDictionary dictionaryWithObjectsAndKeys:
                              font, NSFontAttributeName,
                              textColor, NSForegroundColorAttributeName,
                              nil];
    
    CGSize textSize  = [text sizeWithAttributes:attrDict];
    UIGraphicsBeginImageContextWithOptions(size,NO,0.0);
    
    // optional: add a shadow, to avoid clipping the shadow you should make the context size bigger
    //
    // CGContextRef ctx = UIGraphicsGetCurrentContext();
    // CGContextSetShadowWithColor(ctx, CGSizeMake(1.0, 1.0), 5.0, [[UIColor grayColor] CGColor]);
    if (textColor)
    {
        CGContextSetFillColorWithColor(UIGraphicsGetCurrentContext(), [textColor CGColor]);
    }
    [text drawAtPoint:CGPointMake((size.width - textSize.width)/2, (size.height - textSize.height)/2) withAttributes:attrDict];
    UIImage *image = UIGraphicsGetImageFromCurrentImageContext();
    UIGraphicsEndImageContext();
 
…draws an incorrect image (it’s supposed to be a ‘slideshow’ image)

Jared Henderson (a month ago)
These are icons added by the app, although it's unclear why we're not rendering the expected icons.  If you run the app on iOS, you'll see that these grey squares should contain icons signifying that it's a link to a video, a link to a photo album, etc.

ms-jihua:
Fixes #1661

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1688)

### DIFF
--- a/Frameworks/UIKit/UIApplicationMain.mm
+++ b/Frameworks/UIKit/UIApplicationMain.mm
@@ -140,6 +140,25 @@ int UIApplicationMainInit(NSString* principalClassName,
     [[NSThread currentThread] _associateWithMainThread];
     NSDictionary* infoDict = [[NSBundle mainBundle] infoDictionary];
 
+    // Register fonts listed in app's Info.plist, from the app's bundle
+    // This needs to happen before [UIApplication new]/[objc_getClass(pClassName) new] below,
+    // as they may attempt to use fonts from the app's bundle
+    if (infoDict) {
+        NSArray* fonts = [infoDict objectForKey:@"UIAppFonts"];
+        if (fonts != nil) {
+            NSMutableArray* fontURLs = [NSMutableArray array];
+            for (NSString* curFontName in fonts) {
+                // curFontName contains extension, so pass in nil
+                NSURL* url = [[NSBundle mainBundle] URLForResource:curFontName withExtension:nil];
+                if (url != nil) {
+                    [fontURLs addObject:url];
+                }
+            }
+
+            CTFontManagerRegisterFontsForURLs((CFArrayRef)fontURLs, kCTFontManagerScopeNone, nil);
+        }
+    }
+
     outerPool = [NSAutoreleasePoolWarn new];
     NSAutoreleasePool* pool = [NSAutoreleasePool new];
     NSRunLoop* runLoop = [NSRunLoop currentRunLoop];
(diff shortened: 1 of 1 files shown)


Frameworks/UIKit/UIApplicationMain.mm
```diff
@@ -140,6 +140,25 @@ int UIApplicationMainInit(NSString* principalClassName,
...
+                NSURL* url = [[NSBundle mainBundle] URLForResource:curFontName withExtension:nil];
+                if (url != nil) {
+                    [fontURLs addObject:url];
+                }
+            }
+
+            CTFontManagerRegisterFontsForURLs((CFArrayRef)fontURLs, kCTFontManagerScopeNone, nil);
+        }
+    }
+
```
rajsesh-msft: the auto release pools that are created after this must instead be created first.  In fact, the pools must be created before the [NSThread] line.

DHowett-MSFT:
Could font loading conceivably happen in `-[UIApplication init]`? It must be called by all subclasses, and it happens during `-new` but before the user's code.

ms-jihua:
it probably could, but what do we gain from it? it would still count towards a 'startup load time' bucket, and we'd split the usage of [[NSBundle mainBundle] infoDictionary]

---
In reply to: [272035886](https://github.com/Microsoft/WinObjC/pull/1688#issuecomment-272035886) [](ancestors = 272035886)

DHowett-MSFT:
We gain only separation of responsibilities :smile: only in an idealistic way.

ms-jihua:
we discussed offline that it doesn't make sense to move just this out at this time

---
In reply to: [272037465](https://github.com/Microsoft/WinObjC/pull/1688#issuecomment-272037465) [](ancestors = 272037465)

PR #1699, opened 2017-01-13
Have compiler pre-hash selectors

bviglietta:
This change updates the compiler and libobjc2 runtime to pre-hash selectors at compile time rather than at load time.

This decreases startup time in WOCCatalog by about 80 ms (out of 750 ms total), of which about 25ms comes directly from the elimination of the hash_selector calls. There's no observable reduction in memory usage, although in theory we could get some if we could do more to convince libobjc2 to stop writing to the copy-on-write pages of the data section.

This is a breaking ABI change, so you can't use code built with the new compiler against an old runtime. That's not likely to be common, though. The other way around -- old code with a new runtime -- works fine.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1699)


bviglietta:

bviglietta:

DHowett-MSFT:
This will require libobjc2 to be rebuilt with the updated changes from @pradipd's exception fix.

bbowman:
Looks awesome thanks! 80ms seems like a good savings if I recall right (I think it was ~200-300 ms load time for projections).

GreatKeeper:
The toolset I provided is a chk build, so that we can test this feature. We normally use a signed ret build for the develop branch. I should provide a signed toolset when we are ready to merge, unless I'm misunderstanding something?

bviglietta:
Oh, good point. Yes, we'll need those signed ret binaries whenever you've got them.

---
In reply to: [272340503](https://github.com/Microsoft/WinObjC/pull/1699#issuecomment-272340503) [](ancestors = 272340503)

Issue #1634, opened 2017-01-05
Remove CGBitmapContextCairo and move CT stuff into CGContext


rajsesh-msft:
This is needed for merging CGD2D to develop.

rajsesh-msft:
Fixed with #1729

PR #1700, opened 2017-01-13
Add woc::AutoCF<T>. It's like AutoId<T> for CFTypes.

DHowett-MSFT:
Just like `AutoId<T>`, it takes a lifetime trait template: CFLifetimeRetain
for CFRetain/CFRelease, and CFLifetimeUnsafe for straight assignment.

It builds upon and replaces `woc::unique_cf<T>`.
 * It is copyable (and can share ownership)
 * `AutoCF<T>` can be passed anywhere `T*` is required.
 * It has a fully-loaded `operator&` so you can pass it wherever
   a `T**` is required. The old value will be released.
 * It has counterparts in StrongCF and UnsafeCF, just like StrongId
   and UnsafeId.

By default, a StrongCF will *share ownership* of a CFType.
To transfer ownership, construct it with `std::move(cf)` or use
`woc::MakeStrongCF<T>(CFCreateXxx(...))`.

As a migration measure, `woc::unique_cf<T>` is still provided, though it is
now based on StrongCF. It retains its "transfer ownership on construct"
semantics and provides a few shim methods to adapt it to the
AutoCF/AutoID world (release -> detach, reset -> attach).

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1700)

### DIFF
--- a/build/Tests/UnitTests/Starboard/Starboard.UnitTests.vcxproj
+++ b/build/Tests/UnitTests/Starboard/Starboard.UnitTests.vcxproj
@@ -226,6 +226,7 @@
     <ClCompile Include="$(StarboardBasePath)\tests\unittests\EntryPoint.cpp" />
   </ItemGroup>
   <ItemGroup>
+    <ClangCompile Include="..\..\..\..\Tests\UnitTests\Starboard\AutoCFTests.cpp" />
     <ClangCompile Include="..\..\..\..\tests\unittests\Starboard\AutoIdTests_ARC.mm" />
     <ClangCompile Include="..\..\..\..\tests\unittests\Starboard\AutoIdTests_NoARC.mm" />
     <ClangCompile Include="..\..\..\..\tests\unittests\Starboard\CommonCryptoTests.mm" />
--- a/include/xplat/Starboard/SmartTypes.h
+++ b/include/xplat/Starboard/SmartTypes.h
@@ -305,16 +305,254 @@ bool operator!=(const Any& other, const AutoId<TObj, TLifetimeTraits>& val) {
 
 #ifdef CF_EXPORT // Quick way to detect CoreFoundation.
 namespace woc {
-template <typename T>
-class unique_cf : public std::unique_ptr<typename std::remove_pointer<T>::type, decltype(&CFRelease)> {
-public:
-    unique_cf() : std::unique_ptr<typename std::remove_pointer<T>::type, decltype(&CFRelease)>(nullptr, CFRelease) {
-    }
-
-    explicit unique_cf(T&& val)
-        : std::unique_ptr<typename std::remove_pointer<T>::type, decltype(&CFRelease)>(std::forward<T>(val), CFRelease) {
+// Unsafe - a direct non-refcounting store of a refcountable object.
+struct CFLifetimeUnsafe {
+    inline static CFTypeRef store(CFTypeRef* destination, CFTypeRef value) {
+        *destination = value;
+        return value;
     }
 };
+
+// Retain - a retaining/releasing store of a refcountable object.
+//  - Releases the refcounted value at the destination while never allowing
+//    *destination to hold a non-owning reference.
+struct CFLifetimeRetain {
+    inline static CFTypeRef store(CFTypeRef* destination, CFTypeRef value) {
+        if (*destination == value) {
+            return value;
+        }
+
+        if (*destination) {
+            CFRelease(static_cast<CFTypeRef>(*destination));
+        }
+
+        return *destination = ((value ? CFRetain(value) : nullptr));
+    }
+};
+
+namespace details {
+template <typename T, typename TWrapped>
+class AutoCFRef {
+private:
+    T* ptr;
+
+public:
+    AutoCFRef(T* ptr): ptr(ptr) { }
+
+    operator TWrapped*() const {
+        ptr->attach(nullptr);
+        return &ptr->get();
+    }
+
+    operator void**() const {
+        ptr->attach(nullptr);
+        return &ptr->get();
+    }
+
+    operator TWrapped() const {
+        return ptr->get();
+    }
+
+    operator void*() const {
+        return ptr->get();
+    }
+};
+}
+
+template <typename T, typename TLifetimeTraits = CFLifetimeRetain>
+class AutoCF {
+private:
+    inline CFTypeRef* _addressof() {
+        using TMutable = std::add_pointer_t<std::remove_const_t<std::remove_pointer_t<T>>>;
+        return const_cast<CFTypeRef*>(reinterpret_cast<void**>(const_cast<TMutable*>(&_val)));
+    }
+
+    template <typename TOther>
+    using TCanConvertFrom = std::is_convertible<TOther, T>;
+
+    template <typename TOther>
+    using TCanConvertTo = std::is_convertible<T, TOther>;
+
+    T _val = reinterpret_cast<T>(0);
+
+public:
+    AutoCF() : _val(nullptr) {
+    }
+
+    explicit AutoCF(const T& val) : _val(nullptr) {
+        TLifetimeTraits::store(_addressof(), static_cast<CFTypeRef>(val));
+    }
+
+    explicit AutoCF(T&& val): _val(val) {
+        val = nullptr;
+    }
+
+    template <typename TOtherLifetime>
+    AutoCF(const AutoCF<T, TOtherLifetime>& other)
+        : _val(nullptr) {
+        TLifetimeTraits::store(_addressof(), static_cast<CFTypeRef>(other._val));
+    }
+
+    // Copy from same lifetime.
+    AutoCF(const AutoCF<T, TLifetimeTraits>& other) : _val(nullptr) {
+        TLifetimeTraits::store(_addressof(), static_cast<CFTypeRef>(other._val));
+    }
+
+    // We can only move from the same lifetime, and only with convertible objects.
+    template <typename TOtherLifetime>
+    AutoCF(AutoCF<T, TOtherLifetime>&& other) {
+        static_assert(std::is_same<TLifetimeTraits, TOtherLifetime>::value,
+                      "AutoCF can only move from another AutoCF&& with the same lifetime.");
+    }
+
+    // Move from same lifetime.
+    AutoCF(AutoCF<T, TLifetimeTraits>&& other) : _val(other._val) {
+        other._val = nil;
+    }
+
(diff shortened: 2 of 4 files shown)


DHowett-MSFT:
This is a dependency of #1634, because `__CGContextDrawingState` needs a shareable (copyable) reference to a `CGFont`. `woc::unique_cf` is unsuitable for this purpose.

build/Tests/UnitTests/Starboard/Starboard.UnitTests.vcxproj
```diff
@@ -226,6 +226,7 @@
     <ClCompile Include="$(StarboardBasePath)\tests\unittests\EntryPoint.cpp" />
   </ItemGroup>
   <ItemGroup>
+    <ClangCompile Include="..\..\..\..\Tests\UnitTests\Starboard\AutoCFTests.cpp" />
```
DHowett-MSFT: I'll try to fix the casing on this directory name. Case-sensitive filenames are very much not happy about this.

include/xplat/Starboard/SmartTypes.h
```diff
@@ -305,16 +305,250 @@ bool operator!=(const Any& other, const AutoId<TObj, TLifetimeTraits>& val) {
...
+namespace details {
+template <typename T, typename TWrapped>
+class AutoCFRef {
+private:
+    T* ptr;
+
+public:
+    AutoCFRef(T* ptr): ptr(ptr) { }
+
+    operator TWrapped*() const {
```
rajsesh-msft: You need two of these: 

operator const TWrapped* () const 
operator TWrapped* ()

Note that the version that returns non-const pointer should not be const.

include/xplat/Starboard/SmartTypes.h
```diff
@@ -305,16 +305,250 @@ bool operator!=(const Any& other, const AutoId<TObj, TLifetimeTraits>& val) {
...
+        ptr->attach(nullptr);
+        return &ptr->get();
+    }
+
+    operator void**() const {
```
rajsesh-msft: can't be const, value is changing.

include/xplat/Starboard/SmartTypes.h
rajsesh-msft: Other methods too...

include/xplat/Starboard/SmartTypes.h
DHowett-MSFT: This is nearly 1:1 from the implementation of `WRL::Details::ComPtrRef`, after which this is modelled. Since it's the ref type that's const, we can change the mutable pointer through it (and that is, in fact, what WRL does!)

DHowett-MSFT: (this applies above as well: the const version of their pointer-to-T* operator does not return a const pointer)

rajsesh-msft: That is entirely at odds with std::unique_ptr, etc, which are more correct.  

---
In reply to: [95920587](https://github.com/Microsoft/WinObjC/pull/1700#discussion_r95920587) [](ancestors = 95920587)

include/xplat/Starboard/SmartTypes.h
```diff
@@ -305,16 +305,250 @@ bool operator!=(const Any& other, const AutoId<TObj, TLifetimeTraits>& val) {
...
+        return ptr->get();
+    }
+};
+}
+
+template <typename T, typename TLifetimeTraits = CFLifetimeRetain>
+class AutoCF {
+private:
+    inline CFTypeRef* _addressof() {
+        using TMutable = typename std::add_pointer<typename std::remove_const<typename std::remove_pointer<T>::type>::type>::type;
```
aballway: nit: can use add_pointer_t here?

include/xplat/Starboard/SmartTypes.h
```diff
@@ -305,16 +305,250 @@ bool operator!=(const Any& other, const AutoId<TObj, TLifetimeTraits>& val) {
...
+    }
+
+    template <typename TOtherLifetime>
+    AutoCF(const AutoCF<T, TOtherLifetime>& other)
+        : _val(nullptr) {
+        TLifetimeTraits::store(_addressof(), static_cast<CFTypeRef>(other._val));
+    }
+
+    // Copy from same lifetime.
+    AutoCF(const AutoCF<T, TLifetimeTraits>& other) : _val(nullptr) {
```
aballway: any reason for having this separate?

include/xplat/Starboard/SmartTypes.h
```diff
@@ -305,16 +305,250 @@ bool operator!=(const Any& other, const AutoId<TObj, TLifetimeTraits>& val) {
...
+        TLifetimeTraits::store(_addressof(), static_cast<CFTypeRef>(other._val));
+    }
+
+    // We can only move from the same lifetime, and only with convertible objects.
+    template <typename TOtherLifetime>
+    AutoCF(AutoCF<T, TOtherLifetime>&& other) {
```
aballway: would not defining this for TOtherLifetime and instead just having TLifetimeTraits version accomplish this?

include/xplat/Starboard/SmartTypes.h
```diff
@@ -305,16 +305,250 @@ bool operator!=(const Any& other, const AutoId<TObj, TLifetimeTraits>& val) {
...
+    }
+
+    explicit operator bool() const {
+        return _val != nullptr;
+    }
+
+    // The attach function assumes ownership of an already-refcounted object.
+    void attach(T val) {
+        if (val == _val) {
+            return;
```
aballway: https://media.giphy.com/media/e03UozELSlS3m/giphy.gif

include/xplat/Starboard/SmartTypes.h
```diff
@@ -305,16 +305,250 @@ bool operator!=(const Any& other, const AutoId<TObj, TLifetimeTraits>& val) {
...
+    // Same object, same lifetime.
+    // Without the second, MSVC and clang will generate a direct value copy for
+    // AutoCF<T> x, y; x = y;
+    // thus breaking the refcounting on the contents of x.
+    //
+    // The cross-lifetime copy assignment operator and constructor allow for a retaining AutoCF to take a +1
+    // refcount on another weakly-held AutoCF's value.
+    template <typename TOtherLifetime>
+    AutoCF<T, TLifetimeTraits>& operator=(const AutoCF<T, TOtherLifetime>& other) {
+        TLifetimeTraits::store(_addressof(), static_cast<CFTypeRef>(other._val));
```
aballway: do these need to check (*this != other)?

include/xplat/Starboard/SmartTypes.h
ms-jihua: also the other two _t's

---
In reply to: [96037442](https://github.com/Microsoft/WinObjC/pull/1700#discussion_r96037442) [](ancestors = 96037442)

include/xplat/Starboard/SmartTypes.h
rajsesh-msft: Thinking some more about this, const correctness applies only to C++.  That explains why WRL is also at odds with std::unique_ptr.  Clearing my review status.

include/xplat/Starboard/SmartTypes.h
DHowett-MSFT: The implementations of `store(...)` will stop that from being catastrophic.

include/xplat/Starboard/SmartTypes.h
DHowett-MSFT: Early returns: AJ's kryptonite

include/xplat/Starboard/SmartTypes.h
DHowett-MSFT: I want to explicitly `static_assert` with a useful message, not "lol could not find an applicable constructor, I tried `<these 5 ctors>`"

include/xplat/Starboard/SmartTypes.h
DHowett-MSFT: Yes. I'm fairly certain the assignment operator explanation below applies here as well.

DHowett-MSFT:
Because of template type inference, it's possible to do this:

```cpp
auto cf = woc::MakeStrongCF(CFStringCreate(...))
// typeof(cf) = woc::AutoCF<CFStringRef, CFLifetimeRetaion>
```

However, that reduces clarity. I suggest that an explicit `<CFStringRef>` appear on either the left or right just to be safe.

Thoughts?

include/xplat/Starboard/SmartTypes.h
```diff
@@ -305,16 +305,250 @@ bool operator!=(const Any& other, const AutoId<TObj, TLifetimeTraits>& val) {
...
+
 template <typename T>
-class unique_cf : public std::unique_ptr<typename std::remove_pointer<T>::type, decltype(&CFRelease)> {
+class unique_cf: public AutoCF<T, CFLifetimeRetain> {
 public:
-    unique_cf() : std::unique_ptr<typename std::remove_pointer<T>::type, decltype(&CFRelease)>(nullptr, CFRelease) {
+    unique_cf(): AutoCF<T, CFLifetimeRetain>() {
+    }
+
+    // unique_cf used to have inherit semantics on its default constructor.
```
ms-jihua: >used to have [](start = 17, length = 12)

nit: not sure this comment makes sense considering it still does

DHowett-MSFT: Using the past tense since it's a "compatibility shim" now, and i'm documenting its prior/intended behaviour. I can clear that up if necessary :)

ms-jihua:
(Is this a CC proposal, or did you have an impl change in mind to enforce this?)

I think you picked a bad example since CFStringCreate(...) implies its return type fairly well, but I think I do agree? Not all functions will clearly indicate their return type by name, and I think it's best to not surprise someone searching for instances of CFStringRef that this gets skipped over.

---
In reply to: [272517505](https://github.com/Microsoft/WinObjC/pull/1700#issuecomment-272517505) [](ancestors = 272517505)

DHowett-MSFT:
CC proposal, no impl change. I agree CFStringCreate was a bad example! Maybe more like, `auto x = Make(CFLocaleCopyName)`

include/xplat/Starboard/SmartTypes.h
ms-jihua: Please do. I think that someone lacking the current context may fail to parse this correctly.

DHowett-MSFT:
Merged via squash and rebase (for cherry-pick.)

PR #1701, opened 2017-01-13
Fix UILabel linebreak is not honhored/applied issue and also two regressions in UITextField by my earlier commits

yiyang-msft:
This change contains
1. Apply linebreak semantics to xaml textblock contained in UILabel. best effort to match the semantics to xaml semantics. there are known gaps such as midddleTruncation/HeadTruncation

2. Fix xib2nib tool on UILabel which failed to serialize the lineBreak property earlier

3. Fix two regressions on UITextFields (float round errors and null check groundbackcolor)

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1701)


tools/vsimporter/xib2nib/UILabel.cpp
```diff
@@ -94,20 +100,49 @@ void UILabel::InitFromStory(XIBObject* obj) {
         _numberOfLines = atoi(getAttrAndHandle("numberOfLines"));
     }
 
-    if (getAttrib("textAlignment")) {
-        const char* pAlign = getAttrib("textAlignment");
+    const char* lineBreakModeAttributeString = "lineBreakMode";
+    const char* lineBreakModeAttributeValue = getAttrib(lineBreakModeAttributeString);
```
oliversa-msft: >getAttrib [](start = 46, length = 9)

Why not just call getAttrAndHandle here?

tools/vsimporter/xib2nib/UILabel.cpp
```diff
@@ -94,20 +100,49 @@ void UILabel::InitFromStory(XIBObject* obj) {
...
+            getAttrAndHandle(lineBreakModeAttributeString);
+        }
+        else if (strcmp(lineBreakModeAttributeValue, "tailTruncation") == 0) {
+            _lineBreakMode = 4;
+            getAttrAndHandle(lineBreakModeAttributeString);
+        }
+        else if (strcmp(lineBreakModeAttributeValue, "middleTruncation") == 0) {
+            _lineBreakMode = 5;
+            getAttrAndHandle(lineBreakModeAttributeString);
+        }
```
oliversa-msft: >} [](start = 8, length = 1)

Should we print out a warning message that we extracted an unexpected linebreakMode setting?

tools/vsimporter/xib2nib/UILabel.cpp
```diff
@@ -153,6 +188,10 @@ void UILabel::ConvertStaticMappings(NIBWriter* writer, XIBObject* obj) {
         AddInt(writer, "UINumberOfLines", _numberOfLines);
     }
 
+    if (_lineBreakMode != 4) {
```
jaredhms: >4 [](start = 26, length = 1)

nit: can we pull '4' out into a constant that's used in all places?

Frameworks/UIKit/XamlUtilities.h
```diff
@@ -82,3 +82,6 @@ UIView* GenerateUIKitControlFromXamlType(RTObject* xamlObject);
 // We need a type-safe way to do this with projections.  This is copied verbatim from the projections
 // code and works perfectly for this limited usage, but we don't do any type validation below.
 id CreateRtProxy(Class cls, IInspectable* iface);
+
+// apply LineBreakMode on xaml TextBock
+void ApplyLineBreakModeOnTextBlock(WXCTextBlock* textBlock, UILineBreakMode mode, int numberOfLines);
```
jaredhms: >ApplyLineBreakModeOnTextBlock [](start = 5, length = 29)

not related to change, we should put these all into a XamlUtilities:: namespace so it's more clear at the call site.  XamlUtilities::ApplyLineBreakModeOnTextBlock would be more clear.

Frameworks/UIKit/XamlUtilities.mm
```diff
@@ -323,3 +327,43 @@ id CreateRtProxy(Class cls, IInspectable* iface) {
 
     return ret;
 }
+
+void ApplyLineBreakModeOnTextBlock(WXCTextBlock* textBlock, UILineBreakMode mode, int numberOfLines) {
+    if (numberOfLines == 1) {
+        textBlock.textWrapping = WXTextWrappingNoWrap;
+    } else {
+        textBlock.textWrapping = WXTextWrappingWrap;
+    }
```
jaredhms: nit: a small comment on this would be helpful

samples/XAMLCatalog/XAMLCatalog/UILabelViewController.m
```diff
@@ -0,0 +1,125 @@
...
+                                                 text:@"chararacterWrap is not supported, treated as wordWrap"
+                                             textAlignment:UITextAlignmentLeft
+                                         lineBreakMode:UILineBreakModeCharacterWrap
+                                         numnberOfLines:0]];
+
+    [_labels addObject:[self _createUILabelWithColor:[UIColor blueColor]
+                                                 text:@"Clip and wrapping, the string is wrapped and clipped at the end of second line if the string is long enough"
+                                             textAlignment:UITextAlignmentLeft
+                                         lineBreakMode:UILineBreakModeClip
+                                         numnberOfLines:0]];
```
jaredhms: >0 [](start = 56, length = 1)

any tests for numberOfLines > 1?

samples/XAMLCatalog/XAMLCatalog/UILabelViewController.m
```diff
@@ -0,0 +1,125 @@
...
+@implementation UILabelViewController {
+@private
+    NSMutableArray* _labels;
+}
+
+- (UILabel*)_createUILabelWithColor:(UIColor*)color
+                            text:(NSString *)text
+                            textAlignment:(NSTextAlignment)alignment
+                        lineBreakMode:(UILineBreakMode)lineBreakMode
+                        numnberOfLines:(NSInteger )numnberOfLines {
```
jaredhms: >numnberOfLines [](start = 24, length = 14)

nit: typo across this file

tools/vsimporter/xib2nib/UILabel.cpp
yiyang-msft: because it might be a value that is handled. as you second comment pointed out, i added a warning message in such case.

---
In reply to: [95921835](https://github.com/Microsoft/WinObjC/pull/1701#discussion_r95921835) [](ancestors = 95921835)

tools/vsimporter/xib2nib/UILabel.cpp
yiyang-msft: as much as I like to do that, but only define 4 as a const didn't really make the readability better because we still have other values like 0/1/2/3 for linkBreak in UILabel. and the same property can be used elsewhere, like UIButton. What should happen is that we define enums like UIKit implementation does so that meaningful enums can be used across the board. but this is out of scope of this change. let me know if you disagree.

---
In reply to: [96054766](https://github.com/Microsoft/WinObjC/pull/1701#discussion_r96054766) [](ancestors = 96054766)

jaredhms: ok; can you file a bug to terry about this though?

---
In reply to: [96535041](https://github.com/Microsoft/WinObjC/pull/1701#discussion_r96535041) [](ancestors = 96535041,96054766)

samples/XAMLCatalog/XAMLCatalog/UILabelViewController.m
yiyang-msft: 0 is for this purpose. but i change some 0 to value > 1 to make that clear

---
In reply to: [96056109](https://github.com/Microsoft/WinObjC/pull/1701#discussion_r96056109) [](ancestors = 96056109)

jaredhms: it does seem like they should be separate tests.  does '0' get any special handling vs. a specific line value such as '3'?

---
In reply to: [96539835](https://github.com/Microsoft/WinObjC/pull/1701#discussion_r96539835) [](ancestors = 96539835,96056109)

tools/vsimporter/xib2nib/UILabel.cpp
yiyang-msft: will file one before push.

---
In reply to: [96538840](https://github.com/Microsoft/WinObjC/pull/1701#discussion_r96538840) [](ancestors = 96538840,96535041,96054766)

jaredhms:
:shipit:

PR #1705, opened 2017-01-13
CoreText Performance: Call ID2D1RenderTarget::BeginDraw()/EndDraw() f…

### LINKED ISSUE #1620: Rendertarget EndDraw() calls are expensive. Should batch these through appropriate points.

ms-jihua:
…ewer times

Fixes #1620

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1705)


ms-jihua:
My profiler has been acting up since I took a recent update, so I haven't been able to get a measurement yet. Will make one when I can.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1238,3 +1238,45 @@ void CGContextDrawGlyphRun(CGContextRef ctx, const DWRITE_GLYPH_RUN* glyphRun) {
...
+
+void _CGContextPopEndDraw(CGContextRef ctx) {
+    if (--(ctx->_beginEndDrawDepth) == 0) {
+        ID2D1RenderTarget* imgRenderTarget = ctx->Backing()->DestImage()->Backing()->GetRenderTarget();
+        THROW_HR_IF_NULL(E_UNEXPECTED, imgRenderTarget);
+        THROW_IF_FAILED(imgRenderTarget->EndDraw());
+    }
+}
+
+void _CGContextSafeInnerBeginDraw(CGContextRef ctx) {
```
rajsesh-msft: Where are these two used, and what is their purpose?

Frameworks/UIKit/UIView.mm
```diff
@@ -2252,12 +2251,16 @@ - (CGPoint)convertPoint:(CGPoint)pos fromView:(UIView*)fromView {
 */
 - (void)drawLayer:(CALayer*)layer inContext:(CGContextRef)context {
     UIGraphicsPushContext(context);
+    _CGContextPushBeginDraw(context);
```
rajsesh-msft: This should really go into CALayer display, which displays a layer for one frame.

rajsesh-msft: Also into UIGraphicsPushContext() and PopContext().

ms-jihua: I had the same idea, but it ended up regressing XAMLCatalog (Custom text view would not update at all).

---
In reply to: [96086902](https://github.com/Microsoft/WinObjC/pull/1705#discussion_r96086902) [](ancestors = 96086902)

ms-jihua: Contexts that go onto the UIGraphics context stack cannot actually be safely assumed to only need their results after PopContext().

---
In reply to: [96087015](https://github.com/Microsoft/WinObjC/pull/1705#discussion_r96087015) [](ancestors = 96087015)

Frameworks/CoreGraphics/CGContext.mm
ms-jihua: Not used in this commit, but are needed for CGD2D. it seemed vaguely more appropriate to add them now than to add them when merging to CGD2D, but i don't feel strongly about this.

As an example, CGD2D has __CGContext::DrawToCommandList - it switches render targets in the middle and switches back at the end. Retargeting is illegal if in the middle of a Begin/EndDraw pair, so we need an alternate mechanism to force an EndDraw so we can retarget, then BeginDraw again. (I made a mistake by putting the second command in the body of this function though, oops) According to @DHowett-MSFT, shadows and transparency layers also require a retarget.

---
In reply to: [96086173](https://github.com/Microsoft/WinObjC/pull/1705#discussion_r96086173) [](ancestors = 96086173)

samples/XAMLCatalog/XAMLCatalog/CustomTextControlViewController.m
```diff
@@ -874,7 +874,7 @@ - (void)viewDidLoad {
         textEdit.autoresizingMask = infoLabel.autoresizingMask = focusButton.autoresizingMask =
             UIViewAutoresizingFlexibleLeftMargin | UIViewAutoresizingFlexibleRightMargin | UIViewAutoresizingFlexibleBottomMargin;
 
-        infoLabel.text = @"Click the control to gain focus. Use the keyboard to type and caret nagivate. Hold shift to update the "
+        infoLabel.text = @"Click the control to gain focus. Use the keyboard to type and caret navigate. Hold shift to update the "
```
aballway: +1

Frameworks/include/CGContextInternal.h
```diff
@@ -57,6 +62,9 @@ class __CGContext : private objc_object {
     float scale;
     CGContextImpl* _backing;
 
+    // Reduces the number of BeginDraw() and EndDraw() calls needed, by counting in a stack-like manner
+    uint32_t _beginEndDrawDepth = 0;
```
aballway: Should we have a mutex for updating this?

ms-jihua: i'll std::atomic it

---
In reply to: [96093839](https://github.com/Microsoft/WinObjC/pull/1705#discussion_r96093839) [](ancestors = 96093839)

Frameworks/UIKit/UIView.mm
rajsesh-msft: Fair enough.  We should still do this in CALayer display.

ms-jihua: Er, that is, putting it into [CALayer display] broke things. I could go investigate if you'd like, but I'm not sure how fruitful that'd be - we'd probably have to keep this out of CALayer display anyway.

---
In reply to: [96095561](https://github.com/Microsoft/WinObjC/pull/1705#discussion_r96095561) [](ancestors = 96095561)

Frameworks/CoreGraphics/CGContextCairo.mm
```diff
@@ -105,6 +105,11 @@
 // An OSS request has been submitted. The link for the request is:
 // https://osstool.microsoft.com/palamida/RequestDetails.htm?rid=40072&projectId=1
 void CGContextCairo::_cairoImageSurfaceBlur(cairo_surface_t* surface) {
+    // TODO #1635: It's unsafe to edit the bitmap not through D2D1RenderTarget, during a BeginDraw()
```
ms-jihua: These are a fairly significant hit to performance for the non-text CGContextCairo drawing functions, eating up >50% of time in CGContextFillRect, for instance. This may be unavoidable until the CGD2D merge, and is a tradeoff - text-drawing is still a much bigger bottleneck (CGContextFillRect, while slower, still does not take up much time). Please comment if you have any objections.

Frameworks/CoreGraphics/CGContextCairo.mm
```diff
@@ -105,6 +105,11 @@
...
+    // Escape any Begin/EndDraw() pairs for the lifetime of this function (to be removed during the CGD2D merge)
+    _CGContextEscapeBeginEndDrawStack(_rootContext);
```
aballway: nit: as much as I hate them, this would make a nice macro

ms-jihua: tried it, but it obscures too much for my taste.

---
In reply to: [96971228](https://github.com/Microsoft/WinObjC/pull/1705#discussion_r96971228) [](ancestors = 96971228)

DHowett-MSFT:
:shipit:

PR #1715, opened 2017-01-17
Updating to angle version 2.1.13

rajsesh-msft:
Updating to use angle 2.1.13.  According to the angle team, although 2.1.6 had some support for OpenGL ES 3.0, it is much better supported in the newer version.  Validated the Open GL test apps, which seem to run fine.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1715)


jaredhms:
:shipit:

PR #1717, opened 2017-01-17
[CGD2D] Add woc::AutoCF<T>. It's like AutoId<T> for CFTypes.

DHowett-MSFT:
**THIS IS ADVISORY ONLY**

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1717)


DHowett-MSFT:
Running builds for validation's sake.

PR #1728, opened 2017-01-18
Remove freetype library and references

### LINKED ISSUE #1455: remove freetype library from winobjc
Now that Core text is completely on Dwrite, there is no need for freetype to be bundled.

aballway:
Fixes #1455

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1728)


aballway:
Closing due to dependency on freetype coming from cairo, will remove freetype after CGD2D merges in and we no longer require cairo

PR #1505, opened 2016-12-05
Add additional UIControl events:

davelamb:
TouchUpOutside
TouchDragInside
TouchDragOutside
TouchDragEnter
TouchDragExit

Test added in ButtonsViewController in WOCCatalog.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1505)

PR #1731, opened 2017-01-18
Add missing UIControl events 

jaredhms:
This is a modified version of #1505, rebased onto develop with some additional fixes applied.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1731)


Frameworks/UIKit/UIControl.mm
```diff
@@ -420,7 +428,26 @@ - (void)touchesMoved:(NSSet*)touchSet withEvent:(UIEvent*)event {
...
+        if (currentTouchInside != _touchInside) {
+            _touchInside = currentTouchInside;
+            if (currentTouchInside) {
+                [self sendActionsForControlEvents:UIControlEventTouchDragEnter];
+            } else {
+                [self sendActionsForControlEvents:UIControlEventTouchDragExit];
+            }
+        }
+
+        // TODO: Investigate fat-fingers
```
yiyang-msft: > Investigate fat-fingers [](start = 16, length = 24)

what does this mean?

Frameworks/UIKit/UIControl.mm
```diff
@@ -534,12 +575,10 @@ - (void)_backPressed {
  @Status Interoperable
 */
 - (void)dealloc {
```
yiyang-msft: > (void)dealloc { [](start = 1, length = 16)

nit: can be removed?

Frameworks/UIKit/UIControl.mm
jaredhms: it means we may be a bit too precise on our drag in/out (considering a slack value for a touch point vs. a single mouse point), but I think this is fine for now

---
In reply to: [96942924](https://github.com/Microsoft/WinObjC/pull/1731#discussion_r96942924) [](ancestors = 96942924)

Frameworks/UIKit/UIButton.mm
```diff
@@ -617,6 +618,25 @@ - (void)touchesMoved:(NSSet*)touchSet withEvent:(UIEvent*)event {
...
+    } else {
+        newState = _curState & ~UIControlStateHighlighted;
+    }
+
+    // Relayout when new state is different than old state
+    if(_curState != newState) {
+        _curState = newState;
+        [self setNeedsDisplay];
+        [self setNeedsLayout];
+    }
```
yiyang-msft: should this be moved into UIControl?

jaredhms: I don't think so - only Button knows whether or not it needs to redisplay/relayout.

---
In reply to: [96943884](https://github.com/Microsoft/WinObjC/pull/1731#discussion_r96943884) [](ancestors = 96943884)

yiyang-msft: we can call [self setHightlighted:currentTouchInside] to reuse some logic in UIControl.

---
In reply to: [96944367](https://github.com/Microsoft/WinObjC/pull/1731#discussion_r96944367) [](ancestors = 96944367,96943884)

yiyang-msft:
:shipit:

Frameworks/UIKit/UIButton.mm
jaredhms: great idea; let me give that a shot

---
In reply to: [96945443](https://github.com/Microsoft/WinObjC/pull/1731#discussion_r96945443) [](ancestors = 96945443,96944367,96943884)

PR #1742, opened 2017-01-19
Fix CoreText extra kerning for the first run in a frame

### LINKED ISSUE #1696: Regression in CoreText ExtraKerning
Extra Kerning no longer functions

aballway:
When we changed the Core Text code to optimize out setting the font twice on a frame I forgot that kerning was also set in the loop.  Updates images for post #1723 drawing.

Fixes #1696

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1742)

### DIFF
--- a/Frameworks/CoreText/DWriteWrapper_CoreText.mm
+++ b/Frameworks/CoreText/DWriteWrapper_CoreText.mm
@@ -265,13 +265,7 @@ static HRESULT __DWriteTextLayoutCreate(CFAttributedStringRef string, CFRange ra
     // Used to separate runs for attributes which DWrite does not handle until drawing (e.g. Foreground Color)
     uint32_t incompatibleAttributeFlag = 0;
     CFRange attributeRange;
-
-    // Find the range of the first set of attributes and skip it, since the underlying DWriteTextFormat has already internalized it
-    // If this first set of attributes lasts the entire range, the below for loop is not executed at all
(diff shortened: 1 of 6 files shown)


Frameworks/CoreText/DWriteWrapper_CoreText.mm
```diff
@@ -265,13 +265,7 @@ static HRESULT __DWriteTextLayoutCreate(CFAttributedStringRef string, CFRange ra
...
     uint32_t incompatibleAttributeFlag = 0;
     CFRange attributeRange;
-
-    // Find the range of the first set of attributes and skip it, since the underlying DWriteTextFormat has already internalized it
-    // If this first set of attributes lasts the entire range, the below for loop is not executed at all
-    // attributeRange is populated even if this attribute is not found
-    CFAttributedStringGetAttribute(string, range.location, kCTFontAttributeName, &attributeRange);
-
-    for (CFIndex index = attributeRange.location + attributeRange.length; index < rangeEnd; index += attributeRange.length) {
+    for (CFIndex index = range.location; index < rangeEnd; index += attributeRange.length) {
```
DHowett-MSFT: Why's it not necessary to do the thing this comment said it was doing now?

aballway: It is necessary.  If we don't do it then extra-kerning won't be set for the first attribute range.

Issue #1733, opened 2017-01-19
CGContextClearRect puts context in an invalid state


msft-Jeyaram:
The operation fails and you do not see the results until EndDraw is called (as per documentation)

Upon context entering the invalid state, further calls will invoke FAIL_FAST, thus failure on start up.
Thus applications that are attempting to do this are affected (many of our *sample* applications)

PR #1743, opened 2017-01-19
#1733 CGContextClearRect: Support to clear rect if clipping is not set

msft-Jeyaram:
The current functionality is not implemented properly.
CGContextClearRectwill end up putting the context in an invalid state and any subsequent CGContext[d2d] calls will end up failing. Any application (some of our samples) that uses this call on application init will end up crashing due to failure.

This will unblock our merge to develop, tested with 'sample' applications. They start up and function as expected.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1743)


Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -2068,14 +2068,19 @@ void CGContextShowGlyphsWithAdvances(CGContextRef context, const CGGlyph* glyphs
 
 #pragma region Drawing Operations - Basic Shapes
 /**
- @Status Interoperable
+ @Status Caveat
+ @Notes only supports the scenario where clipping is not set.
```
DHowett-MSFT: Also,
* Only works on untransformed coordinates. The cleared region will be cleared in device space.

Frameworks/CoreGraphics/CGContext.mm~RF70a69f.TMP
```diff
@@ -0,0 +1,2922 @@
+//******************************************************************************
```
DHowett-MSFT: ditch this file

aballway: remove this file

msft-Jeyaram: uhhhh I do not like TMP files

MSFTFox:
:shipit:

PR #1745, opened 2017-01-20
Porting ObjectiveC language services VSIX to VS2017

mukhole:
objc_language-services.vsix has syntax highlighting support which is not needed in VS2017.
The installation targets on this vsix do not cover VS2017.
Our VSIX packaging, does not allow us to have a single VSIX, as buildmonitor and syntax highlighter are added as assets into the actual VSIX, and there is no version check on assets, so Syntax highlighter can get installed on VS2017.

To avoid this the old project files and VSIXManifest is moved to folder legacy, and the original project/solution is migrated to VS2017.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1745)


DHowett-MSFT:
:shipit:

pradipd:
:shipit:

mukhole:
Ping,
Started a new iteration.
Natvis file is removed from VS2017 VSIX.
The contents of the natvis from VS 2017 are being merged with those of VS2015 (just NSObject was required).

DHowett-MSFT:
:shipit:

PR #1748, opened 2017-01-20
Fix our WIC<->CG color format table again.

### LINKED ISSUE #1694: BitmapFormats tests are incorrect and might not validate the results due to uninitialized for loop controller.
The values passed to the bitmapcontextCreate are incorrect and 

The main loop validate code has an uninitialized indexer.

` for (size_t i; i < data.size(); ++i)`

this led to some interesting cases :(

DHowett-MSFT:
Fixes #1694.

This pull request fixes a few issues:
* Bitmap context creation (and image creation) never respected byte order.
* Our little/big endian pixel buffers were swapped.
* The bitmap context test didn't think alpha was premultiplied on the **skip** alpha formats. It turns out that it is!
* A few of our tests were relying on the "default" byte order with an RGBA fallback.
* The RGBA fallback would have led to memory corruption down the line.
* Our "default" endianness specifier was being ignored.
  * Now we make it match a WIC-appropriate endianness. Any consumers relying on "default" to be any particular format should not have been doing so. Including us!
* Some of our WIC->CG pixel formats _still_ had incorrect bpp values.
* It turns out that the reference platform will return a "Default" pixel format even though it never should. For 32bpp RGB, that means "ARGB".
* The format flips were backwards in the testing harness too.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1748)

### DIFF
--- a/Frameworks/CoreGraphics/CGImage.mm
+++ b/Frameworks/CoreGraphics/CGImage.mm
@@ -257,7 +257,9 @@ CGImageRef CGImageCreate(size_t width,
     ComPtr<IWICImagingFactory> imageFactory;
     RETURN_NULL_IF_FAILED(_CGGetWICFactory(&imageFactory));
 
-    REFGUID pixelFormat = _CGImageGetWICPixelFormatFromImageProperties(bitsPerComponent, bitsPerPixel, colorSpace, bitmapInfo);
+    GUID pixelFormat;
+    RETURN_NULL_IF_FAILED(
+        _CGImageGetWICPixelFormatFromImageProperties(bitsPerComponent, bitsPerPixel, colorSpace, bitmapInfo, &pixelFormat));
 
     unsigned char* bytes = static_cast<unsigned char*>(const_cast<void*>(_CGDataProviderGetData(provider)));
     RETURN_NULL_IF_FAILED(
@@ -333,8 +335,9 @@ CGImageRef CGImageMaskCreate(size_t width,
     RETURN_NULL_IF_FAILED(_CGGetWICFactory(&imageFactory));
 
     woc::unique_cf<CGColorSpaceRef> colorSpace(CGColorSpaceCreateDeviceGray());
-    REFGUID pixelFormat =
-        _CGImageGetWICPixelFormatFromImageProperties(bitsPerComponent, bitsPerPixel, colorSpace.get(), kCGBitmapByteOrderDefault);
+    GUID pixelFormat;
+    RETURN_NULL_IF_FAILED(
+        _CGImageGetWICPixelFormatFromImageProperties(bitsPerComponent, bitsPerPixel, colorSpace, kCGBitmapByteOrderDefault, &pixelFormat));
 
     unsigned char* bytes = static_cast<unsigned char*>(const_cast<void*>(_CGDataProviderGetData(provider)));
     RETURN_NULL_IF_FAILED(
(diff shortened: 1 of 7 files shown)


Frameworks/CoreGraphics/CGImage.mm
```diff
@@ -706,98 +710,92 @@ CGImageRef _CGImageLoadImageWithWICDecoder(REFGUID decoderCls, void* bytes, int
...
-                                                     unsigned int bitsPerPixel,
-                                                     CGColorSpaceRef colorSpace,
-                                                     CGBitmapInfo bitmapInfo) {
+// key structure
+// [8b colorspace model] [8b bpp] [16b bitmap info] = 32b
+#define PACK_KEY(colorSpaceModel, bpp, byteOrder, alpha) \
+    (((colorSpaceModel & 0xFF) << 24) | ((bpp & 0xFF) << 16) | ((alpha | byteOrder) & 0xFFFF))
+
+// clang-format off
+static std::map<uint32_t, WICPixelFormatGUID> s_CGThings{
```
DHowett: You seem to have left this named `s_CGThings`. Probably don't do that.

Frameworks/CoreGraphics/CGImage.mm
```diff
@@ -533,8 +536,9 @@ CGImageRef CGImageCreateWithMask(CGImageRef image, CGImageRef mask) {
     size_t width = CGImageGetWidth(image);
     size_t height = CGImageGetHeight(image);
 
-    woc::unique_cf<CGContextRef> context{ CGBitmapContextCreate(
-        nullptr, width, height, 8, width * 4, CGImageGetColorSpace(image), kCGImageAlphaPremultipliedFirst | kCGBitmapByteOrder32Big) };
+    woc::unique_cf<CGContextRef> context{
+        CGBitmapContextCreate(nullptr, width, height, 8, width * 4, CGImageGetColorSpace(image), kCGImageAlphaPremultipliedLast)
```
msft-Jeyaram: default byte order then?

Frameworks/CoreGraphics/CGImage.mm
```diff
@@ -706,98 +710,95 @@ CGImageRef _CGImageLoadImageWithWICDecoder(REFGUID decoderCls, void* bytes, int
...
-REFGUID _CGImageGetWICPixelFormatFromImageProperties(unsigned int bitsPerComponent,
-                                                     unsigned int bitsPerPixel,
-                                                     CGColorSpaceRef colorSpace,
-                                                     CGBitmapInfo bitmapInfo) {
+// CG packed format key
+//  |Color  |bits/px|CGBitmapInfo   |
+//  |-------|-------|---------------|
+// 32      24      16               0
+#define CG_FORMAT_KEY(colorSpaceModel, bpp, byteOrder, alpha) \
+    (((colorSpaceModel & 0xFF) << 24) | ((bpp & 0xFF) << 16) | ((alpha | byteOrder) & 0xFFFF))
```
msft-Jeyaram: >(((colorSpaceModel & 0xFF) << 24) | ((bpp & 0xFF) << 16) | ((alpha | byteOrder) & 0xFFFF)) [](start = 4, length = 90)

that is an awesome key :D

Frameworks/CoreGraphics/CGImage.mm
```diff
@@ -706,98 +710,95 @@ CGImageRef _CGImageLoadImageWithWICDecoder(REFGUID decoderCls, void* bytes, int
...
-        if (bitsPerPixel == 1) {
-            return GUID_WICPixelFormat1bppIndexed;
-        } else if (bitsPerPixel == 2) {
-            return GUID_WICPixelFormat2bppIndexed;
-        } else if (bitsPerPixel == 4) {
-            return GUID_WICPixelFormat4bppIndexed;
-        } else if (bitsPerPixel == 8) {
-            return GUID_WICPixelFormat8bppIndexed;
+    unsigned int byteOrder = bitmapInfo & kCGBitmapByteOrderMask;
+    unsigned int formatImputedBpp = 0;
```
msft-Jeyaram: formatInputedBpp?

PR #1749, opened 2017-01-20
Address feedback and merge issues on packaging branch

bbowman:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1749)


pradipd:
> using System.Reflection;

Are all these files add's or moves?
It would be nice to not lose the git history.

---
Refers to: tools/BuildMonitor/Properties/AssemblyInfo.cs:1 in fb72438. [](commit_id = fb72438ebdccec75fe0466d2dd256c790fcc594d, deletion_comment = False)

bbowman:
I did a git move. I don't know if there is a better way. The BuildMonitor stuff accidentially got deleted so it probably shows up as an add? Without redoing the packaging branch *again* I don't know what else I can do.


tools/VSIX/source.extension.vsixmanifest
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+<PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
+  <Metadata>
+    <Identity Id="VSIX..9c35fff1-f084-44c1-a38e-68c707163aa2" Version="1.5" Language="en-US" Publisher="Microsoft" />
+    <DisplayName>Objective-C Language Service</DisplayName>
+    <Description xml:space="preserve">This package contains Objective-C language service components such as syntax higlighting etc</Description>
+  </Metadata>
+  <Installation>
+    <InstallationTarget Version="[12.0,15.0)" Id="Microsoft.VisualStudio.Community" />
```
rajsesh-msft: >12 [](start = 34, length = 2)

Is the version number here still correct?

rajsesh-msft:
> <?xml version="1.0" encoding="utf-8"?>

Given that we will switch over to VS2017 very soon(ish), does it make sense to simply delete the OBJC highlighting extension?

---
Refers to: tools/ObjectiveC/app.config:1 in a17b5d6. [](commit_id = a17b5d643323b272f05f40ece3fde633d532cbc1, deletion_comment = False)

tools/BuildMonitor/BuildEventProcessor.cs
```diff
@@ -0,0 +1,362 @@
...
+
+            _processorStopwatch = new Stopwatch();
+        }
+
+        #region Public Methods
+        public bool HasIslandwoodProjects()
+        {
+            foreach (Project project in _dte.Solution.Projects)
+            {
+                if (project.Kind.ToString().ToLowerInvariant().Equals(BuildEventProcessor.WinObjCProjectType))
```
rajsesh-msft: I suppose this is the new code in this vsix (or changed?).  This will require some eyes from the quality folks..

memontic-ms:
Remove /build/Package.sln.

memontic-ms:
WinObjC.Language packages up `WinObjC/tools/msvc/*` into `build/msvc/*`
Whereas WinObjC.Tools packages up `WinObjC/tools/WinObjC.Tools/vsimporter-templates/*` into` build/msvc/vsimporter-templates/*`

It seems like it would make more sense to either:
1. Put the current contents of `WinObjC/tools/msvc/*` into `WinObjC/tools/WinObjC.Language/msvc/*`
2. Relocate the vsimporter-templates directory back to `WinObjC/tools/msvc/vsimporter-templates`

To cut down on vsimporter changes I'd prefer number 2.
(vsimporter currently uses a relative path of ../msvc/ to find the vsimporter-templates folder. That only works once it is packaged and won't work from its output path of WinObjC/tools/$(Configuration)/)

memontic-ms:
Not sure if it was intentional, but why isn't Starboard.h making it into any of the packages? 
Not positive why I wasn't hitting this error before but trying to build a winrt project will error telling me that RTHelpers.h (from WinObjC.Frameworks.UWP package) can't find Starboard.h.

memontic-ms:
libxml2.dll is being packaged twice:
winobjc.frameworks: build\deps\prebuilt\Universal Windows\x86\
winobjc.frameworks.core: build\deps\prebuilt\Universal Windows\x86\

Issue #1194, opened 2016-10-20
Recreate render target if EndDraw fails


rajsesh-msft:
This is necessary before we switch CALayer to hw render target, so this doesn't block CGD2D merge to develop.  

https://msdn.microsoft.com/en-us/library/windows/desktop/dd371766(v=vs.85).aspx,

PR #1750, opened 2017-01-20
Merge #1705 to CGD2D

### LINKED ISSUE #1635: Merge #1620 to CGD2D and use it for all draw calls

ms-jihua:
- Add PushBeginDraw/PopEndDraw pairs to UISegment, UIImage
   (not on develop since this likely would've actually hurt performance there)
 - Add Escape/Unescape pairs to CGContext areas where the target is changed

Fixes #1635

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1750)


Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -2138,8 +2174,9 @@ void CGContextClearRect(CGContextRef context, CGRect rect) {
     ComPtr<ID2D1CommandList> commandList;
     RETURN_IF_FAILED(deviceContext->CreateCommandList(&commandList));
 
-    deviceContext->BeginDraw();
+    EscapeBeginEndDrawStack();
     deviceContext->SetTarget(commandList.Get());
```
ms-jihua: @DHowett-MSFT I think this is more correct here, right?

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -2138,8 +2174,9 @@ void CGContextClearRect(CGContextRef context, CGRect rect) {
     ComPtr<ID2D1CommandList> commandList;
     RETURN_IF_FAILED(deviceContext->CreateCommandList(&commandList));
 
-    deviceContext->BeginDraw();
+    EscapeBeginEndDrawStack();
```
ms-jihua: Because of the need to escape before changing targets, and because most functions in this file currently filter down to this function and thus change targets, #1705 has close-to-zero or probably negative impact on CGD2D as is. Since we don't need a new command list each time this is called, I am going to add another iteration to this PR, tying the creation of new command lists to the stack.

Frameworks/UIKit/UISegment.mm
```diff
@@ -224,25 +225,30 @@ - (id)drawRect:(CGRect)inRect {
...
-                CGContextSetFillColorWithColor(UIGraphicsGetCurrentContext(), [_tintColor CGColor]);
-                CGContextFillRect(UIGraphicsGetCurrentContext(), rect);
+                CGContextSetFillColorWithColor(ctx, [_tintColor CGColor]);
+                CGContextFillRect(ctx, rect);
             }
         }
     } else {
         if (!_noDefaultImages) {
-            CGContextRef ctx = UIGraphicsGetCurrentContext();
+            CGContextRef ctx = ctx;
```
aballway: nit: not needed

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -369,6 +377,30 @@ inline bool ShouldDraw() {
...
+
+    inline void PopEndDraw() {
+        if (--(_beginEndDrawDepth) == 0) {
+            THROW_IF_FAILED(deviceContext->EndDraw());
+        }
+    }
+
+    inline void EscapeBeginEndDrawStack() {
+        if ((_beginEndDrawDepth > 0) && ((_escapeBeginEndDrawDepth)++ == 0)) {
+            THROW_IF_FAILED(deviceContext->EndDraw());
```
msft-Jeyaram: >THROW_IF_FAILED [](start = 12, length = 15)

who is catching the throws.
We standardized on FAIL_FAST

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -369,6 +377,30 @@ inline bool ShouldDraw() {
...
 
+    inline void PushBeginDraw() {
+        if ((_beginEndDrawDepth)++ == 0) {
+            deviceContext->BeginDraw();
+        }
+    }
+
+    inline void PopEndDraw() {
+        if (--(_beginEndDrawDepth) == 0) {
+            THROW_IF_FAILED(deviceContext->EndDraw());
```
msft-Jeyaram: same here

msft-Jeyaram:
:shipit:

Frameworks/include/CGContextInternal.h
```diff
@@ -41,6 +41,22 @@ COREGRAPHICS_EXPORT bool CGContextIsPointInPath(CGContextRef c, bool eoFill, flo
...
+// Reduces the number of BeginDraw() and EndDraw() calls needed, by counting in a stack-like manner,
+// and only calling BeginDraw()/EndDraw() when the stack is empty/emptied
+COREGRAPHICS_EXPORT void _CGContextPushBeginDraw(CGContextRef ctx);
+COREGRAPHICS_EXPORT void _CGContextPopEndDraw(CGContextRef ctx);
+
+// If currently in a Begin/EndDraw stack, Escape will EndDraw(), Unescape will BeginDraw()
+// For scenarios where a Begin/EndDraw pair needs to be temporarily escaped, to be returned to at a later time
+// Ie:
+//      - Switching render targets - Illegal to do so if currently in a Begin/EndDraw pair
+//      - Cairo - ID2D1RenderTarget is considered to 'own' the bitmap during Begin/EndDraw,
```
rajsesh-msft: Can safely remove comment about cairo now.

Frameworks/UIKit/UIView.mm
```diff
@@ -2252,14 +2251,18 @@ - (CGPoint)convertPoint:(CGPoint)pos fromView:(UIView*)fromView {
...
+    auto popEnd = wil::ScopeExit([context]() {
+        _CGContextPopEndDraw(context);
+        UIGraphicsPopContext();
+    });
 
     CGRect bounds;
     // TODO(DH)
     bounds = self.bounds;
-    //bounds = CGContextGetClipBoundingBox(context);
+    // bounds = CGContextGetClipBoundingBox(context);
```
rajsesh-msft: nit: remove commented out line.  Also, the whole thing could be CGRect bounds = self.bounds, or even just [self drawRect:self.bounds] since it is not being reused.

build/CoreGraphics/dll/CoreGraphics.def
```diff
@@ -242,6 +242,10 @@ LIBRARY CoreGraphics
         _CGContextCreateWithD2DRenderTarget
         _CGGetD2DFactory
         _CGGetPixelFormatProperties
+        _CGContextPushBeginDraw
+        _CGContextPopEndDraw
+        _CGContextEscapeBeginEndDrawStack
```
rajsesh-msft: These names are not ideal - they are vague (what does escaping the stack mean), and reveal the implementation detail (there is a stack).  The later is not a big deal, but essentially what you doing here is ability to deal with a layer, so it might make sense to just call this _CGContextForceEndDraw() and _CGContextForceBeginDraw(), or _CGContextBeginLayerDraw/EndLayerDraw().

Frameworks/UIKit/UITextView.mm
```diff
@@ -759,7 +763,7 @@ - (CGSize)sizeThatFits:(CGSize)fitSize {
     centerRect.origin.y = 0;
     centerRect.size = fontExtent;
     // TODO(DH)
-    //EbrCenterTextInRectVertically(&centerRect, &fontExtent, _font);
+    // EbrCenterTextInRectVertically(&centerRect, &fontExtent, _font);
```
rajsesh-msft: remove commented code?  also looks like there is a missing TODO issue, lets create it now.

Frameworks/CoreGraphics/CGContext.mm
rajsesh-msft: The standardization we agreed on was to capture exceptions at the API boundary (which is usually the API entry point).  Failing fast in the leaf nodes will lead to code that just arbitrarily quits with no possible recovery.  Particularly when drawing to hardware device contexts, EndDraw() can fail, so we need to handle this with care.  See #1194.

Frameworks/CoreGraphics/CGContext.mm
DHowett-MSFT: I think you're correct. I recall Begin->Set->End being specified in one of the MSDN documents.

ms-jihua:
Then okay, let me just respond to the feedback on the table for now and get this actually performant later.

Frameworks/CoreGraphics/CGContext.mm
ms-jihua: I think I lied and it doesn't matter [https://msdn.microsoft.com/en-us/library/windows/desktop/hh404533(v=vs.85).aspx](https://msdn.microsoft.com/en-us/library/windows/desktop/hh404533(v=vs.85).aspx)

---
In reply to: [97687941](https://github.com/Microsoft/WinObjC/pull/1750#discussion_r97687941) [](ancestors = 97687941)

Frameworks/CoreGraphics/CGContext.mm
ms-jihua: Going to do this in a separate PR instead

---
In reply to: [96997659](https://github.com/Microsoft/WinObjC/pull/1750#discussion_r96997659) [](ancestors = 96997659)

rajsesh-msft: Please open a tracking issue to do this work.

Frameworks/UIKit/UITextView.mm
ms-jihua: @DHowett-MSFT is taking care of this

---
In reply to: [97468215](https://github.com/Microsoft/WinObjC/pull/1750#discussion_r97468215) [](ancestors = 97468215)

Frameworks/UIKit/UIView.mm
ms-jihua: @DHowett-MSFT is taking care of this

---
In reply to: [97467495](https://github.com/Microsoft/WinObjC/pull/1750#discussion_r97467495) [](ancestors = 97467495)

Frameworks/CoreGraphics/CGContext.mm
ms-jihua: #1810

---
In reply to: [97693686](https://github.com/Microsoft/WinObjC/pull/1750#discussion_r97693686) [](ancestors = 97693686)

Frameworks/CoreGraphics/CGContext.mm
ms-jihua: I'm not sure what you're actually advocating I do here, considering handling #1194 appears too large for the scope of this PR. As such, I'm going to take @msft-Jeyaram 's recommendation to follow the standard for now.

---
In reply to: [97468410](https://github.com/Microsoft/WinObjC/pull/1750#discussion_r97468410) [](ancestors = 97468410)

build/CoreGraphics/dll/CoreGraphics.def
ms-jihua: but they don't force an enddraw or a begindraw, nor are they exclusively tied to layers. they are 'enddraw if needed' and 'begindraw if needed', where 'needed' is kind of vague (if currently in the stack) and difficult to describe without mentioning the stack.

i agree that these names are bad, but the suggestions are less factually accurate, so i am won'tfixing this for now.

---
In reply to: [97468082](https://github.com/Microsoft/WinObjC/pull/1750#discussion_r97468082) [](ancestors = 97468082)

Frameworks/UIKit/UIView.mm
rajsesh-msft: @DHowett-MSFT, do we have an issue tracking this TODO?

build/CoreGraphics/dll/CoreGraphics.def
rajsesh-msft: I think BeginDrawIfNeeded(), EndDrawIfNeeded() are more descriptive than EscapeStack.  Won't fix is okay.

Frameworks/UIKit/UIView.mm
DHowett-MSFT: #1811

Frameworks/UIKit/UITextView.mm
DHowett-MSFT: This might not be necessary, but I'll add an issue to track "something about vertical text centering in UITextField"

This is more of a CT concern, but EbrCenterTextInRectVertically was part of CGContext for some unknown reason.

Frameworks/CoreGraphics/CGContext.mm
rajsesh-msft: The recommendation is wrong, we agreed not to FAIL_FAST at arbitrary points in the CG code and instead throw.  Any exceptions must be caught at the "C" entry point, where we decide either to return an error code/null or fail-fast if it is void.  

---
In reply to: [97695158](https://github.com/Microsoft/WinObjC/pull/1750#discussion_r97695158) [](ancestors = 97695158,97468410)

Frameworks/UIKit/UITextView.mm
DHowett-MSFT: #1812

Frameworks/CoreGraphics/CGContext.mm
rajsesh-msft: Although this is *only* one step away from the entry point.  

---
In reply to: [97696710](https://github.com/Microsoft/WinObjC/pull/1750#discussion_r97696710) [](ancestors = 97696710,97695158,97468410)

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -369,6 +377,30 @@ inline bool ShouldDraw() {
...
 
+    inline void PushBeginDraw() {
+        if ((_beginEndDrawDepth)++ == 0) {
+            deviceContext->BeginDraw();
+        }
+    }
+
+    inline void PopEndDraw() {
+        if (--(_beginEndDrawDepth) == 0) {
+            FAIL_FAST_IF_FAILED(deviceContext->EndDraw());
```
DHowett-MSFT: The _true_ recommendation was to have inner functions return `HRESULT` and FAIL_FAST or ignore the error gracefully at the C API boundary.

Frameworks/CoreGraphics/CGContext.mm
rajsesh-msft: I am going to approve this PR given that this is only one step away from the entry point.  

---
In reply to: [97696881](https://github.com/Microsoft/WinObjC/pull/1750#discussion_r97696881) [](ancestors = 97696881,97696710,97695158,97468410)

ms-jihua: i actually already made the change anyway

---
In reply to: [97697849](https://github.com/Microsoft/WinObjC/pull/1750#discussion_r97697849) [](ancestors = 97697849,97696881,97696710,97695158,97468410)

Issue #1497, opened 2016-12-02
UILabel: [UILabel adjustTextLayerSize] does more work than necessary


jaredhms:
We should apply most of the properties to the _textBlock as they're set (within setText, setFont, etc.) rather than doing it within adjustTextLayerSize (which is called quite frequently by most of UILabel's setters).

yiyang-msft:
adjustTextLayerSize is being renamed as _updateXamlElement

jaredhms:
And now _updateXamlElement is being renamed to _updateXamlElementWithAdjustedFont.  Maybe we should fix this bug as part of the rename.

jaredhms:
Not sure what the correct fix is here, but doing minimal work in _updateXamlElementWithAdjustedFont is the goal.

yiyang-msft:
_updateXamlElementWithAdjustedFont is break out to smaller functions and setting correponding properties only calls into corresponding smaller functions. e.g., setTextColor only setting color and only calls setNeedsDisplay.

yiyang-msft:
fixed with 6fcf514c273d32104026c0a973a10ca84d36a5f3

PR #1768, opened 2017-01-21
Fixing Github issue 1602 and 1689 - sizeThatFits and adjustFontSiz eToFitWidth does not work on UILabel

yiyang-msft:
1. For sizeThatFits issue - we used a original font size to implement sizeThatFits, which is not the font we used on textbox, plus we really didn't get the constraint right earlier when calls into the sizeWithFont in order to get the right size given the font and linebreak mode settings. 

2. For adjustFontSizeToFitWidth issue: we used to always assume font is too big, so only shrink the font to fit, and we used to start to the cached orginal font size, which is very small, once it fits, we bail out, instead of increasing the font size to find a bigger size font that can fit.  We used to do linear search as well, now we do binary search, hopefully faster in most cases.

3. additionally Fix UITableViewCell layout issue when accessory view is a UILabel. This has been verified on reference platform, with the change, we have the same behavior as reference platform for tableviewcell accessory view when it is UILabel.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1768)

### DIFF
--- a/Frameworks/UIKit/UILabel.mm
+++ b/Frameworks/UIKit/UILabel.mm
@@ -44,7 +44,6 @@
     UILineBreakMode _lineBreakMode;
     BOOL _adjustFontSize;
     float _minimumFontSize;
-    float _originalFontSize;
     int _numberOfLines;
     BOOL _isDisabled;
     BOOL _isHighlighted;
@@ -52,53 +51,63 @@
     StrongId<WXCTextBlock> _textBlock;
 }
 
-- (void)adjustFontSizeToFit {
+- (void)_adjustFontSizeToFit {
     if (_numberOfLines == 0) {
-        [self adjustTextLayerSize];
+        [self _adjustTextLayerSize];
         return;
     }
 
-    float curFontSize = _originalFontSize;
-    _font = [_font fontWithSize:curFontSize];
-
-    CGRect rect;
-    rect = [self bounds];
-    if (rect.size.width == 0.0f || rect.size.height == 0.0f) {
+    CGRect rect = [self bounds];
+    if (rect.size.width == 0.0f || rect.size.height == 0.0f || _text == nil) {
         return;
     }
 
-    while (curFontSize > _minimumFontSize && curFontSize > 0.0f) {
+    // MaxFitFontSize records currently found maximum font size that fits the bounds,
+    // It starts with _minimumFontSize because it is smallest font size allowed by the UILabel
+    float maxFitFontSize = _minimumFontSize;
+
+    float curFontSize = [_font pointSize];
+    // Start our trial with fonts bigger than _minimumFontSize
+    if (curFontSize < maxFitFontSize) {
+        curFontSize = 2.0 * maxFitFontSize;
+        _font = [_font fontWithSize:curFontSize];
+    }
+
+    // Do binary search on maximum font size that fits the bound
+    do {
+        // Caculating rect size using font and line break settings
         CGSize size = CGSizeZero;
-
-        //  A single line of text should be clipped
         if (_numberOfLines == 1) {
-            size = [_text sizeWithFont:_font constrainedToSize:CGSizeMake(rect.size.width, 0.0f) lineBreakMode:UILineBreakModeClip];
+            size = [_text sizeWithFont:_font constrainedToSize:CGSizeMake(0.0f, 0.0f) lineBreakMode:UILineBreakModeClip];
         } else {
-            //  Multiple lines of text should be wrapped
             size = [_text sizeWithFont:_font constrainedToSize:CGSizeMake(rect.size.width, 0.0f) lineBreakMode:UILineBreakModeWordWrap];
         }
 
         if (size.width < rect.size.width && size.height <= rect.size.height) {
-            break;
+            // Update maxFitFontSize, and double current font size and retry
+            maxFitFontSize = curFontSize;
+            curFontSize *= 2.0f;
+        } else {
+            // Current font size does not fit
+            if (curFontSize - maxFitFontSize <= 1.0f) {
+                // Distance between current font size wwhich does nit fit
+                // and maximum font size which fits is less than 1.0, break out
+                break;
+            } else {
+                // Re-try with middle value between maxFitFontSize and current Font Size
+                curFontSize = (curFontSize + maxFitFontSize) / 2.0f;
+            }
         }
 
-        curFontSize -= 1.0f;
         _font = [_font fontWithSize:curFontSize];
-    }
+    } while (true);
 
-    if (curFontSize < _minimumFontSize) {
-        curFontSize = _minimumFontSize;
-    }
-    if (curFontSize < 0.0f) {
-        curFontSize = 1.0f;
-    }
+    _font = [_font fontWithSize:maxFitFontSize];
 
-    _font = [_font fontWithSize:curFontSize];
-
-    [self adjustTextLayerSize];
+    [self _adjustTextLayerSize];
 }
 
-- (void)adjustTextLayerSize {
+- (void)_adjustTextLayerSize {
     [_textBlock setText:_text];
     [_textBlock setFontSize:[_font pointSize]];
 
@@ -177,9 +186,9 @@
         }
 
         if (_adjustFontSize) {
-            [self adjustFontSizeToFit];
+            [self _adjustFontSizeToFit];
         } else {
-            [self adjustTextLayerSize];
+            [self _adjustTextLayerSize];
         }
     }
 
@@ -231,7 +240,7 @@
 
         _font = [UIFont fontWithName:@"Segoe UI" size:[UIFont labelFontSize]];
 
-        [self adjustTextLayerSize];
+        [self _adjustTextLayerSize];
     }
 
     return self;
@@ -250,7 +259,7 @@
 
         _font = [UIFont fontWithName:@"Segoe UI" size:[UIFont labelFontSize]];
 
-        [self adjustTextLayerSize];
+        [self _adjustTextLayerSize];
     }
 
     return self;
@@ -278,12 +287,10 @@
 - (void)setFont:(UIFont*)font {
     if (![_font isEqual:font]) {
         _font = font;
-        _originalFontSize = [_font pointSize];
-
         if (_adjustFontSize) {
-            [self adjustFontSizeToFit];
+            [self _adjustFontSizeToFit];
         } else {
-            [self adjustTextLayerSize];
+            [self _adjustTextLayerSize];
         }
     }
 }
@@ -309,9 +316,9 @@
         _text.attach([newStr copy]);
 
         if (_adjustFontSize) {
-            [self adjustFontSizeToFit];
+            [self _adjustFontSizeToFit];
         } else {
-            [self adjustTextLayerSize];
+            [self _adjustTextLayerSize];
         }
 
         self.accessibilityValue = newStr;
@@ -362,7 +369,7 @@
 - (void)setTextAlignment:(UITextAlignment)alignment {
     if (alignment != _alignment) {
         _alignment = alignment;
-        [self adjustTextLayerSize];
+        [self _adjustTextLayerSize];
     }
 }
 
@@ -383,7 +390,7 @@
     if (![_textColor isEqual:color]) {
         [[_textColor retain] autorelease];
         _textColor = color;
-        [self adjustTextLayerSize];
+        [self _adjustTextLayerSize];
     }
 }
 
@@ -481,7 +488,7 @@
 */
 - (void)setLineBreakMode:(UILineBreakMode)mode {
     _lineBreakMode = mode;
-    [self adjustTextLayerSize];
+    [self _adjustTextLayerSize];
 }
 
 /**
@@ -501,7 +508,7 @@
     } else {
         [self setBackgroundColor:_savedBackgroundColor];
     }
-    [self adjustTextLayerSize];
+    [self _adjustTextLayerSize];
 }
 
 /**
@@ -539,7 +546,7 @@
     _adjustFontSize = adjusts;
 
     if (_adjustFontSize) {
-        [self adjustFontSizeToFit];
+        [self _adjustFontSizeToFit];
     }
 }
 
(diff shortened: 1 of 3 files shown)


Frameworks/UIKit/UITableViewCell.mm
```diff
@@ -1191,7 +1191,10 @@ - (void)layoutSubviews {
     if (accessoryView) {
         CGRect accessoryRect = { 0 };
 
-        if ([accessoryView isKindOfClass:[UITextField class]]) {
+        // TODO: on reference platform, when UILabel is the accessory view, it uses its own intrinContentsize (similar to UITextField)
```
jaredhms: >// TODO:  [](start = 8, length = 9)

File a bug or is this tracked by a larger 'Finish UITableView' work item?

jaredhms: Seems like fixing bugs here would be preferable at this point.

---
In reply to: [97385306](https://github.com/Microsoft/WinObjC/pull/1768#discussion_r97385306) [](ancestors = 97385306)

Frameworks/UIKit/UILabel.mm
```diff
@@ -44,61 +44,70 @@ @implementation UILabel {
...
 
-    _font = [_font fontWithSize:curFontSize];
+    _font = [_font fontWithSize:maxFitFontSize];
 
-    [self adjustTextLayerSize];
+    [self _adjustTextLayerSize];
 }
 
-- (void)adjustTextLayerSize {
+- (void)_adjustTextLayerSize {
```
jaredhms: >_adjustTextLayerSize [](start = 8, length = 20)

nit: This name is misleading.   How about _applyXamlProperties or _updateXamlElement?

jaredhms: if you do rename it, please update #1497 which tracks getting rid of this method altogether

---
In reply to: [97386064](https://github.com/Microsoft/WinObjC/pull/1768#discussion_r97386064) [](ancestors = 97386064)

Frameworks/UIKit/UILabel.mm
```diff
@@ -44,61 +44,70 @@ @implementation UILabel {
...
+    }
 
-        //  A single line of text should be clipped
+    // Do binary search on maximum font size that fits the bound
+    do {
+        // Caculating rect size using font and line break settings
+        CGSize size = CGSizeZero;
         if (_numberOfLines == 1) {
-            size = [_text sizeWithFont:_font constrainedToSize:CGSizeMake(rect.size.width, 0.0f) lineBreakMode:UILineBreakModeClip];
+            size = [_text sizeWithFont:_font constrainedToSize:CGSizeMake(0.0f, 0.0f) lineBreakMode:UILineBreakModeClip];
```
oliversa-msft: >CGSizeMake(0.0f, 0.0f) [](start = 63, length = 22)

CGSizeZero

Frameworks/UIKit/UILabel.mm
```diff
@@ -44,61 +44,70 @@ @implementation UILabel {
...
         } else {
-            //  Multiple lines of text should be wrapped
             size = [_text sizeWithFont:_font constrainedToSize:CGSizeMake(rect.size.width, 0.0f) lineBreakMode:UILineBreakModeWordWrap];
         }
 
         if (size.width < rect.size.width && size.height <= rect.size.height) {
```
oliversa-msft: >< [](start = 23, length = 1)

Why is this not <=

Frameworks/UIKit/UILabel.mm
```diff
@@ -44,61 +44,70 @@ @implementation UILabel {
...
+            maxFitFontSize = curFontSize;
+            curFontSize *= 2.0f;
+        } else {
+            // Current font size does not fit
+            if (curFontSize - maxFitFontSize <= 1.0f) {
+                // Distance between current font size wwhich does nit fit
+                // and maximum font size which fits is less than 1.0, break out
+                break;
+            } else {
+                // Re-try with middle value between maxFitFontSize and current Font Size
```
oliversa-msft: >Re-try [](start = 19, length = 6)

nit: Retry

Frameworks/UIKit/UILabel.mm
```diff
@@ -563,34 +570,14 @@ - (CGSize)sizeThatFits:(CGSize)curSize {
     CGSize ret = { 0 };
 
     if (_text != nil) {
```
jaredhms: >_text != nil [](start = 8, length = 12)

nit: what about empty strings?  can we just return 0 for them too?

Frameworks/UIKit/UILabel.mm
```diff
@@ -563,34 +570,14 @@ - (CGSize)sizeThatFits:(CGSize)curSize {
     CGSize ret = { 0 };
```
oliversa-msft: >{ 0 } [](start = 17, length = 5)

CGSizeZero

Frameworks/UIKit/UILabel.mm
```diff
@@ -44,61 +44,70 @@ @implementation UILabel {
...
     int _numberOfLines;
     BOOL _isDisabled;
     BOOL _isHighlighted;
     UIBaselineAdjustment _baselineAdjustment;
     StrongId<WXCTextBlock> _textBlock;
 }
 
-- (void)adjustFontSizeToFit {
+- (void)_adjustFontSizeToFit {
     if (_numberOfLines == 0) {
```
jaredhms: Can you please add a comment here?

Frameworks/UIKit/UILabel.mm
```diff
@@ -44,61 +44,70 @@ @implementation UILabel {
...
     }
 
-    float curFontSize = _originalFontSize;
-    _font = [_font fontWithSize:curFontSize];
-
-    CGRect rect;
-    rect = [self bounds];
-    if (rect.size.width == 0.0f || rect.size.height == 0.0f) {
+    CGRect rect = [self bounds];
+    if (rect.size.width == 0.0f || rect.size.height == 0.0f || _text == nil) {
```
jaredhms: >_text == nil [](start = 63, length = 12)

nit: again; what about empty strings?  we don't want to return early for them?

Frameworks/UIKit/UITableViewCell.mm
oliversa-msft: Nit: typo - intrinsicContentSize

samples/XAMLCatalog/XAMLCatalog/UILabelViewController.m
```diff
@@ -46,10 +46,9 @@ - (UILabel*)_createUILabelWithColor:(UIColor*)color
 - (void)viewDidLoad {
     [super viewDidLoad];
 
-    // creating the text Fields
+    // creating lables
```
oliversa-msft: >lables [](start = 16, length = 6)

nit: labels

oliversa-msft:
:shipit:

Frameworks/UIKit/UILabel.mm
jaredhms: >UILineBreakModeClip [](start = 100, length = 19)

why are these different than our currently-set linebreak modes?

Frameworks/UIKit/UILabel.mm
```diff
@@ -44,61 +44,70 @@ @implementation UILabel {
...
-    float _originalFontSize;
     int _numberOfLines;
     BOOL _isDisabled;
     BOOL _isHighlighted;
     UIBaselineAdjustment _baselineAdjustment;
     StrongId<WXCTextBlock> _textBlock;
 }
 
-- (void)adjustFontSizeToFit {
+- (void)_adjustFontSizeToFit {
```
jaredhms: >_adjustFontSizeToFit [](start = 8, length = 20)

Can we add some tests for this?

jaredhms: different frames, different starting/min font sizes, etc.

---
In reply to: [97400039](https://github.com/Microsoft/WinObjC/pull/1768#discussion_r97400039) [](ancestors = 97400039)

jaredhms: Also, if you refactored this method a bit (to take starting/min font sizes and a frame as function arguments, you could return the final font to a calling function that applies it to _font and does the relayout.  At which point you could write a bunch of self-contained UTs against this method.

---
In reply to: [97400127](https://github.com/Microsoft/WinObjC/pull/1768#discussion_r97400127) [](ancestors = 97400127,97400039)

Frameworks/UIKit/UILabel.mm
```diff
@@ -44,61 +44,70 @@ @implementation UILabel {
...
-        CGSize size = CGSizeZero;
+    // MaxFitFontSize records currently found maximum font size that fits the bounds,
+    // It starts with _minimumFontSize because it is smallest font size allowed by the UILabel
+    float maxFitFontSize = _minimumFontSize;
+
+    float curFontSize = [_font pointSize];
+    // Start our trial with fonts bigger than _minimumFontSize
+    if (curFontSize < maxFitFontSize) {
+        curFontSize = 2.0 * maxFitFontSize;
+        _font = [_font fontWithSize:curFontSize];
```
jaredhms: >_font [](start = 8, length = 5)

I'd much prefer a local UIFont to use in the function that only gets set (or returned) at the end, and actually set _font in a single spot.

Frameworks/UIKit/UILabel.mm
```diff
@@ -44,61 +44,70 @@ @implementation UILabel {
...
-    while (curFontSize > _minimumFontSize && curFontSize > 0.0f) {
-        CGSize size = CGSizeZero;
+    // MaxFitFontSize records currently found maximum font size that fits the bounds,
+    // It starts with _minimumFontSize because it is smallest font size allowed by the UILabel
+    float maxFitFontSize = _minimumFontSize;
+
+    float curFontSize = [_font pointSize];
+    // Start our trial with fonts bigger than _minimumFontSize
+    if (curFontSize < maxFitFontSize) {
+        curFontSize = 2.0 * maxFitFontSize;
```
jaredhms: > curFontSize = 2.0 * maxFitFontSize; [](start = 7, length = 36)

Curious why not start at std::max<float>(curFontSize, maxFitFontSize), vs. starting with 2x maxFitFontSize?

Frameworks/UIKit/UILabel.mm
yiyang-msft: use  _updateXamlElement and updated #1497

---
In reply to: [97386744](https://github.com/Microsoft/WinObjC/pull/1768#discussion_r97386744) [](ancestors = 97386744,97386064)

Frameworks/UIKit/UILabel.mm
yiyang-msft: i was wondering so. I changed to <=, seems more logical. 

---
In reply to: [97387666](https://github.com/Microsoft/WinObjC/pull/1768#discussion_r97387666) [](ancestors = 97387666)

Frameworks/UIKit/UILabel.mm
yiyang-msft: the starting point can be randomly chosen. I don't have strong opinion where to start. I changed it to max.

---
In reply to: [97402423](https://github.com/Microsoft/WinObjC/pull/1768#discussion_r97402423) [](ancestors = 97402423)

Frameworks/UIKit/UILabel.mm
yiyang-msft: I got rid of it and adding additional test case for multi-line adjustWidthToFit. 

---
In reply to: [97388485](https://github.com/Microsoft/WinObjC/pull/1768#discussion_r97388485) [](ancestors = 97388485)

Frameworks/UIKit/UILabel.mm
yiyang-msft: because this function intend to fit the string to rect without truncation, even if you set linbreakMode as trailing or any other truncations, we still should not truncate the string with larger potentially font, but should try to fit the string with smaller font without truncation if possible.

---
In reply to: [97389805](https://github.com/Microsoft/WinObjC/pull/1768#discussion_r97389805) [](ancestors = 97389805)

Frameworks/UIKit/UILabel.mm
```diff
@@ -44,61 +44,106 @@ @implementation UILabel {
...
-
-    CGRect rect;
-    rect = [self bounds];
-    if (rect.size.width == 0.0f || rect.size.height == 0.0f) {
-        return;
-    }
+// Helper searches max font size for a string that can fit into given rect using given font/linkbreak config .
+// if no fit is found (e.g., to satisfy minimum font size requirement, nil is returned.
+// Otherwise, it returns the maximum font size for that can fit
+- (UIFont*)_findMaxFontSizeToFit:(CGRect)rect
```
jaredhms: >_findMaxFontSizeToFit [](start = 11, length = 21)

are we adding tests for this in a subsequent pass/set of changes?

yiyang-msft: I added three tests in XamlCatalog, which change the minimum font size to illustrate adjustFontSize behavior is different when we can not find the fit font. Search for. 
 // number of Lines = 2, unlimited, 3 and adjustFontSize YES, TailTruncation, change minimum size to different values

---
In reply to: [98052549](https://github.com/Microsoft/WinObjC/pull/1768#discussion_r98052549) [](ancestors = 98052549)

jaredhms:
:shipit:

Frameworks/UIKit/UILabel.mm
jaredhms: I meant automated tests.  this is a perfect candidate for some unit tests at this point.

---
In reply to: [98056283](https://github.com/Microsoft/WinObjC/pull/1768#discussion_r98056283) [](ancestors = 98056283,98052549)

yiyang-msft: i c. let me do that in UILabel work in this comming iteration. I am expecting a lot more tests will be written for UILabel. and this function should be move to somewhere that can be shared by multiple controls.

---
In reply to: [98065011](https://github.com/Microsoft/WinObjC/pull/1768#discussion_r98065011) [](ancestors = 98065011,98056283,98052549)

yiyang-msft:
pushed.

PR #1771, opened 2017-01-23
UX Functional test pattern - exercising sample app's UIViewController within a functional test

oliversa-msft:
CHANGES:
1. Created a new XAMLCatalog UIViewController with 2 controls (slider, UITextField) to manipulate a UIButton's label
2. This UIViewController's exposes a public UIButton property that can be manipulated from a functional test (UIButtonTests)
3. Created UX helper class, ViewControllerTestHelper, to display visuals using pushVC/dismissVC
4. Created UX helper class, XamlEventSubscription, to register/unregister XAML property change events
5. Updated FunctionalTest VCXPROJ to move external VCs into its own folder

TODO:
1. Migrate NSCondition to Win32 event mechanism

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1771)


tests/functionaltests/MainViewController.mm
```diff
@@ -27,9 +27,9 @@ @interface MainViewController ()
 @implementation MainViewController
 
 - (void)viewDidLoad {
-    self.view.backgroundColor = [UIColor lightGrayColor];
-
     [super viewDidLoad];
+
+    self.view.backgroundColor = [UIColor purpleColor];
```
oliversa-msft: >purpleColor [](start = 41, length = 11)

I could revert this but the upside is that it draws your attention when dismiss the UIViewController ;-)

samples/XAMLCatalog/XAMLCatalog.vsimporter/XAMLCatalog-WinStore10/Package.appxmanifest
```diff
@@ -7,15 +7,15 @@
...
     Publisher="CN=islandwood"
     Version="1.0.0.0" />
 
-  <mp:PhoneIdentity PhoneProductId="22d1821d-adee-4387-9878-038a210b92b0" PhonePublisherId="00000000-0000-0000-0000-000000000000"/>
+  <mp:PhoneIdentity PhoneProductId="610e2818-92fa-4f57-b41f-702cfd837520" PhonePublisherId="00000000-0000-0000-0000-000000000000"/>
 
   <Properties>
     <DisplayName>XAMLCatalog</DisplayName>
-    <PublisherDisplayName>jaredh</PublisherDisplayName>
+    <PublisherDisplayName>oliversa</PublisherDisplayName>
```
yiyang-msft: not sure worthwhile to check in these changes.

samples/XAMLCatalog/XAMLCatalog.vsimporter/XAMLCatalog-WinStore10/XAMLCatalog-Debug-xcvars.txt
```diff
@@ -85,15 +85,15 @@ PRODUCT_BUNDLE_IDENTIFIER = Microsoft.XAMLCatalog
...
+PROJECT_FILE_PATH = E:\GIT_REPOS\winobjc\samples\XAMLCatalog\XAMLCatalog.xcodeproj
 PROJECT_NAME = XAMLCatalog
 PROJECT_TEMP_DIR = sbuild/XAMLCatalog.sbuild
 PROJECT_TEMP_ROOT = sbuild
 PUBLIC_HEADERS_FOLDER_PATH = XAMLCatalog.app/Public
-SDKROOT = I:/enl/vso_1_WinObjC
-SOURCE_ROOT = I:/enl/vso_1_WinObjC/samples/XAMLCatalog
-SRCROOT = I:/enl/vso_1_WinObjC/samples/XAMLCatalog
+SDKROOT = E:/GIT_REPOS/winobjc
+SOURCE_ROOT = E:/GIT_REPOS/winobjc/samples/XAMLCatalog
```
yiyang-msft: >E:/GIT_REPOS/winobjc/ [](start = 14, length = 21)

also this file. seemed related to different user config. should we check this file in at all?

samples/XAMLCatalog/XAMLCatalog.vsimporter/XAMLCatalog-WinStore10/XAMLCatalog-Release-xcvars.txt
```diff
@@ -84,15 +84,15 @@ PRODUCT_BUNDLE_IDENTIFIER = Microsoft.XAMLCatalog
...
+PROJECT_DIR = E:/GIT_REPOS/winobjc/samples/XAMLCatalog
+PROJECT_FILE_PATH = E:\GIT_REPOS\winobjc\samples\XAMLCatalog\XAMLCatalog.xcodeproj
 PROJECT_NAME = XAMLCatalog
 PROJECT_TEMP_DIR = sbuild/XAMLCatalog.sbuild
 PROJECT_TEMP_ROOT = sbuild
 PUBLIC_HEADERS_FOLDER_PATH = XAMLCatalog.app/Public
-SDKROOT = I:/enl/vso_1_WinObjC
-SOURCE_ROOT = I:/enl/vso_1_WinObjC/samples/XAMLCatalog
-SRCROOT = I:/enl/vso_1_WinObjC/samples/XAMLCatalog
+SDKROOT = E:/GIT_REPOS/winobjc
```
yiyang-msft: >GIT_REPOS [](start = 13, length = 9)

same here.

tests/functionaltests/UXTestHelpers.h
```diff
@@ -0,0 +1,91 @@
...
+    ~ViewControllerHelper() {
+        // Uncomment this sleep - only used to visually confirm the view before it is dismissed
+        //[NSThread sleepForTimeInterval:3];
+
+        dispatch_sync(dispatch_get_main_queue(), ^{
+            [_externalViewController dismissViewControllerAnimated:NO completion:nil];
+        });
+
+        // Uncomment this sleep - only used to visually confirm that viewcontroller is dismissed
+        //[NSThread sleepForTimeInterval:3];
```
yiyang-msft: >//[NSThread sleepForTimeInterval:3]; [](start = 8, length = 36)

nit: can we do something like if (timeout != 0) and then [NSThread seelp]. and define timeout as 0 on top. so user does not have to uncomment multiple places if they want to turn on timeout

samples/XAMLCatalog/XAMLCatalog/UIKitControls/UIButtonWithControlsViewController.m
```diff
@@ -0,0 +1,87 @@
...
+//******************************************************************************
+
+#import "UIButtonWithControlsViewController.h"
+#import <UIKit/UIButton.h>
+
+@implementation UIButtonWithControlsViewController {
+    MenuTableViewController* _menuTVC;
+}
+
+- (void)viewDidLoad {
```
yiyang-msft: >viewDidLoad [](start = 8, length = 11)

nit; would be nice to have comments to describe the config of test view controller itself.

samples/XAMLCatalog/XAMLCatalog/ProgrammaticViewController.m
```diff
@@ -41,8 +42,11 @@ - (void)viewDidLoad {
     // UIButton
     [self addMenuItemViewController:[[UIButtonViewController alloc] init] andTitle:@"UIButton"];
 
+    // UIButton
+    [self addMenuItemViewController:[[UIButtonWithControlsViewController alloc] init] andTitle:@"UIButton - controls"];
```
jaredhms: > - controls [](start = 105, length = 11)

nit: The distinction between this and the above UIButton controller isn't clear.  Can we use a more descriptive name?

samples/XAMLCatalog/XAMLCatalog.vsimporter/XAMLCatalog-WinStore10/Package.appxmanifest
oliversa-msft: It was automatically populated with the re-import. I prefer to keep the autogen portion and not manually revert it.

---
In reply to: [97386296](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97386296) [](ancestors = 97386296)

samples/XAMLCatalog/XAMLCatalog/UIKitControls/UIButtonWithControlsViewController.m
oliversa-msft: No problem.

---
In reply to: [97388482](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97388482) [](ancestors = 97388482)

samples/XAMLCatalog/XAMLCatalog/UIKitControls/UIButtonWithControlsViewController.m
```diff
@@ -0,0 +1,87 @@
...
+    _textButtonLabel.delegate = self;
+    [_menuTVC addMenuItemView:_textButtonLabel andTitle:@"Text - normal state"];
+
+    _sliderTitleColorNormal = [[UISlider alloc] initWithFrame:CGRectMake(0.0f, 0.0f, 100.0f, 40.0f)];
+    _sliderTitleColorNormal.minimumValue = 0.0f;
+    _sliderTitleColorNormal.maximumValue = 360.0f;
+    _sliderTitleColorNormal.continuous = YES;
+    _sliderTitleColorNormal.value = 0.0f;
+    [_sliderTitleColorNormal addTarget:self action:@selector(titleColorNormalValueChanged) forControlEvents:UIControlEventValueChanged];
+    [_menuTVC addMenuItemView:_sliderTitleColorNormal andTitle:@"Bkgnd color - normal state"];
```
jaredhms: >Bkgnd [](start = 65, length = 5)

nit: Why abbreviate?

samples/XAMLCatalog/XAMLCatalog/UIKitControls/UIButtonWithControlsViewController.m
```diff
@@ -0,0 +1,87 @@
...
+    [_button setTitle:textField.text forState:UIControlStateNormal];
+
+    return YES;
+}
+
+- (BOOL)textFieldShouldReturn:(UITextField*)textField {
+    [_textButtonLabel resignFirstResponder];
+
+    return YES;
+}
```
jaredhms: this pattern isn't going to work very well with more than one text field - i think we should track individual text fields' text content.

samples/XAMLCatalog/XAMLCatalog/UIKitControls/UIButtonWithControlsViewController.m
```diff
@@ -0,0 +1,87 @@
...
+    _textButtonLabel = [[UITextField alloc] initWithFrame:CGRectMake(0.0f, 0.0f, 150.0f, 40.0f)];
+    _textButtonLabel.textColor = [UIColor blackColor];
+    _textButtonLabel.backgroundColor = [UIColor lightGrayColor];
+    _textButtonLabel.textAlignment = UITextAlignmentCenter;
+    _textButtonLabel.contentVerticalAlignment = UIControlContentVerticalAlignmentCenter;
+    _textButtonLabel.borderStyle = UITextBorderStyleLine;
+    _textButtonLabel.font = [UIFont systemFontOfSize:15.0f];
+    _textButtonLabel.adjustsFontSizeToFitWidth = YES;
+    _textButtonLabel.delegate = self;
+    [_menuTVC addMenuItemView:_textButtonLabel andTitle:@"Text - normal state"];
```
jaredhms: > - normal state [](start = 62, length = 15)

Are you planning to couple this textbox with another combo or set of UISegments for the various possible states?

oliversa-msft: Yes, I'll be adding more controls in future that manipulate the UIButton. But I'll hold that change for the next PR.

---
In reply to: [97408406](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97408406) [](ancestors = 97408406)

tests/functionaltests/MainViewController.mm
jaredhms: sounds fine, but curious about the reordering of calls?  that seems like unnecessary churn, or was it required for some reason?

---
In reply to: [97384555](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97384555) [](ancestors = 97384555)

samples/XAMLCatalog/XAMLCatalog/UIKitControls/UIButtonWithControlsViewController.m
oliversa-msft: I tested it using a small factor phone initially and I didn't like the truncation. I'll revert since it since it's no longer an issue on larger screens.

---
In reply to: [97407985](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97407985) [](ancestors = 97407985)

tests/functionaltests/UXTestHelpers.h
```diff
@@ -0,0 +1,91 @@
...
+                                                             }];
+    }
+
+    ~XamlEventSubscription() {
+        // Unregister the callback
+        [_xamlObject unregisterPropertyChangedCallback:_propertyToObserve token:_callbackToken];
+    }
+
+private:
+    int64_t _callbackToken;
```
jaredhms: why not capture a weak reference in the block?

---
In reply to: [97400288](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97400288) [](ancestors = 97400288)

tests/functionaltests/UXTestHelpers.h
```diff
@@ -0,0 +1,91 @@
...
+        // Register callback and wait for the property changed event to trigger
+        _callbackToken = [_xamlObject registerPropertyChangedCallback:_propertyToObserve
+                                                             callback:^(WXDependencyObject* sender, WXDependencyProperty* dp) {
+                                                                 callbackHandler(sender, dp);
+                                                             }];
+    }
+
+    ~XamlEventSubscription() {
+        // Unregister the callback
+        [_xamlObject unregisterPropertyChangedCallback:_propertyToObserve token:_callbackToken];
```
jaredhms: that assumes you only want it to fire one time.  I don't think that's the pattern we're going for.

---
In reply to: [97400178](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97400178) [](ancestors = 97400178)

samples/XAMLCatalog/XAMLCatalog/ProgrammaticViewController.m
oliversa-msft: Sure - ideally we'll revert update all existing VCs and keep the old naming convention (UISlider, UIButton, etc) but for now I can be more descriptive.

---
In reply to: [97406300](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97406300) [](ancestors = 97406300)

tests/functionaltests/UXTestHelpers.h
```diff
@@ -0,0 +1,91 @@
...
+
+        // Uncomment this sleep - only used to visually confirm that viewcontroller is dismissed
+        //[NSThread sleepForTimeInterval:3];
+    }
+
+private:
+    StrongId<UIViewController> _externalViewController;
+};
+
+typedef void (^XamlEventBlock)(WXDependencyObject*, WXDependencyProperty*);
```
jaredhms: >typedef void  [](start = 0, length = 13)

nit: our CC recommends that we "Prefer alias Declarations to typedefs", which would be something like: using XamlEventBlock = void (^)(WXDependencyObject*, WXDependencyProperty*);

tests/functionaltests/UXTestHelpers.h
jaredhms: just adding a second ctor that takes a timeout would be preferable.

---
In reply to: [97387022](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97387022) [](ancestors = 97387022)

tests/functionaltests/UXTestHelpers.h
```diff
@@ -0,0 +1,91 @@
...
+
+#import "ObjCXamlControls.h"
+#import "UWP/WindowsUIXamlControls.h"
+
+UIWindow* GetCurrentWindow();
+
+NSString* NSStringFromPropertyValue(RTObject* rtPropertyValue);
+NSString* NSStringFromPropertyValue(const Microsoft::WRL::ComPtr<IInspectable>& inspPropertyValue);
+
+class ViewControllerHelper {
```
jaredhms: >class ViewControllerHelper  [](start = 0, length = 27)

can you add comments above this?  what's it for?

jaredhms: also, it looks like a more suitable name may be AutoPresentViewController, or ViewControllerPresenter.

---
In reply to: [97411307](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97411307) [](ancestors = 97411307)

tests/functionaltests/UXTestHelpers.h
```diff
@@ -0,0 +1,91 @@
...
+public:
+    ViewControllerHelper(UIViewController* viewController) : _externalViewController(viewController) {
```
jaredhms: >ViewControllerHelper(UIViewController* viewController) [](start = 4, length = 54)

would it make more sense to take an optional second argument that is the VC on which to present?  The default value could be GetCurrentWindow().rootViewController.

tests/functionaltests/UXTestHelpers.h
```diff
@@ -0,0 +1,91 @@
...
+        dispatch_sync(dispatch_get_main_queue(), ^{
+            UIWindow* currentWindow = GetCurrentWindow();
+            [currentWindow.rootViewController presentViewController:_externalViewController animated:NO completion:nil];
+        });
+
+        // Uncomment this sleep - only used to visually confirm that the view appears on screen
+        //[NSThread sleepForTimeInterval:3];
```
jaredhms: We should move the implementations of these into the .mm file.

tests/functionaltests/UXTestHelpers.h
jaredhms: can we put all of these into a 'UXTestApi' (or similar) namespace for clarity?

tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
```diff
@@ -51,75 +36,98 @@
...
-                    [solidBrush.color b],
-                    [solidBrush.color a]);
+                // Extract the title color for the normal state
+                UIColor* titleColorNormal = [buttonVC.button titleColorForState:UIControlStateNormal];
 
                 CGFloat red, green, blue, alpha;
-                [cellButton.backgroundColor getRed:&red green:&green blue:&blue alpha:&alpha];
-                LOG_INFO("UIButton.backgroundColor (rgba): %.2f,%.2f,%.2f,%.2f", red, green, blue, alpha);
+                [titleColorNormal getRed:&red green:&green blue:&blue alpha:&alpha];
+                LOG_INFO("UIButton.titleColorForState:normal (rgba): %.2f,%.2f,%.2f,%.2f", red, green, blue, alpha);
```
jaredhms: We want EXPECT_EQ calls for these, too.

tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
```diff
@@ -51,75 +36,98 @@
...
 
                 // Validate that the change is reflected on the backing XAML control
                 EXPECT_EQ_MSG(solidBrush.color.r, (int)(red * 255), @"Failed to match red component");
                 EXPECT_EQ_MSG(solidBrush.color.g, (int)(green * 255), @"Failed to match green component");
                 EXPECT_EQ_MSG(solidBrush.color.b, (int)(blue * 255), @"Failed to match blue component");
                 EXPECT_EQ_MSG(solidBrush.color.a, (int)(alpha * 255), @"Failed to match alpha component");
```
jaredhms: nit: it seems like these should go up to where you grabbed solidBrush

jaredhms: ah - you're comparing the UIButton's to the backing Xaml element.  We also want to confirm that they're the expected values though, don't we?

---
In reply to: [97414376](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97414376) [](ancestors = 97414376)

samples/XAMLCatalog/XAMLCatalog/UIKitControls/UIButtonWithControlsViewController.m
```diff
@@ -0,0 +1,87 @@
...
+}
+
+- (BOOL)textFieldShouldReturn:(UITextField*)textField {
+    [_textButtonLabel resignFirstResponder];
+
+    return YES;
+}
+- (void)titleColorNormalValueChanged {
+    CGFloat hue = _sliderTitleColorNormal.value / 359.0f;
+    UIColor* color = [UIColor colorWithHue:hue saturation:hue brightness:0.4f alpha:1.0f];
```
jaredhms: >UIColor* color =  [](start = 4, length = 17)

if you stored this on the VC, you could grab it from the test and validate that the colors set on the button/xaml object were correct.

tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
```diff
@@ -51,75 +36,98 @@
...
+                                                 // Validate the state of the XAML object
+                                                 NSString* text = NSStringFromPropertyValue([sender getValue:dp]);
+                                                 LOG_INFO("TextBlock text: %@", text);
+
+                                                 // Extract the text for the normal state
+                                                 NSString* textNormal = [buttonVC.button titleForState:UIControlStateNormal];
+                                                 LOG_INFO("UIButton title - normal: %@", textNormal);
+
+                                                 // Validate that the change is reflected on the backing XAML control
+                                                 EXPECT_OBJCEQ_MSG(text, textNormal, @"Failed to match strings in XAML and UIButton");
```
jaredhms: same here; we should also make sure the text == @"Functional testing"

tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
```diff
@@ -51,75 +36,98 @@
     ASSERT_TRUE([backingElement isKindOfClass:[WXFrameworkElement class]]);
 }
 
-TEST(UIButton, BackgroundColorChanged) {
-    __block UIButtonViewController* buttonVC;
+TEST(UIButton, TitleColorChanged) {
+    __block BOOL signaled = NO;
     __block NSCondition* condition = [[[NSCondition alloc] init] autorelease];
```
jaredhms: >autorelease [](start = 65, length = 11)

nit: I think it would be nicer to avoid the autorelease pool for all of these, so we have more deterministic cleanup.

tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
oliversa-msft: Yes, that is correct. I took this shortcut because I didn't implement the discrete combobox yet. I could move to a UITextfield and pass through blueColor, redColor etc but I also wanted to leverage another common control such as UISlider for our pattern.

---
In reply to: [97414580](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97414580) [](ancestors = 97414580,97414376)

tests/functionaltests/UXTestHelpers.h
oliversa-msft: Done and done, I like the cleaner overloaded constructor with it. I have noticed though that even with a zero value for the timeout, it keeps displaying the VC.view whereas with the comment, it didn't.

---
In reply to: [97411131](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97411131) [](ancestors = 97411131,97387022)

samples/XAMLCatalog/XAMLCatalog/UIKitControls/UIButtonWithControlsViewController.m
```diff
@@ -0,0 +1,87 @@
...
+    _sliderTitleColorNormal = [[UISlider alloc] initWithFrame:CGRectMake(0.0f, 0.0f, 100.0f, 40.0f)];
+    _sliderTitleColorNormal.minimumValue = 0.0f;
+    _sliderTitleColorNormal.maximumValue = 360.0f;
+    _sliderTitleColorNormal.continuous = YES;
+    _sliderTitleColorNormal.value = 0.0f;
+    [_sliderTitleColorNormal addTarget:self action:@selector(titleColorNormalValueChanged) forControlEvents:UIControlEventValueChanged];
+    [_menuTVC addMenuItemView:_sliderTitleColorNormal andTitle:@"Bkgnd color - normal state"];
+}
+
+- (void)setupButton {
```
tadam-msft: >setupButton [](start = 8, length = 11)

As we move along, it would be great to have the button truly be a default button without any properties set directly on it, including the xxxforState properties.  We could default them in the text boxes that we are using to control the button, for instance _textButtonLabel would be set to Default Button--although we'd need to decide whether we are testing the button's default label or its normal state title.

tests/functionaltests/UXTestHelpers.h
oliversa-msft: Jared and I discussed this offline - my pattern assumed a one-shot event but it's useful to declare the subscriber once and trigger it multiple times with different actions. We still find value in the setter/destructor pattern for this reason.

---
In reply to: [97410140](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97410140) [](ancestors = 97410140,97400178)

samples/XAMLCatalog/XAMLCatalog/UIKitControls/UIButtonWithControlsViewController.m
```diff
@@ -0,0 +1,87 @@
...
+    _button = [[UIButton alloc] initWithFrame:self.view.bounds];
+    [_button setTitle:@"Button Normal" forState:UIControlStateNormal];
+    [_button setTitle:@"Pressed" forState:UIControlStateHighlighted];
+    [_button setTitleColor:[UIColor blackColor] forState:UIControlStateNormal];
+    [_button setTitleColor:[UIColor blueColor] forState:UIControlStateHighlighted];
+
+    _button.backgroundColor = [UIColor lightGrayColor];
+}
+
+- (BOOL)textFieldShouldEndEditing:(UITextField*)textField {
```
tadam-msft: >textFieldShouldEndEditing [](start = 8, length = 25)

As we talked about today, you may want UITextFieldTextDidChangeNotification

samples/XAMLCatalog/XAMLCatalog/UIKitControls/UIButtonWithControlsViewController.m
oliversa-msft: Good point - right now, my pattern was mostly focused on the pipeline from VC to functional test. The only caveat to your suggestion is that sometimes the default state does not render anything visually (white on white, no text, etc.)

---
In reply to: [97436503](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97436503) [](ancestors = 97436503)

tests/functionaltests/UXTestHelpers.h
oliversa-msft: Maybe in the next PR. 

---
In reply to: [97411758](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97411758) [](ancestors = 97411758)

samples/XAMLCatalog/XAMLCatalog/UIKitControls/UIButtonWithControlsViewController.m
oliversa-msft: Jared and I discovered that iOS doesn't actually trigger this callback when the text is set programmatically. We find that strange and we're still investigating why or looking for documentation to explain it.

---
In reply to: [97437560](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97437560) [](ancestors = 97437560)

tests/functionaltests/UXTestHelpers.h
DHowett-MSFT: A weak reference to what exactly?

tests/functionaltests/UXTestHelpers.h
DHowett-MSFT: If you find value in the destructor having hold of the token and unregistering it when it triggers, you'll have to make sure the lifetime of this object extends beyond the lifetime of the possibly async callback.

samples/XAMLCatalog/XAMLCatalog/UIKitControls/UIButtonWithControlsViewController.m
jaredhms: why won't fix this?

---
In reply to: [97408192](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97408192) [](ancestors = 97408192)

samples/XAMLCatalog/XAMLCatalog/UIKitControls/UIButtonWithControlsViewController.m
jaredhms: We found that Xaml doesn't either - at least it wasn't getting fired in our code.  That being said, we're not prevented from following this pattern (by observing changes to the label's text property).

---
In reply to: [97438933](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97438933) [](ancestors = 97438933,97437560)

tests/functionaltests/UXTestHelpers.h
jaredhms: I misunderstood your comment; yes, of course we need to hold it outside the UI thread so it continues to remain registered for the duration of the test.

tests/functionaltests/UXTestHelpers.h
jaredhms: Again, misunderstood your comment.

samples/XAMLCatalog/XAMLCatalog/UIKitControls/UIButtonWithControlsViewController.m
oliversa-msft: As we discussed, we're going to have multiple delegates for different text fields. We only 1 textfield right now so we only need 1 delegate. We'll add more as we add more controls.

---
In reply to: [97445443](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97445443) [](ancestors = 97445443,97408192)

samples/XAMLCatalog/XAMLCatalog.vsimporter/XAMLCatalog-WinStore10/XAMLCatalog-Debug-xcvars.txt
```diff
@@ -85,15 +85,15 @@ PRODUCT_BUNDLE_IDENTIFIER = Microsoft.XAMLCatalog
 PRODUCT_NAME = XAMLCatalog
 PRODUCT_TYPE = com.apple.product-type.application
 PROJECT = XAMLCatalog
-PROJECT_DIR = I:/enl/vso_1_WinObjC/samples/XAMLCatalog
-PROJECT_FILE_PATH = I:\enl\vso_1_WinObjC\samples\XAMLCatalog\XAMLCatalog.xcodeproj
+PROJECT_DIR = E:/GIT_REPOS/winobjc/samples/XAMLCatalog
+PROJECT_FILE_PATH = E:\GIT_REPOS\winobjc\samples\XAMLCatalog\XAMLCatalog.xcodeproj
```
tadam-msft: Do these machine specific changes get checked in?

oliversa-msft: Yes, every time you use VSIMPORTER, these changes are automatically made on your behalf. We could revert them, as Yi suggested, but it's going to resurface each time someone else makes a change.

---
In reply to: [97601767](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97601767) [](ancestors = 97601767)

samples/XAMLCatalog/XAMLCatalog/UIKitControls/UIButtonWithControlsViewController.m
oliversa-msft: A fine suggestion - I'll add it in my in my next commit since it should be quick and easy.

---
In reply to: [97414846](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97414846) [](ancestors = 97414846)

samples/XAMLCatalog/XAMLCatalog/UIKitControls/UIButtonWithControlsViewController.m
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// Delegate to set the UIButton.titleLabel text
+- (void)onTextChanged:(UITextField*)textField {
+    [_button setTitle:textField.text forState:UIControlStateNormal];
+}
+
+// Delegate to dismiss the keyboard
+- (BOOL)textFieldShouldReturn:(UITextField*)textField {
+    [textField resignFirstResponder];
```
yiyang-msft: >[textField resignFirstResponder]; [](start = 4, length = 33)

no need for this.

samples/XAMLCatalog/XAMLCatalog/UIKitControls/UIButtonWithControlsViewController.h
```diff
@@ -0,0 +1,27 @@
...
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#import "MenuTableViewController.h"
+
+@interface UIButtonWithControlsViewController : UIViewController <UITextFieldDelegate>
+
+@property (nonatomic, readonly) UIButton* button;
+@property (nonatomic, readonly) UITextField* textButtonNormal;
```
yiyang-msft: >textButtonNormal [](start = 45, length = 16)

textButton or textField?

samples/XAMLCatalog/XAMLCatalog/UIKitControls/UIButtonWithControlsViewController.h
```diff
@@ -0,0 +1,25 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#import "MenuTableViewController.h"
+
+@interface UIButtonWithControlsViewController : UIViewController <UITextFieldDelegate>
```
yiyang-msft: >UITextFieldDelegate [](start = 66, length = 19)

i don't see you implement any UITextField delegate method, instead, you registered to listen to controChangedlEvent. Are you sure that is what you want?

tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
```diff
@@ -51,75 +36,108 @@
...
+            [titleColorNormal getRed:&red green:&green blue:&blue alpha:&alpha];
+            LOG_INFO("UIButton.titleColorForState:normal (rgba): %.2f,%.2f,%.2f,%.2f", red, green, blue, alpha);
+
+            // Validate that the change is reflected on the backing XAML control
+            EXPECT_EQ_MSG(solidBrush.color.r, (int)(red * 255), @"Failed to match red component");
+            EXPECT_EQ_MSG(solidBrush.color.g, (int)(green * 255), @"Failed to match green component");
+            EXPECT_EQ_MSG(solidBrush.color.b, (int)(blue * 255), @"Failed to match blue component");
+            EXPECT_EQ_MSG(solidBrush.color.a, (int)(alpha * 255), @"Failed to match alpha component");
+
+            // Ensure that the XAML color values match the viewcontroller property
```
yiyang-msft: > // Ensure that the XAML color values match the viewcontroller property [](start = 11, length = 71)

why we care given above verification confirmed backingxaml color is the same as its outter UIButton?

samples/XAMLCatalog/XAMLCatalog/UIKitControls/UIButtonWithControlsViewController.h
oliversa-msft: In view of the other member names,, it would probably be textTitleNormal (similar to sliderTitleColorNormal).

---
In reply to: [97700191](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97700191) [](ancestors = 97700191)

samples/XAMLCatalog/XAMLCatalog/UIKitControls/UIButtonWithControlsViewController.m
oliversa-msft: It still helps when dismissing the keyboard on iOS.

---
In reply to: [97699783](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97699783) [](ancestors = 97699783)

tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
oliversa-msft: We care about the actual color - the pattern in the future might be that we set this property explicitly (e.g. blueColor) Our verification should include that (XAML == UIButton.titleColor) and that both == blueColor.

---
In reply to: [97702035](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97702035) [](ancestors = 97702035)

tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
```diff
@@ -51,75 +36,98 @@
...
-        [backingElement retain];
-
-        // Register callback and wait for the property changed event to trigger
-        int64_t callbackToken = [backingElement
-            registerPropertyChangedCallback:[WXCControl backgroundProperty]
-                            callback:^(WXDependencyObject* sender, WXDependencyProperty* dp) {
+    dispatch_sync(dispatch_get_main_queue(), ^{
+        // Extract UIButton.titleLabel control to verify its visual state
+        WXFrameworkElement* titleElement = [buttonVC.button.titleLabel xamlElement];
+        ASSERT_TRUE(titleElement);
```
oliversa-msft: I had to retract this change - we have to investigate if the _MSG version is or _OBJNE is compatible. I'll file a bug to track. 

---
In reply to: [97399064](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97399064) [](ancestors = 97399064)

samples/XAMLCatalog/XAMLCatalog/UIKitControls/UIButtonWithControlsViewController.m
yiyang-msft: if the delegate return TRUE, it should dismiss keyboard.  Is it not the case?

---
In reply to: [97799074](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97799074) [](ancestors = 97799074,97699783)

yiyang-msft: apparently you have to do it youself. nvm

---
In reply to: [97840037](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97840037) [](ancestors = 97840037,97799074,97699783)

yiyang-msft:
:shipit:

samples/XAMLCatalog/XAMLCatalog/UIKitControls/UIButtonWithControlsViewController.m
oliversa-msft: Correct, without the resign - it doesn't dismiss the keyboard on iOS.

---
In reply to: [97840037](https://github.com/Microsoft/WinObjC/pull/1771#discussion_r97840037) [](ancestors = 97840037,97799074,97699783)

samples/XAMLCatalog/XAMLCatalog/UIKitControls/UIButtonWithControlsViewController.h
```diff
@@ -0,0 +1,31 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#import "MenuTableViewController.h"
+
+// Subclassed UITextField to support functional testing
+@interface TestEnabledUITextField : UITextField
+@end
```
jaredhms: this does seem like something we'll want to use across all of our tests

jaredhms:
:shipit:

PR #1773, opened 2017-01-23
Fixing WocCatalog test app layout issue in Controls view controller on phone and crash on iOS in TextDislayView Controller.

yiyang-msft:
The two button lay out on phone is overlapped because the fixed margin that we set is too big on the phone which pushes two buttons toward the center and then they overlapped.

The crash only happen on iOS. It is because we didn't check the result (those two tests applies only on windows platform given the native font path used), which we used to do but got accidentally removed by earlier commit.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1773)


jaredhms:
:shipit:

Issue #1751, opened 2017-01-20
[IslandwoodNews] Unable to open any articles 


msft-Jeyaram:
Steps:
1. launch IslandwoodNews
2. open an article

Results:
Unable to open the article and application crashes.

This is failing on develop [and has propagated to CGD2D during the merge from Develop->CGD2D]

stack trace:
```
 	LIBOBJC2.DLL!objc_exception_throw(void * object) Line 134	C++
 	FOUNDATION.DLL!-[NSException raise](NSException * self, objc_selector * _cmd) Line 131	Objective-C++
>	FOUNDATION.DLL!+[NSException raise:format:](id * self, objc_selector * _cmd, NSString * name, NSString * format) Line 96	Objective-C++
 	FOUNDATION.DLL!-[NSCFArray objectAtIndex:](NSCFArray * self, objc_selector * _cmd, unsigned int index) Line 86	Objective-C++
 	UIKIT.DLL!-[_UIPageViewScrollView _populatePagesWithOffset:interactive:](_UIPageViewScrollView * self, objc_selector * _cmd, CGPoint * targetOffset, int interactive) Line 491	Objective-C++
 	UIKIT.DLL!-[_UIPageViewScrollView touchesBegan:withEvent:](_UIPageViewScrollView * self, objc_selector * _cmd, NSSet * touches, UIEvent * event) Line 208	Objective-C++
 	FOUNDATION.DLL!-[NSObject performSelector:withObject:withObject:](NSObject * self, objc_selector * _cmd, objc_selector * selector, id * obj1, id * obj2) Line 275	Objective-C++
 	UIKIT.DLL!-[UIView _processPointerEvent:forTouchPhase:](UIView * self, objc_selector * _cmd, WUXIPointerRoutedEventArgs * pointerEventArgs, UITouchPhase touchPhase) Line 609	Objective-C++
 	UIKIT.DLL!__41-[UIView _initPrivWithFrame:xamlElement:]_block_invoke(void * .block_descriptor, RTObject * sender, WUXIPointerRoutedEventArgs * e) Line 767	Objective-C++

```

msft-Jeyaram:
Also the performance seems bad now.
Navigating to the next tab and displaying the listview content is noticeably pretty slow.

aballway:
Can you add a stack trace or more context?

msft-Jeyaram:
```
 	LIBOBJC2.DLL!objc_exception_throw(void * object) Line 134	C++
[...]

PR #1778, opened 2017-01-24
Allow return of non-GDI-compatible family/subfamily names from CTFont…

### LINKED ISSUE #1758: Islandwood News crash when trying to insert nil font into dictionary

ms-jihua:
…Copy(Family)Name et al

Fixes #1758

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1778)


msft-Jeyaram:
This was the root cause of #1751
Please close that with this.

ms-jihua:
@msft-Jeyaram I'm confused what issue #1751 is actually describing - if it's describing an issue in NSLayoutManager with:
CTLineRef fitLine = static_cast<CTLineRef>(CFArrayGetValueAtIndex(CTFrameGetLines(frame.get()), 0));
then that issue still repros, and this was not the root cause (I may have also just opened a dupe at #1779)

msft-Jeyaram:
@ms-jihua  I updated the bug #1751 .
This might not as well be related, but I have seen the crash upon attempting to obtain 
`static_cast(CFArrayGetValueAtIndex(CTFrameGetLines(frame.get()), 0));` and the frame is empty.

Are you able to repo as per #1751 steps?

ms-jihua:
@msft-Jeyaram I am not able to repro the #1751 stack. I get #1758 on develop and #1779 after this change.

PR #1780, opened 2017-01-24
Fix CT drawing tests on ARM

### LINKED ISSUE #1772: CoreGraphics Unit tests are failing on ARM.
The following CoreGraphics Unit tests are failing on ARM.
CTFrame.BasicUnicodeTest
CTFontManager.DrawWithCustomFont 
CTFont.DrawGlyphs
TestDrawingTextInFonts/Fonts.TestFonts/3

aballway:
Changes more tests to be more lenient, and changes the templates for the image comparison to work with the Gtest macros
Fixes #1772

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1780)

### DIFF
--- a/tests/unittests/CoreGraphics.drawing/CTDrawingTests.cpp
+++ b/tests/unittests/CoreGraphics.drawing/CTDrawingTests.cpp
@@ -26,7 +26,7 @@ static NSURL* __GetURLFromPathRelativeToModuleDirectory(NSString* relativePath) 
 }
 #endif // WINOBJC
 
-TEXT_DRAW_TEST_F(CTFrame, BasicDrawingTest, WhiteBackgroundTest<PixelByPixelImageComparator<ComparisonMode::Mask>>) {
+TEXT_DRAW_TEST_F(CTFrame, BasicDrawingTest, WhiteBackgroundTest<PixelByPixelImageComparator<PixelComparisonModeMask<>>>) {
     CGContextRef context = GetDrawingContext();
     CGRect bounds = GetDrawingBounds();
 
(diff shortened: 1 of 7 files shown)


tests/unittests/CoreGraphics.drawing/ImageComparison.h
```diff
@@ -21,7 +21,23 @@
 
 enum class ImageComparisonResult : unsigned int { Unknown = 0, Incomparable, Different, Same };
 
-enum struct ComparisonMode { Exact, Mask };
+struct rgbaPixel {
```
DHowett-MSFT: >rgbaPixel [](start = 7, length = 9)

is there really literally no way to not move this into the header :P

tests/unittests/CoreGraphics.drawing/CTDrawingTests.cpp
```diff
@@ -470,7 +470,7 @@ TEXT_DRAW_TEST_F(CTLine, BasicDrawingTest, WhiteBackgroundTest<PixelByPixelImage
     CTRunDraw(run, context, {});
 }
 
-class Fonts : public WhiteBackgroundTest<PixelByPixelImageComparator<ComparisonMode::Mask>>,
+class Fonts : public WhiteBackgroundTest<PixelByPixelImageComparator<PixelComparisonModeMask<2300>>>,
```
DHowett-MSFT: >2300 [](start = 93, length = 4)

document why 2300, that's _a heck of a lot_

DHowett-MSFT:
:shipit:

tests/unittests/CoreGraphics.drawing/CTDrawingTests.cpp
rajsesh-msft: Yes. Given our reference images are 512x256, this is ~1.5%.  The bigger concern is that we are not accounting for cases such as lines being completely at a different offset, etc.  We need a more "fuzzy" proximity comparison than a pixel count.  But that is an exercise for a different time.

---
In reply to: [97449301](https://github.com/Microsoft/WinObjC/pull/1780#discussion_r97449301) [](ancestors = 97449301)

PR #1793, opened 2017-01-24
Fix leak in NSRunLoop _statesForMode: helper

ehren:
Few notes:

1) I had some trouble adding a MakeStrongId similar to MakeAutoCF

```
template <typename T>
StrongId<T> MakeStrongId(T val) {
    StrongId<T> p;
    p.attach(val);
    return p;
}
```

but maybe something like that would be the better approach?

------

2) some of the existing code uses static_cast<id> to handle objective-c iteration on a StrongId but I just used `.get()`....

maybe I should change them all to `static_cast<NSArray>` ?

-----

3) _getModes could leak but it's dead code so just removed.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1793)

### DIFF
--- a/Frameworks/include/NSRunLoop+Internal.h
+++ b/Frameworks/include/NSRunLoop+Internal.h
@@ -31,7 +31,7 @@
 - (void)_removeInputSource:(NSInputSource*)source forMode:(NSString*)mode;
 - (void)_addObserver:(NSObject*)observer forMode:(NSString*)mode;
 - (void)_removeObserver:(NSObject*)observer forMode:(NSString*)mode;
-- (StrongId<NSArray*>)_statesForMode:(NSString*)mode;
+- (StrongId<NSArray>)_statesForMode:(NSString*)mode;
 - (void)_processMainRunLoop:(int)value;
 - (void)_shutdown;
 - (void)removeTimer:(NSTimer*)timer forMode:(NSString*)mode;
--- a/Frameworks/include/NSRunLoop+Internal.h
+++ b/Frameworks/include/NSRunLoop+Internal.h
@@ -31,7 +31,7 @@
 - (void)_removeInputSource:(NSInputSource*)source forMode:(NSString*)mode;
 - (void)_addObserver:(NSObject*)observer forMode:(NSString*)mode;
 - (void)_removeObserver:(NSObject*)observer forMode:(NSString*)mode;
-- (StrongId<NSArray>)_statesForMode:(NSString*)mode;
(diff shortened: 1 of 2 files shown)


Frameworks/include/NSRunLoop+Internal.h
```diff
@@ -31,7 +31,7 @@
 - (void)_removeInputSource:(NSInputSource*)source forMode:(NSString*)mode;
 - (void)_addObserver:(NSObject*)observer forMode:(NSString*)mode;
 - (void)_removeObserver:(NSObject*)observer forMode:(NSString*)mode;
-- (StrongId<NSArray*>)_statesForMode:(NSString*)mode;
+- (StrongId<NSArray>)_statesForMode:(NSString*)mode;
```
DHowett-MSFT: nit: disprefer returning these C++ things across API boundaries; use an autoreleased NSArray* instead :smile:

DHowett-MSFT: ah, I see. This wasn't initially your code.
I'll downgrade it to a non-showstopper, but it might make the memory management here easier to reckon about.

DHowett-MSFT: Actually, it could be a better solution to simply make that return an autoreleased thing, and drop the StrongId from each of its consumers. Feel free to go either way!

DHowett-MSFT:
As per your comments:
1. Ideally, we'd unify the paradigms behind AutoCF and AutoId -- AutoId could take an rvalue reference (`std::move`) and then `MakeAutoId`/`MakeStrongId` would be perfect.
2. Either way is fine; `.get()` is somewhat clearer. I wish we could have an implicit conversion operator for that expression!
3. Sweet.

Frameworks/include/NSRunLoop+Internal.h
ehren: I would kind of prefer to stick with returning a StrongId since it's a cool way to avoid the autorelease pool (though forgetting to assign the call to a StrongId is a big use-after-free footgun). 

Since it turns out _statesForMode is only used in NSRunLoop.mm I'm thinking the best option is to remove the method declaration from the header but keep it returning a StrongId.

ehren:
I'll change them all to .get()s in the next commit

Frameworks/include/NSRunLoop+Internal.h
ehren: removing the declaration of _statesForMode from the +Internal header would've required another declaration in a class extension in the .mm.... so just went with the autorelease after all.

PR #1804, opened 2017-01-25
[CGD2D] Render buffer-shared G8 images as A8.

### LINKED ISSUE #1747: Sample Application does not render "help" text
If you click on the 'help' button, you should see some 'helpful' text.
It is not present

DHowett-MSFT:
This is riddled with caveats, many of which are documented in the code.

The first commit in this patchset attempts to implement a format converter that would run per-primitive. It's expensive, but it does get the job done.

Unfortunately, asking WIC to convert a PRGBA buffer to a G8 buffer causes it to discard all alpha data -- even if premultiplied, the luminance of partially-transparent pixels is not appropriately reduced in the final G8 image. This impacts alpha-heavy uses such as grayscale font rendering.

The A8 implementation was chosen for maximal compatibiltiy with Cocos2D. Cocos2D renders white (!) text into a "grayscale" buffer that it then passes off to OpenGLES as a `GL_ALPHA8` buffer. Hah.

The A8 "hack" also represents what we currently have on the develop branch.

Fixes #1747.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1804)


Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -252,6 +252,8 @@ inline HRESULT GetTarget(ID2D1Image** pTarget) {
 
     CGAffineTransform textMatrix{ CGAffineTransformIdentity };
 
+    using __CGContextFlushHook = void(*)(CGContextRef);
```
DHowett-MSFT: dead; drop.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -2775,39 +2796,48 @@ CGContextRef CGBitmapContextCreate(void* data,
 
 /**
  @Status Caveat
- @Notes releaseCallback and releaseInfo is ignored.
- We only support formats that are 32 bits per pixel, colorspace and bitmapinfo that are ARGB.
+ @Notes If data is provided, it can only be in one of the few pixel formats Direct2D can render to in system memory: (P)RGBA, (P)BGRA, or Alpha8. If a buffer is
+        provided for a grayscale image, render operations will be carried out into an Alpha8 buffer instead.
```
rajsesh-msft: Grayscale + 8bpp.  Our support for grayscale colorspace for > 8bpp doesn't exist.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -2775,39 +2796,48 @@ CGContextRef CGBitmapContextCreate(void* data,
...
+    RETURN_NULL_IF(!colorSpace);
 
-    // bitsperpixel = ((bytesPerRow/width) * 8bits/byte)
     size_t bitsPerPixel = ((bytesPerRow / width) << 3);
     WICPixelFormatGUID outputPixelFormat;
-    RETURN_NULL_IF_FAILED(_CGImageGetWICPixelFormatFromImageProperties(bitsPerComponent, bitsPerPixel, space, bitmapInfo, &outputPixelFormat));
+    RETURN_NULL_IF_FAILED(_CGImageGetWICPixelFormatFromImageProperties(bitsPerComponent, bitsPerPixel, colorSpace, bitmapInfo, &outputPixelFormat));
     WICPixelFormatGUID pixelFormat = outputPixelFormat;
 
+    void* actualBackingBuffer = data;
```
rajsesh-msft: seems unnecessary.

DHowett-MSFT: There will be a future where we want to keep track of the user's input `data` and destroy it with their release callback, but we don't necessarily want the `CGIWICBitmap` to hold a reference to it.

I'd like to leave this in for that time, but maybe rename it to something more appropriate.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -2743,10 +2743,43 @@ bool CGContextIsPointInPath(CGContextRef context, bool eoFill, CGFloat x, CGFloa
...
+          _width(width),
+          _height(height),
+          _stride(stride),
+          _releaseDataCallback(releaseDataCallback),
+          _releaseInfo(releaseInfo) {
+    }
+
+    ~__CGBitmapContext() {
+        if (_data && _releaseDataCallback) {
+            _releaseDataCallback(_releaseInfo, _data);
```
aballway: will the reference platform call the release callback if data == nullptr?

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -2743,10 +2743,43 @@ bool CGContextIsPointInPath(CGContextRef context, bool eoFill, CGFloat x, CGFloa
 struct __CGBitmapContext : CoreFoundation::CppBase<__CGBitmapContext, __CGContext> {
     woc::unique_cf<CGImageRef> _image;
 
+    void* _data;
```
aballway: nit: I know this one has been around a bit longer and CG has been using it, but I like the m_* naming convention for C++ structs/classes since it makes it more obvious what's C++ and what's objc.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -2743,10 +2743,43 @@ bool CGContextIsPointInPath(CGContextRef context, bool eoFill, CGFloat x, CGFloa
...
+
+    size_t _width;
+    size_t _height;
+    size_t _stride;
+
+    CGBitmapContextReleaseDataCallback _releaseDataCallback;
```
aballway: nit: make these private? nobody should be allowed to muck around with 'em

Frameworks/CoreGraphics/CGContext.mm
DHowett-MSFT: @aballway what is the value in distinguishing Objective-C and C++ code? Members and instance variables are, for all intents and purposes, identical. A fully-qualified out-of-body member function refers to members just like an Objective-C method refers to ivars.

They have similar lifetime requirements, as well.

our CF stuff is closer to Objective-C due to the reference counting; it's only C++ as a measure of convenience.

Also, our coding convention states

> Use the "_" prefix on all data members of Objective C and C++ classes. This is consistent with Objective C instance variables that are synthesized automatically by the compiler.

Frameworks/CoreGraphics/CGContext.mm
DHowett-MSFT: No; there's no data to be released.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -2775,39 +2808,55 @@ CGContextRef CGBitmapContextCreate(void* data,
...
+    void* wicImageBackingBuffer = data;
+
     if (!_CGIsValidRenderTargetPixelFormat(pixelFormat)) {
+        pixelFormat = GUID_WICPixelFormat32bppPRGBA;
+
         if (data) {
-            UNIMPLEMENTED_WITH_MSG(
-                "CGBitmapContext does not currently support input conversion and can only render into 32bpp PRGBA buffers.");
-            return nullptr;
+            if (CGColorSpaceGetModel(colorSpace) == kCGColorSpaceModelMonochrome) {
```
DHowett-MSFT: It would be better to check `outputPixelFormat == GUID_WICPixelFormat8bppGray` for ultimate truthiness.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -2857,7 +2906,11 @@ CGContextRef CGBitmapContextCreateWithData(void* data,
 
     ComPtr<ID2D1RenderTarget> renderTarget;
     RETURN_NULL_IF_FAILED(factory->CreateWicBitmapRenderTarget(customBitmap.Get(), D2D1::RenderTargetProperties(), &renderTarget));
-    CGContextRef context = _CGBitmapContextCreateWithRenderTarget(renderTarget.Get(), image.get(), outputPixelFormat);
+
+    __CGBitmapContext* context = __CGBitmapContext::CreateInstance(
+        kCFAllocatorDefault, renderTarget.Get(), requestedPixelFormat, data, width, height, bytesPerRow, releaseCallback, releaseInfo);
+    __CGContextPrepareDefaults(context);
+    context->SetImage(image);
```
msft-Jeyaram: >image [](start = 22, length = 5)

image.get()?

msft-Jeyaram:
:shipit:

PR #1825, opened 2017-01-25
Initialize m_lastPoint; fixes intermittent test failure.

DHowett-MSFT:
Bah.


DHowett-MSFT:
On merge, the build status here will cascade to `20170125-merge-develop`.

PR #1832, opened 2017-01-26
Allow _UIPopupViewController to present itself in middleware scenarios.

### LINKED ISSUE #1831: _UIPopupViewController is no longer able to present in middleware scenarios
A recent fix/change to UIViewController (75effb97e) regressed one of our middleware scenarios.  UIViewController now makes sure that one is presenting over a currently-visible ViewController.  However, in middleware scenarios, that won't hold true, so we must loosen the check in such cases.

jaredhms:
A recent fix/change to UIViewController (75effb9) regressed one of our middleware scenarios. UIViewController now makes sure that one is presenting over a currently-visible ViewController. However, in middleware scenarios, that won't hold true, so we must loosen the check in such cases.

Fixes #1831.

### DIFF
--- a/Frameworks/UIKit/UIViewController.mm
+++ b/Frameworks/UIKit/UIViewController.mm
@@ -56,6 +56,8 @@
 #import <QuartzCore/CALayer.h>
 #import <QuartzCore/CoreAnimationFunctions.h>
 
+#import "_UIPopupViewController.h"
+#import "CACompositor.h"
 #import "UIApplicationInternal.h"
 #import "UIEmptyView.h"
 #import "UIViewInternal.h"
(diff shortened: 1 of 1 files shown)


ehren:
I'm a bit curious how that worked with the old presentViewController:

https://github.com/Microsoft/WinObjC/blob/fbd5601f539bfdf0e61e59d4155921eb85f88abd/Frameworks/UIKit/UIViewController.mm#L1146

```
- (void)presentViewController:(UIViewController*)controller animated:(BOOL)animated completion:(void (^)(void))completion {
    bool shouldShow = false;
    UIViewController* curController = self;
    while (curController != nil) {
        if (curController->priv->_visibility != controllerNotVisible) {
            shouldShow = true;
        }
        curController = [curController parentViewController];
    }
    if (!shouldShow) {
        TraceWarning(TAG, L"Controller is not visible!");
        return;
    }
```

jaredhms:
Hmm - good question.  I thought the check was new.  I could synch back to check, but I think the fix is correct.  Do you see any issues with it?  Hopefully we'll have a public-facing test for this in the ~near future.

---
In reply to: [275289527](https://github.com/Microsoft/WinObjC/pull/1832#issuecomment-275289527) [](ancestors = 275289527)

ehren:
priv->_visibility is a bit confusing... I think UINavigationController and UIView poke at it too. Perhaps you're presenting over a view controller with _visibility controllerNotVisible.

Don't see any problem with the change though. I know with UINavigationController that pushing onto a completely not visible view controller e.g. in viewDidLoad is fine (not actually sure about the correct logic for presentViewController though)

jaredhms:
We absolutely are - we're presenting over nothing.  It's a middleware scenario where we host UIKit buttons, etc. within a C# app.  For example; a handler for the UIButton needs to present a VC over the C# app's window, so we added _UIPopupViewController to do so (it 'presents' the content within a full-window Xaml Popup).

---
In reply to: [275290885](https://github.com/Microsoft/WinObjC/pull/1832#issuecomment-275290885) [](ancestors = 275290885)

ehren:
I'm wrong that UIView accesses priv->_visible. It's just UIViewController and UINavigationController. I don't have any insight why present with a parent hierarchy all controllerNotVisible worked before though.

pradipd:
:shipit:

pradipd:
Thanks for looking into this.

oliversa-msft:
:shipit:

PR #1833, opened 2017-01-26
ObjC2Winmd: Support for async delegates with multiple parameters

mukhole:
Moved IAsyncOperation to a template implementation.
Added support for multiple return types from async objc blocks.


pradipd:
:shipit:

Issue #1757, opened 2017-01-20
Functional Tests fail during test cleanup.


DHowett-MSFT:
Perhaps due to an over-release.

DHowett-MSFT:
This is due to an over-release in the CALayerAppearanceTests. Assigning to @tadam-msft for triage.

Proximal cause: Overboard use of `__block` causes the nested callback blocks to not retain any referenced objects; some objects are autoreleased and some are manually released before their time on earth is up.

oliversa-msft:
This should not block today's release. It's a rarely repro'ed issue in the test cleanup which does not fail the TAEF test. I'm changing the pattern to these tests in a PR shortly as well.

oliversa-msft:
I attempted to repro the issue after applying the new UX change (which also removes __block decoration) and I've been unable to reproduce the offending stack. 

Previously I was only able to see the leaked object once in about 50 attempts but it's potentially just harder to repro with my local setup. Worth keeping a eye out for any recurrence.

PR #1855, opened 2017-01-27
Created UX test helper class and simplified the tests using the new UX pattern

### LINKED ISSUE #1764: Add functional test helper methods.
Add functional test helper methods to validate properties after waiting for a property change event and also support time out period for waiting.

oliversa-msft:
CHANGES:
1. Created UXEvent class, wrapping NSCondition, and UIColor and WUXMSolidBrush comparison method
2. Added more controls to manipulate a UIButton via XAMLCatalog's viewcontroller
3. Applied new UX test pattern to CALayerAppearance tests

Fixes #1764, #1757

### DIFF
--- a/tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
+++ b/tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
@@ -21,6 +21,8 @@
 #import "UIKitControls/UIButtonViewController.h"
 #import "UIKitControls/UIButtonWithControlsViewController.h"
 
+using namespace UXTestAPI;
+
 TEST(UIButton, CreateXamlElement) {
     // TODO: Switch to UIKit.Xaml projections when they're available.
     Microsoft::WRL::ComPtr<IInspectable> xamlElement(XamlCreateButton());
@@ -37,8 +39,8 @@ TEST(UIButton, GetXamlElement) {
 }
 
 TEST(UIButton, TitleColorChanged) {
-    __block auto uxEvent = std::make_shared<UXTestAPI::UXEvent>();
-    __block auto xamlSubscriber = std::make_shared<UXTestAPI::XamlEventSubscription>();
+    __block auto uxEvent = UXEvent::CreateManual();
+    __block auto xamlSubscriber = std::make_shared<XamlEventSubscription>();
 
     UIButtonWithControlsViewController* buttonVC = [[UIButtonWithControlsViewController alloc] init];
     UXTestAPI::ViewControllerPresenter testHelper(buttonVC);
@@ -58,19 +60,19 @@ TEST(UIButton, TitleColorChanged) {
             EXPECT_TRUE_MSG(UXTestAPI::IsRGBAEqual(solidBrush, titleColorNormal), @"Failed to match XAML- and UIButton title color");
             EXPECT_OBJCEQ_MSG(titleColorNormal, [buttonVC titleColorNormal], @"Failed to match expected color");
 
-            uxEvent->Signal();
+            uxEvent->Set();
         });
 
         // Action - validate this action takes effect on the control
         [buttonVC sliderTitleColorNormal].value = 150.0f;
     });
 
-    ASSERT_TRUE_MSG(uxEvent->Wait(c_testTimeoutInSec), "FAILED: Waiting for property changed event timed out!");
+    ASSERT_FALSE_MSG(uxEvent->WaitFor(c_testTimeoutInSec), "FAILED: Waiting for property changed event timed out!");
 }
 
 TEST(UIButton, TextChanged) {
-    __block auto uxEvent = std::make_shared<UXTestAPI::UXEvent>();
-    __block auto xamlSubscriber = std::make_shared<UXTestAPI::XamlEventSubscription>();
+    __block auto uxEvent = UXEvent::CreateManual();
+    __block auto xamlSubscriber = std::make_shared<XamlEventSubscription>();
 
     NSString* expectedString = @"Functional testing";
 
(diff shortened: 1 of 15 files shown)


tests/functionaltests/UXTestHelpers.h
```diff
@@ -81,4 +84,22 @@ class XamlEventSubscription {
...
+class UXEvent {
+public:
+    UXEvent() : _signaled(false) {
+        _condition.attach([[NSCondition alloc] init]);
+    }
+
+    UXEvent(const UXEvent& other) = delete; // no copy
+    UXEvent& operator=(const UXEvent& other) = delete;
+
+    void Signal();
```
jaredhms: we'll probably want a Reset, too.

jaredhms: nit: I'd probably call this 'Set' to align with 'Reset'

---
In reply to: [98240403](https://github.com/Microsoft/WinObjC/pull/1855#discussion_r98240403) [](ancestors = 98240403)

tests/functionaltests/UXTestHelpers.h
```diff
@@ -81,4 +84,22 @@ class XamlEventSubscription {
     StrongId<WXDependencyProperty> _propertyToObserve;
 }; // class XamlEventSubscription
 
+// Class used to signal and wait during a test run
+class UXEvent {
```
jaredhms: >UXEvent [](start = 6, length = 7)

We'll want an autoreset mode too.  How about we make the ctor private and do:

public:

static std::shared_ptr<UXEvent> UXEvent::CreateManual(); {
    return std::make_shared<UXEvent>(false);
}

static  std::shared_ptr<UXEvent> UXEvent::CreateAuto() {
    return std::make_shared<UXEvent>(true);
}

..

private:

explicit UXEvent(bool autoReset);

oliversa-msft: I can add the constructor UXEvent(autoReset) but using the "factory-like" methods to create UXEvents runs into issues because I'm removed the CC/operator=. As we discussed, I'll keep the constructor public with the enum for Manual vs Auto.

---
In reply to: [98241423](https://github.com/Microsoft/WinObjC/pull/1855#discussion_r98241423) [](ancestors = 98241423)

tests/functionaltests/UXTestHelpers.mm
```diff
@@ -106,4 +114,27 @@
...
+void UXEvent::Signal() {
+    [_condition lock];
+
+    _signaled = true;
+    [_condition signal];
+
+    [_condition unlock];
+}
+
+bool UXEvent::Wait(int timeOutInSeconds) {
```
yiyang-msft: >Wait [](start = 14, length = 4)

nit: 

i feel we can change the name to something like TimeoutIn(seconds). returning true when timed out, false otherwise. so the test can be written as 

ASSERT_FALSE_MSG(uxEvent->TimeoutIn(3), "porperty change didn't happen within 3"

current way of naming wait is bit unclear what return value (true/false) stands for
ASSERT_TRUE_MSG(uxEvent->Wait(c_testTimeoutInSec), "FAILED: Waiting for property changed event timed out!");

tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
```diff
@@ -47,61 +46,33 @@
...
-                     [solidBrush.color r],
-                     [solidBrush.color g],
-                     [solidBrush.color b],
-                     [solidBrush.color a]);
 
-            // Extract the title color for the normal state
+            // Validation
             UIColor* titleColorNormal = [buttonVC.button titleColorForState:UIControlStateNormal];
+            EXPECT_TRUE_MSG(UXTestAPI::IsRGBAEqual(solidBrush, titleColorNormal), @"Failed to match XAML- and UIButton title color");
+            EXPECT_OBJCEQ_MSG(titleColorNormal, [buttonVC titleColorNormal], @"Failed to match expected color");
```
yiyang-msft: ncie: so much shorter!

tests/functionaltests/UXTestHelpers.mm
jaredhms: hmm - i think Wait's more clear and aligns with most other event types.  We could return an enum vs a bool though for clarity, but true for succeeded, false for timed out also seems pretty clear?

---
In reply to: [98267209](https://github.com/Microsoft/WinObjC/pull/1855#discussion_r98267209) [](ancestors = 98267209)

yiyang-msft: agree wait is more aligned, it is just return is unclear 

if read 

ASSERT_TRUE_MSG(uxEvent->Wait(c_testTimeoutInSec), "FAILED: Waiting for property changed event timed out!");

I feel uxEvent->Wait isn't clear enough for reader to know what the expectation is. should we expect true or false. 

---
In reply to: [98268635](https://github.com/Microsoft/WinObjC/pull/1855#discussion_r98268635) [](ancestors = 98268635,98267209)

oliversa-msft: How about WaitFor? WaitFor(3.0f /* timeoutInSeconds */)

---
In reply to: [98268635](https://github.com/Microsoft/WinObjC/pull/1855#discussion_r98268635) [](ancestors = 98268635,98267209)

oliversa-msft: I have no problem changing the API to be WaitFor(....) which returns true if we timed out or false if we were signaled. Incidentally, the iOS documentation for its [NSCondition waitUntilDate:...] is confusing. It returns YES if the condition was signaled. My instincts would expect [NSCondition waitUntilDate:..] should return YES if it waited the full period.

---
In reply to: [98269735](https://github.com/Microsoft/WinObjC/pull/1855#discussion_r98269735) [](ancestors = 98269735,98268635,98267209)

yiyang-msft:
:shipit:

tests/functionaltests/UXTestHelpers.mm
```diff
@@ -106,4 +114,37 @@
...
+    }
+
+    bool timedOut = !_signaled;
+    [_condition unlock];
+
+    if (_eventType == AutoReset) {
+        Reset();
+    }
+
+    return timedOut;
```
jaredhms: >timedOut [](start = 11, length = 8)

we should return true if the wait succeeded, not if it timed out.

yiyang-msft: @ , on retrospection,  return true if the wait succeeded, not if it timed out. is consistent across the board, even with iOS. so i am totally ok with that. 

---
In reply to: [98278040](https://github.com/Microsoft/WinObjC/pull/1855#discussion_r98278040) [](ancestors = 98278040)

jaredhms: yes, otherwise this is just plain confusing. thanks!

---
In reply to: [98278040](https://github.com/Microsoft/WinObjC/pull/1855#discussion_r98278040) [](ancestors = 98278040)

jaredhms:
:shipit:

PR #1857, opened 2017-01-27
Fix islandwood news crash in article view when panning

yiyang-msft:
the caller passing in empty controller array and our protection logic isn't correct which leads to array out of bound.


jaredhms:
:shipit:

oliversa-msft:
:shipit:

Frameworks/UIKit/UIPageViewController.mm
```diff
@@ -450,14 +450,16 @@ - (void)layoutSubviews {
...
                      targetOffset->y,
                      interactive,
-                     _keepAdjacent);
+                     _keepAdjacent,
+                     [_controllers count]);
     }
 
-    if ([_controllers count] == 0 && _skipUpdates) {
+    // if no controller to populate with or was told to skip, return early
+    if ([_controllers count] == 0 || _skipUpdates) {
```
rajsesh-msft: subtle.

PR #1870, opened 2017-01-28
Use _CGContextPushBegin/PopEndDraw for Drawing Tests

### LINKED ISSUE #1869: [CGD2D] CGContextClearRect crashes if used under CGContextPushBegin/PopEndDraw
This is because it's using the raw ->BeginDraw() and ->EndDraw().

I must've missed this during #1750.

ms-jihua:
- Since we generally expect real use-cases to use these functions for perf purposes,
   we should also validate that our drawing is correct when using them.
 - Fixed a stray raw ->BeginDraw() ->EndDraw() in CGContext.mm that was missed
   and could cause crashes if used under PushBegin/PopEndDraw

Fixes #1869

### DIFF
--- a/Frameworks/CoreGraphics/CGContext.mm
+++ b/Frameworks/CoreGraphics/CGContext.mm
@@ -2126,11 +2126,11 @@ void CGContextClearRect(CGContextRef context, CGRect rect) {
     NOISY_RETURN_IF_NULL(context);
     ComPtr<ID2D1DeviceContext> deviceContext = context->DeviceContext();
     if (!context->CurrentGState().clippingGeometry) {
-        deviceContext->BeginDraw();
+        _CGContextPushBeginDraw(context);
         deviceContext->PushAxisAlignedClip(__CGRectToD2D_F(rect), D2D1_ANTIALIAS_MODE_PER_PRIMITIVE);
         deviceContext->Clear(nullptr); // transparent black clear
         deviceContext->PopAxisAlignedClip();
-        deviceContext->EndDraw();
+        _CGContextPopEndDraw(context);
     }
 }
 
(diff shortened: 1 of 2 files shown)


tests/UnitTests/CoreGraphics.drawing/DrawingTest.cpp
```diff
@@ -34,10 +38,14 @@ void testing::DrawTest<TComparator>::SetUp() {
     CGSize size = CanvasSize();
 
     auto deviceColorSpace = woc::MakeStrongCF<CGColorSpaceRef>(CGColorSpaceCreateDeviceRGB());
-    _context.reset(CGBitmapContextCreate(
-        nullptr, size.width, size.height, 8, size.width * 4, deviceColorSpace, kCGImageAlphaPremultipliedFirst));
+    _context.reset(
+        CGBitmapContextCreate(nullptr, size.width, size.height, 8, size.width * 4, deviceColorSpace, kCGImageAlphaPremultipliedFirst));
     ASSERT_NE(nullptr, _context);
 
+#if WINOBJC // Validate that the results are correct even under batched drawing from _CGContextPushBegin/PopEndDraw
```
ms-jihua: I don't know why I didn't think of doing this earlier, sorry everyone!

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -2830,7 +2830,8 @@ CGContextRef CGBitmapContextCreate(void* data,
 
 /**
  @Status Caveat
- @Notes If data is provided, it can only be in one of the few pixel formats Direct2D can render to in system memory: (P)RGBA, (P)BGRA, or Alpha8. If a buffer is
```
DHowett-MSFT: saddest comment. can you reflow this so clang-format doesn't? :smile:

tests/UnitTests/CoreGraphics.drawing/DrawingTest.cpp
DHowett-MSFT: It might even be more correct to do this in CGBitmapContext.

Here's what we can potentially do:

If the user passes `!data`, we push a begin draw. When they call CreateImage/GetImage, we pop an endDraw if we know we are in control of the data.

That way, people using `data` get immediate mode rendering, and people not using data automatically get batched mode rendering.

This opens up some testing avenues that don't require internal implementation details (pass data, do not pass data, etc.).

I do, however, think this is the right thing to do here until we normalize on a strategy for bitmap contexts.

ms-jihua: Agreed, I'm not sure what you're envisioning for how we know if we control the data (will eventually call Create/GetImage). Leaving as-is for now.

---
In reply to: [98318678](https://github.com/Microsoft/WinObjC/pull/1870#discussion_r98318678) [](ancestors = 98318678)

PR #1871, opened 2017-01-28
Adding macros to enhance return and failure checks.

msft-Jeyaram:
This is the commit into develop.


msft-Jeyaram:
It will be one less merge and people (likes of brain) could use it if needed.

DHowett-MSFT:
But it will be one more merge conflict when we resolve CGD2D, which touches the same files.

msft-Jeyaram:
This was a direct cherry pick.

msft-Jeyaram:
Verbal sign off from @DHowett-MSFT

PR #1834, opened 2017-01-26
Fix various leaks in Foundation

ehren:

PR #1872, opened 2017-01-30
Fix leak calling [NSThread currentThread] when underlying thread created outside NSThread API

ehren:
This is an alternative for the naive leak fix in #1834 (which introduces a use after free).

I took a look at supporting the cdecl destructor argument of pthread_key_create but it looks like any fixes there would complicate the implementation of pthread_setspecific/pthread_getspecific. So this commit just calls the Fls functions directly from NSThread.

### DIFF
--- a/Frameworks/Foundation/NSThread.mm
+++ b/Frameworks/Foundation/NSThread.mm
@@ -349,10 +349,6 @@ static void* _threadBody(void* context) {
     return [[_name retain] autorelease];
 }
 
-static VOID WINAPI externalThreadCleanup(_In_ PVOID externalNSThread) {
-    [reinterpret_cast<NSThread*>(externalNSThread) release];
-}
-
 /**
 @Status Interoperable
 */
@@ -366,11 +362,9 @@ static VOID WINAPI externalThreadCleanup(_In_ PVOID externalNSThread) {
         currentThread = [[NSThread alloc] init];
         [currentThread _associateWithCurrentThread];
 
-        // Release it on thread exit
-        DWORD flsIndex = ::FlsAlloc(externalThreadCleanup);
-        if (flsIndex != FLS_OUT_OF_INDEXES) {
-            ::FlsSetValue(flsIndex, reinterpret_cast<PVOID>(currentThread));
-        }
+        // Ensure it's released on thread exit
+        thread_local StrongId<NSThread> current;
+        current.attach(currentThread);
     }
     return currentThread;
 }
--- a/Frameworks/Foundation/NSThread.mm
+++ b/Frameworks/Foundation/NSThread.mm
@@ -363,7 +363,7 @@ static void* _threadBody(void* context) {
 @Status Interoperable
 */
 - (NSString*)name {
-    return _name;
+    return [[_name retain] autorelease];
 }
 
 /**
--- a/Frameworks/Foundation/NSThread.mm
+++ b/Frameworks/Foundation/NSThread.mm
@@ -32,11 +32,35 @@ static const wchar_t* TAG = L"NSThread";
 static const NSUInteger kNSThreadDefaultStackSize = 1024 * 1024;
 static const double kNSThreadDefaultPriority = 0.5;
 
-pthread_key_t tlsNSThread, tlsNSThreadRunLoop;
-
 static std::mutex s_mainThreadMutex;
 static StrongId<NSThread> s_mainThread;
 static BOOL s_isMultiThreaded = NO;
+
+static NSThread* setOrGetCurrentThread(NSThread* thread) {
+    // Use a lazy initialized block scope thread_local to manage the liftime of currentThread.
+    // Note that non-trivial file scope thread_locals are apparently not supported in current clang version.
+    thread_local static StrongId<NSThread> tlsCurrentThread;
+
+    if (thread != nil) {
+        // We've been passed a thread. Designate as current thread and assume ownership.
+        tlsCurrentThread = thread;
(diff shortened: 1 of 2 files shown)


DHowett-MSFT:
Also in CI build :smile:

Frameworks/Foundation/NSThread.mm
```diff
@@ -356,8 +356,15 @@ + (NSThread*)currentThread {
     NSThread* currentThread = [[self class] _threadObjectFromCurrentThread];
 
     if (currentThread == nil) {
-        currentThread = [NSThread new];
+        // Underlying thread created outside the NSThread APIs
+
+        // So create a new instance
+        currentThread = [[NSThread alloc] init];
         [currentThread _associateWithCurrentThread];
```
DHowett-MSFT: Actually, it seems like this could be wrapped up into the current thread machinery. At file scope, a `thread_local StrongId<NSThread>`, which `_associateWith...` and `_threadObjectFrom...` could refer to. That way, its lifetime would be managed by the same thing that keeps track of its current-threadiness.

ehren: That makes sense. I'll include that + some cleanups in the next commit - I can see there's e.g. an unused pthread tls key in that file too.

tests/unittests/Foundation/NSThreadTests.m
```diff
@@ -77,6 +78,24 @@ - (void)countingloop {
...
+        // Grab a strong reference to NSThread.currentThread
+        // on a thread not created via the NSThread APIs
+        current = [[NSThread currentThread] retain];
+    });
+
+    t.join();
+
+    // Ensure the reference grabbed above is
+    // the only one remaining after thread exit
+    EXPECT_EQ(1, [current retainCount]);
```
rajsesh-msft: I am not sure I understand how this is correct.  alloc init should add a reference and currentThread retain should add one?

ehren: In the body of the lambda we're not creating the NSThread via alloc init. We just get a non-owning reference by the call to |currentThread|. The call to |currentThread| does happen to create an owning reference (because it's called from a thread not created via the NSThread API), but this is relinquished on thread exit via the thread_local StrongId destructor. 

I'll add some comments to the test case when I add in @DHowett-MSFT's suggested changes - likely sometime tonight. It might also be clearer to convert that test to ARC - but I had some difficulty attempting this initially.

ehren: @rajsesh-msft there's a few more asserts in the unit test which should improve things

Frameworks/Foundation/NSThread.mm
ehren: @DHowett-MSFT I don't think non-POD file scope thread_locals are supported by this version of clang (c2?). 

e.g. this is fine at file scope

```
struct Blah {
    Blah() {
        abort();
    }

   ~Blah() {
        abort();
   }
};
 
thread_local Blah blah;
```

But even if thread_local worked the lazy approach avoids the static initialization/destruction overhead for every thread created outside the NS* APIs.

I also haven't confirmed but I think this means there's a leak from thread_local use in NSProgress: https://github.com/Microsoft/WinObjC/blob/develop/Frameworks/Foundation/NSProgress.mm#L50

Frameworks/Foundation/NSThread.mm
```diff
@@ -32,12 +32,36 @@
 static const NSUInteger kNSThreadDefaultStackSize = 1024 * 1024;
 static const double kNSThreadDefaultPriority = 0.5;
 
-pthread_key_t tlsNSThread, tlsNSThreadRunLoop;
-
 static std::mutex s_mainThreadMutex;
 static StrongId<NSThread> s_mainThread;
 static BOOL s_isMultiThreaded = NO;
 
+static NSThread* setOrGetCurrentThread(NSThread* thread) {
```
rajsesh-msft: nit: the private functions should be _set* per coding conventions.

Frameworks/Foundation/NSThread.mm
rajsesh-msft: Good find and somewhat worrisome.  Yes, with NSProgress.mm, there is a leak.  s_currentProgressStack could become a thread local static inside _getProgressStackForCurrentThread much like you have done here.

Frameworks/Foundation/NSThread.mm
ehren: @rajsesh-msft I'll open a pull with the NSProgress change

DHowett-MSFT:
Submitted these for CI build again.

DHowett-MSFT:
It looks like these never completed build, but they did submit status updates... @pradipd do you know what happened?
I'll re-queue them in the mean time, but it was a week ago...

rajsesh-msft:
Very timely @DHowett-MSFT.  Take a look at #1937 that @pradipd just opened.

DHowett-MSFT:
_facepalm_
I saw @pradipd's pull request, but didn't put two and two together. Thank you!

ehren:
@pradipd thanks for the test fix!

I took the opportunity to push another change. No change in behavior - but this avoids creating a new thread-lifetime NSThread object unnecessarily (isMainThread is called frequently via UIView RunSynchronouslyOnMainThread)

rajsesh-msft:
Back in CI build.

DHowett-MSFT:
Back in CI build again. Sorry @ehren!

DHowett-MSFT:
Thanks @ehren!

PR #1873, opened 2017-01-30
Fix various leaks in Foundation v2

ehren:
These are the same changes as #1834 but without the NSThread change which is now handled in #1872


DHowett-MSFT:
In CI build! Thanks for the quick resubmit, @ehren!

jaredhms:
:shipit:

PR #1876, opened 2017-01-31
Add benchmarking test project and add text drawing benchmark tests

### LINKED ISSUE #1819: Need performance benchmark tests for CoreText
Would be useful to time these scenarios.  The goal is to have a simple commandline test similar to the unit tests, but rather than pass fail, these would run the perf scenarios and log the various metrics.  Here are some basic metrics we want to have:

1.	Drawing a frame with 100 sentences.
2.	Drawing 10000 lines.
3.	Drawing 100,000 runs.
4.	Drawing 10,000,000 glyphs.
5.	Drawing using NSString drawAtPoint and variations.
6.	Drawing using NSAttributedString drawAtPoint.

Once we have this, we could plug this into nightly build for build-over-build tracking of perf numbers to catch future regressions.

aballway:
Creates a new test project Benchmark in the Unit Test group, which leverages GTest to run automated benchmark tests.  Simple tests can be done with the TEST_BENCHMARK macro, or more complex ones can be used by creating classes to do setup and teardown so only specific parts of code are measured.

Fixes #1819


tests/UnitTests/Benchmark/BenchmarkTest.h
```diff
@@ -0,0 +1,153 @@
...
+template <typename T>
+static void __RunTest(T& test, std::vector<Seconds>& results) {
+    size_t runCount = test.GetRunCount();
+    for (size_t i = 0; i < runCount; ++i) {
+        test.PreRun();
+        auto start = std::chrono::high_resolution_clock::now();
+        test.Run();
+        auto end = std::chrono::high_resolution_clock::now();
+        test.PostRun();
+        std::chrono::duration<Seconds> duration = end - start;
```
msft-Jeyaram: curious why are we doing manual time measurements. 
gtest can do this for us.

tests/UnitTests/Benchmark/BenchmarkTest.h
```diff
@@ -0,0 +1,153 @@
...
+        results.emplace_back(duration.count());
+    }
+
+    Seconds netRuntime = std::accumulate(results.cbegin(), results.cend(), 0.0);
+    Seconds sampleMean = netRuntime / runCount;
+    Seconds sampleStandardDeviation = std::numeric_limits<Seconds>::infinity();
+    if (runCount > 1) {
```
msft-Jeyaram: have we looked into the built in support gtest provides us?

build/build.sln
```diff
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.25123.0
```
ms-jihua: seems dangerous, what was the reasoning behind this?

aballway: Unknown, didn't notice that changed when I added the project.  Good catch

tests/UnitTests/Benchmark/BenchmarkTest.h
rajsesh-msft: Should compute in micro seconds if not nano seconds.

tests/UnitTests/Benchmark/TextBenchmarkTests.mm
```diff
@@ -0,0 +1,399 @@
...
+    inline void Run() {
+        CTFrameDraw(m_frame, m_context);
+    }
+
+    size_t GetRunCount() const {
+        return 100;
+    }
+};
+
+class CTFrameDrawGroup : public CTFrameDrawBase {
```
rajsesh-msft: I'd like to also add tests that time the "create frame, line, etc" to this mix.

tests/UnitTests/Benchmark/TextBenchmarkTests.mm
```diff
@@ -0,0 +1,399 @@
...
+
+    void PostRun() {
+        _CGContextPopEndDraw(m_context);
+    }
+};
+
+TEST_BENCHMARK_F(CoreText, CTFrameDrawSingle);
+TEST_BENCHMARK_F(CoreText, CTFrameDrawGroup);
+
+class CTFrameDrawHuge : public TextBenchmarkBase {
```
rajsesh-msft: nit: CTFrameDrawYuge

aballway:
[Results from running on my machine](https://github.com/Microsoft/WinObjC/files/743396/output.txt)

tests/UnitTests/Benchmark/BenchmarkTest.h
ms-jihua: What was the verdict on this?

---
In reply to: [98574225](https://github.com/Microsoft/WinObjC/pull/1876#discussion_r98574225) [](ancestors = 98574225)

msft-Jeyaram: ++

tests/UnitTests/Benchmark/BenchmarkTest.h
msft-Jeyaram: gtest already provides utility to help you measure running times.
It's part of the UnitTest lib.

aballway: The only sort of timing utility I see in the gtest code is elapsed_time in TestResult.  IIRC that won't work once @bbowman's changes are merged in with #1849.  If that did work, we would still run into the problem of trying to only time part of the code run.  Right now this only measures the time it takes to run the code in the Run method, but from what I can see gtest measures the entire construction of the test as well so we wouldn't be able to ignore setup/teardown time.  Is there some other timing utility in gtest that I'm just not seeing/misunderstanding?

aballway:
Rebased onto develop as this new project depended on freetype which no longer exists 😸 Also ping!

tests/UnitTests/Benchmark/BenchmarkTest.h
```diff
@@ -0,0 +1,156 @@
...
+            Microseconds previousMean = mean;
+            netRuntime += results[i];
+            mean += (results[i] - mean) / (i + 1);
+            q += (results[i] - previousMean) * (results[i] - mean);
+        }
+
+        Microseconds sampleStandardDeviation = std::sqrt(q / (runCount - 1));
+
+        LOG_INFO(
+            "\nNet Runtime: %f microseconds\nSample Mean Runtime: %f microseconds\nSample Standard Deviation: %f microseconds\n%u test "
```
ms-jihua: nit: probably out of scope, but I wonder if we would benefit from having a option to output in a csv format... probably overengineering if it looks like we'll eventually change the framework though.

tests/UnitTests/Benchmark/TextBenchmarkTests.mm
```diff
@@ -0,0 +1,536 @@
...
+    }
+
+    size_t GetRunCount() const {
+        return 100;
+    }
+};
+
+TEST_BENCHMARK_F(CoreText, CTFrameDrawComplete);
+
+class CTFrameDrawYuge : public TextBenchmarkBase {
```
ms-jihua: oh god why did you take @rajsesh-msft seriously

tests/UnitTests/Benchmark/TextBenchmarkTests.mm
```diff
@@ -0,0 +1,536 @@
...
+    inline void Run() {
+        CTFrameDraw(m_frame, m_context);
+    }
+
+    size_t GetRunCount() const {
+        return 100;
+    }
+};
+
+class CTFrameDrawGroup : public CTFrameDrawBase {
```
ms-jihua: DrawGroup tests:
 - should probably abstract out the PreRun PushBeginDraw/PostRun PopEndDraw to something common
 - I'm kind of uncomfortable with the n=1 runs for the group tests. I realize it's 100 frames, and that these tests take a while, but EndDraw only happens once, meaning that some small temporary hiccup on the machine we run this on could lead to us overinterpreting noise. n=5 maybe?

build/Tests/UnitTests/Benchmark/Benchmark.UnitTests.vcxproj
```diff
@@ -0,0 +1,248 @@
...
+      <IncludePaths>$(StarboardBasePath)\Frameworks\include;$(StarboardBasePath)\include\xplat;$(StarboardBasePath)\tests\frameworks\include;$(StarboardBasePath)\tests\frameworks\gtest;$(StarboardBasePath)\tests\frameworks\gtest\include;$(StarboardBasePath)\;%(AdditionalIncludeDirectories)</IncludePaths>
+      <CompileAs>CompileAsObjCpp</CompileAs>
+      <OtherCPlusPlusFlags>-fmsvc-real-char</OtherCPlusPlusFlags>
+      <PreprocessorDefinitions>NO_STUBS;_CRT_SECURE_NO_WARNINGS;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <RuntimeLibrary>MultiThreadedDLL</RuntimeLibrary>
+    </ClangCompile>
+  </ItemDefinitionGroup>
+  <ItemGroup>
+    <ClCompile Include="$(StarboardBasePath)\tests\unittests\Framework\Framework.cpp" />
+    <ClCompile Include="..\..\..\..\tests\UnitTests\Benchmark\EntryPoint.cpp" />
```
ms-jihua: nit: standardize on StarboardBasePath

build/build.sln
ms-jihua: it happened again?

tests/Benchmark/TextBenchmarkTests.mm
```diff
@@ -0,0 +1,545 @@
...
+    inline void Run() {
+        CTFrameDraw(m_frame, m_context);
+    }
+
+    size_t GetRunCount() const {
+        return 100;
+    }
+};
+
+class CTFrameDrawGroup : public CTFrameDrawBase {
```
ms-jihua: DrawGroup:
 - n = 2 is better than n = 1 but probably still too few? other people can object if they think i'm being silly
 - I would still like these Pre/PostRun functions to be abstracted out among the DrawGroup tests

ms-jihua:
optional: For semi-vain reasons, can you post results from after #1878 as well? I'd like to see what the perf impact was more broadly.

tests/Benchmark/TextBenchmarkTests.mm
aballway: Oh I misunderstood what you were getting at.  I'm not sure that they could be abstracted out w/o running into diamond problem.  I changed FrameDrawGroup to use 5, which is giving a pretty consistently low stddev.  I want to keep the default value low, but if others think it's worth bumping up I'll change it.

build/build.sln
aballway: Visual studio needs to chill 🍦

tests/Benchmark/TextBenchmarkTests.mm
```diff
@@ -542,4 +542,16 @@ size_t GetRunCount() const {
...
+    CTFontRef font = CTFontCreateWithName(CFSTR("Arial"), 24, nullptr);
+    CFRelease(font);
+}
+
+TEST_BENCHMARK(CoreText, CTFontDescriptor, 1000) {
+    CTFontDescriptorRef descriptor = CTFontDescriptorCreateWithNameAndSize(CFSTR("Arial"), 88.5);
+    CTFontRef font = CTFontCreateWithFontDescriptor(descriptor, 25.25, nullptr);
+    CFRelease(descriptor);
+    CFRelease(font);
+}
```
ms-jihua: nit: add a newline at the end

tests/Benchmark/TextBenchmarkTests.mm
ms-jihua: 5 seems fine to me

tests/Benchmark/TextBenchmarkTests.mm
```diff
@@ -0,0 +1,557 @@
...
+TEST_BENCHMARK_F(CoreText, CTRunDrawRotated);
+
+class ShowGlyphsBase : public TextBenchmarkBase {
+public:
+    ShowGlyphsBase() : m_glyphs(std::extent<decltype(sc_chars)>::value), m_font(std::move(CGFontCreateWithFontName(CFSTR("Arial")))) {
+        CGContextSetFont(m_context, m_font);
+        CGContextSetFontSize(m_context, 20);
+
+        auto ctFont = woc::MakeAutoCF<CTFontRef>(CTFontCreateWithName(CFSTR("Arial"), 20, nullptr));
+        EXPECT_TRUE(CTFontGetGlyphsForCharacters(ctFont, sc_chars, m_glyphs.data(), std::extent<decltype(sc_chars)>::value));
```
rajsesh-msft: >CTFontGetGlyphsForCharacters [](start = 20, length = 28)

make one more test that benchmarks this api?

tests/Benchmark/EntryPoint.cpp
```diff
@@ -0,0 +1,146 @@
...
+    LOG_INFO("No arguments given to benchmark tests. Defaulting to logging results");
+    s_publisher.reset(new LogBenchmarkTestPublisher());
+}
+
+std::shared_ptr<BenchmarkTestPublisher> BenchmarkTestCreator::GetPublisher() {
+    return s_publisher;
+}
+
+int main(int argc, char** argv) {
+#ifdef WIN32
```
rajsesh-msft: We dont need this to run in reference platform, it that will make things easy w.r.t future integration to TAEF.

rajsesh-msft:
:shipit:

tests/Benchmark/EntryPoint.cpp
```diff
@@ -0,0 +1,152 @@
...
+    virtual void RegisterTestResults(const std::string& testName, const std::vector<Microseconds>& results) {
+        m_maxNameWidth = max(m_maxNameWidth, testName.size());
+        TestPublisherBase::RegisterTestResults(testName, results);
+    }
+
+    virtual void PublishTestResults() {
+        // Align output horizontally
+        std::string spaces(m_maxNameWidth - 5, ' ');
+        std::cout << "Test Name" << spaces << "|"
+                  << "Total Runtime"
```
DHowett-MSFT: You should almost universally prefer string formatting column widths instead of using `\t`.

DHowett-MSFT: `std::setw`!

---
In reply to: [99246724](https://github.com/Microsoft/WinObjC/pull/1876#discussion_r99246724) [](ancestors = 99246724)

tests/Benchmark/BenchmarkTest.h
```diff
@@ -0,0 +1,138 @@
...
+};
+
+namespace internal {
+
+// Helper method shared between BenchmarkTestRunner classes
+// This is called to actually run and benchmark the tests
+// Can't be shared internally because parameterized tests need a different inheritance than non
+template <typename T>
+static void __RunTest(T& test) {
+    using Microseconds = double;
```
DHowett-MSFT: >Microseconds [](start = 10, length = 12)

why not just use `std::duration`?

DHowett-MSFT: like: typedef Microseconds to that, instead of feeding microseconds to it as a template arg

---
In reply to: [99247345](https://github.com/Microsoft/WinObjC/pull/1876#discussion_r99247345) [](ancestors = 99247345)

tests/Benchmark/BenchmarkTest.h
```diff
@@ -0,0 +1,138 @@
...
+        ::testing::internal::__RunTest(m_test);                                                         \
+}                                                                                                       \
+inline void _BENCHMARK_CLASSNAME(test_name, test_case_name)::Run()
+
+#define TEST_BENCHMARK_F(test_name, test_class_name)                                                                                        \
+GTEST_TEST_(test_name, test_class_name, ::testing::internal::BenchmarkTestRunner<test_class_name>, ::testing::internal::GetTestTypeId()) {  \
+        ::testing::internal::__RunTest(m_test);                                                                                             \
+}
+
+#define BENCHMARK_REGISTER_TEST_P(test_name, test_class, generator, ...)                                                \
```
DHowett-MSFT: >BENCHMARK_REGISTER_TEST_P [](start = 8, length = 25)

I don't get why we've cloven to the name `TEST`

We could use macros very similar to gtest's if we simply standardize on calling them `BENCHMARK`.

```
BENCHMARK(a, b, c)

BENCHMARK_F(a, b, c)

INSTANTIATE_BENCHMARK_CASE_P(a, b, c)
```

DHowett-MSFT: Rationale: _they are not tests_

---
In reply to: [99247720](https://github.com/Microsoft/WinObjC/pull/1876#discussion_r99247720) [](ancestors = 99247720)

tests/Benchmark/BenchmarkSampleTests.mm
```diff
@@ -0,0 +1,129 @@
...
+    NSMutableArray* m_array;
+
+public:
+    void PreRun() {
+        m_array = [[NSMutableArray alloc] initWithCapacity:1000];
+    }
+
+    inline void Run() {
+        [m_array release];
+    }
```
DHowett-MSFT: aaaaaaaaaaaaaaaaaa

tests/Benchmark/BenchmarkTest.h
aballway: duration doesn't have operator*, / defined so I needed to convert out to do the calucations for stdDev, which had me saving double rather than Microseconds for my values which I didn't do because I didn't want to obscure that they represent microseconds

tests/Benchmark/BenchmarkSampleTests.mm
aballway: ?

bbowman:
Prerun seems to be conflated with setup. What are the semantics for failures in the prerun method? Does this work with test_p?

tests/Benchmark/EntryPoint.cpp
```diff
@@ -0,0 +1,152 @@
...
+//******************************************************************************
+
+#include <windows.h>
+#include <TestFramework.h>
+#include "BenchmarkTestPublisher.h"
+#include <fstream>
+#include <algorithm>
+
+#ifdef WIN32
+#include <wrl\wrappers\corewrappers.h>
```
DHowett-MSFT: please standardize on `/` for include paths.

There is a visual studio option for this (!)

aballway:
How do you mean conflated? This isn't testing anything in the sense of Unit Tests, so the only kind of failure that we should see would be a fatal failure, and that would just cause that particular test to fail without registering results.  Yes, by using the BENCHMARK_REGISTER_CASE_P macro.

---
In reply to: [277131527](https://github.com/Microsoft/WinObjC/pull/1876#issuecomment-277131527) [](ancestors = 277131527)

aballway:
Ping!

tests/Benchmark/Benchmark.h
```diff
@@ -0,0 +1,142 @@
...
+// Can't be shared internally because parameterized tests need a different inheritance than non
+template <typename T>
+static void __RunCase(T& test, const char* caseName) {
+    using Microseconds = double;
+    size_t runCount = test.GetRunCount();
+    ASSERT_LT(1, runCount);
+    std::vector<Microseconds> results(runCount);
+    @autoreleasepool {
+        for (size_t i = 0; i < runCount; ++i) {
+            test.PreRun();
```
DHowett-MSFT: nit: is it worthwhile to aggregate time spent in setup and teardown?

aballway: That is done in the constructor/destructor of test, this is for things that need to be done once before/after each test run

tests/Benchmark/EntryPoint.cpp
```diff
@@ -0,0 +1,157 @@
...
+    bool created = false;
+    for (int i = 1; i < argc && argv[i]; ++i) {
+        char* arg = argv[i];
+        if (strncmp(arg, "--out=", 6) == 0) {
+            s_publisher.reset(new CSVBenchmarkPublisher(std::move(std::string(arg + 6))));
+            created = true;
+        }
+    }
+
+    if (!created) {
```
DHowett-MSFT: >created [](start = 9, length = 7)

`std::shared_ptr` has an `operator bool` that you could use to check s_publisher's residence.

DHowett-MSFT: (instead of keeping track of this silly thing.)

---
In reply to: [99419315](https://github.com/Microsoft/WinObjC/pull/1876#discussion_r99419315) [](ancestors = 99419315)

tests/Benchmark/TextBenchmarkTests.mm
```diff
@@ -0,0 +1,587 @@
...
+
+    inline void Run() {
+        for (size_t i = 0; i < 100; ++i) {
+            CTFrameDraw(m_frame, m_context);
+        }
+    }
+
+    void PostRun() {
+        _CGContextPopEndDraw(m_context);
+    }
```
DHowett-MSFT: If PostRun time doesn't count, you're ignoring how long it takes to _actually draw_ here; is that your intent?

With the new CG model, every frame draw will be batched and flushed once down here.

DHowett-MSFT: Making _this_ the expensive operation.

---
In reply to: [99419564](https://github.com/Microsoft/WinObjC/pull/1876#discussion_r99419564) [](ancestors = 99419564)

aballway: :O no it is not holy moley

aballway:
@ms-jihua (before vs after drawing change) Yuge Savings HOly guacamole 
image

PR #1877, opened 2017-01-31
Firing UIScrollView's scrollViewDidScroll: delegate call from within …

### LINKED ISSUE #1865: [x86] CTCatalog - Font Fucntions - crashes when attempting to change Select nameKey
This is on the 1701 payload that's in master - 20170127.1

jaredhms:
…Xaml ScrollViewer's ScrollViewerViewChanged event leads to an AV in Xaml if the delegate removes the UIScrollView from the Xaml UIElement tree.  To work around the issue, push the delegate call to a subsequent run of the UI thread.

Fixes #1865.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1877)

### DIFF
--- a/Frameworks/UIKit/UIScrollView.mm
+++ b/Frameworks/UIKit/UIScrollView.mm
@@ -894,7 +894,11 @@ static void clipPoint(UIScrollView* o, CGPoint& p, bool bounce = true) {
     [self setNeedsLayout];
 
     if ([self.delegate respondsToSelector:@selector(scrollViewDidScroll:)]) {
-        [self.delegate scrollViewDidScroll:self];
+        // NOTE: We have to push this off of the ScrollViewer's callback thread to work around
+        // issue #1865.
+        dispatch_async(dispatch_get_main_queue(), ^{
+            [self.delegate scrollViewDidScroll:self];
+        });
     }
 }
 


jaredhms:
I tested several apps with this change and couldn't identify any issues.  That being said, additional testing/validation would be appreciated.

Frameworks/UIKit/UIScrollView.mm
```diff
@@ -894,7 +894,11 @@ - (void)_refreshOrigin {
     [self setNeedsLayout];
 
     if ([self.delegate respondsToSelector:@selector(scrollViewDidScroll:)]) {
-        [self.delegate scrollViewDidScroll:self];
+        // NOTE: We have to push this off of the ScrollViewer's callback thread to work around
+        // issue #1865.
+        dispatch_async(dispatch_get_main_queue(), ^{
```
yiyang-msft: >dispatch_async [](start = 8, length = 14)

my only concern is that we probably should dispatch_async all delegate calls if we do this, otherwise, the sequence between this and those which used to fire after this might be changed. But This might or might not, unfortunately we don't have enough apps to tell.  I 'd prefer to work-around this on app given this is xaml framework issue based on commucation. 

Thoughts?

jaredhms: I'm not positive this is the best solution, but I wanted to get it out for review.  I think spot-fixing apps should be the last resort solution to any problem.  A proper fix from the Xaml team would be ideal, but that would require everyone to be running/testing on latest RS2 builds.  Also, if we do move back to on-thread scrolling, this issue would go away.  Let's synch up in person with Terry after we hear back from the Xaml team.

---
In reply to: [98736679](https://github.com/Microsoft/WinObjC/pull/1877#discussion_r98736679) [](ancestors = 98736679)

oliversa-msft:
:shipit:

rajsesh-msft:
@jaredhms @yiyang-msft - this may not fix everything, but this should at the very least fix one crash we are seeing reliably - #1865 .  Want to merge this?

jaredhms:
@rajsesh-msft; it's probably worth merging.  @yiyang-msft, do you have time to do so?

yiyang-msft:
@jaredhms , i will close this pull request and re-evaulate.  @rajsesh-msft, i will take the bug #1865. Since this change alone does change the delegate firing sequence. We could potentially need do that for all delegate call in order to presverve the right ordering.

Issue #1389, opened 2016-11-15
[CGD2D] Draw primitives with Porter-Duff blend modes


DHowett-MSFT:
Direct2D supports a subset of CG's blend modes as direct composition modes, and a subset of CG's porter-duff blend modes with its blend effect.

The blend effect, however, requires a full target readback and recompose.

The scope of this issue is to support only the basic non-blend-effect composition modes for 1611.

```
kCGBlendModeNormal = kCGBlendModeSourceOver
kCGBlendModeSourceOver
kCGBlendModeDestinationOver
kCGBlendModeSourceIn
kCGBlendModeDestinationIn
kCGBlendModeSourceOut
kCGBlendModeDestinationOut
kCGBlendModeSourceAtop
kCGBlendModeDestinationAtop
kCGBlendModeXOR
kCGBlendModePlusLighter
kCGBlendModeCopy
kCGBlendModePlusDarker // Not supported
```

rajsesh-msft:
Still have some open issues here that need to be sorted out.  May not be required for merge back to develop.

PR #1878, opened 2017-01-31
Change targets less in CGContext

### LINKED ISSUE #1810: Change targets fewer times in post-CGD2D CGContext
Changing targets in ID2D1DeviceContext causes a flush. On CGD2D, every primitive draw in CGContext on CGD2D changes targets to a temporary command list, or something like it. Because of this, #1750 either has no performance impact or actually _hurts_ it. We need to rewrite some of the code to change targets fewer times, possibly tying this to PushBegin/PopEndDraw.

ms-jihua:
- CGContext previously changed targets for all draw primitive functions.
   Since changing targets causes an immediate flush, this significantly negatively impacted performance.
   Changed implementation to only change targets when needed (when shadows are enabled)
 - Added drawing tests for a few edge cases regarding shadows and clipping.
 - Removed unnecessary Escape/Unescape functions - these were not necessary to change targets as previously surmised.

Fixes #1810


tests/unittests/CoreGraphics.Drawing/CGContextDrawing_ShadowTests.cpp
```diff
@@ -50,3 +50,53 @@ DISABLED_DRAW_TEST_F(CGContext, ShadowWithRotatedCTM, WhiteBackgroundTest<>) {
 
     CGContextStrokeRect(context, rect);
 }
+
+DISABLED_DRAW_TEST_F(CGContext, ShadowOverlap, WhiteBackgroundTest<>) {
```
ms-jihua: these three are disabled for the same reason as the other shadow tests - our platform has a minor divergence in the blurriness and darkness of the shadows compared to the reference platform. (this difference is small enough that it's difficult to distinguish by eye, so it's not a blocker) manually verified that these test results diverged only in that respect.

ms-jihua:
Approximate perf impact, as measured in XAMLCatalog custom text view: CTLineDraw previously occupies about 45% of time in [NSString(UIKitAdditions) drawInRect:withFont:lineBreakMode:alignment:], now down to <5%.

Frameworks/include/CGContextInternal.h
```diff
@@ -50,12 +50,4 @@ COREGRAPHICS_EXPORT CGImageRef CGBitmapContextGetImage(CGContextRef ctx);
...
-COREGRAPHICS_EXPORT void _CGContextPopEndDraw(CGContextRef ctx);
-
-// If currently in a Begin/EndDraw stack, Escape will EndDraw(), Unescape will BeginDraw()
-// For scenarios where a Begin/EndDraw pair needs to be temporarily escaped, to be returned to at a later time
-// Ie:
-//      - Switching render targets - Illegal to do so if currently in a Begin/EndDraw pair
-// Also counts in a stack-like manner, so that the escape and unescape only happen once
-COREGRAPHICS_EXPORT void _CGContextEscapeBeginEndDrawStack(CGContextRef ctx);
-COREGRAPHICS_EXPORT void _CGContextUnescapeBeginEndDrawStack(CGContextRef ctx);
+COREGRAPHICS_EXPORT void _CGContextPopEndDraw(CGContextRef ctx);
```
rajsesh-msft: nit: please throw a new line at the end of the line.

tests/unittests/CoreGraphics.Drawing/CGContextDrawing_ShadowTests.cpp
rajsesh-msft: I am not sure if this is how we want it.  For draw tests, our plan is to document the differences with reference platform and then replace the reference images with the baseline winobjc images.  This way, the tests will catch future regressions.  @DHowett-MSFT?

Frameworks/include/CGContextInternal.h
rajsesh-msft: In other files too.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -2253,55 +2191,77 @@ void CGContextClearRect(CGContextRef context, CGRect rect) {
...
+        ComPtr<ID2D1Image> originalTarget; // Cache the original target to restore it later.
+        deviceContext->GetTarget(&originalTarget);
+
+        ComPtr<ID2D1CommandList> commandList;
+        RETURN_IF_FAILED(deviceContext->CreateCommandList(&commandList));
+
+        deviceContext->SetTarget(commandList.Get());
+
+        deviceContext->SetTransform(__CGAffineTransformToD2D_F(transform));
+        RETURN_IF_FAILED(std::forward<Lambda>(drawLambda)(this, deviceContext.Get()));
```
rajsesh-msft: On failure, the transform is not restored, here and below.  I am not sure how relevant that is.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -2253,55 +2191,77 @@ void CGContextClearRect(CGContextRef context, CGRect rect) {
...
+        }
+
+        deviceContext->DrawImage(currentImage.Get());
 
-    if (layer) {
-        deviceContext->PopLayer();
+    } else {
+        deviceContext->SetTransform(__CGAffineTransformToD2D_F(transform));
+        RETURN_IF_FAILED(std::forward<Lambda>(drawLambda)(this, deviceContext.Get()));
+        deviceContext->SetTransform(D2D1::IdentityMatrix());
```
rajsesh-msft: This is the other part.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -2253,55 +2191,77 @@ void CGContextClearRect(CGContextRef context, CGRect rect) {
...
-    ComPtr<ID2D1Effect> shadowEffect;
-    RETURN_IF_FAILED(_CreateShadowEffect(currentImage.Get(), &shadowEffect));
-    if (shadowEffect) {
-        RETURN_IF_FAILED(shadowEffect.As(&currentImage));
-    }
+        // TODO GH#1194: We will need to re-evaluate Direct2D's D2DERR_RECREATE when we move to HW acceleration.
+        this->PopEndDraw();
+    });
+
+    if (state.HasShadow()) {
```
DHowett-MSFT: I'd like to see this abstracted out into something like `context.ShouldBuffer()`; additionally:

If you create a CGBitmapContext and pass `data`, we cannot in good faith buffer.
If you create a CGBitmapContext and pass `!data`, we can buffer. **Except**:
    1. If the user calls `CGBitmapContext(Get|Create)Image`, we must flush the buffer before we give them the image.
    2. If the user calls `CGBitmapContextGetData`, we must flush the buffer _and turn off future buffering_.

We need to make sure that anybody who hasn't signed a buffering contract with us gets immediate mode rendering, to most accurately mimic the reference platform.

We might want to consider an expanded version of `CGBitmapContextCreate(...)` or a custom `CGBitmapContextSet[WOC]BufferingMode()` that allows the user to turn _on_ buffering by agreeing to our contract (must `Flush`)

DHowett-MSFT: Additionally, it might be worthwhile to implement CGContextFlush at this juncture so that the bitmap context APIs can use it.

Frameworks/CoreGraphics/CGContext.mm
ms-jihua: It wouldn't currently cause any issues considering we set it to a new matrix at the start of every draw anyway, but it probably doesn't hurt to make it safer.

---
In reply to: [98774621](https://github.com/Microsoft/WinObjC/pull/1878#discussion_r98774621) [](ancestors = 98774621)

Frameworks/CoreGraphics/CGContext.mm
ms-jihua: I know I deferred you previously, but I still think this is out of scope? Currently, CGBitmapContexts that do not opt into PushBeginDraw()/PopEndDraw() get immediate mode rendering, as per reference platform. What you're asking me to do is to enable buffering for a specific subset of CGBitmapContexts, with caveats to make it conform to the expected behavior. However, I don't believe we've seen evidence that this subset is a significant perf concern, nor that your suggested optimization, with its caveats, would significantly impact such a bottleneck. As such, I still suggest we defer this recommendation until such a bottleneck actually arises.

DHowett-MSFT: I understand. Then, at least, I'd like to keep the suggestion to rename it. While we only have shadows for now, #1389 may require a buffer readback/target swap as well. We'll need the same "create command list before composition" behaviour.

ms-jihua: Done

PR #1879, opened 2017-01-31
Fix the Accelerate page in WOCCatalog to not depend on RGBA buffers.

### LINKED ISSUE #1829: [CGD2D Validation] WOCCatalog Accelerate 1 page crashes because it assumes RGBA
CGImage returning RGBA was an implementation detail; it is incorrect for the vImage test to rely on it.

DHowett-MSFT:
This fixes #1829 and partially addresses some of our format conversion issues.

We will need to evaluate our image loader and the formats it outputs to ensure that UIImage will always (or almost always) return an expected format for vImage users. This pull request does not do that.

Additionally, it removes some dead code!

### DIFF
--- a/samples/WOCCatalog/WOCCatalog/AccelerateViewController.m
+++ b/samples/WOCCatalog/WOCCatalog/AccelerateViewController.m
@@ -33,6 +33,24 @@
 
 static const double meanDivisor = 100;
 
+static CGImageRef _CGImageCreate32bppCopy(CGImageRef image) {
+    size_t originalBpp = CGImageGetBitsPerPixel(image);
+    if (originalBpp == 32) {
+        return CGImageRetain(image);
+    }
+    size_t width = CGImageGetWidth(image);
+    size_t height = CGImageGetHeight(image);
+    CGColorSpaceRef rgbColorSpace = CGColorSpaceCreateDeviceRGB();
+    CGContextRef bitmapContext =
+        CGBitmapContextCreate(NULL, width, height, 8, width * 4, rgbColorSpace, kCGBitmapByteOrder32Big | kCGImageAlphaNoneSkipLast);
+    CGColorSpaceRelease(rgbColorSpace);
+
+    CGContextDrawImage(bitmapContext, (CGRect){ { 0, 0 }, { width, height } }, image);
+    image = CGBitmapContextCreateImage(bitmapContext);
+    CGContextRelease(bitmapContext);
+    return image;
+}
+
 @implementation AccelerateViewController {
     int _accelerateImageNumber;
     int _valueRed;
@@ -99,7 +117,10 @@ static const double meanDivisor = 100;
     _bLabel = [[UILabel alloc] init];
     _boxLabel = [[UILabel alloc] init];
 
-    _img = [UIImage imageNamed:[NSString stringWithFormat:@"photo%d.jpg", _accelerateImageNumber]];
+    UIImage* image = [UIImage imageNamed:[NSString stringWithFormat:@"photo%d.jpg", _accelerateImageNumber]];
+    CGImageRef convertedImage = _CGImageCreate32bppCopy(image.CGImage);
+    _img = [UIImage imageWithCGImage:convertedImage];
+    CGImageRelease(convertedImage);
 }
 
 - (NSInteger)tableView:(UITableView*)tableView numberOfRowsInSection:(NSInteger)section {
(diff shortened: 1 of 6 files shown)


samples/WOCCatalog/WOCCatalog/AccelerateViewController.m
```diff
@@ -33,6 +33,24 @@
 
 static const double meanDivisor = 100;
 
+static CGImageRef _CGImageCreate32bppCopy(CGImageRef image) {
+    size_t originalBpp = CGImageGetBitsPerPixel(image);
```
MSFTFox: no need for temp, just toss this in the if.

msft-Jeyaram:
:shipit:

samples/WOCCatalog/WOCCatalog/AccelerateViewController.m
```diff
@@ -33,6 +33,24 @@
 
 static const double meanDivisor = 100;
 
+static CGImageRef _CGImageCreate32bppCopy(CGImageRef image) {
```
rajsesh-msft: Given that this doesnt address the format conversion issues, i'd rather not fix this, and have an open issue to put in the real fix, which indeed provides the pixels in the proper expected format.

rajsesh-msft:
:clock1:

samples/WOCCatalog/WOCCatalog/AccelerateViewController.m
rajsesh-msft: Synced up offline.  The issue is that the original code assumed UIImage implementation details, this change looks at the actual format instead of making assumptions.  

---
In reply to: [98587425](https://github.com/Microsoft/WinObjC/pull/1879#discussion_r98587425) [](ancestors = 98587425)

DHowett-MSFT:
The title of this pill request is now wrong. There were five different bugs here.

1. Accelerate I depended on 32-bit buffers, but UIImage only returns those as an implementation detail.
    **Fix:** Convert the image necessary.
2. Accelerate I did not have a provision for `kCGBitmapByteOrderDefault`, which iOS returns. It's BE for 32-bit. It was checking "Big Endian or Else". Therefore, its formats were backwards.
3. Accelerate II did not have a provision for `kCGBitmapByteOrderDefault`, and its formats may have been backwards.
4. The WIC format table introduced in 2212f2a had a typo, inserting an RGB where a BGR was required.
5. vImage was attempting to do a format conversion via a no-op CGBitmapContext (creating it with data and then converting it into an image). Now that CGImage supports a wide variety of input formats and CGBitmapContext **doesn't**, this was incorrect.

samples/WOCCatalog/WOCCatalog/AccelerateViewController.m
```diff
@@ -99,7 +117,10 @@ - (void)viewDidLoad {
     _bLabel = [[UILabel alloc] init];
     _boxLabel = [[UILabel alloc] init];
 
-    _img = [UIImage imageNamed:[NSString stringWithFormat:@"photo%d.jpg", _accelerateImageNumber]];
+    UIImage* image = [UIImage imageNamed:[NSString stringWithFormat:@"photo%d.jpg", _accelerateImageNumber]];
+    CGImageRef convertedImage = _CGImageCreate32bppCopy(image.CGImage);
+    _img = [UIImage imageWithCGImage:convertedImage];
+    CGImageRelease(convertedImage);
```
rajsesh-msft: throw this into a static function and reuse?

rajsesh-msft:
:shipit:

msft-Jeyaram:

nit: update

---
Refers to: Frameworks/UIKit/UIImage.mm:4 in 409c5aa. [](commit_id = 409c5aa57bc043d8a6a17d7de2dd81c9fb41fcee, deletion_comment = False)

DHowett-MSFT:
The changes in the actual vImage consumption code in Accelerate tests are _spacing only._

msft-Jeyaram:
:shipit:

samples/WOCCatalog/WOCCatalog/AccelerateViewController.m
DHowett-MSFT: I like it, but I'll defer to what eh have here. Accelerate's integration tests are not really great, and they could use a lot of cleanup. Setting the image twice isn't even necessary here, so we should evaluate just not doing it.

PR #1880, opened 2017-01-31
vsimporter changes for the new packaging world

memontic-ms:
1. vsimporter project source has been modified to create a new packaging project type (.nuproj). 
- This packaging project will automatically be included to the solution if a project is imported with the -genprojections flag.
- A packaging project can be force included/excluded from the generated solution with the -genpackaging flag.

2. vsimporter project templates have been modified. 
- Imported projects will now depend on WinObjC.* packages instead of WINOBJC_SDK_ROOT source

3. WinObjC.Packaging project has been added to tools.sln/WinObjC.Build.sln. 
- This project acts as an extension of Nuget.Build.Packaging with WinObjC specific targets/props.
- In its current state, the WinObjC.Packaging package is only intended to be included from imported projects and not from the WinObjC source projects (foundation, uikit, etc).


tools/WinObjC.Tools/vsimporter-templates/packageable.project.json
```diff
@@ -0,0 +1,21 @@
+{
+  "dependencies": {
+    "WinObjC.Packaging": "*",
+    "WinObjC.Frameworks": "*",
+    "WinObjC.Frameworks.UWP": "*",
+    "WinObjC.Frameworks.Core": "*",
+    "WinObjC.Language": "*",
+    "WinObjC.Logging": "*"
```
pradipd: Should it be all the nuget packages?  Or just the WinOjbc.Frameworks.UWP?  I thought there was a dependency chain.

tools/WinObjC.Tools/vsimporter-templates/packageable.project.json
```diff
@@ -0,0 +1,21 @@
+{
+  "dependencies": {
+    "WinObjC.Packaging": "*",
```
pradipd: Just so I understand (from the other CR).

project.json is for iOS apps that run through vsimporter.

packageable.project.json is for iOS middleware that runs through vsimporter.  So, we need to include winobjc.packaging for iOS middleware to create a nuget package.

tools/WinObjC.Tools/vsimporter-templates/packageable.project.json
pradipd: >"*" [](start = 23, length = 3)

Should this be * or the last SDK release?
Same with project.json

I assume a user doesn't want to automatically upgrade to a new SDK without knowing it.

tools/WinObjC.Packaging/WinObjC.Packaging.nuproj
```diff
@@ -0,0 +1,38 @@
...
+    <ProjectGuid>ddc2aa5f-c8ae-48cb-adb7-9278985c3db7</ProjectGuid>
+  </PropertyGroup>
+  <PropertyGroup Label="Configuration" />
+  <ItemGroup>
+    <None Include="WinObjC.Packager.props" />
+    <None Include="WinObjC.Packager.targets" />
+    <None Include="WinObjC.Packageable.props" />
+    <None Include="WinObjC.Packageable.targets" />
+    <None Include="project.json" />
+    <None Include="ImportedByThisProj.targets" />
```
pradipd: >ImportedByThisProj [](start = 19, length = 18)

Do you have to import and include the project?

tools/WinObjC.Tools/vsimporter-templates/packageable.project.json
memontic-ms: To be decided. I was having some difficulties since these packages were all being built as prereleases. Need input from Brian.

---
In reply to: [98955335](https://github.com/Microsoft/WinObjC/pull/1880#discussion_r98955335) [](ancestors = 98955335)

tools/WinObjC.Tools/vsimporter-templates/WinStore10-App/project.json
```diff
@@ -0,0 +1,20 @@
+{
```
memontic-ms: This file can be removed since it is a duplicate of vsimporter-templates/project.json

tools/WinObjC.Tools/vsimporter-templates/packageable.project.json
memontic-ms: Yes, but project.json can be also for iOS middleware that does not wish to be packageable. Although I'd assume most middleware would like to be packaged, this gives them the option to opt out of packaging.

Currently, the default for importing any project is to not add the packaging project to the solution, ie. make it not packageable. 
However, projects imported with -genprojections will automatically add the packaging project to the solution.
Should the default instead be to include the packaging project?

---
In reply to: [98952538](https://github.com/Microsoft/WinObjC/pull/1880#discussion_r98952538) [](ancestors = 98952538)

tools/WinObjC.Packaging/ImportedByAnyDependent.targets
```diff
@@ -0,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8" standalone="no"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Condition="$(MSBuildProjectExtension) == '.nuproj'" Project="WinObjC.Packager.props" />
+  <Import Condition="$(MSBuildProjectExtension) == '.nuproj'" Project="WinObjC.Packager.targets" />
```
memontic-ms: ><Import Condition="$(MSBuildProjectExtension) == '.nuproj'" Project="WinObjC.Packager.targets" /> [](start = 2, length = 97)

This line and line 7 could go inside their .props files instead. I'm fine with it as it is unless this way is wrong/not a best practice.

tools/WinObjC.Tools/vsimporter-templates/packageable.project.json
bbowman: There is a dependency chain described already. The only difference is what version is picked up due to the constraints being satisfied with an older version or not. We should expect the typical case to just depend on the tip even if our internal samples etc depend on all to force the latest of everything even in prerelease mode.

tools/WinObjC.Tools/vsimporter-templates/packageable.project.json
bbowman: This mostly a question of what it looks like to invoke vsimporter on middleware VS an app. How does vsimporter know it's middleware? Do we need more user obvious knobs to control this?

tools/WinObjC.Tools/vsimporter-templates/packageable.project.json
bbowman: The star is latest *stable* version which I think is a fair thing to auto upgrade to. This is less overhead all around I think.

tools/WinObjC.Packaging/WinObjC.Packaging.nuproj
bbowman: What does that mean @pradipd

memontic-ms:
At some point I deleted the vsimporter templates for the WinStore10-Package props and targets that would be included in WinStore10-Package's package. I'm currently working on adding those back in and checking functionality so that the packages created are actually usable. Making a few upgrades as well.

memontic-ms:
Closing this PR and opening new one.

PR #1883, opened 2017-01-31
Winmd2ObjC update: 

mukhole:
Fixed inconsistencies with create method and added annotations.


DHowett-MSFT:
I don't think any header changes are in here.

mukhole:
They are there, you just cannot see them on github because of the large number of files modified.

---
In reply to: [276532250](https://github.com/Microsoft/WinObjC/pull/1883#issuecomment-276532250) [](ancestors = 276532250)

PR #1891, opened 2017-02-01
Rewrite portions of EbrFile handling to use UTF16 

### LINKED ISSUE #1875: UTF8 encoded file paths don't work as expected on windows
All of the EbrFile* apis use char* and assume "ascii" encoding, which works okay for English. All of core foundation URL apis assume UTF-8 encoding (specifically url % encoding), and we assume it is ascii.

The most visible impact of this is that NSUserDefaults does not synchronize settings if the user name is in non-english and uses accent characters.

rajsesh-msft:
Only big change here was to rewrite CPathMapper.  To prevent regressions there, unit tests were added before refactor.  We pass better than we did before.  Also added a NSFileManager test, to watch path names change from UTF16 to UTF8 back to UTF16.  Some day, we will just go UTF16 everywhere, but that requires rework on the CFURL* front.

Fixes #1875

    * Rewrite path mapper to simplify code and support UTF16 path names.
    * Rewrite Ebr file apis to use new path mapper.
    * Remove dead code.
    * fixed unit tests to build after refactor.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1891)

### DIFF
--- a/Frameworks/Starboard/AssetFile.cpp
+++ b/Frameworks/Starboard/AssetFile.cpp
@@ -32,8 +32,6 @@ static const wchar_t* TAG = L"AssetFile";
 
 #define strtok_r strtok_s
 
-char CPathMapper::currentDir[4096];
-
 void appendPath(char* curpath, const char* path) {
     char copy[4096];
 
@@ -199,7 +197,7 @@ CPathMapper::CPathMapper(const char* path) {
     strcpy_s(relativePath, "");
 
     if (path[0] != '/') {
-        appendPath(relativePath, currentDir);
+        appendPath(relativePath, "");
     }
     appendPath(relativePath, path);
     fixedValid = fixPath(fixedPath, relativePath);
--- a/Frameworks/Starboard/AssetFile.cpp
+++ b/Frameworks/Starboard/AssetFile.cpp
@@ -79,8 +79,9 @@ bool EbrFSDirReader::readNext(EbrDir* curDir, EbrDirEnt* ent) {
 EbrDir* EbrOpenDir(const char* path) {
     CPathMapper map(path);
 
-    if (!map)
+    if (!map) {
         return NULL;
+    }
 
     EbrDirReader* fsReader = EbrFSDirReader::open(map.MappedPath());
     if (!fsReader) {
--- a/Frameworks/Starboard/AssetFile.cpp
+++ b/Frameworks/Starboard/AssetFile.cpp
@@ -123,6 +123,7 @@ bool EbrIsDir(const char* path) {
 }
 
 int EbrStat(const char* filename, struct stat* ret) {
+#ifdef _USE_32BIT_TIME_T
     CPathMapper map(filename);
     if (!map) {
         TraceError(TAG, L"EbrStat failure!");
@@ -137,6 +138,12 @@ int EbrStat(const char* filename, struct stat* ret) {
     }
 
     return _wstat32(map, reinterpret_cast<struct _stat32*>(ret));
+#else
+    // This is from stat.h, unfortunately, it doesn't define the wstat apis for non 32-bit time_t
+    // essentially on our platform, EbrStat is same as EbrStat64i32.
+    // keeping this #ifdef just in case.
+    return EbrStat64i32(filename, reinterpret_cast<struct _stat64i32*>(ret));
+#endif
 }
 
 int EbrStat64i32(const char* filename, struct _stat64i32* ret) {
--- a/Frameworks/Starboard/AssetFile.cpp
+++ b/Frameworks/Starboard/AssetFile.cpp
@@ -14,198 +14,15 @@
 //
 //******************************************************************************
 
-#include <Windows.h>
-#include <objbase.h>
-#include <algorithm>
-#include <map>
-#include <sys/stat.h>
-#include <direct.h>
-#include <io.h>
-
 #include "Starboard.h"
-#include "Platform/EbrPlatform.h"
+#include "ErrorHandling.h"
+#include "StringHelpers.h"
+#include <sys/stat.h>
+
 #include "AssetFile.h"
 #include "PathMapper.h"
-#include "LoggingNative.h"
 
 static const wchar_t* TAG = L"AssetFile";
-
-#define strtok_r strtok_s
-
-void appendPath(char* curpath, const char* path) {
-    char copy[4096];
-
-    strcpy_s(copy, path);
-
-    char* save;
-    char* curToken = strtok_r(copy, "/\\", &save);
-    char* curpathEnd = curpath + strlen(curpath);
-    while (curToken) {
-        if (strlen(curToken) == 0) {
-            curToken = strtok_r(NULL, "/\\", &save);
-        }
-        int tokenLen = strlen(curToken);
-        if (strcmp(curToken, "~") == 0) {
-            strcpy_s(curpath, 2048, "/home");
-            curpathEnd = curpath + strlen(curpath);
-        } else {
-            strcat_s(curpathEnd, 2048, "/");
-            curpathEnd++;
-            strcat_s(curpathEnd, 2048, curToken);
-            curpathEnd += tokenLen;
-        }
-        curToken = strtok_r(NULL, "/\\", &save);
-    }
-    if (strcmp(curpath, "") == 0)
-        strcpy_s(curpath, 2048, "/");
-}
-
-static void EscapePath(char* dest, const char* src) {
-    while (*dest)
-        dest++;
-
-    while (*src) {
-        switch (*src) {
-            case '?':
-                *dest = '+';
-                break;
-
-            default:
-                *dest = *src;
-        }
-
-        dest++;
-        src++;
-    }
-
-    *dest = 0;
-}
-
-bool convertPath(char* filePath, const char* relativePath) {
-    if (!EbrGetRootMapping(NULL, filePath, 4096))
-        return false;
-    int curComponent = 0;
-
-    char copy[4096];
-
-    strcpy_s(copy, relativePath);
-
-    char* save;
-    char* curToken = strtok_r(copy, "/", &save);
-    char* filePathEnd = filePath + strlen(filePath);
-    while (curToken) {
-        if (strlen(curToken) == 0 || strcmp(curToken, ".") == 0) {
-            curToken = strtok_r(NULL, "/", &save);
-            continue;
-        }
-
-        if (curComponent == 0) {
-            if (!EbrGetRootMapping(curToken, filePath, 4096))
-                return false;
-            filePathEnd = filePath + strlen(filePath);
-        } else {
-            strcat_s(filePathEnd, 2048, "\\");
-            filePathEnd++;
-            EscapePath(filePathEnd, curToken);
-            filePathEnd += strlen(curToken);
-        }
-        curComponent++;
-
-        curToken = strtok_r(NULL, "/\\", &save);
-    }
-
-    if (strcmp(filePath, "") == 0)
-        return false;
-
-    return true;
-}
-
-bool fixPath(char* outPath, const char* relativePath) {
-    strcpy_s(outPath, 2048, "");
-    int curComponent = 0;
-
-    if (relativePath[0] != '/')
-        return false;
-
-    char copy[4096];
-
-    strcpy_s(copy, relativePath);
-
-    char* save;
-    char* curToken = strtok_r(copy, "/", &save);
-    char* outPathEnd = outPath + strlen(outPath);
-    while (curToken) {
-        if (strlen(curToken) == 0 || strcmp(curToken, ".") == 0) {
-            curToken = strtok_r(NULL, "/", &save);
-            continue;
-        }
-        if (strcmp(curToken, "..") == 0) {
-            if (curComponent == 0) {
-                return false;
-            }
-            //  Move back one directory
-            char* curPos = outPath + strlen(outPath);
-            while (curPos >= outPath) {
-                if (*curPos == '/') {
-                    *curPos = 0;
-                    break;
-                }
-                curPos--;
-            }
-            //  Impossible!
-            if (curPos < outPath) {
-                return false;
-            }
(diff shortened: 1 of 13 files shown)


Frameworks/Starboard/AssetFile.cpp
```diff
@@ -259,29 +70,19 @@ bool EbrFSDirReader::readNext(EbrDir* curDir, EbrDirEnt* ent) {
...
 
-    // Note that we're doing wcslen here, which is number of characters. If we've got continues
-    // in the stream then we may truncate the buffer.
-    std::string conv(data.cFileName, data.cFileName + wcslen(data.cFileName));
-    strcpy_s(ent->fileName, conv.c_str());
-    char tmpPath[4096];
-    strcpy_s(tmpPath, startPath);
-    strcat_s(tmpPath, "//");
-    strcat_s(tmpPath, conv.c_str());
+    const std::string filename = Strings::WideToNarrow(std::wstring(data.cFileName));
```
DHowett-MSFT: Is WideToNarrow a Multibyte conversion?

Frameworks/Starboard/PathMapper.cpp
```diff
@@ -0,0 +1,136 @@
...
+#include "StringHelpers.h"
+
+#include <algorithm>
+#include <list>
+#include <regex>
+
+#include "Platform/EbrPlatform.h"
+#include "PathMapper.h"
+
+// utility function to tokenize string using delmiters
```
DHowett-MSFT: nit: "delimiters"

Frameworks/Starboard/PathMapper.cpp
```diff
@@ -0,0 +1,136 @@
...
+// d:\src/winobjc ==> d:, src, winobjc
+// /src/winobjc ==> "", src, winobjc
```
DHowett-MSFT: Does the presence of a "" always mean it's an absolute unix path?

Frameworks/Starboard/PathMapper.cpp
```diff
@@ -0,0 +1,136 @@
...
+    std::size_t pos = str.find_first_of(L'?');
+    while (pos != std::wstring::npos) {
+        str[pos] = L'+';
+        pos = str.find_first_of(L'?', pos + 1);
+    }
+}
+
+// normalize relative path
+//  1 remove any "." or "" components
+//  2 from reverse, remove any ".." and preceeding component
```
DHowett-MSFT: nit: I don't understand this comment; it implies we'll walk backwards to eliminate the `..`s.

What it looks like is we iterate forward, and when we encounter a `..` we step the iterator backwards, delete it and the component preceding it.

Therefore, something like `a/b/c/../..` folds to `a`?

Frameworks/Starboard/PathMapper.cpp
```diff
@@ -0,0 +1,136 @@
...
+            continue;
+        } else if (*it == c_UpDir) {
+            // remove previous component if it exists
+            if (it == components.begin()) {
+                // error, no previous directory exists
+                components.clear();
+                break;
+            }
+            // we have to do this in case the last component is ..
+            it = components.erase(--it);
```
DHowett-MSFT: Are we concerned about reverse-iterating a std::list? I thought they were singly-linked, making `--` a taxing operator, but perhaps they're not!

Frameworks/Starboard/PathMapper.cpp
```diff
@@ -0,0 +1,136 @@
...
+static const std::wstring c_specialFolders[] = { L"Documents", L"cache", L"Library", L"AppSupport", L"tmp", L"shared" };
+
+std::wstring _MapPathRoot(const std::wstring& root) {
+    if (root == c_currentDir) {
+        return root;
+    } else if (root == L"~") {
+        return std::wstring(L".\\home");
+    }
+
+    std::wregex drive(L"[a-zA-Z]:");
```
DHowett-MSFT: I'm worried about the performance of regex matching for every path we map.
It'd be easier/cheaper if we were to take the `length >= 2 && wisalpha(root[0]) && root[1] == L':'` approach below.

Frameworks/Starboard/PathMapper.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+    if (std::regex_match(root, drive)) {
+        return root;
+    }
+
+    for (int i = 0; i < _countof(c_specialFolders); ++i) {
+        if (_wcsicmp(root.c_str(), c_specialFolders[i].c_str()) == 0) {
+            return Strings::NarrowToWide<std::wstring>(EbrGetWritableFolder()) + std::wstring(L"\\") + c_specialFolders[i];
```
DHowett-MSFT: Is there room to exchange `EbrGetWritableFolder` with a new `_EbrGetWritableFolder`, its UTF-16 counterpart?

Frameworks/Starboard/PathMapper.cpp
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    std::wregex drive(L"[a-zA-Z]:");
+
+    if (std::regex_match(root, drive)) {
+        return root;
+    }
+
+    for (int i = 0; i < _countof(c_specialFolders); ++i) {
+        if (_wcsicmp(root.c_str(), c_specialFolders[i].c_str()) == 0) {
```
DHowett-MSFT: Since both of these are `std::wstring`s, it'll be better to use == comparison (it can optimize for length differences better than _wcsicmp)

Frameworks/Starboard/PathMapper.cpp
```diff
@@ -0,0 +1,136 @@
...
+        return std::wstring(L".\\home");
+    }
+
+    std::wregex drive(L"[a-zA-Z]:");
+
+    if (std::regex_match(root, drive)) {
+        return root;
+    }
+
+    for (int i = 0; i < _countof(c_specialFolders); ++i) {
```
DHowett-MSFT: nit: `for(auto& specialFolder: c_specialFolders) {` (range for is awesome!)

Frameworks/Starboard/AssetFile.cpp
```diff
@@ -259,29 +70,19 @@ bool EbrFSDirReader::readNext(EbrDir* curDir, EbrDirEnt* ent) {
...
 
 EbrDir* EbrOpenDir(const char* path) {
     CPathMapper map(path);
-    char* fixedName = map.FixedPath();
-    if (!fixedName) {
-        return NULL;
-    }
 
-    if (!map.MappedPath())
+    if (!map)
```
aballway: nit: braces, or if you're so inclined, use one of our fancy macros

Frameworks/Starboard/PathMapper.cpp
aballway: They can be double linked I thought, but given how we're using it might make more sense to just use a std::vector.  FWIU it's generally better to take a hit every once in a while for resizing if we need to than having to jump around when iterating.

Frameworks/Starboard/PathMapper.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+    return std::wstring(L".\\") + root;
+}
+
+std::wstring _PathFromComponents(const std::list<std::wstring>& components) {
+    std::wstring path = components.front();
+
+    auto it = components.begin();
+
+    std::for_each(++it, components.end(), [&path](const std::wstring& comp) {
```
aballway: nit: use std::accumulate

Frameworks/Starboard/PathMapper.cpp
```diff
@@ -0,0 +1,136 @@
...
+        components.emplace_back(str.substr(start, delimPos - start));
+        start = str.find_first_not_of(delims, delimPos);
+        delimPos = str.find_first_of(delims, start);
+    }
+    return components;
+}
+
+// converts any ? to +, we may add more later if needed for compat
+
+static void _EscapeIllegalPathCharacters(std::wstring& str) {
```
aballway: nit: why not return wstring rather than taking a mutable reference?

Frameworks/Starboard/PathMapper.cpp
```diff
@@ -0,0 +1,136 @@
...
+        pos = str.find_first_of(L'?', pos + 1);
+    }
+}
+
+// normalize relative path
+//  1 remove any "." or "" components
+//  2 from reverse, remove any ".." and preceeding component
+// if components becomes empty, return "."
+
+static void _NormalizeRelativePathComponents(std::list<std::wstring>& components) {
```
aballway: same as above

Frameworks/Starboard/PathMapper.cpp
aballway: also don't like this expression it = something(--it), hard to figure out exactly what's going on

Frameworks/Starboard/PathMapper.cpp
```diff
@@ -0,0 +1,136 @@
...
+    for (auto it = components.begin(); it != components.end();) {
```
aballway: put ++it in here rather than in last else?

tests/unittests/Foundation/NSFileManagerTests.mm
```diff
@@ -213,3 +213,14 @@
     // Verify data.
     ASSERT_OBJCEQ([content dataUsingEncoding:NSUTF8StringEncoding], [NSData dataWithContentsOfURL:destURL]);
 }
+
+TEST(NSFileManager, DirectoryWithUTF16Chars) {
+    wchar_t* specialFolder = L"oÖo winobjc";
```
aballway: should add more variety of characters

tests/unittests/Foundation/NSFileManagerTests.mm
```diff
@@ -213,3 +213,14 @@
...
+    NSString* directoryName = [NSString stringWithCharacters:(const unichar*)specialFolder length:wcslen(specialFolder) * sizeof(wchar_t)];
+    NSString* testPath = getPathToFile(directoryName);
```
aballway: test this is the value we expect it to be

Frameworks/Starboard/AssetFile.cpp
rajsesh-msft: Yes, using UTF8.

Frameworks/Starboard/PathMapper.cpp
rajsesh-msft: Yes.  It just means that it was an absolute path.

Frameworks/Starboard/PathMapper.cpp
rajsesh-msft: in-place modification was the goal.

Frameworks/Starboard/PathMapper.cpp
rajsesh-msft: Wrong comment :).  I had intended to go backwards, but realized it was far more effective to go forward.

Frameworks/Starboard/PathMapper.cpp
rajsesh-msft: Same as above :).

Frameworks/Starboard/PathMapper.cpp
rajsesh-msft: No.  continue also goes to ++.

Frameworks/Starboard/PathMapper.cpp
rajsesh-msft: std::list is bidirectional.  Not changing this to vector as sometime in the future when we do properly support > MAX_PATH, there will be several elements here.  And leading '.' is pretty common.  Since the list is all constructed at the same time, locality of reference should not be a problem.

tests/unittests/Foundation/NSFileManagerTests.mm
rajsesh-msft: That is covered elsewhere in NSString (existing) tests.

Frameworks/Starboard/PathMapper.cpp
rajsesh-msft: Unfortunately, i need it to be case insensitive and == isn't.

Frameworks/Starboard/PathMapper.cpp
rajsesh-msft: The right thing would be to replace all the Ebr* methods that take UTF-8 encoded strings to take UTF-16 encoded strings, but that would be a much larger change (many callers).  However, given that EbrGet/SetWritableFolder is not that commonly used, I'll upgrade that pair to UTF16.

tests/unittests/Foundation/NSFileManagerTests.mm
rajsesh-msft: possibly, but this demonstrates the problem adequately.

Frameworks/Starboard/PathMapper.cpp
```diff
@@ -98,15 +99,13 @@ std::wstring _MapPathRoot(const std::wstring& root) {
         return std::wstring(L".\\home");
     }
 
-    std::wregex drive(L"[a-zA-Z]:");
-
-    if (std::regex_match(root, drive)) {
+    if (root.length() == 2 && iswalpha(root[0]) && root[1] == L':') {
```
DHowett-MSFT: :+1:

tests/unittests/Starboard/PathMapperTests.mm
```diff
@@ -69,6 +71,8 @@
 
     CPathMapper mapper16("/AppSupport/.\\?/");
     EXPECT_EQ_MSG(0, wcscmp(mapper16, L"d:\\temp\\AppSupport\\+"), "%S", (const wchar_t*)mapper16);
```
DHowett-MSFT: nit: for these guys, prefer `EXPECT_STREQ(expected, actual)`; gtest/taef will properly log the inputs and use the wchar_t overload of StrHelperCmpEQ

DHowett-MSFT: (it means we don't have to rely on the logging infrastructure in the `_MSG` macro variants, and gtest can print expected/actual next to eachother)

Issue #1724, opened 2017-01-18
Archival.BackwardsCompatibilityWithOldWinObjCArchives test fails on OSX


ms-jihua:
[ RUN      ] Archival.BackwardsCompatibilityWithOldWinObjCArchives
ArchivalTests.mm:444: Failure
Value of: rootObject
  Actual: NULL
Expected: @"RootObject"
Which is: RootObject
[  FAILED  ] Archival.BackwardsCompatibilityWithOldWinObjCArchives (0 ms)

Judging by the name, it's possible that this should be an OSX_DISABLED_TEST

DHowett-MSFT:
oops, this is a winobjc-specific test :cry:

PR #1907, opened 2017-02-02
Disabling windows specific test on OSX

msft-Jeyaram:
#1724


bbowman:
FOO_DISABLED_TEST is supposed to indicate that tests are *temporarily* disabled and will be fixed pending some change. We already have a windows only test binary for foundation for tests that are truly windows specific in their operation. Let's not muddy the waters here...

ms-jihua:
+1 to what @bbowman said (it totally slipped my mind)

DHowett-MSFT:
When you commit this, _please_ reword the commit to actually say what changed.

msft-Jeyaram:
@ping some movements please take a look

bbowman:
:shipit:

PR #1908, opened 2017-02-02
Disabling some NSDecimal subtraction tests on OSX

msft-Jeyaram:


bbowman:
@ram why are we making this change?

msft-Jeyaram:
Why are we disabling this?

 OSX isn't doing a good job with subtraction.
1000-999 = 1, but for them they take the double computation which ends up being 0.99999999997
We can assert near but this would mean we would have to take the double value of the NSDecimalNumber (which will have rounding errors) and sadly thats not fully implemented yet. Given we only have subtraction and addition implemented as of now.

PR #1912, opened 2017-02-03
Use block scope thread_local static for NSProgress progress stack

ehren:
per comments in #1872. Works around issue with file_scope thread_local in current clang. Leak check unit test added.

### DIFF
--- a/Frameworks/Foundation/NSProgress.mm
+++ b/Frameworks/Foundation/NSProgress.mm
@@ -46,16 +46,15 @@ struct CurrentProgress {
     bool childCreated;
 };
 
-// Stack of NSProgress objects that have becomeCurrent
-thread_local static std::shared_ptr<std::stack<CurrentProgress>> s_currentProgressStack;
-
 // Returns the stack of NSProgress objects that have becomeCurrent on the current thread, initializing it if necessary
-static decltype(s_currentProgressStack) & _getProgressStackForCurrentThread() {
-    if (!s_currentProgressStack) {
-        s_currentProgressStack = std::make_shared<std::stack<CurrentProgress>>();
+static std::shared_ptr<std::stack<CurrentProgress>> & _getProgressStackForCurrentThread() {
+    thread_local static std::shared_ptr<std::stack<CurrentProgress>> tlsCurrentProgressStack;
+
+    if (!tlsCurrentProgressStack) {
+        tlsCurrentProgressStack = std::make_shared<std::stack<CurrentProgress>>();
     }
 
-    return s_currentProgressStack;
+    return tlsCurrentProgressStack;
 }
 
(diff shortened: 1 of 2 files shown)


tests/unittests/Foundation/NSProgressTests.mm
```diff
@@ -583,3 +583,16 @@ - (void)addParentTo:(NSProgress*)child {
...
+    NSProgress* progress = [NSProgress progressWithTotalUnitCount:100];
+
+    std::thread t([&progress]() {
+        [progress becomeCurrentWithPendingUnitCount:50];
+    });
+
+    t.join();
+
+    // Progress unfinished upon thread exit. Ensure we're not keeping progress alive:
+    EXPECT_EQ(1, [progress retainCount]);
```
ehren: Not the biggest fan of asserting on the retain count everywhere in tests.... but a conversion to ARC would be straightforward.

Frameworks/Foundation/NSProgress.mm
```diff
@@ -46,16 +46,15 @@
...
-
 // Returns the stack of NSProgress objects that have becomeCurrent on the current thread, initializing it if necessary
-static decltype(s_currentProgressStack) & _getProgressStackForCurrentThread() {
-    if (!s_currentProgressStack) {
-        s_currentProgressStack = std::make_shared<std::stack<CurrentProgress>>();
+static std::shared_ptr<std::stack<CurrentProgress>> & _getProgressStackForCurrentThread() {
+    thread_local static std::shared_ptr<std::stack<CurrentProgress>> tlsCurrentProgressStack;
+
+    if (!tlsCurrentProgressStack) {
+        tlsCurrentProgressStack = std::make_shared<std::stack<CurrentProgress>>();
```
rajsesh-msft: These three lines could be collapsed into thread_local static std::shared_ptr<std::stack<CurrentProgress>> tlsCurrentProgressStack = std::make_shared<std::stack<CurrentProgress>>()

ehren: @rajsesh-msft fixed. Also, it's not expressly allowed by the style guide but what about using auto& / return type deduction to replace the removed use of decltype?

DHowett-MSFT: @ehren we do that in a lot of places where the type is expressed fully on the righthand side :)

ehren: @DHowett-MSFT cool, switched in latest.

ehren: the double auto might be skirting the rules a bit... no prob either way to switch.

rajsesh-msft:
thank you @ehren!

Issue #1904, opened 2017-02-02
Convert Ebr* file apis to use wchar_t instead of char*


rajsesh-msft:
Currently, we have a stop gap with #1875 to support UTF8 characters, but there is back and forth conversion between UTF8 and UTF16 across the Ebr boundary.  

- We can reduce these conversions and get better performance.
- We can finally support long paths (additional work needed).

edvv:
@rajsesh-msft Can you please inform me about this: Are long paths only available with UTF16 API? I thought long paths was more a function of UWP settings and not a particular file system API.

rajsesh-msft:
https://msdn.microsoft.com/library/windows/desktop/aa365247.aspx#maxpath - this is new in the anniversary update (1607), we do need to support 1512 update for the immediate future.  Also, the ascii version of the apis such as mkdir won't support > MAX_PATH.  As such, all the Ebr apis use ascii versions of c runtime file apis, so all bets are off.

edvv:
Ah. On Unix MAX_PATH is 1024. It appears that MAX_PATH is 260 on Windows - which is a pretty severe limitation. So, are you saying that if I convert to wchar_t API such as using _wmkdir() then I get a max math length of 32767?

That fact (if it is true?) is not mentioned at: https://msdn.microsoft.com/en-us/library/2fkk4dzw.aspx

rajsesh-msft:
@eddv a small correction - we did convert the path apis to use the _w version.  If you set the manifest correctly you should have support for long path, but i haven't actually verified everything would work.  This work is to avoid the unnecessary conversions as well as to ensure that long paths indeed works.

PR #1916, opened 2017-02-03
SetTempFolder() should take wchar_t instead of char so that we don't

### LINKED ISSUE #1915: UIApplication SetTemporaryFolder() is losing string encoding information
This causes failures in all code paths using temporary folders.  Thanks @jboich for reporting this.

rajsesh-msft:
lose encoding information. Fixes #1915

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1916)


Frameworks/UIKit/StarboardXaml/ApplicationMain.mm
```diff
@@ -179,6 +179,6 @@ int ApplicationMainStart(const char* principalName,
     return 0;
 }
 
-void SetTemporaryFolder(const char* folder) {
-    NSSetTemporaryDirectory([NSString stringWithCString:folder]);
+void SetTemporaryFolder(const wchar_t* folder) {
+    NSSetTemporaryDirectory([NSString stringWithCharacters:(const unichar*)folder length:wcslen(folder)]);
```
rajsesh-msft: I could have just changed this function to use NSString stringWithUTF8String, but there is no need for back and forth conversions, so just roll with UTF16.

rajsesh-msft:
I reset the approval status.  I changed instances where we were using stringWithCString without encoding for file access.  All of these were potential points of failure, so i replaced them with stringWithUTF8String.

tests/unittests/Accelerate/vImageTest.mm
```diff
@@ -27,390 +27,1017 @@
 
 vImage_Buffer src, dest;
 vImage_Error g_error;
-Pixel_8888 background = {10, 20, 30, 40};
+Pixel_8888 background = { 10, 20, 30, 40 };
```
aballway: nit: undo the formatting changes on this to not muddy history

Frameworks/UIKit/StarboardXaml/ApplicationMain.mm
msft-Jeyaram: C++ style cast?

msft-Jeyaram: C++ style cast

Frameworks/Foundation/NSFileManager.mm
```diff
@@ -267,9 +267,9 @@ - (NSArray*)subpathsAtPath:(NSString*)path {
  @Notes attributes parameter not supported. error parameter not supported.
 */
 - (BOOL)createDirectoryAtURL:(NSURL*)url
- withIntermediateDirectories:(BOOL)createIntermediates
-                  attributes:(NSDictionary*)attrs
-                       error:(NSError**)err {
+    withIntermediateDirectories:(BOOL)createIntermediates
```
msft-Jeyaram: I assume the formats got corrected here?

Frameworks/UIKit/StarboardXaml/ApplicationMain.mm
DHowett-MSFT: Darn. If there was some way to get a HSTRING out of the layer above this one, we could pass it via the NSString HSTRING bridge. That'd save us a whole copy :)

Frameworks/Foundation/NSFileManager.mm
```diff
@@ -844,7 +841,7 @@ - (const char*)fileSystemRepresentationWithPath:(id)pathAddr {
 */
 - (id)stringWithFileSystemRepresentation:(const char*)path length:(NSUInteger)length {
     UNIMPLEMENTED();
-    return [NSString stringWithCString:path length:length];
+    return [NSString stringWithUTF8String:path length:length];
```
rajsesh-msft: Not an actual method.

Frameworks/UIKit/StarboardXaml/ApplicationMain.mm
rajsesh-msft: @DHowett-MSFT yes, and in fact, we do get the value from a HSTRING.  But this being a temp folder, it is going to be copied when added to any file path.

Frameworks/UIKit/StarboardXaml/ApplicationCompositor.cpp
```diff
@@ -39,14 +39,8 @@ void InitializeApp() {
...
-    auto pathData = Windows::Storage::ApplicationData::Current->LocalFolder->Path;
-    IwSetWritableFolder(pathData->Data());
-
-    char writableFolder[2048];
-    size_t outLen;
-    auto tempPathData = Windows::Storage::ApplicationData::Current->TemporaryFolder->Path;
-    wcstombs_s(&outLen, writableFolder, tempPathData->Data(), sizeof(writableFolder) - 1);
-    SetTemporaryFolder(writableFolder);
+    IwSetWritableFolder(Windows::Storage::ApplicationData::Current->LocalFolder->Path->Data());
+    SetTemporaryFolder(Windows::Storage::ApplicationData::Current->TemporaryFolder->Path->Data());
```
bbowman: This is really, really silly. Why do we need to set these at runtime like this. When would this not be what we want. Seems like we should have the getters return this ...

Frameworks/UIKit/StarboardXaml/ApplicationMain.mm
bbowman: Yes it will be copied there anyway. It would still save a copy to not have a useless extra layer here to set it ... 

---
In reply to: [99445013](https://github.com/Microsoft/WinObjC/pull/1916#discussion_r99445013) [](ancestors = 99445013)

Frameworks/UIKit/StarboardXaml/ApplicationCompositor.cpp
bbowman: .

rajsesh-msft: ? the first one is required because starboard is removed from the application layer.  The second is rather bizarre, but NSPathUtilties depends on it.  Sure, NSPathUtiltities can do this instead.

bbowman: They are both silly. It doesn't make sense that the application compositor sets any sort of file system stuff. Starboard should know how to do this itself. That would be more proper encapsulation ... 

---
In reply to: [99649338](https://github.com/Microsoft/WinObjC/pull/1916#discussion_r99649338) [](ancestors = 99649338)

bbowman:
:clock1:

Frameworks/UIKit/StarboardXaml/ApplicationCompositor.cpp
rajsesh-msft: @bbowman any re-write on this should be considered through #1904. what we have is not ideal.

jaredhms:
is there any way to prevent usage of stringWithCString  from creeping back in?

---
In reply to: [277388030](https://github.com/Microsoft/WinObjC/pull/1916#issuecomment-277388030) [](ancestors = 277388030)

jaredhms:
:shipit:

PR #1917, opened 2017-02-04
Fix bug in sizeWithFont

aballway:
NSString+UIKitAdditions sizeWithFont:constrainedToSize: should return the height that the text requires rather than the height of the text that can actually fit in the region that CTFramesetterSuggestFrameSizeWithConstraints returns.

### DIFF
--- a/Frameworks/UIKit/NSString+UIKitAdditions.mm
+++ b/Frameworks/UIKit/NSString+UIKitAdditions.mm
@@ -259,9 +259,8 @@ static NSDictionary* _getDefaultUITextAttributes() {
         size.width = std::numeric_limits<CGFloat>::max();
     }
 
-    if (size.height == 0.0) {
-        size.height = std::numeric_limits<CGFloat>::max();
-    }
+    // Return the height that the text actually requires rather than the constrained size
+    size.height = std::numeric_limits<CGFloat>::max();
 
     return CTFramesetterSuggestFrameSizeWithConstraints(framesetter.get(), CFRangeMake(0, self.length), nullptr, size, nullptr);
 }
--- a/Frameworks/UIKit/NSString+UIKitAdditions.mm
+++ b/Frameworks/UIKit/NSString+UIKitAdditions.mm
@@ -244,6 +244,11 @@ static NSDictionary* _getDefaultUITextAttributes() {
 // Returns the bounding box size this string would occupy when drawn as specified
 // All sizeWith... functions in this file funnel to this
 - (CGSize)_sizeWithAttributes:(NSDictionary<NSString*, id>*)attributes constrainedToSize:(CGSize)size {
+    UIFont* font = attributes[NSFontAttributeName];
+    if (font == nil) {
+        return CGSizeZero;
+    }
+
     if ([attributes objectForKey:NSParagraphStyleAttributeName]) {
         NSMutableDictionary* copied = [NSMutableDictionary dictionaryWithDictionary:attributes];
         woc::unique_cf<CTParagraphStyleRef>
@@ -266,7 +271,6 @@ static NSDictionary* _getDefaultUITextAttributes() {
     CGSize ret = CTFramesetterSuggestFrameSizeWithConstraints(framesetter.get(), CFRangeMake(0, self.length), nullptr, size, nullptr);
 
     // If the constrained height is less than a line height sizeWithFont will increase the returned size to fit at least one line
-    UIFont* font = attributes[NSFontAttributeName];
     CGFloat lineHeight = [font ascender] - [font descender];
     ret.height = std::max(ret.height, lineHeight);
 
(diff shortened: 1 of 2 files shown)


Frameworks/UIKit/NSString+UIKitAdditions.mm
```diff
@@ -259,9 +259,8 @@ - (CGSize)_sizeWithAttributes:(NSDictionary<NSString*, id>*)attributes constrain
         size.width = std::numeric_limits<CGFloat>::max();
     }
 
-    if (size.height == 0.0) {
```
yiyang-msft: >if (size.height == 0.0) { [](start = 4, length = 25)

no wonder with this, it works. :)

yiyang-msft:
:shipit:

Frameworks/UIKit/NSString+UIKitAdditions.mm
```diff
@@ -259,9 +259,8 @@ - (CGSize)_sizeWithAttributes:(NSDictionary<NSString*, id>*)attributes constrain
...
-        size.height = std::numeric_limits<CGFloat>::max();
-    }
+    // Return the height that the text actually requires rather than the constrained size
+    size.height = std::numeric_limits<CGFloat>::max();
```
ms-jihua: so wait the expectation of the constrainedToSize: contract is that the height passed in doesn't matter at all? that seems... wrong...

jaredhms: that does seem a bit odd

---
In reply to: [99448184](https://github.com/Microsoft/WinObjC/pull/1917#discussion_r99448184) [](ancestors = 99448184)

aballway: @yiyang-msft and I reproduced this on the reference platform, but it's likely that this only applies to a case where we are only drawing one line but unable to fit it.  I'm going to be doing more testing to see where height is respected and where it is not, and if this also applies to CTFramesetterSuggestFrameSizeWithConstraints.

tests/unittests/UIKit/NSString+UIKitAdditionsTests.mm
```diff
@@ -17,9 +17,21 @@
 #import <TestFramework.h>
 #import <UIKit/UIKit.h>
 
+static constexpr double c_errorDelta = .0001;
```
rajsesh-msft: nit: use numEquals

jaredhms:
:shipit:

Frameworks/UIKit/NSString+UIKitAdditions.mm
yiyang-msft: @alballw@microsoft.com , make sense.

---
In reply to: [99671953](https://github.com/Microsoft/WinObjC/pull/1917#discussion_r99671953) [](ancestors = 99671953)

yiyang-msft: also,please verified with MM. check with the introduction label on first screen.

---
In reply to: [99942712](https://github.com/Microsoft/WinObjC/pull/1917#discussion_r99942712) [](ancestors = 99942712,99671953)

aballway: With this change the introduction label in Islandwood house is being scaled properly

---
In reply to: [99943020](https://github.com/Microsoft/WinObjC/pull/1917#discussion_r99943020) [](ancestors = 99943020,99942712,99671953)

Frameworks/UIKit/NSString+UIKitAdditions.mm
```diff
@@ -263,7 +263,14 @@ - (CGSize)_sizeWithAttributes:(NSDictionary<NSString*, id>*)attributes constrain
         size.height = std::numeric_limits<CGFloat>::max();
     }
 
-    return CTFramesetterSuggestFrameSizeWithConstraints(framesetter.get(), CFRangeMake(0, self.length), nullptr, size, nullptr);
+    CGSize ret = CTFramesetterSuggestFrameSizeWithConstraints(framesetter.get(), CFRangeMake(0, self.length), nullptr, size, nullptr);
+
+    // If the constrained height is less than a line height sizeWithFont will increase the returned size to fit at least one line
+    UIFont* font = attributes[NSFontAttributeName];
```
DHowett-MSFT: What if there's no font?

jaredhms: i figured you'd just end up with a 0 size, so you'd get ret's height.

DHowett-MSFT: If the compiler is working as intended, yes :) We had issues with returning floats from nil receivers before.

aballway: Looks like reference platform just returns CGSizeZero, will update to reflect that

PR #1920, opened 2017-02-06
This change includes the addition of a few new features to UIButton.

### LINKED ISSUE #1442: Sytem UIButton does not update when being pressed
Phone: Lumia 640 device (14393 build)
Master commit: https://github.com/Microsoft/WinObjC/commit/2e0afafbf36d23b1cee7bd09f1dc8cf230c5c8a1

Build/run WoCCatalog in Debug.
Go to ButtonsViewController.
Press Simple Button.

Expected: Simple Button makes some sort of visible change when pressed.
Actual: No visible change when pressed.

Image Button does change when pressed.

### LINKED ISSUE #1671: need to scrub the usage of applyTemplate in UIButton
we have following pattern in UIButton. In order to retrieve temlateImage/Text, we invoke xamlButton applyTemplate and hope it force load all templated parts. But this pattern is not guaranteed to work until layout happen. Currently UIButton is working but might as well be lucky. We need come up with a solid pattern to make this approach more determinstic.

On the other hand hand, a separate pattern is being used in UITextField, which utilizes onLoadedEvent. But this requires caching value which isn't idea.

- (void)_initUIButton {
    _xamlButton = rt_dynamic_cast<WXCButton>([self xamlElement]);
... 
   // Force-load the template, and get the TextBlock and Image for use in our proxies.
    [_xamlButton applyTemplate];
...
    WXCImage* templateImage = rt_dynamic_cast([WXCImage class], [_xamlButton getTemplateChild:@"buttonImage"]);
    WXCTextBlock* templateText = rt_dynamic_cast([WXCTextBlock class], [_xamlButton getTemplateChild:@"buttonText"]);

...
}

### LINKED ISSUE #1885: [UIButton] Scrub Annotations
Review Annotations and update to match current set of annotations.

### LINKED ISSUE #1921: Support UIButton buttonWithType
Support System button type in addition to the Custom button type already supported.  Other button types functionality can be easily duplicated with a custom button.

### LINKED ISSUE #1922: UIButton should have different visual for disabled state

jaredhms:
Some functionality deviates from the reference platform, but we feel that it aligns more closely with the default look and feel of Xaml buttons on Windows because buttons 'look' pressed in more cases with these changes, and there are straightforward ways to opt out of this behavior if it's not desired.

**adjustsImageWhenDisabled/adjustsImageWhenHighlighted**:
We're adding WinObjC-specific support for adjustsImageWhenDisabled and adjustsImageWhenHighlighted.  On the reference platform, this tinting/darkening behavior is ONLY applied to the button's background an foreground images, but on our platform, the tintening/darkening is applied to the entire button's bounds (including its background color), but NOT to its label text.  The primary justification for this approach is that there's not a straighforward way to tint Xaml Images on RS1, yet this approach also helps us implement simple SystemButton support and it makes UIButtons appear a bit more 'windows-like' in many cases, since they show more distinct pressed behavior.  The behavior is as follows; if adjustsImageWhenDisabled is set to YES, and if no UIControlStateDisabled properties are set on the button, then the entire button contents (aside from its label) receives a lightening effect.  Similar behavior is implemented for adjustsImageWhenHighlighted.  The way to opt-out of this behavior is to either set a single UIControlStateDisabled/UIControlStateHighlighted property, or to set adjustsImageWhenDisabled/adjustsImageWhenHighlighted to NO.

**SystemButton (aka buttonWithType)**:
We're adding support for UIButtonTypeSystem, which behaves as follows:

	- Default background is white.
	- Receives a default text color of (blue/green) [UIColor colorWithRed:0.0f green:0.47843137f blue:1.0f alpha:1.0f] which applies to all states (unless set otherwise)
	- Default to UIColor lightTextColor for disabled titleLabelColor

**SystemButton - Deviations from the Reference Platform**:

	- When selected, the entire label background turns the default text color (blue/green) [UIColor colorWithRed:0.0f green:0.47843137f blue:1.0f alpha:1.0f], and the text color is changed to white.  This seems like a corner case that we're not implementing until we need to for a customer.
	- When highlighted, the titleLabel contents (its text and any sublayer content) are tinted to a darker/greyed out version of the normal state.
		- Instead of this behavior, we're piggy-backing off of our adjustsImageWhenHighlighted implementation to avoid needing to do this extra work.  The button *looks* highlighted, so we don't have to change the text color.
	- On the reference platform, when disabled, text is grey by default, UNLESS any custom normal or disabled text colors are set.  On our platform, we DEFAULT to disabled to grey text and then you can change it, but changing the normal color won't affect the default color (for the sake of simplicity).

Scrubbed Annotations:
All UIButton annotations have been scrubbed and updated with 'NotInPlan' status as needed.

Fixes #1442
Fixes #1885
Fixes #1671
Fixes #1671
Fixes #1921 
Fixes #1922


Frameworks/UIKit/UIButton.mm
```diff
@@ -71,16 +69,20 @@ @implementation UIButton {
...
+    UIButtonType _buttonType;
+
+    ComPtr<IInspectable> _inspectableAdjustsWhenDisabledBrush;
+    ComPtr<IInspectable> _inspectableAdjustsWhenHighlightedBrush;
 }
 
 /**
  @Status Caveat
- @Notes May not be fully implemented
+ @Notes Not all properties are supported.
```
jaredhms: @tadam-msft, do we have a work item tracking the additiona of all missing UIButton xib2nib conversions?

Frameworks/UIKit/UIButton.mm
```diff
@@ -456,42 +456,83 @@ - (BOOL)isEnabled {
...
 /**
- @Status Interoperable
- @Notes The xaml element may be modified directly and we could still return stale values.
+@Status Interoperable
+@Notes The xaml element may be modified directly and we could return stale values.
 */
-- (UIImage*)backgroundImageForState:(UIControlState)state {
-    return _states[state].backgroundImage;
+-(UIImage*)imageForState:(UIControlState)state {
+    const auto& statePair = _states.find(state);
```
jaredhms: maybe could use a comment.  a subtle change that avoids inserting empty values in to the map (which wastes space) when retrieving empty values

Frameworks/UIKit/UIButton.mm
```diff
@@ -784,17 +872,18 @@ - (UIEdgeInsets)contentEdgeInsets {
...
 }
 
 /**
- @Status Stub
+ @Status NotInPlan
+ @Notes Xaml doesn't have built in support for showing 'glowing' touch feedback.
 */
 - (BOOL)showsTouchWhenHighlighted {
-    UNIMPLEMENTED();
+    UNIMPLEMENTED_WITH_MSG("Xaml doesn't have built in support for showing 'glowing' touch feedback.");
```
DHowett-MSFT: propagate this message to the setter as well; people will be calling the setter _far_ more often than the getter.

Frameworks/UIKit/UIControl.mm
```diff
@@ -221,7 +221,7 @@ - (NSSet*)allTargets {
 */
 - (void)setEnabled:(BOOL)enabled {
     // only update the value and then relayout if value actually changed
-    if ((_curState & UIControlStateDisabled) != enabled) {
+    if ((_curState & UIControlStateDisabled) != !enabled) {
```
DHowett-MSFT: This doesn't seem like a type-safe comparison. `_curState & x` evaluates to an int, not a boolean. There's probably a cleaner way to structure this check.

Frameworks/UIKit/UIView.mm
```diff
@@ -1405,10 +1395,8 @@ - (void)setFrame:(CGRect)frame {
                      frame.size.height);
     }
 
-    CGRect startFrame = frame;
-
-    if (frame.origin.x == doRound(curFrame.origin.x) && frame.origin.y == doRound(curFrame.origin.y) &&
-        frame.size.width == doRound(curFrame.size.width) && frame.size.height == doRound(curFrame.size.height)) {
+    // Don't bother doing work if there's no work to be done
+    if (memcmp(&frame, &curFrame, sizeof(CGRect)) == 0) {
```
DHowett-MSFT: This only works if/when the floats in CGRect are packed back-to-back. While that's true now, it might not be universally true. Worth being concerned about?

Frameworks/UIKit/UIView.mm
```diff
@@ -1505,6 +1494,7 @@ - (void)setOrigin:(CGPoint)origin {
     curFrame = [layer frame];
     curFrame.origin = origin;
 
+    // TODO: Should we be rounding to the nearest pixel like we do in setFrame?
```
DHowett-MSFT: Do we need an issue tracking this TODO?

Frameworks/include/UIViewInternal.h
```diff
@@ -29,10 +29,33 @@
 #import "UWP/InteropBase.h"
 #import "UWP/WindowsUIXamlControls.h"
 
+#import "UIViewInternal.h"
```
DHowett-MSFT: Is this header including itself?
If not, why does it have the same name?

jaredhms: thanks; unintentional modification.

---
In reply to: [99915207](https://github.com/Microsoft/WinObjC/pull/1920#discussion_r99915207) [](ancestors = 99915207)

Frameworks/UIKit/UIButton.mm
jaredhms: great point; an oversight for sure.

---
In reply to: [99914391](https://github.com/Microsoft/WinObjC/pull/1920#discussion_r99914391) [](ancestors = 99914391)

jaredhms: looks like this is the only one i missed

---
In reply to: [99922832](https://github.com/Microsoft/WinObjC/pull/1920#discussion_r99922832) [](ancestors = 99922832,99914391)

Frameworks/UIKit/UIView.mm
jaredhms: probably not worth worrying about right now; this is existing code moved down from above to avoid two checks.  i suppose we could just keep the distinct field checks instead, but the memcmp seems fine to me.

---
In reply to: [99915102](https://github.com/Microsoft/WinObjC/pull/1920#discussion_r99915102) [](ancestors = 99915102)

Frameworks/UIKit/UIControl.mm
jaredhms: this just follows precedence from elsewhere.  it's unclear to me why this isn't safe?

---
In reply to: [99914946](https://github.com/Microsoft/WinObjC/pull/1920#discussion_r99914946) [](ancestors = 99914946)

Frameworks/include/UIViewInternal.h
```diff
@@ -33,6 +33,27 @@
 @class WXFrameworkElement;
 @class WUXIPointerRoutedEventArgs;
 
+// Round subpixel values to be able to perform per-pixel UI placement/calculations
+inline float doPixelRound(float f) {
```
oliversa-msft: >doPixelRound [](start = 13, length = 12)

Probably explains why I see Canvas.Left values of 101 when I expect 100. Experienced this during inset testing

Frameworks/UIKit/UIButton.mm
```diff
@@ -39,28 +39,26 @@
...
 
 struct ButtonState {
+    // Returns whether or not the state contains any currently-set values
+    bool IsEmpty() const {
+        return !title && !textColor && !image && !backgroundImage;
+    }
+
     StrongId<UIImage> image;
     StrongId<UIImage> backgroundImage;
     StrongId<UIColor> textColor;
```
oliversa-msft: >textColor [](start = 22, length = 9)

nit: titleColor would have been in line with the iOS fields.

Frameworks/include/UIViewInternal.h
jaredhms: maybe, plus you are looking at a label proxy right?  the layout code is different for the proxies.

---
In reply to: [100634412](https://github.com/Microsoft/WinObjC/pull/1920#discussion_r100634412) [](ancestors = 100634412)

oliversa-msft: I may have seen it on images as well - not sure but definitively on label.

---
In reply to: [100638812](https://github.com/Microsoft/WinObjC/pull/1920#discussion_r100638812) [](ancestors = 100638812,100634412)

jaredhms: the image proxy still exists, so this probably won't help them

---
In reply to: [100639129](https://github.com/Microsoft/WinObjC/pull/1920#discussion_r100639129) [](ancestors = 100639129,100638812,100634412)

Frameworks/UIKit/UIControl.mm
yiyang-msft: >!enabled [](start = 48, length = 8)

sutble

Frameworks/UIKit/UIButtonProxies.h
```diff
@@ -35,16 +35,6 @@
...
 
-// NOTE: current _UILabel_Proxy is merely a bridge to xaml TextBlock, which is only
-// used to in a very confined scope - that is when accessing some properties
-// of UIButton.titleLabel.  So it has very limited usage right now.
-// However, we will replace UILabel the full implemention of _UILabel in the future
-@interface _UILabel_Proxy : _UIView_Proxy
-
-- (instancetype)initWithXamlElement:(WXFrameworkElement*)xamlElement font:(UIFont*)font;
-
-@end
```
yiyang-msft: sweat!

Frameworks/UIKit/UIButton.mm
```diff
@@ -167,19 +169,29 @@ - (void)_initUIButton {
...
+    self.adjustsImageWhenDisabled = YES;
+    self.adjustsImageWhenHighlighted = YES;
+
     // Force-load the template, and get the TextBlock and Image for use in our proxies.
     [_xamlButton applyTemplate];
-
+    [_xamlButton updateLayout];
+
+    // Create our child UILabel; its frame will be updated in layoutSubviews
+    // TODO: Ideally we'd grab this directly from the Xaml, but we're not able to launch some apps when doing so due to a
```
yiyang-msft: >directly from the Xaml [](start = 36, length = 22)

so close!

jaredhms: yeah, bummer. it actually worked fine with WoCCatalog and XamlCatalog; must have something to do with our import flow.  

---
In reply to: [100646740](https://github.com/Microsoft/WinObjC/pull/1920#discussion_r100646740) [](ancestors = 100646740)

Frameworks/UIKit/UIButton.mm
```diff
@@ -503,44 +545,34 @@ - (UIImage*)currentBackgroundImage {
...
-        _states[state].inspectableTitle = nullptr;
-    }
-
-    // Update the Xaml elements immediately, so the proxies reflect reality
-    XamlButtonApplyVisuals([_xamlButton comObj],
-                           _currentInspectableTitle(self),
-                           _currentInspectableImage(self),
-                           _currentInspectableTitleColor(self));
+    // Update our title label immediately
+    self.titleLabel.text = self.currentTitle;
```
yiyang-msft: >self.titleLabel.text = self.currentTitle; [](start = 4, length = 41)

should we check if current state is the title state before updating this?

Frameworks/UIKit/UIButton.mm
```diff
@@ -549,65 +581,69 @@ - (UIImage*)imageForState:(UIControlState)state {
...
-        _states[state].inspectableTitleColor = nullptr;
-    }
-
-    // Update the Xaml elements immediately, so the proxies reflect reality
-    XamlButtonApplyVisuals([_xamlButton comObj],
-                           _currentInspectableTitle(self),
-                           _currentInspectableImage(self),
-                           _currentInspectableTitleColor(self));
+    // Update our title label color
+    self.titleLabel.textColor = self.currentTitleColor;
```
yiyang-msft: >self.titleLabel.textColo [](start = 4, length = 24)

similar question as title state.

Frameworks/UIKit/UIButton.mm
```diff
@@ -891,22 +946,18 @@ - (UIImageView*)imageView {
...
  @Status Interoperable
 */
 - (CGSize)intrinsicContentSize {
     CGSize ret = CGSizeZero;
 
     UIImage* img = self.currentImage;
     UIEdgeInsets contentInsets = self.contentEdgeInsets;
-    CGSize textSize = [[self currentTitle] sizeWithFont:self.font];
+
+    // TODO: Should we be asking our label for its intrinsicContentSize?
```
yiyang-msft: >// TODO: Should we be asking our label for its intrinsicContentSize? [](start = 4, length = 68)

I guess not. 

intrinsicContentSize

The natural size for the receiving view, considering only properties of the view itself.
[https://developer.apple.com/reference/uikit/uiview?language=objc](https://developer.apple.com/reference/uikit/uiview?language=objc)

Frameworks/UIKit/UIButton.mm
jaredhms: What do you mean?  I don't think we should do a string comparison if that's what you're asking?

---
In reply to: [100649772](https://github.com/Microsoft/WinObjC/pull/1920#discussion_r100649772) [](ancestors = 100649772)

Frameworks/UIKit/UIButton.mm
jaredhms: That makes it sound as though we shouldn't be taking the text size into account at all then.  I think in this case 'the view itself' may mean the button and its internal content/views.

---
In reply to: [100653027](https://github.com/Microsoft/WinObjC/pull/1920#discussion_r100653027) [](ancestors = 100653027)

Frameworks/UIKit/UIButton.mm
yiyang-msft: sorry. I mean, when I setTitle for dsiabled state, but button current state is normal. and i still end up with updating the title for normal state?

---
In reply to: [100653158](https://github.com/Microsoft/WinObjC/pull/1920#discussion_r100653158) [](ancestors = 100653158,100649772)

Frameworks/UIKit/UIButton.mm
yiyang-msft: wish the doc is more clear, looks good candidate for an experiment later.

---
In reply to: [100653270](https://github.com/Microsoft/WinObjC/pull/1920#discussion_r100653270) [](ancestors = 100653270,100653027)

Frameworks/UIKit/UIButton.mm
jaredhms: Ah - self.currentTitle gives you the text for the current state.  

---
In reply to: [100653633](https://github.com/Microsoft/WinObjC/pull/1920#discussion_r100653633) [](ancestors = 100653633,100653158,100649772)

yiyang-msft: got you, setting value on other states might impact current state title, you need re-evaluate.

---
In reply to: [100653762](https://github.com/Microsoft/WinObjC/pull/1920#discussion_r100653762) [](ancestors = 100653762,100653633,100653158,100649772)

Frameworks/UIKit/UIButton.mm
yiyang-msft: Oh,  i think i missunerstand your point earlier.

It is a little weird why we are not asking TitleLabel for its size, instead, we are using the sizeWithFont with that string. This does not make sense. some of properties on UIButton such as linkbreakMode is just a pass through to UILbael.

---
In reply to: [100653705](https://github.com/Microsoft/WinObjC/pull/1920#discussion_r100653705) [](ancestors = 100653705,100653270,100653027)

yiyang-msft:
:shipit:

Frameworks/UIKit/UIButton.mm
yiyang-msft: and i guess we have not run into a button which should display two line of text yet. :)

---
In reply to: [100654429](https://github.com/Microsoft/WinObjC/pull/1920#discussion_r100654429) [](ancestors = 100654429,100653705,100653270,100653027)

PR #1923, opened 2017-02-07
Improve performance in __CF_IsBridgedObject/__CF_IsCFObject

### LINKED ISSUE #1490: CFRetain and CFRelease could be expensive for bridged classes
_CFLocaleCopyCurrentGuts for example spends most of its time in CFRetain because CFLocale is bridged to NSLocale.  @DHowett-MSFT has partial fix for this [here](https://github.com/DHowett-MSFT/WinObjC/commits/201609-nscftype).

ms-jihua:
- __CF_IsCFObject() calls __CF_IsBridgedObject()
 - Previously, __CF_IsBridgedObject() would do an O(n^2) comparison against the entire runtime class table.
   (O(n) traversal of table * O(n) check of inheritance tree)
 - These two functions are used in CF memory management functions, such as CFRetain/Release,
   and due to their prevalence, took up substantial portions of CPU time.
 - Changed bridged classes to adopt an empty 'bridged object' protocol when they are bridged/added to the class table.
   This protocol can be quickly checked against, and obviates the need to traverse the table.
 - Removed some redundant _cfisa checks in _CF_IsCFObject

Fixes #1490

### DIFF
--- a/Frameworks/CoreFoundation/Base.subproj/CFInternal.h
+++ b/Frameworks/CoreFoundation/Base.subproj/CFInternal.h
@@ -712,7 +712,7 @@ CF_INLINE bool __CF_IsBridgedObject(CFTypeRef obj) {
     // However, this would not grant access to anything that could not be accessed by more conventional means
     // (ie: overriding an existing CF/Foundation class)
     // and as such would not constitute an escalation of privilege.
-    return [object conformsToProtocol:__CFRuntimeGetBridgeProtocol()];
+    return class_conformsToProtocol((Class)(object->_cfisa), __CFRuntimeGetBridgeProtocol());
 }
 
 
--- a/Frameworks/CoreFoundation/Base.subproj/CFInternal.h
+++ b/Frameworks/CoreFoundation/Base.subproj/CFInternal.h
@@ -369,6 +369,17 @@ CF_PRIVATE Boolean __CFProcessIsRestricted();
 extern "C" Class _OBJC_CLASS__NSCFNumber;
 extern "C" Class _OBJC_CLASS__NSCFBoolean;
 
+// WINOBJC: Returns an empty protocol that is assigned only to bridged classes, so that bridged objects can be distinguished quickly
+static Protocol* __CFRuntimeGetBridgeProtocol() {
+    static Protocol* s_bridgeProtocol = []() {        
+        Protocol* ret = objc_allocateProtocol("_NSCFBridgedType");
+        objc_registerProtocol(ret);
+        return ret;
+    }();
+
+    return s_bridgeProtocol;
+}
+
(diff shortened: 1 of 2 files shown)


ms-jihua:
Benchmarks: 
image

Frameworks/CoreFoundation/Base.subproj/CFInternal.h
```diff
@@ -696,28 +706,23 @@ CF_INLINE bool __CF_IsBridgedObject(CFTypeRef obj) {
...
-    }
-
-    return false;
+    // Check if the object has the 'bridged object' protocol
+    // NOTE: because our class_conformsToProtocol() checks based only on name, 
+    // it is possible to spoof a bridged object by stating an object conforms to a protocol with the same name.
+    // However, this would not grant access to anything that could not be accessed by more conventional means
+    // (ie: overriding an existing CF/Foundation class)
+    // and as such would not constitute an escalation of privilege.
+    return [object conformsToProtocol:__CFRuntimeGetBridgeProtocol()];
```
DHowett-MSFT: Prefer `class_conformsToProtocol(object->_cfisa, ...)` -- that will remove Objective-C from purely C++ CFInternal.h consumers.

DHowett-MSFT: (this header is shared externally -- CoreGraphics, CoreText, etc. use it)

Frameworks/CoreFoundation/Base.subproj/CFInternal.h
```diff
@@ -696,28 +706,23 @@ CF_INLINE bool __CF_IsBridgedObject(CFTypeRef obj) {
...
-    for (unsigned int i = 0; i < __CFRuntimeClassTableSize; i++) {
-        if ((__CFRuntimeObjCClassTable[i] != 0) && 
-            ((object->_cfisa == __CFRuntimeObjCClassTable[i]) ||
-            ([object isKindOfClass:(Class)(__CFRuntimeObjCClassTable[i])]))) {
-            return true;
-        }
-    }
-
-    return false;
+    // Check if the object has the 'bridged object' protocol
```
aballway: nit: should have WINOBJC: here

ms-jihua: This is already a WINOBJC: function, so not necessary

ms-jihua:
Benchmarks after handling NSCFType properly:

image

ms-jihua:
Before/after: 
image

ms-jihua:
Note that for these changes to propagate, it may be necessary to do a rebuild - our incremental build seems to be bad at picking up changes to CF_INLINE functions, though nothing seems to break.

PR #1933, opened 2017-02-08
Add support for all CGTextDrawingModes

### LINKED ISSUE #1935: Text drawn through CTFrameDraw will shadow and alpha compose itself
Because CTFrameDraw (eventually) calls `__CGContext::DrawGlyphRun` once per line, it invokes the CG shadowing and composition stack multiple times.
If a shadow is applied, this will result in later lines shadowing earlier lines. The same applies to rendering non-opaque text: overlapping regions will be composed instead of merged into a single draw region.

aballway:
Several changes have been added to address #1933 and #1935 

CGContextDrawGlyphRun has been changed to _CGContextDrawGlyphRuns and __CGContext::DrawGlyphRun -> __CGContext::DrawGlyphsRuns, which takes a dummy struct GlyphRunData, which holds the DWRITE_GLYPH_RUN, relative position of the glyph run (to the text position), and attributes of the glyph run (fill color, etc.).

Text drawing methods such as CTFrameDraw now accumulate an array of GlyphRunDatas and pass them to _CGContextDrawGlyphRuns, which in turn calls context->DrawGlyphRuns.

To batch all run drawing for clipping/stroking into the same geometry, the runs have to be drawn with the same transformation, so position has to be given (hence relative position in GlyphRunData).  The position is transformed by the inverse of the text transform, so when the glyph is drawn and transformed by the real text transform it will have the correct position.  This has the added benefit of allowing us to draw a glyph run which has a transformed text matrix all together, rather than glyph-by-glyph as we did previously.

Because text drawing is being batched, attributes which are handled at draw time on a run-per-run basis such as fill color need to be done in DrawGlyphRuns.  The attribute names are defined in CoreText, though all of them are just CoreGraphics attributes, so local values have been added to CGContext.mm rather than making an internal version and defining the CT/UIKit versions as CG.

Images have been updates since changing this seems to have enabled antialiasing to some degree on several tests. 

Fixes #1933 
Fixes #1935


Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1898,13 +1908,22 @@ void CGContextSetPatternPhase(CGContextRef context, CGSize phase) {
 #pragma endregion
 
 #pragma region Drawing Operations - Text
+
+// Attributes are needed in DrawGlyphRuns but names are defined in CoreText
+const CFStringRef _kCGForegroundColorAttributeName = static_cast<CFStringRef>(@"NSForegroundColor");
```
ms-jihua: why not CFSTR()?

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1898,13 +1908,22 @@ void CGContextSetPatternPhase(CGContextRef context, CGSize phase) {
 #pragma endregion
 
 #pragma region Drawing Operations - Text
+
+// Attributes are needed in DrawGlyphRuns but names are defined in CoreText
```
ms-jihua: worried about these getting out of sync, would possibly prefer defining here and letting CoreText = these

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1918,66 +1937,134 @@ void CGContextSetPatternPhase(CGContextRef context, CGSize phase) {
     textTransform.ty = std::round(textTransform.ty);
 
     CGAffineTransform finalTextTransform = CGAffineTransformConcat(textTransform, deviceTransform);
-    if ((fabs(finalTextTransform.a * glyphRun->fontEmSize) <= c_glyphThreshold &&
-         fabs(finalTextTransform.b * glyphRun->fontEmSize) <= c_glyphThreshold) ||
-        (fabs(finalTextTransform.d * glyphRun->fontEmSize) <= c_glyphThreshold &&
-         fabs(finalTextTransform.c * glyphRun->fontEmSize) <= c_glyphThreshold)) {
-        TraceWarning(TAG, L"Glyphs too small to be rendered");
+    for (size_t i = 0; i < runsCount; ++i) {
```
ms-jihua: this fails drawing for every run if even a single run is too small to be drawn, that seems wrong?

Frameworks/CoreGraphics/CGContext.mm
aballway: Just used what was in CTStringAttributes.mm, is one preferred over the other?

Frameworks/CoreGraphics/CGContext.mm
aballway: I'm not so sure.  This is only a problem on our platform, but idk if we'd just want to draw the rest of a line if part is so small that it would be causing a crash.

include/CoreGraphics/CGContext.h
```diff
@@ -85,16 +85,19 @@ typedef CF_ENUM(CFIndex, CGBlendMode) {
...
-    kCGTextStroke,
-    kCGTextFillStroke,
-    kCGTextInvisible,
-    kCGTextFillClip,
-    kCGTextStrokeClip,
-    kCGTextFillStrokeClip,
-    kCGTextClip
+// clang-format off
+typedef CF_OPTIONS(CFIndex, CGTextDrawingMode) {
+//  CG Drawing Bitfield     |FILL  |   |STROKE|   |CLIP  |
```
ms-jihua: nice

Frameworks/CoreGraphics/CGContext.mm
ms-jihua: runs within a line don't necessarily have the same font size, right?

---
In reply to: [100872254](https://github.com/Microsoft/WinObjC/pull/1933#discussion_r100872254) [](ancestors = 100872254)

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1918,66 +1937,134 @@ void CGContextSetPatternPhase(CGContextRef context, CGSize phase) {
...
+                    DWRITE_GLYPH_OFFSET{ transformedPosition.x + glyphRuns[i].run->glyphOffsets[j].advanceOffset,
+                                         std::round(transformedPosition.y + glyphRuns[i].run->glyphOffsets[j].ascenderOffset) };
+
+                // Translate position of next glyph by current glyph's advance
+                runningPosition.x += glyphRuns[i].run->glyphAdvances[j];
+            }
+        } else {
+            // CG text drawing mathods use the text matrix to transform about the origin, not about each glyph's origin
+            // We can simply use the positions directly from the glyph run
+            std::copy(glyphRuns[i].run->glyphOffsets,
```
ms-jihua: let's not do an expensive copy here, especially for what we can expect to be the 'normal' case

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1918,66 +1937,134 @@ void CGContextSetPatternPhase(CGContextRef context, CGSize phase) {
...
-
-                         // Uses glyphAdvances to move each glyph
-                         runningGlobalTransform = CGAffineTransformTranslate(runningGlobalTransform, glyphRun->glyphAdvances[i], 0);
-                     }
-                     return S_OK;
-                 }));
+            Factory()->CreateTransformedGeometry(pathGeometry.Get(), __CGAffineTransformToD2D_F(finalTextTransform), &transformedGeometry));
+    }
+
+    HRESULT ret = Draw(_kCGCoordinateModeDeviceSpace, nullptr, [&](CGContextRef context, ID2D1DeviceContext* deviceContext) {
```
ms-jihua: >[&] [](start = 63, length = 3)

CC: explicitly write out all your captures.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1918,66 +1937,134 @@ void CGContextSetPatternPhase(CGContextRef context, CGSize phase) {
...
+            // First glyph's origin is at the given relative position for the glyph run
+            CGPoint runningPosition{ glyphRuns[i].relativePosition.x, std::round(glyphRuns[i].relativePosition.y) };
+            DWRITE_GLYPH_OFFSET previousOffset{ 0, 0 };
+            for (size_t j = 0; j < glyphRuns[i].run->glyphCount; ++j) {
+                // Invert position by text transformation
+                CGPoint transformedPosition = CGPointApplyAffineTransform(runningPosition, invertedTextTransformation);
+
+                // Set current glyph's position to be the actual position transformed by the inverted text position
+                // So when the space is transformed by the text position it will be drawn in the correct real position
+                positionsVectors[i][j] =
```
ms-jihua: probably worth assigning names to some things to lessen the number of calls to getters

Frameworks/CoreGraphics/CGContext.mm
ms-jihua: ugh I guess no one else in this file did it either, so I don't blame you, but it'd be nice if you did.

---
In reply to: [100875768](https://github.com/Microsoft/WinObjC/pull/1933#discussion_r100875768) [](ancestors = 100875768)

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1918,66 +1937,134 @@ void CGContextSetPatternPhase(CGContextRef context, CGSize phase) {
...
+            // CG text drawing mathods use the text matrix to transform about the origin, not about each glyph's origin
+            // We can simply use the positions directly from the glyph run
+            std::copy(glyphRuns[i].run->glyphOffsets,
+                      glyphRuns[i].run->glyphOffsets + glyphRuns[i].run->glyphCount,
+                      positionsVectors[i].begin());
+        }
     }
 
-    auto& state = CurrentGState();
+    // DWRITE will misbehave if we don't give it a pointer to zero advances so just create one vector for all advances
```
ms-jihua: how did this work before when we were passing nullptr?

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1918,66 +1937,134 @@ void CGContextSetPatternPhase(CGContextRef context, CGSize phase) {
...
+                if (glyphRuns[i].attributes) {
+                    CGColorRef fontColor = (CGColorRef)CFDictionaryGetValue(glyphRuns[i].attributes, _kCGForegroundColorAttributeName);
+                    if (!fontColor) {
+                        CFBooleanRef useContextColor =
+                            (CFBooleanRef)CFDictionaryGetValue(glyphRuns[i].attributes, _kCGForegroundColorFromContextAttributeName);
+                        if (!useContextColor || !CFBooleanGetValue(useContextColor)) {
+                            // Neither given a color nor use current context color, so set the fill color to black
+                            CGContextSetRGBFillColor(context, 0, 0, 0, 1);
+                        }
+                    } else {
```
ms-jihua: nit: think success case first makes more sense here

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1918,66 +1937,134 @@ void CGContextSetPatternPhase(CGContextRef context, CGSize phase) {
...
+                        CGContextSetFillColorWithColor(context, fontColor);
+                    }
+                }
+
+                DWRITE_GLYPH_RUN transformedGlyphRun{ glyphRuns[i].run->fontFace,   glyphRuns[i].run->fontEmSize,
```
ms-jihua: it's not the worst thing but i dislike having to construct this in the ordinary case where we don't transform by glyph.

Frameworks/CoreGraphics/CGContext.mm
ms-jihua: why not just RETURN_IF_FAILED up here instead of saving the HR?

Frameworks/CoreGraphics/CGContext.mm
ms-jihua: if you avoided the copy, you could also avoid the resize up top :0

---
In reply to: [100875602](https://github.com/Microsoft/WinObjC/pull/1933#discussion_r100875602) [](ancestors = 100875602)

Frameworks/CoreGraphics/CGContext.mm
aballway: the macro breaks with all of this stuff in it

ms-jihua: yeesh

---
In reply to: [100879340](https://github.com/Microsoft/WinObjC/pull/1933#discussion_r100879340) [](ancestors = 100879340)

Frameworks/CoreGraphics/CGContext.mm
ms-jihua: I think we prefer CFSTR() when we're in the Core_____ frameworks

---
In reply to: [100870707](https://github.com/Microsoft/WinObjC/pull/1933#discussion_r100870707) [](ancestors = 100870707)

Frameworks/CoreGraphics/CGContext.mm
aballway: Right, they could be very different sizes.

Frameworks/CoreGraphics/CGContext.mm
aballway: the only time we were passing nullptr was the single glyph case, where the advance wasn't being used

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -415,7 +421,11 @@ inline HRESULT PopEndDraw() {
     HRESULT Draw(_CGCoordinateMode coordinateMode, CGAffineTransform* additionalTransform, Lambda&& drawLambda);
 
     HRESULT DrawGeometry(_CGCoordinateMode coordinateMode, ID2D1Geometry* pGeometry, CGPathDrawingMode drawMode);
-    HRESULT DrawGlyphRun(const DWRITE_GLYPH_RUN* glyphRun, bool transformByGlyph = true);
+    HRESULT DrawGeometryInternal(ID2D1Geometry* geometry,
```
ms-jihua: if it's internal, let's put it in the private section with a _ name

Frameworks/CoreGraphics/CGContext.mm
aballway: Even if we don't transform by glyph if we have > 1 runs we will need to do this to clip/stroke at once to prevent the shadowing/alpha issues.  I suppose we could make a fast path for "simple text transform w/ just fill" as that's going to be the most common scenario but I think that'd lead to a good amount of code duplication.

Frameworks/CoreText/CTFrame.mm
```diff
@@ -113,28 +113,39 @@ void CTFrameDraw(CTFrameRef frameRef, CGContextRef ctx) {
...
+        });
+
+        std::vector<GlyphRunData> runs;
 
         for (size_t i = 0; i < frame->_lineOrigins.size() && (frame->_lineOrigins[i].y < frame->_frameRect.size.height); ++i) {
             _CTLine* line = static_cast<_CTLine*>([frame->_lines objectAtIndex:i]);
-            CGContextSetTextPosition(ctx, frame->_lineOrigins[i].x, frame->_lineOrigins[i].y);
-            CTLineDraw(static_cast<CTLineRef>(line), ctx);
+            CGPoint relativePosition = frame->_lineOrigins[i];
+            for (size_t i = 0; i < [line->_runs count]; ++i) {
```
ms-jihua: redefinition of i here

aballway: ! how does this work?

Frameworks/CoreGraphics/CGContext.mm
aballway: It was before but I ended up capturing **a lot** so changed it to let the compiler figure it out.  I also don't like [&] but it seems clearer to me here than a whole line of variables being captured

Frameworks/CoreGraphics/CGContext.mm
aballway: This isn't the normal case, just the CGShowText methods.  How would we get away with not copying?

Frameworks/CoreGraphics/CGContext.mm
aballway: I don't know that it would make sense to define these in CG and export them for CT (especially since we're not sharing values w/ UIKit consts).  Not stubborn on this so if others think this is worth changing I'll fix it.

ms-jihua:
Some of the test images don't appear to have changed despite being in the change?

Frameworks/CoreGraphics/CGContext.mm
ms-jihua: it sounds like we should still draw the ones that are big enough to be shown, then.

---
In reply to: [100880090](https://github.com/Microsoft/WinObjC/pull/1933#discussion_r100880090) [](ancestors = 100880090)

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1918,66 +1948,161 @@ void CGContextSetPatternPhase(CGContextRef context, CGSize phase) {
...
-                     return S_OK;
-                 }));
+        // Get the linear transformation that is the inverse of the text transform
+        // We transform the positions of each glyph by the inverse so they will be positioned correctly
+        // While still being transformed per glyph by the text transformation
+        CGAffineTransform invertedTextTransformation = CGAffineTransformInvert(
+            CGAffineTransformScale(CGAffineTransformMake(textTransform.a, textTransform.b, textTransform.c, textTransform.d, 0, 0), 1, -1));
+
+        // Vectors which hold the positions of each glyph that we give to DWRITE for drawing
+        std::vector<std::vector<DWRITE_GLYPH_OFFSET>> positionsVectors(runsCount);
```
ms-jihua: for both here, and below, where i mentioned that i disliked the unnecessary operations for the most common case, i agree that code duplication is bad. the solution i would've rather you taken would've been something like, store pointers, pref smart ones, instead of actual vectors/structs. that way you can just have a pointer to the original vector/struct for the most common case.

ms-jihua: thinking on it more it might actually be that you can't avoid constructing the glyph runs below, in which case you could construct them up here, and use their glyphOffsets field to store a ptr to either original glyphOffsets or transformed glyphOffsets (positionsVectors and glyphRuns sizes would become intentionally out of sync)

---
In reply to: [100935594](https://github.com/Microsoft/WinObjC/pull/1933#discussion_r100935594) [](ancestors = 100935594)

aballway:
Triple checked this to make sure the changes are consistent 😹 
image

Frameworks/CoreText/CTLine.mm
```diff
@@ -228,15 +229,19 @@ void CTLineDraw(CTLineRef lineRef, CGContextRef ctx) {
...
-            CGPoint curTextPos = CGContextGetTextPosition(ctx);
-            CGContextSetTextPosition(ctx, curTextPos.x + curRun->_relativeXOffset, curTextPos.y);
+            relativePosition.x += curRun->_relativeXOffset;
         }
+        runs.emplace_back(GlyphRunData{ &curRun->_dwriteGlyphRun, relativePosition, (CFDictionaryRef)curRun->_attributes.get() });
+    }
 
-        CTRunDraw(static_cast<CTRunRef>(curRun), ctx, CFRange{});
+    if (!runs.empty()) {
+        _CGContextDrawGlyphRuns(ctx, runs.data(), runs.size());
```
DHowett-MSFT: There's no point in doing a push begin and pop end if `runs.empty()`. Since we're not managing gstates, it would also be reasonable to get rid of the scopeguard.

Frameworks/CoreText/CTFrame.mm
```diff
@@ -113,28 +113,41 @@ void CTFrameDraw(CTFrameRef frameRef, CGContextRef ctx) {
...
+                }
+                runs.emplace_back(GlyphRunData{ &curRun->_dwriteGlyphRun, relativePosition, (CFDictionaryRef)curRun->_attributes.get() });
+            }
         }
 
-        // Restore CTM and Text Matrix to values before we modified them
-        CGContextRestoreGState(ctx);
-        CGContextSetTextMatrix(ctx, textMatrix);
+        if (!runs.empty()) {
+            _CGContextDrawGlyphRuns(ctx, runs.data(), runs.size());
```
DHowett-MSFT: There's no point in doing a push begin and pop end if `runs.empty()`. You could concentrate all the drawing logic here.

tests/unittests/CoreGraphics.drawing/CTDrawingTests.cpp
```diff
@@ -569,3 +569,56 @@ TEXT_DRAW_TEST_F(CTFont, DrawGlyphs, WhiteBackgroundTest<PixelByPixelImageCompar
...
+
+    woc::unique_cf<CTParagraphStyleRef> paragraphStyle{ CTParagraphStyleCreate(setting, std::extent<decltype(setting)>::value) };
+    woc::unique_cf<CTFontRef> myCFFont{ CTFontCreateWithName(CFSTR("Arial"), 20, nullptr) };
+
+    CFStringRef keys[3] = { kCTFontAttributeName, kCTParagraphStyleAttributeName, kCTForegroundColorFromContextAttributeName };
+    CFTypeRef values[3] = { myCFFont.get(), paragraphStyle.get(), kCFBooleanTrue };
+
+    CGContextSetRGBFillColor(context, 1, 0, 0, 1);
+    CGContextSetRGBStrokeColor(context, 0, 1, 0, 1);
+    CGContextSetShadow(context, { 3, 5 }, 1);
```
DHowett-MSFT: I must admit, I was hoping for the shadow test to be separate. One per text drawing mode, including clip, but i suppose it doesn't matter since this is a strict superset.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1918,66 +1948,161 @@ void CGContextSetPatternPhase(CGContextRef context, CGSize phase) {
...
+         !transformByGlyph) &&
+        state.textDrawingMode == kCGTextFill) {
+        // Able to fast-path simple text drawing that doesn't need to clip/stroke nor has text transformations
+        RETURN_IF_FAILED(Draw(_kCGCoordinateModeDeviceSpace, nullptr, [&](CGContextRef context, ID2D1DeviceContext* deviceContext) {
+            D2D1_POINT_2F origin{ 0, 0 };
+            for (size_t i = 0; i < runsCount; ++i) {
+                if (glyphRuns[i].attributes) {
+                    CGColorRef fontColor = (CGColorRef)CFDictionaryGetValue(glyphRuns[i].attributes, _kCGForegroundColorAttributeName);
+                    if (fontColor) {
+                        CGContextSetFillColorWithColor(context, fontColor);
```
DHowett-MSFT: Should this manipulate global drawing state, or should it set the D2D brush to use later?

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1913,71 +1943,155 @@ void CGContextSetPatternPhase(CGContextRef context, CGSize phase) {
...
+        state.textDrawingMode == kCGTextFill;
+
+    std::vector<GlyphRunData> runs;
+
+    size_t maxGlyphCount = std::max_element(glyphRuns, glyphRuns + runsCount, [](const GlyphRunData& left, const GlyphRunData& right) {
+                               return left.run->glyphCount < right.run->glyphCount;
+                           })->run->glyphCount;
+
+    // Shared between transformed glyph runs
+    std::unique_ptr<CGFloat[]> zeroAdvances(new CGFloat[maxGlyphCount]());
```
DHowett-MSFT: What is the benefit of this over a vector? Is it that it doesn't painstakingly initialize each member?

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1913,71 +1943,155 @@ void CGContextSetPatternPhase(CGContextRef context, CGSize phase) {
...
+
+    HRESULT ret = Draw(_kCGCoordinateModeDeviceSpace, nullptr, [&](CGContextRef context, ID2D1DeviceContext* deviceContext) {
+        if (state.textDrawingMode & kCGTextFill) {
+            deviceContext->SetTransform(__CGAffineTransformToD2D_F(combinedTransform));
+            D2D1_POINT_2F origin{ 0, 0 };
+            for (auto& runData : runs) {
+                if (runData.attributes) {
+                    CGColorRef fontColor = (CGColorRef)CFDictionaryGetValue(runData.attributes, _kCGForegroundColorAttributeName);
+                    if (fontColor) {
+                        CGContextSetFillColorWithColor(context, fontColor);
```
DHowett-MSFT: Should this manipulate global state or simply change which ID2D1Brush is used?

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1913,71 +1943,155 @@ void CGContextSetPatternPhase(CGContextRef context, CGSize phase) {
...
+    RETURN_IF_FAILED(ret);
+
+    if (state.textDrawingMode & kCGTextClip) {
+        RETURN_IF_FAILED(state.IntersectClippingGeometry(transformedGeometry.Get(), kCGPathFill));
+    }
+
+    if (!fastPathTextDrawing) {
+        // Need to delete all of the created glyphOffsets and DWRITE_GLYPH_RUNS
+        for (auto& runData : runs) {
+            delete[] runData.run->glyphOffsets;
```
DHowett-MSFT: Curious; there should be some sort of RAII container for this, right?

aballway: Ideally, but I'm not sure there is one that could be created in a smaller scope than it's needed in, especially since the DWRITE_GLYPH_RUN doesn't have a RAII container for the glyphOffset

Frameworks/CoreGraphics/CGContext.mm
aballway: Our current implementation has it modifying the global state

Frameworks/CoreGraphics/CGContext.mm
aballway: I'm not sure there are major gains by using a unique_ptr rather than vector in this instance, just thought it would be clearer since we don't need to iterate over it or use any other vector functionality

ms-jihua:
looking at this chart, some of your post-change times are higher than mine after #1923. could just be debug/release and machine differences, but...

---
In reply to: [279803769](https://github.com/Microsoft/WinObjC/pull/1933#issuecomment-279803769) [](ancestors = 279803769)

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1913,71 +1943,155 @@ void CGContextSetPatternPhase(CGContextRef context, CGSize phase) {
...
+    std::unique_ptr<CGFloat[]> zeroAdvances(new CGFloat[maxGlyphCount]());
+
+    for (size_t i = 0; i < runsCount; ++i) {
+        auto& run = glyphRuns[i].run;
+        if ((fabs(combinedTransform.a * run->fontEmSize) <= c_glyphThreshold &&
+             fabs(combinedTransform.b * run->fontEmSize) <= c_glyphThreshold) ||
+            (fabs(combinedTransform.d * run->fontEmSize) <= c_glyphThreshold &&
+             fabs(combinedTransform.c * run->fontEmSize) <= c_glyphThreshold)) {
+            // Not a failure state! Not drawing the glyphs is *okay*.
+            TraceWarning(TAG, L"Glyphs too small to be rendered");
```
ms-jihua: probably use continue here, to get rid of the else below.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1913,71 +1943,155 @@ void CGContextSetPatternPhase(CGContextRef context, CGSize phase) {
...
-    CGAffineTransform finalTextTransform = CGAffineTransformConcat(textTransform, deviceTransform);
-    if ((fabs(finalTextTransform.a * glyphRun->fontEmSize) <= c_glyphThreshold &&
-         fabs(finalTextTransform.b * glyphRun->fontEmSize) <= c_glyphThreshold) ||
-        (fabs(finalTextTransform.d * glyphRun->fontEmSize) <= c_glyphThreshold &&
-         fabs(finalTextTransform.c * glyphRun->fontEmSize) <= c_glyphThreshold)) {
-        TraceWarning(TAG, L"Glyphs too small to be rendered");
-
-        // Not a failure state! Not drawing the glyphs is *okay*.
-        return S_FALSE;
+    bool fastPathTextDrawing =
```
ms-jihua: didn't you have more granularity in iteration 1 than this? (can you fast-path some stuff even if textDrawingMode isn't Fill?)

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1913,71 +1943,155 @@ void CGContextSetPatternPhase(CGContextRef context, CGSize phase) {
...
+            } else {
+                // Get the linear transformation that is the inverse of the text transform
+                // We transform the positions of each glyph by the inverse so they will be positioned correctly
+                // While still being transformed per glyph by the text transformation
+                CGAffineTransform invertedTextTransformation = CGAffineTransformInvert(
+                    CGAffineTransformScale(CGAffineTransformMake(textTransform.a, textTransform.b, textTransform.c, textTransform.d, 0, 0),
+                                           1,
+                                           -1));
+
+                DWRITE_GLYPH_OFFSET* positions = new DWRITE_GLYPH_OFFSET[run->glyphCount];
```
ms-jihua: ? just keep this in a vector<vector>> like you had?

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1913,71 +1943,155 @@ void CGContextSetPatternPhase(CGContextRef context, CGSize phase) {
...
+                    // So when the space is transformed by the text position it will be drawn in the correct real position
+                    positions[j] = DWRITE_GLYPH_OFFSET{ transformedPosition.x + run->glyphOffsets[j].advanceOffset,
+                                                        std::round(transformedPosition.y + run->glyphOffsets[j].ascenderOffset) };
+
+                    // Translate position of next glyph by current glyph's advance
+                    runningPosition.x += run->glyphAdvances[j];
+                }
+
+                DWRITE_GLYPH_RUN* transformedGlyphRun =
+                    new DWRITE_GLYPH_RUN{ run->fontFace,      run->fontEmSize, run->glyphCount, run->glyphIndices,
```
ms-jihua: std::make_shared this instead of new-ing it. if you don't want to change the contract of GlyphRunData to take a shared_ptr instead of a *, then you can just stuff this in some unrelated vector so that it can get properly scope-released

Frameworks/CoreGraphics/CGContext.mm
ms-jihua: I commented above about how you can make this more RAII

---
In reply to: [101130382](https://github.com/Microsoft/WinObjC/pull/1933#discussion_r101130382) [](ancestors = 101130382)

Frameworks/CoreGraphics/CGContext.mm
aballway: If it's not just fill then we have to do the transformations for clip/stroke,

Frameworks/CoreGraphics/CGContext.mm
aballway: It can't be a shared_ptr because it points to the DWRITE_GLYPH_RUN in the _CTRun, and I don't think moving it up in scope to a vector makes sense because then we'd be doing a whole bunch more copying in the fast case.

ms-jihua: why would you be doing a bunch more copying?

---
In reply to: [101147313](https://github.com/Microsoft/WinObjC/pull/1933#discussion_r101147313) [](ancestors = 101147313)

aballway: I would need to copy the input DWRITE_GLYPH_RUNS into the new vector

ms-jihua: why wouldn't you use a vector of shared_ptr<DWRITE_GLYPH_RUNS> so you could avoid that?

---
In reply to: [101148834](https://github.com/Microsoft/WinObjC/pull/1933#discussion_r101148834) [](ancestors = 101148834)

aballway: that'd have the same problem as using make_shared, we'd be deleting the _CTRun's DWRITE_GLYPH_RUN when it gets destroyed

ms-jihua: sorry, lost my train of thought for a second there.
you can put only the DWRITE_GLYPH_RUNs you would've new-ed in a vector for RAII, and leave the _CTRun-owned DWRITE_GLYPH_RUNs alone.

---
In reply to: [101151683](https://github.com/Microsoft/WinObjC/pull/1933#discussion_r101151683) [](ancestors = 101151683)

aballway: But we wouldn't be able to take a ptr into the vector since we're adding glyph runs and it could be invalidated.  We'd need a set size vector, but to do that we'd need to loop through the GlyphRunDatas and copy all that aren't too small first, then have an extra vector for possible positions and DWRITE_GLYPH_RUNS, and I think that complicates the code more than having the raw pointers to glyphOffsets and runs

ms-jihua: discussed this offline, i'm bad at communicating

---
In reply to: [101155233](https://github.com/Microsoft/WinObjC/pull/1933#discussion_r101155233) [](ancestors = 101155233)

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1898,86 +1906,197 @@ void CGContextSetPatternPhase(CGContextRef context, CGSize phase) {
...
+    // Shared between transformed glyph runs
+    std::unique_ptr<CGFloat[]> zeroAdvances(new CGFloat[maxGlyphCount]());
+
+    // DWrite will crash if we try to give it glyphs that are below this threshold
+    // Though this value is approximate, it is small enough to not be noticeable while still safe
+    static constexpr float c_glyphThreshold = 0.5f;
+    std::vector<GlyphRunData> runs;
+
+    // For cases where we need to create glyphOffsets and runs
+    std::vector<std::vector<DWRITE_GLYPH_OFFSET>> createdOffsets;
```
ms-jihua: is it worth it to .reserve() on these?

aballway: Possibly, but adds unneeded memory use on the fast path, & we're not certain of the size we'll need

PR #1937, opened 2017-02-09
Removing an unnecessary call to run UI appliation on the main thread.

pradipd:
Ui application initialize already forces itself to run on the main thread.

This is issue was discovered when testing https://github.com/Microsoft/WinObjC/pull/1872.
FunctionalTestSetupUIApplication calls 
  RunSynchronouslyOnMainThread calls

which looks like:
inline void RunSynchronouslyOnMainThread(void (^block)()) {
    if ([NSThread isMainThread]) {
        block(); //block =  UIApplicationInitializeFunctionalTest
    } else {
        dispatch_sync(dispatch_get_main_queue(), block);  //block = UIApplicationInitializeFunctionalTest
    }

The problem is isMainThread is not properly initialize till after UIApplicationInitializeFunctionalTest is called.

Up until now, we have just been lucky NSThread isMainThread returned true.  Even though, in functional tests (TAEF) we know we are not on the main thread.  If you look at isMainThread.
+ (BOOL)isMainThread {
    return [self _threadObjectFromCurrentThread] == [self mainThread];
}
this evaluates to nil == nil for functional tests before UIApplicationInitializeFunctionalTest is called.

But, our functional tests have always worked because UIApplicationInitializeFunctionalTest itself already takes care of calling the important pieces on the main thread.

So, the call to RunSynchronouslyOnMainThread in FunctionalTestSetupUIApplication  seems redundant since UIApplicationInitializeFunctionalTest  already knows how to initialize itself.

And, with https://github.com/Microsoft/WinObjC/pull/1872, this caused the FT's to hang, because isMainThread returns false (which is correct!), but, now we call libdispatch to get the main queue, but that hasn't been setup yet because UIApplicationInitializeFunctionalTest.

tl;dr 
Call UIApplicationInitializeFunctionalTest before relying on any objective-c thread related activities.

Additional Validation: 
I also tested this change with https://github.com/Microsoft/WinObjC/pull/1872 and confirmed the PR worked.

### DIFF
--- a/Frameworks/UIKit.Xaml/Label.xaml.cpp
+++ b/Frameworks/UIKit.Xaml/Label.xaml.cpp
@@ -61,6 +61,7 @@ Canvas^ Label::SublayerCanvas::get() {
 
 Windows::Foundation::Size Label::ArrangeOverride(Windows::Foundation::Size finalSize) {
     // Make sure we render vertically-centered text if possible, else cap at the containing layer's height.
+    // TODO: Issue #1946: Do we actually need this call to Measure?
     TextBlock->Measure(finalSize);
     if (TextBlock->DesiredSize.Height >= finalSize.Height) {
         TextBlock->VerticalAlignment = Windows::UI::Xaml::VerticalAlignment::Top;
--- a/Frameworks/UIKit.Xaml/Layer.xaml.cpp
+++ b/Frameworks/UIKit.Xaml/Layer.xaml.cpp
@@ -105,6 +105,11 @@ void Layer::_RegisterDependencyProperties() {
 // ObjectiveC Interop
 ////////////////////////////////////////////////////////////////////////////////////
 
+// Initializes our library for use; registers dependency properties, etc.
+UIKIT_XAML_EXPORT void UIKitXamlInitialize() {
+    UIKit::Xaml::Private::CoreAnimation::Layer::_RegisterDependencyProperties();
+}
+
 // Set one or more layer properties for the specified target xaml element
 UIKIT_XAML_EXPORT void XamlSetFrameworkElementLayerProperties(
     const Microsoft::WRL::ComPtr<IInspectable>& targetElement,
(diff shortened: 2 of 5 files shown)


pradipd:

pradipd:

tests/functionaltests/FunctionalTestHelpers.mm
```diff
@@ -30,11 +30,7 @@
 
 // Launches the functional test app
 bool FunctionalTestSetupUIApplication() {
-    RunSynchronouslyOnMainThread(^{
```
jaredhms: >RunSynchronouslyOnMainThread [](start = 4, length = 28)

Maybe we should be using CoreDispatcher APIs rather than NSThread, etc for this?  How can you be positive that FunctionalTestSetupUIApplication is getting called from the UI thread?

pradipd: I guarantee that it is *not* being called on the UI thread.

UIApplicationInitializeFunctionalTest uses Coredispatcher to call the important parts on the UI thread. 


---
In reply to: [100440302](https://github.com/Microsoft/WinObjC/pull/1937#discussion_r100440302) [](ancestors = 100440302)

tests/functionaltests/FunctionalTestHelpers.mm
```diff
@@ -30,11 +30,7 @@
...
-        // The name of our default 'AppDelegate' class
-        UIApplicationInitializeFunctionalTest(nullptr, Strings::NarrowToWide<std::wstring>(NSStringFromClass([AppDelegate class])).c_str());
-    });
-
+    UIApplicationInitializeFunctionalTest(nullptr, Strings::NarrowToWide<std::wstring>(NSStringFromClass([AppDelegate class])).c_str());
```
jaredhms: >    UIApplicationInitializeFunctionalTest(nullptr, St [](start = 0, length = 53)

Ah! in that case, please add a detailed comment so we don't do this again

pradipd: Is that what those // are for?
Will do.

---
In reply to: [100440979](https://github.com/Microsoft/WinObjC/pull/1937#discussion_r100440979) [](ancestors = 100440979)

jaredhms:
:shipit:

PR #1938, opened 2017-02-10
Implement Antialiasing render modes for CGContext.

### LINKED ISSUE #1177: Propagate CGContext drawing properties into Direct2D
- [ ] interpolation quality
- [ ] blend mode (#1176)
- [x] shadows (#1176)
- [ ] rendering intent (#1176)
- [ ] antialiasing
- [ ] font subpixel positioning
- [ ] font smoothing
- [ ] font subpixel quantization

### LINKED ISSUE #1349: Antialias for images, curves and fonts
Dear WinObjC Engineers:

I notice that compositing for UIImage is not re-sampled in a interpolative way (misnomer: antialiased). This is an integral feature for displaying a high resolution image on different DPI devices (normal and retina) instead of using the `@2x , @3x`, etc. method. I have the following requests:

- CGContextSetShouldAntialias() be implemented.

- UIImage apply a resampling filter better than the current one.

- Antialiased fonts.

Sincerely,

Ed

Reference:

void CGContextSetShouldAntialias(CGContextRef ctx, bool shouldAntialias) {
    UNIMPLEMENTED();
    ctx->Backing()->CGContextSetShouldAntialias(shouldAntialias);
}

MSFTFox:
Match CGContext antialiasing parameters to D2D properties in draw APIs for CGContext. D2D defaults are respected unless an API is explicitly called.

Fixes #1177 
Fixes #1349


Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -121,6 +122,14 @@
     woc::StrongCF<CGFontRef> font;
     CGFloat fontSize = 0.f;
 
+    // Antialiasing
+    _CGAntialiasToggle shouldAntialias = _kCGToggleDefault;
+
+    // Subpixels
```
MSFTFox: Font Antialiasing*

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -292,6 +304,84 @@ inline bool _ShouldDrawToCommandList() {
...
+            CurrentGState().shouldSubpixelPosition == _kCGToggleOn && allowsFontSubpixelPositioning == _kCGToggleOn) {
+            if (CurrentGState().shouldSubpixelQuantizeFonts == _kCGToggleOn && allowsFontSubpixelQuantization == _kCGToggleOn) {
+                return DWRITE_RENDERING_MODE_NATURAL_SYMMETRIC;
+            }
+            return DWRITE_RENDERING_MODE_NATURAL;
+        }
+        return DWRITE_RENDERING_MODE_DEFAULT;
+    }
+
+    inline IDWriteRenderingParams* GetTextRenderingParams(IDWriteRenderingParams* originalParams) {
```
MSFTFox: Follow D2DWrapper factory code style to respect HRESULTs in their proper place as well as in place construction of DWrite interface without passing raw interface pointers.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1352,47 +1442,47 @@ CGPoint CGContextGetTextPosition(CGContextRef context) {
...
  @Status Stub
 */
 void CGContextSetShouldSmoothFonts(CGContextRef context, bool shouldSmooth) {
     NOISY_RETURN_IF_NULL(context);
-    UNIMPLEMENTED();
+    context->SetShouldSmoothFonts(shouldSmooth ? _kCGToggleOn : _kCGToggleOff);
 }
 
 /**
  @Status Stub
```
MSFTFox: Update API status.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -58,6 +58,7 @@
 static const float s_kCGGradientOffsetPoint = 1E-45;
 
 enum _CGCoordinateMode : unsigned int { _kCGCoordinateModeDeviceSpace = 0, _kCGCoordinateModeUserSpace };
+enum _CGAntialiasToggle : unsigned int { _kCGToggleDefault = 2, _kCGToggleOff = 0, _kCGToggleOn = 1 };
```
DHowett-MSFT: This isn't antialiasing-specific; perhaps something like _CGTrinary would be better.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -151,6 +160,10 @@ inline bool ShouldDraw() {
         return std::fpclassify(alpha) != FP_ZERO && std::fpclassify(globalAlpha) != FP_ZERO;
     }
 
+    inline bool ShouldAntialias() {
```
DHowett-MSFT: bool?

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -246,11 +259,10 @@ inline HRESULT GetTarget(ID2D1Image** pTarget) {
     // See the comments above _CGContextSetShadowProjectionTransform.
     CGAffineTransform shadowProjectionTransform{ CGAffineTransformIdentity };
 
-    // TODO(DH) GH#1070 evaluate these defaults; they should be set by context creators.
-    bool allowsAntialiasing = false;
-    bool allowsFontSmoothing = false;
-    bool allowsFontSubpixelPositioning = false;
-    bool allowsFontSubpixelQuantization = false;
+    _CGAntialiasToggle allowsAntialiasing = _kCGToggleDefault;
```
DHowett-MSFT: These do not have a direct mapping to D2D. The default should be _true, allow_. They aren't trimodal.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -292,6 +304,84 @@ inline bool _ShouldDrawToCommandList() {
     HRESULT _CreateShadowEffect(ID2D1Image* inputImage, ID2D1Effect** outShadowEffect);
 
 public:
+    inline D2D1_ANTIALIAS_MODE GetAntialiasMode() {
+        if (CurrentGState().shouldAntialias == _kCGToggleDefault && allowsAntialiasing == _kCGToggleDefault) {
```
DHowett-MSFT: Dropping "default" from the _allow_ states will simplify these.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -292,6 +304,84 @@ inline bool _ShouldDrawToCommandList() {
...
+            return D2D1_ANTIALIAS_MODE_PER_PRIMITIVE;
+        }
+
+        return CurrentGState().shouldAntialias == _kCGToggleOn && allowsAntialiasing == _kCGToggleOn ? D2D1_ANTIALIAS_MODE_PER_PRIMITIVE :
+                                                                                                       D2D1_ANTIALIAS_MODE_ALIASED;
+    }
+
+    inline D2D1_TEXT_ANTIALIAS_MODE GetTextAntialiasMode() {
+        if (CurrentGState().shouldSmoothFonts == _kCGToggleDefault && allowsFontSmoothing == _kCGToggleDefault) {
+            return D2D1_TEXT_ANTIALIAS_MODE_DEFAULT;
```
DHowett-MSFT: I think I misunderstood what you meant when you said Default. I assumed that _nothing would be set on the render target_ if the value was "default"

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -292,6 +304,84 @@ inline bool _ShouldDrawToCommandList() {
...
+            CurrentGState().shouldSmoothFonts == _kCGToggleOn && allowsFontSmoothing == _kCGToggleOn) {
+            return D2D1_TEXT_ANTIALIAS_MODE_CLEARTYPE;
+        } else {
+            return D2D1_TEXT_ANTIALIAS_MODE_GRAYSCALE;
+        }
+
+        return D2D1_TEXT_ANTIALIAS_MODE_ALIASED;
+    }
+
+    inline DWRITE_RENDERING_MODE GetRenderingMode() {
```
DHowett-MSFT: nit: GetTextRenderingMode

Frameworks/CoreGraphics/CGContext.mm
DHowett-MSFT: perhaps `CreateDerivedTextRenderingParams`, and these can all be private and underscore-prefixed since they're consumed by `::Draw`

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -292,6 +304,84 @@ inline bool _ShouldDrawToCommandList() {
...
+        if (FAILED(DWriteCreateFactory(DWRITE_FACTORY_TYPE_SHARED, __uuidof(IDWriteFactory), &dwriteFactory))) {
+            return originalParams;
+        }
+        if (!originalParams) {
+            dwriteFactory->CreateRenderingParams(&workingParams);
+        }
+
+        IDWriteRenderingParams* textRenderingParams;
+
+        dwriteFactory->CreateCustomRenderingParams(workingParams->GetGamma(),
```
DHowett-MSFT: I am really confused as to why there are three rendering params being passed all around.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -292,6 +304,84 @@ inline bool _ShouldDrawToCommandList() {
...
+
+    inline IDWriteRenderingParams* GetTextRenderingParams(IDWriteRenderingParams* originalParams) {
+        IDWriteRenderingParams* workingParams = originalParams;
+
+        ComPtr<IDWriteFactory> dwriteFactory;
+        if (FAILED(DWriteCreateFactory(DWRITE_FACTORY_TYPE_SHARED, __uuidof(IDWriteFactory), &dwriteFactory))) {
+            return originalParams;
+        }
+        if (!originalParams) {
+            dwriteFactory->CreateRenderingParams(&workingParams);
```
DHowett-MSFT: you're doing this to get the defaults?
you're leaking it.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -292,6 +304,84 @@ inline bool _ShouldDrawToCommandList() {
...
+        }
+
+        IDWriteRenderingParams* textRenderingParams;
+
+        dwriteFactory->CreateCustomRenderingParams(workingParams->GetGamma(),
+                                                   workingParams->GetEnhancedContrast(),
+                                                   workingParams->GetClearTypeLevel(),
+                                                   workingParams->GetPixelGeometry(),
```
DHowett-MSFT: Is the "RGB" or "BGR" pixel geometry a better approximation for `ShouldSubpixelPosition`? It definitely sounds like that flag should tell dwrite that RGB are in separate places. The other option is FLAT, which implies that there is no subpixel positioning.

ClearType is for RGB antialiasing; perhaps something from CG maps to Grayscale antialiasing?

build/Tests/UnitTests/CoreGraphics.Drawing/CoreGraphics.Drawing.UnitTests.vcxproj
```diff
@@ -237,6 +237,7 @@
     </ClangCompile>
   </ItemDefinitionGroup>
   <ItemGroup>
+    <ClangCompile Include="$(StarboardBasePath)\tests\unittests\CoreGraphics.drawing\CGContextDrawing_AntiAliasTests.cpp" />
```
DHowett-MSFT: nit: move this down with its friends. The entry point and framework are here, the tests are below.

Frameworks/CoreGraphics/CGContext.mm
rajsesh-msft: I question a separate value for default, shouldn't it be one of the two real options?

Frameworks/CoreGraphics/CGContext.mm
rajsesh-msft: while we are at it, it should be ShouldAntialias() const

Frameworks/CoreGraphics/CGContext.mm
rajsesh-msft: There is no direct mapping between how Mac does subpixel positioning and Dwrite does it.  From the "intention" standpoint, subpixel positioning should map to clear type.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1352,47 +1442,47 @@ CGPoint CGContextGetTextPosition(CGContextRef context) {
 */
 void CGContextSetAllowsFontSmoothing(CGContextRef context, bool allows) {
     NOISY_RETURN_IF_NULL(context);
-    UNIMPLEMENTED();
+    context->allowsFontSmoothing = allows ? _kCGToggleOn : _kCGToggleOff;
 }
 
 /**
  @Status Stub
```
rajsesh-msft: please update annotations, these are no longer stubs.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -246,11 +255,10 @@ inline HRESULT GetTarget(ID2D1Image** pTarget) {
     // See the comments above _CGContextSetShadowProjectionTransform.
     CGAffineTransform shadowProjectionTransform{ CGAffineTransformIdentity };
 
-    // TODO(DH) GH#1070 evaluate these defaults; they should be set by context creators.
-    bool allowsAntialiasing = false;
-    bool allowsFontSmoothing = false;
-    bool allowsFontSubpixelPositioning = false;
-    bool allowsFontSubpixelQuantization = false;
+    _CGTrinary allowsAntialiasing = _kCGTrinaryDefault;
```
DHowett-MSFT: I still don't think there's a need for a terinary for _allow_.

build/Tests/UnitTests/CoreGraphics.Drawing/CoreGraphics.Drawing.UnitTests.vcxproj
DHowett-MSFT: nit: move this down with its friends. The entry point and framework are here, the tests are below.

Frameworks/CoreGraphics/CGContext.mm
MSFTFox: The point of a default value here is to avoid setting any Anti aliasing information when we don't have to. It avoids the work of updating the D2D1 drawing parameters when nothing has been changed and allows D2D to make the decisions as to the best AA to use.

DHowett-MSFT: I understand the need for a trinary; I just don't think that we need to have a "yes/default/no" optionset for "Allow". Allow's default is On. If allow+default, do nothing.
If !allow+default, disable (effectively, set our default for "disallow" to be "no dwrite antialiasing" or w/e property)

---
In reply to: [100920282](https://github.com/Microsoft/WinObjC/pull/1938#discussion_r100920282) [](ancestors = 100920282)

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -2211,6 +2282,25 @@ void CGContextClearRect(CGContextRef context, CGRect rect) {
         this->PopEndDraw();
     });
 
+    // If the context has requested antialiasing other than the defaults we now need to update the device context.
+    if (CurrentGState().shouldAntialias != _kCGTrinaryDefault || allowsAntialiasing != _kCGTrinaryDefault) {
```
aballway: It would probably be cleaner to just use a bool for these values and some sort of flag that gets set whenever one of these values is changed

tests/UnitTests/CoreGraphics.drawing/CGContextDrawingTests.cpp
```diff
@@ -378,3 +378,43 @@ DRAW_TEST(CGContext, PremultipliedAlphaImage) {
     CGContextSetRGBFillColor(context, 1.0, 0.0, 0.0, 0.5);
     CGContextFillRect(context, { 0, 0, 100, 100 });
 }
+
+DRAW_TEST_F(CGContext, AntialiasToggleTranscendsGState, WhiteBackgroundTest<>) {
```
aballway: I'm not sure that this should be a drawing test, seems more like a regular unit test.

tests/unittests/CoreGraphics.drawing/CGContextDrawing_AntiAliasTests.cpp
```diff
@@ -0,0 +1,187 @@
...
+    CGPoint origin = ::testing::get<4>(GetParam());
+    __SetContextToggles(context, useAA, useSmoothing, useSubpixel, useQuantization);
+
+    std::vector<CGGlyph> glyphs{ __CreateGlyphs() };
+    __SetFontForContext(context);
+    CGContextSetTextPosition(context, origin.x, origin.y);
+    CGContextShowGlyphs(context, glyphs.data(), glyphs.size());
+}
+
+int toggles[] = { 2, 1, 0 };
```
aballway: nit: use an enum for these values

tests/UnitTests/CoreGraphics.drawing/CGContextDrawingTests.cpp
DHowett-MSFT: @aballway it is impossible to query the antialias state of the context, and this allowed us to check against the reference platform

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -2315,6 +2386,25 @@ void CGContextClearRect(CGContextRef context, CGRect rect) {
...
+
+        // If any text rendering parameters have been updated, we need to update the device context.
+        if ((CurrentGState().shouldSmoothFonts != _kCGTrinaryDefault && allowsFontSmoothing) ||
+            (CurrentGState().shouldSubpixelPosition != _kCGTrinaryDefault && allowsFontSubpixelPositioning) ||
+            (CurrentGState().shouldSubpixelQuantizeFonts != _kCGTrinaryDefault && allowsFontSubpixelQuantization)) {
+            ComPtr<IDWriteRenderingParams> originalTextRenderingParams;
+            deviceContext->GetTextRenderingParams(&originalTextRenderingParams);
+
+            ComPtr<IDWriteRenderingParams> customParams;
+            GetTextRenderingParams(&originalTextRenderingParams, &customParams);
```
DHowett-MSFT: Actually, this is bad. & calls ReleaseAndGetAddressOf.
We should make it not take a pointer to the original params; it can simply create new ones on the fly for the custom params.

DHowett-MSFT: Sorry: not take a pointer to a pointer to the original params. That just complicates the ownership rules.
I guess returning an HRESULT would be safe too, since it does COM interop.

PR #1947, opened 2017-02-10
Adding more UIButton tests - mostly UIControlState transitions tests for title, titleColor, image, backgroundImage ...

oliversa-msft:
There are 74 new UIButton tests. These tests drive an existing XAMLCatalog view controller with various UI controls to manipulate the target UIButton. Most tests perform their validation by comparing the UIKit state against the backing XAML element.

CHANGES:
=======
1. Updated XAMLCatalog with new controls to manipulate UIButtons (default, custom, system)
2. The tests are designed to drive the XAMLCatalog VC using its controls (mostly text fields) to set the UIButton in the required state
3. Added 64 state transition functional tests (e.g. setTitle:forState) 
4. Added 12 more tests that cover the public properties (e.g. currentTitleColor) and initialization defaults
5. ImageTests are performed using existing assets from XAMLCatalog - no pixel comparison just dimensions re compared.

TODO:
====
1. Now that we're based on TAEF, leverage parameterized or combination test patterns
2. Create more test-enabled controls to XAMLCatalog (UIControlState multi-segment selector, pickers, etc)
3. Consolidate more setup code into fixtures so it is re-usable.


tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
```diff
@@ -118,68 +117,1128 @@ - (void)dealloc {
...
+            UIColor* expectedBackgroundColor = [UIColor colorWithRed:1.0f green:1.0f blue:1.0f alpha:0.0f];
+            EXPECT_TRUE(UXTestAPI::IsRGBAEqual(solidBackgroundBrush, expectedBackgroundColor));
+        });
+
+        // Title color
+        dispatch_sync(dispatch_get_main_queue(), ^{
+            WUXMSolidColorBrush* solidForegroundBrush = rt_dynamic_cast([WUXMSolidColorBrush class], textBlock.foreground);
+
+            // BUGBUG: Thought the default should be whiteColor
+            UIColor* expectedTitleColor = [UIColor blackColor];
```
oliversa-msft: This test will be updated shortly - the title color is being updated to a white color but the timing of this test picks up the XAML default.

include/UIKit/UIButton.h
```diff
@@ -86,7 +86,7 @@ UIKIT_EXPORT_CLASS
 @property (nonatomic) UIEdgeInsets contentEdgeInsets;
 @property (nonatomic) UIEdgeInsets imageEdgeInsets;
 @property (nonatomic) UIEdgeInsets titleEdgeInsets;
-@property (nonatomic, readonly) UIButtonType buttonType;
+@property (nonatomic, readonly) UIButtonType buttonType STUB_PROPERTY;
```
bbowman: seems like backwards progress? was this just wrong before?

build/Tests/FunctionalTests/FunctionalTests.vcxproj
```diff
@@ -295,8 +293,12 @@
   <ItemGroup>
     <Media Include="$(MSBuildThisFileDirectory)..\..\..\tests\functionaltests\Tests\AssetsLibraryTests\AssetsLibraryTestVideo.mp4" />
   </ItemGroup>
+  <ItemGroup>
+    <UXResources Include="$(MSBuildThisFileDirectory)..\..\..\samples\XAMLCatalog\XAMLCatalog\Images\*" />
```
bbowman: >UXResources [](start = 5, length = 11)

this makes me a little sad. Could we use ReferenceCopyLocalPaths instead and let the build just do its thing to copy?

bbowman:
> 

just whitespace change? Revert this file?

---
Refers to: tests/functionaltests/FunctionalTests.cpp:58 in ea546ee. [](commit_id = ea546ee5c9150219c8ce17d76f9277172198cfba, deletion_comment = False)

bbowman:
Didn't look at the test themselves since I don't know the UI stuff but looks good otherwise.

bbowman:
:shipit:

include/UIKit/UIButton.h
oliversa-msft: To be honest, I can remove this change. Jared has a PR out that adds this feature. I'll remove.

build/Tests/FunctionalTests/FunctionalTests.vcxproj
oliversa-msft: I'll take a shot at it - VS wasn't happy when I specified them as resources/content so I followed the same pattern used by CTDrawingTests.

tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
```diff
@@ -118,68 +117,1128 @@ - (void)dealloc {
...
         __block auto xamlSubscriber = std::make_shared<XamlEventSubscription>();
+        __block NSString* expectedBackgroundImage;
+
+        WXFrameworkElement* xamlElement = [buttonToTest xamlElement];
+        ASSERT_OBJCNE(xamlElement, nil);
+
+        // Wait for the layerContent to be part of the visual tree
+        dispatch_sync(dispatch_get_main_queue(), ^{
+            ert = [xamlElement addLayoutUpdatedEvent:^(RTObject* sender, RTObject* args) {
+                dispatch_async(dispatch_get_main_queue(), ^{
```
yiyang-msft: >dispatch_async(dispatch_get_main_queue() [](start = 16, length = 40)

isn't this event firing on UI thread? do you need dispatch again?

oliversa-msft: Correct, it was not firing on the UI thread.

tests/functionaltests/FunctionalTests.cpp
```diff
@@ -54,4 +54,4 @@ MODULE_SETUP(ModuleSetup) {
 
 MODULE_CLEANUP(ModuleCleanup) {
     return SUCCEEDED(FrameworkHelper::DisplayRequest(false));
-}
\ No newline at end of file
+}
```
yiyang-msft: just to confirm. we don't have to add any "connector" code here any more? or is connector code being put somewhere else? which i didn't find.

oliversa-msft: Correct, no more trampolining/connector code required. All nice and self-contained within the mm file.

---
In reply to: [100679868](https://github.com/Microsoft/WinObjC/pull/1947#discussion_r100679868) [](ancestors = 100679868)

tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
```diff
@@ -118,68 +117,1144 @@ - (void)dealloc {
...
         __block auto xamlSubscriber = std::make_shared<XamlEventSubscription>();
 
-        UIButtonWithControlsViewController* buttonVC = [[UIButtonWithControlsViewController alloc] init];
+        dispatch_async(dispatch_get_main_queue(), ^{
+            // Register RAII event subscription handler
+            xamlSubscriber->Set(textBlock, [WXCTextBlock foregroundProperty], ^(WXDependencyObject* sender, WXDependencyProperty* dp) {
+                WUXMSolidColorBrush* solidBrush = rt_dynamic_cast([WUXMSolidColorBrush class], [sender getValue:dp]);
+
+                // Validation
+                if (UXTestAPI::IsRGBAEqual(solidBrush, [UIColor whiteColor])) {
```
yiyang-msft: >whiteColor [](start = 64, length = 10)

is whiteColor our contract for titleColor for default (normal) state for Default type UIButton? 

I was playing around and seems The ios contract is that normal state title color for normal state for default type of UIButton is default to Tint Color which is set to Blue for UIButton. or at least this is on iOS 10.

tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
```diff
@@ -118,68 +117,1144 @@ - (void)dealloc {
...
+            [buttonVC textTitle].text = @"Normal";
+            [buttonVC textTitleColorStateField].text = @"N;";
+            [buttonVC textTitleColor].text = @"redColor";
+
+            // Register RAII event subscription handler
+            xamlWidthSubscriber->Set(xamlElement, [WXCImage widthProperty], ^(WXDependencyObject* sender, WXDependencyProperty* dp) {
+                double width = DoubleFromPropertyValue([sender getValue:dp]);
+
+                // Validation
+                EXPECT_EQ(width, 40.0f);
```
yiyang-msft: >40.0f [](start = 33, length = 5)

are these defauts, if so, we should define them.

tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
oliversa-msft: Are you're sure it's not UIButtonTypeSystem? That one is blue by default. In my testing, UIButtonTypeCustom (and default) is white. IslandWood also sets its default to white.

---
In reply to: [100680288](https://github.com/Microsoft/WinObjC/pull/1947#discussion_r100680288) [](ancestors = 100680288)

tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
oliversa-msft: These are the values XAML output when we set the insets to (80, 80, 80, 80) inside a (200x200) button. It's a calculation that factors in the insets and frame of the content container. This is just one example that we can use for regression testing.

---
In reply to: [100681068](https://github.com/Microsoft/WinObjC/pull/1947#discussion_r100681068) [](ancestors = 100681068)

tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
yiyang-msft: wait, i thought default button is using UIButtonTypeSystem, yes, the blue color i was talking about is for UIButtonTypeSystem button. 

---
In reply to: [100697674](https://github.com/Microsoft/WinObjC/pull/1947#discussion_r100697674) [](ancestors = 100697674,100680288)

oliversa-msft: Default refers to the allocation method: [[alloc] init] vs [buttonWithType]. I believe [[alloc] init] is actually UIButtonTypeCustom under the covers.

---
In reply to: [100863125](https://github.com/Microsoft/WinObjC/pull/1947#discussion_r100863125) [](ancestors = 100863125,100697674,100680288)

yiyang-msft: ok. cool.

---
In reply to: [100864193](https://github.com/Microsoft/WinObjC/pull/1947#discussion_r100864193) [](ancestors = 100864193,100863125,100697674,100680288)

tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
yiyang-msft: we should make that calculation (contract) clear. otherwise, if the test failed with a value, let us say 41. it becomes difficult to track down it is our contract failed vs impl regression?

---
In reply to: [100697740](https://github.com/Microsoft/WinObjC/pull/1947#discussion_r100697740) [](ancestors = 100697740,100681068)

tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
```diff
@@ -119,69 +118,1156 @@ - (void)dealloc {
...
+            [buttonVC textTitle].text = @"Normal";
+            [buttonVC textTitleColorStateField].text = @"N;";
+            [buttonVC textTitleColor].text = @"redColor";
+
+            // Register RAII event subscription handler
+            xamlWidthSubscriber->Set(xamlElement, [WXCCanvas widthProperty], ^(WXDependencyObject* sender, WXDependencyProperty* dp) {
+                double width = DoubleFromPropertyValue([sender getValue:dp]);
+
+                // Validation
+                EXPECT_EQ(width, 30.0f); // now 0
```
jaredhms: >// now 0 [](start = 41, length = 8)

0? Are you sure you're looking at the correct element?

jaredhms:
:shipit:

tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
oliversa-msft: I was looking at contentCanvas under VS's Live Property Explorer. I've disabled tests for ContentEdgeInsets and TitleEdgeInsets after the last merge as these tests need to be updated and I'd like understand if I'm looking at the right values.

---
In reply to: [101126737](https://github.com/Microsoft/WinObjC/pull/1947#discussion_r101126737) [](ancestors = 101126737)

tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
oliversa-msft: TBH the contract to calculate insets is more complex than I initially expected. To fully understand this contract will require more time.

---
In reply to: [100920302](https://github.com/Microsoft/WinObjC/pull/1947#discussion_r100920302) [](ancestors = 100920302,100697740,100681068)

tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
jaredhms: contentCanvas should always be the same size as the Button.

jaredhms: From Button.Xaml:

<Canvas x:Name="contentCanvas" **_Width="{TemplateBinding Width}" Height="{TemplateBinding Height}"_** Background="{TemplateBinding Background}">

yiyang-msft:
:shipit:

tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
```diff
@@ -119,69 +118,1165 @@ - (void)dealloc {
...
+
+    TEST_METHOD(UIButton_BackgroundImageForStateCumulative) {
+        StrongId<UIButtonWithControlsViewController> buttonVC;
+        buttonVC.attach([[UIButtonWithControlsViewController alloc] init]);
+        UXTestAPI::ViewControllerPresenter testHelper(buttonVC);
+
+        UIButton* buttonToTest = [buttonVC defaultButton];
+
+        __block auto uxLayoutEvent = UXEvent::CreateManual();
+        __block StrongId<WXFrameworkElement> layerContent = nil;
```
jaredhms: > = nil [](start = 57, length = 6)

nit: unnecessary assignment to nil

tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
```diff
@@ -119,69 +118,1165 @@ - (void)dealloc {
...
+        ASSERT_OBJCNE(xamlElement, nil);
+
+        // Wait for the layerContent to be part of the visual tree
+        dispatch_sync(dispatch_get_main_queue(), ^{
+            EventRegistrationToken ert = [xamlElement addLayoutUpdatedEvent:^(RTObject* sender, RTObject* args) {
+                dispatch_async(dispatch_get_main_queue(), ^{
+                    // Initially there is no LayerContent element unless we set a background image
+                    if (layerContent == nil) {
+                        // Initially there is no LayerContent element
+                        layerContent = FindXamlChild(xamlElement, @"LayerContent");
```
jaredhms: >FindXamlChild(xamlElement, @"LayerContent"); [](start = 39, length = 44)

we need to retain because this returns an autoreleased object??

oliversa-msft: Exactly, the object was released more aggressively under release which made the repro more obvious.

PR #1949, opened 2017-02-11
Move UIKit.Xaml DependencyProperty registration out of UIKit.Layer's …

### LINKED ISSUE #1930: _RegisterDependencyProperties should be moved out of the Layer::Layer constructor
We cannot guarantee the the layer will be created first when we call _initWithXamlElement and expect that all dependency properties have been registered. We should move the registration of layer dependency properties into higher level initialization method such XamlCompositor::Initialize.

jaredhms:
…constructor.

Fixes #1930.


oliversa-msft:
:shipit:

PR #1952, opened 2017-02-13
Add support for colored and non-colored Patterns

### LINKED ISSUE #1591: Patterns should be anchored to the bottom (0,0) vs top (0,0) origin

msft-Jeyaram:
This fixes #1591 (Updating tests to render with proper values).
#1592


tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_PatternTests.cpp
```diff
@@ -0,0 +1,178 @@
...
+class CGPatternColored : public WhiteBackgroundTest<>,
+                         public ::testing::WithParamInterface<::testing::tuple<PatternCallback, CGFloat, int>> {
+    CFStringRef CreateOutputFilename() {
+        PatternCallback pattern = ::testing::get<0>(GetParam());
+        CGFloat alpha = ::testing::get<1>(GetParam());
+        int size = ::testing::get<2>(GetParam());
+
+        return CFStringCreateWithFormat(nullptr,
+                                        nullptr,
+                                        CFSTR("TestImage.CGContext.CGPatternColored.%s.size.%d.A%1.01f.png"),
```
msft-Jeyaram: >ternColored.%s.size.%d.A%1.01f.pn [](start = 72, length = 33)

need light naming fix.

Frameworks/CoreGraphics/CGPattern.mm
```diff
@@ -233,3 +233,8 @@ CGRect _CGPatternGetFinalPatternSize(CGPatternRef pattern) {
     RETURN_RESULT_IF_NULL(pattern, CGRectNull);
     return { CGPointZero, { ((CGPattern*)pattern)->xStep, ((CGPattern*)pattern)->yStep } };
 }
+
+bool _CGPatternIsColored(CGPatternRef pattern) {
+    FAIL_FAST_IF_NULL(pattern);
```
MSFTFox: nit: we shouldn't fail fast in private functions. Our pattern has been to return an HRESULT and fail in the base CG API function. That said, I don't think you need the null check in particular here. Since it's a private method, move the fail_fast_if_null to CGContext. If you plan on using this method in more places though I would recommend keeping the fail fast here and returning a struct for {HRESULT, returnvalue} which has been our preferred pattern or passing a return bool by reference.

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_PatternTests.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+    CGPatternRef pattern = CGPatternCreate(
+        NULL, CGRectMake(0, 0, size, size), CGAffineTransformIdentity, size, size, kCGPatternTilingConstantSpacing, false, &callbacks);
+
+    CGContextSetFillPattern(context, pattern, stencilColor);
+    CGPatternRelease(pattern);
+    CGContextFillRect(context, bounds);
+}
+
+INSTANTIATE_TEST_CASE_P(CGPatternTests,
```
MSFTFox: The stars with the lowest size don't appear. I don't think it's worth having a test for those sizes especially when the images are blank.

msft-Jeyaram: it's because the drawing gets clipped due to the size.
I can remove those tests, but they work as intended

rajsesh-msft:
There are a number of tests in this that are probably redundant.  For example, what value does having drawStencilStar.red.241.png add over drawStencilStar.purple.png?

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1770,8 +1770,10 @@ static CGImageRef __CGContextCreateRenderableImage(CGImageRef image) {
 #pragma region Drawing Parameters - Stroke / Fill Patterns
 
 template <typename ContextStageLambda> // Takes the form HRESULT(*)(CGContextRef)
-static HRESULT _CreatePatternBrush(
-    CGContextRef context, CGPatternRef pattern, const CGFloat* components, ID2D1BitmapBrush1** brush, ContextStageLambda&& contextStage) {
+static HRESULT _CreatePatternBrush(CGContextRef context,
+                                   CGPatternRef pattern,
+                                   ID2D1BitmapBrush1** brush,
+                                   ContextStageLambda&& contextStage) {
     // TODO #1592: change to support grayscale (masks) after dustins change.
```
rajsesh-msft: TODO is obsolete now.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1781,20 +1783,23 @@ static HRESULT _CreatePatternBrush(
...
-    CGBitmapInfo bitmapInfo = kCGImageAlphaPremultipliedFirst | kCGBitmapByteOrder32Big;
+    CGBitmapInfo bitmapInfo = kCGImageAlphaPremultipliedFirst | kCGBitmapByteOrderDefault;
 
     woc::unique_cf<CGContextRef> patternContext{ CGBitmapContextCreate(
         nullptr, tileSize.size.width, tileSize.size.height, bitsPerComponent, bytesPerRow, colorspace.get(), bitmapInfo) };
     RETURN_HR_IF_NULL(E_UNEXPECTED, patternContext);
 
+    // Determine if this pattern is a colored pattern (the coloring is specified in the pattern callback) or if it is stencil pattern (the
+    // color is set outside and not within the pattern callback)
+    bool isColored = _CGPatternIsColored(pattern);
```
rajsesh-msft: Nit: can you just do this in the lambda? will simplify the code.

Frameworks/CoreGraphics/CGPattern.mm
rajsesh-msft: Yes.  also, it null cannot happen here, just the code below should fail_fast it for you 😄

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_PatternTests.cpp
rajsesh-msft: Do we need to verify so many times that they are getting clipped (with different colors? ) 😸

msft-Jeyaram:
@rajsesh-msft 
removing some of the redundant sizes.

Frameworks/CoreGraphics/CGContext.mm
msft-Jeyaram: i have one more commit on top of this, looks like i didn't push it.
It takes the comment out and creates the bitmap based on colorspace.

Frameworks/CoreGraphics/CGContext.mm
msft-Jeyaram: it'll be duplicated in both stoke and fill.
I'll leave it here as the comments will also get duplicated.

---
In reply to: [100935286](https://github.com/Microsoft/WinObjC/pull/1952#discussion_r100935286) [](ancestors = 100935286)

Frameworks/CoreGraphics/CGContext.mm
msft-Jeyaram: uh still true.

---
In reply to: [100935769](https://github.com/Microsoft/WinObjC/pull/1952#discussion_r100935769) [](ancestors = 100935769)

msft-Jeyaram:
@rajsesh-msft @DHowett-MSFT  Ping!
Could you please review, i need to get this in by today.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1770,9 +1770,10 @@ static CGImageRef __CGContextCreateRenderableImage(CGImageRef image) {
 #pragma region Drawing Parameters - Stroke / Fill Patterns
 
 template <typename ContextStageLambda> // Takes the form HRESULT(*)(CGContextRef)
```
aballway: nit: update the comment w/ new params

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1848,9 +1852,18 @@ void CGContextSetFillPattern(CGContextRef context, CGPatternRef pattern, const C
     }
 
     ComPtr<ID2D1BitmapBrush1> bitmapBrush;
-    FAIL_FAST_IF_FAILED(_CreatePatternBrush(context, pattern, components, &bitmapBrush, [&](CGContextRef drawingContext) {
+    FAIL_FAST_IF_FAILED(_CreatePatternBrush(context, pattern, &bitmapBrush, [&](CGContextRef drawingContext, bool isColored) {
```
aballway: is it possible that we'll have other pattern variables to address in the future? Even if not, I think it would make more sense for the function to take (CGContextRef, CGPatternRef) rather than a bool flag, especially since the lambdas are almost identical already.

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_PatternTests.cpp
```diff
@@ -0,0 +1,373 @@
...
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#include "DrawingTest.h"
+
+#pragma region Stencil Pattern
+
+static void drawStencilStar(void* info, CGContextRef context) {
+    double r = 0.8 * 16 / 2;
```
aballway: nit: why all of the 16 / 2? make this a constant?

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_PatternTests.cpp
```diff
@@ -0,0 +1,373 @@
...
+                                           ::testing::ValuesIn(patternsizes)));
+
+#pragma endregion Stencil Pattern
+
+#pragma region Colored Pattern
+
+void drawColoredPatternMultiSquare(void* info, CGContextRef context) {
+    CGFloat subunit = 5; // the pattern cell itself is 16 by 18
+
+    CGRect myRect1 = { { 0, 0 }, { subunit, subunit } }, myRect2 = { { subunit, subunit }, { subunit, subunit } },
```
aballway: nit: this is not one of those cases where multiple variables on the same line makes sense 🌮

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_PatternTests.cpp
```diff
@@ -0,0 +1,373 @@
...
+    CGContextFillRect(context, myRect3);
+    CGContextSetRGBFillColor(context, .5, 0, .5, 0.5);
+    CGContextFillRect(context, myRect4);
+}
+
+PatternCallback ColoredPatternCallback[] = { { "drawColoredPatternMultiSquare", &drawColoredPatternMultiSquare } };
+
+//<PatternCallback,alpha,pattern size>
+class CGPatternColored : public WhiteBackgroundTest<>,
+                         public ::testing::WithParamInterface<::testing::tuple<PatternCallback, CGFloat, int>> {
```
aballway: super duper nit: should use CGFloat rather than int for CGSize width, height values

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_PatternTests.cpp
```diff
@@ -0,0 +1,373 @@
...
+
+static CGFloat alphas[] = { 0.25f, 0.5f, 1.f };
+
+INSTANTIATE_TEST_CASE_P(CGPatternTests,
+                        CGPatternColored,
+                        ::testing::Combine(::testing::ValuesIn(ColoredPatternCallback),
+                                           ::testing::ValuesIn(alphas),
+                                           ::testing::ValuesIn(patternsizes)));
+
+static void drawPatternWindowsLogo(void* info, CGContextRef context) {
```
aballway: +1

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_PatternTests.cpp
```diff
@@ -0,0 +1,373 @@
...
+    CGContextSetStrokeColorSpace(context, patternColorSpace.get());
+
+    CGContextSetStrokePattern(context, pattern.get(), &alpha);
+}
+
+static void _SetPatternForFill(
+    CGContextRef context, CGRect rect, float xStep, float yStep, CGPatternDrawPatternCallback drawpattern, CGFloat alpha) {
+    CGPatternCallbacks coloredPatternCallbacks = { 0, drawpattern, NULL };
+
+    woc::unique_cf<CGPatternRef> pattern{
```
aballway: nit: convert these and all raw CG*Refs to Auto/StrongCF

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_PatternTests.cpp
```diff
@@ -0,0 +1,373 @@
...
+    : public WhiteBackgroundTest<>,
+      public ::testing::WithParamInterface<::testing::tuple<PatternCallback, CGFloat, float, float, CGRect>> {
+    CFStringRef CreateOutputFilename() {
+        PatternCallback pattern = ::testing::get<0>(GetParam());
+        CGFloat alpha = ::testing::get<1>(GetParam());
+        float xStep = ::testing::get<2>(GetParam());
+        float yStep = ::testing::get<3>(GetParam());
+        CGRect rect = ::testing::get<4>(GetParam());
+        char rectDesc[100];
+        snprintf(rectDesc, sizeof(rectDesc), "(%0.0f.%0.0f)%0.0fx%0.0f", rect.origin.x, rect.origin.y, rect.size.width, rect.size.height);
```
aballway: I think this is a case where the default name is better than a custom one.  Might start running into path length issues too...

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_PatternTests.cpp
```diff
@@ -370,4 +370,14 @@ DISABLED_DRAW_TEST_F(CGPatternTests, PatternDrawPath, UIKitMimicTest<>) {
     CGPathRelease(theFirstPath);
     CGPathRelease(theSecondPath);
 }
+
+DISABLED_DRAW_TEST_F(CGPatternTests, PatternFillWindowsLogoRotate, UIKitMimicTest<>) {
```
aballway: why disabled?

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1770,9 +1770,10 @@ static CGImageRef __CGContextCreateRenderableImage(CGImageRef image) {
 #pragma region Drawing Parameters - Stroke / Fill Patterns
 
 template <typename ContextStageLambda> // Takes the form HRESULT(*)(CGContextRef)
-static HRESULT _CreatePatternBrush(
-    CGContextRef context, CGPatternRef pattern, const CGFloat* components, ID2D1BitmapBrush1** brush, ContextStageLambda&& contextStage) {
-    // TODO #1592: change to support grayscale (masks) after dustins change.
+static HRESULT _CreatePatternBrush(CGContextRef context,
+                                   CGPatternRef pattern,
+                                   ID2D1BitmapBrush1** brush,
+                                   ContextStageLambda&& contextStage) {
```
msft-Jeyaram: it's much nicer to have the lambda last in my opinion. 

---
In reply to: [101327104](https://github.com/Microsoft/WinObjC/pull/1952#discussion_r101327104) [](ancestors = 101327104)

Frameworks/CoreGraphics/CGContext.mm
msft-Jeyaram: no and we don't need to keep on passing it.
I just did it at one place, to add the comments on what's happening with the colored/non-colored indication.

---
In reply to: [101328139](https://github.com/Microsoft/WinObjC/pull/1952#discussion_r101328139) [](ancestors = 101328139)

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_PatternTests.cpp
msft-Jeyaram: this is just a test, doesn't make sense to make everything in tests as consts.
if so  paths test would have done mad :).


---
In reply to: [101328691](https://github.com/Microsoft/WinObjC/pull/1952#discussion_r101328691) [](ancestors = 101328691)

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_PatternTests.cpp
msft-Jeyaram: doesn't work as expected. 
Bugs associated. 

---
In reply to: [101332405](https://github.com/Microsoft/WinObjC/pull/1952#discussion_r101332405) [](ancestors = 101332405)

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_PatternTests.cpp
aballway: also, why snprintf when this is already using CFStringCreateWithFormat?

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1781,20 +1782,23 @@ static HRESULT _CreatePatternBrush(
...
         nullptr, tileSize.size.width, tileSize.size.height, bitsPerComponent, bytesPerRow, colorspace.get(), bitmapInfo) };
     RETURN_HR_IF_NULL(E_UNEXPECTED, patternContext);
 
+    // Determine if this pattern is a colored pattern (the coloring is specified in the pattern callback) or if it is stencil pattern (the
+    // color is set outside and not within the pattern callback)
+    bool isColored = _CGPatternIsColored(pattern);
+
     // Stage the drawing context
-    RETURN_IF_FAILED(std::forward<ContextStageLambda>(contextStage)(patternContext.get()));
+    RETURN_IF_FAILED(std::forward<ContextStageLambda>(contextStage)(patternContext.get(), isColored));
```
aballway: On more thought, the lambda is really only saving us from having to check if it's stroke vs fill.  For this case I think it'd be better to pass in a flag (enum or just bool since this'll be the only difference) and inline the lambda stuff here.

msft-Jeyaram: yes but this helps make this more general.
Leaving it as it is.

---
In reply to: [101356532](https://github.com/Microsoft/WinObjC/pull/1952#discussion_r101356532) [](ancestors = 101356532)

msft-Jeyaram:
Ping! :)

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_PatternTests.cpp
DHowett-MSFT: just put this into CFStringCreateWithFormat :)

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1781,20 +1782,23 @@ static HRESULT _CreatePatternBrush(
 
     size_t bitsPerComponent = 8;
     size_t bytesPerRow = 4 * tileSize.size.width;
-    CGBitmapInfo bitmapInfo = kCGImageAlphaPremultipliedFirst | kCGBitmapByteOrder32Big;
+    CGBitmapInfo bitmapInfo = kCGImageAlphaPremultipliedFirst | kCGBitmapByteOrderDefault;
```
rajsesh-msft: How does this work?  We should not be using alphafirst formats since that is not natively supported by d2d.

DHowett-MSFT: @rajsesh-msft alpha first + default = `BGRA` 😸

DHowett-MSFT: However, @msft-Jeyaram: If you want to avoid unnecessary bitmap copies, please use AlphaLast|Big: That is `RGBA`, and won't have the overhead of a format-swizzling copy when we give it to D2D.

DHowett-MSFT: Sorry: premultiplied alpha last | big

msft-Jeyaram: @DHowett-MSFT  updating 😄

PR #1954, opened 2017-02-13
Avoid D2D1Group fill mode conversion for CGPath.

### LINKED ISSUE #1375: [CGD2D] CGPath's default fill mode is Alternate (EO), but CG should default to Winding.
Current fix: "convert" all D2D Geometries to the appropriate fill mode, even if they are already compliant.

Suggested fix: Enlighten CGPath as to its fill mode and provide a CGPath internal function for obtaining a geometry in a different fill mode.

MSFTFox:
A path can avoid the ID2D1GeometryGroup step when filling because a default fill mode has been set.

Fixes #1375


Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -2250,17 +2250,25 @@ void CGContextClearRect(CGContextRef context, CGRect rect) {
...
         auto& state = context->CurrentGState();
         if (drawMode & kCGPathFill) {
             state.fillBrush->SetOpacity(state.alpha);
 
             ComPtr<ID2D1Geometry> geometryToFill;
-            D2D1_FILL_MODE d2dFillMode = (drawMode & kCGPathEOFill) == kCGPathEOFill ? D2D1_FILL_MODE_ALTERNATE : D2D1_FILL_MODE_WINDING;
-            RETURN_IF_FAILED(_CGConvertD2DGeometryToFillMode(geometry.Get(), d2dFillMode, &geometryToFill));
-
+            // If the geometry is a path and has requested the default fill mode we can avoid the fill mode conversion step.
+            if (!isPathGeometry || drawMode == kCGPathEOFill) {
```
rajsesh-msft: drawMode is a flag based on the next line in this method, so you'll need to do the same check here as below.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -414,7 +414,7 @@ inline HRESULT PopEndDraw() {
     template <typename Lambda> // Lambda takes the form HRESULT (*)(CGContextRef, ID2D1DeviceContext*)
     HRESULT Draw(_CGCoordinateMode coordinateMode, CGAffineTransform* additionalTransform, Lambda&& drawLambda);
 
-    HRESULT DrawGeometry(_CGCoordinateMode coordinateMode, ID2D1Geometry* pGeometry, CGPathDrawingMode drawMode);
+    HRESULT DrawGeometry(_CGCoordinateMode coordinateMode, ID2D1Geometry* pGeometry, CGPathDrawingMode drawMode, bool isPathGeometry);
```
rajsesh-msft: Not sure if I like this.  Since the cases where it is "false" you are creating a geometry just before calling DrawGeometry() can you do the conversion there?  More fundamentally, why do we need this conversion for built-in geometry like Rectangle, seems wasteful?

MSFTFox: It's still required since a rectangle geometry is created without a particular fill mode. There's no way to set a particular default for simply an ID2D1RectangleGeometry (Similarly for others). An equivalent comparison would be creating a path, streaming a rectangle geometry into that path, and setting the fill mode on the geometrysink. I think the groups work better for this though.

As far as I know, you can't just assign a fill mode to a rectangle geometry or any normal ID2D1Geometry for that matter and have to do it through a group.

Frameworks/CoreGraphics/CGContext.mm
MSFTFox: I'm not sure what you mean by that. drawMode here is passed in. We already know we have to fill the geometry but if it's a winding fill and a path geometry we're allowed to skip this conversion step since CGPath has already set the fill mode by default.

msft-Jeyaram:
:shipit:

Frameworks/CoreGraphics/CGContext.mm
DHowett-MSFT: But the fill mode for a simple shape like a rectangle doesn't matter, as it has no overlapping sections.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -271,11 +271,29 @@ HRESULT AddGeometryToPathWithTransformation(const ID2D1Geometry* geometry, const
...
+
     RETURN_IF_FAILED(path->ClosePath());
-    path->pathGeometry.CopyTo(pGeometry);
+    if (fillMode == kCGPathEOFill || fillMode == kCGPathEOFillStroke) {
+        ID2D1Geometry* geometry = path->GetPathGeometry();
+        ComPtr<ID2D1Factory> factory;
+        geometry->GetFactory(&factory);
+
+        ComPtr<ID2D1GeometryGroup> geometryGroup;
+        RETURN_IF_FAILED(factory->CreateGeometryGroup(fillMode == kCGPathEOFill ? D2D1_FILL_MODE_ALTERNATE : D2D1_FILL_MODE_WINDING,
```
rajsesh-msft: >kCGPathEOFill [](start = 66, length = 13)

aha!

MSFTFox:
Well the previous check made sure this was always expecting alternate in the ternary. It was just an unnecessary extra check for which fill mode to use.

tests/unittests/CoreGraphics.drawing/CGContextDrawing_FillModeTests.cpp
```diff
@@ -0,0 +1,124 @@
...
+        case kCGPathFillStroke:
+            CGContextFillPath(context);
+            CGContextStrokePath(context);
+            break;
+        case kCGPathEOFill:
+            CGContextEOFillPath(context);
+            break;
+        case kCGPathEOFillStroke:
+            CGContextEOFillPath(context);
+            CGContextStrokePath(context);
```
DHowett-MSFT: so funny story!
Fill clears the path, and stroke will not have a path to stroke.

The sheer joy of using this `fillMode` method is that you can just do...

```
CGContextDrawPath(context, fillMode);
```

it will do the right thing :)

DHowett-MSFT:
All glory to various hypnotoads.

PR #1958, opened 2017-02-14
Vsimporter packaging changes and WinObjC.Packaging proj

memontic-ms:


tools/vsimporter/src/vsimporter.cpp
```diff
@@ -282,15 +290,11 @@ int main(int argc, char* argv[])
   // Initialize global settings
   String binaryDir = sb_dirname(getBinaryLocation());
   sbValidateWithTelemetry(!binaryDir.empty(), "Failed to resolve binary directory.");
+  if (!templatesPath.empty()) {
+    settingsManager.setGlobalVar("VSIMPORTER_TEMPLATES_DIR", joinPaths(getcwd(), templatesPath));
```
bbowman: >getcwd() [](start = 71, length = 8)

don't do this. Its a bad idea to assume that the cwd is the vsimporter directory. I think this should be binaryDir with some relative path?

tools/WinObjC.Packaging/ImportedByAnyDependent.props
```diff
@@ -0,0 +1,15 @@
...
+  PACKAGING PROPS
+  Description: The property definitions specific to dependents of the WinObjC.Packaging package
+  For Use By: A NuProj packager or a packageable VcxProj
+-->
+
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!-- Props for NuGet NuProj packager projects -->
+  <Import Condition="$(MSBuildProjectExtension) == '.nuproj'" Project="WinObjC.Packager.props" />
+
+  <!-- Props for WinObjC VcxProj packageable projects -->
```
bbowman: >VcxProj [](start = 25, length = 7)

maybe add some *** around this to highlight the difference.

tools/WinObjC.Packaging/ImportedByAnyDependent.targets
```diff
@@ -0,0 +1,15 @@
...
+  Description: The target definitions specific to dependents of the WinObjC.Packaging package
+  For Use By: A NuProj packager or a packageable VcxProj
+-->
+
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!-- Targets for NuGet NuProj packager projects -->
+  <Import Condition="$(MSBuildProjectExtension) == '.nuproj'" Project="WinObjC.Packager.targets" />
+
+  <!-- Targets for WinObjC VcxProj packageable projects -->
+  <Import Condition="$(MSBuildProjectExtension) == '.vcxproj'" Project="WinObjC.Packageable.targets" />
```
bbowman: should this key off of PackageId or ShouldPackage or some other nuetizer metadata that knows if it is a packaging project or not? Probably fine as is.

tools/WinObjC.Packaging/Package.Consumption.targets
```diff
@@ -0,0 +1,64 @@
...
+    <_ImportLibsFromPackage Include="$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\*.lib"/>    
+    <_ImportLibsFromPackage Include="$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\$(Configuration)\*.lib"/>
+
+    <_XamlFilesFromPackage Include="$(MSBuildThisFileDirectory)\xaml\**\*.xaml" />
+    <_BundleFilesFromPackage Include="$(MSBuildThisFileDirectory)\bundle\**\*.*" />
+    <_ContentFilesFromPackage Include="$(MSBuildThisFileDirectory)\content\**\*.*" />
+  </ItemGroup>
+
+  <!-- Add references to package files for projects that consume this package -->
+  <ItemGroup Condition="'$(ConfigurationType)' != 'StaticLibrary'">
```
bbowman: >ItemGroup Condition="'$(ConfigurationType)' != 'StaticLibrary' [](start = 3, length = 62)

was thinking and we might actually want to remove this so that the import lib is bundled in a static lib too. The scenario I encountered that made me think of this was if we had a dll project that is straight c++ and just has a project reference to a static lib that is objC. The dll would need to also need to depend on the nuget to get it to link sucessfully but thats a little weird. 


So I think what we want to do is always link but only do the ReferenceCopyLocalPaths if its not a static lib.

tools/WinObjC.Packaging/Package.Consumption.targets
```diff
@@ -0,0 +1,64 @@
...
+  <!-- Add dependency paths needed by projects that consume this package -->
+  <ItemDefinitionGroup>
+    <ClangCompile>
+      <InternalSystemIncludePaths>$(MSBuildThisFileDirectory)\include\;%(InternalSystemIncludePaths);</InternalSystemIncludePaths>
+    </ClangCompile>
+
+    <!-- Add in the link bits only if its a "final" linked product like a dll or exe -->
+    <Link Condition="'$(ConfigurationType)' != 'StaticLibrary'">
+      <AdditionalLibraryDirectories>$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\;$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\$(Configuration)\;%(AdditionalLibraryDirectories);</AdditionalLibraryDirectories>
+    </Link>
```
bbowman: Shouldn't be needed with the above Link include (at least if you pull out the fullpath i think). If its not working try doing:

    
    <Link Include="@(_ImportLibsFromPackage -> %(FullPath))"/>

I think.

tools/WinObjC.Packaging/WinObjC.Packageable.props
```diff
@@ -0,0 +1,25 @@
...
+    <IncludeOutputsInPackage>false</IncludeOutputsInPackage>
+    <IncludeContentInPackage>false</IncludeContentInPackage>
+
+    <!-- WinObjC.Packaging specific properties -->
+    <PresetPackageContents Condition="'$(PresetPackageContents)' == ''">true</PresetPackageContents>
+  </PropertyGroup>
+
+  <ItemDefinitionGroup>
+    <_MainProjectOutput>
+      <IsPackable>$(IsPackable)</IsPackable>
```
bbowman: >IsPackable [](start = 7, length = 10)

Yeah this one (my other comment about keying off of metadata instead of extension)

tools/vsimporter/src/vsimporter.cpp
memontic-ms: My workflow was was running vsimporter.exe from the xcodeproj dir. So it was relative to that location. I think it makes a little more sense to have it relative to the binary but there's a few options here. Preferably it would be run with the full path to the templates, not a relative one.

tools/WinObjC.Packaging/WinObjC.Packageable.targets
```diff
@@ -0,0 +1,88 @@
...
+  <PropertyGroup>
+    <GetPackageContentsDependsOn>$(GetPackageContentsDependsOn);$(ComputeLinkInputsTargets);FindPublicHeaders</GetPackageContentsDependsOn>
+  </PropertyGroup>
+
+  <!-- Add MSBuild outputs into the package -->
+  <Target Name="AddBuiltOutput" BeforeTargets="GetPackageContents"  DependsOnTargets="$(GetPackageContentsDependsOn)" Condition="'$(PresetPackageContents)' == 'true'">
+    <!-- Framework / Static Library Projects -->
+    <ItemGroup Condition="'$(ConfigurationType)' == 'StaticLibrary'">
+      <!-- .lib Files -->
+      <_ImportLibraryFileNames Include="$(OutDir)*.lib" />
```
bbowman: >OutDir [](start = 42, length = 6)

Is this what we want? I was thinking that TargetPath was the more generalized thing or whatever I did for the frameworks ones ...

      <_ImportLibraryFileNames Include="@(Link -> '%(ImportLibrary)')">
      </_ImportLibraryFileNames>


      <PackageFile Include="@(_ImportLibraryFileNames->Distinct())" Condition="Exists('%(FullPath)')">
        <PackagePath>$(PackageContentBasePath)\%(Filename)%(Extension)</PackagePath>
      </PackageFile>


The ImportLibrary metadata on Link is telling it what to make the importLibrary called / where to output it so I think that is the safest? spot? I think OutDir*.lib is going to maybe pick up more than you want if other packages etc dump stuff to the output folder.

tools/WinObjC.Packaging/ImportedByAnyDependent.targets
memontic-ms: The nugetizer properties aren't set at this point.

tools/WinObjC.Packaging/WinObjC.Packageable.targets
```diff
@@ -0,0 +1,88 @@
...
+        <PackagePath>$(PackageContentBasePath)\%(Filename)%(Extension)</PackagePath>
+      </PackageFile>
+      <!-- Resources -->
+      <PackageFile Include="@(SBResourceCopy->'%(FullPath)')" Condition="Exists('%(FullPath)')">
+        <PackagePath>$(PackageResourcesBasePath)\%(Filename)%(Extension)</PackagePath>
+      </PackageFile>
+    </ItemGroup>
+
+    <!-- Bundle Projects -->
+    <ItemGroup Condition="'$(ConfigurationType)' == 'Utility'">
```
bbowman: I dunno what this is... can you make this comment more descriptive? I'm also a little worried that we are overloading the ConfigurationType for this. Do bundle projects actually build C/C++/ObjC code like other .vcxproj files?

tools/WinObjC.Packaging/WinObjC.Packageable.targets
bbowman: Oh I see a little more what you are doing here but its still wrong. You have this under Condition="'$(ConfigurationType)' == 'StaticLibrary'"

but a static lib doesn't have an import lib. An import lib is paired with a dll so you want this bit to go below and to copy the BuiltProjectOutputGroupOutput to this block as the static lib is the main project output for a static library config.

---
In reply to: [100939160](https://github.com/Microsoft/WinObjC/pull/1958#discussion_r100939160) [](ancestors = 100939160)

tools/WinObjC.Packaging/WinObjC.Packager.targets
```diff
@@ -0,0 +1,70 @@
...
+
+  <Target Name="AddSingleConfigBuildOutput"  Returns="@(_PackageContentsForPlat)">
+    <MSBuild Projects="$(SolutionPath)"
+       Targets="GetSolutionConfigurationContents"
+       Properties="Configuration=$(Configuration);Platform=$(Platform)">
+      <Output TaskParameter="TargetOutputs" ItemName="_SolutionConfigurationContents" />
+    </MSBuild>
+
+    <Message Text="AddSingleConfigBuildOutput Platform: $(Platform)" Importance="High"/>
+
```
bbowman: I used this for debugging. Maybe turn this into a more human friendly version with a lower pri if we feel its still nice to keep around.

tools/WinObjC.Packaging/WinObjC.Packageable.targets
memontic-ms: It'd be the .vcxproj created from the WinStore10-Bundle template, which is mimicking the xcode bundle type targets. They shouldn't be building anything. Instead they'll have SBResourceCopy items to include. I'm not clear what your concern is here.

tools/WinObjC.Packaging/WinObjC.Packager.targets
```diff
@@ -0,0 +1,70 @@
...
+    <ItemGroup>
+      <PackageFile Include="$(MSBuildProjectDirectory)\$(PackageId).props" Condition="Exists('$(MSBuildProjectDirectory)\$(PackageId).props')">
+        <PackagePath>build\$(PackageId).props</PackagePath>
+      </PackageFile>
+      <PackageFile Include="$(MSBuildProjectDirectory)\$(PackageId).targets" Condition="Exists('$(MSBuildProjectDirectory)\$(PackageId).targets')">
+        <PackagePath>build\$(PackageId).targets</PackagePath>
+      </PackageFile>
+      <PackageFile Include="$(MSBuildThisFileDirectory)Package.Linker.targets" Condition="Exists('$(MSBuildThisFileDirectory)Package.Linker.targets')">
+        <PackagePath>build\$(PackageId).Linker.targets</PackagePath>
+      </PackageFile>
```
bbowman: Is this the right name? You had the other one as packagable

tools/WinObjC.Packaging/WinObjC.Packageable.targets
memontic-ms: It'd be the .vcxproj created from the WinStore10-Bundle template, which is mimicking the xcode bundle type targets. They shouldn't be building anything. Instead they'll have SBResourceCopy items to include. I'm not clear what your concern is here.

---
In reply to: [100939382](https://github.com/Microsoft/WinObjC/pull/1958#discussion_r100939382) [](ancestors = 100939382)

tools/WinObjC.Packaging/ImportedByAnyDependent.targets
memontic-ms: The nugetizer properties aren't set at this point... But let me double check that.

---
In reply to: [100937477](https://github.com/Microsoft/WinObjC/pull/1958#discussion_r100937477) [](ancestors = 100937477)

tools/vsimporter/src/vsimporter.cpp
memontic-ms: My workflow was was running vsimporter.exe from the xcodeproj dir. So it was relative to that location. I think it makes a little more sense to have it relative to the binary but there's a few options here. Preferably it would be run with the full path to the templates, not a relative one.

---
In reply to: [100936693](https://github.com/Microsoft/WinObjC/pull/1958#discussion_r100936693) [](ancestors = 100936693)

tools/WinObjC.Packaging/WinObjC.Packageable.targets
bbowman: if it should be a .vcxproj at all. Its probably outside the scope of this change but .vcxproj files imply something about what the project is doing (building native code) and that is not the case for this. If its just adding SBResourceCopy items it also doesn't seem like a stand alone project is necessarily correct.

---
In reply to: [100939382](https://github.com/Microsoft/WinObjC/pull/1958#discussion_r100939382) [](ancestors = 100939382)

tools/WinObjC.Packaging/WinObjC.Packager.targets
```diff
@@ -0,0 +1,70 @@
...
+
+  <Target Name="AddBuiltOutput" BeforeTargets="GetPackageContents" Condition="'$(SkipGetPackageContentsForOtherPlatforms)' != 'true' and '$(PresetPackageContents)' == 'true'">
+    <ItemGroup>
+      <PackageFile Include="$(MSBuildProjectDirectory)\$(PackageId).props" Condition="Exists('$(MSBuildProjectDirectory)\$(PackageId).props')">
+        <PackagePath>build\$(PackageId).props</PackagePath>
+      </PackageFile>
+      <PackageFile Include="$(MSBuildProjectDirectory)\$(PackageId).targets" Condition="Exists('$(MSBuildProjectDirectory)\$(PackageId).targets')">
+        <PackagePath>build\$(PackageId).targets</PackagePath>
+      </PackageFile>
+      <PackageFile Include="$(MSBuildThisFileDirectory)Package.Linker.targets" Condition="Exists('$(MSBuildThisFileDirectory)Package.Linker.targets')">
```
memontic-ms: >Linker [](start = 133, length = 6)

This should be "Consumption"

tools/WinObjC.Tools/vsimporter-templates/packageable.project.json
```diff
@@ -0,0 +1,21 @@
+{
+  "dependencies": {
+    "WinObjC.Packaging": "*",
+    "WinObjC.Frameworks": "*",
```
bbowman: should just need Packaging and Frameworks I think?

bbowman: this is the one that will end up in the imported project's directory right?

---
In reply to: [100940627](https://github.com/Microsoft/WinObjC/pull/1958#discussion_r100940627) [](ancestors = 100940627)

tools/WinObjC.Tools/vsimporter-templates/project.json
```diff
@@ -0,0 +1,20 @@
+{
+  "dependencies": {
+    "WinObjC.Frameworks": "*",
```
bbowman: >"WinObjC.Frameworks": "*", [](start = 4, length = 26)

should just need Frameworks I think?

tools/WinObjC.Packaging/WinObjC.Packageable.targets
memontic-ms: From what I discussed with Artem, there is no VS equivalent of a bundle project so he settled on this.

---
In reply to: [100940481](https://github.com/Microsoft/WinObjC/pull/1958#discussion_r100940481) [](ancestors = 100940481,100939382)

tools/WinObjC.Tools/vsimporter-templates/project.json
bbowman: do we want to just go ahead and give them packaging anyway? Unless they have a .nuproj its harmless and would make it easier for them to add after the fact?

---
In reply to: [100940735](https://github.com/Microsoft/WinObjC/pull/1958#discussion_r100940735) [](ancestors = 100940735)

tools/WinObjC.Packaging/WinObjC.Packageable.targets
bbowman: lets file an issue to think about that some more. I'm not convinced we want this approach long term.

---
In reply to: [100940736](https://github.com/Microsoft/WinObjC/pull/1958#discussion_r100940736) [](ancestors = 100940736,100940481,100939382)

tools/WinObjC.Tools/vsimporter-templates/project.json
memontic-ms: It's just one more thing to go wrong, imo. I'm not totally against the idea though.

---
In reply to: [100940824](https://github.com/Microsoft/WinObjC/pull/1958#discussion_r100940824) [](ancestors = 100940824,100940735)

tools/WinObjC.Tools/vsimporter-templates/packageable.project.json
memontic-ms: Yes.

---
In reply to: [100940695](https://github.com/Microsoft/WinObjC/pull/1958#discussion_r100940695) [](ancestors = 100940695,100940627)

tools/WinObjC.Packaging/WinObjC.Packager.targets
memontic-ms: Not such what you're pointing at here

---
In reply to: [100939995](https://github.com/Microsoft/WinObjC/pull/1958#discussion_r100939995) [](ancestors = 100939995)

tools/WinObjC.Tools/vsimporter-templates/WinStore10-Package/Package.targets
```diff
@@ -0,0 +1,14 @@
+﻿<?xml version="1.0" encoding="utf-8" standalone="no"?>
+
+<!--
+  PACKAGE TARGETS
+  Description: The target definitions specific to $projectname$
+  For Use By: A VcxProj that depends on the $projectname$ package
+-->
+
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildThisFileDirectory)\$(MSBuildThisFileName).Consumption.targets" Condition="Exists('$(MSBuildThisFileDirectory)\$(MSBuildThisFileName).Consumption.targets')" />
```
bbowman: ><Import Project="$(MSBuildThisFileDirectory)\$(MSBuildThisFileName).Consumption.targets" Condition="Exists('$(MSBuildThisFileDirectory)\$(MSBuildThisFileName).Consumption.targets')" /> [](start = 2, length = 184)

This line confuses me. Isn't that what this file is doing?

memontic-ms: Bringing in the package's Consumption.targets file, which should be right next to it in the package. Nuget does something similar but they had some special syntax to get the package's installed path.

---
In reply to: [100941133](https://github.com/Microsoft/WinObjC/pull/1958#discussion_r100941133) [](ancestors = 100941133)

tools/vsimporter/src/vsimporter.cpp
bbowman: vsimporter.exe and templates will eventually live in a Tools package in some random not super easy to get to path. 

Full path to the templates is best if thats knowable ahead of time. Was thinking that at least we can know the relative directory structure between vsimporter.exe and the templates and can get the .exe path at runtime

---
In reply to: [100940407](https://github.com/Microsoft/WinObjC/pull/1958#discussion_r100940407) [](ancestors = 100940407,100936693)

bbowman:
:clock1:

tools/WinObjC.Tools/vsimporter-templates/WinStore10-Package/Package.targets
bbowman: Perhaps I don't understand what Package.targets is here but I thought this is what is going to be the consumption targets anyway. It seems you are making another level of indirection and i"m not sure why.

---
In reply to: [100941520](https://github.com/Microsoft/WinObjC/pull/1958#discussion_r100941520) [](ancestors = 100941520,100941133)

tools/WinObjC.Packaging/WinObjC.Packageable.targets
memontic-ms: Updated in latest iteration (5). Please confirm.

---
In reply to: [100939635](https://github.com/Microsoft/WinObjC/pull/1958#discussion_r100939635) [](ancestors = 100939635,100939160)

tools/WinObjC.Packaging/Package.Default.props
```diff
@@ -0,0 +1,64 @@
...
+    <_ImportLibsFromPackage Include="$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\*.lib"/>    
+    <_ImportLibsFromPackage Include="$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\$(Configuration)\*.lib"/>
+
+    <_XamlFilesFromPackage Include="$(MSBuildThisFileDirectory)\xaml\**\*.xaml" />
+    <_BundleFilesFromPackage Include="$(MSBuildThisFileDirectory)\bundle\**\*.*" />
+    <_ContentFilesFromPackage Include="$(MSBuildThisFileDirectory)\content\**\*.*" />
+  </ItemGroup>
+
+  <!-- Add references to package files for projects that consume this package -->
+  <ItemGroup Condition="'$(ConfigurationType)' != 'StaticLibrary'">
```
bbowman: my comment from Iteration 2 stands. I think we want to do part of this for static libs too.

tools/WinObjC.Packaging/Package.Default.props
```diff
@@ -0,0 +1,64 @@
...
+    <!-- Add references for bundle files -->
+    <ReferenceCopyLocalPaths Include="@(_BundleFilesFromPackage)">
+      <DestinationSubDirectory>%(RecursiveDir)</DestinationSubDirectory>
+    </ReferenceCopyLocalPaths>
+  </ItemGroup>
+
+  <!-- Add dependency paths needed by projects that consume this package -->
+  <ItemDefinitionGroup>
+    <ClangCompile>
+      <InternalSystemIncludePaths>$(MSBuildThisFileDirectory)\include\;%(InternalSystemIncludePaths);</InternalSystemIncludePaths>
```
bbowman: >InternalSystemIncludePaths [](start = 7, length = 26)

This should probably be IncludePaths not InternalSystemIncludePaths ... We use the latter in frameworks because Foundattion et al are "system" frameworks unlike what a 3rd party component would be. We would need to xmlPoke or something fancy here to allow a flag on the production side determine this when we have the frameworks consume the packaging nuget.

tools/WinObjC.Packaging/Package.Default.props
```diff
@@ -0,0 +1,64 @@
...
+    </ClangCompile>
+
+    <!-- Add in the link bits only if its a "final" linked product like a dll or exe -->
+    <Link Condition="'$(ConfigurationType)' != 'StaticLibrary'">
```
bbowman: >Link [](start = 5, length = 4)

Still valid comments here about this being not needed if the other part is done with full paths. Probably better that way so that the metadata here isn't as polluted and harder for the user to mess up

tools/WinObjC.Packaging/WinObjC.Packageable.props
```diff
@@ -0,0 +1,27 @@
...
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup Label="Globals">
+    <!-- NuGet specific properties -->
+    <IncludeOutputsInPackage>false</IncludeOutputsInPackage>
+    <IncludeContentInPackage>false</IncludeContentInPackage>
+
+    <!-- Packaging defaults from WinObjC.Packaging -->
+    <IncludeDefaultPackageContents Condition="'$(IncludeDefaultPackageContents)' == ''">true</IncludeDefaultPackageContents>
+    <!-- If the package is intended to only include WinRT components set IncludeStaticLibraryInPackage to false -->
+    <IncludeStaticLibraryInPackage Condition="'$(IncludeStaticLibraryInPackage)' == ''">true</IncludeStaticLibraryInPackage>
```
bbowman: nit: add a new line in between these. The comment doesn't need to mention WinRT. And I might switch it to be off by default since I think that is going to be the common case.

tools/WinObjC.Packaging/WinObjC.Packageable.targets
```diff
@@ -0,0 +1,88 @@
...
+      <PackageFile Include="@(_IncludeFiles)" Condition="Exists('%(FullPath)')">
+        <PackagePath>$(PackageHeaderBasePath)\%(Filename)%(Extension)</PackagePath>
+      </PackageFile>
+      <!-- Resources -->
+      <PackageFile Include="@(SBResourceCopy->'%(FullPath)')" Condition="Exists('%(FullPath)')">
+        <PackagePath>$(PackageResourcesBasePath)\%(Filename)%(Extension)</PackagePath>
+      </PackageFile>
+    </ItemGroup>
+
+    <!-- WinRT Projects -->
```
bbowman: >WinRT [](start = 9, length = 5)

normal dlls could go through this path too and I think then you'd want to include the public headers here too.

The weird / hard part is going to be if its a WinRT component, we do want to include the generated WinRT headers instead of the normal headers (for C++\Cx consumptions since it doesn't technically go through the .winmd). 

We can use the winmd presense of an indicator to look for these though. I'd talk to muktesh / pradip about if / how objc2winmd produces these.

tools/WinObjC.Packaging/WinObjC.Packageable.targets
```diff
@@ -0,0 +1,88 @@
...
+      <!-- .dll -->
+      <_BuiltProjectOutputImportLib Include="@(BuiltProjectOutputGroupKeyOutput -> '%(FinalOutputPath)')" />
+      <PackageFile Include="@(_BuiltProjectOutputImportLib->Distinct())" Condition="Exists('%(FullPath)')">
+        <PackagePath>$(PackageContentBasePath)\%(Filename)%(Extension)</PackagePath>
+      </PackageFile>
+      <!-- .winmd -->
+      <PackageFile Include="@(WinMDFullPath)" Condition="'@(WinMDFullPath)' != '' And Exists('@(WinMDFullPath)')">
+        <PackagePath>$(PackageContentBasePath)\%(Filename)%(Extension)</PackagePath>
+      </PackageFile>
+      <!-- Resources -->
```
bbowman: >Resources [](start = 11, length = 9)

the resources bit look common? could we factor this out different? maybe not cause of bundle projects?

tools/WinObjC.Packaging/WinObjC.Packager.targets
```diff
@@ -0,0 +1,79 @@
...
+      <PackageFile Include="$(MSBuildThisFileDirectory)Package.targets" Condition="Exists('$(MSBuildThisFileDirectory)Package.targets')">
+        <PackagePath>build\$(PackageId).targets</PackagePath>
+      </PackageFile>
+      <PackageFile Include="$(MSBuildThisFileDirectory)Package.Default.props" Condition="Exists('$(MSBuildThisFileDirectory)Package.Default.props')">
+        <PackagePath>build\$(PackageId).Default.props</PackagePath>
+      </PackageFile>
+      <PackageFile Include="$(MSBuildThisFileDirectory)Package.Default.targets" Condition="Exists('$(MSBuildThisFileDirectory)Package.Default.targets')">
+        <PackagePath>build\$(PackageId).Default.targets</PackagePath>
+      </PackageFile>
+      <PackageFile Include="$(MSBuildProjectDirectory)\$(PackageId).Custom.targets" Condition="Exists('$(MSBuildProjectDirectory)\$(PackageId).Custom.targets')">
```
bbowman: >MSBuildProjectDirectory [](start = 105, length = 23)

can you put a comment for the custom ones about them living in the project folder. Just scanning this its hard to see this one is special if someone comes along later and fixes something up

bbowman:
This is kickass. Love it.

tools/WinObjC.Packaging/Package.Default.props
```diff
@@ -0,0 +1,65 @@
...
+      <AdditionalLibraryDirectories>$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\;$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\$(Configuration)\;%(AdditionalLibraryDirectories);</AdditionalLibraryDirectories>
+    </Link>
```
bbowman: This can be removed since you are adding things to Link directly with the full path now.

PR #1966, opened 2017-02-14
Bbowman/packaging merge fix

bbowman:
Pull in latest develop and fix issues that came up from the merge.


bbowman:
The last commit is the interesting one (fixing merge conflicts). Dunno the best way to review it.

include/Platform/Universal Windows/UWP/RTHelpers.h
```diff
@@ -51,8 +51,8 @@
 #include <UWP/interopBase.h>
 
 #if __has_feature(objc_arc)
-#define AUTORELEASE(x)
-#define RETAIN(x)
+#define AUTORELEASE(x) x
```
memontic-ms: There's an extra x floating around on the two lines. Don't see that in develop or packaging.

build/Tests/UnitTests/Accelerate/Accelerate.UnitTests.vcxproj
```diff
@@ -1,5 +1,5 @@
-<?xml version="1.0" encoding="utf-8"?>
-<Project DefaultTargets="Build" ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
memontic-ms: Might want to roll these all back to 14.0 for now?

include/Platform/Universal Windows/UWP/RTHelpers.h
bbowman: They are the "return" values here. some of these guys are supposed to return themselves so you can [[foo retain] bar ] which was creating errors somewhere. I was surprised it worked before.

build/Tests/UnitTests/Accelerate/Accelerate.UnitTests.vcxproj
bbowman: done

include/Platform/Universal Windows/UWP/RTHelpers.h
memontic-ms: Sounds good. I was concerned it was a merge error.

memontic-ms:
:shipit:

PR #1980, opened 2017-02-16
Re-enable CancelResume test without flakey data size comparison checks

aballway:
Checking data sizes in NSURL::NSURLSession_DownloadTaskWithURL_WithCancelResume would intermittently cause failures, so re-enabling this test with just state checks.  This is not going to catch as many types of regressions as the previous implementation but it shouldn't fail for passing code either.


ms-jihua:
nit: should we add a final check for the size when the download is finished?

aballway:
we're not waiting for it to finish right now, and adding a wait for that might bring back some of the flakeyness

---
In reply to: [280432942](https://github.com/Microsoft/WinObjC/pull/1980#issuecomment-280432942) [](ancestors = 280432942)

PR #1988, opened 2017-02-16
Annotate CoreText and CoreFoundation apis that are not in plan

rajsesh-msft:
This is just internal book keeping for APIs that we don't plan to implement in near future.  Subject to change of course.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1988)

### DIFF
--- a/Frameworks/CoreFoundation/String.subproj/CFString.c
+++ b/Frameworks/CoreFoundation/String.subproj/CFString.c
@@ -6720,21 +6720,19 @@ void CFShowStr(CFStringRef str) {
 #undef HANGUL_NCOUNT
 
 /**
- @Status Stub
- @Notes
+ @Status NotInPlan
+ @Notes String hypenation is non-trivial and not used widely.  For regular text layout, Coretext provides hypenation through Dwrite.
 */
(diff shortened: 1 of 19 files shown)


Frameworks/CoreFoundation/String.subproj/CFString.c
```diff
@@ -6720,21 +6720,19 @@ void CFShowStr(CFStringRef str) {
 #undef HANGUL_NCOUNT
 
 /**
- @Status Stub
- @Notes
+ @Status NotInPlan
+ @Notes String hypenation is non-trivial and not used widely.  For regular text layout, Coretext provides hypenation through Dwrite.
```
ms-jihua: nit:hyphenation

Frameworks/CoreText/CTRun.mm
```diff
@@ -303,10 +303,10 @@ CGAffineTransform CTRunGetTextMatrix(CTRunRef run) {
 }
 
 /**
- @Status Stub
- @Notes
+ @Status NotInPlan
+ @Notes this would require us to move to using briged type implementation, seems of little value at this point
```
ms-jihua: nit:bridged

Frameworks/CoreText/CTLine.mm
```diff
@@ -265,7 +266,7 @@ CFRange CTLineGetStringRange(CTLineRef line) {
 }
 
 /**
- @Status Stub
+ @Status Caveat
```
aballway: Shouldn't this remain Stub?

PR #2001, opened 2017-02-17
Fix #1999 'MoveMe' button in HelloUI moves erratically

### LINKED ISSUE #1999: 'MoveMe' button in HelloUI moves erratically
Launch HelloUI and drag the 'MoveMe' button around; it jumps around and ultimately disappears from the window.  This _appears_ to have regressed with 1ec1a845a; reverting that commit fixes the issue.

yiyang-msft:
We removed the places which set curPanList in earlier change, which is used to support UIScollivew for gesture chaining but now that is not needed. 

but we keep curPanList around which didn't get set anymore. But it was still used in setTranslation in PanGesture. So when app resets the translation, we failed to do that because curPanList is nil. The fix is to remove enumerating the curPanList and set the translation to self, which is correct way of doing it.


jaredhms:
:shipit:

Issue #1998, opened 2017-02-17
CGBitmapContext should support 0 bytesPerRow


msft-Jeyaram:

DHowett-MSFT:
@msft-Jeyaram I have work to fix this here: https://github.com/DHowett-MSFT/WinObjC/tree/201701-0bpp

It has not been submitted for review.

msft-Jeyaram:
@DHowett-MSFT  awesome!!

PR #2002, opened 2017-02-17
Allow a 0b stride bitmap context to calculate its own stride.

### LINKED ISSUE #1997: CGBitmapContextCreate() regression when using latest HEAD of master branch.
We use some code like the following to resize a UIImage using CoreGraphics.

When using a build of winobjc from the HEAD of master (ffe05fa6e5c7819a6cb2c9105ed9694245656842) this code no longer works and the created `newImage` is nil.

Picking through it, it seems `CGBitmapContextCreate()` returns a nil `CGContextRef`.

This code works fine using the 0.2 170201 1st Feb release.

```
UIImage *image = [UIImage imageNamed:@"someImage"];
CGSize size = CGSizeMake(64.0, 64.0);
CGFloat scaleFactor = [UIScreen mainScreen].scale;
CGRect newRect = CGRectIntegral(CGRectMake(0, 0, size.width, size.height));
CGImageRef imageRef = image.CGImage;
	
CGContextRef bitmap = CGBitmapContextCreate(NULL,
                                                newRect.size.width * scaleFactor,
                                                newRect.size.height * scaleFactor,
						CGImageGetBitsPerComponent(imageRef),
						0,
						CGImageGetColorSpace(imageRef),
						CGImageGetBitmapInfo(imageRef));
	
CGContextScaleCTM(bitmap, scaleFactor, scaleFactor);

CGContextDrawImage(bitmap, newRect, imageRef);
	
CGImageRef newImageRef = CGBitmapContextCreateImage(bitmap);
UIImage *newImage = [UIImage imageWithCGImage:newImageRef scale:scaleFactor orientation:UIImageOrientationUp];
	
CGContextRelease(bitmap);
CGImageRelease(newImageRef);
```

DHowett-MSFT:
This adds a few more failure cases for bad bitmap contexts,
and changes format imputation in the CG<->WIC mapper.

Fixes #1997, #1998.

### DIFF
--- a/Frameworks/CoreGraphics/CGImage.mm
+++ b/Frameworks/CoreGraphics/CGImage.mm
@@ -710,6 +710,36 @@ NSData* _CGImageRepresentation(CGImageRef image, REFGUID guid, float quality) {
     return nil;
 }
 
+size_t _CGImageImputeBitsPerPixelFromFormat(CGColorSpaceRef colorSpace, size_t bitsPerComponent, CGBitmapInfo bitmapInfo) {
+    unsigned int alphaInfo = bitmapInfo & kCGBitmapAlphaInfoMask;
+    unsigned int byteOrder = bitmapInfo & kCGBitmapByteOrderMask;
+
+    // Try byte order first: The user can specify 32 or 16 directly.
+    switch (byteOrder) {
+        case kCGBitmapByteOrder32Little:
+        case kCGBitmapByteOrder32Big:
+            return 32;
+        case kCGBitmapByteOrder16Little:
+        case kCGBitmapByteOrder16Big:
+            return 16;
+    }
+
+    // Otherwise, try to figure out how many components there are.
+    size_t nComponents = CGColorSpaceGetNumberOfComponents(colorSpace);
+    switch (alphaInfo) {
+        case kCGImageAlphaNoneSkipFirst:
+        case kCGImageAlphaPremultipliedFirst:
+        case kCGImageAlphaFirst:
+        case kCGImageAlphaNoneSkipLast:
+        case kCGImageAlphaPremultipliedLast:
+        case kCGImageAlphaLast:
+            nComponents += 1;
+            break;
+    }
+
+    return (bitsPerComponent * nComponents);
+}
+
 // CG packed format key
 //  |Color  |bits/px|CGBitmapInfo   |
 //  |-------|-------|---------------|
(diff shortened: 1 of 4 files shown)


Frameworks/CoreGraphics/CGImage.mm
```diff
@@ -710,6 +710,36 @@ CGImageRef _CGImageLoadImageWithWICDecoder(REFGUID decoderCls, void* bytes, int
...
+        case kCGImageAlphaPremultipliedFirst:
+        case kCGImageAlphaFirst:
+        case kCGImageAlphaNoneSkipLast:
+        case kCGImageAlphaPremultipliedLast:
+        case kCGImageAlphaLast:
+            nComponents += 1;
+            break;
+    }
+
+    return (bitsPerComponent * nComponents);
```
DHowett-MSFT: This _can_ return something weird like 20 (5bpp + alpha) or 15 (5bpp no alpha) when it should return 16 for both cases.

Seeing as we don't support 16bpp contexts, however ...

aballway: nit: Add a comment in the code w/ todo#

Frameworks/CoreGraphics/CGImage.mm
```diff
@@ -710,6 +710,36 @@ CGImageRef _CGImageLoadImageWithWICDecoder(REFGUID decoderCls, void* bytes, int
...
+            return 32;
+        case kCGBitmapByteOrder16Little:
+        case kCGBitmapByteOrder16Big:
+            return 16;
+    }
+
+    // Otherwise, try to figure out how many components there are.
+    size_t nComponents = CGColorSpaceGetNumberOfComponents(colorSpace);
+    switch (alphaInfo) {
+        case kCGImageAlphaNoneSkipFirst:
```
aballway: are there cases where we don't have alpha to add? Maybe add those at end

tests/unittests/CoreGraphics/CGBitmapContextTests.mm
```diff
@@ -288,3 +288,43 @@ void _TestPixelFormat(const CGRect bounds, const CGBitmapInfo info, DWORD expect
...
+
+    {
+        // Provided 2 bytes per row, requested 32bpp, but using 3-component no alpha = 24bpp (3 byte/pixel)
+        // Context imputes format to be 32, but provided values should not match.
+        auto context = woc::MakeStrongCF<CGContextRef>(CGBitmapContextCreateWithData(nullptr, 1, 1, 8, 2, colorSpace, kCGImageAlphaNone | kCGBitmapByteOrder32Big, nullptr, nullptr));
+        ASSERT_EQ(nullptr, context);
+    }
+}
+
+TEST(CGBitmapContext, ImputeStrideRGBA) {
```
msft-Jeyaram: could you add a grayscale ones?
16s?

Frameworks/CoreGraphics/CGImage.mm
```diff
@@ -710,6 +710,36 @@ CGImageRef _CGImageLoadImageWithWICDecoder(REFGUID decoderCls, void* bytes, int
     return nil;
 }
 
+size_t _CGImageImputeBitsPerPixelFromFormat(CGColorSpaceRef colorSpace, size_t bitsPerComponent, CGBitmapInfo bitmapInfo) {
```
rajsesh-msft: nit: could also be static

Frameworks/CoreGraphics/CGImage.mm
```diff
@@ -710,6 +710,36 @@ CGImageRef _CGImageLoadImageWithWICDecoder(REFGUID decoderCls, void* bytes, int
...
+    unsigned int byteOrder = bitmapInfo & kCGBitmapByteOrderMask;
+
+    // Try byte order first: The user can specify 32 or 16 directly.
+    switch (byteOrder) {
+        case kCGBitmapByteOrder32Little:
+        case kCGBitmapByteOrder32Big:
+            return 32;
+        case kCGBitmapByteOrder16Little:
+        case kCGBitmapByteOrder16Big:
+            return 16;
```
rajsesh-msft: Given that we don't support 16 bit contexts, should we failfast or early error here?

DHowett-MSFT: We will failfast/error out later for this, when it comes time to make the context. This table/function is used for both CGImage and CGBitmapContext, and CGImage supports a wider range of formats!

Frameworks/CoreGraphics/CGImage.mm
DHowett-MSFT: This is used in CGBitmapContext :)

tests/unittests/CoreGraphics/CGBitmapContextTests.mm
DHowett-MSFT: We cannot create these contexts, so there is no point in imputing their formats.

Frameworks/CoreGraphics/CGImage.mm
DHowett-MSFT: there are such cases; they have been elided from the table.

PR #2020, opened 2017-02-22
Vsimporter won't generate packaging project if solution only contains an app

### LINKED ISSUE #2014: VSImporter is overly aggressive in trying to package solutions
It shouldn't do this if the project is an app.

memontic-ms:
Fixes #2014

### DIFF
--- a/tools/vsimporter/src/SBProject.cpp
+++ b/tools/vsimporter/src/SBProject.cpp
@@ -163,6 +163,17 @@ bool SBProject::checkTargetCycle(const PBXTarget* target)
   }
 }
 
+bool SBProject::containsPackagebleProject()
+{
+  for (auto target : m_existingTargets) {
+    TargetProductType productType = target.second->getProductType();
+    if (productType == TargetStaticLib || productType == TargetBundle || target.first->getType() == "PBXAggregateTarget") {
+      return true;
+    }
+  }
+  return false;
+}
+
 void SBProject::selectBuildConfigurations(const StringSet* configNames)
 {
   // Use previously-selected configurations, if possible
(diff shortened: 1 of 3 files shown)


tools/vsimporter/src/SBProject.cpp
```diff
@@ -163,6 +163,17 @@ bool SBProject::checkTargetCycle(const PBXTarget* target)
   }
 }
 
+bool SBProject::containsPackagebleProject()
+{
+  for (auto target : m_existingTargets) {
+    TargetProductType productType = target.second->getProductType();
+    if (productType == TargetStaticLib || productType == TargetBundle || target.first->getType() == "PBXAggregateTarget") {
```
bbowman: This seems wrong. I think we should try to figure out the "end product" instead of just containing a packagable thing.

bbowman:
@memontoc-ms you should kick off a CI build

tools/vsimporter/src/SBProject.cpp
memontic-ms: What do you mean? At import we can only tell what kind of xcode targets they are trying to import.

---
In reply to: [102370117](https://github.com/Microsoft/WinObjC/pull/2020#discussion_r102370117) [](ancestors = 102370117)

memontic-ms:
> 

@bbowman Similar logic

---
Refers to: tools/vsimporter/src/SBProject.cpp:340 in e4b2e39. [](commit_id = e4b2e39e1ddc09d3596a788e59614e0736e8d72b, deletion_comment = False)

bbowman:
:shipit:

PR #2022, opened 2017-02-22
Properly support RTL text drawing in CoreText

### LINKED ISSUE #1932: Properly support RTL languages in core text
https://github.com/Microsoft/WinObjC/blob/develop/Frameworks/CoreGraphics/CGContext.mm#L1906, when drawing glyphs individually, we won't draw the RTL languages correctly.   Should use IDWriteFactory4::ComputeGlyphOrigins to compute the origins of individual glyphs.

aballway:
Now properly draws RTL text with support for text matrix manipulation.
Change CTLineGetOffsetForStringIndex and CTFramesetterCreateFrame to remove assumption of LTR text.

Fixes #1932

### DIFF
--- a/Frameworks/CoreGraphics/CGContext.mm
+++ b/Frameworks/CoreGraphics/CGContext.mm
@@ -2089,6 +2089,12 @@ HRESULT __CGContext::DrawGlyphRuns(GlyphRunData* glyphRuns, size_t runsCount, bo
             // First glyph's origin is at the given relative position for the glyph run
             CGPoint runningPosition{ glyphRuns[i].relativePosition.x, std::round(glyphRuns[i].relativePosition.y) };
             for (size_t j = 0; j < run->glyphCount; ++j) {
+                if (run->bidiLevel & 1) {
+                    // Translate position of glyph by advance
+                    // Need to translate each glyph by its own advance because it's RTL
+                    runningPosition.x -= run->glyphAdvances[j];
+                }
+
                 // Invert position by text transformation
                 CGPoint transformedPosition = CGPointApplyAffineTransform(runningPosition, invertedTextTransformation);
 
@@ -2097,10 +2103,13 @@ HRESULT __CGContext::DrawGlyphRuns(GlyphRunData* glyphRuns, size_t runsCount, bo
                 positions[j] = DWRITE_GLYPH_OFFSET{ transformedPosition.x + run->glyphOffsets[j].advanceOffset,
                                                     std::round(transformedPosition.y + run->glyphOffsets[j].ascenderOffset) };
 
-                // Translate position of next glyph by current glyph's advance
-                runningPosition.x += run->glyphAdvances[j];
+                if ((run->bidiLevel & 1) == 0) {
+                    // Translate position of next glyph by current glyph's advance
+                    runningPosition.x += run->glyphAdvances[j];
+                }
             }
 
+            // Even if text is RTL set bidiLevel to 0 to draw in correct orientation for proper transformation by glyph
             auto transformedGlyphRun = std::make_shared<DWRITE_GLYPH_RUN>(DWRITE_GLYPH_RUN{ run->fontFace,
                                                                                             run->fontEmSize,
                                                                                             run->glyphCount,
@@ -2108,7 +2117,7 @@ HRESULT __CGContext::DrawGlyphRuns(GlyphRunData* glyphRuns, size_t runsCount, bo
                                                                                             zeroAdvances.get(),
                                                                                             positions.data(),
                                                                                             run->isSideways,
-                                                                                            run->bidiLevel });
+                                                                                            0 });
             createdRuns.emplace_back(transformedGlyphRun);
             runs.emplace_back(GlyphRunData{ transformedGlyphRun.get(), CGPointZero, glyphRuns[i].attributes });
         }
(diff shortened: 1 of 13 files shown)


Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -2089,6 +2089,12 @@ void CGContextSetPatternPhase(CGContextRef context, CGSize phase) {
             // First glyph's origin is at the given relative position for the glyph run
             CGPoint runningPosition{ glyphRuns[i].relativePosition.x, std::round(glyphRuns[i].relativePosition.y) };
             for (size_t j = 0; j < run->glyphCount; ++j) {
+                if (run->bidiLevel & 1) {
```
DHowett-MSFT: nit: Since we use this in two different places, but with flipped senses, it might be reasonable to abstract it out into a helper macro or inlinable function. `if(!RUN_IS_BIDI(...))` seems clearer than that we currently have.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -2097,18 +2103,21 @@ void CGContextSetPatternPhase(CGContextRef context, CGSize phase) {
...
 
-                // Translate position of next glyph by current glyph's advance
-                runningPosition.x += run->glyphAdvances[j];
+                if ((run->bidiLevel & 1) == 0) {
+                    // Translate position of next glyph by current glyph's advance
+                    runningPosition.x += run->glyphAdvances[j];
+                }
             }
 
+            // Even if text is RTL set bidiLevel to 0 to draw in correct orientation for proper transformation by glyph
```
DHowett-MSFT: nit: comment "we have compensated for the RTL glyph positions, so we don't need DWrite to do it for us."

Frameworks/CoreText/CTLine.mm
```diff
@@ -370,15 +374,22 @@ CGFloat CTLineGetOffsetForStringIndex(CTLineRef lineRef, CFIndex charIndex, CGFl
                                 run->_stringIndices.begin() - 1;
 
                     if (index >= 0) {
-                        ret = std::accumulate(run->_dwriteGlyphRun.glyphAdvances, run->_dwriteGlyphRun.glyphAdvances + index, ret);
+                        if (run->_dwriteGlyphRun.bidiLevel & 1) {
+                            ret += std::accumulate(run->_dwriteGlyphRun.glyphAdvances,
+                                                   run->_dwriteGlyphRun.glyphAdvances + index,
+                                                   run->_relativeXOffset,
+                                                   std::minus<CGFloat>());
```
DHowett-MSFT: lol accumulate

Frameworks/CoreText/DWriteWrapper_CoreText.mm
```diff
@@ -505,7 +509,10 @@ HRESULT STDMETHODCALLTYPE GetPixelsPerDip(_In_opt_ void* clientDrawingContext, _
             line->_strRange.length = stringRange;
             line->_glyphCount = glyphCount;
             line->_relativeXOffset = static_cast<_CTRun*>(line->_runs[0])->_relativeXOffset;
-            line->_relativeYOffset = static_cast<_CTRun*>(line->_runs[0])->_relativeYOffset;
+            if (static_cast<_CTRun*>([line->_runs objectAtIndex:0])->_dwriteGlyphRun.bidiLevel & 1) {
```
DHowett-MSFT: nit: cache `line->_runs[0]` to avoid double (or triple!) dynamic dispatch

Frameworks/include/CoreTextInternal.h
```diff
@@ -78,7 +78,6 @@ inline void _SafeRelease(T** p) {
 @public
     CFRange _strRange;
     CGFloat _relativeXOffset;
-    CGFloat _relativeYOffset;
```
DHowett-MSFT: no more relative y offset!
how uh
how do we handle top-to-bottom scripts?

aballway: We don't currently support TtB scripts.

Frameworks/CoreText/CTLine.mm
```diff
@@ -370,15 +374,22 @@ CGFloat CTLineGetOffsetForStringIndex(CTLineRef lineRef, CFIndex charIndex, CGFl
...
+                        } else {
+                            ret += std::accumulate(run->_dwriteGlyphRun.glyphAdvances,
+                                                   run->_dwriteGlyphRun.glyphAdvances + index,
+                                                   run->_relativeXOffset);
+                        }
                     }
```
ms-jihua: nit: this is probably an awful idea, but you _could_ use a ternary for the last parameter instead of having the if/else

aballway: I tried that some other place but it _really_ didn't like it, can't remember why though.

rajsesh-msft:
Given #2027, are we sure nothing really regressed here?

DHowett-MSFT:
right: it would behoove us to avoid merging anything until 2027 is merged with test fixes.

rajsesh-msft:
The recommendation in the issue #1932 is to use ComputeGlyphOrigins but we decided to roll out our own implementation instead?

rajsesh-msft:
ComputeGlyphOrigins is not available on windows < anniversary update, so we can't use it here.  Will wait for #2027

aballway:
Tests passed post #2027, please take another look 🍏

aballway:
Rebased on top of develop to resolve merge conflict, tests passed again.

PR #2035, opened 2017-02-22
ObjC2Winmd binary update: 

mukhole:
Added support for containers with heterogeneous types


pradipd:
:shipit:

pradipd:
What caused the projections binaries to be updated?

pradipd:
nevermind.

---
In reply to: [281839514](https://github.com/Microsoft/WinObjC/pull/2035#issuecomment-281839514) [](ancestors = 281839514)

mukhole:
I changed the header RTHelpers.h which is used by all the projection binaries.

PR #2088, opened 2017-02-24
Fix build issues with bundle project templates

### LINKED ISSUE #2044: Vsimporter needs to add AppContainerApplication to Bundle projects
`<AppContainerApplication>true</AppContainerApplication>`

needs added to generated projects, otherwise nuget gets confused on the TFM since the sdk appcontainer .props/.targets set it to UAP but old school win32 apps don't have one.

The **real** fix here would be to not use a .vcxproj for these at all though as they don't build native code as implied by .vcxproj ... .msbuildproj is likely more appropriate.

memontic-ms:
fixes #2044
Bonus: removes bundle project's need for the WinObjC.Language package

### DIFF
--- a/tools/WinObjC.Packaging/WinObjC.Packageable.targets
+++ b/tools/WinObjC.Packaging/WinObjC.Packageable.targets
@@ -29,7 +29,8 @@
   </PropertyGroup>
 
   <PropertyGroup>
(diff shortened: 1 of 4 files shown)


bbowman:
:shipit:

Issue #1965, opened 2017-02-14
Address lack of proper nil-initialization in WinObjC compiler


jaredhms:
There were some speculative patches for objc_msgSend, but it turns out that the compiler skips the send entirely if the target is known to be nil.  We need to investigate either implementing a nil-initializer in the compiler, or make a thorough pass through the codebase to make sure we're not relying on this behavior for complex types, followed by educating the team against sending messages to nil.  

For example; this call is returning a completely bogus CGRect when 'image' is nil:

    CGRect stretch = [image _imageStretch];
    if (!CGRectEqualToRect([layer contentsCenter], stretch)) {
         [layer setContentsCenter:stretch];
    }

    0:007:x86> dd &stretch
    034ff37c  034ff398 09b41e9c 0ae98df8 c68a9ac4

edvv:
Regarding the workaround, if layer is nil then CGRectEqualToRect compares garbage. The only result is a no-op on a nil anyways, unless the garbage contains a NaN value (highly improbable) which could result in an arithmethic exception...

alakoring:
In my code, I rely on sending messages to nil to work as defined (doing nothing and returning 0). If WinObjC does not behave the same, it would be a bug.

edvv:
@alakoring I think the title to this thread is poorly worded and the actual issue is how are the (function return) registers initialized when the receiver is nil. For a word length it is zero'd out but for anything greater than a word length (such as double or a struct where sizeof is greater than word length) the registers are not necessarily zero'd past a word length (32 bits in this case, or 64 bits on a 64bit implementation). This is the same way it works on iOS and Mac and its always been that way ever since Objective-C was invented. It is pretty standard that a pointer byte length (an id type) is also the same as a word length.

alakoring:
Makes sense, thanks

DHowett-MSFT:
@edvv You're totally right. Thanks for the detailed write up.

In light of that, I'm duping this to #1365. There's more info about the compiler side of the bug and some attempted workarounds over there.

edvv:
@DHowett-MSFT : @jaredhms wrote: "or make a thorough pass through the codebase to make sure we're not relying on this behavior for complex types". ... that is a good idea and is the main takeaway that I got from this thread. (in fact, the entire codebase can use a good cleaning and methodical going over).

jaredhms:
[...]

Issue #1365, opened 2016-11-11
Message sends to nil return structs return garbage values on ARM


aballway:

DHowett-MSFT:
*Specifically*: message sends to nil which return floats or uniform float aggregates will return garbage values on ARM.

**Root Cause**
objc_msgSend zeroes out `r0` and `r1` if `r0` is nil, but it does nothing for `d0-d3`.

***Fix***
```diff
diff --git a/msvc/objc_msgSend.arm.asm b/msvc/objc_msgSend.arm.asm
index 3b29c96..90e95ed 100644
--- a/msvc/objc_msgSend.arm.asm
+++ b/msvc/objc_msgSend.arm.asm
@@ -89,6 +89,10 @@
 #line DBG_SKIP
     movs  r1, #0
 #line DBG_SKIP
+    ldr r2, =LZeroVFPRegisters
+#line DBG_SKIP
+    vldm r2, {d0-d3}
+#line DBG_SKIP
     mov   ip, lr
 #line DBG_SKIP
     b     %5f
@@ -172,5 +176,7 @@
     EXTERN SmallObjectClasses
 LSmallIntClass DCD SmallObjectClasses

+LZeroVFPRegisters DCD 0, 0, 0, 0
+
 #line DBG_SKIP
     END
```

However, I have not had time to validate this fix.

DHowett-MSFT:
On the reference platform, Clang emits an `if(unlikely(!receiver)) { memset(&dest, 0, sizeof(dest)) } ` before every stret-possible message send.

DHowett-MSFT:
Without a compiler fix, we cannot address the general case here.
We can, however, address the <=24-byte VFP aggregate case and the single floating point value cases.

Since getting a `memset(0)`'d struct from a nil message send is relatively new (compared to the rest of Objective-C), the real bug is that a single float or double does not get cleared on ARM. Fixing this should be sufficient for most cases.

rajsesh-msft:
@DHowett-MSFT I have a repro of this on x86 as well.

DHowett-MSFT:
It turns out that we can't fix this in libobjc2.

Assuming the following:
```objc
@interface X
- (T)message;
@end
...
T value = [x message];
```

For non-pointer-sized aggregate returns, the Mac Objective-C codegen shipped with clang emits code like the following:
```objc
T value;
static T _global_T_zeroValue_1; // zero-initialized by virtue of being static; C++ ctors called.
[...]

PR #2094, opened 2017-02-24
Protecting UIButton's CGSize calculations against nil receivers; more…

### LINKED ISSUE #2066: [TypographyInsight] Crash selecting "Kerning" tile - release only
Install TypeographyInsight from release drop share
Open TypeographyInsight app
Select the "Kerning" tile

Expected Result:
Page explaining kerning appears

Actual Result:
App crashes

jaredhms:
… fallout from #1965 and #1365.  Fixes #2066.

### DIFF
--- a/Frameworks/UIKit/UIButton.mm
+++ b/Frameworks/UIKit/UIButton.mm
@@ -371,21 +371,32 @@ static CGRect calculateContentRect(UIButton* self, CGSize size, CGRect contentRe
  @Status Interoperable
 */
 - (CGRect)imageRectForContentRect:(CGRect)contentRect {
-    CGSize titleSize = [self.currentTitle sizeWithFont:self.font];
-    // TODO  #1365 :: Currently cannot assume getting size from nil will return CGSizeZero
-    CGSize imageSize = CGSizeZero;
-    CGRect insetsRect = UIEdgeInsetsInsetRect(contentRect, self.imageEdgeInsets);
-
-    if (!self.currentImage) {
+    /////////////////////////////////////////////////////////////////////
+    // Note: #1365 All size calculations here must check for nil, 
+    // as we cannot assume getting a size from nil will return CGSizeZero
+    /////////////////////////////////////////////////////////////////////
+    UIImage* currentImage = self.currentImage;
+    if (!currentImage) {
+        // No image, so we return zero for the entire content rect
         return CGRectZero;
     }
 
-    imageSize = self.currentImage.size;
+    // Grab the size of the current title
+    // TODO: Should we be asking our label for this?
+    NSString* currentTitle = self.currentTitle;
+    CGSize titleSize = CGSizeZero;
+    if (currentTitle) {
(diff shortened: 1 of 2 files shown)


Frameworks/UIKit/UIButton.mm
```diff
@@ -371,21 +371,32 @@ - (CGRect)contentRectForBounds:(CGRect)bounds {
...
+        // No image, so we return zero for the entire content rect
         return CGRectZero;
     }
 
-    imageSize = self.currentImage.size;
+    // Grab the size of the current title
+    // TODO: Should we be asking our label for this?
+    NSString* currentTitle = self.currentTitle;
+    CGSize titleSize = CGSizeZero;
+    if (currentTitle) {
```
oliversa-msft: >currentTitle [](start = 8, length = 12)

[currentTitle length] > 0?

jaredhms: not sure, and i think the safest approach is to stick to functional equivalence just to be safe

---
In reply to: [103035505](https://github.com/Microsoft/WinObjC/pull/2094#discussion_r103035505) [](ancestors = 103035505)

Frameworks/UIKit/UIButton.mm
```diff
@@ -404,19 +415,34 @@ - (CGRect)imageRectForContentRect:(CGRect)contentRect {
...
+    // Note: #1365 All size calculations here must check for nil, 
+    // as we cannot assume getting a size from nil will return CGSizeZero
+    /////////////////////////////////////////////////////////////////////
 
-    if ([self currentTitle].length == 0) {
+    // Grab the size of the current title
+    // TODO: Should we be asking our label for this?
+    NSString* currentTitle = self.currentTitle;
+    CGSize titleSize = CGSizeZero;
+    if (currentTitle && currentTitle.length > 0) {
```
DHowett-MSFT: per oliver's comment; `nil.length` is 0, so this is a redundant check

jaredhms: i know, but if we don't leave the explicit nil check, someone may forget next time they come through here?

---
In reply to: [103035907](https://github.com/Microsoft/WinObjC/pull/2094#discussion_r103035907) [](ancestors = 103035907)

Frameworks/UIKit/UIButton.mm
oliversa-msft: Are you saying we should *not* do the same check like below: currentTitle && currentTitle.length > 0?

---
In reply to: [103035505](https://github.com/Microsoft/WinObjC/pull/2094#discussion_r103035505) [](ancestors = 103035505)

jaredhms: i'm saying, 'i don't know'; do we have tests to validate this behavior?  I just went through and kept the existing functionality.  The other functions check the length, and this one doesn't, because that's how it worked previously.

---
In reply to: [103035705](https://github.com/Microsoft/WinObjC/pull/2094#discussion_r103035705) [](ancestors = 103035705,103035505)

oliversa-msft:
:shipit:

PR #2097, opened 2017-02-24
Update scripts for running vsimporter on all the samples

memontic-ms:
These changes are needed for packaging's CI builds


scripts/build/unzip.ps1
```diff
@@ -1,14 +1,16 @@
...
 #>
 param(
-    #NOTE: using Parameter(Mandatory) will hang the VSO build job since powershell will block waiting for user input.
+    #NOTE: using Parameter(Mandatory) will hang the VSO build job since
+    #      powershell will block waiting for user input.
     [string]$zipFile = $(throw 'Mandatory parameter "zipFile" is not set.'),
     [string]$outputDir = $(throw 'Madatory parameter "outputDir" is not set.')
     )
 
+$zipFile = gci $zipFile
```
bbowman: why?

scripts/build/VSImport-Samples.ps1
```diff
@@ -14,15 +15,18 @@ $ErrorActionPreference = "Stop"
 
 & $PSScriptRoot\unzip.ps1 -zipFile $zipFile -outputDir $tempDir
 
-$XCodeProjects = gci -recurse $tempDir\*.xcodeproj
+$XCodeProjects = gci -recurse $PSScriptRoot\..\..\samples\*.xcodeproj
+$XCodeProjects += gci -recurse $PSScriptRoot\..\..\tests\testapps\*.xcodeproj
```
bbowman: I think we should use the more generic thing where the xcode folder is passed in.

bbowman:
:clock1:

scripts/build/unzip.ps1
memontic-ms: Converts to absolute path (it was useful to have a * in the path)

---
In reply to: [103060152](https://github.com/Microsoft/WinObjC/pull/2097#discussion_r103060152) [](ancestors = 103060152)

scripts/build/VSImport-Samples.ps1
memontic-ms: Why? Then we'd have to run the script twice and so we'd unzip tools twice. At that point we should just be running vsimporter directly and not using this script at all. 

---
In reply to: [103060288](https://github.com/Microsoft/WinObjC/pull/2097#discussion_r103060288) [](ancestors = 103060288)

scripts/build/unzip.ps1
bbowman: is that the normal way? gci is usually used to get a child item of the current item.

---
In reply to: [103061610](https://github.com/Microsoft/WinObjC/pull/2097#discussion_r103061610) [](ancestors = 103061610,103060152)

scripts/build/VSImport-Samples.ps1
bbowman: I was thinking you'd pass in \build\ and let it recurse from there

---
In reply to: [103061909](https://github.com/Microsoft/WinObjC/pull/2097#discussion_r103061909) [](ancestors = 103061909,103060288)

memontic-ms: ..why build? There's no xcodeprojs under build.

---
In reply to: [103062579](https://github.com/Microsoft/WinObjC/pull/2097#discussion_r103062579) [](ancestors = 103062579,103061909,103060288)

scripts/build/unzip.ps1
memontic-ms: I suppose get-item would be more correct

---
In reply to: [103062479](https://github.com/Microsoft/WinObjC/pull/2097#discussion_r103062479) [](ancestors = 103062479,103061610,103060152)

scripts/build/unzip.ps1
```diff
@@ -1,15 +1,17 @@
...
     .SYNOPSIS
     This script wraps c#'s io.compression.zipfile to unzip a zip file.
-    The script is intended to be used within the VSO build to unzip the winobjc SDK.
+    The script is intended to be used within the VSO build to unzip a NuGet package.
 #>
 param(
-    #NOTE: using Parameter(Mandatory) will hang the VSO build job since powershell will block waiting for user input.
+    #NOTE: using Parameter(Mandatory) will hang the VSO build job since
+    #      powershell will block waiting for user input.
     [string]$zipFile = $(throw 'Mandatory parameter "zipFile" is not set.'),
```
pradipd: I'm not sure what the build def looks like, but, do we need to do this at all?  I thought we don't have  a zip file anymore.

bbowman: We will need to "install" the tools package. I thought it would be better for build isolation to not actually install it and instead just unzip the package and run vsimporter from the unzipped package

bbowman: If and when we move vsimporter into the VSIX for "new project" support, well need to come up with a way to test that in isolation. Not sure what that will look like.

pradipd:
:shipit:

Issue #2098, opened 2017-02-25
Reevaluate the CGImage <-> DisplayTexture cache.


DHowett-MSFT:
We can probably do something smarter when we cache CGImages.

Initially, I thought that we were caching very short-lived CGImages that were created for format conversion.
Fortunately, we're not. We have a cache mapping the original image (pre-conversion) to its converted XAML DisplayTexture.

DHowett-MSFT:
Potential fix:

Use an NSHashTable with weak pointers -- this will obviate the need for a destruction listener.

Use an NSCache -- this will allow for proper behaviour under constrained memory.

jaredhms:
We should also make the cache entry/removal instance-based (off of the IDisplayTexture?) rather than using the static listener pattern.

PR #2099, opened 2017-02-25
Reintroduce support for the CGImage destruction listener.

### LINKED ISSUE #1928: Add support for CGImageAddDestructionListener mechanism
As of now XamlCompositor is using this to release display texture for cgimage.
We need to support this or look into a better way to do this

### LINKED ISSUE #2096: Pori app gets confused and starts to render seemingly-random images
Navigate around a bit and you'll quickly see various images getting swapped out for others.

image

DHowett-MSFT:
Fixes #1928.
Fixes #2096.
Refs #2098.

### DIFF
--- a/Frameworks/CoreGraphics/CGImage.mm
+++ b/Frameworks/CoreGraphics/CGImage.mm
@@ -39,14 +39,12 @@ using namespace Microsoft::WRL;
 
 static const wchar_t* TAG = L"CGImage";
 
-// TODO #1124: remove old code
-#pragma region OLD_CODE
+// This is used by XamlCompositor to flush the DisplayTexture cache.
+// TODO GH#2098 look at where we're using the image cache and what we can do to avoid it.
 static std::vector<CGImageDestructionListener> _imageDestructionListeners;
 COREGRAPHICS_EXPORT void CGImageAddDestructionListener(CGImageDestructionListener listener) {
     _imageDestructionListeners.push_back(listener);
 }
-
-#pragma endregion OLD_CODE
 
 #pragma region CGImageImplementation
 
@@ -224,6 +222,12 @@ struct __CGImage : CoreFoundation::CppBase<__CGImage> {
     inline __CGImage& SetRenderingIntent(CGColorRenderingIntent intent) {
         _impl.renderingIntent = intent;
         return *this;
+    }
+
+    ~__CGImage() {
+        for (auto listener: _imageDestructionListeners) {
+            listener(this);
+        }
     }
 };
 


Frameworks/CoreGraphics/CGImage.mm
```diff
@@ -39,15 +39,13 @@
 
 static const wchar_t* TAG = L"CGImage";
 
-// TODO #1124: remove old code
-#pragma region OLD_CODE
+// This is used by XamlCompositor to flush the DisplayTexture cache.
+// TODO GH#2098 look at where we're using the image cache and what we can do to avoid it.
 static std::vector<CGImageDestructionListener> _imageDestructionListeners;
```
jaredhms: >_imageDestructionListeners [](start = 47, length = 26)

should this be 'g_' ?

Frameworks/CoreGraphics/CGImage.mm
```diff
@@ -225,6 +223,12 @@ inline WICPixelFormatGUID PixelFormat() const {
         _impl.renderingIntent = intent;
         return *this;
     }
+
+    ~__CGImage() {
+        for (auto listener: _imageDestructionListeners) {
```
DHowett-MSFT: It is the listener's job to make sure that it can be called on any thread.

Access to the listener list is not currently synchronized because there is ONE consumer.

In the future, _that consumer will be removed_.

Frameworks/CoreGraphics/CGImage.mm
```diff
@@ -225,6 +223,12 @@ inline WICPixelFormatGUID PixelFormat() const {
...
+            listener(this);
+        }
```
jaredhms: do we need a lock around this?  also, kind of odd to have an instance free all static listeners....are you sure this is working correctly?

jaredhms: :o you read my mind

---
In reply to: [103062474](https://github.com/Microsoft/WinObjC/pull/2099#discussion_r103062474) [](ancestors = 103062474)

Frameworks/CoreGraphics/CGImage.mm
DHowett-MSFT: No, it's not global. It's static.

Frameworks/CoreGraphics/CGImage.mm
DHowett-MSFT: :smile:

DHowett-MSFT: A listener is added for _all_ CGImages, not just one. W

Frameworks/CoreGraphics/CGImage.mm
jaredhms: ok, then  's_' for clarity

---
In reply to: [103062541](https://github.com/Microsoft/WinObjC/pull/2099#discussion_r103062541) [](ancestors = 103062541)

jaredhms:
:shipit:

PR #2100, opened 2017-02-25
Implementation of CGRectDivide + tests

msft-Jeyaram:
#2042

### DIFF
--- a/Frameworks/CoreGraphics/CGGeometry.mm
+++ b/Frameworks/CoreGraphics/CGGeometry.mm
@@ -1,6 +1,6 @@
 //******************************************************************************
 //
-// Copyright (c) 2016 Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
 //
 // This code is licensed under the MIT License (MIT).
 //
@@ -17,17 +17,21 @@
 #import <StubReturn.h>
 #import <Starboard.h>
 #import <CoreGraphics/CGGeometry.h>
+#import <LoggingNative.h>
 #import <limits>
+#import <algorithm>
 
 #import "Etc.h"
+
+static const wchar_t* TAG = L"CGGeometry";
 
 const CGRect CGRectInfinite = CGRectMake(0, 0, std::numeric_limits<float>::infinity(), std::numeric_limits<float>::infinity());
 const CGPoint CGPointZero = CGPointMake(0, 0);
 const CGRect CGRectZero = CGRectMake(0, 0, 0, 0);
 const CGSize CGSizeZero = CGSizeMake(0, 0);
 
-/** 
- @Status caveat 
+/**
+ @Status caveat
  @Notes TODO: unclear the value is correct or not based on reference doc. Need Revisit
 */
 const CGRect CGRectNull = { std::numeric_limits<float>::infinity(), std::numeric_limits<float>::infinity(), 0.0f, 0.0f };
@@ -87,11 +91,94 @@ bool CGRectMakeWithDictionaryRepresentation(CFDictionaryRef dict, CGRect* rect) 
 }
 
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 void CGRectDivide(CGRect rect, CGRect* slice, CGRect* remainder, CGFloat amount, CGRectEdge edge) {
-    UNIMPLEMENTED();
+    RETURN_IF(slice == nullptr);
+    RETURN_IF(remainder == nullptr);
+
+    if (CGRectIsNull(rect)) {
+        *slice = CGRectNull;
+        *remainder = CGRectNull;
+        return;
+    }
+
+    amount = std::max(0.0f, amount);
+    rect = CGRectStandardize(rect);
+
+    /*
+    * (0,0)                      (width,0)
+    * x-----------------------------x
+    * |                             |
+    * |                             |
+    * |                             |
+    * |                             |
+    * |                             |
+    * x-----------------------------x
+    * (0,height)               (width,height)
+    *
+    *
+    * The division is based on CGRectEdge, which edge to divide.
+    * e.g CGRectMinYEdge
+    *
+    *   (0,0)                      (width,0)
+    *    x-----------------------------x
+    *    | slice            | (amount) |
+    *    |                  v          |
+    * (0,amount)------------------(width,amount)
+    *    | reminder                    |
+    *    |                             |
+    *    x-----------------------------x
+    *   0,height-amount)               (width,height-amount)
+    */
+
+    // Set both to rect, then update the sizes.
+    *slice = rect;
+    *remainder = rect;
+
+    switch (edge) {
+        case CGRectMinYEdge:
+            amount = (amount > rect.size.height) ? rect.size.height : amount;
+            remainder->origin.y += amount;
+            if (amount >= rect.size.height) {
+                remainder->size.height = 0;
+            } else {
+                slice->size.height = amount;
+                remainder->size.height -= amount;
+            }
+            break;
+        case CGRectMaxYEdge:
+            if (amount >= rect.size.height) {
+                remainder->size.height = 0;
+            } else {
+                slice->origin.y += (rect.size.height - amount);
+                slice->size.height = amount;
+                remainder->size.height -= amount;
+            }
+            break;
+        case CGRectMinXEdge:
+            amount = (amount > rect.size.width) ? rect.size.width : amount;
+            remainder->origin.x += amount;
+            if (amount >= rect.size.width) {
+                remainder->size.width = 0;
+            } else {
+                slice->size.width = amount;
+                remainder->size.width -= amount;
+            }
+            break;
+        case CGRectMaxXEdge:
+            if (amount >= rect.size.width) {
+                remainder->size.width = 0;
+            } else {
+                remainder->size.width -= amount;
+                slice->origin.x += rect.size.width - amount;
+                slice->size.width = amount;
+            }
+            break;
+        default:
+            TraceError(TAG, L"Invalid CGRectEdge.");
+            break;
+    }
 }
 
 /**
--- a/Frameworks/CoreGraphics/CGGeometry.mm
+++ b/Frameworks/CoreGraphics/CGGeometry.mm
@@ -25,16 +25,14 @@
 
 static const wchar_t* TAG = L"CGGeometry";
 
-const CGRect CGRectInfinite = CGRectMake(0, 0, std::numeric_limits<float>::infinity(), std::numeric_limits<float>::infinity());
+const CGRect CGRectInfinite = CGRectMake(std::numeric_limits<float>::infinity(),
+                                         std::numeric_limits<float>::infinity(),
+                                         std::numeric_limits<float>::infinity(),
+                                         std::numeric_limits<float>::infinity());
 const CGPoint CGPointZero = CGPointMake(0, 0);
 const CGRect CGRectZero = CGRectMake(0, 0, 0, 0);
 const CGSize CGSizeZero = CGSizeMake(0, 0);
-
-/**
- @Status caveat
- @Notes TODO: unclear the value is correct or not based on reference doc. Need Revisit
-*/
-const CGRect CGRectNull = { std::numeric_limits<float>::infinity(), std::numeric_limits<float>::infinity(), 0.0f, 0.0f };
+const CGRect CGRectNull = CGRectMake(std::numeric_limits<float>::infinity(), std::numeric_limits<float>::infinity(), 0.0f, 0.0f);
 
 /**
  @Status Stub
--- a/include/CoreGraphics/CGGeometry.h
+++ b/include/CoreGraphics/CGGeometry.h
@@ -38,12 +38,7 @@ typedef struct {
     CGFloat dy;
 } CGVector;
 
-typedef CF_ENUM(CFIndex, CGRectEdge) {
-    CGRectMinXEdge,
-    CGRectMinYEdge,
-    CGRectMaxXEdge,
-    CGRectMaxYEdge
-};
+typedef CF_ENUM(CFIndex, CGRectEdge) { CGRectMinXEdge, CGRectMinYEdge, CGRectMaxXEdge, CGRectMaxYEdge };
 
 COREGRAPHICS_EXPORT const CGRect CGRectInfinite;
 COREGRAPHICS_EXPORT const CGPoint CGPointZero;
@@ -237,4 +232,4 @@ COREGRAPHICS_EXPORT CFDictionaryRef CGRectCreateDictionaryRepresentation(CGRect)
 COREGRAPHICS_EXPORT bool CGPointMakeWithDictionaryRepresentation(CFDictionaryRef dict, CGPoint* point) STUB_METHOD;
 COREGRAPHICS_EXPORT bool CGSizeMakeWithDictionaryRepresentation(CFDictionaryRef dict, CGSize* size) STUB_METHOD;
 COREGRAPHICS_EXPORT bool CGRectMakeWithDictionaryRepresentation(CFDictionaryRef dict, CGRect* rect) STUB_METHOD;
-COREGRAPHICS_EXPORT void CGRectDivide(CGRect rect, CGRect* slice, CGRect* remainder, CGFloat amount, CGRectEdge edge) STUB_METHOD;
+COREGRAPHICS_EXPORT void CGRectDivide(CGRect rect, CGRect* slice, CGRect* remainder, CGFloat amount, CGRectEdge edge);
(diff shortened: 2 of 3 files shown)


tests/unittests/CoreGraphics/CGGeometryTests.mm
```diff
@@ -52,4 +52,4 @@
     CGSize inverseTransformedSize = CGSizeApplyAffineTransform(transformedSize, CGAffineTransformInvert(rotate));
     EXPECT_NEAR(size.width, inverseTransformedSize.width, 0.001);
     EXPECT_NEAR(size.height, inverseTransformedSize.height, 0.001);
-}
```
msft-Jeyaram: adding few test cases , aside from the drawing tests.

msft-Jeyaram:
Yep definitely, added the comment:)
Coming up in the iteration, i think the drawings are valid to have

DHowett-MSFT:
I think the drawing are very cool but they are unnecessary burden on our repository and LFS storage. They have to be downloaded and kept up to date and they don't test the feature at hand.
They're actually just testing rendering rectangles that are next to eachother in different layouts, which is _not_ a CGRectDivide test.

msft-Jeyaram:
Given CGRectDiv is mainly used for layouts within the view. The drawings are valid to have, but they should really be toned down
The math ones should be more, but given it's a parameterized test and two outcomes, it would be odd to do a pure parameterized test. I'd rather not repeat the calculations in the test nor have expected values for all those combinations

DHowett-MSFT:
Having expected values for the small and big split is the only valid way to test rect divisions. It's pure math! If we regress rect drawing, we will have an invalid failure on “rect division”.

Frameworks/CoreGraphics/CGGeometry.mm
```diff
@@ -26,8 +26,8 @@
 const CGRect CGRectZero = CGRectMake(0, 0, 0, 0);
 const CGSize CGSizeZero = CGSizeMake(0, 0);
 
-/** 
- @Status caveat 
+/**
+ @Status caveat
```
rajsesh-msft: Could we go ahead and fix this?

rajsesh-msft:
Let me weigh in on the usage of drawing tests here.  The idea behind the drawing tests is purely to validate drawing code which cannot be validated at an API level.  So please add extensive math tests if you have concerns on the validity of CGRectDivide.  There is no need for drawing tests here and those should be removed.

Frameworks/CoreGraphics/CGGeometry.mm
```diff
@@ -87,11 +87,85 @@ bool CGRectMakeWithDictionaryRepresentation(CFDictionaryRef dict, CGRect* rect)
...
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 void CGRectDivide(CGRect rect, CGRect* slice, CGRect* remainder, CGFloat amount, CGRectEdge edge) {
-    UNIMPLEMENTED();
+    RETURN_IF(slice == nullptr);
+    RETURN_IF(remainder == nullptr);
+
+    amount = (amount < 0) ? 0 : amount;
```
aballway: nit: std::max

Frameworks/CoreGraphics/CGGeometry.mm
```diff
@@ -87,11 +87,85 @@ bool CGRectMakeWithDictionaryRepresentation(CFDictionaryRef dict, CGRect* rect)
...
+
+    // Set both to rect, then update the sizes.
+    *slice = rect;
+    *remainder = rect;
+
+    switch (edge) {
+        case CGRectMinYEdge:
+            remainder->origin.y += amount;
+            if (amount >= rect.size.height) {
+                remainder->size.height = 0;
```
aballway: This would return a weird value for invalid rects. What does the reference platform return for these cases?

Frameworks/CoreGraphics/CGGeometry.mm
```diff
@@ -87,11 +87,85 @@ bool CGRectMakeWithDictionaryRepresentation(CFDictionaryRef dict, CGRect* rect)
...
+            if (amount >= rect.size.width) {
+                remainder->size.width = 0;
+            } else {
+                remainder->size.width -= amount;
+                slice->origin.x += rect.size.width - amount;
+                slice->size.width = amount;
+            }
+            break;
+        default:
+            break;
```
aballway: nit: should we log a warning here?

Frameworks/CoreGraphics/CGGeometry.mm
msft-Jeyaram: weird ones :p
The tests validate it

---
In reply to: [103261259](https://github.com/Microsoft/WinObjC/pull/2100#discussion_r103261259) [](ancestors = 103261259)

msft-Jeyaram:
@DHowett @rajsesh-msft @aballway  ping!

tests/unittests/CoreGraphics/CGGeometryTests.mm
```diff
@@ -52,4 +52,122 @@
...
+    CGRect rRemainder = ::testing::get<4>(GetParam());
+
+    CGRect slice, remainder;
+    CGRectDivide(rect, &slice, &remainder, amount, edge);
+
+    EXPECT_TRUE(CGRectEqualToRect(rSlice, slice));
+    EXPECT_TRUE(CGRectEqualToRect(rRemainder, remainder));
+}
+
+std::tr1::tuple<CGRect, CGFloat, CGRectEdge, /*slice*/ CGRect, /*remainder*/ CGRect> const cCGRectDivideResults[] = {
```
msft-Jeyaram: making it static

Frameworks/CoreGraphics/CGGeometry.mm
```diff
@@ -17,21 +17,23 @@
...
 #import "Etc.h"
 
-const CGRect CGRectInfinite = CGRectMake(0, 0, std::numeric_limits<float>::infinity(), std::numeric_limits<float>::infinity());
+static const wchar_t* TAG = L"CGGeometry";
+
+const CGRect CGRectInfinite = CGRectMake(std::numeric_limits<float>::infinity(),
+                                         std::numeric_limits<float>::infinity(),
+                                         std::numeric_limits<float>::infinity(),
+                                         std::numeric_limits<float>::infinity());
+const CGRect CGRectNull = { std::numeric_limits<float>::infinity(), std::numeric_limits<float>::infinity(), 0.0f, 0.0f };
```
aballway: nit: use CGRectMake for consistency, or {{,},{,}} since CGRect isn't flat

include/CoreGraphics/CGGeometry.h
```diff
@@ -38,12 +38,7 @@ typedef struct {
     CGFloat dy;
 } CGVector;
 
-typedef CF_ENUM(CFIndex, CGRectEdge) {
-    CGRectMinXEdge,
-    CGRectMinYEdge,
-    CGRectMaxXEdge,
-    CGRectMaxYEdge
-};
+typedef CF_ENUM(CFIndex, CGRectEdge) { CGRectMinXEdge, CGRectMinYEdge, CGRectMaxXEdge, CGRectMaxYEdge };
```
DHowett-MSFT: stupid clang-format.

tests/unittests/CoreGraphics/CGGeometryTests.mm
DHowett-MSFT: you should _not_ be reaching into `::tr1`..?

tests/unittests/CoreGraphics/CGGeometryTests.mm
```diff
@@ -52,4 +52,122 @@
...
+                          10.35,
+                          CGRectMaxYEdge,
+                          CGRectMake(0, 89.65, 100, 10.35),
+                          CGRectMake(0, 0, 100, 89.65)),
+    ::testing::make_tuple(CGRectMake(0, 0, 100, 100),
+                          10.35,
+                          CGRectMinYEdge,
+                          CGRectMake(0, 0, 100, 10.35),
+                          CGRectMake(0, 10.35, 100, 89.65)),
+    ::testing::make_tuple(CGRectMake(0, 0, 900, 800), -50, CGRectMaxXEdge, CGRectMake(900, 0, 0, 800), CGRectMake(0, 0, 900, 800)),
```
DHowett-MSFT: might be worth forcing these to format the same as the others and disabling clang-format in this zone

tests/unittests/CoreGraphics/CGGeometryTests.mm
```diff
@@ -52,4 +52,122 @@
...
+    CGRect rect = ::testing::get<0>(GetParam());
+    CGFloat amount = ::testing::get<1>(GetParam());
+    CGRectEdge edge = ::testing::get<2>(GetParam());
+    CGRect rSlice = ::testing::get<3>(GetParam());
+    CGRect rRemainder = ::testing::get<4>(GetParam());
+
+    CGRect slice, remainder;
+    CGRectDivide(rect, &slice, &remainder, amount, edge);
+
+    EXPECT_TRUE(CGRectEqualToRect(rSlice, slice));
```
aballway: nit: use EXPECT_EQ so it will print rect values on failure

tests/unittests/CoreGraphics/CGGeometryTests.mm
```diff
@@ -52,4 +52,122 @@
...
+    EXPECT_TRUE(CGRectEqualToRect(rRemainder, remainder));
+}
+
+std::tr1::tuple<CGRect, CGFloat, CGRectEdge, /*slice*/ CGRect, /*remainder*/ CGRect> const cCGRectDivideResults[] = {
+
+    ::testing::make_tuple(CGRectMake(0, 0, 100, 100),
```
DHowett-MSFT: please add brief comments explaining what each test does.

```
// Advances X (normal case)
...
// Advances X (outside of rectangle)
..
// Reduces height (below 0)
..
```

tests/unittests/CoreGraphics/CGGeometryTests.mm
```diff
@@ -52,4 +52,122 @@
...
+                          CGRectMake(100, 100, 125, 250)),
+    ::testing::make_tuple(CGRectNull, -50.24, CGRectMaxXEdge, CGRectNull, CGRectNull),
+    ::testing::make_tuple(CGRectNull, 150.240000, CGRectMinXEdge, CGRectNull, CGRectNull),
+    ::testing::make_tuple(CGRectNull, -50.24, CGRectMaxYEdge, CGRectNull, CGRectNull),
+    ::testing::make_tuple(CGRectNull, 150.240000, CGRectMinYEdge, CGRectNull, CGRectNull),
+};
+
+INSTANTIATE_TEST_CASE_P(CGGeometry, CGGeometryDivRect, ::testing::ValuesIn(cCGRectDivideResults));
+
+TEST(CGGeometry, RectDivideInf) {
```
DHowett-MSFT: hm, why can't this be part of the other thinger?

tests/unittests/CoreGraphics/CGGeometryTests.mm
aballway: use testing::Combine?

tests/unittests/CoreGraphics/CGGeometryTests.mm
msft-Jeyaram: it just makes this odd case stand out. also makes life easier + we don't have to compute CGRectInf

---
In reply to: [105253463](https://github.com/Microsoft/WinObjC/pull/2100#discussion_r105253463) [](ancestors = 105253463)

tests/unittests/CoreGraphics/CGGeometryTests.mm
msft-Jeyaram: not useful if we have expected values. pre populated 

---
In reply to: [105253532](https://github.com/Microsoft/WinObjC/pull/2100#discussion_r105253532) [](ancestors = 105253532)

Frameworks/CoreGraphics/CGGeometry.mm
```diff
@@ -87,11 +89,92 @@ bool CGRectMakeWithDictionaryRepresentation(CFDictionaryRef dict, CGRect* rect)
...
+            if (amount >= rect.size.width) {
+                remainder->size.width = 0;
+            } else {
+                remainder->size.width -= amount;
+                slice->origin.x += rect.size.width - amount;
+                slice->size.width = amount;
+            }
+            break;
+        default:
+            TraceError(TAG, L"Invalid CGRectEdge.");
```
DHowett-MSFT: :+1:

DHowett-MSFT:
You should fully revert `CGContextDrawing_FillTests.cpp`; right now you have introduced a conflict on a practically-unchanged file.

tests/unittests/CoreGraphics/CGGeometryTests.mm
msft-Jeyaram: isn't supported

---
In reply to: [105253190](https://github.com/Microsoft/WinObjC/pull/2100#discussion_r105253190) [](ancestors = 105253190)

tests/unittests/CoreGraphics/CGGeometryTests.mm
msft-Jeyaram: hmm but it makes it soo readable :)

---
In reply to: [105253114](https://github.com/Microsoft/WinObjC/pull/2100#discussion_r105253114) [](ancestors = 105253114)

tests/unittests/CoreGraphics/CGGeometryTests.mm
aballway: import CppUtils.h

tests/unittests/CoreGraphics/CGGeometryTests.mm
msft-Jeyaram: looking at the CGRectEdge and the amount you reduce by, the test itself indicates what it does.

---
In reply to: [105253335](https://github.com/Microsoft/WinObjC/pull/2100#discussion_r105253335) [](ancestors = 105253335)

DHowett-MSFT: The point of comments is to clarify difficult-to-read sections of code. Since this is a pile of constant values in no discernible order, it is a difficult-to-read section of code.

Adding comments will only serve to increase its clarity.

It also helps us think about all the test cases and whether there are gaps.

tests/unittests/CoreGraphics/CGGeometryTests.mm
DHowett-MSFT: It *will* be worth forcing these to format the same as the others and disabling clang-format in this zone

tests/unittests/CoreGraphics/CGGeometryTests.mm
```diff
@@ -52,4 +53,122 @@
...
+    CGRect expectedSlice = CGRectInfinite;
+    expectedSlice.size.width = 0;
+
+    CGRect slice, remainder;
+    CGRectDivide(CGRectInfinite, &slice, &remainder, -50.24, CGRectMaxXEdge);
+    EXPECT_TRUE(CGRectEqualToRect(expectedSlice, slice));
```
aballway: nit: these as well

tests/unittests/CoreGraphics.Drawing/CGContextDrawing_FillTests.cpp
```diff
@@ -133,4 +133,4 @@ static int directions[] = { 0, 1 };
 // TODO: enable when #2062 is fixed.
 INSTANTIATE_TEST_CASE_P(DISABLED_CGContextFill,
                         CGContextArcFill,
-                        ::testing::Combine(::testing::ValuesIn(sweep), ::testing::ValuesIn(directions)));
+                        ::testing::Combine(::testing::ValuesIn(sweep), ::testing::ValuesIn(directions)));
```
aballway: nit: revert this file

tests/unittests/CoreGraphics/CGGeometryTests.mm
```diff
@@ -52,4 +53,122 @@
...
+    CGRect rRemainder = ::testing::get<4>(GetParam());
+
+    CGRect slice, remainder;
+    CGRectDivide(rect, &slice, &remainder, amount, edge);
+
+    EXPECT_EQ(rSlice, slice);
+    EXPECT_EQ(rRemainder, remainder);
+}
+
+static const std::tuple<CGRect, CGFloat, CGRectEdge, /*slice*/ CGRect, /*remainder*/ CGRect> cCGRectDivideResults[] = {
```
aballway: super nit: use ::testing::tuple

msft-Jeyaram:
@DHowett-MSFT  ping

msft-Jeyaram:
@DHowett-MSFT  tests validated on ARM

PR #2106, opened 2017-02-27
GLKTexture load files without assuming it's PNG

### LINKED ISSUE #2102: GLKitComplex does not show background
This happens in the new release candidate.

msft-Jeyaram:
GLKTexture cubeMapWithContentsOfFiles was loading all files assuming it's PNG.
Updated the function to handle all types of files.

Fixes #2102


msft-Jeyaram:
Looks like there are few other instance I need to update.
Updating it

Frameworks/GLKit/GLKTexture.mm~RF14aac35d.TMP
```diff
@@ -0,0 +1,767 @@
+//******************************************************************************
```
rajsesh-msft: This .tmp file should be removed (from clang format?)

Frameworks/GLKit/GLKTexture.mm
```diff
@@ -277,7 +279,8 @@ @implementation GLKTextureLoader {
 */
 + (GLKTextureInfo*)textureWithContentsOfFile:(NSString*)fname options:(NSDictionary*)opts error:(NSError**)err {
     CGDataProviderRef provider = CGDataProviderCreateWithFilename([fname UTF8String]);
-    CGImageRef img = CGImageCreateWithPNGDataProvider(provider, NULL, NO, kCGRenderingIntentDefault);
+    unsigned char* dataBytes = static_cast<unsigned char*>(const_cast<void*>(_CGDataProviderGetData(provider)));
+    CGImageRef img = _CGImageGetImageFromData(dataBytes, _CGDataProviderGetSize(provider));
```
rajsesh-msft: Nit: I would add a _CGImageGetImageFromDataProvider() instead of these two lines.

Frameworks/GLKit/GLKTexture.mm
```diff
@@ -277,7 +279,8 @@ @implementation GLKTextureLoader {
 */
 + (GLKTextureInfo*)textureWithContentsOfFile:(NSString*)fname options:(NSDictionary*)opts error:(NSError**)err {
     CGDataProviderRef provider = CGDataProviderCreateWithFilename([fname UTF8String]);
-    CGImageRef img = CGImageCreateWithPNGDataProvider(provider, NULL, NO, kCGRenderingIntentDefault);
+    unsigned char* dataBytes = static_cast<unsigned char*>(const_cast<void*>(_CGDataProviderGetData(provider)));
```
aballway: nit: should make a _CGImageCreateImageFromDataProvider rather than copying this around three times. Also saves from having to pull in CGDataProviderInternal here.

aballway: Not super related to this, but _CGImageGetImageFromData returns a +1 rn which breaks the CF naming convention, so that needs to be changed at some point

Frameworks/GLKit/GLKTexture.mm~RF14aac35d.TMP
aballway: !

Frameworks/GLKit/GLKTexture.mm
```diff
@@ -395,7 +398,9 @@ + (GLKTextureInfo*)cubeMapWithContentsOfFile:(NSString*)fname options:(NSDiction
     if (!provider) {
         return nil;
     }
-    CGImageRef img = CGImageCreateWithPNGDataProvider(provider, NULL, NO, kCGRenderingIntentDefault);
+
+    unsigned char* dataBytes = static_cast<unsigned char*>(const_cast<void*>(_CGDataProviderGetData(provider)));
```
aballway: is there any way that we can add a unit test so this won't regress?

Frameworks/GLKit/GLKTexture.mm
msft-Jeyaram: was planning on then switched back, adding.

---
In reply to: [103335956](https://github.com/Microsoft/WinObjC/pull/2106#discussion_r103335956) [](ancestors = 103335956)

Frameworks/GLKit/GLKTexture.mm
msft-Jeyaram: hmm looks like others functions too, loadJPEG, etc.


---
In reply to: [103336909](https://github.com/Microsoft/WinObjC/pull/2106#discussion_r103336909) [](ancestors = 103336909)

Frameworks/CoreGraphics/CGImage.mm
```diff
@@ -650,6 +652,12 @@ CGImageRef _CGImageCreateCopyWithPixelFormat(CGImageRef image, WICPixelFormatGUI
     return imageRef;
 }
 
+CGImageRef _CGImageGetImageFromDataProvider(CGDataProviderRef provider) {
```
DHowett-MSFT: Since you're introducing a totally new function, can you name it `_CGImageCreateFromDataProvider`? While the other `Get` is wrong, this one doesn't have to be!

Frameworks/CoreGraphics/CGImage.mm
```diff
@@ -650,6 +652,12 @@ CGImageRef _CGImageCreateCopyWithPixelFormat(CGImageRef image, WICPixelFormatGUI
...
+    RETURN_NULL_IF(!provider);
+    unsigned char* dataBytes = static_cast<unsigned char*>(const_cast<void*>(_CGDataProviderGetData(provider)));
+    return _CGImageGetImageFromData(dataBytes, _CGDataProviderGetSize(provider));
```
aballway: nit, just pass into method directly rather than const void* -> void* -> unsigned char* -> void*

PR #2110, opened 2017-02-28
Winmd2ObjC binary update: 

mukhole:
Replace [NSAutorelease alloc] with @autoreleasepool blocks


DHowett-MSFT:
@mukhole I don't know why this is necessary. We are bumping hundreds of binaries for an unnecessary change.

mukhole:
We are making this change to avoid allocating NSAutorelease objects each time we dispatch a runtime call or invoke a callback.

Issue #2095, opened 2017-02-24
Pori app saves images upside down and cropped


jaredhms:
![image](https://cloud.githubusercontent.com/assets/13719817/23321562/ea8d46ec-fa94-11e6-9bac-0f6df4cdbc8f.png)

DHowett-MSFT:
This is perhaps the funniest thing I've seen all week.

jaredhms:
@yiyang-msft, last time this regressed, it was a bug in UIScrollView.  Can you please take a look?

DHowett-MSFT:
You're going to love this.

PR #2117, opened 2017-02-28
Fix some rendering issues that were causing #2095.

DHowett-MSFT:
* `-[CALayer renderInContext:]` has the ability to render a cached version of itself; it was doing so upside-down because it was using a Cairo implementation detail.
* `-[UIImage drawAtPoint:]` squashed every image it rendered because it switched width & height.

The bug in UIImage dates back to the Cairo implementation as well; the invalid size was passed to a function that ignored it. When our refactor made use of that parameter, things rapidly went south.
We did not catch the second regression because all the images we drew with it were square.

### DIFF
--- a/Frameworks/QuartzCore/CALayer.mm
+++ b/Frameworks/QuartzCore/CALayer.mm
@@ -427,14 +427,20 @@ CGContextRef CreateLayerContentsBitmapContext32(int width, int height, float sca
             [priv->delegate drawLayer:self inContext:ctx];
         }
     } else {
-        CGRect rect;
+        // If the layer has cached contents, blit them directly.
 
-        rect.origin.x = 0;
-        rect.origin.y = priv->bounds.size.height * priv->contentsScale;
-        rect.size.width = priv->bounds.size.width * priv->contentsScale;
-        rect.size.height = -priv->bounds.size.height * priv->contentsScale;
+        // Since the layer was rendered in Quartz referential (ULO) AND the current context
+        // is assumed to be Quartz referential (ULO), BUT the layer's cached contents
+        // were captured in a CGImage (CGImage referential, LLO), we have to flip
+        // the context again before we render it.
 
-        _CGContextDrawImageRect(ctx, priv->contents, rect, destRect);
+        // |1  0 0| is the transformation matrix for flipping a rect anchored at 0,0 about its Y midpoint.
+        // |0 -1 0|
+        // |0  h 1|
+        CGContextSaveGState(ctx);
+        CGContextConcatCTM(ctx, CGAffineTransformMake(1, 0, 0, -1, 0, destRect.size.height));
+        CGContextDrawImage(ctx, destRect, priv->contents);
+        CGContextRestoreGState(ctx);
     }
 
     //  Draw sublayers
--- a/Frameworks/UIKit/UIImage.mm
+++ b/Frameworks/UIKit/UIImage.mm
@@ -491,8 +491,8 @@ static inline CGImageRef getImage(UIImage* uiImage) {
 
     CGRect pos;
     pos.origin = point;
-    pos.size.width = (img_height / _scale);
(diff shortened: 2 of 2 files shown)


Frameworks/UIKit/UIImage.mm
```diff
@@ -491,8 +491,8 @@ - (void)drawAtPoint:(CGPoint)point blendMode:(CGBlendMode)mode alpha:(float)alph
 
     CGRect pos;
     pos.origin = point;
-    pos.size.width = (img_height / _scale);
-    pos.size.height = (img_width / _scale);
+    pos.size.height = (img_height / _scale);
+    pos.size.width = (img_width / _scale);
```
DHowett-MSFT: @aballway might like this

yiyang-msft: typical think too fast or type too fast.

---
In reply to: [103579177](https://github.com/Microsoft/WinObjC/pull/2117#discussion_r103579177) [](ancestors = 103579177)

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -427,14 +427,15 @@ - (void)renderInContext:(CGContextRef)ctx {
...
-        rect.size.height = -priv->bounds.size.height * priv->contentsScale;
-
-        _CGContextDrawImageRect(ctx, priv->contents, rect, destRect);
+        // If the layer has cached contents, blit them directly.
+
+        // |1  0 0| is the transformation matrix for flipping a rect anchored at 0,0 about its Y midpoint.
+        // |0 -1 0|
+        // |0  h 1|
+        CGContextSaveGState(ctx);
+        CGContextConcatCTM(ctx, CGAffineTransformMake(1, 0, 0, -1, 0, destRect.size.height));
```
yiyang-msft: >CGAffineTransformMake [](start = 32, length = 21)

mind adding a comment calling out why this this transform is needed so that other people without the context can understand the reasoning behind. Thx!

yiyang-msft:
:shipit:

oliversa-msft:
:shipit:

Frameworks/UIKit/UIImage.mm
aballway: nit: position is a very silly name for a rect 😝

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -427,14 +427,15 @@ - (void)renderInContext:(CGContextRef)ctx {
...
-        rect.size.width = priv->bounds.size.width * priv->contentsScale;
-        rect.size.height = -priv->bounds.size.height * priv->contentsScale;
-
-        _CGContextDrawImageRect(ctx, priv->contents, rect, destRect);
+        // If the layer has cached contents, blit them directly.
+
+        // |1  0 0| is the transformation matrix for flipping a rect anchored at 0,0 about its Y midpoint.
+        // |0 -1 0|
+        // |0  h 1|
+        CGContextSaveGState(ctx);
```
aballway: is saving/popping g-state relatively free?  we should be able to get away with just reverting the CTM if not.

DHowett-MSFT: It's very cheap.

Frameworks/UIKit/UIImage.mm
DHowett-MSFT: nit: oh well :)

Frameworks/QuartzCore/CALayer.mm
DHowett-MSFT: @yiyang-msft done

DHowett-MSFT:
This is in CI build; git-lfs is being particularly slow today.

DHowett-MSFT:
@rajsesh-msft if you end up being the one to merge this (I'll try to monitor the build status!), would you mind copying the Pull Request body into the squash message body?

Thanks :)

PR #2019, opened 2017-02-22
Fixing internal naming conventions in CoreGraphics

MSFTFox:
Remove any unused Core Graphics internal APIs as well as normalize naming conventions across files.

Fix #1301

PR #2119, opened 2017-03-01
Add support for "enhanced error handling" to CGContext and CALayer

### LINKED ISSUE #1194: Recreate render target if EndDraw fails

DHowett-MSFT:
Direct2D can emit a failure or remove the backing hardware device from your render target at the end of any drawing operation. However, CoreGraphics does not natively have support for communicating failure states out to interested consumers.

Enhanced Error Handling, when enabled, allows a sufficiently motivated developer to become part of CGContext's rendering lifecycle.

The error of first and foremost importance is `D2DERR_RECREATE_TARGET`: It is not an error, but a plea to retry the last rendering operation after regenerating the render target. While CoreGraphics could batch up a list of pending drawing operations and replay them on a new target, it opts instead to report this to an Enhanced Error Handling consumer through `kCGContextErrorDeviceReset`.

All other failing drawing operations result in `kCGContextErrorInvalidParameter`.

In this pull request, **CALayer** has been augmented (in a bid to better prepare us for a hardware-accelerated future) to retry its rendering up to three (3) times if the device goes away. Since a layer can draw itself from start to finish, `-display` is the best point at which to handle the newly-minted CGContext errors.

Fixes #1194.


Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -47,6 +47,7 @@
 #import <cmath>
 #import <list>
 #import <vector>
+#import <array>
```
DHowett-MSFT: unused.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -471,21 +475,32 @@ inline void ClearPath() {
     }
 
     inline bool ShouldDraw() {
-        return CurrentGState().ShouldDraw();
+        return FAILED(_firstErrorHr) || CurrentGState().ShouldDraw();
```
DHowett-MSFT: Hilarious. prefer `SUCCEEDED` or `!FAILED`

DHowett-MSFT: And probably `&&`

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -518,69 +521,73 @@ - (void)display {
...
+                CGRect wholeRect = CGRectMake(0, 0, width, height);
+                CGContextFillRect(drawContext, wholeRect);
+                CGContextRestoreGState(drawContext);
+            }
 
-            CGContextSaveGState(drawContext);
-            CGContextSetFillColorWithColor(drawContext, [static_cast<UIColor*>(priv->_backgroundColor) CGColor]);
+            // UIKit and CALayer consumers expect the origin to be in the top left.
+            // CoreGraphics defaults to the bottom left, so we must flip and translate the canvas.
+            CGContextTranslateCTM(drawContext, 0, heightInPoints);
```
aballway: nit: squish this into one transformation like in the other places this is done

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -518,69 +521,73 @@ - (void)display {
...
             }
-        }
 
-        CGContextReleaseLock(drawContext);
+            CGImageRef target = CGBitmapContextGetImage(drawContext);
+            priv->ownsContents = TRUE;
+            priv->savedContext = CGContextRetain(drawContext);
+            priv->contents = CGImageRetain(target);
+            break;
+        } while (tries < 3);
```
aballway: nit: this is a for loop with a fake nose and a handlebar mustache

aballway: also, no magic number. Maybe stuff it in the context (if we keep the enabled flag in there) or a global that can be set by user

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -518,69 +521,73 @@ - (void)display {
...
-            CGImageRelease(target);
-            CGContextRelease(drawContext);
-            priv->savedContext = NULL;
-            priv->contents = NULL;
-        } else {
-            priv->contents = target;
+        if (!priv->contents) {
+            NSTraceError(TAG, @"Failed to render layer %@", self);
         }
     } else if (priv->contents) {
```
aballway: nit: I know you didn't modify this but please get rid of the if here

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -518,69 +521,73 @@ - (void)display {
...
-            } else {
-                [priv->delegate drawLayer:self inContext:drawContext];
+            woc::StrongCF<CFErrorRef> renderError;
+            if (CGContextIwGetError(drawContext, &renderError)) {
+                switch (CFErrorGetCode(renderError)) {
+                    case kCGContextErrorDeviceReset:
+                        NSTraceInfo(TAG, @"Hardware device disappeared when rendering %@; retrying.", self);
+                        ++tries;
+                        continue;
+                    default: {
```
aballway: nit: use the constant here for this case, and have a default "something went terribly wrong" case.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -2367,6 +2382,10 @@ void CGContextClearRect(CGContextRef context, CGRect rect) {
 HRESULT __CGContext::Draw(_CGCoordinateMode coordinateMode, CGAffineTransform* additionalTransform, Lambda&& drawLambda) {
     auto& state = CurrentGState();
 
+    if (FAILED(_firstErrorHr)) {
```
msft-Jeyaram: nit: I assume S_FALSE is valid to have here?
or should we take care of that?

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -471,21 +475,32 @@ inline void ClearPath() {
...
+        if (!FAILED(_firstErrorHr) && (_beginEndDrawDepth)++ == 0) {
             deviceContext->BeginDraw();
         }
     }
 
     inline HRESULT PopEndDraw() {
+        HRESULT hr = S_OK;
         if (--(_beginEndDrawDepth) == 0) {
-            RETURN_IF_FAILED(deviceContext->EndDraw());
+            hr = deviceContext->EndDraw();
```
msft-Jeyaram: i'm thinking about S_FALSE, looking at the documentation, i don't see much about it for EndDraw(), but given it's technically a success, I assume we don't have to worry about it?

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -2402,12 +2421,6 @@ void CGContextClearRect(CGContextRef context, CGRect rect) {
     }
 
     auto cleanup = wil::ScopeExit([this, layer]() {
```
msft-Jeyaram: > auto cleanup = wil::ScopeExit([this, layer]() { [](start = 3, length = 48)

no longer needed?

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -518,69 +521,73 @@ - (void)display {
...
 
-        if (priv->_backgroundColor == nil || (int)[static_cast<UIColor*>(priv->_backgroundColor) _type] == solidBrush) {
-            CGContextClearToColor(drawContext,
-                                  priv->backgroundColor.r,
-                                  priv->backgroundColor.g,
-                                  priv->backgroundColor.b,
-                                  priv->backgroundColor.a);
-        } else {
-            CGContextClearToColor(drawContext, 0, 0, 0, 0);
+                CGRect wholeRect = CGRectMake(0, 0, width, height);
```
msft-Jeyaram: >ect wholeRect = CGRectMake(0, 0, width, height); [](start = 19, length = 48)

nit: you can directly pass it

include/CoreGraphics/CGContext.h
```diff
@@ -261,4 +261,22 @@ COREGRAPHICS_EXPORT CGPoint CGContextConvertPointToUserSpace(CGContextRef c, CGP
...
+COREGRAPHICS_EXPORT CGRect CGContextConvertRectToUserSpace(CGContextRef c, CGRect rect);
+
+// [WinObjC Extension]
+// CGContextIwEnableEnhancedErrorHandling allows a developer to turn on advanced context error
+// handling. A CGContext that encounters errors when rendering will, by default, assert and abort.
+// Enhanced handling both silences asserts and enables an interested consumer to determine whether a
+// context has encountered a fatal drawing error.
+// A developer using enhanced error handling MUST check the status of a completed set of drawing
+// operations using CGContextIwGetError() and determine whether to redraw the frame.
+COREGRAPHICS_EXPORT void CGContextIwEnableEnhancedErrorHandling(CGContextRef context);
```
msft-Jeyaram: maybe we should have some tag to indicate we have an extension?

Also it would make it easier to look at all the extensions we support/have

include/CoreGraphics/CGContext.h
```diff
@@ -261,4 +261,22 @@ COREGRAPHICS_EXPORT CGPoint CGContextConvertPointToUserSpace(CGContextRef c, CGP
...
+
+COREGRAPHICS_EXPORT const CFStringRef kCGErrorDomainIslandwood;
+
+typedef CF_ENUM(CFIndex, CGContextIwErrorCode) {
+    kCGContextErrorDeviceReset,
```
msft-Jeyaram: nit: set the values explicitly to reduce future mishaps (if it were to ever happen)

include/CoreGraphics/CGContext.h
```diff
@@ -261,4 +261,22 @@ COREGRAPHICS_EXPORT CGPoint CGContextConvertPointToUserSpace(CGContextRef c, CGP
...
+    kCGContextErrorInvalidParameter,
+};
+
+COREGRAPHICS_EXPORT bool CGContextIwGetError(CGContextRef context, CFErrorRef* /* returns-retained */ error);
```
msft-Jeyaram: >/* returns-retained */ [](start = 79, length = 22)

since this is a public api and we are adding it, we should add c/c++ style document to it, to explicitly say what it returns and what's going on.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -471,21 +475,32 @@ inline void ClearPath() {
...
+                // If we haven't yet stored an error, and we're about to return an error (not S_OK), store it.
+                _firstErrorHr = hr;
+                hr = S_OK;
+            }
         }
-        return S_OK;
+        return hr;
+    }
+
+    inline void EnableEnhancedErrorHandling() {
```
DHowett-MSFT: @aballway no; external bitmap context consumers are not enlightened as to error handling.

Frameworks/CoreGraphics/CGContext.mm
DHowett-MSFT: `S_FALSE` is a valid "success" return :)

Frameworks/QuartzCore/CALayer.mm
ms-jihua: I don't know about exposing it, but I think giving it a name/placing it next to the declaration of tries might be slightly more readable.

Frameworks/QuartzCore/CALayer.mm
aballway: aye but it could also be "something went terribly horribly wrong" where we don't even have a name for it 😜

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -475,20 +475,22 @@ inline void ClearPath() {
...
     inline bool ShouldDraw() {
-        return FAILED(_firstErrorHr) || CurrentGState().ShouldDraw();
+        return SUCCEEDED(_firstErrorHr) && CurrentGState().ShouldDraw();
     }
 
     inline void PushBeginDraw() {
-        if (!FAILED(_firstErrorHr) && (_beginEndDrawDepth)++ == 0) {
-            deviceContext->BeginDraw();
+        if ((_beginEndDrawDepth)++ == 0) {
+            if (__builtin_expect(SUCCEEDED(_firstErrorHr), true)) {
```
ms-jihua: fancy

DHowett-MSFT:
The doc comment format here matches that used in CoreFoundation (!)

Frameworks/CoreGraphics/CGContext.mm
ms-jihua: still here

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -2917,25 +2934,46 @@ void CGContextClearToColor(CGContextRef context, CGFloat r, CGFloat g, CGFloat b
...
-    UNIMPLEMENTED();
+bool __CGContext::GetError(CFErrorRef* /* returns-retained */ outError) {
+    RETURN_RESULT_IF(!_useEnhancedErrorHandling, false);
+
+    HRESULT hr = _firstErrorHr;
+    if (SUCCEEDED(hr)) {
+        return false;
+    }
+
+    CGContextIwErrorCode errorCode = kCGContextErrorInvalidParameter;
```
aballway: nit: nice case for a ternary

Frameworks/CoreGraphics/CGContext.mm
DHowett-MSFT: oof

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -85,6 +86,10 @@
 NSString* const kCAFilterNearest = @"kCAFilterNearest";
 NSString* const kCAFilterTrilinear = @"kCAFilterTrilinear";
 
+// The number of rendering attempts a CALayer will make when
+// its backing device disappears.
+static const unsigned int _kCALayerRenderAttempts = 3;
```
msft-Jeyaram: super nit: _skCA.. ?

aballway: or s_kCA... or sc_CA...

DHowett-MSFT: How about `static_const_CFStyleConstantWithAK_CALayer...`

aballway: or just inline the value since it's only used in one place 😉

DHowett-MSFT:
That can't be good.

```
2017-03-01T23:15:21.8217323Z ##[error]frameworks\coregraphics\cgcontext.mm(482,0): Error C1001: An internal error has occurred in the compiler.
```

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -471,22 +474,36 @@ inline void ClearPath() {
...
 
     inline bool ShouldDraw() {
-        return CurrentGState().ShouldDraw();
+        return SUCCEEDED(_firstErrorHr) && CurrentGState().ShouldDraw();
     }
 
     inline void PushBeginDraw() {
         if ((_beginEndDrawDepth)++ == 0) {
-            deviceContext->BeginDraw();
+            if (__builtin_expect(SUCCEEDED(_firstErrorHr), true)) {
```
rajsesh-msft: other than the ICE, there is really no need to use __builtin_expect to optimize here - this is not a hot code path :).

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -2917,25 +2934,47 @@ void CGContextClearToColor(CGContextRef context, CGFloat r, CGFloat g, CGFloat b
...
+bool CGContextIsPointInPath(CGContextRef context, bool eoFill, CGFloat x, CGFloat y) {
     NOISY_RETURN_IF_NULL(context, StubReturn());
-    return true;
+    UNIMPLEMENTED();
+    return StubReturn();
 }
+#pragma endregion
+
+#pragma region Enhanced Error Handling
+const CFStringRef kCGErrorDomainIslandwood = CFSTR("kCGErrorDomainIslandwood");
```
rajsesh-msft: I like standardizing on WinOBJC :)

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -518,69 +525,73 @@ - (void)display {
...
+            woc::StrongCF<CFErrorRef> renderError;
+            if (CGContextIwGetError(drawContext, &renderError)) {
+                switch (CFErrorGetCode(renderError)) {
+                    case kCGContextErrorDeviceReset:
+                        NSTraceInfo(TAG, @"Hardware device disappeared when rendering %@; retrying.", self);
+                        ++tries;
+                        continue;
+                    default: {
+                        NSTraceError(TAG, @"Error %@ rendering %@.", renderError.get(), self);
+                        break;
```
rajsesh-msft: Fail_fast here instead.  It would give us enough to up the number of retries. Another thing to note is that we should probably have another counter that falls back to warp if the particular device fails many times.  That is, if each frame takes 2 attempts to draw, we are probably much better off using warp.  This of course isn't necessary until we move to h/w acceleration.

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -577,7 +577,10 @@ - (void)display {
                         ++tries;
                         continue;
                     default: {
-                        NSTraceError(TAG, @"Error %@ rendering %@.", renderError.get(), self);
+                        FAIL_FAST_MSG("Failed to render <%hs %p>: %hs",
```
DHowett-MSFT: The error description can be augmented to include the failing HRESULT, printing it plus the layer should allow us to correlate events and layer classes.

rajsesh-msft:
Nit: Could you update the description with Fixes# before merge?

Frameworks/CoreGraphics/CGContext.mm
rajsesh-msft: I promise I removed the comment above before submitting my review.  :/

DHowett-MSFT:
@msft-jeyaram: Whichever of this pull request or #2133 merges second will have to add error handling to `__CGContext::ClearRect`

DHowett-MSFT:
@msftfox: Whichever of this pull request or #2019 merges second will have to endure a merge conflict.

DHowett-MSFT:
Ugh, the build agent is stuck downloading LFS file again.

DHowett-MSFT:
The latest change here adds this cool thing:

```
E/WIL: E:\src\WinObjC\Frameworks\QuartzCore\CALayer.mm(583)\(null)!01E2C3CE:
    (caller: 01E2B337) FailFast(1) tid(5438) 8000FFFF Catastrophic failure
    Msg:[Failed to render <CALayer 13A3A984>:
    Error Domain=kCGErrorDomainIslandwood Code=2 "(null)"
    UserInfo={hresult=-2003238902}] [-[CALayer display]]
```

The important bit being `hresult=-2003238902` :smile:

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -538,15 +538,7 @@ - (void)display {
...
-            if (priv->_backgroundColor == nil || (int)[static_cast<UIColor*>(priv->_backgroundColor) _type] == solidBrush) {
-                CGContextClearToColor(drawContext,
-                                      priv->backgroundColor.r,
-                                      priv->backgroundColor.g,
-                                      priv->backgroundColor.b,
-                                      priv->backgroundColor.a);
-            } else {
-                CGContextClearToColor(drawContext, 0, 0, 0, 0);
-
+            if (priv->_backgroundColor != nil && (int)[static_cast<UIColor*>(priv->_backgroundColor) _type] != solidBrush) {
```
DHowett-MSFT: @msftfox this is straight from your pull request. however, i'm concerned about what it's checking.

It looks like it will NOT fill the background if there *is* a solid color/solid brush background color?

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -524,51 +531,69 @@ - (void)display {
             return;
         }
 
-        // Create the contents
-        CGContextRef drawContext = CreateLayerContentsBitmapContext32(width, height, priv->contentsScale);
+        unsigned int tries = 0;
```
aballway: nit: for (unsigned int tries = 0; tries < _kCA...; ++tries)

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -524,51 +531,69 @@ - (void)display {
...
+                    case kCGContextErrorDeviceReset:
+                        NSTraceInfo(TAG, @"Hardware device disappeared when rendering %@; retrying.", self);
+                        ++tries;
+                        continue;
+                    default: {
+                        FAIL_FAST_MSG("Failed to render <%hs %p>: %hs",
+                                      object_getClassName(self),
+                                      self,
+                                      [[static_cast<NSError*>(renderError.get()) debugDescription] UTF8String]);
+                        break;
```
aballway: nit: do we need a break if we're FAIL_FAST?

PR #2129, opened 2017-03-01
Fix#1973: rt_dynamic_cast should only be visible in objc++

mukhole:
rt_dynamic_cast uses COM objects and thus cannot be used from ObjC and therefore, should only visible to ObjC++.


mukhole:
agreed.

PR #2132, opened 2017-03-01
Fix #1973:Fixed crash in rt_dynamic_cast caused by nil objects

### LINKED ISSUE #1973: Using rt_dynamic_cast() and WFIPropertyValue objects.
I'm using `[WSFIStorageItemExtraProperties retrievePropertiesAsync:success:failure:]` to retrieve specific properties for a WSStorageFile.

The code looks something like:
```
 [self.file.properties retrievePropertiesAsync:@[ @"System.Comment" ] success:^(NSDictionary *result) {
  if ([result objectForKey:@"System.Comment"]) {
    WFIPropertyValue *value = rt_dynamic_cast([WFIPropertyValue class], [result objectForKey:@"System.Comment"]);
  }
}
failure:^(NSError *error){}]; 
```

For some files the `rt_dynamic_cast()` causes a crash. I've tracked this down to the fact that the `[RTObject comObj]` property for the result value in some cases is null.

The `comObj` property is only accessible from objc++ and I was just wandering if there is a way to safely check if `rt_dynamic_cast()` will work from a objc class?

In `InteropBase.h` it mentions
> Does a safe cast of rtObject into a derived projected class type. Throws if it is an invalid cast. Throws if it is an invalid cast.

But it doesn't seem to throw a catchable execption in this case.

mukhole:


include/Platform/Universal Windows/UWP/InteropBase.h
```diff
@@ -51,7 +51,9 @@ WINRT_EXPORT
 
 @end
 
-// Does a safe cast of rtObject into a derived projected class type. Throws if it is an invalid cast.
+// Does a safe cast of rtObject into a derived projected class type.
+// Returns nil if cast fails.
```
pradipd: >Returns nil if cast fails. [](start = 3, length = 26)

Thanks for updating the comment.

pradipd:
:shipit:

PR #2133, opened 2017-03-02
Implementation of ClearRect with masking support

msft-Jeyaram:


msft-Jeyaram:
Note: the second iteration is the change from the last CR.
The implementation was correct, it was the way how tests were that caused some failures, which are fixed now.

@DHowett-MSFT please take a look at the new heart art :)

msft-Jeyaram:
Looks like git lfs is failing on build machines

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -2297,22 +2300,46 @@ void CGContextShowGlyphsWithAdvances(CGContextRef context, const CGGlyph* glyphs
 #pragma endregion
 
 #pragma region Drawing Operations - Basic Shapes
+
+HRESULT __CGContext::ClearRect(CGRect rect) {
+    PushBeginDraw();
+    auto endDraw = wil::ScopeExit([this]() { PopEndDraw(); });
```
DHowett-MSFT: prefer `return PopEndDraw()` at the end; otherwise, we lose the HRESULT from EndDraw.

msft-Jeyaram:
so on the build machine, the generated image for one of the tests differ by a pixel :(
making it such that we check if the pixel is present.

msft-Jeyaram:
Verbal extra sign off by @DHowett-MSFT  merging .

PR #2134, opened 2017-03-02
Improve CAShapeLayer

### LINKED ISSUE #2039: CAShapeLayer does an end run around CALayer rendering
Instead of keeping track of whether it is dirty and using the CALayer rendering pipeline, it renders its own bitmap into its own bitmap context and splats that into ANOTHER layer inside it.

Additionally, it does its own state-keeping to determine whether it is dirty instead of using `setNeedsDisplay`.

Active impediment to work on #2038.

DHowett-MSFT:
This pull request fixes **CAShapeLayer** the issues mentioned in #2039, plus adds support for EO fill and line caps.

CAShapeLayer originally used a nested layer that it manually poked an image into; doing this is incompatible with #1194 (and therefore #2119). It also had a secret dependency on `CreateLayerContentsBitmapContext32`, a private import from CALayer.

Testing was carried out manually in WOCCatalog on 1.0, 2.0, and 3.5-scale devices.

Fixes #2039.


include/QuartzCore/CAShapeLayer.h
```diff
@@ -34,8 +34,8 @@ CA_EXPORT_CLASS
 
 @property CGPathRef path;
 @property CGColorRef fillColor;
-@property (copy) NSString* fillRule STUB_PROPERTY;
-@property (copy) NSString* lineCap STUB_PROPERTY;
+@property (copy) NSString* fillRule;
```
rajsesh-msft: Does woccatalog have tests for fillRule and lineCap?

Frameworks/QuartzCore/CAShapeLayer.mm
```diff
@@ -14,131 +14,78 @@
...
+@interface CAShapeLayer () {
+    woc::StrongCF<CGPathRef> _path;
+    woc::StrongCF<CGColorRef> _strokeColor;
+    woc::StrongCF<CGColorRef> _fillColor;
 
-@implementation CAShapeLayer {
-    CGPathRef _path;
+    CGPathDrawingMode _fillMode;
+    CGLineCap _lineCap;
     float _lineWidth;
```
msft-Jeyaram: nit: CGFloat

include/QuartzCore/CAShapeLayer.h
aballway: How does (copy) work w/ overriding default getter/setter?

DHowett-MSFT: It doesn't; however, there's no `the object you pass in will be ignored completely, as objects go` annotation; assign would be truer, but copy is what the reference platform specifies and it is expected for all strings.

Since we've implemented it in terms of an enumerated value, we hold true to one of the core tenets of `copy`: The string they pass into the setter and the string they get back from the getter *may* not be the same pointer ;)

rajsesh-msft: To clarify question above - lets add the variation to woccatalog test to demonstrate this works.

DHowett-MSFT:
The new test images look like:

image

DHowett-MSFT:
@rajsesh-msft ping! :smile: I'd like to make sure I address your concerns

PR #2136, opened 2017-03-02
Fix CGPathAddArc and add additional tests.

### LINKED ISSUE #2062: CGContextAddArc does not create arcs as expected
It looks like CGContextAddArc isn't creating the proper arc.

We differ from the reference platform.

Sample:
```
CGContextAddArc(context, 50, 50, 50, 0.0, M_PI, 0);
CGContextFillPath(context);
```

Result:
capture

Tested with various angles and it's not as expected.

MSFTFox:
Correctly translates the CoreGraphics arc parameters into D2D1_ARC_SEGMENT parameters. Added additional testing to cover previously known failure situations.

Fix #2062


Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -150,6 +152,14 @@ CGPoint GetStartingPoint() const {
         return startingPoint;
     }
 
+    bool GetIsFirstCall() {
+        return isFirstCall;
+    }
+
+    void SetIsFirstCall(bool firstCall) {
+        isFirstCall = firstCall;
+    }
```
DHowett-MSFT: I very much do not like this mechanism.

DHowett-MSFT:
Also, does this allow you to enable @msft-jeyaram's arc clearing test?

tests/unittests/CoreGraphics.Drawing/CGPathDrawingTests.cpp
```diff
@@ -291,6 +291,154 @@ DRAW_TEST_F(CGPath, PathApplyDraw, UIKitMimicTest<>) {
...
+            break;
+        case 2:
+            CGPathMoveToPoint(thepath, &transformation, xstart + .75 * width, ystart + .5 * height);
+            break;
+        case 0:
+        default:
+            break;
+    }
+    if (moveToPoint == 0) {
+    }
```
msft-Jeyaram: ? needed?

msft-Jeyaram:
@MSFTFox  where are the generated base images for the tests?

msft-Jeyaram:
@MSFTFox I've also added some FillMode tests for Arcs.
These were parameterized tests that I've added to ensure arcs was working.
Please enable them when #2133 is in.
There is a TODO with issue id #2062 

Also depending on who gets the changes in first, arc clearing tests need to be enabled.
The disabled tests are tagged with the Arc issue ids.

Frameworks/CoreGraphics/CGPath.mm
rajsesh-msft: According to the docs - it is not the "first call", rather the existence of an open sub path - that should be the mechanism used here.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -231,8 +231,12 @@ HRESULT ClosePath() {
         return S_OK;
     }
 
+    bool IsFigureOpen() {
+        return geometrySink->IsFigureOpen();
+    }
+
     void BeginFigure() {
-        if (!geometrySink->IsFigureOpen()) {
+        if (geometrySink && !geometrySink->IsFigureOpen()) {
```
DHowett-MSFT: should this logic be in the above function?

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -482,6 +486,21 @@ static HRESULT _createPathReadyForFigure(CGPathRef previousPath,
     return S_OK;
 }
 
+static CGFloat __normalizeAngle(CGFloat originalAngle) {
```
DHowett-MSFT: A comment explaining what this one is doing would be exceptionally helpful.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -567,6 +570,9 @@ void CGPathAddArcToPoint(
 
 /**
  @Status Interoperable
+ @Notes For the scenario of drawing a full circle, any distance of 2pi will result in a circle being drawn
+        regardless of the direction being drawn. The reference platform however, will not draw a circle
```
DHowett-MSFT: nit: `, however,` (missing preceding comma)

DHowett-MSFT: this is too wordy, and could be trimmed down to a succinct

`All arcs from 0 to 2pi in either direction will result in a circle; the reference platform, by contrast, does not honor this behavior.`

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -482,6 +486,24 @@ static HRESULT _createPathReadyForFigure(CGPathRef previousPath,
...
+    CGFloat returnAngle = fmod(originalAngle, 2 * M_PI);
+    if (returnAngle < .00001) {
+        returnAngle = 0;
+    }
+    if (returnAngle == 0) {
+        if (abs(originalAngle) > 0) {
+            return 2 * M_PI;
+        }
+    }
+    if (returnAngle < 0) {
```
aballway: This case should never happen because you're setting it to 0 above if < .00001

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -482,6 +486,24 @@ static HRESULT _createPathReadyForFigure(CGPathRef previousPath,
     return S_OK;
 }
 
+// This function will return a normalized angle in radians between 0 and 2pi. This is to standardize
+// the calculations for arcs since 0, 2pi, 4pi, etc... are all visually the same angle.
+static CGFloat __normalizeAngle(CGFloat originalAngle) {
+    CGFloat returnAngle = fmod(originalAngle, 2 * M_PI);
+    if (returnAngle < .00001) {
```
aballway: super nit: pull this out into some constant, c_zeroAngleThreshold or something

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -482,6 +486,24 @@ static HRESULT _createPathReadyForFigure(CGPathRef previousPath,
...
+        returnAngle = 0;
+    }
+    if (returnAngle == 0) {
+        if (abs(originalAngle) > 0) {
```
aballway: nit: this would be more clear as (originalAngle != 0) since you're taking the abs

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -565,32 +584,66 @@ void CGPathAddArc(CGMutablePathRef path,
...
     ComPtr<ID2D1GeometrySink> newSink;
+    if (!path->IsFigureOpen()) {
+        CGPathMoveToPoint(path, transform, startPoint.x, startPoint.y);
+    }
     FAIL_FAST_IF_FAILED(_createPathReadyForFigure(path, startPoint, &newPath, &newSink));
 
-    newSink->AddArc(arcSegment);
+    // This will only happen when drawing a circle in the clockwise direction from 2pi to 0, a scenario
+    // supported on the reference platform.
+    if (abs(abs(rawDifference) - 2 * M_PI) < .00001) {
```
aballway: nit: rawDifference is already abs() from above

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -565,32 +584,66 @@ void CGPathAddArc(CGMutablePathRef path,
                   bool clockwise) {
     RETURN_IF(!path);
 
-    CGPoint startPoint = CGPointMake(x + radius * cos(startAngle), y + radius * sin(startAngle));
-    CGPoint endPoint = CGPointMake(x + radius * cos(endAngle), y + radius * sin(endAngle));
+    // Normalize the angles to work with to values between 0 and 2*PI
+    CGFloat normalizedStartAngle = __normalizeAngle(startAngle);
+    CGFloat normalizedEndAngle = __normalizeAngle(endAngle);
+
+    CGPoint startPoint = CGPointMake(x + radius * cos(normalizedStartAngle), y + radius * sin(normalizedStartAngle));
```
aballway: nit: you're calculating radius * sin, cos twice

MSFTFox: They're on different angles.

aballway: 😮 well by golly they are

Frameworks/CoreGraphics/CGPath.mm
MSFTFox: But could still be negative from a counterclockwise rotation. This negative is important later on but the strict value being close enough to zero here is the important factor for when to draw a full circle.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -482,6 +486,24 @@ static HRESULT _createPathReadyForFigure(CGPathRef previousPath,
...
+    if (returnAngle < .00001) {
+        returnAngle = 0;
+    }
+    if (returnAngle == 0) {
+        if (abs(originalAngle) > 0) {
+            return 2 * M_PI;
+        }
+    }
+    if (returnAngle < 0) {
+        returnAngle += M_PI * 2;
```
ms-jihua: super super nit: consistent ordering with your (2*pi)s

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -565,32 +584,66 @@ void CGPathAddArc(CGMutablePathRef path,
...
+    if (!path->IsFigureOpen()) {
+        CGPathMoveToPoint(path, transform, startPoint.x, startPoint.y);
+    }
     FAIL_FAST_IF_FAILED(_createPathReadyForFigure(path, startPoint, &newPath, &newSink));
 
-    newSink->AddArc(arcSegment);
+    // This will only happen when drawing a circle in the clockwise direction from 2pi to 0, a scenario
+    // supported on the reference platform.
+    if (abs(abs(rawDifference) - 2 * M_PI) < .00001) {
+        CGFloat midPointAngle = normalizedStartAngle + rawDifference / 2;
```
ms-jihua: nit: do we wanna just round the latter term to +/- pi here?

Frameworks/CoreGraphics/CGPath.mm
ms-jihua: nit: should this equation be an inline function?

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -482,6 +486,20 @@ static HRESULT _createPathReadyForFigure(CGPathRef previousPath,
     return S_OK;
 }
 
+static const CGFloat sc_zeroAngleThreshold = .00001;
+// This function will return a normalized angle in radians between 0 and 2pi. This is to standardize
+// the calculations for arcs since 0, 2pi, 4pi, etc... are all visually the same angle.
+static CGFloat __normalizeAngle(CGFloat originalAngle) {
+    CGFloat returnAngle = fmod(originalAngle, 2 * M_PI);
```
DHowett-MSFT: I'm not certain what this function is doing anymore.

Does it boil down to

```c++
return std::min(abs(angle), 2 * M_PI);
```

?

MSFTFox: Not quite, since that would ignore any angle like 7pi/2 and instead return 2 *M_PI. We only want to return an angle of 2pi if the angle is a multiple of 2pi and not 0. That's what the extra checks are for.

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_ClearRectTests.cpp
```diff
@@ -141,10 +141,8 @@ DRAW_TEST_P(CGClearRectArc, ClearArc) {
 }
 
 static CGPoint sweep[] = { CGPointMake(0, M_PI), CGPointMake(M_PI, 0), CGPointMake(0.3 * M_PI, 0.6 * M_PI), CGPointMake(0, 1.9 * M_PI) };
-// TODO: enable when #2062 is fixed.
-INSTANTIATE_TEST_CASE_P(DISABLED_CGContextTests,
```
msft-Jeyaram: thanks

MSFTFox: 👍

PR #2139, opened 2017-03-02
Fill/Stroke with color should not crash, if color is null

msft-Jeyaram:


tests/UnitTests/CoreGraphics.drawing/CGContextDrawingTests.cpp
```diff
@@ -213,4 +213,21 @@ DRAW_TEST_F(CGContext, AntialiasToggleTranscendsGState, WhiteBackgroundTest<>) {
         { 15, 5 }, { bounds.size.width, bounds.size.height - 10 },
     };
     CGContextStrokeLineSegments(context, blurryLineSegments2, 2);
-}
\ No newline at end of file
+}
+
+DRAW_TEST_F(CGContext, NullFillColor, WhiteBackgroundTest<>) {
```
msft-Jeyaram: hmm not sure if we need the drawing test.
Thoughts?

aballway: We only need to test that the fill/stroke colors aren't changed when nullptr is passed in correct? Is there a way to get the color from the context to check?

DHowett-MSFT:
Does this hold on the reference platform? I have a post-it on my desk saying that we should support `NULL` colors; that implies to me that it's slightly more complicated than ignoring them.

msft-Jeyaram:
@DHowett-MSFT  this does hold on reference platform.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1650,6 +1650,8 @@ void CGContextSetLineWidth(CGContextRef context, CGFloat width) {
 */
 void CGContextSetStrokeColor(CGContextRef context, const CGFloat* components) {
     NOISY_RETURN_IF_NULL(context);
+    NOISY_RETURN_IF_NULL(components);
```
msft-Jeyaram: for these it uses the existing value.
It's covered in the test.

PR #2148, opened 2017-03-03
Fix some missed CGContext API annotations.

DHowett-MSFT:
(This also nukes CGContextGetBlendMode, which was completely unused and not a public API.)

### DIFF
--- a/Frameworks/CoreGraphics/CGContext.mm
+++ b/Frameworks/CoreGraphics/CGContext.mm
@@ -3045,7 +3045,10 @@ private:
 
 /**
  @Status Caveat
- We only support formats that are 32 bits per pixel, colorspace and bitmapinfo that are ARGB.
+ @Notes If data is provided, it can only be in one of the few pixel formats Direct2D can render to in system memory:
+        (P)RGBA, (P)BGRA, or Alpha8.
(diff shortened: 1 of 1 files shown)


msft-Jeyaram:
>  We only support formats that are 32 bits per pixel, colorspace and bitmapinfo that are ARGB.

nit: Notes

---
Refers to: Frameworks/CoreGraphics/CGContext.mm:3048 in d66f2b5. [](commit_id = d66f2b5b1d6d42dfbac39f650762e2ec4afe89e2, deletion_comment = False)

DHowett-MSFT:
Funny enough, the comment isn’t even correct.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1088,7 +1085,6 @@ void CGContextAddLines(CGContextRef context, const CGPoint* points, unsigned cou
 #pragma region Global State - Path Queries
 /**
  @Status Interoperable
- @Notes
 */
 CGPathRef CGContextCopyPath(CGContextRef context) {
     NOISY_RETURN_IF_NULL(context, nullptr);
```
rajsesh-msft: Please remove the TODO's in this file while you are at it.

PR #2165, opened 2017-03-06
Force all loaded textures to 32bpp in GLKit

### LINKED ISSUE #2150: [ARM] [x86] GLKitComplex images are missing
On the most recent master build - 20170228.1

Image on the Teapot is missing for these scenes - 	
		Vertex Lit, Specularity, Texture
		Solid Color, Textured w/ Decal	 

And - Vertex Colors, Textured is black and white and the Text “Tea” is missing on the labels on the tea pot

msft-Jeyaram:
GLKit's implementation of GLKTexture only handled an image format of 32bpp properly (other formats e.g 8bpp indexed,etc) was not handled properly (the code was never exercised).

The root cause of this was that the earlier Cairo implementation treated every image as a 32bpp. Thus when we created an image with it's proper pixel format, GLKTexture code didn't know how to handle it and thus the errors were produced.

fixes #2150

The fix converts the images to 32bpp, since GLKTexture does not support any other format properly.
Eventually, GLKTexure needs to be re-written properly. 

Current Stage: 
image

image

image

image

image

image

### DIFF
--- a/Frameworks/CoreGraphics/CGImage.mm
+++ b/Frameworks/CoreGraphics/CGImage.mm
@@ -941,19 +941,8 @@ HRESULT _CGImageConvertToMaskCompatibleWICBitmap(CGImageRef image, IWICBitmap** 
 CGImageRef _CGImageCreateFromFileWithWICFormat(CFStringRef filename, WICPixelFormatGUID format) {
     RETURN_NULL_IF(!filename);
 
-    std::vector<char> buffer;
-    const char* fname = CFStringGetCStringPtr(filename, kCFStringEncodingUTF8);
-    if (fname == nullptr) {
-        CFIndex length = CFStringGetLength(filename) + 1;
-        buffer.reserve(length);
-        if (CFStringGetCString(filename, buffer.data(), length, kCFStringEncodingUTF8)) {
-            fname = buffer.data();
-        }
-    }
-
-    RETURN_NULL_IF(!fname);
-
-    woc::StrongCF<CGDataProviderRef> provider{ woc::MakeStrongCF(CGDataProviderCreateWithFilename(fname)) };
+    woc::StrongCF<CFURLRef> url{ woc::MakeStrongCF(CFURLCreateWithFileSystemPath(nullptr, filename, kCFURLWindowsPathStyle, NO)) };
+    woc::StrongCF<CGDataProviderRef> provider{ woc::MakeStrongCF(CGDataProviderCreateWithURL(url)) };
 
     woc::StrongCF<CGImageRef> image{ woc::MakeStrongCF(_CGImageCreateFromDataProvider(provider)) };
 
--- a/Frameworks/CoreGraphics/CGImage.mm
+++ b/Frameworks/CoreGraphics/CGImage.mm
@@ -934,4 +934,30 @@ HRESULT _CGImageConvertToMaskCompatibleWICBitmap(CGImageRef image, IWICBitmap** 
     return _CGImageGetWICImageSource(convertedImage.get(), pBitmap);
 }
 
+/**
+* Creates an image from file, if the image is not in the requested format, it is converted to
+* the requested format and returned.
+*/
+CGImageRef _CGImageCreateFromFileWithWICFormat(CFStringRef filename, WICPixelFormatGUID format) {
+    RETURN_NULL_IF(!filename);
+
+    std::vector<char> buffer;
+    const char* fname = CFStringGetCStringPtr(filename, kCFStringEncodingUTF8);
+    if (fname == nullptr) {
+        CFIndex length = CFStringGetLength(filename) + 1;
+        buffer.reserve(length);
+        if (CFStringGetCString(filename, buffer.data(), length, kCFStringEncodingUTF8)) {
+            fname = buffer.data();
+        }
+    }
+
+    RETURN_NULL_IF(!fname);
+
+    woc::StrongCF<CGDataProviderRef> provider{ woc::MakeStrongCF(CGDataProviderCreateWithFilename(fname)) };
+
+    woc::StrongCF<CGImageRef> image{ woc::MakeStrongCF(_CGImageCreateFromDataProvider(provider)) };
+
+    return _CGImageCreateCopyWithPixelFormat(image, format);
+}
+
 #pragma endregion WIC_HELPERS
--- a/Frameworks/GLKit/GLKTexture.mm
+++ b/Frameworks/GLKit/GLKTexture.mm
@@ -26,6 +26,7 @@
 #import <GLKit/GLKTexture.h>
 #import "NSLogging.h"
 #import "CGImageInternal.h"
+#import "CGDataProviderInternal.h"
 #import <Starboard/SmartTypes.h>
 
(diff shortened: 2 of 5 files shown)


Frameworks/GLKit/GLKTexture.mm
```diff
@@ -274,18 +275,26 @@ @implementation GLKTextureLoader {
...
 
 /**
- @Status Interoperable
+* Creates an image from file, if the image is not a GUID_WICPixelFormat32bppPRGBA format, it is converted to
+* GUID_WICPixelFormat32bppPRGBA and returned.
 */
-+ (GLKTextureInfo*)textureWithContentsOfFile:(NSString*)fname options:(NSDictionary*)opts error:(NSError**)err {
-    CGDataProviderRef provider = CGDataProviderCreateWithFilename([fname UTF8String]);
-    CGImageRef img = _CGImageCreateFromDataProvider(provider);
+static inline CGImageRef __Create32bppPRGBAImageFromFile(NSString* fname) {
```
rajsesh-msft: _CGImageCreateFromFileWithFormat(name, format) and export it from CG?

Frameworks/GLKit/GLKTexture.mm
```diff
@@ -299,7 +308,13 @@ + (GLKTextureInfo*)textureWithContentsOfData:(NSData*)data options:(NSDictionary
 /**
  @Status Interoperable
 */
-+ (GLKTextureInfo*)textureWithCGImage:(CGImageRef)img options:(NSDictionary*)opts error:(NSError**)err {
++ (GLKTextureInfo*)textureWithCGImage:(CGImageRef)image options:(NSDictionary*)opts error:(NSError**)err {
+    woc::StrongCF<CGImageRef> img = { woc::MakeStrongCF(_CGImageCreateCopyWithPixelFormat(image, GUID_WICPixelFormat32bppPRGBA)) };
+
+    if (!img) {
```
rajsesh-msft: no macro here?

Frameworks/GLKit/GLKTexture.mm
msft-Jeyaram: I need like we are bloating CG. we already have an internal method, and we keep on building on top of it for every case.

msft-Jeyaram: but i'll go ahead and make that change

rajsesh-msft: CG has 600+ apis :). What belongs there has to belong there...

DHowett-MSFT: nit: if you do make this API, please perhaps call it `_CGImageCreateFromFileWithWICFormat` -- specifically, we need to differentiate it from the CGBitmapInfo/CGAlphaInfo APIs.

DHowett-MSFT: And it'll have to take a CFStringRef.

Frameworks/GLKit/GLKTexture.mm
```diff
@@ -414,13 +421,10 @@ + (GLKTextureInfo*)cubeMapWithContentsOfFile:(NSString*)fname options:(NSDiction
...
     if (h != 6 * w) {
         NSTraceError(TAG, @"ERROR - Unexpected cube map format, expected 6 square textures aligned vertically.");
-        CGImageRelease(img);
-        CGDataProviderRelease(provider);
         return nil;
     }
 
-    CGDataProviderRelease(provider);
-    provider = CGImageGetDataProvider(img);
+    CGDataProviderRef provider = CGImageGetDataProvider(img);
```
DHowett-MSFT: Since you are fixing this code, you might as well make it use `_CGDataProviderGetBytes`.

DHowett-MSFT:
Also nit: When you merge this, please fix the name to be `GLKit`, not `GLKKit`

DHowett-MSFT:
I have renamed this pull request. Please add "Fixes #2150."

msft-Jeyaram:
I think taking in char * would be fine, given that's what CGDataProviderCreateWithFilename takes not.
so no need for CFString

DHowett-MSFT:
@msft-Jeyaram please do not do that. We have strong, expressive string types for a reason.

NSString is bridged directly to CFStringRef; unless you want to deal with UTF8<->UTF16 issues like #1875 and #1891 I would _implore_ you to use CFString.

msft-Jeyaram:
fixes #2150

Frameworks/CoreGraphics/CGImage.mm
```diff
@@ -934,4 +934,30 @@ HRESULT _CGImageConvertToMaskCompatibleWICBitmap(CGImageRef image, IWICBitmap**
...
+        CFIndex length = CFStringGetLength(filename) + 1;
+        buffer.reserve(length);
+        if (CFStringGetCString(filename, buffer.data(), length, kCFStringEncodingUTF8)) {
+            fname = buffer.data();
+        }
+    }
+
+    RETURN_NULL_IF(!fname);
+
+    woc::StrongCF<CGDataProviderRef> provider{ woc::MakeStrongCF(CGDataProviderCreateWithFilename(fname)) };
```
aballway: super nit: these can be "auto provider = "

Frameworks/GLKit/GLKTexture.mm
```diff
@@ -309,10 +311,8 @@ + (GLKTextureInfo*)textureWithCGImage:(CGImageRef)img options:(NSDictionary*)opt
         NSTraceWarning(TAG, @"WARNING - Image (%dx%d) - expected row size %d, got row size %d\n", w, h, expectedRowSize, rowSize);
     }
 
-    CGDataProviderRef provider = CGImageGetDataProvider(img);
-    NSData* data = (id)CGDataProviderCopyData(provider);
-    [data autorelease];
-    auto bytesIn = (unsigned char*)[data bytes];
+    unsigned char* bytesIn = static_cast<unsigned char*>(const_cast<void*>(_CGDataProviderGetData(CGImageGetDataProvider(img))));
```
aballway: super nit: a c-style cast from const void* -> char* should suffice

Frameworks/CoreGraphics/CGImage.mm
```diff
@@ -934,4 +934,30 @@ HRESULT _CGImageConvertToMaskCompatibleWICBitmap(CGImageRef image, IWICBitmap**
...
 }
 
+/**
+* Creates an image from file, if the image is not in the requested format, it is converted to
+* the requested format and returned.
+*/
+CGImageRef _CGImageCreateFromFileWithWICFormat(CFStringRef filename, WICPixelFormatGUID format) {
+    RETURN_NULL_IF(!filename);
+
+    std::vector<char> buffer;
```
aballway: On second look, we'd be better off here creating a CFURL from the filename and using CGDataProviderCreateWithURL, because that's what CGDataProviderCreateWithFilename goes through, and we get the added bonus of not creating an extra CFStringRef as well as cutting down on the length of this function quite a bit.

msft-Jeyaram: and here was me thinking it was the other way around.

PR #2170, opened 2017-03-06
Add logic to use latest tag instead of fork point which doesn't work …

bbowman:
…for some reason in our release methodology.


rajsesh-msft:
:shipit:

memontic-ms:
:shipit:

DHowett-MSFT:
I take it that’s a no? :P

PR #2171, opened 2017-03-06
Implement CTFontDescriptorCreateMatchingFontDescriptor(s)

### LINKED ISSUE #2024: Implement CTFontDescriptorCreateMatchingFontDescriptor and CTFontDescriptorCreateMatchingFontDescriptors

aballway:
CTFontDescriptorCreateMatchingFontDescriptor(s) on the reference platform can return unexpected and inconsistent results, so rather than matching those we are going for a consistent return policy.

Any supported attributes in the mandatoryAttributes set will be required for all matching fonts, so any returned font descriptors will have those attributes.  If no fonts match the attributes, no descriptors will be returned.

Any non-mandatory attributes will attempt to be matched, so if there are still descriptors that match the mandatory and an additional we will only return descriptors that match both.

Fixes #2024


Frameworks/CoreGraphics/DWriteFontCollectionHelper.mm
```diff
@@ -130,6 +130,32 @@
     return S_FALSE;
 }
 
+HRESULT DWriteFontCollectionHelper::GetAvailableFonts(std::vector<ComPtr<IDWriteFont>>& outFontList) {
+    std::lock_guard<std::recursive_mutex> lock(m_lock);
+
+    ComPtr<IDWriteFontCollection> fontCollection = GetFontCollection();
+    if (!fontCollection) {
+        return S_OK;
+    }
```
ms-jihua: RETURN_HR_IF_NULL

Frameworks/CoreGraphics/DWriteFontCollectionHelper.mm
```diff
@@ -130,6 +130,32 @@
...
+
+    uint32_t count = fontCollection->GetFontFamilyCount();
```
ms-jihua: nit: more specific name, const

Frameworks/CoreGraphics/DWriteFontCollectionHelper.mm
```diff
@@ -130,6 +130,32 @@
...
+    for (uint32_t i = 0; i < count; ++i) {
+        ComPtr<IDWriteFontFamily> fontFamily;
+        RETURN_IF_FAILED(fontCollection->GetFontFamily(i, &fontFamily));
+        ComPtr<IDWriteFontList> fontList;
+        RETURN_IF_FAILED(
+            fontFamily->GetMatchingFonts(DWRITE_FONT_WEIGHT_NORMAL, DWRITE_FONT_STRETCH_NORMAL, DWRITE_FONT_STYLE_NORMAL, &fontList));
+        uint32_t fontCount = fontList->GetFontCount();
```
ms-jihua: nit: more specific name, const

Frameworks/UIKit/UIFontDescriptor.mm
```diff
@@ -173,8 +173,7 @@ - (instancetype)initWithFontAttributes:(NSDictionary*)attributes {
  @Status Stub
```
ms-jihua: Caveat?

Frameworks/CoreGraphics/DWriteFontCollectionHelper.mm
```diff
@@ -130,6 +130,32 @@
     return S_FALSE;
 }
 
+HRESULT DWriteFontCollectionHelper::GetAvailableFonts(std::vector<ComPtr<IDWriteFont>>& outFontList) {
```
DHowett-MSFT: I don't entirely like this return-through-reference-param pattern.

Frameworks/CoreText/CTFontDescriptor.mm
```diff
@@ -182,21 +182,32 @@ CTFontDescriptorRef CTFontDescriptorCreateCopyWithFeature(CTFontDescriptorRef or
 }
 
 /**
- @Status Stub
- @Notes
+ @Status Caveat
+ @Notes Will have different return values than reference platform due to reference platform's inconsistency
+        This implementation will return fonts that match all mandatoryAttributes, and will attempt to reduce the list by any other
+ attributes in the descriptor, but will no do so if no font descriptors match
```
msft-Jeyaram: >no [](start = 40, length = 2)

not?

DHowett-MSFT: nit: `will not do` instead of `will no do`

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -208,6 +208,173 @@ HRESULT _DWriteCreateFontFaceWithFontDescriptor(CTFontDescriptorRef fontDescript
     return E_INVALIDARG;
 }
 
+// Helper methods for removing fonts that do not match given attributes for CTFontDescriptorCreateMatchingFontDescriptors
+#pragma region Helper Filter Methods
+
+void __FilterFontsByTraits(std::vector<ComPtr<IDWriteFont>>& fonts, CFDictionaryRef traits, bool forceMatch) {
```
DHowett-MSFT: double underscore, but not static?

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -208,6 +208,173 @@ HRESULT _DWriteCreateFontFaceWithFontDescriptor(CTFontDescriptorRef fontDescript
...
+    }
+
+    if (!matchingFonts.empty() || forceMatch) {
+        fonts = std::move(matchingFonts);
+    }
+
+    return S_OK;
+}
+
+HRESULT __FilterFontsByInformationalString(std::vector<ComPtr<IDWriteFont>>& fonts,
```
DHowett-MSFT: these four functions are largely the same. can they be deduplicated somehow?

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -208,6 +208,173 @@ HRESULT _DWriteCreateFontFaceWithFontDescriptor(CTFontDescriptorRef fontDescript
...
+    _DWriteFontProperties properties = __DWriteFontPropertiesFromTraits(traits);
+    std::vector<ComPtr<IDWriteFont>> matchingFonts;
+    matchingFonts.reserve(fonts.size());
+    for (auto& font : fonts) {
+        if (font->GetWeight() == properties.weight && font->GetStretch() == properties.stretch && font->GetStyle() == properties.style) {
+            matchingFonts.emplace_back(font);
+        }
+    }
+
+    if (!matchingFonts.empty() || forceMatch) {
```
ms-jihua: should probably comment here about defaulting to no-filtering if there's no match and forceMatch isn't true

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -208,6 +208,173 @@ HRESULT _DWriteCreateFontFaceWithFontDescriptor(CTFontDescriptorRef fontDescript
...
+        return S_FALSE;
+    }
+}
+
+#pragma endregion // Helper Filter Methods
+
+/**
+ * Creates a CFArray of CTFontDescriptors values for all fonts that match the given attributes best and completely match mandatoryKeys
+ */
+HRESULT _DWriteCreateMatchingFontDescriptors(CFDictionaryRef attributes, CFSetRef mandatoryKeys, CFArrayRef* matchingNames) {
```
msft-Jeyaram: >_DWriteCreateMatchingFontDescriptors [](start = 8, length = 36)

 you should guard these, as they are at the lowest level in terms of this function.
In the future if someone were to call this with a pointer that ends up being invalid, we will blow up.
We saw this happen in CG ;).

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -208,6 +208,173 @@ HRESULT _DWriteCreateFontFaceWithFontDescriptor(CTFontDescriptorRef fontDescript
...
+        CFIndex count = CFSetGetCount(mandatoryKeys);
+        std::vector<CFStringRef> keys(count);
+        CFSetGetValues(mandatoryKeys, (const void**)keys.data());
+        for (auto& key : keys) {
+            if (!CFDictionaryContainsKey(attributes, key)) {
+                // Mandatory key does not exist, so no fonts can match
+                return S_OK;
+            }
+
+            auto value = woc::MakeAutoCF<CFTypeRef>(CFDictionaryGetValue(attributes, key));
```
DHowett-MSFT: !bad!

This will overrelease the dictionary member value.

DHowett-MSFT: just use an unguarded local. no need to retain it and then release it across a single sequence point.

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -208,6 +208,173 @@ HRESULT _DWriteCreateFontFaceWithFontDescriptor(CTFontDescriptorRef fontDescript
...
+        std::vector<CFTypeRef> values(count);
+        CFDictionaryGetKeysAndValues(attributes, (const void**)keys.data(), (const void**)values.data());
+        for (size_t i = 0; i < count; ++i) {
+            if (!mandatoryKeys || !CFSetContainsValue(mandatoryKeys, keys[i])) {
+                RETURN_IF_FAILED(__FilterFontsByKeyAndValue(fonts, keys[i], values[i], false));
+            }
+        }
+    }
+
+    std::unique_ptr<CFTypeRef[]> outDescriptors(new CFTypeRef[fonts.size()]);
```
DHowett-MSFT: `unique_iw`?

DHowett-MSFT: vector?

DHowett-MSFT: this is just a vector in disguise.

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -208,6 +208,173 @@ HRESULT _DWriteCreateFontFaceWithFontDescriptor(CTFontDescriptorRef fontDescript
...
+    for (size_t i = 0; i < fonts.size(); ++i) {
+        BOOL exists;
+        ComPtr<IDWriteLocalizedStrings> dwriteString;
+        RETURN_IF_FAILED(fonts[i]->GetInformationalStrings(DWRITE_INFORMATIONAL_STRING_POSTSCRIPT_NAME, &dwriteString, &exists));
+        CFStringRef name = _CFStringFromLocalizedString(dwriteString.Get());
+        outDescriptors[i] = CFAutorelease(CTFontDescriptorCreateWithNameAndSize(name, kCTFontSystemFontSize));
```
DHowett-MSFT: CFAutorelease is not really a great API to use; can we avoid it somehow?

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -208,6 +208,173 @@ HRESULT _DWriteCreateFontFaceWithFontDescriptor(CTFontDescriptorRef fontDescript
...
+            }
+        }
+    }
+
+    std::unique_ptr<CFTypeRef[]> outDescriptors(new CFTypeRef[fonts.size()]);
+    for (size_t i = 0; i < fonts.size(); ++i) {
+        BOOL exists;
+        ComPtr<IDWriteLocalizedStrings> dwriteString;
+        RETURN_IF_FAILED(fonts[i]->GetInformationalStrings(DWRITE_INFORMATIONAL_STRING_POSTSCRIPT_NAME, &dwriteString, &exists));
+        CFStringRef name = _CFStringFromLocalizedString(dwriteString.Get());
```
DHowett-MSFT: it is strange that `_CFStringFromLo...` doesn't contain `Get` or `Create`, so we do not know who owns its return value...

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -208,6 +208,173 @@ HRESULT _DWriteCreateFontFaceWithFontDescriptor(CTFontDescriptorRef fontDescript
...
+    for (auto& font : fonts) {
+        ComPtr<IDWriteLocalizedStrings> dwriteFontName;
+        RETURN_IF_FAILED(font->GetFaceNames(&dwriteFontName));
+        CFStringRef name = _CFStringFromLocalizedString(dwriteFontName.Get());
+        if (CFStringCompare(name, styleName, kCFCompareCaseInsensitive) == kCFCompareEqualTo) {
+            matchingFonts.emplace_back(font);
+        }
+    }
+
+    if (!matchingFonts.empty() || forceMatch) {
```
ms-jihua: nit: move these, and possibly the reserve() logic, into __FilterFontsByKeyAndValue

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -208,6 +208,173 @@ HRESULT _DWriteCreateFontFaceWithFontDescriptor(CTFontDescriptorRef fontDescript
...
+}
+
+HRESULT __FilterFontsByStyleName(std::vector<ComPtr<IDWriteFont>>& fonts, CFStringRef styleName, bool forceMatch) {
+    std::vector<ComPtr<IDWriteFont>> matchingFonts;
+    matchingFonts.reserve(fonts.size());
+    for (auto& font : fonts) {
+        ComPtr<IDWriteLocalizedStrings> dwriteFontName;
+        RETURN_IF_FAILED(font->GetFaceNames(&dwriteFontName));
+        CFStringRef name = _CFStringFromLocalizedString(dwriteFontName.Get());
+        if (CFStringCompare(name, styleName, kCFCompareCaseInsensitive) == kCFCompareEqualTo) {
```
ms-jihua: nit: I think CFStringFindWithOptions() is more efficient here since it can early return in the failure case?

ms-jihua: same with the rest of the CFStringCompares here

---
In reply to: [104538394](https://github.com/Microsoft/WinObjC/pull/2171#discussion_r104538394) [](ancestors = 104538394)

Frameworks/CoreText/DWriteWrapper_CTFont.mm
ms-jihua: also should declare the TypeRef -> CFStringRef cast here

ms-jihua:
> HRESULT _DWriteCreateFontFaceWithFontDescriptor(CTFontDescriptorRef fontDescriptor, IDWriteFontFace** fontFace) {

should probably rewrite this to make use of your new logic?

---
Refers to: Frameworks/CoreText/DWriteWrapper_CTFont.mm:170 in 3610687. [](commit_id = 36106871de405ecfcf9af0243db0ae4c166369ef, deletion_comment = False)

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -208,6 +208,173 @@ HRESULT _DWriteCreateFontFaceWithFontDescriptor(CTFontDescriptorRef fontDescript
...
+
+/**
+ * Creates a CFArray of CTFontDescriptors values for all fonts that match the given attributes best and completely match mandatoryKeys
+ */
+HRESULT _DWriteCreateMatchingFontDescriptors(CFDictionaryRef attributes, CFSetRef mandatoryKeys, CFArrayRef* matchingNames) {
+    *matchingNames = nullptr;
+
+    // Get list of all possible fonts
+    std::vector<ComPtr<IDWriteFont>> fonts;
+    RETURN_IF_FAILED(_DWriteGetAllFonts(fonts));
```
ms-jihua: euuurgh. this is a pretty huge number of fonts and i'd like to not fetch all of them in as many cases as possible.

ms-jihua: ie: in the case where name or family name is specified, the 'right way' probably involves using FindFamilyName, GetFontFamily, rather than filtering ourselves. our abstraction should reflect that.

---
In reply to: [104540071](https://github.com/Microsoft/WinObjC/pull/2171#discussion_r104540071) [](ancestors = 104540071)

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -208,6 +208,173 @@ HRESULT _DWriteCreateFontFaceWithFontDescriptor(CTFontDescriptorRef fontDescript
...
+    std::unique_ptr<CFTypeRef[]> outDescriptors(new CFTypeRef[fonts.size()]);
+    for (size_t i = 0; i < fonts.size(); ++i) {
+        BOOL exists;
+        ComPtr<IDWriteLocalizedStrings> dwriteString;
+        RETURN_IF_FAILED(fonts[i]->GetInformationalStrings(DWRITE_INFORMATIONAL_STRING_POSTSCRIPT_NAME, &dwriteString, &exists));
+        CFStringRef name = _CFStringFromLocalizedString(dwriteString.Get());
+        outDescriptors[i] = CFAutorelease(CTFontDescriptorCreateWithNameAndSize(name, kCTFontSystemFontSize));
+    }
+
+    *matchingNames = CFArrayCreate(nullptr, (const void**)outDescriptors.release(), fonts.size(), &kCFTypeArrayCallBacks);
```
ms-jihua: seems like you should just use CFMutableArray instead of being fancy?

Frameworks/CoreText/DWriteWrapper_CTFont.mm
aballway: yes but I don't believe vector has a detach(), so I went for unique_ptr.  Does unique_iw give any advantage over unique_ptr?

Frameworks/CoreGraphics/DWriteFontCollectionHelper.mm
aballway: It's sort of a "out param pointer" with a little bit less room for error, and though it's not ideal I'm not sure there's a better way to pass around a dynamic list of IDWriteFonts

Frameworks/CoreText/DWriteWrapper_CTFont.mm
aballway: whoops, nice catch!

Frameworks/CoreText/DWriteWrapper_CTFont.mm
aballway: We can't because it could be a CFDictionaryRef for Traits

Frameworks/CoreText/DWriteWrapper_CTFont.mm
aballway: I was originally doing it that way but it ends up being a pretty big web of ifs and I thought the runtime tradeoff of having to check all fonts was better than a tangled mess of code to handle all cases for names more efficiently

Frameworks/CoreText/DWriteWrapper_CTFont.mm
aballway: that would make a whole lot more sense yup

Frameworks/CoreText/DWriteWrapper_CTFont.mm
ms-jihua: I think you could probably manage to make it both nice and performant. ie off the top of my head (you can probably improve on this) - you could change the pattern so that instead of getting all fonts, and checking them here, you instead provide the check as a lambda/enum to something like _DWriteGetAllFonts(). You do some rearranging here to make sure names are processed before the other filter conditions. Then, if the new _DWriteGet___Fonts() is being called for the first time (ie you passed it a null/empty array, instead of a previously filtered one), then you can do the aforementioned fast path.

Frameworks/CoreText/DWriteWrapper_CTFont.mm
ms-jihua: ah, I misread the signature for __FilterFontsByKeyAndValue

Frameworks/CoreText/DWriteWrapper_CTFont.mm
aballway: I can strip them down until they're only loops w/ logic inside, but unfortunately the HRESULT returns aren't easily composable so I wouldn't be able to share any more without doing some crazy higher-order stuff that (though it would be pretty neat imo) almost certainly isn't worth it.

Frameworks/CoreText/DWriteWrapper_CTFont.mm
aballway: I think we'd be safe with attribute keys, but I'm not so sure about font names/families, as we might run into problems with "Arial" and "ArialJK" or some other user-defined font nonsense.  Maybe style and display names would be safe? I'm not sure, what do you think?

ms-jihua: ? you could just make sure the ranges are correct/lengths match

aballway: 😮 mm yea once again that makes a whole lot of sense.

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -225,11 +225,14 @@ static void __FilterFontsByTraits(const std::vector<ComPtr<IDWriteFont>>& fonts,
 static HRESULT __FilterFontsByStyleName(const std::vector<ComPtr<IDWriteFont>>& fonts,
                                         CFStringRef styleName,
                                         std::vector<ComPtr<IDWriteFont>>& matchingFonts) {
+    CFIndex styleLength = CFStringGetLength(styleName);
     for (auto& font : fonts) {
         ComPtr<IDWriteLocalizedStrings> dwriteFontName;
         RETURN_IF_FAILED(font->GetFaceNames(&dwriteFontName));
         CFStringRef name = _CFStringFromLocalizedString(dwriteFontName.Get());
-        if (CFStringCompare(name, styleName, kCFCompareCaseInsensitive) == kCFCompareEqualTo) {
+        CFIndex fontStyleLength = CFStringGetLength(name);
```
ms-jihua: nit: I'd just make a helper for the compare. You'd have to do more GetLengths but that shouldn't matter too much.

Frameworks/CoreText/DWriteWrapper_CTFont.mm
ms-jihua: fwiw I read it as an autorelease value. if I write it like [NSString stringFromLocalizedString:...] that looks a lot like an autorelease even though it doesn't have 'get', right?

---
In reply to: [104537218](https://github.com/Microsoft/WinObjC/pull/2171#discussion_r104537218) [](ancestors = 104537218)

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -208,6 +208,207 @@ HRESULT _DWriteCreateFontFaceWithFontDescriptor(CTFontDescriptorRef fontDescript
...
+            if (mandatoryKeys) {
+                // Not matching is not a fatal error, but we can just return now rather than try to match against any other attributes
+                return S_FALSE;
+            }
+        } else {
+            fonts.emplace_back(matchingFont);
+        }
+    }
+
+    if (fonts.empty() && CFDictionaryContainsKey(attributes, kCTFontFamilyNameAttribute) &&
```
ms-jihua: should this be an else if?

aballway: No, because font if we've gone into fontName and it failed but it's not required we still want to go into the family branch.

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -208,6 +208,207 @@ HRESULT _DWriteCreateFontFaceWithFontDescriptor(CTFontDescriptorRef fontDescript
...
+HRESULT _DWriteCreateMatchingFontDescriptors(CFDictionaryRef attributes, CFSetRef mandatoryKeys, CFArrayRef* matchingNames) {
+    RETURN_HR_IF(E_POINTER, !attributes || !matchingNames);
+    *matchingNames = nullptr;
+
+    std::vector<ComPtr<IDWriteFont>> fonts;
+    if (CFDictionaryContainsKey(attributes, kCTFontNameAttribute) &&
+        (!mandatoryKeys || CFSetContainsValue(mandatoryKeys, kCTFontNameAttribute))) {
+        // Font name is either mandatory or there are no mandatory keys, so we can start match from font
+        ComPtr<IDWriteFont> matchingFont;
+        if (FAILED(_DWriteCreateFontWithName(static_cast<CFStringRef>(CFDictionaryGetValue(attributes, kCTFontNameAttribute)),
```
ms-jihua: ugh you were right, this if nesting is really ugly...

I think I'd prefer the failure case second, since there's a case where it just no-ops, and it's more intuitive to see that if the eye doesn't have to read the else block afterward

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -208,6 +208,207 @@ HRESULT _DWriteCreateFontFaceWithFontDescriptor(CTFontDescriptorRef fontDescript
...
+                                             &matchingFont))) {
+            if (mandatoryKeys) {
```
ms-jihua: nit: possibly use a macro here

Frameworks/CoreText/DWriteWrapper_CTFont.mm
ms-jihua: write this as a comment?

---
In reply to: [105239780](https://github.com/Microsoft/WinObjC/pull/2171#discussion_r105239780) [](ancestors = 105239780)

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -40,6 +40,15 @@
                     { DWRITE_FONT_WEIGHT_ULTRA_BLACK, kCTFontWeightBlack } };
 
 /**
+ * Helper function that effeciently compares two CFStringRef and returns true if they are equal (case insensitive)
```
ms-jihua: nit: efficiently

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -331,28 +331,26 @@ HRESULT _DWriteCreateMatchingFontDescriptors(CFDictionaryRef attributes, CFSetRe
...
-        if (FAILED(_DWriteCreateFontWithName(static_cast<CFStringRef>(CFDictionaryGetValue(attributes, kCTFontNameAttribute)),
-                                             &matchingFont))) {
+        if (SUCCEEDED(_DWriteCreateFontWithName(static_cast<CFStringRef>(CFDictionaryGetValue(attributes, kCTFontNameAttribute)),
+                                                &matchingFont))) {
+            fonts.emplace_back(matchingFont);
+        } else {
             if (mandatoryKeys) {
                 // Not matching is not a fatal error, but we can just return now rather than try to match against any other attributes
                 return S_FALSE;
+            } else {
```
ms-jihua: nit: this else seems silly and probably doesn't actually help readability

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -331,28 +331,26 @@ HRESULT _DWriteCreateMatchingFontDescriptors(CFDictionaryRef attributes, CFSetRe
         (!mandatoryKeys || CFSetContainsValue(mandatoryKeys, kCTFontNameAttribute))) {
         // Font name is either mandatory or there are no mandatory keys, so we can start match from font
         ComPtr<IDWriteFont> matchingFont;
-        if (FAILED(_DWriteCreateFontWithName(static_cast<CFStringRef>(CFDictionaryGetValue(attributes, kCTFontNameAttribute)),
-                                             &matchingFont))) {
+        if (SUCCEEDED(_DWriteCreateFontWithName(static_cast<CFStringRef>(CFDictionaryGetValue(attributes, kCTFontNameAttribute)),
+                                                &matchingFont))) {
+            fonts.emplace_back(matchingFont);
+        } else {
             if (mandatoryKeys) {
```
ms-jihua: nit: could just use a macro here

Frameworks/CoreGraphics/DWriteFontCollectionHelper.mm
```diff
@@ -130,6 +130,30 @@
...
+        ComPtr<IDWriteFontFamily> fontFamily;
+        RETURN_IF_FAILED(fontCollection->GetFontFamily(i, &fontFamily));
+        ComPtr<IDWriteFontList> fontList;
+        RETURN_IF_FAILED(
+            fontFamily->GetMatchingFonts(DWRITE_FONT_WEIGHT_NORMAL, DWRITE_FONT_STRETCH_NORMAL, DWRITE_FONT_STYLE_NORMAL, &fontList));
+        const uint32_t fontCount = fontList->GetFontCount();
+        for (size_t j = 0; j < fontCount; ++j) {
+            ComPtr<IDWriteFont> font;
+            RETURN_IF_FAILED(fontList->GetFont(j, &font));
+            outFontList.emplace_back(font);
```
DHowett-MSFT: nit: `std::move` this to avoid an addref/release.

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -208,6 +217,203 @@ HRESULT _DWriteCreateFontFaceWithFontDescriptor(CTFontDescriptorRef fontDescript
...
+static HRESULT __FilterFontsByInformationalString(const std::vector<ComPtr<IDWriteFont>>& fonts,
+                                                  CFStringRef name,
+                                                  DWRITE_INFORMATIONAL_STRING_ID id,
+                                                  std::vector<ComPtr<IDWriteFont>>& matchingFonts) {
+    for (auto& font : fonts) {
+        BOOL exists;
+        ComPtr<IDWriteLocalizedStrings> dwriteString;
+        RETURN_IF_FAILED(font->GetInformationalStrings(id, &dwriteString, &exists));
+        if (exists) {
+            CFStringRef fontName = _CFStringFromLocalizedString(dwriteString.Get());
```
DHowett-MSFT: nit: this ain't a font name

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -208,6 +217,203 @@ HRESULT _DWriteCreateFontFaceWithFontDescriptor(CTFontDescriptorRef fontDescript
...
+                                                            static_cast<CFStringRef>(value),
+                                                            DWRITE_INFORMATIONAL_STRING_FULL_NAME,
+                                                            matchingFonts));
+    } else if (__CFStringEqual(key, kCTFontStyleNameAttribute)) {
+        RETURN_IF_FAILED(__FilterFontsByStyleName(fonts, static_cast<CFStringRef>(value), matchingFonts));
+    } else if (__CFStringEqual(key, kCTFontTraitsAttribute)) {
+        __FilterFontsByTraits(fonts, static_cast<CFDictionaryRef>(value), matchingFonts);
+    } else {
+        // Other attributes are currently unsupported by CTFont
+        UNIMPLEMENTED();
```
DHowett-MSFT: `UNIMPLEMENTED_WITH_MSG` may be more helpful, since these keys are strings and their values can be logged.

PR #2180, opened 2017-03-08
Add GitVersion.yml actual binary will come later when it is decided w…

bbowman:
…hich version to use and how to consume it.


GitVersion.yml
```diff
@@ -0,0 +1,28 @@
...
+    is-release-branch: true
+    is-mainline: true
+  develop:
+    regex: develop
+    increment: None
+    tag: 'dev'
+    track-merge-target: true
+    tracks-release-branches: true
+    is-release-branch: false
+  .*:
```
memontic-ms: Is increment an optional config? Briefly looking at the docs it seems this should be "increment: Inherit".

memontic-ms:
:shipit:

GitVersion.yml
bbowman: I can try that out.

bbowman: or actually none. I think.

PR #2181, opened 2017-03-08
Added projections for Microsoft Ads (Microsoft Store Serices SDK 10.0.1)

mukhole:


DHowett-MSFT:
Why did this require all the existing projected libraries to be rebuilt?

mukhole:
This is phase 1 of the implementation, in this we generate a single solution for all meta data files using our tool.
Phase 2 will have separate solutions and then we will have only MS Ads binaries.

---
In reply to: [285151637](https://github.com/Microsoft/WinObjC/pull/2181#issuecomment-285151637) [](ancestors = 285151637)

PR #2182, opened 2017-03-08
Bbowman/packaging merge in latest develop

bbowman:


memontic-ms:
:shipit: As long as the build passes

PR #2190, opened 2017-03-09
Objc2winmd binary update: Fix issues in projecting multipe types from async blocks

mukhole:


pradipd:
:shipit:

ms-jihua:
Seeing an issue when building FunctionalTests after pulling this:

LNK1104	cannot open file 'ObjCUWP_Windows_Media_Core_Protection_Playback_Capture_Devices.lib'

The lib checked in is named 'ObjCUWP_Windows_Media_Playback_Protection_Core_Capture_Devices.lib' instead of the above ordering of namespaces - was there a misnaming somewhere?

DHowett-MSFT:
@ms-jihua You may have an old `.obj` laying around; clean the offending project and it should work :smile:

ms-jihua:
@DHowett-MSFT Yup, that fixed it. Sorry for the trouble!

PR #2194, opened 2017-03-09
UIViewController modal present/dismiss test

ehren:
A version of this test hacked into a test app passes on the ref platform. 

No prob to let this bake - not sure if this is the right approach for validating the correct order of view events (an alternative approach would just be to build a big list of 'events' and validate them at the end - though validating the state of the presented and presenting view controller during the viewDid/Will events is easier when sticking the EXPECTs in the test view controller itself). 

Also, when checking WinObjC behavior vs the ref platform, I've just added verbose logging into this testcase; perhaps doing this with a new LOG_DEBUG macro would be nice.

Issue #2115, opened 2017-02-28
Adding a controller owned view directly to a view hierarchy shouldn't automatically invoke viewDidAppear


ehren:
Code like this:

    [window addSubview:controller.view]; // view owned by controller added directly to a view hierarchy

should result in viewWillAppear: sent to the controller. viewDidAppear: should be the responsibility of the user to manually call. However, current handling in UIView.mm calls viewDidAppear: as well.

Note: This is only relevant for very old code (documentation referencing the above behaviour has been deleted - despite the ref platform conforming to it). It should be fixed (without breaking appearance event handling for the root view controller and UINavigationController) - but only after view appearance event handling in UIViewController is sorted.

Note: Filling this only for reference as this issue complicates the event handling logic in UIViewController.mm

Issue #2114, opened 2017-02-28
UIViewController viewDidAppear/viewDidDisappear called too soon


ehren:
During a modal dismiss, the current implementation tears down the presented-presenting relationship just before calling the user completion block (but after viewDidDisappear: is called on the modal). This doesn't match the ref platform where e.g. presentingViewController should be nil on viewDidDisappear: of the modal.

(cause of a crash in our app due to some tricky conditional logic checking if a view controller is presenting a modal during a view event).

edvv:
The beingDismissed and beingPresented properties are just stubs, so it is difficult to be picky about this issue. There are a lot of code sections that need thorough thinking out, and this is one of them. This is also an issue on iOS itself, and there is a lot of implementation-dependent code that attempts to navigate the view controller calling sequences.

tadam-msft:
Fixed with #2195

PR #2195, opened 2017-03-09
UIViewController modal appearance event fixes and addition of UIModalPresentationFormSheet overlay

### LINKED ISSUE #2113: UIViewController UIModalPresentationFormSheet needs an overlay
Parent window of modal needs to by covered by a view when a modal is presented. Currently it's possible to interact with the parent while it's presenting a modal - divergence from ref platform (note: patch in progress).

ehren:
Fixes #2113 and #2114. In addition, fixes to bring the invocation order of the view[Did/Will][Appear/Disappear] events in line with the ref platform have had to take #2115 into account (commit 20a4224 from March 19 has a better approach for working around #2115 than the initial commits to this pull).

Notes:

I went with the block based animation to avoid need for a CATransaction (form sheet overlay and the modal view itself have separate layers).

I've also been using a modified version of the testcase in #2194 to validate the order of the [will/did]MoveTo[Superview/Window] events. WinObjC differs from the ref platform (which for example, seems to move the view of a controller completely covered by a modal to an offscreen dummy window first and then removes it from that dummy window entirely, while WinObjC just leaves them in the view hierarchy). However, these changes do a better job of the ordering of the window/superview events vs the view controller appearance events (for example the window/superview events should be finished by the time of viewDidAppear); though, this is not something validated by the testcase in #2194.

### DIFF
--- a/Frameworks/UIKit/UIPopoverPresentationController.mm
+++ b/Frameworks/UIKit/UIPopoverPresentationController.mm
@@ -23,12 +23,6 @@
 #import "UIPopoverPresentationControllerInternal.h"
 #import "UIPopoverControllerInternal.h"
 
-@interface UIPopoverPresentationController ()
-
-- (void)_handleDismiss;
-
-@end
-
 @interface _UIPopoverControllerDelegateInternal : NSObject <UIPopoverControllerDelegate>
 
 @property (nonatomic, assign) UIPopoverPresentationController* controller;
@@ -66,8 +60,6 @@
     if ([_delegate respondsToSelector:@selector(popoverPresentationControllerDidDismissPopover:)]) {
         [_delegate popoverPresentationControllerDidDismissPopover:_controller];
     }
-
-    [_controller _handleDismiss];
 }
 
 @end
@@ -79,7 +71,6 @@
     StrongId<UIBarButtonItem*> _barButtonItem;
     StrongId<UIView*> _sourceView;
     CGRect _sourceRect;
-    StrongId<dispatch_block_t> _dismissCompletion;
 }
 
 /**
@@ -97,9 +88,8 @@
     return self;
 }
 
-- (void)_presentAnimated:(BOOL)animated presentCompletion:(dispatch_block_t)presentCompletion dismissCompletion:(dispatch_block_t)dismissCompletion {
+- (void)_presentAnimated:(BOOL)animated presentCompletion:(dispatch_block_t)presentCompletion {
     _popoverControllerInternal->_presentCompletion.attach([presentCompletion copy]);
-    _dismissCompletion.attach([dismissCompletion copy]);
 
     if ([[_delegateInternal delegate] respondsToSelector:@selector(prepareForPopoverPresentation:)]) {
         [[_delegateInternal delegate] prepareForPopoverPresentation:self];
@@ -119,16 +109,6 @@
 
 - (void)_dismissAnimated:(BOOL)animated completion:(dispatch_block_t)dismissCompletion {
     [_popoverControllerInternal _dismissPopoverAnimated:animated completion:dismissCompletion];
-}
-
-- (void)_handleDismiss {
-    // Stay alive while we invoke _dismissCompletion (UIViewController tears
-    // down its strong ref to us in this completion; we must avoid our
-    // deallocation while this block we own is executing).
-    StrongId<UIPopoverPresentationController> strongSelf = self;
-
-    _dismissCompletion();
-    _dismissCompletion = nil;
 }
 
 /**
--- a/Frameworks/UIKit/UIViewController.mm
+++ b/Frameworks/UIKit/UIViewController.mm
@@ -1114,14 +1114,6 @@ NSMutableDictionary* _pageMappings;
     [self setToolbarItems:newItems animated:FALSE];
 }
 
-- (BOOL)_managesViewEvents {
-    return priv->_managesViewEvents;
-}
-
-- (void)_setManagesViewEvents:(BOOL)managesViewEvents {
-    priv->_managesViewEvents = managesViewEvents;
-}
-
 - (void)_setBeingPresented:(BOOL)beingPresented {
     _beingPresented = beingPresented;
 }
@@ -1271,14 +1263,8 @@ NSMutableDictionary* _pageMappings;
     [presented->priv->_modalOverlayView removeFromSuperview];
     presented->priv->_modalOverlayView = nil;
 
-    // Prevent default view appearance handling in UIView occuring when we
-    // remove the presented controller's view from its window.
-    presented->priv->_managesViewEvents = YES;
-
     UIView* presentedView = [presented view];
     [presentedView removeFromSuperview];
-
-    presented->priv->_managesViewEvents = NO;
 
     UIView* view = [self view];
     [[view superview] bringSubviewToFront:view];
@@ -1477,16 +1463,8 @@ NSMutableDictionary* _pageMappings;
         [parentWindow addSubview:priv->_modalOverlayView];
     }
 
-    // Avoid default view appearance mechanisms in UIView.mm (which e.g. handle view
-    // appearence events for views with an owning controller manually added to a window).
-    // Note that mechanism in UIView.mm currently overeagerly calls viewDidAppear when adding
-    // a controller owned view to a window.
-    self->priv->_managesViewEvents = YES;
-
     UIView* view = [self view];
     [parentWindow addSubview:view];
-
-    self->priv->_managesViewEvents = NO;
 
     if (animated) {
         CGPoint origPos = [view center];
@@ -2128,8 +2106,10 @@ static UIInterfaceOrientation findOrientation(UIViewController* self) {
  @Status Caveat
  @Notes  WinObjC-only extension providing access
          to UIModalPresentationFormSheet overlay.
+         Not exposed in header (clients must access
+         via performSelector).
 */
-- (UIView*)modalOverlayView {
+- (UIView*)_modalOverlayView {
     return priv->_modalOverlayView;
 }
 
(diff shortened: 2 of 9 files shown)


Frameworks/UIKit/UIViewController.mm
```diff
@@ -1207,43 +1196,40 @@ - (void)presentViewController:(UIViewController*)controller animated:(BOOL)anima
...
     [controller view];
+    priv->_presentedViewController = controller;
     controller->priv->_parentViewController = self;
     controller->priv->_presentingViewController = self;
 
     if ([controller modalPresentationStyle] == UIModalPresentationPopover) {
         controller->priv->_popoverPresentationController.attach(
             [[UIPopoverPresentationController alloc] initWithPresentedViewController:controller presentingViewController:self]);
 
+        // Keep popoverPresentationController alive until presentation complete. Fixes a corner case crash where the
```
ehren: crash exposed by testcase in #2194

Frameworks/UIKit/UIPopoverPresentationController.mm
```diff
@@ -79,7 +71,6 @@ @implementation UIPopoverPresentationController {
     StrongId<UIBarButtonItem*> _barButtonItem;
     StrongId<UIView*> _sourceView;
     CGRect _sourceRect;
-    StrongId<dispatch_block_t> _dismissCompletion;
```
ehren: No need for a private dismissCompletion API anymore (for the one user UIViewController.mm). This was used for the unlinking of the modal from the parent which is now handled in WYPopoverController (doing the unlinking after didDismissPopover is way too late).

ehren:
The addition of the extension to access the modalOverlayView is to make hacks like http://stackoverflow.com/questions/9102497/dismiss-modal-view-form-sheet-controller-on-outside-tap much easier than on the ref platform.

Frameworks/UIKit/UIViewController.mm
```diff
@@ -1207,43 +1196,40 @@ - (void)presentViewController:(UIViewController*)controller animated:(BOOL)anima
 
     [[UIApplication sharedApplication] beginIgnoringInteractionEvents];
 
-    if ([controller _hidesParent]) {
-        [self _notifyViewWillDisappear:animated];
-    }
-
-    priv->_presentedViewController = controller;
     [controller view];
```
ehren: reason for setting the presentedViewController after [controller view] (which implicitly can call viewDidLoad) is to match ref platform behavior; but this is not something validated by the #2194 testcase.

Frameworks/UIKit/UIViewController.mm
```diff
@@ -1254,40 +1240,64 @@ - (void)dismissModalViewControllerAnimated:(BOOL)animated {
...
     if (controller != nil) {
-        [controller _notifyViewDidDisappear:TRUE];
+        [controller _notifyViewDidDisappear:animated];
     }
 }
 
-- (void)_dismissTransitionStopped:(id)anim finished:(BOOL)finished {
-    UIView* view = [priv->_dismissController view];
-    [[view layer] removeAnimationForKey:@"ModalDismiss"];
+- (void)_unlinkPresentedController {
```
ehren: rename of method made the diff worse but |_unlinkPresentedController| is the same method as the previous |_childDismissCleanup|.

Frameworks/UIKit/UIViewController.mm
```diff
@@ -1380,85 +1380,66 @@ - (void)dismissViewControllerAnimated:(BOOL)animated completion:(void (^)(void))
...
-        [presented _notifyViewDidDisappear:NO];
-        [self _notifyViewDidAppear:animated];
-
-        cleanupCompletion();
+        [self _finishModalDismissAnimated:NO completion:completion];
     }
 }
 
-- (void)_notifyDidAppearAnimated:(UIView*)view {
+- (void)_notifyDidAppear:(UIView*)view animated:(BOOL)animated {
```
ehren: This is the scariest change in that there's no test coverage (previously only the animated code path used this method to recursively call viewDidAppear on any view/subview with an owning controller (likely the existing code relied on the default handling in UIView.mm to handle this in the non-animated case). I wanted to make sure there was no incorrect behavior change after disabling the default appearance handling in UIView.mm so this method is now invoked in the non-animated case as well. Note that the underlying scenario (the need to invoke appearance events for a controller whose view is a subview of a controller being presented via presentViewController) seems obscure. Same comments apply for the similar method that handles viewDidDisappear.

include/UIKit/UIViewController.h
```diff
@@ -117,6 +117,7 @@ UIKIT_EXPORT_CLASS
 @property (readonly, nonatomic, strong) NSExtensionContext* extensionContext STUB_PROPERTY;
 @property (readonly, nonatomic, strong) UISearchDisplayController* searchDisplayController;
 @property (readonly, nonatomic, strong) UIView* viewIfLoaded STUB_PROPERTY;
+@property (readonly, nonatomic, strong) UIView* modalOverlayView; // WinObjC extension
```
ehren: If there's no objection, I'm going to push a commit in the coming days to remove this added WinObjC extension from the header. I'll just leave an _ prefixed method in the .mm so that clients can access the modalOverlayView via performSelector.

Frameworks/UIKit/WYPopoverController.mm
```diff
@@ -2634,8 +2639,12 @@ - (void)dismissPopoverAnimated:(BOOL)aAnimated
         }
 #endif
     };
+
+    UIPopoverPresentationController* popoverPresentationController = [viewController popoverPresentationController];
     
     void (^completionBlock)() = ^() {
+        // Keep popoverPresentationController alive via block-capture during teardown.
+        (void)popoverPresentationController;
```
jaredhms: > (void)popoverPresentationController; [](start = 7, length = 37)

nit: might be a bit more clear to do an assignment to a local variable 'strongPopoverPresentationController', rather than this.

Frameworks/UIKit/UIViewController.mm
jaredhms: can you add a comment here so we know not to reorder it in the future?

---
In reply to: [106151392](https://github.com/Microsoft/WinObjC/pull/2195#discussion_r106151392) [](ancestors = 106151392)

Frameworks/UIKit/UIViewController.mm
```diff
@@ -1207,43 +1188,40 @@ - (void)presentViewController:(UIViewController*)controller animated:(BOOL)anima
...
 
     if ([controller modalPresentationStyle] == UIModalPresentationPopover) {
         controller->priv->_popoverPresentationController.attach(
             [[UIPopoverPresentationController alloc] initWithPresentedViewController:controller presentingViewController:self]);
 
+        // Keep popoverPresentationController alive until presentation complete. Fixes a corner case crash where the
+        // popover is dismissed, without animation, in the presentViewController completion block itself.
+        UIPopoverPresentationController* popover = [controller->priv->_popoverPresentationController retain];
+
         dispatch_block_t popoverPresent = ^{
```
jaredhms: >popoverPresent [](start = 25, length = 14)

nit: should this be called 'presentCompletion' for clarity?

Frameworks/UIKit/UIViewController.mm
```diff
@@ -2148,6 +2103,17 @@ - (UIModalPresentationStyle)modalPresentationStyle {
...
 /**
+ @Status Caveat
+ @Notes  WinObjC-only extension providing access
+         to UIModalPresentationFormSheet overlay.
+         Not exposed in header (clients must access
+         via performSelector).
+*/
+- (UIView*)_modalOverlayView {
+    return priv->_modalOverlayView;
+}
```
jaredhms: I don't see this used anywhere; is it needed?  Also, we shouldn't be annotating internal/private methods.

ehren: That's needed to allow users to interact with the overlay view, typically to dismiss the modal on overlay-touch (by attaching a gesture recognizer). I didn't want to expose this in the public header, instead forcing the user to access via performSelector (since it requires private API on iOS, but it's often done). What about a commit adding dismiss via touch on overlay to the modal example in WOCCatalog? (will also remove the annotations from the method comment while I'm at it)

jaredhms: sounds good - i'm about to pull your changes and try them out locally

---
In reply to: [109035577](https://github.com/Microsoft/WinObjC/pull/2195#discussion_r109035577) [](ancestors = 109035577)

Frameworks/UIKit/WYPopoverController.mm
ehren: There's an implicitly strong local defined just above (ARC's enabled) but I need to keep its lifetime tied to that completionBlock via a capture. Unfortunately, I can't say do ```assert(popoverPresentationController != nil);``` in the body of the block since it can be nil in the current implementation with the deprecated UIPopoverController API. I'm inclined to leave as is but the best alternative would probably be to use ```popoverPresentationController``` in a logging statement in the block body.

Frameworks/UIKit/UIViewController.mm
ehren: fixed in latest.

Frameworks/UIKit/WYPopoverController.mm
ehren: or an explicit ivar is doable too

jaredhms: I was suggesting something like: UIPopoverPresentationController* blockCapture = popoverPresentationController;

---
In reply to: [109064009](https://github.com/Microsoft/WinObjC/pull/2195#discussion_r109064009) [](ancestors = 109064009)

jaredhms: although this is probably fine - i just kicked off a CI build

---
In reply to: [109265389](https://github.com/Microsoft/WinObjC/pull/2195#discussion_r109265389) [](ancestors = 109265389,109064009)

PR #2203, opened 2017-03-09
Create version swapping task and change WOCCatalog to use it as an ex…

bbowman:
…ample.


common/winobjc.packagereference.override.targets
```diff
@@ -0,0 +1,83 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup>
+    <WinObjCPackageOverridePath Include="$(WINOBJC_SDK_ROOT)\build\OutputPackages"/>
+    <WinObjCPackageOverridePath Include="$(WINOBJC_SDK_ROOT)\tools\OutputPackages"/>
```
memontic-ms: >(WINOBJC_SDK_ROOT [](start = 42, length = 17)

:'( I'd rather not need to have this set

memontic-ms:
Still looks good.

PR #2204, opened 2017-03-09
Support NSArray KVO for elements

### LINKED ISSUE #2068: Finish NSArray KVO methods
These should throw exceptions per documentation.

addObserver:forKeyPath:options:context:
removeObserver:forKeyPath:
removeObserver:forKeyPath:context:

These should be implemented:

addObserver:toObjectsAtIndexes:forKeyPath:options:context:
removeObserver:fromObjectsAtIndexes:forKeyPath:context:
removeObserver:fromObjectsAtIndexes:forKeyPath:

aballway:
Causes an NSException to be thrown if addObserver:forKeyPath* or removeObserver:forKeyPath* are called on NSArray as per documentation.

Implements addObserver:toObjectsAndIndexes* and removeObserver:fromObjectsAtIndexes* for NSArray by calling addObserver and remvoeObserver on each specified element. Though documentation states that this method _is not_ a simple convenience wrapper, we cannot do so because our current implementation of KVO cannot support a single observer intermediary for multiple observees without a larger refactor than this issue calls for.

Fixes #2068


Frameworks/include/NSEnumeratorInternal.h
```diff
@@ -18,6 +18,11 @@
 
 #include <Foundation/NSEnumerator.h>
 
+#define NS_COLLECTION_THROW_ILLEGAL_KVO(KEYPATH)                                                                                         \
+    do {                                                                                                                                 \
+        [NSException raise:NSInvalidArgumentException format:@"%@ is not supported. Key path: %@", NSStringFromSelector(_cmd), KEYPATH]; \
```
DHowett-MSFT: nit: no shouty `KEYPATH`.
Please add `-[%hs %hs]` with `object_getClassName(self), sel_getName(_cmd)`; this will add the class (obvs.) and avoid the NSString intermediary for `_cmd`.

Frameworks/include/NSEnumeratorInternal.h
```diff
@@ -18,6 +18,11 @@
 
 #include <Foundation/NSEnumerator.h>
 
+#define NS_COLLECTION_THROW_ILLEGAL_KVO(KEYPATH)                                                                                         \
```
DHowett-MSFT: enumerator is an unusual header into which to put this

aballway: I was looking for a shared internal place for the NS* classes where KVO is illegal.  What's the reasonable place to stick this?

DHowett-MSFT: Reckon we could go with `NSKeyValueObserving-Internal.h`, but iunno.

aballway:
OK so the tests passed, but the PR still reports the Packaging tests failed so merging is locked 😦

aballway:
ping!

aballway:
ping!

PR #2207, opened 2017-03-10
remove more WIN_OBJC_SDK_ROOT and add a mutex around GetVersion

bbowman:


common/winobjc.nuproj.common.targets
```diff
@@ -121,4 +121,69 @@
     </PropertyGroup>
   </Target>
 
+
+    <!-- Inline tasks (and targets) to make sure GitVersion only runs one at a time. This is an improvement that should
```
memontic-ms: >     [](start = 0, length = 4)

nit: extra space here and at end of file before </Project>
We really need a pre-commit xml formatting step.

memontic-ms:
:shipit:

PR #2208, opened 2017-03-10
Implementation of CGContextFlush. #2046

msft-Jeyaram:
The implementation is based on D2D:EndDraw vs D2D:Flush.

This is mainly due to the fact that D2D goes through D3D, the flush would flush the D2D batch command and will not flush D3D batch for
WicBitmapRenderTargert. Also there is an issue of a potential shadow copy used by D2D, and flush does not copy over the shadow copy into the original buffer.
EndDraw supports both of the issues addressed above.

### DIFF
--- a/Frameworks/CoreGraphics/CGContext.mm
+++ b/Frameworks/CoreGraphics/CGContext.mm
@@ -373,6 +373,20 @@ public:
         return S_OK;
     }
 
+    inline HRESULT Flush() {
+        // We cannot use use ID2D1::Flush here, mainly due to:
+        // 1. Flush flushes the D2D batch, but does not flush the D3D batch, we do not have a way to flush
+        // the D3D batch for WicBitmapRenderTarget. (Not: During the EndDraw procedure this occurs.)
+        // 2. Sometimes a shadow copy of our buffer can be used to do the actual rendering, Flush does not force the copy
+        // over of the shadow copy into the original buffer (EndDraw does)
+        // Thus the best way is to implement Flush is through EndDraw
+        if (_beginEndDrawDepth > 0) {
+            RETURN_IF_FAILED(PopEndDraw(true));
+            PushBeginDraw(true);
+        }
+        return S_OK;
+    }
+
     inline void SetShouldAntialias(__CGTrinary shouldAntialias) {
         CurrentGState().shouldAntialias = shouldAntialias;
     }
@@ -490,17 +504,17 @@ public:
         return SUCCEEDED(_firstErrorHr) && CurrentGState().ShouldDraw();
     }
 
-    inline void PushBeginDraw() {
-        if ((_beginEndDrawDepth)++ == 0) {
+    inline void PushBeginDraw(bool forceBeginDraw = false) {
+        if (forceBeginDraw || ((_beginEndDrawDepth)++ == 0)) {
             if (SUCCEEDED(_firstErrorHr)) {
                 deviceContext->BeginDraw();
             }
         }
     }
 
-    inline HRESULT PopEndDraw() {
+    inline HRESULT PopEndDraw(bool forceEndDraw = false) {
         HRESULT hr = S_OK;
-        if (--(_beginEndDrawDepth) == 0) {
+        if (forceEndDraw || (--(_beginEndDrawDepth) == 0)) {
             hr = deviceContext->EndDraw();
             if (_useEnhancedErrorHandling && SUCCEEDED(_firstErrorHr) && FAILED(hr)) {
                 // If we haven't yet stored an error, and we're about to return an error (not S_OK), store it.
@@ -612,11 +626,12 @@ void CGContextRestoreGState(CGContextRef context) {
 
 #pragma region Global State - Context Maintenance
 /**
- @Status Stub
+ @Status Interoperable
+ @Notes The flush can operate on a CGBitmapContext.
 */
 void CGContextFlush(CGContextRef context) {
     NOISY_RETURN_IF_NULL(context);
-    UNIMPLEMENTED();
+    FAIL_FAST_IF_FAILED(context->Flush());
 }
 
 /**
@@ -2522,7 +2537,7 @@ public:
             deviceContext->Clear({ 0, 0, 0, 0 }); // Clear the original target to transparent black.
             RETURN_IF_FAILED(inputBitmap.As(&copiedImage));
         }
-        return __super::Stage(context, deviceContext);
+        return __super ::Stage(context, deviceContext);
     }
 };
 
@@ -2570,13 +2585,13 @@ public:
         RETURN_IF_FAILED(deviceContext->CreateEffect(CLSID_D2D1Composite, &_compositeEffect));
         _compositeEffect->SetInputEffect(0, affineTransformEffect.Get());
 
-        return __super::Stage(context, deviceContext);
+        return __super ::Stage(context, deviceContext);
     }
 
     HRESULT Complete(CGContextRef context, ID2D1DeviceContext* deviceContext) override {
-        RETURN_IF_FAILED(__super::Complete(context, deviceContext));
-        _shadowEffect->SetInput(0, __super::commandList.Get());
-        _compositeEffect->SetInput(1, __super::commandList.Get());
+        RETURN_IF_FAILED(__super ::Complete(context, deviceContext));
+        _shadowEffect->SetInput(0, __super ::commandList.Get());
+        _compositeEffect->SetInput(1, __super ::commandList.Get());
         deviceContext->DrawImage(_compositeEffect.Get());
         return S_OK;
     }
@@ -2597,7 +2612,7 @@ public:
     // into the member "copiedImage".
 
     HRESULT Complete(CGContextRef context, ID2D1DeviceContext* deviceContext) override {
-        RETURN_IF_FAILED(__super::Complete(context, deviceContext));
+        RETURN_IF_FAILED(__super ::Complete(context, deviceContext));
 
         // The blend effect takes two images and a blend mode.
         //
@@ -2609,8 +2624,8 @@ public:
         ComPtr<ID2D1Effect> blendEffect;
         RETURN_IF_FAILED(deviceContext->CreateEffect(CLSID_D2D1Blend, &blendEffect));
         blendEffect->SetValue(D2D1_BLEND_PROP_MODE, _blendMode);
-        blendEffect->SetInput(0, __super::copiedImage.Get());
-        blendEffect->SetInput(1, __super::commandList.Get());
+        blendEffect->SetInput(0, __super ::copiedImage.Get());
+        blendEffect->SetInput(1, __super ::commandList.Get());
 
         deviceContext->SetPrimitiveBlend(D2D1_PRIMITIVE_BLEND_COPY);
         deviceContext->DrawImage(blendEffect.Get());
@@ -2638,12 +2653,12 @@ public:
                                                         nullptr,
                                                         D2D1_LAYER_OPTIONS1_INITIALIZE_FROM_BACKGROUND),
                                  nullptr);
-        return __super::Stage(context, deviceContext);
+        return __super ::Stage(context, deviceContext);
     }
 
     HRESULT Complete(CGContextRef context, ID2D1DeviceContext* deviceContext) override {
-        RETURN_IF_FAILED(__super::Complete(context, deviceContext));
-        deviceContext->DrawImage(__super::commandList.Get(), D2D1_INTERPOLATION_MODE_NEAREST_NEIGHBOR, _compositeMode);
+        RETURN_IF_FAILED(__super ::Complete(context, deviceContext));
+        deviceContext->DrawImage(__super ::commandList.Get(), D2D1_INTERPOLATION_MODE_NEAREST_NEIGHBOR, _compositeMode);
         deviceContext->PopLayer();
         deviceContext->SetPrimitiveBlend(D2D1_PRIMITIVE_BLEND_SOURCE_OVER);
         return S_OK;
@@ -2660,8 +2675,8 @@ public:
     }
 
     HRESULT Complete(CGContextRef context, ID2D1DeviceContext* deviceContext) override {
-        RETURN_IF_FAILED(__super::Complete(context, deviceContext));
-        deviceContext->DrawImage(__super::commandList.Get(), D2D1_INTERPOLATION_MODE_NEAREST_NEIGHBOR, _compositeMode);
+        RETURN_IF_FAILED(__super ::Complete(context, deviceContext));
+        deviceContext->DrawImage(__super ::commandList.Get(), D2D1_INTERPOLATION_MODE_NEAREST_NEIGHBOR, _compositeMode);
         return S_OK;
     }
 };
--- a/tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_FlushTests.cpp
+++ b/tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_FlushTests.cpp
@@ -1,0 +1,132 @@
+//******************************************************************************
+//
+// Copyright (c) Microsoft. All rights reserved.
+//
+// This code is licensed under the MIT License (MIT).
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#include "DrawingTest.h"
+
+#if WINOBJC
+#include "CGContextInternal.h"
+
+DRAW_TEST_F(CGContextFlush, FillFlush, WhiteBackgroundTest<>) {
+    CGContextRef context = GetDrawingContext();
+    CGRect bounds = GetDrawingBounds();
+    _CGContextPushBeginDraw(context);
+
+    CGContextSetRGBFillColor(context, 1, 0, 0, 1);
+    CGContextFillRect(context, bounds);
+
+    // Flush the red fill rect
+    CGContextFlush(context);
+
+    CGContextSetRGBFillColor(context, 0, 0, 1, 1);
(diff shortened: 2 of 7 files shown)


tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_FlushTests.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+    BYTE* dataPtr = static_cast<BYTE*>(CGBitmapContextGetData(context));
+    ASSERT_NE(dataPtr, nullptr);
+
+    EXPECT_EQ(dataPtr[0], 00);
+    EXPECT_EQ(dataPtr[1], 00);
+    EXPECT_EQ(dataPtr[2], 0xff);
+    EXPECT_EQ(dataPtr[3], 0xff);
+
+    _CGContextPopEndDraw(context);
```
msft-Jeyaram: this can be a unit test, but I wanted to validate the drawings did render properly.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -362,6 +362,20 @@ inline HRESULT GetTextRenderingParams(IDWriteRenderingParams* originalParams, ID
         return S_OK;
     }
 
+    inline HRESULT Flush() {
+        // We cannot use use ID2D1::Flush here, mainly due to:
+        // 1. Flush flushes the D2D batch, but does not flush the D3D batch, we do not have a way to flush
+        // the D3D batch for WicBitmapRenderTarget. (Not: During the EndDraw procedure this occurs.)
+        // 2. Sometimes a show copy of our buffer can be used to do the actual rendering, Flush does not force the copy
```
DHowett-MSFT: nit: show?

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -362,6 +362,20 @@ inline HRESULT GetTextRenderingParams(IDWriteRenderingParams* originalParams, ID
...
+        // over of the shadow copy into the original buffer (EndDraw does)
+        // Thus the best way is to implement Flush is through EndDraw
+        if (_beginEndDrawDepth > 0) {
+            RETURN_IF_FAILED(deviceContext->EndDraw());
```
DHowett-MSFT: This will discard the error value. We have to go through PopEndDraw to collect all of the error values from all of the draw operations.

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_FlushTests.cpp
msft-Jeyaram: adding some functionality after the flush

DHowett-MSFT: offline conversation about UT:
this should be a UT, but it can be a drawing test if we look at what happens if we flush+draw+flush+draw.
not 100% okay with testing something out of our control (d2d batching params)

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -3002,7 +3017,8 @@ void CGContextIwEnableEnhancedErrorHandling(CGContextRef context) {
 
         CFTypeRef key = CFSTR("hresult"); // This matches the hresult exception key used in Foundation, which we can't import.
         CFTypeRef value = embeddedHresult.get();
-        auto userInfo = woc::MakeStrongCF<CFDictionaryRef>(CFDictionaryCreate(nullptr, &key, &value, 1, &kCFCopyStringDictionaryKeyCallBacks, &kCFTypeDictionaryValueCallBacks));
+        auto userInfo = woc::MakeStrongCF<CFDictionaryRef>(
+            CFDictionaryCreate(nullptr, &key, &value, 1, &kCFCopyStringDictionaryKeyCallBacks, &kCFTypeDictionaryValueCallBacks));
```
MSFTFox: Formatter?

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_FlushTests.cpp
msft-Jeyaram: As earlier common indicated, it should be, but with more drawings it can be a valid drawing test.

:D

---
In reply to: [105513897](https://github.com/Microsoft/WinObjC/pull/2208#discussion_r105513897) [](ancestors = 105513897)

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_FlushTests.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+    CGContextSetRGBFillColor(context, 1, 0, 0, 1);
+    CGContextFillRect(context, bounds);
+
+    CGContextFlush(context);
+
+    BYTE* dataPtr = static_cast<BYTE*>(CGBitmapContextGetData(context));
+    ASSERT_NE(dataPtr, nullptr);
+
+    EXPECT_EQ(dataPtr[0], 00);
```
MSFTFox: A note on which pixels (I presume that's what this is?) these are checking would be nice. The draw test itself is just checking that we can successfully draw a rectangle and the real test is here in the EXPECTs. The same can certainly be accomplished in a normal test with the data point checking before and after a flush.

msft-Jeyaram: please see the comments below to address why it's a drawing test.
Also I want to ensure that he final EndDraw via DrawingTests.cpp works as expected. 

---
In reply to: [105514587](https://github.com/Microsoft/WinObjC/pull/2208#discussion_r105514587) [](ancestors = 105514587)

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -493,17 +493,17 @@ inline bool ShouldDraw() {
         return SUCCEEDED(_firstErrorHr) && CurrentGState().ShouldDraw();
     }
 
-    inline void PushBeginDraw() {
-        if ((_beginEndDrawDepth)++ == 0) {
+    inline void PushBeginDraw(bool forceBeginDraw = false) {
+        if ((_beginEndDrawDepth)++ == 0 || forceBeginDraw) {
```
DHowett-MSFT: If we're going to go with the "override" flag, putting it _first_ will save unnecessary manipulation of the counter.

msft-Jeyaram:
Validated pass on ARM

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_FlushTests.cpp
```diff
@@ -0,0 +1,76 @@
...
+    CGContextFillRect(context, bounds);
+
+    // Flush the red fill rect
+    CGContextFlush(context);
+
+    CGContextSetRGBFillColor(context, 0, 0, 1, 1);
+    CGContextFillRect(context, CGRectMake(0, 0, 300, 300));
+
+    // We should still have red & blue rectangle should not show up.
+    BYTE* dataPtr = static_cast<BYTE*>(CGBitmapContextGetData(context));
```
aballway: nit: no shouty

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_FlushTests.cpp
```diff
@@ -0,0 +1,76 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#include "DrawingTest.h"
+
+#if WINOBJC
+#include "CGContextInternal.h"
+DRAW_TEST_F(CGContextFlush, FillFlush, WhiteBackgroundTest<>) {
```
aballway: This could just be a windows-only unit test rather than a drawing test since we're checking the data directly.

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_FlushTests.cpp
```diff
@@ -0,0 +1,76 @@
...
+    CGContextRef context = GetDrawingContext();
+    CGRect bounds = GetDrawingBounds();
+    _CGContextPushBeginDraw(context);
```
aballway: Should add test with multiple _CGContextPushBeginDraw calls (should force through them all) and multiple Flush calls in a row (only first one should modify data)

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_FlushTests.cpp
```diff
@@ -0,0 +1,76 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#include "DrawingTest.h"
+
+#if WINOBJC
+#include "CGContextInternal.h"
```
ms-jihua: nit: #import, add a newline after

Frameworks/CoreGraphics/CGContext.mm
ms-jihua: @DHowett-MSFT out of curiosity, what would be your proposed alternative to the override flag?

rajsesh-msft: Would we not need to update the counter even if it is "force drawn"?

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -362,6 +362,20 @@ inline HRESULT GetTextRenderingParams(IDWriteRenderingParams* originalParams, ID
...
 
+    inline HRESULT Flush() {
+        // We cannot use use ID2D1::Flush here, mainly due to:
+        // 1. Flush flushes the D2D batch, but does not flush the D3D batch, we do not have a way to flush
+        // the D3D batch for WicBitmapRenderTarget. (Not: During the EndDraw procedure this occurs.)
+        // 2. Sometimes a shadow copy of our buffer can be used to do the actual rendering, Flush does not force the copy
+        // over of the shadow copy into the original buffer (EndDraw does)
+        // Thus the best way is to implement Flush is through EndDraw
+        if (_beginEndDrawDepth > 0) {
+            RETURN_IF_FAILED(PopEndDraw(true));
```
rajsesh-msft: The way this is written: if PopEndDraw failed due to some reason (device lost), we will leave the begin/end draw commands out of balance.

Frameworks/CoreGraphics/CGContext.mm
msft-Jeyaram: @DHowett-MSFT other alternative suggestions are welcome 😄

Frameworks/CoreGraphics/CGContext.mm
msft-Jeyaram: if PopEndDraw fails, CGContextFlush would FailFast so all states are lost.

---
In reply to: [105751972](https://github.com/Microsoft/WinObjC/pull/2208#discussion_r105751972) [](ancestors = 105751972)

Frameworks/CoreGraphics/CGContext.mm
msft-Jeyaram: Given forceEndDraw can be used by anyone, i would prefer the counter to be manipulated.
regardless of manipulation, if someone does not start a PushBeginDraw.forceEndDraw after doing a PopEndDraw.forceEndDraw, we'll get into an invalid state.
So again it depends on how it's being used.

---
In reply to: [105750699](https://github.com/Microsoft/WinObjC/pull/2208#discussion_r105750699) [](ancestors = 105750699)

msft-Jeyaram:
ping!

Frameworks/CoreGraphics/CGContext.mm
aballway: Though the contract of this is to return the HRESULT, not FAIL.  We should FAIL in here if we're not going to balance the states.  A consumer who handles a failed flush may not FAIL and be stuck with a broken Context

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_FlushTests.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+    // Validate only the red fill rect is executed.
+    EXPECT_EQ(dataPtr[0], 00);
+    EXPECT_EQ(dataPtr[1], 00);
+    EXPECT_EQ(dataPtr[2], 0xff);
+    EXPECT_EQ(dataPtr[3], 0xff);
+
+    for (int i = 0; i < 3; ++i) {
+        // Multiple flushes should work.
+        CGContextFlush(context);
```
aballway: for this should check that the data doesn't change after the redundant flushes

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_FlushTests.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+    CGContextSetRGBFillColor(context, 0, 0, 1, 1);
+    CGContextFillRect(context, CGRectMake(0, 0, 300, 300));
+
+    // We should still have red & blue rectangle should not show up.
+    unsigned char* dataPtr = static_cast<unsigned char*>(CGBitmapContextGetData(context));
+    ASSERT_NE(dataPtr, nullptr);
+
+    // Validate only the red fill rect is executed.
+    EXPECT_EQ(dataPtr[0], 00);
```
aballway: nit: 0x00, 00 would be octal, and though they're equivalent it's nice to be consistent

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_FlushTests.cpp
```diff
@@ -0,0 +1,132 @@
...
+    for (int i = 0; i < sDrawCount; ++i) {
+        _CGContextPushBeginDraw(context);
+    }
+
+    CGContextSetRGBFillColor(context, 1, 0, 0, 1);
+    CGContextFillRect(context, bounds);
+
+    for (int i = 0; i < sDrawCount; ++i) {
+        // Multiple flushes should work.
+        CGContextFlush(context);
```
aballway: same here

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -479,17 +493,17 @@ inline bool ShouldDraw() {
...
+    inline void PushBeginDraw(bool forceBeginDraw = false) {
+        if (forceBeginDraw || ((_beginEndDrawDepth)++ == 0)) {
             if (SUCCEEDED(_firstErrorHr)) {
                 deviceContext->BeginDraw();
             }
         }
     }
 
-    inline HRESULT PopEndDraw() {
+    inline HRESULT PopEndDraw(bool forceEndDraw = false) {
```
aballway: could get around the flag here by pulling out the innards into some private _PopEndDraw which actually calls the end draw, which would be called by this (if beginEndDrawDepth == 1) and directly from Flush (since it already does the _beginEndDrawDepth check).

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_FlushTests.cpp
msft-Jeyaram: it's validated below, so no need here.

---
In reply to: [106073225](https://github.com/Microsoft/WinObjC/pull/2208#discussion_r106073225) [](ancestors = 106073225)

Frameworks/CoreGraphics/CGContext.mm
msft-Jeyaram: if the state is out of balance, any subsequent requests that end up with D2D End will fail. I rather not break the pattern here and start introducing failures.

---
In reply to: [106073049](https://github.com/Microsoft/WinObjC/pull/2208#discussion_r106073049) [](ancestors = 106073049)

Frameworks/CoreGraphics/CGContext.mm
msft-Jeyaram: The whole point was to make it more usable, we don't want to start making private functions for every single case.
This seems like too much dancing around to do the thing we are going to do any way.

---
In reply to: [106077266](https://github.com/Microsoft/WinObjC/pull/2208#discussion_r106077266) [](ancestors = 106077266)

msft-Jeyaram:
@aballway  pingy

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_FlushTests.cpp
aballway: nit: this is not resolved

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_FlushTests.cpp
aballway: It only validates the first 4 bytes, it doesn't validate that after every flush call the data doesn't change.

Frameworks/CoreGraphics/CGContext.mm
aballway: Which is why we shouldn't let it get out of balance and maintain a functional context.

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_FlushTests.cpp
```diff
@@ -0,0 +1,132 @@
...
+    EXPECT_EQ(dataPtr[1], 0x00);
+    EXPECT_EQ(dataPtr[2], 0x00);
+    EXPECT_EQ(dataPtr[3], 0xff);
+
+    // Add some extra drawings
+    CGContextClearRect(context, CGRectMake(100, 100, 200, 300));
+
+    CGPoint center = _CGRectGetCenter(bounds);
+
+    CGMutablePathRef concentricCirclesPath = CGPathCreateMutable();
```
aballway: nit: StrongCF

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_FlushTests.cpp
msft-Jeyaram: include standard is fine here

---
In reply to: [106512085](https://github.com/Microsoft/WinObjC/pull/2208#discussion_r106512085) [](ancestors = 106512085)

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_FlushTests.cpp
msft-Jeyaram: again not needed, the drawing would catch it.
there is enough variable-ness to this. if we really want to go mad, we can check after every single call from start to end,.

---
In reply to: [106512296](https://github.com/Microsoft/WinObjC/pull/2208#discussion_r106512296) [](ancestors = 106512296)

Frameworks/CoreGraphics/CGContext.mm
msft-Jeyaram: but that can happen any time...
This is to be generic.

---
In reply to: [106513837](https://github.com/Microsoft/WinObjC/pull/2208#discussion_r106513837) [](ancestors = 106513837)

PR #2213, opened 2017-03-12
Rebase PackageReference work from Memontic-ms and add overrides.

bbowman:


bbowman:
>     <ProjectReference Include="$(MSBuildThisFileDirectory)..\WinObjC.Frameworks.ThirdParty\WinObjC.Frameworks.ThirdParty.nuproj" />

This effectively removes the need to do the BuildingFrameworksCore check, which is incorrect right now. (@memontic-ms you'd want BuildingFrameworksCore to be in each of these projects and then to have this project reference the tip of the subtree).

---
Refers to: build/WinObjC.Frameworks.Core/WinObjC.Frameworks.Core.nuproj:64 in 38ccfc2. [](commit_id = 38ccfc2f2425a89173998a071b58d15fff4791a3, deletion_comment = False)

build/WinObjC.Frameworks/WinObjC.Frameworks.Packaging.props
```diff
@@ -2,7 +2,7 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
     
   <PropertyGroup>
-      <PackageId>WinObjC.Frameworks</PackageId>
+      <PackageId Condition="'$(IsPackagingProject)' == 'true'">WinObjC.Frameworks</PackageId>
```
bbowman: >'$(IsPackagingProject)' == 'true' [](start = 28, length = 33)

Probably remove these since no one weird imports these now.

PR #2218, opened 2017-03-13
Implement NSArray read&write URL

### LINKED ISSUE #2069: Implement NSArray URL init/write methods
Although write to http url (post) doesn't work now, we should be able to implement this in a protocol agnostic way.  

initWithContentsOfURL:
arrayWithContentsOfURL:
writeToURL:atomically:

aballway:
Implement [NSArray  writeToURL:atomically:], [NSArray arrayWithContentsOfURL:], and [NSArray initWithContentsOfURL:].

Fixes #2069

### DIFF
--- a/Frameworks/Foundation/NSArray.mm
+++ b/Frameworks/Foundation/NSArray.mm
@@ -37,6 +37,8 @@
 
 static const wchar_t* TAG = L"NSArray";
 
+static const CFPropertyListFormat sc_plistFormat = kCFPropertyListBinaryFormat_v1_0;
+
 @implementation NSArray
 
 BASE_CLASS_REQUIRED_IMPLS(NSArray, NSArrayPrototype, CFArrayGetTypeID);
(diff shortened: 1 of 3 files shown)


Frameworks/Foundation/NSArray.mm
```diff
@@ -996,21 +995,29 @@ - (NSIndexSet*)indexesOfObjectsAtIndexes:(NSIndexSet*)indexSet
...
+    NSData* data = [NSData dataWithContentsOfURL:aURL];
+    return static_cast<NSArray*>(
+        CFPropertyListCreateWithData(nullptr, static_cast<CFDataRef>(data), kCFPropertyListImmutable, nullptr, nullptr));
 }
 
 /**
- @Status Stub
- @Notes
+ @Status Caveat
+ @Notes Only file:// URLs supported. atomically parameter not supported.
```
DHowett-MSFT: I'm not sure about noting either of these here; they're gaps from NSData, really, and if we fix those this works "for free"

aballway: Whenever NSData is fixed yes this would work for free, but this still isn't interoperable, and marking it as so would be incorrect.

PR #2221, opened 2017-03-13
 Implement CGPathCreateCopyByStrokingPath

MSFTFox:
Implement CopyByStrokingPath by creating a new geometry and widening the current geometry into it. A side effect of this process is that all curves are approximated into straight lines.

CGPathCreateCopyByDashingPath cannot be implemented in this was as the Widen API will create a path around the current path and cannot be used to simply apply a stroke style to modify the path itself.


Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -273,6 +273,44 @@ HRESULT AddGeometryToPathWithTransformation(const ID2D1Geometry* geometry, const
...
+                                                    0,
+                                                    &newStrokeStyle));
+
+        D2D1_MATRIX_3X2_F d2dTransform = D2D1::IdentityMatrix();
+        if (transform) {
+            d2dTransform = __CGAffineTransformToD2D_F(*transform);
+        }
+        RETURN_IF_FAILED(path->GetPathGeometry()->Widen(lineWidth, newStrokeStyle.Get(), d2dTransform, newBackingSink.Get()));
+
+        newBackingSink->SetFillMode(D2D1_FILL_MODE_WINDING);
```
DHowett-MSFT: do we need to do this _before_ we widen into it?

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -273,6 +273,44 @@ HRESULT AddGeometryToPathWithTransformation(const ID2D1Geometry* geometry, const
...
+
+        pathGeometry = newPath;
+        geometrySink = Make<_CGPathCustomSink>(newBackingSink.Get());
```
DHowett-MSFT: do we know this new sink is in a state that matches up with the externals we're tracking? It seems like we might corrupt currentPoint, lastTransform, etc.

tests/unittests/CoreGraphics.Drawing/CGPathDrawingTests.cpp
```diff
@@ -663,4 +663,61 @@ DRAW_TEST_F(CGPath, AddLinesWithoutMove, UIKitMimicTest<>) {
...
+    CGPathAddLineToPoint(thepath, NULL, xstart + .25 * width, ystart + .9 * height);
+    CGPathAddLineToPoint(thepath, NULL, xstart + .5 * width, ystart + .7 * height);
+    CGPathAddLineToPoint(thepath, NULL, xstart + .75 * width, ystart + .9 * height);
+    CGPathAddLineToPoint(thepath, NULL, xstart + .7 * width, ystart + .5 * height);
+    CGPathAddLineToPoint(thepath, NULL, xstart + .9 * width, ystart + .3 * height);
+    CGPathAddLineToPoint(thepath, NULL, xstart + .65 * width, ystart + .3 * height);
+    CGPathAddLineToPoint(thepath, NULL, xstart + .5 * width, ystart + .1 * height);
+    CGPathCloseSubpath(thepath);
+
+    CGPathRef widenedPath = CGPathCreateCopyByStrokingPath(thepath, NULL, 5, kCGLineCapRound, kCGLineJoinRound, 3);
```
DHowett-MSFT: please use `woc::MakeStrongCF` for this; you're leaking it.

tests/unittests/CoreGraphics.Drawing/CGPathDrawingTests.cpp
```diff
@@ -663,4 +663,61 @@ DRAW_TEST_F(CGPath, AddLinesWithoutMove, UIKitMimicTest<>) {
     CGContextStrokePath(context);
 
     CGPathRelease(thepath);
+}
+
+DRAW_TEST_F(CGPath, WidenTest_Stroking, UIKitMimicTest<>) {
+    CGContextRef context = GetDrawingContext();
+    applyBounds = GetDrawingBounds();
```
DHowett-MSFT: ? ? how does this compile

Frameworks/CoreGraphics/CGPath.mm
MSFTFox: That actually doesn't need to be there, but the fill mode does not affect the widen call in any way. This call should be removed since the baking sink already has this fill mode applied to it from the initial creation of this new path.

Frameworks/CoreGraphics/CGPath.mm
MSFTFox: That's a good point, we need additional tests to track behavior beyond the creation of this widened/stroked path.

MSFTFox:
As for the various parameters to pass in, I will two tests (for straight lines as well as curves) into parameter tests with the various line caps, joins, and some widening widths.

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -1025,10 +1025,6 @@ void CGContextReplacePathWithStrokedPath(CGContextRef context) {
 
     auto& state = context->CurrentGState();
 
-// TODO GH#xxxx When CGPathCreateCopyByStrokingPath is no longer stubbed, remove the diagnostic suppression.
-#pragma clang diagnostic push
```
aballway: 👍

tests/unittests/CoreGraphics.Drawing/CGPathDrawingTests.cpp
aballway: It's static, and used in a bunch of tests here.  Just change this to local in each test.

MSFTFox: applyBounds was a static Rect used earlier in the file. It should not be used here. Fixing + other tests.

tests/unittests/CoreGraphics.drawing/CGContextDrawing_WidenPathTests.cpp
```diff
@@ -0,0 +1,205 @@
...
+            default:
+                break;
+        }
+
+        return CFStringCreateWithFormat(
+            nullptr, nullptr, CFSTR("TestImage.CopyByStroking_Lines.%f.%f.%s.%s.png"), lineWidth, miterLimit, lineJoinName, lineCapName);
+    }
+};
+
+class CGPathCopyByStroking_Curves : public WhiteBackgroundTest<>,
```
DHowett-MSFT: These can be serviced by the same class.

Your `DRAW_TEST_P(...)` can have the same class on the left and a different name on the right.

To differentiate them in `CreateOutputFilename`, you can look at the blending tests.

tests/unittests/CoreGraphics.drawing/CGContextDrawing_WidenPathTests.cpp
```diff
@@ -0,0 +1,205 @@
...
+            default:
+                break;
+        }
+
+        return CFStringCreateWithFormat(
+            nullptr, nullptr, CFSTR("TestImage.CopyByStroking_Arcs.%f.%f.%s.%s.png"), lineWidth, miterLimit, lineJoinName, lineCapName);
+    }
+};
+
+void drawStrokeTestWithParams(
```
DHowett-MSFT: static this

tests/unittests/CoreGraphics.drawing/CGContextDrawing_WidenPathTests.cpp
```diff
@@ -0,0 +1,205 @@
...
+    drawStrokeTestWithParams(context, thepath, lineCap, lineJoin, lineWidth, miterLimit, useContext);
+
+    CGPathRelease(thepath);
+}
+
+INSTANTIATE_TEST_CASE_P(CGPathWidenTest,
+                        CGPathCopyByStroking_Lines,
+                        ::testing::Combine(::testing::ValuesIn(lineCaps),
+                                           ::testing::ValuesIn(lineJoins),
+                                           ::testing::ValuesIn(lineWidths),
```
DHowett-MSFT: we might want to test the zero-width cases separately so we don't spit out a hundred identical test images with different names.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -67,14 +68,47 @@
...
+    }
+
+    STDMETHOD_(void, AddQuadraticBezier)(_In_ CONST D2D1_QUADRATIC_BEZIER_SEGMENT* bezier) {
+        m_geometrySink->AddQuadraticBezier(bezier);
+        m_lastPoint = bezier->point2;
+    }
+
+    STDMETHOD_(void, AddQuadraticBeziers)(_In_reads_(beziersCount) CONST D2D1_QUADRATIC_BEZIER_SEGMENT* beziers, UINT32 beziersCount) {
+        m_geometrySink->AddQuadraticBeziers(beziers, beziersCount);
+        m_lastPoint = beziers[beziersCount - 1].point2;
```
DHowett-MSFT: since for arcs these are the control points, is this accurate?

tests/unittests/CoreGraphics.drawing/CGContextDrawing_WidenPathTests.cpp
```diff
@@ -0,0 +1,359 @@
...
+                                        addLine ? "AddLine" : "NoLine",
+                                        useContext ? "Context" : "Path");
+    }
+};
+
+class CGPathCopyByStroking_Curves : public WhiteBackgroundTest<>,
+                                    public ::testing::WithParamInterface<::testing::tuple<CGLineCap, CGFloat, bool, bool>> {
+    CFStringRef CreateOutputFilename() {
+        CGLineCap lineCap = ::testing::get<0>(GetParam());
+        CGFloat miterLimit = ::testing::get<1>(GetParam());
```
DHowett-MSFT: since curves aren't mitered, you could even remove _this_ parameter!

tests/unittests/CoreGraphics.drawing/CGContextDrawing_WidenPathTests.cpp
```diff
@@ -0,0 +1,359 @@
...
+                break;
+        }
+
+        return CFStringCreateWithFormat(nullptr,
+                                        nullptr,
+                                        CFSTR("TestImage.CopyByStroking_Lines.%f.%s.%s.%s.png"),
+                                        miterLimit,
+                                        lineJoinName,
+                                        lineCapName,
+                                        addLine ? "AddLine" : "NoLine",
```
DHowett-MSFT: what happens when you add a line from a completed stroked path is sorta an implementation detail; you could consider dropping that parameter from every test to half the search space.

Frameworks/CoreGraphics/CGPath.mm
MSFTFox: A quadratic bezier curve has the form, {point1, point2} where point1 is the only control point and point2 is the ending point of the curve.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -46,7 +46,22 @@
...
+          m_lastPoint{ 0, 0 },
+          m_isFigureOpen(false),
+          m_allowsFigureCalls(false),
+          m_hasGeometryStarted(false),
+          m_lastClose(D2D1_FIGURE_END_CLOSED),
+          m_allowsFinalFigureCall(false),
+          m_delayedEndFigure(false) {
+    }
+
+    STDMETHOD_(void, _delayedEndFigureCall)() {
```
rajsesh-msft: >STDMETHOD_ [](start = 4, length = 10)

This doesnt have to be STDMETHOD_() because this is not part of the interface ID2D1GeometrySink

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -46,7 +46,22 @@
         return m_geometrySink.Get();
     }
 
-    _CGPathCustomSink(_In_ ID2D1GeometrySink* sink) : m_geometrySink(sink), m_lastPoint{ 0, 0 }, m_isFigureOpen(false) {
+    _CGPathCustomSink(_In_ ID2D1GeometrySink* sink)
+        : m_geometrySink(sink),
+          m_lastPoint{ 0, 0 },
+          m_isFigureOpen(false),
+          m_allowsFigureCalls(false),
+          m_hasGeometryStarted(false),
```
rajsesh-msft: >m_hasGeometryStarted [](start = 10, length = 20)

some comments on what these variables are tracking will be helpful.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -58,42 +73,105 @@
...
+        m_lastPoint = arc->point;
+    }
+
+    STDMETHOD_(void, SetAllowsFigureCalls)(bool allows) {
+        m_allowsFigureCalls = allows;
+    }
+
+    STDMETHOD_(void, _BeginFigure)(D2D1_POINT_2F startPoint, D2D1_FIGURE_BEGIN figureBegin) {
+        _delayedEndFigureCall();
         if (m_isFigureOpen) {
```
rajsesh-msft: _EndFigure() is already doing this check, so you can remove this.

build/Tests/UnitTests/CoreGraphics.Drawing/CoreGraphics.Drawing.UnitTests.vcxproj
```diff
@@ -232,6 +232,7 @@
   </ItemDefinitionGroup>
   <ItemGroup>
     <ClangCompile Include="$(StarboardBasePath)\tests\unittests\CoreGraphics.Drawing\TAEFEntryPoint.cpp" />
+    <ClangCompile Include="..\..\..\..\tests\unittests\CoreGraphics.drawing\CGContextDrawing_WidenPathTests.cpp" />
```
msft-Jeyaram: >..\..\..\..\ [](start = 27, length = 12)

use (StarboardBasePath)\

Frameworks/CoreGraphics/CGPath.mm
msft-Jeyaram: nit: can this be private? it's not being used outside the function

msft-Jeyaram: also comment on why this is being applied to some of the functions vs the other will help.

tests/testapps/CGCatalog/CGCatalog/CGPathAddArcFullCircleViewController.m
```diff
@@ -49,7 +49,7 @@ - (void)loadView {
 
         CGPathRelease(thepath);
 
-        [super drawComparisonCGImageFromImageName:@"AddArc" intoContext:currentContext];
```
rajsesh-msft: >AddArc [](start = 52, length = 6)

Still need this image elsewhere?

tests/testapps/CGCatalog/CGCatalog.vsimporter/CGCatalog-WinStore10/CGCatalog.vcxproj
```diff
@@ -192,6 +192,7 @@
     <SBResourceCopy Include="..\..\CGCatalog\Resources\PathAddRect.png" />
     <SBResourceCopy Include="..\..\CGCatalog\Resources\PathApply.png" />
     <SBResourceCopy Include="..\..\CGCatalog\Resources\RoundedRect.png" />
+    <Image Include="..\..\CGCatalog\Resources\FullCircles.png" />
```
rajsesh-msft: >FullCircles [](start = 46, length = 11)

SBResourceCopy this guy?

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -105,10 +183,46 @@
...
+
+    STDMETHOD_(void, SetStartingPoint)(D2D1_POINT_2F startPoint) {
+        m_startPoint = startPoint;
+    }
+
+    STDMETHOD_(void, SetBackingSink)(ID2D1GeometrySink* sink) {
+        m_geometrySink = sink;
+    }
+
+    STDMETHOD_(void, AllowAllExceptFinalEndFigure)(bool ignore) {
```
rajsesh-msft: >ignore [](start = 56, length = 6)

From the name of the method, and the variable you are setting, it looks like m_allowsFinalFigureCall = !ignore?  Also, will pick a different name for the variable.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -58,42 +73,105 @@
...
+        m_lastPoint = beziers[beziersCount - 1].point2;
+    }
+
+    STDMETHOD_(void, AddArc)(_In_ CONST D2D1_ARC_SEGMENT* arc) {
+        _delayedEndFigureCall();
+        m_geometrySink->AddArc(arc);
+        m_lastPoint = arc->point;
+    }
+
+    STDMETHOD_(void, SetAllowsFigureCalls)(bool allows) {
```
rajsesh-msft: These methods that are not part of the ID2D1 geometry sink should not be STDMETHOD (rather they dont have to be)

include/CoreGraphics/CGPath.h
```diff
@@ -43,8 +43,8 @@ typedef CF_ENUM(CFIndex, CGLineCap) {
 };
 
 typedef CF_ENUM(CFIndex, CGLineJoin) {
-    kCGLineJoinMiter = 0, // Default
-    kCGLineJoinBevel = 3, // D2D: Miter or Bevel
+    kCGLineJoinMiter = 0, // Default/Miter
+    kCGLineJoinBevel = 1, // D2D: Bevel
```
msft-Jeyaram: why did this change?

tests/unittests/CoreGraphics.drawing/CGContextDrawing_WidenPathTests.cpp
```diff
@@ -0,0 +1,325 @@
...
+};
+
+class CGPathCopyByStroking_Curves : public WhiteBackgroundTest<>,
+                                    public ::testing::WithParamInterface<::testing::tuple<CGLineCap, CGFloat, bool>> {
+    CFStringRef CreateOutputFilename() {
+        CGLineCap lineCap = ::testing::get<0>(GetParam());
+        CGFloat miterLimit = ::testing::get<1>(GetParam());
+        bool useContext = ::testing::get<2>(GetParam());
+
+        char* lineCapName;
```
msft-Jeyaram: const char

tests/unittests/CoreGraphics.drawing/CGContextDrawing_WidenPathTests.cpp
```diff
@@ -0,0 +1,325 @@
...
+};
+
+class CGPathCopyByStroking_OpenFigure : public WhiteBackgroundTest<>,
+                                        public ::testing::WithParamInterface<::testing::tuple<CGLineJoin, CGFloat, bool>> {
+    CFStringRef CreateOutputFilename() {
+        CGLineJoin lineJoin = ::testing::get<0>(GetParam());
+        CGFloat miterLimit = ::testing::get<1>(GetParam());
+        bool useContext = ::testing::get<2>(GetParam());
+
+        char* lineJoinName;
```
msft-Jeyaram: same here and other places.
You are assigning a const char to char

tests/unittests/CoreGraphics.drawing/CGContextDrawing_WidenPathTests.cpp
```diff
@@ -0,0 +1,325 @@
...
+
+DRAW_TEST_P(CGPathCopyByStroking_Lines, PathWiden) {
+    CGContextRef context = GetDrawingContext();
+    CGRect drawingBounds = GetDrawingBounds();
+    CGFloat width = drawingBounds.size.width;
+    CGFloat height = drawingBounds.size.height;
+    CGFloat xstart = drawingBounds.origin.x;
+    CGFloat ystart = drawingBounds.origin.y;
+
+    CGMutablePathRef thepath = CGPathCreateMutable();
```
msft-Jeyaram: >thepath [](start = 21, length = 7)

AutoCF

tests/unittests/CoreGraphics.drawing/CGContextDrawing_WidenPathTests.cpp
```diff
@@ -0,0 +1,325 @@
...
+}
+
+void CreateOpenFigurePaths(CGContextRef context,
+                           CGFloat xstart,
+                           CGFloat ystart,
+                           CGFloat width,
+                           CGFloat height,
+                           CGLineJoin lineJoin,
+                           CGFloat miterLimit,
+                           bool useContext) {
```
msft-Jeyaram: static

tests/testapps/CGCatalog/CGCatalog/CGPathAddArcFullCircleViewController.m
MSFTFox: This image is for another test.

PR #2234, opened 2017-03-14
Fixed WinObjC.Packaging versioning

memontic-ms:
It was not updated to GitVersionTask and there was a syntax error


bbowman:
>   

Change.

---
Refers to: tools/WinObjC.Packaging/WinObjC.Packager.props:21 in f71c8bd. [](commit_id = f71c8bd49802986dbdb68657b7d8b7d1183ece31, deletion_comment = False)

PR #2235, opened 2017-03-14
Add GitVersionTask and NuGet.Build.Packaging as WinObjC.Packaging deps

memontic-ms:
The WinObjC.Packaging package was relying on InferLegacyPackageReferences being set to true. Recent changes in winobjc.nuproj.common.props now have that set as false. Decided to instead include the package dependency without relying on the project.json.

Note 1: We may want to consider updating the versions of these two included packages, GitVersionTask and NuGet.Build.Packaging. They are currently set to a version that is known to work for WinObjC.Packaging.

Note 2: Once tools.sln is switched to PackageReference instead of project.json we can do fancy things here to use the same versions of GitVersionTask and NuGet.Build.Packaging that the WinObjC.Packaging project is using.


bbowman:
Looks Good!

PR #2236, opened 2017-03-15
Implement NSArray index(es)Of* and enumerateObjectsAtIndexes methods

### LINKED ISSUE #2070: finish NSArray index* methods that are stubbed out
indexOfObjectIdenticalTo:inRange:
indexOfObjectWithOptions:passingTest:

The following may need completion of NSIndexSet, in which case we will break this out.

indexOfObjectAtIndexes:options:passingTest:
indexesOfObjectsPassingTest:
indexesOfObjectsWithOptions:passingTest:
indexesOfObjectsAtIndexes:options:passingTest:

### LINKED ISSUE #2072: Implement NSArray enumerateObjectWithIndexes:options:usingBlock:

aballway:
fixes #2070 
fixes #2072


Frameworks/Foundation/NSArray.mm
```diff
@@ -965,33 +941,44 @@ - (NSString*)descriptionWithLocale:(id)locale indent:(NSUInteger)level {
 }
 
 /**
- @Status Stub
+ @Status Interoperable
  @Notes
```
DHowett-MSFT: nit: remove empty notes. `<ESC>:g/@Notes$/d`

Frameworks/Foundation/NSArray.mm
```diff
@@ -1041,32 +1028,43 @@ - (NSArray*)sortedArrayUsingFunction:(NSInteger (*)(id, id, void*))comparator co
...
 */
 - (NSUInteger)indexOfObjectIdenticalTo:(id)anObject inRange:(NSRange)range {
-    UNIMPLEMENTED();
-    return StubReturn();
+    if (range.location + range.length > [self count]) {
+        [NSException raise:NSRangeException
+                    format:@"Range {%d, %d} larger than array of size %d.", range.location, range.length, [self count]];
```
DHowett-MSFT: This doesn't match the other exception formats NSArray uses

Frameworks/Foundation/NSArray.mm
```diff
@@ -1041,32 +1028,43 @@ - (NSArray*)sortedArrayUsingFunction:(NSInteger (*)(id, id, void*))comparator co
...
+    }
+
+    NSUInteger end = NSMaxRange(range);
```
DHowett-MSFT: Move this above 1035 and use `end` in the `if(...)`

Frameworks/Foundation/NSArray.mm
```diff
@@ -1041,32 +1028,43 @@ - (NSArray*)sortedArrayUsingFunction:(NSInteger (*)(id, id, void*))comparator co
...
+    for (NSUInteger i = range.location; i < end; ++i) {
+        if ([self objectAtIndex:i] == anObject) {
```
DHowett-MSFT: This is curious; if you prefer `getObjects:range:`, a "fast" NSArray implementation can turn that into an `n`-wide buffer copy into your provided buffer. Right now, _NSCFArray doesn't do that... but it could!

Arguably, it should, for optimization purposes. It would call `CFArrayGetValues` under the hood.

DHowett-MSFT: This can apply anywhere you iterate manually over a range.

Frameworks/Foundation/NSArray.mm
```diff
@@ -1041,32 +1028,43 @@ - (NSArray*)sortedArrayUsingFunction:(NSInteger (*)(id, id, void*))comparator co
...
- @Status Stub
+ @Status Interoperable
  @Notes
 */
 - (NSUInteger)indexOfObjectAtIndexes:(NSIndexSet*)indexSet
                              options:(NSEnumerationOptions)opts
                          passingTest:(BOOL (^)(id, NSUInteger, BOOL*))predicate {
-    UNIMPLEMENTED();
-    return StubReturn();
+    NSIndexSet* matching = [self indexesOfObjectsAtIndexes:indexSet options:opts passingTest:predicate];
```
DHowett-MSFT: Optimization:
If you add another level of block here, you can stop enumeration on the first hit instead of continuing to aggregate every index and then discarding the rest of them.

Frameworks/Foundation/NSArray.mm
aballway: Reference platform throws an NSRangeException for invalid ranges, should I opt for consistency here?

DHowett-MSFT: Sorry, I meant the wording was inconsistent :)

Frameworks/Foundation/NSArray.mm
aballway: should I make that part of these changes? I'm wary of changing too much in NSArray for possibly unnecessary optimization

Frameworks/Foundation/NSArray.mm
aballway: Reference platform will blow up on invalid ranges, even if a valid element is found, and this gives us that.

Frameworks/Foundation/NSArray.mm
aballway: Updated to reflect exception generated on reference platform

Frameworks/Foundation/NSArray.mm
```diff
@@ -1031,8 +1031,10 @@ - (NSUInteger)indexOfObjectIdenticalTo:(id)anObject inRange:(NSRange)range {
                            [self count]];
     }
 
+    id objects[range.length];
+    [self getObjects:objects range:range];
     for (NSUInteger i = range.location; i < end; ++i) {
```
DHowett-MSFT: nit: why not `i=0, < range.length`?

aballway: Considered it but this was the path of least change, didn't think it made a substantial difference in readability or perf

Frameworks/Foundation/NSArray.mm
```diff
@@ -1033,40 +1010,54 @@ - (NSArray*)sortedArrayWithOptions:(NSSortOptions)opts usingComparator:(NSCompar
...
- @Notes
+ @Status Interoperable
 */
 - (NSUInteger)indexOfObjectIdenticalTo:(id)anObject inRange:(NSRange)range {
-    UNIMPLEMENTED();
-    return StubReturn();
+    NSUInteger end = NSMaxRange(range);
+    if (end > [self count]) {
+        [NSException raise:NSRangeException
+                    format:@"[%s %s]: range {%d, %d} extends beyond bounds [ 0 .. %d]",
```
DHowett-MSFT: nit: missing `-` before `[%s`

Frameworks/Foundation/NSArray.mm
DHowett-MSFT: I fail to see what that has to do with this; using `*stop = YES;` to abort the iteration inside the block still has us calling `indexesOfObjects...` and getting the blowing-up behaviour.

Issue #2225, opened 2017-03-14
Implement NSMutableOrderedSet


alakoring:
Very handy implementation of a “mutable, ordered collection of distinct objects.”

Unclear if this is being implemented (see https://github.com/Microsoft/WinObjC/pull/482).

For my purposes, performance is not critical.

RG-J:
I am implementing most of the functions, expect to see a review out soon.

RJ

ArnOmsft:
@RG-J, thank you!

rajsesh-msft:
@RG-J how is this coming along?

RG-J:
@rajsesh-msft  I'll push my changes soon, just pulled it up to the latest code and setting up VS2017.

PR #2240, opened 2017-03-15
Re-implement NSOperationQueue

### LINKED ISSUE #869: NSOperationQueue is suboptimal (but it does work)

### LINKED ISSUE #1655: Crash in -[NSOperationQueue _doMainWork]
We’re seeing the following crash in NSOperationQueue / NSAtomicList:

```
HEAP_CORRUPTION_ACTIONABLE_LFHSegmentDoubleFree_DOUBLE_FREE_c0000374_Foundation.dll!NSAtomicListPop

Frame	Image	Function	Offset
0	ntdll	RtlpHpLfhSubsegmentFreeBlock	0xE5	
1	ntdll	RtlpHpSegFree	0x5C	
2	ntdll	RtlpHpFreeHeap	0x18A	
3	ntdll	RtlpFreeHeapInternal	0x4F	
4	ntdll	RtlpHpFreeWithExceptionProtection	0x2E	
5	ntdll	RtlFreeHeap	0x4AADB	
6	ucrtbase	_free_base	0x1B	
7	ucrtbase	free	0x18	
8	Foundation	NSAtomicListPop	0x85	
9	Foundation	RunOperationFromLists	0x96	
10	Foundation	-[NSOperationQueue _doMainWork]	0xD0	
11	Foundation	-[NSOperationQueue _workThread]	0x1B	
12	Foundation	-[NSThread main]	0x4A	
13	Foundation	_threadBody	0x4D	
14	Starboard	ThreadRoutine	0x3F	
15	kernel32	BaseThreadInitThunk	0x24	
16	ntdll	__RtlUserThreadStart	0x2F	
17	ntdll	_RtlUserThreadStart	0x1B	
```

### LINKED ISSUE #2228: Add support for NSOperation queuePriority
We would like to see support for `NSOperation.queuePriority` to change the relative priority of an operation in a queue.

This might depend on or be part of #869.

ms-jihua:
- Add support for
    - queuePriority
    - underlyingQueue
    - currentQueue
    - maxConcurrentOperationCount
    - KVO-compliance
- Fixes some edge cases (ie: hanging on certain non-ready operations)
- Add unit tests
- Add performance benchmark tests

Fixes #2228
Fixes #1655
Fixes #869


Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -1,443 +1,381 @@
...
+    switch (quality) {
+        case NSQualityOfServiceUserInteractive:
+            return QOS_CLASS_USER_INTERACTIVE;
+        case NSQualityOfServiceUserInitiated:
+            return QOS_CLASS_USER_INITIATED;
+        case NSQualityOfServiceUtility:
+            return QOS_CLASS_UTILITY;
+        case NSQualityOfServiceBackground:
+            // currently generates a warning:
+            // 'QOS_CLASS_BACKGROUND' is deprecated: QOS_CLASS_BACKGROUND is the same as DISPATCH_QUEUE_PRIORITY_LOW on WinObjC
```
DHowett-MSFT: do we have an issue ID for this?

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -1,443 +1,381 @@
...
-#import "Platform/EbrPlatform.h"
-#import "Foundation/NSString.h"
-#import "Foundation/NSOperationQueue.h"
-#import "Foundation/NSThread.h"
-#import "Foundation/NSAutoreleasePool.h"
-#import "Foundation/NSLock.h"
-#import <time.h>
-#import "LoggingNative.h"
-#import "NSOperationInternal.h"
+@interface NSOperationQueue_MainQueue : NSOperationQueue
```
DHowett-MSFT: If this is file-internal, i'd prefer a `_` prepended :smile:

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -1,443 +1,381 @@
...
+- (void)_setDispatchQueueUsingQualityOfService;
+@end
 
-    NSOperationQueuePriv() {
-        memset(this, 0, sizeof(NSOperationQueuePriv));
-        _threadRunningLock = PTHREAD_MUTEX_INITIALIZER;
-    }
-};
+@implementation NSOperationQueue {
+@private
```
DHowett-MSFT: nit: prefer ivars on class extension above instead of implementation

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -1,443 +1,381 @@
...
-    id workAvailable;
-    id suspendedCondition;
-    id allWorkDone;
-    id curOperation;
-    BOOL isSuspended;
+#pragma region NSOperationQueue
 
-    id _name;
+@interface NSOperationQueue ()
+// Private member functions
```
DHowett-MSFT: nit: comment redundant with location

DHowett-MSFT: (and with naming)

ms-jihua:
Here are perf numbers, comparing on Release with this change, develop, and #1676 which we backed out:
image

Things to note:
- The time it takes to add 1000 operations individually has doubled from develop. However, if we look under the hood we see that 
image
90% of this time is spent in KVO (!!), which we are looking at next.
- The time it takes to add 1000 operations together, and wait for them to finish, has largely remained the same. However, unlike develop, not waiting for them to finish provides no significant time discount - it appears that unlike develop, we are bottlenecked by KVO rather than the process of executing the operations, so our perf should improve once we look at KVO.

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -1,443 +1,381 @@
...
-NSAtomicListRef NSAtomicListSteal(NSAtomicListRef* listPtr) {
-    NSAtomicListRef ret;
-    do {
-        ret = *listPtr;
-    } while (CompareExchangePointer((void**)listPtr, NULL, ret) != ret);
-    return ret;
-}
+    // Protects _dispatchQueue, _underlyingQueue, and _qualityOfService
+    // Should never be locked before _concurrentOperationCountLock
+    std::recursive_mutex _dispatchQueueLock;
```
DHowett-MSFT: it worries me that we're mixing _dispatch_ queues with _STL_ mutexes and atomics.

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -1,443 +1,381 @@
...
 
-void NSAtomicListReverse(NSAtomicListRef* listPtr) {
-    struct NSAtomicListNode* cur = *listPtr;
-    struct NSAtomicListNode* prev = NULL;
-    struct NSAtomicListNode* next = NULL;
+    StrongId<NSMutableArray> _operations;
```
DHowett-MSFT: Are operations inherently indexed? Do we need to use an ordered set here for quick removal?

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -1,443 +1,381 @@
...
+    std::lock_guard<std::recursive_mutex> lock(_concurrentOperationCountLock);
+    NSInteger maxOps = self.maxConcurrentOperationCount;
+    return (maxOps <= 0) || (_currentConcurrentOperationCount < maxOps);
 }
 
-NSObject* NSAtomicListPop(NSAtomicListRef* listPtr) {
-    struct NSAtomicListNode* node = *listPtr;
-    if (!node) {
-        return NULL;
+// Private helper that tries to starts a single NSOperation on the dispatch queue and manages associated state.
```
DHowett-MSFT: nit: "tries to starts"

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -1,443 +1,381 @@
...
-    }
+        ++_currentConcurrentOperationCount;
+    } // _concurrentOperationCountLock scope
 
-    return node->elt;
-}
+    std::lock_guard<std::recursive_mutex> lock(_dispatchQueueLock);
+    dispatch_async(_dispatchQueue, ^{
+        // Check if currentQueue needs to be changed
+        NSMutableDictionary* threadDictionary = [[NSThread currentThread] threadDictionary];
```
DHowett-MSFT: [Curiosity only]

Why'd you choose to use NSThread's thread dictionary instead of a thread_local?

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -1,443 +1,381 @@
...
+        case NSQualityOfServiceUserInteractive:
+            return QOS_CLASS_USER_INTERACTIVE;
+        case NSQualityOfServiceUserInitiated:
+            return QOS_CLASS_USER_INITIATED;
+        case NSQualityOfServiceUtility:
+            return QOS_CLASS_UTILITY;
+        case NSQualityOfServiceBackground:
+            // currently generates a warning:
+            // 'QOS_CLASS_BACKGROUND' is deprecated: QOS_CLASS_BACKGROUND is the same as DISPATCH_QUEUE_PRIORITY_LOW on WinObjC
+            // should go away once libdispatch is updated
```
ms-jihua: #2241

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -1,443 +1,381 @@
...
-    NSOperationQueuePriv() {
-        memset(this, 0, sizeof(NSOperationQueuePriv));
-        _threadRunningLock = PTHREAD_MUTEX_INITIALIZER;
-    }
-};
+@implementation NSOperationQueue {
+@private
+    // The actual dispatch queue to which operations are dispatched
+    // Currently, changing between dispatch queues is done by reassigning this member
+    // TODO #:  WinObjC's libdispatch cannot create concurrent dispatch queues (DISPATCH_QUEUE_CONCURRENT just returns a serial queue)
```
ms-jihua: #2241

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -1,443 +1,381 @@
...
+        if (needToChangeCurrentQueue) {
+            [threadDictionary setObject:self forKey:_NSOperationQueueCurrentQueueKey];
+        }
 
-    do {
-        node->next = *listPtr;
-    } while (CompareExchangePointer((void**)listPtr, node, node->next) != node->next);
-}
+        // Restore state once the operation is done, and attempt to start any queued operations
+        auto cleanup = wil::ScopeExit([self, threadDictionary, currentQueue, needToChangeCurrentQueue]() {
```
DHowett-MSFT: clarity: Why's the queue drained as a "side effect of exiting"? This doesn't seem clearer than sequencing the operation start before the code inside this block.

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -1,443 +1,381 @@
...
-        *listPtr = NSAtomicListSteal(sourceListPtr);
-        // source lists are in LIFO order, but we want to execute operations in the order they were enqueued
-        // so we reverse the list before we do anything with it
-        NSAtomicListReverse(listPtr);
-        op = NSAtomicListPop(listPtr);
-    }
+// Private helper that sets _dispatchQueue based on the current quality of service
+- (void)_setDispatchQueueUsingQualityOfService {
+    std::lock_guard<std::recursive_mutex> lock(_dispatchQueueLock);
+    _dispatchQueue = dispatch_get_global_queue(_QOSClassForNSQualityOfService(self.qualityOfService), 0);
```
DHowett-MSFT: nit: no need for the getter call for QOS here

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -1,443 +1,381 @@
...
+        _maxConcurrentOperationCount = NSOperationQueueDefaultMaxConcurrentOperationCount;
+        _qualityOfService = NSQualityOfServiceDefault;
 
-    while (!stop) {
-        [self _doMainWork];
-        pthread_mutex_lock(&priv->_threadRunningLock);
-        if (![self hasMoreWork]) {
-            stop = TRUE;
-            priv->_thread = nil;
+        if (dispatch_queue_t underlyingQueue = [self underlyingQueue]) {
```
DHowett-MSFT: ! ! !

This wasn't allowed before (C++11); do we have CC on whether we can do this?

tests/unittests/Foundation/NSOperationTests.mm
```diff
@@ -586,4 +690,628 @@ - (void)main {
...
+    ASSERT_NO_THROW([mainQueue setUnderlyingQueue:nil]);
+    ASSERT_EQ([mainQueue underlyingQueue], dispatch_get_main_queue());
+    ASSERT_NO_THROW([mainQueue setUnderlyingQueue:dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0)]);
+    ASSERT_EQ([mainQueue underlyingQueue], dispatch_get_main_queue());
+    ASSERT_NO_THROW([mainQueue setUnderlyingQueue:dispatch_get_main_queue()]);
+    ASSERT_EQ([mainQueue underlyingQueue], dispatch_get_main_queue());
+}
+
+TEST(NSOperation, CurrentQueue) {
+// TODO #: WinObjC's implementation of NSThread does not consider this context the main thread - this is a bug
```
ms-jihua: #2242

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -1,443 +1,381 @@
...
+}
 
-        for (int i = 0; i < NSOperationQueuePriority_Count; i++) {
-            [priv->suspendedCondition lock];
-            while (priv->isSuspended) {
-                [priv->suspendedCondition wait];
+- (void)observeValueForKeyPath:(NSString*)keyPath ofObject:(id)object change:(NSDictionary*)change context:(void*)context {
+    if ([keyPath isEqualToString:@"isFinished"]) {
+        NSOperation* op = static_cast<NSOperation*>(object);
+        if (op.isFinished) {
```
DHowett-MSFT: nit: if we request the 'new' value from the change, we can avoid the synchronized call back into operation to query it.

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -1,443 +1,381 @@
...
+            // Minor optimization:
+            // Because of constraints in addOperation:, it is safe to assume that there is exactly one instance of op in the array
+            // removeObject: keeps going after finding first instance of object, which is unnecessary work
+            // Use indexOfObject:op instead
+            @synchronized(_operations.get()) {
+                NSUInteger index = [_operations indexOfObject:op];
```
DHowett-MSFT: Prefer `indexOfObjectIdenticalTo:`, as it will require far fewer `isEqual:` calls

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -1,443 +1,381 @@
...
+#import "NSOperationInternal.h"
+#import "Starboard.h"
 
-The above copyright notice and this permission notice shall be included in all copies or substantial portions of the
-Software.
+#import <atomic>
+#import <mutex>
+#import <queue>
+
+static const wchar_t* g_logTag = L"NSOperationQueue";
```
DHowett-MSFT: I know this name makes more sense, but our convention everywhere else is the shouty `TAG`

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -1,443 +1,381 @@
...
+            // Clean up state from addOperation:
+            [op removeObserver:self forKeyPath:@"isFinished" context:nil];
+            [op removeObserver:self forKeyPath:@"isReady" context:nil];
         }
-    } while (didWork);
-
-    [innerPool release];
-
-    return self;
+    } else if ([keyPath isEqualToString:@"isReady"]) {
```
DHowett-MSFT: Can `isReady` ever be transitioning to false? If so, we may have to handle that.

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -1,443 +1,381 @@
...
 /**
  @Status Interoperable
 */
-- (BOOL)hasMoreWork {
-    if (priv->curOperation != nil) {
-        return TRUE;
++ (NSOperationQueue*)currentQueue {
+    NSThread* currentThread = [NSThread currentThread];
+    if ([currentThread isMainThread]) {
+        return [[self class] mainQueue];
```
DHowett-MSFT: While it won't break anything, using `[self class]` in a `+` method is ambiguous as to whether it should be a meta-metaclass, or just a metaclass.

DHowett-MSFT: (technically, it is well-defined, but I thought this was an instance method when i saw the `-class` call)

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -1,443 +1,381 @@
...
-    /*
-    //  Add any dependencies
-    id dependencies = [op dependencies];
-    int count = [dependencies count];
-    for ( int i = 0; i < count; i ++ ) {
-    id curDep = [dependencies objectAtIndex:i];
-    [self addOperation:curDep];
+- (void)addOperation:(NSOperation*)op {
+    if (op.isExecuting || op.isFinished) {
+        @throw [NSException exceptionWithName:NSInvalidArgumentException reason:@"operation is already or finished executing" userInfo:nil];
```
DHowett-MSFT: We may want to have an exception formatter that prepends `-[ClassName selector:name:]` to our exceptions; i know we have a couple scattered around.

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -1,443 +1,381 @@
...
-    [priv->workAvailable signal];
+    [self willChangeValueForKey:@"operations"];
+    [self willChangeValueForKey:@"operationCount"];
 
-    pthread_mutex_lock(&priv->_threadRunningLock);
-    if (priv->_thread == nil) {
-        priv->_thread = [[NSThread alloc] initWithTarget:self selector:@selector(_workThread) object:nil];
-        [priv->_thread start];
-        [priv->_thread release];
+    @synchronized(_operations.get()) {
```
DHowett-MSFT: Should we be guarding the will/did?

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -1,443 +1,381 @@
...
-*/
-- (void)addOperationWithBlock:(void (^)())block {
-    id op = [NSOperation new];
-    [op setCompletionBlock:block];
-    [self addOperation:op];
-    [op release];
+    [self didChangeValueForKey:@"operationCount"];
+    [self didChangeValueForKey:@"operations"];
+
+    [op addObserver:self forKeyPath:@"isFinished" options:0 context:nil];
```
DHowett-MSFT: Can the operation change status between when we add it to the array and when we register for its notifications?

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -1,443 +1,381 @@
...
-- (id)operations {
-    id ret = [NSMutableArray array];
-
-    TraceVerbose(TAG, L"Should lock queue for this");
-    id cur = priv->curOperation;
-    if (cur != nil) {
-        [ret addObject:cur];
+- (NSArray<__kindof NSOperation*>*)operations {
+    @synchronized(_operations.get()) {
+        return [_operations copy];
```
DHowett-MSFT: copy + autorelease; otherwise, this leaks

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -446,138 +384,112 @@ - (void)cancelAllOperations {
...
 - (void)waitUntilAllOperationsAreFinished {
-    BOOL isWorking;
-
-    [priv->workAvailable lock];
-    isWorking = [self hasMoreWork];
-
-    if (isWorking) {
-        [priv->allWorkDone wait];
-        isWorking = [self hasMoreWork];
+    while (NSOperation* lastOp = [_operations lastObject]) {
```
DHowett-MSFT: What if another operation is added while we're waiting for them all to finish?

Frameworks/Foundation/NSOperationQueue.mm
DHowett-MSFT: (since it's under the same lock)

include/Foundation/NSOperationQueue.h
```diff
@@ -19,25 +19,23 @@
...
 FOUNDATION_EXPORT_CLASS
 @interface NSOperationQueue : NSObject
 + (NSOperationQueue*)currentQueue;
 + (NSOperationQueue*)mainQueue;
 - (void)addOperation:(NSOperation*)operation;
-- (void)addOperations:(NSArray*)ops waitUntilFinished:(BOOL)wait;
+- (void)addOperations:(NSArray<NSOperation*>*)ops waitUntilFinished:(BOOL)wait;
 - (void)addOperationWithBlock:(void (^)(void))block;
-@property (readonly, copy) NSArray* operations;
+@property (readonly, copy) NSArray<__kindof NSOperation*>* operations;
```
DHowett-MSFT: we may have to add the `suspended` property

tests/Benchmark/NSOperationQueueBenchmarkTests.mm
```diff
@@ -0,0 +1,87 @@
...
+        }
+    }
+}
+class AddOperation : public ::benchmark::BenchmarkCaseBase {
+    StrongId<NSOperationQueue> m_queue;
+
+public:
+    void PreRun() {
+        @autoreleasepool {
+            m_queue = [[NSOperationQueue new] autorelease];
```
DHowett-MSFT: you could use .attach here and avoid the round trip

Frameworks/Foundation/NSOperationQueue.mm
DHowett-MSFT: You beat me to it, and the other issue ID questions. Deleting.

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -1,443 +1,381 @@
...
+        return;
+    }
 
-    NSObject* elt = [[node->elt retain] autorelease];
-    delete node;
-    return elt;
-}
+    {
+        std::lock_guard<std::recursive_mutex> lock(_concurrentOperationCountLock);
+        if ((self.isSuspended) || (![self _belowMaxConcurrentOperations])) {
```
DHowett-MSFT: when we add the property for suspended, this should become `self.suspended`; the getter form is for `[self isSuspended]` use

Frameworks/Foundation/NSOperationQueue.mm
ms-jihua: define 'everywhere else', we've also had non-zero g_logTags...

Frameworks/Foundation/NSOperationQueue.mm
ms-jihua: oops somehow completely missed this

Frameworks/Foundation/NSOperationQueue.mm
ms-jihua: i forgot thread_local existed derp

DHowett-MSFT:
FWIW we have 2 `g_logTag`s, both in CoreText; we have 182 `TAG`s :)

Frameworks/Foundation/NSOperationQueue.mm
ms-jihua: don't we generally prefer wil::ScopeExit for safer cleanups if the inbetween function is relatively unknown/risky?

Frameworks/Foundation/NSOperationQueue.mm
ms-jihua: subclasses could override though?

Frameworks/Foundation/NSOperationQueue.mm
ms-jihua: we do it all the time as 'if (self = [super init])' though

Frameworks/Foundation/NSOperationQueue.mm
DHowett-MSFT: Absolutely. It feels like this is a little bigger than "cleanup" though, it'll even go and reschedule more operations.

Aside: Should we pay any heed to trying to run operations back-to-back without spinning up another async dispatch?

Frameworks/Foundation/NSOperationQueue.mm
DHowett-MSFT: We don't use the type-declaring version, though.

Frameworks/Foundation/NSOperationQueue.mm
ms-jihua: oh hm, i didn't think about this angle. i'll hold the lock on the operation then.

Frameworks/Foundation/NSOperationQueue.mm
ms-jihua: won't lastOp be set to a new value every loop?

include/Foundation/NSOperationQueue.h
ms-jihua: ? it's there, it's just hidden under the github cutoff where it doesn't like to show you the whole file

tests/Benchmark/NSOperationQueueBenchmarkTests.mm
ms-jihua: derp derp

include/Foundation/NSOperationQueue.h
DHowett-MSFT: Oh jeez, I was looking at the wrong file locally T_T

ms-jihua:
@DHowett-MSFT I am defeated

Frameworks/Foundation/NSOperationQueue.mm
ms-jihua: how come? this more has to do with synchronizing these members' getters/setters

Frameworks/Foundation/NSOperationQueue.mm
ms-jihua: Not inherently indexed, but the ordering needs to be consistent. Ordered set would be good, I kind of neglected other possibilities just because the operations property is an array. However, we don't actually have NSMutableOrderedSet yet. I'll keep an eye on #2225, and change this to use it once it's in.

Frameworks/Foundation/NSOperationQueue.mm
ms-jihua: You have a good point. I moved this to be async, and to be triggered in observing isFinished instead.

include/Foundation/NSOperation.h
```diff
@@ -64,7 +55,7 @@ FOUNDATION_EXPORT_CLASS
 - (void)addDependency:(NSOperation*)operation;
 - (void)removeDependency:(NSOperation*)operation;
 @property (readonly, copy) NSArray* dependencies;
-@property NSQualityOfService qualityOfService STUB_PROPERTY;
+@property NSQualityOfService qualityOfService;
 @property double threadPriority STUB_PROPERTY;
```
rajsesh-msft: >threadPriority [](start = 17, length = 14)

can we add support for this as well?

Frameworks/Foundation/NSOperationQueue.mm
ms-jihua: it seems easier than doing the dictionary lookup, cast to nsnumber, and get boolvalue, though

include/Foundation/NSOperation.h
ms-jihua: yeah - i seem to have lost a commit somewhere

Frameworks/Foundation/NSOperationQueue.mm
ms-jihua: I don't think it really matters. KVO has an expectation that the timing might be a little off, right?

Frameworks/Foundation/NSOperationInternal.h
```diff
@@ -18,4 +18,5 @@
 
 @interface NSOperation ()
 @property (nonatomic, assign, getter=_completionQueue, setter=_setCompletionQueue:) dispatch_queue_t _completionQueue;
+- (BOOL)_acquirePermissionToAddToQueue;
```
rajsesh-msft: >_acquirePermissionToAddToQueue [](start = 8, length = 30)

something about this name doesn't fit right with me.  We are not requesting any permissions here, perhaps just make this an atomic property InQueue and set it directly.

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -1,443 +1,381 @@
...
+            return QOS_CLASS_USER_INTERACTIVE;
+        case NSQualityOfServiceUserInitiated:
+            return QOS_CLASS_USER_INITIATED;
+        case NSQualityOfServiceUtility:
+            return QOS_CLASS_UTILITY;
+        case NSQualityOfServiceBackground:
+            // currently generates a warning:
+            // 'QOS_CLASS_BACKGROUND' is deprecated: QOS_CLASS_BACKGROUND is the same as DISPATCH_QUEUE_PRIORITY_LOW on WinObjC
+            // should go away once libdispatch is updated
+            return QOS_CLASS_BACKGROUND;
```
rajsesh-msft: >QOS_CLASS_BACKGROUND [](start = 19, length = 20)

Could we not introduce a "deprecated" warning in our new code if avoidable.

include/Foundation/NSOperation.h
ms-jihua: no wait i was thinking of another thing, i actually just missed this, sorry!

Frameworks/Foundation/NSOperationQueue.mm
ms-jihua: I think I'd prefer to disable the warning in this instance, since it's one that'll go away once we Do The Right Thing and update libdispatch.

Frameworks/Foundation/NSOperationInternal.h
ms-jihua: i like having it as an atomic operation so that it's harder to break contract, i'll try to think of a better name...

ms-jihua: going with _markInQueue for now...

Frameworks/Foundation/NSOperationQueue.mm
ms-jihua: no wait i didn't think this all the way through.

- isFinished is set at the end of [operation start]. if we schedule more operations when we observe isFinished, we end up doing so before we decrement _currentConcurrentOperationCount, causing _popAndStart to do the wrong thing.
- we can move the decrement to the isFinished observation, but i dislike this because 1) the increment and decrement ideally should be in the same function 2) an operation can finish through means besides _startOperation (ie: if a user decides to start it manually for whatever reason)
- as it currently is, we're basically checking if new operations need to be started when an old operation has finished, using the same async context. i think that this makes intuitive sense, so i'm going to leave it as is for now. (it might look less scary once you see the thread_local version also)

ms-jihua:
I just realized I pushed the branch to the main repo instead of my own fork. I'll keep the PR open for now and delete the branch when I'm done.

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -38,27 +38,52 @@ static inline long _QOSClassForNSQualityOfService(NSQualityOfService quality) {
...
+
+// STL compare for NSOperation*, sorting by queuePriority, putting the greatest priority first
+struct _NSOperation_ComparePrioritySTL {
+    bool operator()(NSOperation* lhs, NSOperation* rhs) {
+        return lhs.queuePriority > rhs.queuePriority;
+    }
+};
+
+// Used for [NSOperationQueue currentQueue]
+thread_local static StrongId<NSOperationQueue*> tls_currentQueue;
```
DHowett-MSFT: So this is sorta silly; ehren ran into an issue with destructors on file-level thread_locals; they never get called and then leak.

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -116,83 +133,90 @@ @implementation NSOperationQueue {
...
-
-        // If so, cache the previous currentQueue stored in the threadDictionary, and store the new currentQueue
-        if (needToChangeCurrentQueue) {
-            [threadDictionary setObject:self forKey:_NSOperationQueueCurrentQueueKey];
-        }
+        std::lock_guard<std::recursive_mutex> lock(_dispatchQueueLock);
+        dispatch_async(_dispatchQueue, ^{
+            // Check if currentQueue needs to be changed
+            StrongId<NSOperationQueue> currentQueue = tls_currentQueue;
+            bool needToChangeCurrentQueue = (![[NSThread currentThread] isMainThread]) && (![self isEqual:tls_currentQueue.get()]);
```
DHowett-MSFT: `isEqual` is likely overkill here; `==` will do

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -38,27 +38,52 @@ static inline long _QOSClassForNSQualityOfService(NSQualityOfService quality) {
...
+    bool operator()(NSOperation* lhs, NSOperation* rhs) {
+        return lhs.queuePriority > rhs.queuePriority;
+    }
+};
+
+// Used for [NSOperationQueue currentQueue]
+thread_local static StrongId<NSOperationQueue*> tls_currentQueue;
+
+// Used for key-value observing on held operations - faster than comparing keypaths
+static void* _NSOperationQueue_IsFinishedContext = &_NSOperationQueue_IsFinishedContext;
```
DHowett-MSFT: nit: you can use use `static char _xyz` and then `&_xyz` later; no need for initialization (or emitting a binary relocation that requires updating a pointer :smile:)

DHowett-MSFT: (doing it the aforementioned way is common in objc associated object scenarios; they use `key`s which are just `void*`)

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -246,32 +270,31 @@ - (void)dealloc {
...
-                [self willChangeValueForKey:@"operationCount"];
-                [_operations removeObjectAtIndex:index];
-                [self didChangeValueForKey:@"operationCount"];
-                [self didChangeValueForKey:@"operations"];
+    NSOperation* op = static_cast<NSOperation*>(object);
+    if ((context == _NSOperationQueue_IsFinishedContext) && (op.finished)) {
+        // Minor optimization:
+        // Because of constraints in addOperation:, it is safe to assume that there is exactly one instance of op in the array
+        // removeObject: keeps going after finding first instance of object, which is unnecessary work
+        // Use indexOfObject:op instead
```
DHowett-MSFT: nit: comment out of date

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -281,64 +304,70 @@ - (void)observeValueForKeyPath:(NSString*)keyPath ofObject:(id)object change:(NS
...
+        }
 
-    [self didChangeValueForKey:@"operationCount"];
-    [self didChangeValueForKey:@"operations"];
+        [self didChangeValueForKey:@"operationCount"];
+        [self didChangeValueForKey:@"operations"];
 
-    [op addObserver:self forKeyPath:@"isFinished" options:0 context:nil];
-    [op addObserver:self forKeyPath:@"isReady" options:0 context:nil];
+        [op addObserver:self forKeyPath:@"isFinished" options:NSKeyValueObservingOptionNew context:_NSOperationQueue_IsFinishedContext];
```
DHowett-MSFT: since you didn't in the end decide to use the change dictionary, you'll save a few cycles killing the **new** option here.

tests/unittests/Foundation/NSOperationTests.mm
```diff
@@ -88,14 +88,15 @@ - (instancetype)initWithBlock:(void (^)())block;
 @implementation BlockThread
 - (instancetype)initWithBlock:(void (^)())block {
     if (self = [super init]) {
-        self->_block = block;
-        self->_finishCondition = [[_NSBooleanCondition new] autorelease];
+        self.block = block;
```
DHowett-MSFT: huh, is block observable? 🌟 never knew!

DHowett-MSFT: though, wait; we generally avoid using setters in init, because they can be overridden by subclasses who are not yet initialized.

DHowett-MSFT: even though that doesn't apply here

DHowett-MSFT:
strangely, it marked some of my new comments as outdated.

Frameworks/Foundation/NSOperationQueue.mm
ms-jihua: oh yeah i remember that...

Frameworks/Foundation/NSOperationQueue.mm
ms-jihua: oops oops
i should version control a little more aggressively so i avoid these sorts of mistakes

tests/unittests/Foundation/NSOperationTests.mm
ms-jihua: er, not sure what you're getting excited about here, actually. if a block were observable it'd be on the block member changing, not the block finishing.

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -1,443 +1,415 @@
...
+    } else if (pri1 < pri2) {
+        return NSOrderedDescending;
+    }
+    return NSOrderedSame;
+}
+
+// STL compare for NSOperation*, sorting by queuePriority, putting the greatest priority first
+struct _NSOperation_ComparePrioritySTL {
+    bool operator()(NSOperation* lhs, NSOperation* rhs) {
+        return lhs.queuePriority > rhs.queuePriority;
```
aballway: This would put the lowest priority on top of the pqueue

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -1,443 +1,415 @@
...
+
+            ++_currentConcurrentOperationCount;
+        } // _concurrentOperationCountLock scope
+
+        std::lock_guard<std::recursive_mutex> lock(_dispatchQueueLock);
+        dispatch_async(_dispatchQueue, ^{
+            // Check if currentQueue needs to be changed
+            StrongId<NSOperationQueue*>& currentQueue = _getCurrentQueue();
+            StrongId<NSOperationQueue*> prevCurrentQueue(currentQueue.get());
+            bool needToChangeCurrentQueue = (![[NSThread currentThread] isMainThread]) && (!(self == currentQueue.get()));
```
aballway: nit: prefer !(A || B) to (!A && !B)

tests/unittests/Foundation/TestUtils.mm
```diff
@@ -52,4 +52,4 @@ void deleteFile(NSString* name) {
     if ([[NSFileManager defaultManager] fileExistsAtPath:fullPath]) {
         [[NSFileManager defaultManager] removeItemAtPath:fullPath error:nil];
     }
-};
+};
```
aballway: nit: revert this file

Frameworks/Foundation/NSOperationQueue.mm
ms-jihua: oops, i didn't realize priority queue had opposite sorting from objective-c ;;;;

Frameworks/Foundation/NSOperation.mm
```diff
@@ -349,4 +355,37 @@ - (void)dealloc {
...
+    } else if (queuePriority > NSOperationQueuePriorityVeryLow) {
+        _queuePriority = NSOperationQueuePriorityLow;
+
+    } else {
+        _queuePriority = NSOperationQueuePriorityVeryLow;
+    }
+}
+
+- (BOOL)_markInQueue {
+    return !_inQueue.fetch_or(YES); // only returns true for the first caller
```
DHowett-MSFT: nit: since it's truly a `bool`, you'll want `true`; though we're using an `atomic_bool` to back a `BOOL` property o_O

ms-jihua: it's an atomic<BOOL>

ms-jihua:
Ping @rajsesh-msft - I want to make sure you have no additional issues with this before it merges.

PR #2244, opened 2017-03-16
Colorspace support for CGColor and improvements to the class.

### LINKED ISSUE #2041: Implement some colorspace apis (with caveats)
CGColorGetColorSpace
CGContextSetStrokeColorSpace
CGContextSetFillColorSpace

These can be trivially implemented to support RGB (and grayscale to the extent we support it).

msft-Jeyaram:
The CGColor is still a UIColor class [re-implementation is planned], but we have improved the functionality.

fixes #2041

### DIFF
--- a/Frameworks/CoreGraphics/CGColor.mm
+++ b/Frameworks/CoreGraphics/CGColor.mm
@@ -1,6 +1,6 @@
 //******************************************************************************
 //
-// Copyright (c) 2016 Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
 //
 // This code is licensed under the MIT License (MIT).
 //
@@ -31,25 +31,34 @@ static IWLazyClassLookup __LazyUIColor("UIColor");
  @Notes Limited constants supported
 */
 CGColorRef CGColorGetConstantColor(CFStringRef name) {
+    CGColorRef ret = nullptr;
+    woc::StrongCF<CGColorSpaceRef> colorspace;
     if (CFEqual(kCGColorBlack, name)) {
-        return [[__LazyUIColor blackColor] CGColor];
+        ret = static_cast<CGColorRef>([[__LazyUIColor blackColor] CGColor]);
+        colorspace = woc::MakeStrongCF<CGColorSpaceRef>(CGColorSpaceCreateDeviceGray());
     } else if (CFEqual(kCGColorWhite, name)) {
-        return [[__LazyUIColor whiteColor] CGColor];
+        ret = static_cast<CGColorRef>([[__LazyUIColor whiteColor] CGColor]);
+        colorspace = woc::MakeStrongCF<CGColorSpaceRef>(CGColorSpaceCreateDeviceGray());
     } else if (CFEqual(kCGColorClear, name)) {
-        return [[__LazyUIColor clearColor] CGColor];
+        ret = static_cast<CGColorRef>([[__LazyUIColor clearColor] CGColor]);
+        colorspace = woc::MakeStrongCF<CGColorSpaceRef>(CGColorSpaceCreateDeviceGray());
+    } else {
+        UNIMPLEMENTED_WITH_MSG("CGColorGetConstantColor does not support color %s", CFStringGetCStringPtr(name, kCFStringEncodingUTF8));
     }
 
-    UNIMPLEMENTED_WITH_MSG("CGColorGetConstantColor does not support color %s", CFStringGetCStringPtr(name, kCFStringEncodingUTF8));
-    return nullptr;
+    if (ret) {
+        [static_cast<UIColor*>(ret) setColorSpace:colorspace];
+    }
+
+    return ret;
 }
 
 /**
  @Status Interoperable
 */
 void CGColorRelease(CGColorRef color) {
-    if (color != nullptr) {
-        CFRelease(color);
-    }
+    RETURN_IF(!color);
+    CFRelease(color);
 }
 
 /**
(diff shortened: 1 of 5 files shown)


Frameworks/UIKit/UIColor.mm~RF5a7c3ee.TMP
```diff
@@ -0,0 +1,850 @@
+//******************************************************************************
```
aballway: .TMP 😆

Frameworks/UIKit/UIColor.mm
```diff
@@ -226,6 +226,8 @@ @implementation UIColor {
     UIImage* _image;
     id _pattern;
     __CGColorQuad _components;
+    // Note: This should be part of the CGColor struct
+    woc::StrongCF<CGColorSpaceRef> _colorSpace;
```
aballway: I really don't think this is the way to fix this.  This change set should be included in the separation of CGColor from UIColor, rather than piling on additional short term fixes in UIColor which will add more things to remove.

msft-Jeyaram: Talked to raj regarding this, ideally, the whole thing should be fixed!
but given the load this iteration and the re-write is coming up, this should be sufficient. 
as much as I don't like doing this, until the re-write happens this would fix up some stuff for us.

Frameworks/CoreGraphics/CGColor.mm
```diff
@@ -31,25 +31,31 @@
...
+        ret = static_cast<CGColorRef>([[__LazyUIColor clearColor] CGColor]);
+    } else {
+        UNIMPLEMENTED_WITH_MSG("CGColorGetConstantColor does not support color %s", CFStringGetCStringPtr(name, kCFStringEncodingUTF8));
     }
 
-    UNIMPLEMENTED_WITH_MSG("CGColorGetConstantColor does not support color %s", CFStringGetCStringPtr(name, kCFStringEncodingUTF8));
-    return nullptr;
+    if (ret) {
+        auto grayColorSpace = woc::MakeStrongCF<CGColorSpaceRef>(CGColorSpaceCreateDeviceGray());
+        [static_cast<UIColor*>(ret) setColorSpace:grayColorSpace];
```
DHowett-MSFT: This is only the appropriate fix for black/white/clear!

Frameworks/CoreGraphics/CGColor.mm
```diff
@@ -62,18 +68,20 @@ CGColorRef CGColorRetain(CGColorRef color) {
...
+ @Status Caveat
+ @Notes only supports RGB
 */
 CGColorRef CGColorCreate(CGColorSpaceRef colorSpace, const CGFloat* components) {
-    CGColorRef ret =
-        (CGColorRef)[[__LazyUIColor colorWithRed:components[0] green:components[1] blue:components[2] alpha:components[3]] retain];
-
+    RETURN_NULL_IF(!colorSpace);
+    RETURN_NULL_IF(!components);
+    CGColorRef ret = static_cast<CGColorRef>(
```
DHowett-MSFT: it might be safe to assert that the #components in the color space is 3-4. Just so that we don't index beyond the end of components.

Frameworks/CoreGraphics/CGColor.mm
```diff
@@ -149,10 +161,10 @@ CGColorSpaceRef CGColorGetColorSpace(CGColorRef color) {
 
 /**
  @Status Caveat
- @Notes Not all colors have 4 components, but all of the ones we currently support do!
+ @Notes Limited support for colorspace
 */
 size_t CGColorGetNumberOfComponents(CGColorRef color) {
-    return 4;
+    return CGColorSpaceGetNumberOfComponents(CGColorGetColorSpace(color)) + 1;
```
DHowett-MSFT: this +1 is somewhat worrying

Frameworks/include/UIColorInternal.h
```diff
@@ -37,7 +37,7 @@ inline void _ClearColorQuad(__CGColorQuad& color) {
 }
 + (UIColor*)_colorWithCGPattern:(CGPatternRef)pattern;
 + (UIColor*)_windowsTableViewCellSelectionBackgroundColor;
-
 - (const __CGColorQuad*)_getColors;
 - (BrushType)_type;
+@property CGColorSpaceRef colorSpace;
```
DHowett-MSFT: whoa boy, give this a lifetime and an atomicity specifier!

tests/unittests/CoreGraphics/CGColorTests.mm
```diff
@@ -15,21 +15,30 @@
...
+// TODO #2243: Remove the UIKit dependency
+
 #define EXPECT_EQ_COMPONENTS(a, b) \
     EXPECT_EQ((a)[0], (b)[0]);     \
     EXPECT_EQ((a)[1], (b)[1]);     \
     EXPECT_EQ((a)[2], (b)[2]);     \
     EXPECT_EQ((a)[3], (b)[3])
 
-DISABLED_TEST(CGColor, CGColorGetComponents) {
+TEST(CGColor, CGColorGetComponents) {
```
DHowett-MSFT: it doesn't look like this test fixture anchors the UIKit symbols.

Frameworks/CoreGraphics/CGColor.mm
DHowett-MSFT: specifically: not all colours have alphas

Frameworks/CoreGraphics/CGColor.mm
msft-Jeyaram: the Caveat indicates it's limited. Should probably move it up so the 'future' is not clouded.

Frameworks/CoreGraphics/CGColor.mm
msft-Jeyaram: hmm i don't want to do this, as it's only supporting RGB, not even grayscale.
This is not the main purpose of the CR. The main purpose is to move it to Add CGColorGetColorSpace only.
These will be implemented properly in the next iteration.

---
In reply to: [106506478](https://github.com/Microsoft/WinObjC/pull/2244#discussion_r106506478) [](ancestors = 106506478)

Frameworks/CoreGraphics/CGColor.mm
msft-Jeyaram: this is based off of the limited support. Again a while fix to the colorspace is needed. 

---
In reply to: [106507058](https://github.com/Microsoft/WinObjC/pull/2244#discussion_r106507058) [](ancestors = 106507058)

msft-Jeyaram:
@DHowett  @aballway ping!

PR #2251, opened 2017-03-16
Make UTs build on reference platform

### LINKED ISSUE #2252: Foundation.UnitTests should build on reference platform

aballway:
With files required for Foundation.UnitTests being moved the unit tests no longer build on the reference platform.  This resolves the issue by removing the dependency on the moved files as they should not be used by unit tests on the reference platform.

Fixes #2252

### DIFF
--- a/tests/unittests/Foundation/ReferenceFoundation/TestNSBundle.mm
+++ b/tests/unittests/Foundation/ReferenceFoundation/TestNSBundle.mm
@@ -25,7 +25,6 @@
 #import <Foundation/Foundation.h>
 #import <Starboard/SmartTypes.h>
 #import <TestFramework.h>
-#import <LoggingNative.h>
 
 TEST(NSBundle, Paths) {
     auto bundle = [NSBundle mainBundle];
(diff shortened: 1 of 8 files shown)


DHowett-MSFT:
We introduced this dependency on purpose. Let's chat; I don't think these tests should be disabled.

bbowman:
I agree we shouldn't comment our tests but there isn't a need for Iw anything on the reference platform by definition. Looks like a std unique ptr would fit the bill fine here. It's up to the reference platform to choose the right heap. As "app code" in these tests we shouldn't need to be extra fancy / helpful

DHowett-MSFT:
@bbowman generally, I agree. IwMalloc and friends, however, are an extension API that apps can consume.

bbowman:
On our platform yes(for now). On the ref platform it is meaningless since there is only one controllable side of the equation. The entire purpose of IwMalloc and friends is actually so you don't need the extension. It's to match the app developers "normal" heap so that blind memory alloc / dealloc between components doesn't explode with wrath of 1000 suns.

bbowman:
:shipit:

tests/unittests/Foundation/ReferenceFoundation/TestNSBundle.mm
```diff
@@ -116,18 +115,14 @@ virtual void SetUp() {
 
         NSError* error = nil;
         [[NSFileManager defaultManager] createDirectoryAtPath:tempDir withIntermediateDirectories:false attributes:nil error:&error];
-        if (error) {
-            TraceError(L"TestNSBundle", L"%hs", [[error description] UTF8String]);
-            return;
-        }
+        // WINOBJC: Change to ASSERT macro to remove LoggingNative reference
```
DHowett-MSFT: fwiw i don't think this needs commenting, as the reference tests didn't have `TraceError` to remove to begin with

tests/unittests/Foundation/ReferenceFoundation/TestNSURL.mm
```diff
@@ -71,52 +70,55 @@
     NSError* error = nil;
     id testRoot =
         [NSPropertyListSerialization propertyListWithData:data options:static_cast<NSPropertyListReadOptions>(0) format:nil error:&error];
+
+    // WINOBJC: Change to macros to remove LoggingNative
```
DHowett-MSFT: as above

tests/unittests/UT.common.mk
```diff
@@ -28,12 +28,12 @@ OSX_VERSION_MIN := 10.10
 _UT_OBJECTS := $(addsuffix .o,$(addprefix $(_UT_OBJ_DIR)/,$(subst $(_UT_PROJECT_DIR)/,,$(UT_FILES))))
 
 # Common shims for all tests.
-_UT_OBJECTS += $(_UT_OBJ_DIR)/windows.o $(_UT_OBJ_DIR)/IwMalloc.o $(_UT_OBJ_DIR)/LoggingNative.o
+_UT_OBJECTS += $(_UT_OBJ_DIR)/windows.o
```
DHowett-MSFT: please make sure the drawing tests still compile.

DHowett-MSFT:
(specifically on OS X)

tests/unittests/UT.common.mk
aballway: They still compile 😁

Issue #2206, opened 2017-03-10
[Packaging] Upgrade ToolsVersion and MinimumVisualStudioVersion to 15.0


memontic-ms:
PackageReference requires Visual Studio 2017 and its toolset.

Most/some of tools.sln has been upgraded already. All of build.sln needs upgrading.

weswmsft:
Work completed

PR #2255, opened 2017-03-16
Upgrade build.sln projects to VS 2017 and toolset v141

memontic-ms:
Hopefully this will avoid any trouble with VS 2017 installs finding the v140 toolset.


bviglietta:
:shipit:

bbowman:
:shipit:

memontic-ms:
Okay, *now* all of build.sln should be using platform toolset v141.

memontic-ms:
This PR partially resolves #2206

PR #2256, opened 2017-03-16
Implement NSArray descriptionWithLocale* methods

### LINKED ISSUE #2073: Implement NSArray descriptionWithLocale* apis

aballway:
Fixes #2073

### DIFF
--- a/Frameworks/Foundation/NSArray.mm
+++ b/Frameworks/Foundation/NSArray.mm
@@ -749,31 +749,7 @@ static CFComparisonResult _CFComparatorFunctionFromComparator(const void* val1, 
  @Status Interoperable
 */
 - (NSString*)description {
-    thread_local unsigned int indent = 0;
-    NSMutableString* s = [NSMutableString string];
-    for (unsigned int i = 0; i < indent; ++i) {
-        [s appendString:@"    "];
-    }
-    [s appendString:@"(\n"];
-    {
-        ++indent;
-        auto deferPop = wil::ScopeExit([]() { --indent; });
-        for (id val in self) {
-            for (unsigned int i = 0; i < indent; ++i) {
-                [s appendString:@"    "];
-            }
-            [s appendFormat:@"%@,\n", val];
-        }
-
-        if ([self count] > 0) {
-            [s deleteCharactersInRange:{[s length] - 2, 1 }];
-        }
-    }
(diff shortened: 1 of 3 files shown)


Frameworks/Foundation/NSArray.mm
```diff
@@ -947,21 +923,69 @@ + (NSArray*)arrayWithContentsOfURL:(NSURL*)aURL {
...
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 - (NSString*)descriptionWithLocale:(id)locale indent:(NSUInteger)level {
-    UNIMPLEMENTED();
-    return StubReturn();
+    thread_local unsigned int indent = 0;
+    NSMutableString* s = [NSMutableString string];
+    NSString* indentStr = (level == 0) ? @"    " : @"\t";
```
DHowett-MSFT: I'm torn on how this is supposed to work. Shouldn't it call the next one with the next `level`, and always make sure to indent itself to _this_ `level`? The thread local becomes superfluous, and it's initialized with the wrong value anyway.

aballway: Nope, level is different from our thread_local indent, it's a parameter to the method.  0 means use 4 spaces, 1 mean use a tab.  https://developer.apple.com/reference/foundation/nsarray/1416257-descriptionwithlocale

Frameworks/Foundation/NSArray.mm
```diff
@@ -947,21 +923,68 @@ + (NSArray*)arrayWithContentsOfURL:(NSURL*)aURL {
...
+        [s appendString:indentStr];
+    }
+
+    [s appendString:@"(\n"];
+
+    {
+        ++level;
+        auto deferPop = wil::ScopeExit([&level]() { --level; });
+        for (id val in self) {
+            for (unsigned int i = 0; i < level; ++i) {
```
DHowett-MSFT: I'm still not sure this is totally correct; if you nest arrays, you'll double-indent them because you're appending `level*"spaces"`, and then passing level to the nested value which will prepend `level*"spaces"`

aballway: This should be expected behavior, our test for NSArray description depth still passes w/ this change

DHowett-MSFT: :(

i guess we should file a task to do this for dictionary too.

PR #2258, opened 2017-03-16
Fix all the clang windows that sometimes pop up all over the place

bbowman:


msft-Jeyaram:
😱

pradipd:
The fix looks good.
Is this a new issue?  I don't recall ever hearing about this?

pradipd:
:shipit:

pradipd:
If this is a recent regression, my concern is that .Net changed their behavior and perhaps they could like to know about it.

---
In reply to: [287396506](https://github.com/Microsoft/WinObjC/pull/2258#issuecomment-287396506) [](ancestors = 287396506)

bbowman:
It's a recent regression due to packaging or vs2017. Our best guess right now is that there was a change in how msbuild is being invoked and before we were just inheriting the parent process handle so it didn't matter but now that parent process can spawn UI in some circumstances

 @rainersigwald may know more on recent msbuild changes? 

Seems like a long standing bug that we got lucky with either way.

rainersigwald:
VS2017 moved some MSBuild evaluations to a separate process (a worker `MSBuild.exe`, instead of the main VS `devenv.exe`). I suspect that you're right and `devenv.exe` was squelching the windows but MSBuild itself isn't. I don't think anything in MSBuild itself would have changed this recently.

PR #2259, opened 2017-03-17
Try to reduce commandline restore errors by adding <DeployExtension>F…

bbowman:
…alse</DeployExtension>


tools/BuildMonitor/BuildMonitor.csproj
```diff
@@ -51,7 +52,6 @@
     <DefineConstants>DEBUG;TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
-    <DeployExtension>False</DeployExtension>
```
memontic-ms: Debug builds for BuildMonitor have already failed during restore. I'm not so sure this will fix this issue.

PR #2263, opened 2017-03-17
Support for Pattern transformation.

### LINKED ISSUE #2108: CGPattern CTM rotation should not affect the brush
If a rotation is applied to the context, it should not affect the patterns being drawn.
Patterns should always be of orientation 0.

msft-Jeyaram:
fixes #2108

### DIFF
--- a/Frameworks/CoreGraphics/CGContext.mm
+++ b/Frameworks/CoreGraphics/CGContext.mm
@@ -82,7 +82,7 @@ struct __CGContextDrawingState {
 
     // Strokes
     ComPtr<ID2D1Brush> strokeBrush{ nullptr };
-    D2D1_STROKE_STYLE_PROPERTIES strokeProperties{
+    D2D1_STROKE_STYLE_PROPERTIES1 strokeProperties{
         D2D1_CAP_STYLE_FLAT,
         D2D1_CAP_STYLE_FLAT,
         D2D1_CAP_STYLE_FLAT,
@@ -90,12 +90,13 @@ struct __CGContextDrawingState {
         10.f, // Default from Reference Docs
         D2D1_DASH_STYLE_SOLID,
         0.f,
+        D2D1_STROKE_TRANSFORM_TYPE_FIXED,
     };
     std::vector<CGFloat> dashes{};
     CGFloat lineWidth = 1.0f;
 
     // Computed from the above at draw time
-    ComPtr<ID2D1StrokeStyle> strokeStyle{ nullptr };
+    ComPtr<ID2D1StrokeStyle1> strokeStyle{ nullptr };
 
     CGFloat flatness = 0.0f;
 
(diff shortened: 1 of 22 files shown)


Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -2537,6 +2533,18 @@ void CGContextClearRect(CGContextRef context, CGRect rect) {
...
+        // pre transform the geometry
+        ComPtr<ID2D1TransformedGeometry> transformedGeometry;
+        FAIL_FAST_IF_FAILED(
+            Factory()->CreateTransformedGeometry(geometry.Get(),
+                                                 __CGAffineTransformToD2D_F(CGContextGetUserSpaceToDeviceSpaceTransform(this)),
+                                                 &transformedGeometry));
+
+        // Draw it in device space, so transformation is not applied to the render target.
+        coordinateMode = _kCGCoordinateModeDeviceSpace;
+        geometry.Attach(transformedGeometry.Detach());
```
msft-Jeyaram: note: As per D2D DrawGeo.. a transformed geometry will have a thinner line width, than an untransformed geometry + transformation.

Working on a fix in the coming iteration.

DHowett-MSFT:
Do we have any tests for _stroking_ with a pattern brush??

msft-Jeyaram:
@DHowett-MSFT  TEST_P(CGPatternColoredRectBasedStroke, PatternStrokeRegion) {
Cover the stroke

tests/UnitTests/CoreGraphics.drawing/CGContextDrawing_PatternTests.cpp
```diff
@@ -365,21 +365,113 @@ DISABLED_DRAW_TEST_F(CGPatternTests, PatternDrawPath, UIKitMimicTest<>) {
...
+DRAW_TEST_F(CGPatternTests, PatternFillTransformationWindowsLogoRotate, UIKitMimicTest<>) {
+    CGContextRef context = GetDrawingContext();
+    CGRect bounds = GetDrawingBounds();
+    CGContextRotateCTM(context, 0.4);
+    _SetPatternForFill(context, CGRectMake(0, 0, 100, 100), 100, 100, drawCoolStar, 1);
+
+    CGContextFillRect(context, bounds);
+}
+
+DRAW_TEST_F(CGPatternTests, PatternFillTransformationWindowsLogoTranslate, UIKitMimicTest<>) {
```
DHowett-MSFT: these can be parameterized on CGAffineTransform

DHowett-MSFT: also, we don't know what happens to stroke pattern brushes under transformation

Frameworks/CoreGraphics/CGContext.mm
```diff
@@ -144,15 +143,23 @@ inline void ComputeStrokeStyle(ID2D1DeviceContext* deviceContext) {
         ComPtr<ID2D1Factory> factory;
         deviceContext->GetFactory(&factory);
 
+        ComPtr<ID2D1Factory1> factory1;
```
rajsesh-msft: >factory1 [](start = 30, length = 8)

factory.As(&factory1)? also no need for fail fasts here, we will crash right afterwards anyway.

PR #2264, opened 2017-03-17
Update version overrides to take lastest pre-release where possible.

bbowman:


pradipd:
Sorry, it's not 100% clear to me what this change is for.
Is this so that we give precedence to the packages we locally built ?

pradipd:
Where did the gitverstion task come from?  How do we update it (if necessary)?

bbowman:
@pradipd 
Before this change, the precedence was:
1. Use Locally built packages if possible. 
2. Otherwise use "*" Version (latest stable package)

After this change it goes:
1. Use Locally built packages if possible. 
2. Otherwise Use Latest pre-release package if building pre-release (determined by gitversion)
3. Otherwise use "*" Version (latest stable package)

This would allow us to get the latest daily builds when building on develop. (Answers the question you asked in person this morning about trying out the fix for the clang windows)

In terms of where gitversion came from, this was the nuget package installed on my machine for version 4.0.0-beta0011, as shown in the path. Its a pre-restored package. To update we'd need to check in new binaries here. The other option is to have init.ps1 or other script get the latest version before the build but this is error prone if people forget, clean their repo etc.

common/winobjc.packagereference.override.targets
```diff
@@ -67,9 +78,24 @@
...
 
-  <Target Name="FixupWinObjCProjectReferences" BeforeTargets="_GetRestoreProjectStyle">
-      <FixupVersionOfWinObjCProjectReferences PackageReferences="@(PackageReference)"
-         SearchPaths="@(WinObjCPackageOverridePath)">
+  <PropertyGroup>
+    <FixupWinObjCProjectReferencesDependsOn>_GetRestoreProjectStyle</FixupWinObjCProjectReferencesDependsOn>
+    <FixupWinObjCProjectReferencesDependsOn Condition="'$(IsNuGetized)' != ''">GetPackageContents;%(FixupWinObjCProjectReferencesDependsOn)</FixupWinObjCProjectReferencesDependsOn>
+  </PropertyGroup>
+  
+  <Target Name="FixupWinObjCProjectReferences" BeforeTargets="$(FixupWinObjCProjectReferencesDependsOn)" DependsOnTargets="GetVersion">
```
memontic-ms: >FixupWinObjCProjectReferences [](start = 16, length = 29)

Should this be FixupWinObjCPackageReferences instead?

common/winobjc.packagereference.override.targets
```diff
@@ -67,9 +78,24 @@
...
+
+    <!-- Only set a prerelease version to use if its a pre-release build itself. Stable builds should depend on stable packages. -->
```
memontic-ms: nit

common/winobjc.packagereference.override.targets
bbowman: lol. yes

memontic-ms:
> CodeFlow for GitHub does not download executable files. If you trust this file and would like to run it, please download it yourself through GitHub.

Don't think we need this file

---
Refers to: common/gitversiontask/4.0.0-beta0011/build/lib/linux/x86_64/libgit2-baa87df.so:1 in 680a7de. [](commit_id = 680a7de8d42100dbb651f1b7e3a48ea3d88cfa80, deletion_comment = False)

bbowman:
> CodeFlow for GitHub does not download executable files. If you trust this file and would like to run it, please download it yourself through GitHub.

maybe I just did the whole nuget so #won't fix.

---
In reply to: [287474738](https://github.com/Microsoft/WinObjC/pull/2264#issuecomment-287474738) [](ancestors = 287474738)

---
Refers to: common/gitversiontask/4.0.0-beta0011/build/lib/linux/x86_64/libgit2-baa87df.so:1 in 680a7de. [](commit_id = 680a7de8d42100dbb651f1b7e3a48ea3d88cfa80, deletion_comment = False)

pradipd:
Awesome!.  Thanks for the explanation.

---
In reply to: [287472748](https://github.com/Microsoft/WinObjC/pull/2264#issuecomment-287472748) [](ancestors = 287472748)

pradipd:
:shipit:

common/winobjc.packagereference.override.targets
```diff
@@ -67,9 +78,24 @@
     </Task>
   </UsingTask>
 
-  <Target Name="FixupWinObjCProjectReferences" BeforeTargets="_GetRestoreProjectStyle">
-      <FixupVersionOfWinObjCProjectReferences PackageReferences="@(PackageReference)"
-         SearchPaths="@(WinObjCPackageOverridePath)">
+  <PropertyGroup>
+    <FixupWinObjCProjectReferencesDependsOn>_GetRestoreProjectStyle</FixupWinObjCProjectReferencesDependsOn>
+    <FixupWinObjCProjectReferencesDependsOn Condition="'$(IsNuGetized)' != ''">GetPackageContents;%(FixupWinObjCProjectReferencesDependsOn)</FixupWinObjCProjectReferencesDependsOn>
```
memontic-ms: >GetPackageContents; [](start = 79, length = 19)

I'm a bit confused why it cares about GetPackageContents

bbowman: yeah that is a little weird
but its so that if we do a dynamic version replacement
we correctly describe what version was used when making a package
this is seen in Frameworks -> Language
right now Frameworks .nuspec says it dpends on "*" for Language even if it was built with a specific version

memontic-ms:
>     TaskName="FixupVersionOfWinObjCProjectReferences"

I know these comments are basically inline, but could you add the summary of what version it'll assign at somewhere at the top here? Just what you had in another PR comment would be useful for when we look back at this in the future:
1. Use Locally built packages if possible.
2. Otherwise Use Latest pre-release package if building pre-release (determined by gitversion)
3. Otherwise use "*" Version (latest stable package)

---
Refers to: common/winobjc.packagereference.override.targets:9 in 680a7de. [](commit_id = 680a7de8d42100dbb651f1b7e3a48ea3d88cfa80, deletion_comment = False)

bbowman:
>     TaskName="FixupVersionOfWinObjCProjectReferences"

sure.

---
In reply to: [287477586](https://github.com/Microsoft/WinObjC/pull/2264#issuecomment-287477586) [](ancestors = 287477586)

---
Refers to: common/winobjc.packagereference.override.targets:9 in 680a7de. [](commit_id = 680a7de8d42100dbb651f1b7e3a48ea3d88cfa80, deletion_comment = False)

memontic-ms:
:shipit:

memontic-ms:
Add check to not override package version if it has already been set to specific version.
As in, only override if version has a * in it

bbowman:
@DHowett-MSFT @memontic-ms updated for "*" and packages that are substrings.

common/winobjc.packagereference.override.targets
```diff
@@ -28,6 +28,7 @@
       <Reference Include="System.IO" />
       <Using Namespace="System" />
       <Using Namespace="System.IO" />
+      <Using Namespace=" System.Text.RegularExpressions" />
```
DHowett-MSFT: nit: space?

common/winobjc.packagereference.override.targets
```diff
@@ -36,9 +37,11 @@
               // Try to find a matching package in one of the search paths.
               string packageName = packageItem.GetMetadata("Identity");
               string packageNameSearchPattern =  packageName + "*.nupkg";
+              var searchRegex= new Regex(packageName + @"\.\d+\.\d+\.\d+(-.*)?");
```
DHowett-MSFT: nit: the space from above probably belongs before this `=`

PR #2265, opened 2017-03-17
Removing unnecessary layout callback code from UIKit.Xaml.Button

jaredhms:


Frameworks/UIKit/UIButton.mm
```diff
@@ -226,15 +226,6 @@ - (void)_initUIButton {
                                               e.handled = YES;
                                               [weakSelf _processPointerCaptureLostCallback:sender eventArgs:e];
                                           });
-
-    XamlControls::HookLayoutEvent(_xamlButton, ^(RTObject*, WUXIPointerRoutedEventArgs*) {
-        // Since we are using XAML Button behind the scene, the intrinsicContentSize calculation is done
-        // by XAML.
-        // The size of XAML elements(for eg Image) is calculated at runtime and then the
-        // intrinsicContentSize is invalidated.
```
yiyang-msft: so this conclusion is not valid any more?

jaredhms: Correct; this isn't needed.  We handle image loading on this side and trigger layout as needed.

---
In reply to: [106746566](https://github.com/Microsoft/WinObjC/pull/2265#discussion_r106746566) [](ancestors = 106746566)

yiyang-msft:
:shipit:

PR #2275, opened 2017-03-19
Move #pragma statements so that they do not interfere with API annota…

asimonov-msft:
…tion block.

### DIFF
--- a/Frameworks/UIKit/UIViewController.mm
+++ b/Frameworks/UIKit/UIViewController.mm
@@ -666,12 +666,12 @@ NSMutableDictionary* _pageMappings;
     }
 }
 
+#pragma clang diagnostic push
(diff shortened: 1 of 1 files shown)


DHowett-MSFT:
@bbowman I fully agree, but I don't think that is worth rejecting this hotfix for. :smile:

bbowman:
Can @asimonov-msft file an issue for this then?

jaredhms:
:shipit:

arjun-msft:
:shipit:

asimonov-msft:
@bbowman @DHowett-MSFT Unfortunately, this is an issue with how libclang associates the doxygen-style comment blocks with entries in the AST. As a result, this is not something that we can easily fix.

PR #2278, opened 2017-03-20
Adding 'MinApp' test apps for perf comparisons

jaredhms:


tests/testapps/MinAppCppCx/MinAppCppCx/MinAppCppCx.vcxproj
```diff
@@ -0,0 +1,213 @@
+<?xml version="1.0" encoding="utf-8"?>
```
bbowman: You'll want to use PackageReference style nuget deps with the override targets here if you want to use the latest / local packages.

tests/testapps/MinApp/MinApp.vsimporter/MinApp-WinStore10/project.json
```diff
@@ -0,0 +1,17 @@
+{
+  "dependencies": {
+    "WinObjC.Frameworks": "*",
```
bbowman: You'll want to move to PackageReference instead. Additionally, you'll want to add all packages (except tools and packaging) here to handle cases where you update a single pacakge that isn't a leaf node and want to test local updates.

bbowman: Look at WOCCatalog as an example.

jaredhms: Why didn't vsimporter 'do the right thing' here?

bbowman: @jaredhms it did for a normal app developer. This is not a normal app. The two leaf nodes with a "*" version is exactly what we want most external developers to use as it is the simplest way to have them stay up to date on the latest pre-release version.

We don't currently have a switch in VSImporter to do version swapping but it is something we know about and are working on. See https://github.com/Microsoft/WinObjC/issues/2199

jaredhms: ah, that's right - thanks for the clarification.

---
In reply to: [107011503](https://github.com/Microsoft/WinObjC/pull/2278#discussion_r107011503) [](ancestors = 107011503)

yiyang-msft:
:shipit:

PR #2279, opened 2017-03-20
Implement NSIndexPath

### LINKED ISSUE #2079: Finish NSIndexPath implementation
init 😮 
indexPathByRemovingLastIndex:
getIndexes:range:

aballway:
Changes NSIndexPath implementation to use std::vector rather than a raw pointer & size value, included some additional cleanup while I was in the file.

Fixes #2079


Frameworks/Foundation/NSIndexPath.mm
```diff
@@ -90,114 +104,90 @@ - (BOOL)isEqual:(id)other {
...
+    return std::accumulate(_indexes.cbegin(), _indexes.cend(), 0);
 }
 
 /**
  @Status Interoperable
 */
-- (int)compare:(id)otherObj {
-    int len1 = [self length];
-    int len2 = [otherObj length];
+- (NSComparisonResult)compare:(id)otherObj {
```
ms-jihua: nit: can we add a comment about how NSIndexPaths are sorted in depth-first traversal order? was surprised by this, would like to save people future trips to the docs

Frameworks/Foundation/NSIndexPath.mm
```diff
@@ -18,6 +18,17 @@
 #import "StubReturn.h"
 #import "Foundation/NSIndexPath.h"
 
+#import <vector>
+#import <numeric>
+
+static NSString* c_archiveKey = @"_NS.IP";
```
msft-Jeyaram: >static NSString* c [](start = 0, length = 18)

static const?

Frameworks/Foundation/NSIndexPath.mm
```diff
@@ -38,22 +49,28 @@ + (NSIndexPath*)indexPathWithIndexes:(const NSUInteger*)indexes length:(NSUInteg
  @Status Interoperable
 */
 + (NSIndexPath*)indexPathForRow:(NSInteger)row inSection:(NSInteger)section {
-    unsigned indexes[] = { section, row };
-
+    NSUInteger indexes[] = { section, row };
```
msft-Jeyaram: > NSUInteger indexes[] = { section, row }; [](start = 3, length = 41)

NSInteger to NSUnsignedInteger

Frameworks/Foundation/NSIndexPath.mm
```diff
@@ -38,22 +49,28 @@ + (NSIndexPath*)indexPathWithIndexes:(const NSUInteger*)indexes length:(NSUInteg
...
-    NSUInteger indexes[] = { section, row };
++ (NSIndexPath*)indexPathForItem:(NSInteger)item inSection:(NSInteger)section {
+    NSUInteger indexes[] = { section, item };
     return [[[self alloc] initWithIndexes:indexes length:2] autorelease];
 }
 
 /**
  @Status Interoperable
 */
+- (instancetype)init {
```
msft-Jeyaram: >init [](start = 16, length = 4)

which one is the default initializer? 
We should respect the order

Frameworks/Foundation/NSIndexPath.mm
```diff
@@ -62,16 +79,13 @@ - (instancetype)initWithIndex:(NSUInteger)index {
  @Status Interoperable
 */
 - (instancetype)initWithIndexes:(const NSUInteger*)indexes length:(NSUInteger)length {
-    _length = length;
-    _indexes = (unsigned*)IwMalloc(length * sizeof(unsigned));
-    if (_indexes == nil) {
-        [self release];
-        return nil;
+    if (self = [super init]) {
+        if (length != 0 && indexes != nullptr) {
```
msft-Jeyaram: >(length != 0 && indexes != nullptr)  [](start = 11, length = 36)

nit: if(length && indexes) ?

Frameworks/Foundation/NSIndexPath.mm
```diff
@@ -62,16 +79,13 @@ - (instancetype)initWithIndex:(NSUInteger)index {
...
+            _indexes.resize(length);
+            memcpy(_indexes.data(), indexes, length * sizeof(NSUInteger));
```
msft-Jeyaram: reserve & insert??

tests/UnitTests/Foundation/NSIndexPathTests.mm
```diff
@@ -0,0 +1,167 @@
...
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#include <TestFramework.h>
+#import <Foundation/Foundation.h>
+#import <Starboard/SmartTypes.h>
+
+TEST(NSIndexPath, Init) {
+    NSIndexPath* path = [[[NSIndexPath alloc] init] autorelease];
```
msft-Jeyaram: >[[[NSIndexPath alloc] init] autorelease]; [](start = 24, length = 41)

nite: new?

tests/UnitTests/Foundation/NSIndexPathTests.mm
```diff
@@ -0,0 +1,167 @@
...
+    ASSERT_NE(nil, path);
+    EXPECT_EQ(0, [path length]);
+    EXPECT_EQ(NSNotFound, [path indexAtPosition:0]);
+}
+
+TEST(NSIndexPath, InitWithIndex) {
+    NSIndexPath* path = [[[NSIndexPath alloc] initWithIndex:5] autorelease];
```
msft-Jeyaram: >th = [[[NSIndexPath alloc] initWithIndex:5] autorelease]; [](start = 19, length = 57)

indexPathWithIndex
This way you get to test both functions ;)

msft-Jeyaram: same at other places.

---
In reply to: [107549224](https://github.com/Microsoft/WinObjC/pull/2279#discussion_r107549224) [](ancestors = 107549224)

tests/UnitTests/Foundation/NSIndexPathTests.mm
```diff
@@ -0,0 +1,167 @@
...
+OSX_DISABLED_TEST(NSIndexPath, IndexPathForRowInSection) {
+    NSIndexPath* path = [NSIndexPath indexPathForRow:2 inSection:1];
+    ASSERT_NE(nil, path);
+    EXPECT_EQ(2, [path length]);
+    EXPECT_EQ(1, [path indexAtPosition:0]);
+    EXPECT_EQ(2, [path indexAtPosition:1]);
+    EXPECT_EQ(NSNotFound, [path indexAtPosition:2]);
+}
+
+OSX_DISABLED_TEST(NSIndexPath, IndexPathForItemInSection) {
```
msft-Jeyaram: hmm is this the same case?
If it's not supported (why?) add the comment here too.

tests/UnitTests/Foundation/NSIndexPathTests.mm
```diff
@@ -0,0 +1,167 @@
...
+    NSIndexPath* path = [NSIndexPath indexPathForItem:2 inSection:1];
+    ASSERT_NE(nil, path);
+    EXPECT_EQ(2, [path length]);
+    EXPECT_EQ(1, [path indexAtPosition:0]);
+    EXPECT_EQ(2, [path indexAtPosition:1]);
+    EXPECT_EQ(NSNotFound, [path indexAtPosition:2]);
+}
+
+TEST(NSIndexPath, IndexPathByAddingIndex) {
+    NSIndexPath* path = [[[NSIndexPath alloc] init] autorelease];
```
msft-Jeyaram: >[[[NSIndexPath alloc] init] autorelease]; [](start = 24, length = 41)

nit: new

Frameworks/Foundation/NSIndexPath.mm
```diff
@@ -90,114 +104,90 @@ - (BOOL)isEqual:(id)other {
...
 */
 - (NSUInteger)hash {
-    unsigned ret = 0;
-
-    for (unsigned int i = 0; i < _length; i++) {
-        ret += _indexes[i];
-    }
-
-    return ret;
+    // Allegedly the behavior of the reference platform, does not seem to be so for test machines but it will suffice
```
rajsesh-msft: >machines [](start = 89, length = 8)

not sure what this comment is implying :)

Frameworks/Foundation/NSIndexPath.mm
```diff
@@ -215,54 +205,51 @@ - (NSInteger)section {
...
 }
 
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 - (void)encodeWithCoder:(NSCoder*)coder {
-    UNIMPLEMENTED();
+    if ([self length] > 0L) {
```
rajsesh-msft: >[self length] > 0L [](start = 8, length = 18)

0 length indexpath is not serialized then?

Frameworks/Foundation/NSIndexPath.mm
rajsesh-msft: According to documentation, init is the designated initializer, which means all other initializers must call it and this should be the one calling super.  This helps with subclassing.

---
In reply to: [107547799](https://github.com/Microsoft/WinObjC/pull/2279#discussion_r107547799) [](ancestors = 107547799)

tests/UnitTests/Foundation/NSIndexPathTests.mm
```diff
@@ -0,0 +1,167 @@
...
+    EXPECT_EQ(3, [path length]);
+    EXPECT_EQ(1, [path indexAtPosition:0]);
+    EXPECT_EQ(2, [path indexAtPosition:1]);
+    EXPECT_EQ(3, [path indexAtPosition:2]);
+    EXPECT_EQ(NSNotFound, [path indexAtPosition:4]);
+}
+
+// These are disabled because the methods are not yet supported on the reference testing platform
+OSX_DISABLED_TEST(NSIndexPath, IndexPathForRowInSection) {
+    NSIndexPath* path = [NSIndexPath indexPathForRow:2 inSection:1];
```
aballway: Nope, it seems like macOS never got around to implementing it...

Frameworks/Foundation/NSIndexPath.mm
aballway: Other encodeWithEncoder implementations had no serialization done for empty objects and then when calling initWithEncoder returning the empty object when there's no data to create from, and I'd made the (possibly unsound) assumption that NSIndexPath followed suit.  Will check on reference platform and update should it be incorrect.

Frameworks/Foundation/NSIndexPath.mm
aballway: According to documentation initWithIndexes:length: is also the designated initializer 🤣 (https://developer.apple.com/reference/foundation/nsindexpath/1416906-initwithindexes?language=objc) The previous implementation used initWithIndexes:length: as default so I'm going to leave it as is

tests/UnitTests/Foundation/NSIndexPathTests.mm
aballway: 👍

Frameworks/Foundation/NSIndexPath.mm
aballway: I was told this implementation was written this way because of observed values, but the reference platform seems to have updated the hash implementation.  I'll change this to something more suitable as well.  It won't be equivalent to reference platform, but I don't think it needs to be

aballway: Actually this doesn't seem to be a documented method on the reference platform, so I don't think we need to support it.  @jaredhms do you know of any places in UIKit where this might be used? If we don't have any I'm just going to remove it

Frameworks/Foundation/NSIndexPath.mm
aballway: 😮 good catch

Frameworks/Foundation/NSIndexPath.mm
aballway: should be s_, but can't be const because encode/decode expect non-const

Frameworks/Foundation/NSIndexPath.mm
rajsesh-msft: Ugh, it is an NSObject protocol method, we do need to implement it :).  

---
In reply to: [107951981](https://github.com/Microsoft/WinObjC/pull/2279#discussion_r107951981) [](ancestors = 107951981)

Frameworks/Foundation/NSIndexPath.mm
```diff
@@ -90,114 +112,95 @@ - (BOOL)isEqual:(id)other {
...
-
-    NSUInteger* indexCopy = (NSUInteger*)IwMalloc((_length + 1) * sizeof(DWORD));
-    memcpy(indexCopy, _indexes, _length * sizeof(DWORD));
-    indexCopy[_length] = newIndex;
-
-    [ret initWithIndexes:indexCopy length:_length + 1];
-
-    IwFree(indexCopy);
+    std::vector<NSUInteger> indexes = _indexes;
+    indexes.emplace_back(newIndex);
```
jaredhms: the extra copy here is unfortunate; how about just alloc'ing an empty NSIndex path, then std::move(indexes) into the new object's _indexes vector?

aballway:
ping @rajsesh-msft @msft-Jeyaram

aballway:
ping @msft-Jeyaram

Frameworks/Foundation/NSIndexPath.mm
```diff
@@ -215,54 +223,45 @@ - (NSInteger)section {
...
 }
 
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 - (void)encodeWithCoder:(NSCoder*)coder {
-    UNIMPLEMENTED();
+    NSData* dataToEncode = [self length] > 0L ?
```
DHowett-MSFT: We can't do this; especially if we want peoples' preferences to eventually roam to ARM and other platforms whose NSIntegers are different from x86.

Frameworks/Foundation/NSIndexPath.mm
```diff
@@ -18,6 +18,17 @@
 #import "StubReturn.h"
 #import "Foundation/NSIndexPath.h"
 
+#import <vector>
+#import <numeric>
+
+static NSString* s_archiveKey = @"_NS.IP";
```
msft-Jeyaram: nit: const

msft-Jeyaram: >"_NS.IP" [](start = 33, length = 8)

nit: is this the standard format we are using?
I recall seeing a different one.

Frameworks/Foundation/NSIndexPath.mm
```diff
@@ -38,22 +49,36 @@ + (NSIndexPath*)indexPathWithIndexes:(const NSUInteger*)indexes length:(NSUInteg
  @Status Interoperable
 */
 + (NSIndexPath*)indexPathForRow:(NSInteger)row inSection:(NSInteger)section {
-    unsigned indexes[] = { section, row };
+    if (row < 0 || section < 0) {
+        return [[[self alloc] initWithIndexes:nullptr length:0] autorelease];
+    }
 
+    NSUInteger indexes[] = { section, row };
```
msft-Jeyaram: >{ section, row } [](start = 27, length = 16)

Still assigning an signed integer to an unsigned integer?

Frameworks/Foundation/NSIndexPath.mm
```diff
@@ -38,22 +49,36 @@ + (NSIndexPath*)indexPathWithIndexes:(const NSUInteger*)indexes length:(NSUInteg
...
     return [[[self alloc] initWithIndexes:indexes length:2] autorelease];
```
msft-Jeyaram: nit: std::extent<decltype(indexes)>::value

Frameworks/Foundation/NSIndexPath.mm
```diff
@@ -38,22 +49,36 @@ + (NSIndexPath*)indexPathWithIndexes:(const NSUInteger*)indexes length:(NSUInteg
...
 */
-+ (NSIndexPath*)indexPathForItem:(NSInteger)row inSection:(NSInteger)section {
-    NSUInteger indexes[] = { section, row };
++ (NSIndexPath*)indexPathForItem:(NSInteger)item inSection:(NSInteger)section {
+    if (item < 0 || section < 0) {
+        return [[[self alloc] initWithIndexes:nullptr length:0] autorelease];
+    }
+
+    NSUInteger indexes[] = { section, item };
     return [[[self alloc] initWithIndexes:indexes length:2] autorelease];
```
msft-Jeyaram: >2 [](start = 57, length = 1)

nit: std::extent<decltype(indexes)>::value

Frameworks/Foundation/NSIndexPath.mm
```diff
@@ -38,22 +49,36 @@ + (NSIndexPath*)indexPathWithIndexes:(const NSUInteger*)indexes length:(NSUInteg
  @Status Interoperable
 */
 + (NSIndexPath*)indexPathForRow:(NSInteger)row inSection:(NSInteger)section {
-    unsigned indexes[] = { section, row };
+    if (row < 0 || section < 0) {
```
msft-Jeyaram: >row < 0 || section < 0) { [](start = 8, length = 25)

is this valid? given the fact that they are passing in a NSInteger vs NSUInteger.
is this the expected behavior on ref platform?

Frameworks/Foundation/NSIndexPath.mm
```diff
@@ -62,16 +87,13 @@ - (instancetype)initWithIndex:(NSUInteger)index {
...
-    _length = length;
-    _indexes = (unsigned*)IwMalloc(length * sizeof(unsigned));
-    if (_indexes == nil) {
-        [self release];
-        return nil;
+    if (self = [super init]) {
+        if (length != 0 && indexes != nullptr) {
+            _indexes.reserve(length);
+            _indexes.insert(_indexes.begin(), indexes, indexes + length);
+        }
```
msft-Jeyaram: You have it such that, if you pass in an invalid index, you get a object rather nil.
I don't think that's what you want

Frameworks/Foundation/NSIndexPath.mm
```diff
@@ -81,7 +103,7 @@ - (instancetype)initWithIndexes:(const NSUInteger*)indexes length:(NSUInteger)le
 - (BOOL)isEqual:(id)other {
     if ([other isKindOfClass:[NSIndexPath class]]) {
         NSIndexPath* otherPath = other;
-        return _length == otherPath->_length && (memcmp(_indexes, otherPath->_indexes, _length * sizeof(DWORD)) == 0);
+        return _indexes == otherPath->_indexes;
     }
 
     return FALSE;
```
msft-Jeyaram: >FALSE [](start = 11, length = 5)

nit: NO as it's the proper usage for BOOL

Frameworks/Foundation/NSIndexPath.mm
```diff
@@ -90,114 +112,100 @@ - (BOOL)isEqual:(id)other {
...
-    NSUInteger* indexCopy = (NSUInteger*)IwMalloc((_length + 1) * sizeof(DWORD));
-    memcpy(indexCopy, _indexes, _length * sizeof(DWORD));
-    indexCopy[_length] = newIndex;
-
-    [ret initWithIndexes:indexCopy length:_length + 1];
+    NSIndexPath* ret = [NSIndexPath indexPathWithIndexes:nullptr length:0];
+    if (ret) {
+        ret->_indexes.reserve(_indexes.size() + 1);
+        ret->_indexes.insert(ret->_indexes.begin(), _indexes.cbegin(), _indexes.cend());
+        ret->_indexes.emplace_back(newIndex);
```
msft-Jeyaram: please don't do this.
move this to a init method, rather than individually manipulating in a '+' method

Frameworks/Foundation/NSIndexPath.mm
```diff
@@ -90,114 +112,100 @@ - (BOOL)isEqual:(id)other {
...
+    return CFHash(dataToHash);
 }
 
 /**
  @Status Interoperable
 */
-- (int)compare:(id)otherObj {
-    int len1 = [self length];
-    int len2 = [otherObj length];
+- (NSComparisonResult)compare:(id)otherObj {
```
msft-Jeyaram: nit: if they are the same object, we can short circuit earlier vs doing more method calls and validating.

tests/UnitTests/Foundation/NSIndexPathTests.mm
msft-Jeyaram: why? I don't see any issues using new here.

---
In reply to: [107549082](https://github.com/Microsoft/WinObjC/pull/2279#discussion_r107549082) [](ancestors = 107549082)

aballway: There's also no issue leaving it as alloc/init

Frameworks/Foundation/NSIndexPath.mm
aballway: I've seen differing archiveKeys, went with this one since I was borrowing from IndexSet

Frameworks/Foundation/NSIndexPath.mm
aballway: This is what I want, tested on the reference platform

Frameworks/Foundation/NSIndexPath.mm
aballway: The reference platform hasn't yet implemented this so I cannot say...
So long as row >= 0 && section >= 0 NSUInteger can hold it

Frameworks/Foundation/NSIndexPath.mm
aballway: Is there some way to do that without having a double-copy like before?

Frameworks/Foundation/NSIndexPath.mm
msft-Jeyaram: ?

tests/UnitTests/Foundation/NSIndexPathTests.mm
msft-Jeyaram: much more concise

---
In reply to: [109024879](https://github.com/Microsoft/WinObjC/pull/2279#discussion_r109024879) [](ancestors = 109024879,107549082)

Frameworks/Foundation/NSIndexPath.mm
```diff
@@ -38,22 +49,36 @@ + (NSIndexPath*)indexPathWithIndexes:(const NSUInteger*)indexes length:(NSUInteg
  @Status Interoperable
 */
 + (NSIndexPath*)indexPathForRow:(NSInteger)row inSection:(NSInteger)section {
```
DHowett-MSFT: `indexPathForRow:inSection:` is supposed to be a category method implemented in UIKit.

Frameworks/Foundation/NSIndexPath.mm
```diff
@@ -38,22 +49,36 @@ + (NSIndexPath*)indexPathWithIndexes:(const NSUInteger*)indexes length:(NSUInteg
...
+    NSUInteger indexes[] = { section, row };
     return [[[self alloc] initWithIndexes:indexes length:2] autorelease];
 }
 
 /**
  @Status Interoperable
 */
-+ (NSIndexPath*)indexPathForItem:(NSInteger)row inSection:(NSInteger)section {
-    NSUInteger indexes[] = { section, row };
++ (NSIndexPath*)indexPathForItem:(NSInteger)item inSection:(NSInteger)section {
```
DHowett-MSFT: As above

aballway:
ping @msft-Jeyaram

Frameworks/UIKit/NSIndexPath+UITableView.mm
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    NSUInteger indexes[] = { section, item };
+    return [[[self alloc] initWithIndexes:indexes length:2] autorelease];
+}
+
+/**
+ @Status Interoperable
+*/
+- (NSInteger)item {
```
MSFTFox: Are these actually correct? The documentation says to use the section property which doesn't seem to line up with these constant values.

DHowett-MSFT:
The documentation states that the "item" (and "row") properties identify the item or row INSIDE the section identified by "section".

Issue #2291, opened 2017-03-21
NSBundle loadNibNamed:options: emits a 'stub' (deprecated) warning even though we implement it in UIKit


jaredhms:
Not sure how best to address this.  We implement this method in NSBundle+UIKitAdditions (albeit with caveats), but we mark it as stubbed in NSBundle.h, which results in misleading build output.

PR #2296, opened 2017-03-22
UINib nibWithNibName:bundle: should look in the MainBundle if no bundle is specified

jaredhms:
This fixes a nib-loading issue in an app we're investigating.


yiyang-msft:
:clock1:

rajsesh-msft:
Please fix #2291 along with this.  It is just a removal of a macro and fits right with this work.

jaredhms:
Added fix for #2291.

Fixes #2291.

PR #2297, opened 2017-03-22
Support custom fonts in Xaml through _compatibleFamilyName

### LINKED ISSUE #2273: Custom fonts are not properly applied to Xaml-backed UIViews
When working with custom, app-specific fonts, UIFont's _compatibleFamilyName gives us a family name that Xaml doesn't understand.  We should enhance it to append a full path to the font file + family name if the font is backed by a custom font file (if not a custom font, the current implementation works fine).

For example; this would allow the font to work properly:

ms-appx:///thefontfile.ttf#The Font Family

We should also add tests for this to XamlCatalog, for each caller of _compatibleFamilyName (currently UILabel and UITextField).

aballway:
Our implementation of supporting user-supplied fonts could not propagate to XAML because we cannot modify the system's font collection and instead have a separate font collection where we store user-supplied fonts.  XAML can't access this separate font collection, and instead has to pull the font from the file directly.  We already had a private method _compatibleFamilyName which returned the family name required for XAML as CTFontGetFamilyName would return different strings for some fonts, so this has been modified to return the string which XAML will use to get the specified font from its file.

This will be in format `@"ms-appx///[Absolute Path To File].ttf#FamilyName"`.  This however requires that the family name requested from the font be used for that specific font, as families can have multiple files.  For example, WinObjC-Bold.ttf has family name WinObjC as does WinObjC.ttf, but getting _compatibleFamilyName from a regular WinObjC font will not return the correct path to construct a Bold font, so **one should be careful to create the font UIFont with required attributes before trying to convert to XAML**.

We will get the path to the file by saving the CGDataConsumer which contains the CFURL it is registered with in the IDWriteFontFile we create.  Then, when we go to get the _compatibleFamilyName we retrieve the CGDataConsumer from which we retrieve the CFURL, pull out the path, and return the correct value.

This can be manually confirmed in XamlCatalog through the Custom Text Editor, which now uses our WinObjC font (the character 'A' is represented by a solid black circle).

Fixes #2273

### DIFF
--- a/Frameworks/CoreGraphics/CGDataProvider.mm
+++ b/Frameworks/CoreGraphics/CGDataProvider.mm
@@ -59,6 +59,11 @@ public:
         return memcmp(GetData(), rhs.GetData(), GetSize()) == 0;
     }
 
+    // Subclasses containing CFURL should override this
+    virtual CFURLRef GetURL() const {
+        return nullptr;
+    }
+
     CFHashCode Hash() const {
         std::call_once(m_hashFlag, &__CGDataProviderInternal::SetHash, this);
         return m_hashValue;
@@ -250,6 +255,10 @@ public:
         std::call_once(m_dataFlag, &__CFURLDataProvider::GenerateData, this);
         return CFDataGetLength(m_data.get());
     }
+
+    CFURLRef GetURL() const override {
+        return m_url.get();
+    }
 };
 #pragma endregion // __CGDataProviderInternal
 
@@ -293,6 +302,10 @@ struct __CGDataProvider : CoreFoundation::CppBase<__CGDataProvider> {
 
     size_t GetSize() const {
         return m_internal->GetSize();
+    }
+
+    CFURLRef GetURL() const {
+        return m_internal->GetURL();
     }
 
     bool operator==(const __CGDataProvider& other) const {
@@ -433,6 +446,10 @@ const void* _CGDataProviderGetData(CGDataProviderRef provider) {
 
 size_t _CGDataProviderGetSize(CGDataProviderRef provider) {
     return provider ? provider->GetSize() : 0;
+}
+
+CFURLRef _CGDataProviderGetURL(CGDataProviderRef provider) {
+    return provider ? provider->GetURL() : nullptr;
 }
 
 #pragma endregion // Internal CGDataProvider Functions
--- a/Frameworks/CoreGraphics/DWriteWrapper.mm
+++ b/Frameworks/CoreGraphics/DWriteWrapper.mm
@@ -224,20 +224,35 @@ CFStringRef _DWriteGetCompatibleFamilyName(CFStringRef fontName, IDWriteFontFace
             return userFontInfo->familyName.get();
         }
 
-        UINT32 fileCount = 1;
-        ComPtr<IDWriteFontFile> fontFile;
-        RETURN_NULL_IF_FAILED(fontFace->GetFiles(&fileCount, &fontFile));
+        // Retrieve the font file that was used to create this font face
+        UINT32 fileCount;
+        fontFace->GetFiles(&fileCount, nullptr);
+        RETURN_RESULT_IF(fileCount == 0, userFontInfo->familyName.get());
+        if (fileCount > 1) {
+            CFIndex familyNameLength = CFStringGetLength(userFontInfo->familyName.get());
+            std::vector<UniChar> chars(familyNameLength + 1);
+            CFStringGetCharacters(userFontInfo->familyName.get(), { 0, familyNameLength }, chars.data());
+            TraceWarning(TAG, L"The font %ls has %d files, defaulting to the first", chars.data(), fileCount);
+            fileCount = 1;
+        }
 
-        // GetReferenceKey sets the input value to the pointer to fontFile's pointer to the data provider
+        ComPtr<IDWriteFontFile> fontFile;
+        RETURN_RESULT_IF_FAILED(fontFace->GetFiles(&fileCount, &fontFile), userFontInfo->familyName.get());
+
+        // Get the CGDataProvider used to register this font
+        // Note that GetReferenceKey will set fontProvider to a pointer to the file's internal CGDataProviderRef
         CGDataProviderRef* fontProvider = nullptr;
         UINT32 keySize;
         RETURN_RESULT_IF_FAILED(fontFile->GetReferenceKey((const void**)&fontProvider, &keySize), userFontInfo->familyName.get());
         RETURN_RESULT_IF(keySize != sizeof(fontProvider) || !fontProvider || !(*fontProvider), userFontInfo->familyName.get());
 
+        // From the CGDataProvider get the URL
         CFURLRef fontURL = _CGDataProviderGetURL(*fontProvider);
         RETURN_RESULT_IF(!fontURL, userFontInfo->familyName.get());
 
-        static const CFStringRef sc_prefix = CFSTR("ms-appx///");
+        // Now construct the XAML compatible family name, which should be in the format
+        // ms-appx:///[PATH TO FONT FILE]/font_file.ttf#family_name
+        static const CFStringRef sc_prefix = CFSTR("ms-appx:///");
         auto ret = woc::MakeAutoCF<CFMutableStringRef>(CFStringCreateMutableCopy(nullptr, 0, sc_prefix));
         auto path = woc::MakeAutoCF<CFStringRef>(CFURLCopyFileSystemPath(fontURL, kCFURLWindowsPathStyle));
         CFStringAppend(ret, path);
--- a/Frameworks/CoreGraphics/DWriteWrapper.mm
+++ b/Frameworks/CoreGraphics/DWriteWrapper.mm
@@ -210,7 +210,7 @@ CFArrayRef _DWriteCopyFontNamesForFamilyName(CFStringRef familyName) {
  * Note: This function currently uses a cache, meaning that fonts installed during runtime will not be reflected,
  * unless registered to the user font collection
  */
-CFStringRef _DWriteGetCompatibleFamilyName(CFStringRef fontName, IDWriteFontFace* fontFace) {
+CFStringRef _DWriteGetXamlCompatibleFamilyName(CFStringRef fontName, IDWriteFontFace* fontFace) {
     woc::unique_cf<CFStringRef> upperFontName(_CFStringCreateUppercaseCopy(fontName));
 
     const auto& info = __GetSystemFontCollectionHelper()->GetFontPropertiesFromUppercaseFontName(upperFontName);
@@ -220,19 +220,16 @@ CFStringRef _DWriteGetCompatibleFamilyName(CFStringRef fontName, IDWriteFontFace
 
     const auto& userFontInfo = __GetUserFontCollectionHelper()->GetFontPropertiesFromUppercaseFontName(upperFontName);
     if (userFontInfo) {
-        if (!fontFace) {
-            return userFontInfo->familyName.get();
-        }
+        RETURN_RESULT_IF(!fontFace, userFontInfo->familyName.get());
 
         // Retrieve the font file that was used to create this font face
         UINT32 fileCount;
-        fontFace->GetFiles(&fileCount, nullptr);
+        RETURN_RESULT_IF_FAILED(fontFace->GetFiles(&fileCount, nullptr), userFontInfo->familyName.get());
         RETURN_RESULT_IF(fileCount == 0, userFontInfo->familyName.get());
         if (fileCount > 1) {
-            CFIndex familyNameLength = CFStringGetLength(userFontInfo->familyName.get());
-            std::vector<UniChar> chars(familyNameLength + 1);
-            CFStringGetCharacters(userFontInfo->familyName.get(), { 0, familyNameLength }, chars.data());
-            TraceWarning(TAG, L"The font %ls has %d files, defaulting to the first", chars.data(), fileCount);
+            TraceWarning(TAG, L"The font has %d files, defaulting to the first file found", fileCount);
+
+            // Set fileCount to 1 so we just get the first file
             fileCount = 1;
         }
 
@@ -252,14 +249,10 @@ CFStringRef _DWriteGetCompatibleFamilyName(CFStringRef fontName, IDWriteFontFace
 
         // Now construct the XAML compatible family name, which should be in the format
         // ms-appx:///[PATH TO FONT FILE]/font_file.ttf#family_name
-        static const CFStringRef sc_prefix = CFSTR("ms-appx:///");
-        auto ret = woc::MakeAutoCF<CFMutableStringRef>(CFStringCreateMutableCopy(nullptr, 0, sc_prefix));
+        static const CFStringRef sc_format = CFSTR("ms-appx:///%@#%@");
         auto path = woc::MakeAutoCF<CFStringRef>(CFURLCopyFileSystemPath(fontURL, kCFURLWindowsPathStyle));
-        CFStringAppend(ret, path);
-
-        static const UniChar numberSign = '#';
-        CFStringAppendCharacters(ret, &numberSign, 1);
-        CFStringAppend(ret, userFontInfo->familyName.get());
+        auto ret =
+            woc::MakeAutoCF<CFStringRef>(CFStringCreateWithFormat(nullptr, nullptr, sc_format, path.get(), userFontInfo->familyName.get()));
         return (CFStringRef)CFAutorelease(ret.detach());
     }
 
--- a/Frameworks/CoreGraphics/DWriteWrapper.mm
+++ b/Frameworks/CoreGraphics/DWriteWrapper.mm
@@ -28,6 +28,7 @@
 #import "DWriteFontBinaryDataLoader.h"
 #import "DWriteFontCollectionHelper.h"
 #import "DWriteFontBinaryDataCollectionLoader.h"
+#import "CGDataProviderInternal.h"
 
 using namespace Microsoft::WRL;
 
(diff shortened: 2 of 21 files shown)


Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -209,8 +210,45 @@ CFArrayRef _DWriteCopyFontNamesForFamilyName(CFStringRef familyName) {
...
+        static const CFStringRef sc_prefix = CFSTR("ms-appx///");
+        auto ret = woc::MakeAutoCF<CFMutableStringRef>(CFStringCreateMutableCopy(nullptr, 0, sc_prefix));
+        auto path = woc::MakeAutoCF<CFStringRef>(CFURLCopyFileSystemPath(fontURL, kCFURLWindowsPathStyle));
+        CFStringAppend(ret, path);
+
+        static const UniChar numberSign = '#';
+        CFStringAppendCharacters(ret, &numberSign, 1);
+        CFStringAppend(ret, userFontInfo->familyName.get());
+        return (CFStringRef)CFAutorelease(ret.detach());
+    }
```
jaredhms: extensive comments would be helpful

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -209,8 +210,45 @@ CFArrayRef _DWriteCopyFontNamesForFamilyName(CFStringRef familyName) {
...
+
+    const auto& userFontInfo = __GetUserFontCollectionHelper()->GetFontPropertiesFromUppercaseFontName(upperFontName);
+    if (userFontInfo) {
+        if (!fontFace) {
+            return userFontInfo->familyName.get();
+        }
+
+        UINT32 fileCount = 1;
+        ComPtr<IDWriteFontFile> fontFile;
+        RETURN_NULL_IF_FAILED(fontFace->GetFiles(&fileCount, &fontFile));
```
jaredhms: if fileCount > 1, should we log a warning?

aballway: Probably a good idea

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -209,8 +210,45 @@ CFArrayRef _DWriteCopyFontNamesForFamilyName(CFStringRef familyName) {
...
+        // GetReferenceKey sets the input value to the pointer to fontFile's pointer to the data provider
+        CGDataProviderRef* fontProvider = nullptr;
+        UINT32 keySize;
+        RETURN_RESULT_IF_FAILED(fontFile->GetReferenceKey((const void**)&fontProvider, &keySize), userFontInfo->familyName.get());
+        RETURN_RESULT_IF(keySize != sizeof(fontProvider) || !fontProvider || !(*fontProvider), userFontInfo->familyName.get());
+
+        CFURLRef fontURL = _CGDataProviderGetURL(*fontProvider);
+        RETURN_RESULT_IF(!fontURL, userFontInfo->familyName.get());
+
+        static const CFStringRef sc_prefix = CFSTR("ms-appx///");
```
jaredhms: >ms-appx/// [](start = 52, length = 10)

weird - shouldn't this be ms-appx:/// ?

tests/unittests/CoreText/CTFontTests.mm
```diff
@@ -782,3 +783,24 @@ virtual void TearDown() {
...
+    EXPECT_EQ(nullptr, error);
+
+    auto font = woc::MakeAutoCF<CTFontRef>(CTFontCreateWithName(fontName, 20, nullptr));
+    EXPECT_NE(nullptr, font);
+
+    CFStringRef familyName = _CTFontGetCompatibleFamilyName(font);
+    EXPECT_NE(nil, familyName);
+
+    // CompatibleFamilyName returns an absolute path, so can only check prefix and suffix
+    EXPECT_TRUE([static_cast<NSString*>(familyName) hasPrefix:@"ms-appx///"]);
```
jaredhms: >ms-appx/// [](start = 64, length = 10)

ms-appx:/// ?

jaredhms:
:clock1:

Frameworks/CoreGraphics/DWriteWrapper.mm
aballway: It should, interesting that it still works 😨

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -209,8 +210,60 @@ CFArrayRef _DWriteCopyFontNamesForFamilyName(CFStringRef familyName) {
...
+
+        // Retrieve the font file that was used to create this font face
+        UINT32 fileCount;
+        fontFace->GetFiles(&fileCount, nullptr);
+        RETURN_RESULT_IF(fileCount == 0, userFontInfo->familyName.get());
+        if (fileCount > 1) {
+            CFIndex familyNameLength = CFStringGetLength(userFontInfo->familyName.get());
+            std::vector<UniChar> chars(familyNameLength + 1);
+            CFStringGetCharacters(userFontInfo->familyName.get(), { 0, familyNameLength }, chars.data());
+            TraceWarning(TAG, L"The font %ls has %d files, defaulting to the first", chars.data(), fileCount);
```
jaredhms: > first" [](start = 76, length = 7)

nit: ...first file found.

super-nit: also maybe useful to log the file name found?

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -209,8 +210,60 @@ CFArrayRef _DWriteCopyFontNamesForFamilyName(CFStringRef familyName) {
...
+            fileCount = 1;
```
jaredhms: nit: a bit subtle to do this here. maybe could use a comment why you're setting it to 1.

Frameworks/CoreGraphics/DWriteWrapper.mm
ms-jihua: re: super-nit: could potentially be PII, best not to, I think

jaredhms: hmm - font file names are PII?

ms-jihua: i can pretty easily imagine a scenario where say, someone makes a font of their own handwriting and names it FirstNameLastName.ttf, for instance. in general i think we should avoid logging file names/paths.

build/CoreText/dll/CoreText.def
```diff
@@ -139,6 +139,7 @@ LIBRARY CoreText
         _CTFontGetDWriteWeight
         _CTFontGetDWriteStretch
         _CTFontGetDWriteStyle
+        _CTFontGetCompatibleFamilyName
```
rajsesh-msft: >_CTFontGetCompatibleFamilyName [](start = 8, length = 30)

why the two sets of apis? why not let this go directly through CG instead (just like before?)

Frameworks/CoreGraphics/DWriteWrapper.mm
rajsesh-msft: All this trouble to trace a warning? %@ won't work?

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -209,8 +210,60 @@ CFArrayRef _DWriteCopyFontNamesForFamilyName(CFStringRef familyName) {
  * Note: This function currently uses a cache, meaning that fonts installed during runtime will not be reflected,
  * unless registered to the user font collection
  */
-CFStringRef _DWriteGetFamilyNameForFontName(CFStringRef fontName) {
-    return _DWriteGetFontPropertiesFromName(fontName)->familyName.get();
+CFStringRef _DWriteGetCompatibleFamilyName(CFStringRef fontName, IDWriteFontFace* fontFace) {
```
rajsesh-msft: Might want to rename this _DWriteGetXAMLCompatibleFamilyName

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -209,8 +210,60 @@ CFArrayRef _DWriteCopyFontNamesForFamilyName(CFStringRef familyName) {
...
+        // ms-appx:///[PATH TO FONT FILE]/font_file.ttf#family_name
+        static const CFStringRef sc_prefix = CFSTR("ms-appx:///");
+        auto ret = woc::MakeAutoCF<CFMutableStringRef>(CFStringCreateMutableCopy(nullptr, 0, sc_prefix));
+        auto path = woc::MakeAutoCF<CFStringRef>(CFURLCopyFileSystemPath(fontURL, kCFURLWindowsPathStyle));
+        CFStringAppend(ret, path);
+
+        static const UniChar numberSign = '#';
+        CFStringAppendCharacters(ret, &numberSign, 1);
+        CFStringAppend(ret, userFontInfo->familyName.get());
+        return (CFStringRef)CFAutorelease(ret.detach());
```
rajsesh-msft: All of these append calls can be collapsed into one CFStringAppendFormatAndArguments()

Frameworks/CoreText/CTFont.mm
```diff
@@ -945,6 +945,12 @@ CFTypeID CTFontGetTypeID() {
     return __kCTFontTypeID;
 }
 
+// Private function for getting the XAML-compatible family name
+CFStringRef _CTFontGetCompatibleFamilyName(CTFontRef font) {
```
rajsesh-msft: Get rid of this and call the _DWrite version in UIKit. Or insert XAML between Get and Compatible :)

aballway: Can't call directly from UIKit since the _DWrite version needs the _dwriteFontFace, so rename it is 🔹

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -209,8 +210,60 @@ CFArrayRef _DWriteCopyFontNamesForFamilyName(CFStringRef familyName) {
  * Note: This function currently uses a cache, meaning that fonts installed during runtime will not be reflected,
  * unless registered to the user font collection
  */
-CFStringRef _DWriteGetFamilyNameForFontName(CFStringRef fontName) {
-    return _DWriteGetFontPropertiesFromName(fontName)->familyName.get();
+CFStringRef _DWriteGetCompatibleFamilyName(CFStringRef fontName, IDWriteFontFace* fontFace) {
+    woc::unique_cf<CFStringRef> upperFontName(_CFStringCreateUppercaseCopy(fontName));
```
msft-Jeyaram: >oc::unique_cf<CFStringRef> upperFontName [](start = 5, length = 40)

use StrongCF?

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -209,8 +210,60 @@ CFArrayRef _DWriteCopyFontNamesForFamilyName(CFStringRef familyName) {
...
+    }
+
+    const auto& userFontInfo = __GetUserFontCollectionHelper()->GetFontPropertiesFromUppercaseFontName(upperFontName);
+    if (userFontInfo) {
+        if (!fontFace) {
+            return userFontInfo->familyName.get();
+        }
+
+        // Retrieve the font file that was used to create this font face
+        UINT32 fileCount;
```
msft-Jeyaram: >UINT32 [](start = 8, length = 6)

UInt32 ????

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -209,8 +210,60 @@ CFArrayRef _DWriteCopyFontNamesForFamilyName(CFStringRef familyName) {
...
+        fontFace->GetFiles(&fileCount, nullptr);
+        RETURN_RESULT_IF(fileCount == 0, userFontInfo->familyName.get());
```
msft-Jeyaram: mix of macros and ifs deep within the function.
why cant above be the same or vise versa?

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -209,8 +210,60 @@ CFArrayRef _DWriteCopyFontNamesForFamilyName(CFStringRef familyName) {
...
+
+    const auto& userFontInfo = __GetUserFontCollectionHelper()->GetFontPropertiesFromUppercaseFontName(upperFontName);
+    if (userFontInfo) {
+        if (!fontFace) {
+            return userFontInfo->familyName.get();
+        }
+
+        // Retrieve the font file that was used to create this font face
+        UINT32 fileCount;
+        fontFace->GetFiles(&fileCount, nullptr);
```
msft-Jeyaram: >  fontFace->GetFiles(&fileCount, nullptr); [](start = 6, length = 42)

the error is falling through, validate if the attempt to obtain the count was successful

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -209,8 +210,60 @@ CFArrayRef _DWriteCopyFontNamesForFamilyName(CFStringRef familyName) {
...
+        }
+
+        ComPtr<IDWriteFontFile> fontFile;
+        RETURN_RESULT_IF_FAILED(fontFace->GetFiles(&fileCount, &fontFile), userFontInfo->familyName.get());
+
+        // Get the CGDataProvider used to register this font
+        // Note that GetReferenceKey will set fontProvider to a pointer to the file's internal CGDataProviderRef
+        CGDataProviderRef* fontProvider = nullptr;
+        UINT32 keySize;
+        RETURN_RESULT_IF_FAILED(fontFile->GetReferenceKey((const void**)&fontProvider, &keySize), userFontInfo->familyName.get());
```
msft-Jeyaram: >(const void**)&fontProvider [](start = 58, length = 27)

woah that casting game

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -209,8 +210,60 @@ CFArrayRef _DWriteCopyFontNamesForFamilyName(CFStringRef familyName) {
...
+        RETURN_RESULT_IF(keySize != sizeof(fontProvider) || !fontProvider || !(*fontProvider), userFontInfo->familyName.get());
```
msft-Jeyaram: are you starting to use macros everywhere? :p

msft-Jeyaram: if so be consistent 

---
In reply to: [107769673](https://github.com/Microsoft/WinObjC/pull/2297#discussion_r107769673) [](ancestors = 107769673)

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -209,8 +210,60 @@ CFArrayRef _DWriteCopyFontNamesForFamilyName(CFStringRef familyName) {
...
+
+        // From the CGDataProvider get the URL
+        CFURLRef fontURL = _CGDataProviderGetURL(*fontProvider);
+        RETURN_RESULT_IF(!fontURL, userFontInfo->familyName.get());
+
+        // Now construct the XAML compatible family name, which should be in the format
+        // ms-appx:///[PATH TO FONT FILE]/font_file.ttf#family_name
+        static const CFStringRef sc_prefix = CFSTR("ms-appx:///");
```
msft-Jeyaram: >"ms-appx:///" [](start = 51, length = 13)

make it a const var

Frameworks/CoreGraphics/DWriteWrapper.mm
msft-Jeyaram: >CFStringRef)CFA [](start = 16, length = 15)

nit: static cast

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -209,8 +210,60 @@ CFArrayRef _DWriteCopyFontNamesForFamilyName(CFStringRef familyName) {
...
+        auto path = woc::MakeAutoCF<CFStringRef>(CFURLCopyFileSystemPath(fontURL, kCFURLWindowsPathStyle));
+        CFStringAppend(ret, path);
+
+        static const UniChar numberSign = '#';
+        CFStringAppendCharacters(ret, &numberSign, 1);
+        CFStringAppend(ret, userFontInfo->familyName.get());
+        return (CFStringRef)CFAutorelease(ret.detach());
+    }
+
+    return nullptr;
```
msft-Jeyaram: you are using early returns anyways, just switch it up, makes things simpler.

RETURN_NULL_IF(!userFontInfo);

then do the rest of the logic.

Frameworks/CoreGraphics/DWriteWrapper.mm
aballway: Expected type for GetFiles #ByDesign

Frameworks/CoreGraphics/DWriteWrapper.mm
aballway: GetFontPropertiesFromUppercaseFontName expects a unique_cf and changing that entire path is beyond the scope of this PR #ByDesign

msft-Jeyaram: :( we should of never let that occur in the first place.

---
In reply to: [107799297](https://github.com/Microsoft/WinObjC/pull/2297#discussion_r107799297) [](ancestors = 107799297)

Frameworks/CoreGraphics/DWriteWrapper.mm
msft-Jeyaram: curious why are you just mixing early returns and late returns.
be consistent. 

---
In reply to: [107775106](https://github.com/Microsoft/WinObjC/pull/2297#discussion_r107775106) [](ancestors = 107775106)

tests/unittests/CoreGraphics/DWriteWrapperTests.mm
```diff
@@ -19,14 +19,14 @@
...
-    EXPECT_OBJCEQ(@"Times New Roman", (id)_DWriteGetFamilyNameForFontName(CFSTR("Times New Roman Bold")));
-    EXPECT_OBJCEQ(@"Times New Roman", (id)_DWriteGetFamilyNameForFontName(CFSTR("Times New Roman Bold Italic")));
-    EXPECT_OBJCEQ(nil, (id)_DWriteGetFamilyNameForFontName(CFSTR("NotAFont")));
-}
\ No newline at end of file
+    EXPECT_OBJCEQ(@"Arial", (id)_DWriteGetXamlCompatibleFamilyName(CFSTR("Arial"), nullptr));
+    EXPECT_OBJCEQ(@"Arial", (id)_DWriteGetXamlCompatibleFamilyName(CFSTR("Arial Italic"), nullptr));
+    EXPECT_OBJCEQ(@"Arial", (id)_DWriteGetXamlCompatibleFamilyName(CFSTR("Arial Bold"), nullptr));
+    EXPECT_OBJCEQ(@"Arial", (id)_DWriteGetXamlCompatibleFamilyName(CFSTR("Arial Bold Italic"), nullptr));
+    EXPECT_OBJCEQ(@"Arial", (id)_DWriteGetXamlCompatibleFamilyName(CFSTR("Arial Black"), nullptr));
```
msft-Jeyaram: >   EXPECT_OBJCEQ(@"Arial", (id)_DWriteGetXamlCompatibleFamilyName(CFSTR("Arial Black"), nullptr)); [](start = 1, length = 98)

do these pass in OSX?

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -209,8 +210,53 @@ CFArrayRef _DWriteCopyFontNamesForFamilyName(CFStringRef familyName) {
...
+        RETURN_RESULT_IF_FAILED(fontFile->GetReferenceKey((const void**)&fontProvider, &keySize), userFontInfo->familyName.get());
+        RETURN_RESULT_IF(keySize != sizeof(fontProvider) || !fontProvider || !(*fontProvider), userFontInfo->familyName.get());
+
+        // From the CGDataProvider get the URL
+        CFURLRef fontURL = _CGDataProviderGetURL(*fontProvider);
+        RETURN_RESULT_IF(!fontURL, userFontInfo->familyName.get());
+
+        // Now construct the XAML compatible family name, which should be in the format
+        // ms-appx:///[PATH TO FONT FILE]/font_file.ttf#family_name
+        static const CFStringRef sc_format = CFSTR("ms-appx:///%@#%@");
```
msft-Jeyaram: >("ms-appx:///%@#%@") [](start = 50, length = 20)

what happened here?

Frameworks/CoreGraphics/CGDataProvider.mm
```diff
@@ -435,4 +448,8 @@ size_t _CGDataProviderGetSize(CGDataProviderRef provider) {
     return provider ? provider->GetSize() : 0;
 }
 
+CFURLRef _CGDataProviderGetURL(CGDataProviderRef provider) {
+    return provider ? provider->GetURL() : nullptr;
```
msft-Jeyaram: mix of macros and ternary within this PR.
Consistency is not going well :(

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -209,8 +210,53 @@ CFArrayRef _DWriteCopyFontNamesForFamilyName(CFStringRef familyName) {
...
+    const auto& userFontInfo = __GetUserFontCollectionHelper()->GetFontPropertiesFromUppercaseFontName(upperFontName);
+    if (userFontInfo) {
+        RETURN_RESULT_IF(!fontFace, userFontInfo->familyName.get());
+
+        // Retrieve the font file that was used to create this font face
+        UINT32 fileCount;
+        RETURN_RESULT_IF_FAILED(fontFace->GetFiles(&fileCount, nullptr), userFontInfo->familyName.get());
+        RETURN_RESULT_IF(fileCount == 0, userFontInfo->familyName.get());
+        if (fileCount > 1) {
+            TraceWarning(TAG, L"The font has %d files, defaulting to the first file found", fileCount);
```
msft-Jeyaram: >d [](start = 46, length = 1)

%u

tests/unittests/CoreGraphics/DWriteWrapperTests.mm
aballway: This is for DWrite so no 😛

Frameworks/CoreGraphics/DWriteWrapper.mm
aballway: Converted to one CreateWithFormat as per @rajsesh-msft's suggestion

Frameworks/CoreGraphics/CGDataProvider.mm
aballway: This is consistent w/ the file, and is a case where a ternary is just as if not more clear than a macro IMO

Frameworks/CoreGraphics/DWriteWrapper.mm
aballway: Depends on where it's clearer, here I think it makes sense to use an if

Frameworks/CoreGraphics/DWriteWrapper.mm
aballway: why do you think that?

PR #2306, opened 2017-03-23
Removed circular dependency between Frameworks.Core and Frameworks.UWP.

mukhole:
Separated Frameworks.UWP into Frameworks.UWP.Core (for RTObjCInteorp) and Frameworks.UWP (for projections).


pradipd:
Are there 2 changes here?
1. Remove circular dependency between Frameworks.core and frameworks.UWP 
2. separate rtobjcinterop into uwp.core

If so, the first one looks fine.  But, I thought we weren't going to do the 2nd change and instead move all frameworks to use c++/winrt.

pradipd:
:shipit:

bbowman:
:shipit:

PR #2309, opened 2017-03-23
Add support for UIButton presses/clicks via the keyboard and accessibility tools

### LINKED ISSUE #2308: UIButton should respond to keyboard and accessibility 'clicks' and 'presses'
UIButton's are focusable, but not yet functional via the keyboard (or narrator).  We should fix that; space and enter should work to press a button, and button 'clicks' should fire in such cases.

jaredhms:
Fixes #2308.

### DIFF
--- a/Frameworks/UIKit/UIButton.mm
+++ b/Frameworks/UIKit/UIButton.mm
@@ -88,7 +88,7 @@ struct ButtonState {
 
 static UIEdgeInsets _decodeUIEdgeInsets(NSCoder* coder, NSString* key) {
     id object = [coder decodeObjectForKey:key];
-    UIEdgeInsets* rawInsets = (UIEdgeInsets*)((char *)[object bytes] + 1);
+    UIEdgeInsets* rawInsets = (UIEdgeInsets*)((char*)[object bytes] + 1);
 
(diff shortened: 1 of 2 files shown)


Frameworks/UIKit/UIButton.mm
```diff
@@ -719,6 +747,24 @@ - (void)touchesMoved:(NSSet*)touchSet withEvent:(UIEvent*)event {
     [super touchesMoved:touchSet withEvent:event];
 }
 
+// Updates our highlighted state as needed
+- (void)_updateHighlightedState {
+    [self setHighlighted:(_isInTouchSequence || _isXamlElementPressed)];
```
tadam-msft: >_isXamlElementPressed [](start = 48, length = 21)

This method needs to be triggered by IsPressed changed; but can't this just access the xaml button's IsPressed property directly?

jaredhms: good point; we shouldn't need this extra variable

---
In reply to: [107766623](https://github.com/Microsoft/WinObjC/pull/2309#discussion_r107766623) [](ancestors = 107766623)

yiyang-msft:
:shipit:

PR #2311, opened 2017-03-23
Redo timestamp file change to allow partial rebuilds.

bbowman:
Fix previous implementation of creating the timestamp file at the beginning by
being more careful with project references between nuproj files and projects
pretending to be .nuprojs. The new implementation ensures that a .nuproj<->.nuproj
project reference is only evaluated in one configuration so that the timestamp file
creation doesn't happen more than once (happens once per conifg of .nuproj being built).

The full explanation of what was occuring is as follows:
```
A.nuproj -> B.nuproj -> NuprojInit.msbuildproj
         |--------------^

In the Any CPU config the build order was as follows:

A.nuproj (Any CPU) -> B.nuproj (Any CPU) -> NuprojInit.msbuildproj (Any CPU)
   |                     |
   |                     |-> B.nuproj (x86)
   |                     |-> B.nuproj (ARM)
   |
   |--> A.nuproj (x86)--> B.nuproj (x86) -> NuprojInit.msbuildproj (x86)
   |--> A.nuproj (ARM)--> B.nuproj (ARM) -> NuprojInit.msbuildproj (ARM)
```
The new build order is:
```
A.nuproj (Any CPU) -> B.nuproj (Any CPU) -> NuprojInit.msbuildproj (Any CPU)
   |                     |
   |                     |-> B.nuproj (x86)
   |                     |-> B.nuproj (ARM)
   |
   |--> A.nuproj (x86)
   |--> A.nuproj (ARM)
```
The key difference being NuprojInit.msbuildproj only being built in the original
configuration. An additional benefit of this change is that B.nuproj will be rebuilt less.


build/WinObjC.Frameworks/WinObjC.Frameworks.nuproj
```diff
@@ -110,6 +110,9 @@
     <ProjectReference Include="$(MSBuildThisFileDirectory)..\UIKit.Xaml\dll\UIKit.Xaml.vcxproj" />
     <ProjectReference Include="$(MSBuildThisFileDirectory)..\UIKit\dll\UIKit.vcxproj" />
     <ProjectReference Include="$(MSBuildThisFileDirectory)..\WebKit\dll\WebKit.vcxproj" />
+
+  <!-- NuProj Project References. Ensure that these only get evaluated once by checking that this isn't the second time through this project -->
+  <ItemGroup Condition="'$(SkipGetPackageContentsForOtherPlatforms)' != 'true'">
```
memontic-ms: ><ItemGroup Condition="'$(SkipGetPackageContentsForOtherPlatforms)' != 'true'"> [](start = 2, length = 78)

So this check is needed anytime a nuproj has a ProjectReference on another nuproj? As in, we will need to make sure to edit the .nuproj file and add this check whenever we add a reference to another nuproj?
Additionally, when does SkipGetPackageContentsForOtherPlatforms get set?

bbowman: Correct. It gets set [here](https://github.com/Microsoft/WinObjC/blob/develop/common/winobjc.nuproj.common.targets#L17) which is when the .nuproj gets re-evaluated for the other configs.

tools/legacy/BuildMonitor/Legacy_BuildMonitor.csproj
```diff
@@ -38,6 +38,7 @@
     <IncludeDebugSymbolsInLocalVSIXDeployment>true</IncludeDebugSymbolsInLocalVSIXDeployment>
     <CopyBuildOutputToOutputDirectory>true</CopyBuildOutputToOutputDirectory>
     <CopyOutputSymbolsToOutputDirectory>false</CopyOutputSymbolsToOutputDirectory>
+    <DeployExtension>False</DeployExtension>
```
rajsesh-msft: Did these just sneak in from a different change?

rajsesh-msft: never mind, found what happened here.


---
In reply to: [107792278](https://github.com/Microsoft/WinObjC/pull/2311#discussion_r107792278) [](ancestors = 107792278)

bbowman: these were in the in original. They are coming back from reverting the revert. I was seeing some instability in tools.sln builds without this. @memontic-ms says I don't these now but they don't hurt anything to my knowledge so I left them.

PR #2312, opened 2017-03-23
Plug leak in NSURLProtocol_WinHTTP and NSURLSession

### LINKED ISSUE #2232: Memory leaks in NSURLProtocol_WinHTTP.mm
The method `[NSURLProtocol_WinHTTP _asyncHttpOperation:completedWithStatus:]` leaks objects. I think this is because the method is called on a system thread that does not have a NSAutoreleasePool setup.

This can be reproduced using the code:

```
for (int i = 0; i < 100; i++) {    
    @autoreleasepool{
        NSURLSession *session = [NSURLSession sharedSession];
        NSMutableURLRequest *request = [NSMutableURLRequest requestWithURL:[NSURL URLWithString:@"https://www.bing.com"]];
        NSURLSessionDataTask  *task = [session dataTaskWithRequest:request completionHandler:^(NSData *data, NSURLResponse *response, NSError *error){
            NSLog(@"TASK completed");
        }];
        [task resume];
    }
}
```

Attached is a screenshot of the memory profile capture in VS.

image

aballway:
The significant changes are:

- Placing the contents of `-[asyncHttpOperation:completedWithStatus:]` in an `@autoreleasepool {}` because it is on a thread that may not have a pool set up
- Autoreleasing the results from `WRLToNSCollection` for NSDictionary and NSURLSession creation methods

Fixes #2232

### DIFF
--- a/Frameworks/Foundation/NSURLProtocol_WinHTTP.mm
+++ b/Frameworks/Foundation/NSURLProtocol_WinHTTP.mm
@@ -239,7 +239,7 @@ static std::function<R(Args...)> bindObjC(id instance, SEL _cmd) {
     StrongId<NSObject> strongInstance(instance);
     return [strongInstance, _cmd](Args... args) -> R {
         // args are not perfectly forwarded here, as objc_msgSend will not know what to do with them.
-        return ((R (*)(id, SEL, Args...))objc_msgSend)(strongInstance, _cmd, args...);
+        return ((R(*)(id, SEL, Args...))objc_msgSend)(strongInstance, _cmd, args...);
     };
 }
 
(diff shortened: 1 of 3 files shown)


Frameworks/include/CollectionHelpers.h
```diff
@@ -81,7 +81,7 @@ HRESULT WRLToNSCollection<ABI::Windows::Foundation::Collections::IMap<HSTRING, H
         });
     RETURN_IF_FAILED(hr);
 
-    *pDictionary = collection;
+    *pDictionary = [collection autorelease];
```
rajsesh-msft: >autorelease [](start = 31, length = 11)

nit: why not add this directly in the creation spot?

rajsesh-msft: In the other places too


---
In reply to: [107799430](https://github.com/Microsoft/WinObjC/pull/2312#discussion_r107799430) [](ancestors = 107799430)

Frameworks/Foundation/NSURLProtocol_WinHTTP.mm
```diff
@@ -290,150 +290,152 @@ - (void)stopLoading {
...
+                case HttpVersion_Http20:
+                    httpVersionStr = @"HTTP/2.0";
+                    break;
+                default:
+                    httpVersionStr = nil;
+            }
+            NSHTTPURLResponse* nsResponse = [[[NSHTTPURLResponse alloc] initWithURL:[_request URL]
+                                                                         statusCode:statusCode
+                                                                        HTTPVersion:httpVersionStr
+                                                                       headerFields:nsHeaders] autorelease];
```
bbowman: StrongId?

Frameworks/Foundation/NSURLProtocol_WinHTTP.mm
```diff
@@ -290,150 +290,152 @@ - (void)stopLoading {
...
-        ComPtr<IAsyncInfo> asyncInfo;
-        THROW_IF_FAILED(operation.As(&asyncInfo));
-
-        if (status == AsyncStatus::Error) {
-            HRESULT hr = S_OK;
-            THROW_IF_FAILED(asyncInfo->get_ErrorCode(&hr));
-            THROW_HR(hr);
-        } else if (status != AsyncStatus::Completed) {
-            THROW_HR(E_UNEXPECTED);
+    @autoreleasepool {
```
bbowman: >@autoreleasepool { [](start = 4, length = 18)

Is there something more general we should do to make sure our threads have pools set up?

rajsesh-msft: *Our* threads have autorelease pools setup.  This is something we need to watchout for any callbacks that happen in a system threadpool thread.  Has to be case-by-case?

Frameworks/include/CollectionHelpers.h
bbowman: Agreeed. if we hit an exception before here, there will be sadness. 

---
In reply to: [107801632](https://github.com/Microsoft/WinObjC/pull/2312#discussion_r107801632) [](ancestors = 107801632,107799430)

Frameworks/Foundation/NSURLProtocol_WinHTTP.mm
aballway: I'm not sure since this is using AsyncHttpOperation and Microsoft::WRL::Callback directly.  Any outside consumer should be either using objc methods to spawn threads or autoreleasepooling directly, so the only place I think this should be an issue is our code when we're using non-objc async stuffs

PR #2321, opened 2017-03-24
Switching to legacy {Binding} vs. x:Bind in ContentDialog.xaml so our…

### LINKED ISSUE #2320: WOCCatalog Alerts page doesn't show OK\Cancel buttons
Run 20170323.2 build of WOCCatalog, Click on Alerts menu item, then click on any of the Alerts options, buttons should be present on the alert that allow it to be closed.  On this build the buttons do not show up.  This happens on both Desktop and Phone.

jaredhms:
… ContendDialog Items are rendered properly.  Fixes #2320.

### DIFF
--- a/Frameworks/UIKit.Xaml/ContentDialog.xaml
+++ b/Frameworks/UIKit.Xaml/ContentDialog.xaml
@@ -119,7 +119,7 @@
         <ListView.ItemTemplate>
             <DataTemplate x:DataType="local:Item">
                 <!-- Consider making these buttons for accessibility purposes -->
-                <TextBlock Text="{x:Bind Title}" HorizontalAlignment="Left" TextTrimming="WordEllipsis" />
+                <TextBlock Text="{Binding Title}" HorizontalAlignment="Left" TextTrimming="WordEllipsis" />
             </DataTemplate>
         </ListView.ItemTemplate>
     </ListView>


Frameworks/UIKit.Xaml/ContentDialog.xaml
```diff
@@ -119,7 +119,7 @@
         <ListView.ItemTemplate>
             <DataTemplate x:DataType="local:Item">
                 <!-- Consider making these buttons for accessibility purposes -->
-                <TextBlock Text="{x:Bind Title}" HorizontalAlignment="Left" TextTrimming="WordEllipsis" />
```
yiyang-msft: >x:Bind [](start = 34, length = 6)

any known issue with xBind?

yiyang-msft:
:shipit:

Frameworks/UIKit.Xaml/ContentDialog.xaml
jaredhms: I'm not aware of any issues, and I didn't debug into the actual failure.  However, this change fixes the issue.  Perhaps we should send a repro to the Xaml team in case this is a bug in the platform.

---
In reply to: [108000838](https://github.com/Microsoft/WinObjC/pull/2321#discussion_r108000838) [](ancestors = 108000838)

PR #2322, opened 2017-03-24
Run clangformat over vsimporter

aballway:
The only substantive change beyond formatting is in scripts/git/clang-format.ps1, where I added support for *.hpp files and @memontic-ms cleaned up the File/Directory split a little bit.

### DIFF
--- a/scripts/git/clang-format.ps1
+++ b/scripts/git/clang-format.ps1
@@ -34,22 +34,22 @@ try {
     }
 
     # Process a single file if specified
-    if ($File) {
+    if ($File -ne "True") {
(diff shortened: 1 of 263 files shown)


scripts/git/clang-format.ps1
```diff
@@ -34,22 +34,22 @@ try {
     }
 
     # Process a single file if specified
-    if ($File) {
+    if ($File -ne "True") {
```
aballway: I'm sure there's some better way than this to check that $File is empty but they didn't seem to work, would appreciate input from powershell gurus

PR #2323, opened 2017-03-24
Improve perf in KVO and NSString

### LINKED ISSUE #2040: Investigate/fix performance issues in kvo codepath
after #1490, the perf in the kvo scenario has improved significantly.  We likely have other areas to go after.

ms-jihua:
- Fixed a bug in CFStringGetCStringPtr where it called _fastCStringContents with an NSString encoding,
  but _fastCStringContents expected a CFString encoding. As a result, CFStringGetCStringPtr now actually returns something for most strings.
    - Similarly, [NSString UTF8String] on a literal now returns a fast pointer rather than an autorelease buffer.
- Changed [NSString isEqualToString:], [NSConstantString hash], and [NSConstantString isEqualToString:] to rely on faster functions
    - Made a similar change to speed up a local helper (sorry @aballway i misled you)
- Changed _NSKVCSplitKeypath, [Class keyPathsForValuesAffectingKey:] to use C-string parsing
    - added a private helper for keyPathsForValuesAffectingKey: to not construct the empty set, for internal usage
- Changed NSKVOKeypathObserver to store an observer as assign instead of as weak.
  This makes us theoretically less safe if an observer deallocs without removing itself as an observer.
  However, the observed object already threw once it reached its dealloc and still had observers,
  and the above case is considered incorrect anyway,
  so it only allowed us to fail quietly in the case where the observed object has a much longer lifetime, which is of questionable value.
  Opting in favor of assign here, as it provided a substantial (-~40%) decrease in time it takes to addObserver:
- Changed _dispatchWillChange to just take an NSMutableDictionary instead of doing an unnecessary mutableCopy
- Minor change to setObservationInfo: to be atomic on the associated object table rather than the object itself
- Added more benchmark tests for KVO, String scenarios

Fixes #2040

### DIFF
--- a/Frameworks/CoreText/DWriteWrapper_CTFont.mm
+++ b/Frameworks/CoreText/DWriteWrapper_CTFont.mm
@@ -43,9 +43,7 @@ static const struct {
  * Helper function that efficiently compares two CFStringRef and returns true if they are equal (case insensitive)
  */
 static inline bool __CFStringEqual(CFStringRef left, CFStringRef right) {
-    CFIndex leftLength = CFStringGetLength(left);
-    return leftLength == CFStringGetLength(right) &&
-           CFStringFindWithOptions(left, right, { 0, leftLength }, kCFCompareCaseInsensitive, nullptr);
+    return CFStringCompare(left, right, kCFCompareCaseInsensitive) == kCFCompareEqualTo;
 }
 
 /**
--- a/Frameworks/Foundation/NSKVOSupport.mm
+++ b/Frameworks/Foundation/NSKVOSupport.mm
@@ -234,7 +234,7 @@ static NSSet* _keyPathsForValuesAffectingValueForKey(Class self, NSString* key) 
             // fast path using c string manipulation, will cover most cases, as most keyPaths are short
             char selectorName[sc_bufferSize] = "keyPathsForValuesAffecting"; // 26 chars
             selectorName[sc_prefixLength] = toupper(rawKey[0]);
-            strcat_s(&selectorName[sc_prefixLength + 1], sc_bufferSize - sc_prefixLength - 1, &rawKey[1]);
+            strcpy_s(&selectorName[sc_prefixLength + 1], sc_bufferSize - sc_prefixLength - 1, &rawKey[1]);
             sel = sel_registerName(selectorName);
 
         } else {
--- a/Frameworks/Foundation/NSKVOSupport.mm
+++ b/Frameworks/Foundation/NSKVOSupport.mm
@@ -66,9 +66,7 @@ NSString* const NSKeyValueChangeNotificationIsPriorKey = @"NSKeyValueChangeNotif
 #pragma region Keypath Observer
 @interface _NSKVOKeypathObserver () {
     long _changeDepth;
-    void* _rawObserver;
 }
-@property (nonatomic, readonly) void* _rawObserver;
 @end
 
 @implementation _NSKVOKeypathObserver
@@ -79,8 +77,7 @@ NSString* const NSKeyValueChangeNotificationIsPriorKey = @"NSKeyValueChangeNotif
                        context:(void*)context {
     if (self = [super init]) {
         _object = object;
-        objc_storeWeak(&_observer, observer);
-        _rawObserver = _observer;
+        _observer = observer;
         _keypath = [keypath copy];
         _options = options;
         _context = context;
@@ -89,21 +86,13 @@ NSString* const NSKeyValueChangeNotificationIsPriorKey = @"NSKeyValueChangeNotif
 }
 
 - (void)dealloc {
-    objc_destroyWeak(&_observer);
     [_keypath release];
     [_pendingChange release];
     [super dealloc];
 }
 
 - (id)observer {
-    return objc_loadWeak(&_observer);
-}
-
-- (void*)_rawObserver {
-    // We might need to remove an observer during its own deallocation;
-    // We therefore need the ability to grab the raw observer pointer during
-    // deregistration.
-    return _rawObserver;
+    return _observer;
 }
 
 - (bool)pushWillChange {
@@ -210,12 +199,10 @@ NSString* const NSKeyValueChangeNotificationIsPriorKey = @"NSKeyValueChangeNotif
 @end
 
 static _NSKVOObservationInfo* _createObservationInfoForObject(id object) {
-    @synchronized(object) {
-        _NSKVOObservationInfo* observationInfo = [_NSKVOObservationInfo new];
-        [object setObservationInfo:observationInfo];
-        [observationInfo release];
-        return observationInfo;
-    }
+    _NSKVOObservationInfo* observationInfo = [_NSKVOObservationInfo new];
+    [object setObservationInfo:observationInfo];
+    [observationInfo release];
+    return observationInfo;
 }
 #pragma endregion
 
@@ -225,6 +212,45 @@ static _NSKVOKeyObserver* _addKeypathObserver(id object,
                                               _NSKVOKeypathObserver* keyPathObserver,
                                               NSArray* affectedObservers);
 static void _removeKeyObserver(_NSKVOKeyObserver* keyObserver);
+
+// Private helper that backs the default implementation of keyPathsForValuesAffectingValueForKey:
+// Returns nil instead of constructing an empty set if no keyPaths are found
+// Also used internally as a minor optimization, to avoid constructing an empty set when it is not needed
+static NSSet* _keyPathsForValuesAffectingValueForKey(Class self, NSString* key) {
+    // This function can be a KVO bottleneck, so it will prefer to use c string manipulation when safe
+    NSUInteger keyLength = key.length;
+    if (keyLength > 0) {
+        static const size_t sc_bufferSize = 128;
+        static const size_t sc_prefixLength = strlen("keyPathsForValuesAffecting"); // 26
+
+        // max length of a key that can guaranteed fit in the char buffer,
+        // even if UTF16->UTF8 conversion causes length to double, or a null terminator is needed
+        static const size_t sc_safeKeyLength = (sc_bufferSize - sc_prefixLength) / 2 - 1; // 50
+
+        const char* rawKey = [key UTF8String];
+        SEL sel;
+
+        if (keyLength <= sc_safeKeyLength) {
+            // fast path using c string manipulation, will cover most cases, as most keyPaths are short
+            char selectorName[sc_bufferSize] = "keyPathsForValuesAffecting"; // 26 chars
+            selectorName[sc_prefixLength] = toupper(rawKey[0]);
+            strcat_s(&selectorName[sc_prefixLength + 1], sc_bufferSize - sc_prefixLength - 1, &rawKey[1]);
+            sel = sel_registerName(selectorName);
+
+        } else {
+            // guaranteed path
+            std::string selectorName("keyPathsForValuesAffecting");
+            selectorName += toupper(rawKey[0]);
+            selectorName += &rawKey[1];
+            sel = sel_registerName(selectorName.data());
+        }
+
+        if ([self respondsToSelector:sel]) {
+            return [self performSelector:sel];
+        }
+    }
+    return nil;
+}
 
 // Add all observers with declared dependencies on this one:
 // * All keypaths that could trigger a change (keypaths for values affecting us).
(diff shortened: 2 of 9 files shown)


ms-jihua:
Perf numbers, with some numbers from before #2240 for comparison image

Improvements across the board, but we're still slower than the old NSOperationQueue on adding operations (though not executing them). (Sorry for misleading you on this @rajsesh-msft, I forgot I switched hardware recently and would get lower numbers). The cost of addOperation: is currently about 2/3 KVO addObserver: and 1/3 libdispatch.

The biggest cost in KVO as it currently stands is construction of objc objects - NSKVOObservationInfo et al seem like they could easily be lighter-weight structs, but this would be a fairly heavy change that I didn't want to get into while @DHowett-MSFT was out - I will attempt to raise the issue with him once he returns.

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -226,6 +213,45 @@ - (bool)isEmpty {
                                               NSArray* affectedObservers);
 static void _removeKeyObserver(_NSKVOKeyObserver* keyObserver);
 
+// Private helper that backs the default implementation of keyPathsForValuesAffectingValueForKey:
+// Returns nil instead of constructing an empty set if no keyPaths are found
+// Also used internally as a minor optimization, to avoid constructing an empty set when it is not needed
+static NSSet* _keyPathsForValuesAffectingValueForKey(Class self, NSString* key) {
```
aballway: nit: rename self to something that's not a keyword.

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -43,9 +43,7 @@
  * Helper function that efficiently compares two CFStringRef and returns true if they are equal (case insensitive)
  */
 static inline bool __CFStringEqual(CFStringRef left, CFStringRef right) {
-    CFIndex leftLength = CFStringGetLength(left);
-    return leftLength == CFStringGetLength(right) &&
-           CFStringFindWithOptions(left, right, { 0, leftLength }, kCFCompareCaseInsensitive, nullptr);
+    return CFStringCompare(left, right, kCFCompareCaseInsensitive) == kCFCompareEqualTo;
```
aballway: 😆

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -446,15 +472,8 @@ + (BOOL)automaticallyNotifiesObserversForKey:(NSString*)key {
...
-    if ([key length] > 0) {
-        const char* rawKey = [key UTF8String];
-        auto selectorName = woc::string::format("keyPathsForValuesAffecting%c%s", toupper(rawKey[0]), rawKey + 1);
-        SEL sel = sel_registerName(selectorName.c_str());
-        if ([self respondsToSelector:sel]) {
-            return [self performSelector:sel];
-        }
-    }
-    return [NSSet set];
+    NSSet* ret = _keyPathsForValuesAffectingValueForKey(self, key);
```
aballway: _keyPathsForValuesAffectingValueForKey takes a Class but this is an NSObject?

ms-jihua: This is a + function, so self is a Class.

Frameworks/CoreText/DWriteWrapper_CTFont.mm
ms-jihua: Sorry! Dx

Frameworks/Foundation/NSKVOSupport.mm
ms-jihua: I think we have other instances of 'self' being what it would normally be, were this an objc function. Wontfixing this.

build/Tests/Benchmark/Framework.Benchmark.vcxproj
```diff
@@ -221,7 +221,9 @@
   </ItemGroup>
   <ItemGroup>
     <ClangCompile Include="$(StarboardBasePath)\tests\Benchmark\BenchmarkSampleTests.mm" />
+    <ClangCompile Include="$(StarboardBasePath)\tests\Benchmark\KVOBenchmarkTests.mm" />
```
rajsesh-msft: There is also a tests\testapps\kvoperf, which is probably obsolete now if we move all those tests here.

ms-jihua: I forgot about that - I don't think I'm comfortable removing it while @DHowett-MSFT is out, but I'll take a look and see if anything can be copied from there to here.

Frameworks/CoreText/DWriteWrapper_CTFont.mm
```diff
@@ -43,9 +43,7 @@
  * Helper function that efficiently compares two CFStringRef and returns true if they are equal (case insensitive)
```
rajsesh-msft: >efficiently [](start = 24, length = 11)

oh the irony!

tests/Benchmark/KVOBenchmarkTests.mm
```diff
@@ -0,0 +1,147 @@
...
+        @autoreleasepool {
+            for (size_t i = 0; i < 1000; ++i) {
+                [_testObservers addObject:[[NSObject new] autorelease]];
+            }
+        }
+    }
+
+    inline void Run() {
+        for (NSObject* obj in _testObservers.get()) {
+            [_testKVOObject addObserver:obj forKeyPath:@"intProperty" options:0 context:nullptr];
```
rajsesh-msft: NSArray now has the helper function to add observers for every object in the arraay, so you can cut this loop.

aballway: Though that has the added cost of making an NSIndexSet* and is only a convenience wrapper

build/Tests/Benchmark/Framework.Benchmark.vcxproj
rajsesh-msft: That app is not used at all (it is a manual tests).  When it was written, we didn't have the benchmark tests.  It should be safe to remove *once* all the pieces of it are salvaged.  

---
In reply to: [108012483](https://github.com/Microsoft/WinObjC/pull/2323#discussion_r108012483) [](ancestors = 108012483)

ms-jihua: It's more or less the same scenario as the ObserveChange benchmark, but also measures the amount of time spent in the actual notification part, which the benchmark framework doesn't really have a mechanism for.

Frameworks/Foundation/NSKeyValueCoding.mm
```diff
@@ -399,10 +402,10 @@ static bool quickSet(id self, SEL setter, id value, const char* valueType) {
 }
 
 #define QUICK_SET_CASE(type, name, capitalizedName, encodingChar) \
-    case encodingChar: \
-        if (quickSet<type>(self, setter, value, valueType)) { \
-            return true; \
-        } \
+    case encodingChar:                                            \
```
rajsesh-msft: No changes other than formatting in this file?

build/Tests/Benchmark/Framework.Benchmark.vcxproj
ms-jihua: (as opposed to ObserveChange, it is also a scenario with just one observer - do we want to add that to benchmarks, or do we consider the many-observer scenario mostly representative of health?)

Before/after:
image

Frameworks/Foundation/NSKeyValueCoding.mm
ms-jihua: Will back this out.

tests/Benchmark/KVOBenchmarkTests.mm
ms-jihua: I think it would end up being harder to read because of the need to construct an NSIndexSet. Leaving as-is.

build/Tests/Benchmark/Framework.Benchmark.vcxproj
rajsesh-msft: The multiple observer scenario is important.  We do see some improvement there so thats good to see.

---
In reply to: [108014282](https://github.com/Microsoft/WinObjC/pull/2323#discussion_r108014282) [](ancestors = 108014282)

ms-jihua:
Failure did not repro locally - i'll assume it's spurious and queue another CI build.

Frameworks/Foundation/NSString.mm
```diff
@@ -740,13 +743,11 @@ - (NSString*)description {
  @Status Interoperable
 */
 - (BOOL)isEqualToString:(NSString*)compStr {
-    RETURN_FALSE_IF(!compStr);
-
-    if ([compStr isKindOfClass:[NSString class]]) {
```
rajsesh-msft: >isKindOfClass [](start = 17, length = 13)

We can't use the short path for custom classes (such as NSStringOnHStting).  Why is this removed?

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -226,6 +213,45 @@ - (bool)isEmpty {
...
+        static const size_t sc_safeKeyLength = (sc_bufferSize - sc_prefixLength) / 2 - 1; // 50
+
+        const char* rawKey = [key UTF8String];
+        SEL sel;
+
+        if (keyLength <= sc_safeKeyLength) {
+            // fast path using c string manipulation, will cover most cases, as most keyPaths are short
+            char selectorName[sc_bufferSize] = "keyPathsForValuesAffecting"; // 26 chars
+            selectorName[sc_prefixLength] = toupper(rawKey[0]);
+            strcat_s(&selectorName[sc_prefixLength + 1], sc_bufferSize - sc_prefixLength - 1, &rawKey[1]);
```
rajsesh-msft: >strcat_s [](start = 12, length = 8)

strcpy instead since you have already computed the offset.

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -424,7 +450,7 @@ - (void*)observationInfo {
 @Status Interoperable
 */
 - (void)setObservationInfo:(void*)observationInfo {
-    objc_setAssociatedObject(self, &s_kvoObservationInfoAssociationKey, (__bridge id)observationInfo, OBJC_ASSOCIATION_RETAIN_NONATOMIC);
+    objc_setAssociatedObject(self, &s_kvoObservationInfoAssociationKey, (__bridge id)observationInfo, OBJC_ASSOCIATION_RETAIN);
```
rajsesh-msft: >OBJC_ASSOCIATION_RETAIN [](start = 102, length = 23)

why this subtle change here?

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -238,7 +264,7 @@ static void _addNestedObserversAndOptionallyDependents(_NSKVOKeyObserver* keyObs
 
     // Aggregate all keys whose values will affect us.
     if (dependents) {
-        NSSet* valueInfluencingKeys = [[object class] keyPathsForValuesAffectingValueForKey:key];
+        NSSet* valueInfluencingKeys = _keyPathsForValuesAffectingValueForKey([object class], key);
```
rajsesh-msft: >valueInfluencingKeys [](start = 15, length = 20)

could be nil. messages to nil object should be okay but i would double check this on ARM (we have had funky issues there but not with integral type rets).

Frameworks/Foundation/NSKVOSupport.mm
ms-jihua: Changed to this from an @ synchronized elsewhere. More of a consistency change (this had maybe ~3% perf impact) - there was another call to this function that didn't do the synchronized for whatever reason, and the synchronization should probably occur within the body of this function rather than around it. If you want, I can change this to @ synchronized instead of using the associated objects table's locking, but I don't think it makes much of a difference.

Frameworks/Foundation/NSString.mm
ms-jihua: am confused - NSHSTRINGString isKindOfClass:[NSString class] would have returned YES here...

Frameworks/Foundation/NSKVOSupport.mm
ms-jihua: seems to be fine - my only test failures were the CGGradient tests, which are a known issue. i doublechecked the nil message behavior in an isolated test as well.

Frameworks/Foundation/NSString.mm
rajsesh-msft: Yes it would.  Looking at CFStringCompareWithOptions, it can handle both non CF NSStrings as well as CFStrings, so this should be okay.  Still confused about the original intent here :)

---
In reply to: [108258325](https://github.com/Microsoft/WinObjC/pull/2323#discussion_r108258325) [](ancestors = 108258325)

bbowman:
:shipit:

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -573,7 +591,7 @@ - (void)willChangeValueForKey:(NSString*)key {
     if (![self observationInfo]) {
         return;
     }
-    _dispatchWillChange(self, key, @{ NSKeyValueChangeKindKey : @(NSKeyValueChangeSetting) });
+    _dispatchWillChange(self, key, [NSMutableDictionary dictionaryWithObject:@(NSKeyValueChangeSetting) forKey:NSKeyValueChangeKindKey]);
```
DHowett-MSFT: Oh no! @aballway this is what actually introduced the "information leak" I was talking about.

aballway: 😝 I told ya it was already there

PR #2325, opened 2017-03-24
Fix performance issue with NSData base64EncodedStringWithOptions

### LINKED ISSUE #1996: NSData base64EncodedStringWithOptions works very slow
Hi.
Converting of audio file of size around 7MB to string using base64EncodedStringWithOptions takes 2 minutes. Also to be able to playback audio file I have to delete all newlines  in the string I get from base64EncodedStringWithOptions and this operation takes 1 minute. So the app won't respond to user for 3 minutes because of these convertings.

```
NSLog(@"start encode data");
NSString *str = [audioData base64EncodedStringWithOptions:NSDataBase64Encoding64CharacterLineLength];
NSLog(@"end encode data");
str = [str stringByReplacingOccurrencesOfString:@"\r\n" withString:@""];
NSLog(@"end replacing newlines");
```

I took a look into the bridge and if I delete the code that adds linebreaks than the operation base64EncodedStringWithOptions takes 0.094 seconds
```

`/*  // Add new line every (lineLength) chars
            // Iterate backwards to avoid compensating for changes in length
            for (size_t i = newLineCount; i > 0; i--) {
                [mutableEncodedString insertString:newLineChar atIndex:(i * lineLength)];
            }

			*/`

```

Why are these newlines added? Is it possible to get rid of them or make some optimization?

Thank you in advance!
Kind regards,
Julia

rajsesh-msft:
Our perf was terrible because we were inserting new lines into a mutable string and making a copy of it.  Fixing by processing everything in a raw buffer and creating the resulting string.  

Perf on release build for a 3MB buffer:

Before: 21s (21,347,000.00Ms)
After: 28ms (28,929.70 Ms)

Fixes #1996

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2325)

### DIFF
--- a/Frameworks/Foundation/NSData.mm
+++ b/Frameworks/Foundation/NSData.mm
@@ -124,7 +124,7 @@ BASE_CLASS_REQUIRED_IMPLS(NSData, NSDataPrototype, CFDataGetTypeID);
                 memcpy(dest, source, (rawLength % lineLength)*sizeof(wchar_t));
             }
 
-            return [NSString stringWithCharacters:reinterpret_cast<unichar*>(result.get()) length:newLength];
+            return [NSString stringWithCharacters:reinterpret_cast<UniChar*>(result.get()) length:newLength];
         }
     } 
 
--- a/Frameworks/Foundation/NSData.mm
+++ b/Frameworks/Foundation/NSData.mm
@@ -80,11 +80,6 @@ BASE_CLASS_REQUIRED_IMPLS(NSData, NSDataPrototype, CFDataGetTypeID);
     Wrappers::HString encodedString;
     RETURN_NULL_IF_FAILED(cryptographicBufferStatics->EncodeToBase64String(wrlBuffer.Get(), encodedString.GetAddressOf()));
 
-    unsigned int rawLength;
-    const wchar_t* rawEncodedString = WindowsGetStringRawBuffer(encodedString.Get(), &rawLength);
-
-    NSString* nsEncodedString = [NSString _stringWithHSTRING:encodedString.Get()];
-
     // Do any necessary post-processing for options
     // If line length specified, place a newline character (/r, /n, or /r/n every (specified number) characters)
     if ((options & NSDataBase64Encoding64CharacterLineLength) || (options & NSDataBase64Encoding76CharacterLineLength)) {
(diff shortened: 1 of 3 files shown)


Frameworks/Foundation/NSData.mm
```diff
@@ -80,49 +80,56 @@ - (NSString*)base64EncodedStringWithOptions:(NSDataBase64EncodingOptions)options
...
+                dest += newLineLength;
+            }
+            
+            if (rawLength % lineLength != 0) {
+                // finally add the remaining characters
+                memcpy(dest, source, (rawLength % lineLength)*sizeof(wchar_t));
             }
 
-            return [NSString stringWithString:mutableEncodedString];
+            return [NSString stringWithCharacters:reinterpret_cast<unichar*>(result.get()) length:newLength];
```
aballway: super nit: UniChar*

tests/Benchmark/NSDataBenchmarkTests.mm
```diff
@@ -0,0 +1,42 @@
...
+#import <Starboard/SmartTypes.h>
+#import "Benchmark.h"
+#import <CppUtils.h>
+
+class EncodeStringWithOptions : public ::benchmark::BenchmarkCaseBase {
+    StrongId<NSMutableData> m_data;
+
+public:
+    void PreRun() {
+        m_data = [NSMutableData dataWithLength:3*1024*1024];
```
aballway: nit: do this in the constructor so it's only made once vs once per run.

Frameworks/Foundation/NSData.mm
```diff
@@ -80,49 +80,56 @@ - (NSString*)base64EncodedStringWithOptions:(NSDataBase64EncodingOptions)options
...
+
+            for (size_t i = 0; i < newLineCount; ++i) {
+                memcpy(dest, source, lineLength * sizeof(wchar_t));
+                dest += lineLength;
+                source += lineLength;
+                memcpy(dest, newLineChar, newLineLength * sizeof(wchar_t));
+                dest += newLineLength;
+            }
+            
+            if (rawLength % lineLength != 0) {
```
aballway: super nit: save in a variable rather than calculating twice

rajsesh-msft:
@aballway NSDataTests.mm already has good 'ol fashion tests to handle the corner cases :).

aballway:
I thought there were only single-line tests, but upon closer inspection it plays with a couple of lines :D

---
In reply to: [289160436](https://github.com/Microsoft/WinObjC/pull/2325#issuecomment-289160436) [](ancestors = 289160436)

rajsesh-msft:
Through mathematical induction, we have proven that it will work for any number of lines :).

---
In reply to: [289161023](https://github.com/Microsoft/WinObjC/pull/2325#issuecomment-289161023) [](ancestors = 289161023,289160436)

Frameworks/Foundation/NSData.mm
rajsesh-msft: optimizer should deal with this.

---
In reply to: [108012500](https://github.com/Microsoft/WinObjC/pull/2325#discussion_r108012500) [](ancestors = 108012500)

Frameworks/Foundation/NSData.mm
```diff
@@ -80,49 +80,56 @@ - (NSString*)base64EncodedStringWithOptions:(NSDataBase64EncodingOptions)options
...
             }
 
-            NSMutableString* mutableEncodedString = [NSMutableString stringWithString:nsEncodedString];
+            const size_t newLineLength = wcslen(newLineChar);
+            const size_t newLength = rawLength + newLineCount * newLineLength;
 
-            // Add new line every (lineLength) chars
-            // Iterate backwards to avoid compensating for changes in length
-            for (size_t i = newLineCount; i > 0; i--) {
-                [mutableEncodedString insertString:newLineChar atIndex:(i * lineLength)];
```
ms-jihua: bit concerned that such a basic NSMutableString functio performs so poorly...

Frameworks/Foundation/NSData.mm
```diff
@@ -80,49 +80,56 @@ - (NSString*)base64EncodedStringWithOptions:(NSDataBase64EncodingOptions)options
...
+            std::unique_ptr<wchar_t> result(new wchar_t[newLength]);
+
+            const wchar_t* source = rawEncodedString;
+            wchar_t* dest = result.get();
+
+            for (size_t i = 0; i < newLineCount; ++i) {
+                memcpy(dest, source, lineLength * sizeof(wchar_t));
```
ms-jihua: what are your feelings on taking a bit of a perf-for-safety tradeoff and using std::wstring instead?

Frameworks/Foundation/NSData.mm
rajsesh-msft: interestingly, even inserting at end performed poorly - likely due to different encodings (UTF8 vs UniChar).  Inserting in the middle of the string was 10x faster, but still terrible.  Hence this raw buffer manipulation.

---
In reply to: [108017657](https://github.com/Microsoft/WinObjC/pull/2325#discussion_r108017657) [](ancestors = 108017657)

Frameworks/Foundation/NSData.mm
rajsesh-msft: What is the safety concern here? 

---
In reply to: [108017729](https://github.com/Microsoft/WinObjC/pull/2325#discussion_r108017729) [](ancestors = 108017729)

ms-jihua: easier maintenance, mostly - we can be mostly confident that the current impl is correct, but it's a lot harder for someone to come by later and mess it up if it's a simple += to a wstring

rajsesh-msft: I rewrote that part using std::wstring, it was still 3x slower than this version.  Since this is not overly complicated, I am leaving in the better performing code.

PR #2328, opened 2017-03-27
Fix #2313: UITableView items no longer stretch to the size of their parent on window resize 

### LINKED ISSUE #2313: UITableView items no longer stretch to the size of their parent on window resize
If you launch WOCCatalog and then resize the window, the UITableView cells don't stretch to compensate:

image

I suspect this is a regression from 2646eb8.

yiyang-msft:
when canvas width is smaller than scrollview, we need adjust canvas's width to be width of scrollview.

### DIFF
--- a/Frameworks/UIKit.Xaml/ScrollView.xaml.cpp
+++ b/Frameworks/UIKit.Xaml/ScrollView.xaml.cpp
@@ -65,6 +65,17 @@ ScrollViewer^ ScrollView::InnerScrollViewer::get() {
     return scrollViewer;
 }
 
+Windows::Foundation::Size ScrollView::ArrangeOverride(Windows::Foundation::Size finalSize) {
+    if (this->SublayerCanvas->Width < Width) {
+        // if the canvas's width is smaller than that of scrollview
+        // need adjust canvas's width to be same of the scrollview
+        // in order to have proper resizing behavior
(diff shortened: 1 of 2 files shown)


Frameworks/UIKit.Xaml/ScrollView.xaml.cpp
```diff
@@ -65,6 +65,17 @@ ScrollViewer^ ScrollView::InnerScrollViewer::get() {
     return scrollViewer;
 }
 
+Windows::Foundation::Size ScrollView::ArrangeOverride(Windows::Foundation::Size finalSize) {
+    if (this->SublayerCanvas->Width < Width) {
+        // if the canvas's width is smaller than that of scrollview
+        // need adjust canvas's width to be same of the scrollview
+        // in order to have proper resizing behavior
+        this->SublayerCanvas->Width = finalSize.Width;
+    }
```
jaredhms: what about height?

yiyang-msft: we don't adjust the height when height is smaller because you don't want to stretch the height vertically - which is current hehavior.

---
In reply to: [108253765](https://github.com/Microsoft/WinObjC/pull/2328#discussion_r108253765) [](ancestors = 108253765)

jaredhms:
:shipit:

PR #2330, opened 2017-03-27
Make sample and test apps open in Visual Studio 2017

### LINKED ISSUE #2327: Sample projects open in VS2015 by default, then fail to load
The WOCCatalog-WinStore10.sln solution opens in VS 2015 by default, with the following error:
image

Trying to reload project produces the following error:
image

To fix this, the solution files will need to be updated to VS2017.

memontic-ms:
fixes #2327

These apps require Visual Studio 2017 because they use PackageReference


bbowman:
:shipit:

PR #2333, opened 2017-03-27
Auto-Restore Build.sln

bbowman:


bbowman:
Note: This only works when building in visual studio. Also this means that the CI build won't accurately reflect this change. If we want other builds to be able to take part, they need to add this project to their solution which is super unfortunate.

bbowman:
The reason for both points above is that VS uses a bottom up approach that first analyzes dependencies and then builds the projects bottom to top so this new project will be loaded and built before anyone that depends on it. In msbuild (command line) it operates in the reverse (top down) so there isn't a chance to restore first as the project is already loaded.

---
In reply to: [289613699](https://github.com/Microsoft/WinObjC/pull/2333#issuecomment-289613699) [](ancestors = 289613699)

build/build.sln
```diff
@@ -795,6 +795,8 @@ Project("{5DD5E4FA-CB73-4610-85AB-557B54E96AA9}") = "WinObjC.Frameworks.UWP.Core
 EndProject
 Project("{13B669BE-BB05-4DDF-9536-439F39A36129}") = "NuprojInitializer", "..\common\NuprojInitializer.msbuildproj", "{417B0D00-F90E-47BB-9085-BE7E412C9DA8}"
 EndProject
+Project("{13B669BE-BB05-4DDF-9536-439F39A36129}") = "NugetRestore", "..\common\NugetRestore.msbuildproj", "{B2C6186E-B340-49D8-888E-76ED83752E43}"
+EndProject
```
jaredhms: should we do this for all test/sample apps?

common/winobjc.packagereference.override.targets
```diff
@@ -179,4 +179,13 @@
   </Target>
 
 
+  <ItemGroup>
```
jaredhms: nit: worth adding a comment about this?

jaredhms:
:shipit:

PR #2336, opened 2017-03-27
Change Packaging's header include path to a non-clang path

### LINKED ISSUE #2277: [Packaging] WinObjC Packaging package does not correctly set the include paths for the package consumers
The props file (for consumers) is using InternalSystemIncludePaths, it should be using UserIncludes or AdditionalIncludes instead.

memontic-ms:
Fixes #2277

WinObjC.Packaging's default header include paths now modify "AdditionalIncludeDirectories"

This path should not rely on a ClangCompile path since WinObjC.Packaging (and any package created with it) does not depend on WinObjC.Language.


bbowman:
:shipit:

PR #2338, opened 2017-03-28
Bbowman/restore tests samples

bbowman:


samples/CoreAnimationTest/CoreAnimationTest-WinStore10.sln
```diff
@@ -6,33 +6,52 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "CoreAnimationTest", "CoreAn
 EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "CoreAnimationTest", "CoreAnimationTest.vsimporter\CoreAnimationTest-WinStore10\CoreAnimationTest.vcxproj", "{9F755F4B-5A21-4F6A-A690-5C9AA0ABE302}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "BuildSupport", "BuildSupport", "{52059437-69B4-4E0F-B713-416B841A59DC}"
```
bbowman: Only for packagereferences and only until it's fixed. It should have a copy of this project though instead of pointing to this one.

common/winobjc.packagereference.override.targets
```diff
@@ -194,4 +194,16 @@
     </ProjectReference>
   </ItemGroup>
 
+       NOTE: To make this work, this project MUST be added to the solution so that Visual Studio can see it and build it before building any projects that
```
jaredhms: >        [](start = 0, length = 7)

where's the starting <!-- ?

bbowman: bad merge maybe. let me look.

PR #2347, opened 2017-03-28
Make vsimporter generate for Visual Studio 2017

memontic-ms:
This leaves no compatibility with Visual Studio 2015 for solutions and projects generated with vsimporter.

Leaving in code-only switches for the possibility to generate for VS 2015 if we need to support both in the future. The exe will only allow generation for VS 2017.

I'm doing manual validation locally. I'll update the PR if I find anything wrong.


tools/vsimporter/src/vswriter/VSSolution.cpp
```diff
@@ -228,11 +228,16 @@ void VSSolution::write(std::ostream& out) const {
...
         out << "VisualStudioVersion = 12.0.21005.1" << std::endl;
-    } else {
+        out << "MinimumVisualStudioVersion = 10.0.40219.1" << std::endl;
+    } else if (m_version == 14) {
         out << "# Visual Studio 14" << std::endl;
         out << "VisualStudioVersion = 14.0.22823.1" << std::endl;
+        out << "MinimumVisualStudioVersion = 14.0.22823.1" << std::endl;
+    } else {
+        out << "# Visual Studio 15" << std::endl;
+        out << "VisualStudioVersion = 15.0.26228.4" << std::endl;
```
memontic-ms: Just picked 15.0.26228.4 and 15.0.26206.0 based on what I've seen but VS doesn't really care about VisualStudioVersion and MinimumVisualStudioVersion anyway. It'll update these two numbers as it pleases.

It really only cares about "# Visual Studio 15". That's where it's deciding compatibility for 2015 vs 2017.

pradipd:
Excellent!  I was just about to file a bug about this.

pradipd:
:shipit:

PR #2350, opened 2017-03-29
Fixing override-projectjson to handle project.json with a depth > 2.

pradipd:
By default converto-json only handles json data with a depth of 2.  After that it starts "toString()'ing" things.
https://msdn.microsoft.com/en-us/powershell/reference/5.1/microsoft.powershell.utility/convertto-json

I changed the max depth to 100,  this is the max supported value.


pradipd:

Issue #2352, opened 2017-03-29
[Foundation Dependency Removal] CoreGraphics removal of Foundation usage.


msft-Jeyaram:
- [x] CGPDFContext: Change the const strings to use CFSTR vs NSStrings
- [x] CGColorSpace: Change the const strings to use CFSTR vs NSStrings
- [x] CGImage: Change function return from NSData to CFData

msft-Jeyaram:
Task complete with above Pull requests being merged in.

PR #2361, opened 2017-03-29
[Foundation Dependency Removal] Should use use CoreFoundation vs Foundation: Strings should be created via CFSTR vs NS

msft-Jeyaram:
#2352

### DIFF
--- a/Frameworks/CoreGraphics/CGColorSpace.mm
+++ b/Frameworks/CoreGraphics/CGColorSpace.mm
@@ -27,18 +27,18 @@
 
 static const wchar_t* TAG = L"CGColorSpace";
 
-const CFStringRef kCGColorSpaceGenericGray = static_cast<CFStringRef>(@"kCGColorSpaceGenericGray");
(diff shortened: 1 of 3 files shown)


aballway:
> #import "_CGLifetimeBridgingType.h"

Should this go as well?

---
Refers to: Frameworks/CoreGraphics/CGColorSpace.mm:25 in 19da71a. [](commit_id = 19da71a769bc7fdcd5db7b38ca18efcb24d3d4d8, deletion_comment = False)

ms-jihua:
Is there broader scoped context to this change? (Are we looking to remove enough dependencies that we can make some sort of build change, or are we just removing dependencies for code quality's sake?)

msft-Jeyaram:
@aballway  yes yes yes
@ms-jihua  this is a small part of the grand plan of completely removing Foundation/UIKit dependencies from CoreText and CoreGraphics

msft-Jeyaram:
There are few more instances, updating CR.

PR #2366, opened 2017-03-29
CGContextTest:ContextCoordinateTest should use CGBitmapContext than creating a WIC render target

### LINKED ISSUE #2365: ContextCoordinateTest should use CGBitmapContext rather than creating a WIC render target

msft-Jeyaram:
fixes  #2365


msft-Jeyaram:
nuget restore failure

PR #2372, opened 2017-03-30
UIKit.Xaml post-packaging cleanup

### LINKED ISSUE #2261: Address UIKit cleanup with packaging
1. App.xaml
2. UIKit xaml files
3. xib2nib <-> UIKit dependencies?

### LINKED ISSUE #2319: UIKit should not depend on protected WinRT interfaces and methods.
UIKit depends on getTemplateChild call, this method is protected and should not be used.

jaredhms:
Formalizing how we use our UIKit.Xaml markup at runtime; we now ensure that all relevant UIKit.Xaml markup is included in each app's resources.pri, so we no longer need them on disk; fixes #2261

Also removing use of protected WinRT API 'getTemplateChild'; fixes #2319.


jaredhms:
Thanks, @bbowman for all of your help with this!

Frameworks/UIKit/UILabel.mm
```diff
@@ -188,10 +188,6 @@ - (void)_initUILabel {
     WXCTextBlock* textBlock = nil;
```
yiyang-msft: >WXCTextBlock* textBlock = nil; [](start = 4, length = 30)

super nit: move closer to its usage.

jaredhms: how much closer could it be? :)

---
In reply to: [108989011](https://github.com/Microsoft/WinObjC/pull/2372#discussion_r108989011) [](ancestors = 108989011)

yiyang-msft: nvm. :S

---
In reply to: [108989240](https://github.com/Microsoft/WinObjC/pull/2372#discussion_r108989240) [](ancestors = 108989240,108989011)

yiyang-msft:
:shipit:

tools/WinObjC.Packaging/Package.Default.props
```diff
@@ -18,7 +18,6 @@
     <_ImportLibsFromPackage Include="$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\*.lib"/>    
     <_ImportLibsFromPackage Include="$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\$(Configuration)\*.lib"/>
 
-    <_XamlFilesFromPackage Include="$(MSBuildThisFileDirectory)\xaml\**\*.xaml" />
```
bbowman: ><_XamlFilesFromPackage Include="$(MSBuildThisFileDirectory)\xaml\**\*.xaml" /> [](start = 4, length = 78)

Why is this file modified but doesn't have the new .pri file consumption targets?

Package.Default.props is the default consumption props/targets for middleware using WinObjC

If they have xaml components, what needs to happen to ensure it works for them?

msvc/sdk-build.targets
```diff
@@ -38,7 +38,7 @@
       <GetPackageContentsDependsOn>$(GetPackageContentsDependsOn);$(ComputeLinkInputsTargets)</GetPackageContentsDependsOn>
   </PropertyGroup>
 
-  <Target Name="AddBuiltOutput" BeforeTargets="GetPackageContents"  DependsOnTargets="$(GetPackageContentsDependsOn)" Condition="'$(ConfigurationType)' == 'DynamicLibrary' or '$(ConfigurationType)' == 'Application'">
+  <Target Name="AddBuiltOutput" BeforeTargets="GetPackageContents" DependsOnTargets="$(GetPackageContentsDependsOn);_GenerateProjectPriFileCore" Condition="'$(ConfigurationType)' == 'DynamicLibrary' or '$(ConfigurationType)' == 'Application'">
```
bbowman: >_GenerateProjectPriFileCore [](start = 116, length = 27)

This should probably be in a conditioned property group to += it onto GetPackageContentsDependsOn based on if its an AppContainerApplication project. 

Since this is being added to build.sln's common targets, you should probably consider adding it to the Packaging Package's default *production* targets so that the same things happen for middleware consumers.

msvc/sdk-build.targets
```diff
@@ -64,7 +64,15 @@
       <PackageFile Include="@(_WinMDFileNames->Distinct())" Condition="Exists('%(FullPath)')">
         <PackagePath>$(PackageContentBasePath)\%(Filename)%(Extension)</PackagePath>
       </PackageFile>
-    </ItemGroup>
+      
+      <!-- Package up .pri files associated with any .winmds -->
+      <_PriFileNames Include="$(ProjectPriFullPath)">
+      </_PriFileNames>
+      <PackageFile Include="@(_PriFileNames)" Condition="Exists('%(RootDir)%(Directory)%(Filename).winmd')">
```
bbowman: > Condition="Exists('%(RootDir)%(Directory)%(Filename).winmd') [](start = 45, length = 61)

This condition really isn't what we want.

You should be checking for the existence of the .pri file not the .winmd and then And ing with the notion that "this project is going to produce a winmd". 

Are there any scenarios where a pri file is needed but a winmd isn't created?

msvc/sdk-build.targets
jaredhms: 1) example?

2) where is that?

---
In reply to: [108993817](https://github.com/Microsoft/WinObjC/pull/2372#discussion_r108993817) [](ancestors = 108993817)

tools/WinObjC.Packaging/Package.Default.props
jaredhms: the changes we made to WinObjC.Frameworks.targets?

---
In reply to: [108993288](https://github.com/Microsoft/WinObjC/pull/2372#discussion_r108993288) [](ancestors = 108993288)

msvc/sdk-build.targets
jaredhms: No, there's no such scenario.  I don't understand what's wrong with this; we only want to grab .pri files that have associated .winmds.  If you have a better solution, please share it.

---
In reply to: [108994152](https://github.com/Microsoft/WinObjC/pull/2372#discussion_r108994152) [](ancestors = 108994152)

tools/WinObjC.Packaging/Package.Default.props
bbowman: WinObjC.Frameworks.targets are the consumption rules for the WinObjc.Frameworks package, so yes you want to think about how to generalize what is going on there to work for middleware.

---
In reply to: [108994382](https://github.com/Microsoft/WinObjC/pull/2372#discussion_r108994382) [](ancestors = 108994382,108993288)

jaredhms: What's middleware's version of WinObjC.Frameworks.targets?

---
In reply to: [108994710](https://github.com/Microsoft/WinObjC/pull/2372#discussion_r108994710) [](ancestors = 108994710,108994382,108993288)

build/UIKit.Xaml/dll/UIKit.Xaml.vcxproj
```diff
@@ -225,18 +225,12 @@
   <ImportGroup Label="ExtensionTargets">
     <Import Project="$(StarboardBasePath)\msvc\sdk-build.targets" />
   </ImportGroup>
-  <ItemGroup>
-    <ReferenceCopyLocalPaths Include="$(MSBuildThisFileDirectory)$(XamlGeneratedOutputPath)\**\*.xaml">
-      <DestinationSubDirectory>$(OutputName)\%(RecursiveDir)</DestinationSubDirectory>
-    </ReferenceCopyLocalPaths>
-  </ItemGroup>
-  <Target Name="AddExtraPackageItems" BeforeTargets="GetPackageContents">
+  <Target Name="CopyXbfFiles" BeforeTargets="_CopyFilesMarkedCopyLocal" DependsOnTargets="$(AfterBuildCompileTargets)">
```
bbowman: > <Target Name="CopyXbfFiles" BeforeTargets="_CopyFilesMarkedCopyLocal" DependsOnTargets="$(AfterBuildCompileTargets)"> [](start = 1, length = 118)

So why was this target needed again? The pri files that we package have the xbf data embedded in them right? 

This is *strictly* for functional tests?

If so please add a comment for this target explaining.

tools/WinObjC.Packaging/Package.Default.props
bbowman: See my above comment, 

"Package.Default.props is the default consumption props/targets for middleware using WinObjC"

---
In reply to: [108995058](https://github.com/Microsoft/WinObjC/pull/2372#discussion_r108995058) [](ancestors = 108995058,108994710,108994382,108993288)

build/UIKit.Xaml/dll/UIKit.Xaml.vcxproj
jaredhms: Right; this was for functional tests; i'll add a comment.

---
In reply to: [108995251](https://github.com/Microsoft/WinObjC/pull/2372#discussion_r108995251) [](ancestors = 108995251)

msvc/sdk-build.targets
bbowman: Right above is an example.

```xml
  <PropertyGroup Condition="'$(ConfigurationType)' == 'DynamicLibrary' or '$(ConfigurationType)' == 'Application'">
      <GetPackageContentsDependsOn>$(GetPackageContentsDependsOn);$(ComputeLinkInputsTargets)</GetPackageContentsDependsOn>
  </PropertyGroup>
```


---
In reply to: [108994265](https://github.com/Microsoft/WinObjC/pull/2372#discussion_r108994265) [](ancestors = 108994265,108993817)

tools/WinObjC.Packaging/Package.Default.props
jaredhms: how do i test these changes? are CI builds sufficient?

---
In reply to: [108995360](https://github.com/Microsoft/WinObjC/pull/2372#discussion_r108995360) [](ancestors = 108995360,108995058,108994710,108994382,108993288)

msvc/sdk-build.targets
bbowman: You aren't considering whats really going on here. Remember that this is in sdk-build.targets (and you are going to add something similar to the middleware targets). This needs to work generically. Things to consider:

1. You can't package a file that doesn't exist.

 Its imperative that when you add a PackageFile item, it exists. Currently you don't check at all if this exists. Its possible in a different scenario that on disk the pri file isn't at ProjectPriFullPath or that ProjectPriFullPath is defined but a Pri file isn't made. Perhaps in the scenario of a non AppContainerApplication project, no pri file is created.

2. The winmd file may not be located at %(RootDir)%(Directory)%(Filename).winmd

People can change where the winmd is output similar to changing their output directory. This combination is using metadata (the % ) off of each item from the expansion of _PriFileNames, so you are assuming that the pri file is *always* next to the winmd. That is true for UIKit.Xaml but is that true always? 

The precise condition that you are wanting to capture is "Only add a pri file if I have a pri file on disk and that this project produces a winmd"

the current condition is "Only add a pri file if there is a winmd file on disk in the same folder where I might write a pri file"

in the UIKit.xaml case this is equivalent but in the general case I would like to be more exact.

---
In reply to: [108994641](https://github.com/Microsoft/WinObjC/pull/2372#discussion_r108994641) [](ancestors = 108994641,108994152)

tools/WinObjC.Packaging/Package.Default.props
bbowman: Yeah probably. Talk to @memontic-msft about that.

---
In reply to: [108996733](https://github.com/Microsoft/WinObjC/pull/2372#discussion_r108996733) [](ancestors = 108996733,108995360,108995058,108994710,108994382,108993288)

bbowman:
:shipit:

PR #2375, opened 2017-03-30
REvert version string change for chocolatey since they don't support …

bbowman:
…prerelease strings starting with numbers

### DIFF
--- a/common/winobjc.nuproj.common.props
+++ b/common/winobjc.nuproj.common.props
@@ -12,7 +12,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <PackageVersion_PreReleaseFormat>{PackageVersion_Timestamp}-{PackageVersion_PreReleaseLabel}</PackageVersion_PreReleaseFormat>
+    <PackageVersion_PreReleaseFormat>{PackageVersion_PreReleaseLabel}-{PackageVersion_Timestamp}</PackageVersion_PreReleaseFormat>
   </PropertyGroup>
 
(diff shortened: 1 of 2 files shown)


pradipd:
Is this just for winobjc.tools?    Or is this all the packages?

bbowman:
Just tools

pradipd:
So we will have different prerelease string for tools vs other packages?


---
In reply to: [290313159](https://github.com/Microsoft/WinObjC/pull/2375#issuecomment-290313159) [](ancestors = 290313159)

bbowman:
@pradipd Yeah tools will be different. I can revert them all but that makes me more sad.

bbowman:
@pradipd / @jaredhms / @rajsesh-msft  / @memontic-ms would you prefer all of them to be reverted?

The pros / cons or reverting them all are as follows:

Pros:
1. They are all the same
2. You can use 0.2.170229-dev-* as a version string to say I want the latest pre-release package off of develop or 0.2.170229-pr-* to say I want the latest pre-release package off of a private branch

Cons:
1.  pr > dev > [0-9]+ meaning that in version strings like 0.2.170229-*, private branch packages will always be preferred over develop daily packages, etc. We've already pushed a package to nuget.org with -pr-### so its being preferred over our develop packages. If this happens again we need to wait for a stable package release to "reset" it. With the timestamp first, each day resets it so its more likely a daily build will be preferred.

pradipd:
:shipit:

PR #2381, opened 2017-03-30
Make packaging proj generation disabled by default

memontic-ms:
#### Previous behavior:
-genpackaging automatically enabled when:
1. Any number of non-app projects are included in the vsimporter import
2. -genprojections is specified

#### New behavior:
-genpackaging disabled by default
Must set -genpackaging or -genpackaging=1 to generate packaging (nuproj) project for any vsimporter import
i.e. If you are creating middleware you should specify -genprojections AND -genpackaging. It is no longer automatic.

I'll update the wiki's vsimporter page to reflect this change.

### DIFF
--- a/tools/vsimporter/src/vsimporter.cpp
+++ b/tools/vsimporter/src/vsimporter.cpp
@@ -111,7 +111,7 @@ int main(int argc, char* argv[]) {
     int workspaceSet = 0;
     int interactiveFlag = 0;
     int genProjectionsFlag = 0;
-    int genPackagingFlag = 1;
+    int genPackagingFlag = 0;
     int allTargets = 0;
     int allSchemes = 0;
     int mode = GenerateMode;


pradipd:
I'm fine with this change, but, I'm curious what is the reason for the change?

memontic-ms:
nuproj's, despite using project.json, do not automatically restore properly in VS 2017. Therefore, the nuproj will not build successfully without additional intervention (must command line nuget restore).   
Additionally, the packaging nuproj was added when ANY non-app was importer. For example, the case of importing an app plus a lib it depends on would add the packaging project to the solution.   
Therefore, to not break regular app importers, default has been set to OFF.

---
In reply to: [290533234](https://github.com/Microsoft/WinObjC/pull/2381#issuecomment-290533234) [](ancestors = 290533234)

bbowman:
Looks good.

pradipd:
I sign off, but codflow won't let me.

---
In reply to: [290541570](https://github.com/Microsoft/WinObjC/pull/2381#issuecomment-290541570) [](ancestors = 290541570)

PR #2382, opened 2017-03-30
Switch from projections to C++/WinRT

bviglietta:
This change switches some Framework code from using projections to using C++/WinRT when interacting with WinRT objects. This is beneficial because C++/WinRT is lighter-weight than the projections. They're header-only, so we don't have to link with the ObjCUWP_* binaries or load zillions of Objective-C classes.

In this change I've switched UIKit away from using the XAML projections, along with a couple of other smaller frameworks. Eventually none of the frameworks should be using projections, and the projections will exist solely for use by application code. I'll send mail to the team describing how to migrate away from projections and onto C++/WinRT.

This pull request is large. Sorry about that. I'll annotate certain parts of it that are especially interesting.


bviglietta:

Frameworks/include/CppWinRTHelpers.h
```diff
@@ -0,0 +1,139 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#pragma once
+
+#import <Foundation/Foundation.h>
+#import "StringHelpers.h"
+
```
bbowman: Is this file something we will want to redirstibute to people using our project that also want to use CppWinRT?

Frameworks/include/UITouchInternal.h
```diff
@@ -17,29 +17,31 @@
...
-    StrongId<WUXIPointerRoutedEventArgs> _routedEventArgs;
+    TrivialDefaultConstructor<winrt::Windows::UI::Xaml::Input::PointerRoutedEventArgs> _routedEventArgs;
     NSTimeInterval _timestamp;
     float _touchX, _touchY;
     float _previousTouchX, _previousTouchY;
     unsigned _tapCount;
     UITouchPhase _phase;
     StrongId<UIView> _view;
-    WUIPointerPoint* _pointerPoint; // Only valid during the processing of a pointer event; don't try to retain it.
+    TrivialDefaultConstructor<winrt::Windows::UI::Input::PointerPoint> _pointerPoint;
```
bbowman: >TrivialDefaultConstructor [](start = 4, length = 25)

I'm a little confused why this is needed. Is this because PointerPoint doesn't have a default ctor in the cpp winrt types?

Frameworks/include/CppWinRTHelpers.h
bviglietta: Clients are free to use C++/WinRT if they want to, but making it easier for them is not currently one of our goals.

Frameworks/include/UITouchInternal.h
bviglietta: That is a good question, because it's not very obvious.

C++/WinRT objects have a default constructor that _instantiates the object_. It is not like ComPtr or other smart types where the default constructor leaves the object uncreated.

C++/WinRT also have a constructor that takes nullptr_t that leaves the object initialized to null, but fields in Objective-C++ can only be initialized with the default constructor, so the nullptr_t constructor is not available. Therefore I've created TrivialDefaultConstructor<> that wraps a C++/WinRT object and forwards the default constructor to the nullptr_t constructor.

include/Platform/Universal Windows/UWP/InteropBase.h
```diff
@@ -26,19 +26,19 @@
 #include <inttypes.h>
 #include <wchar.h>
 
-#ifndef WINRT_EXPORT
```
bviglietta: >WINRT_EXPORT [](start = 8, length = 12)

This macro conflicts with one used in the C++/WinRT headers, so I renamed ours.

Frameworks/UIKit/UITextField.mm
```diff
@@ -116,13 +122,21 @@ @implementation UITextField {
...
-        _textBox.selectionLength = 0;
+        _textBox.Text(objcwinrt::string(text));
+
+        try {
+            // Ensure caret at end of field in case we programmatically
+            // gain focus (becomeFirstResponder) after the text is set:
+            _textBox.SelectionStart([text length]);
+            _textBox.SelectionLength(0);
+        } catch (const winrt::hresult_error& e) {
+            // Bug in XAML TextBox causes it to return S_FALSE
```
bviglietta: >Bug in XAML TextBox [](start = 15, length = 19)

Newer versions of the C++/WinRT headers have been changed to swallow these exceptions. Once we pick up those headers, we can remove this try/catch block.

tests/functionaltests/Tests/UIKitTests/UILabelTests.mm
```diff
@@ -299,29 +307,30 @@ void enumerateNumberOfLinesLinBreakFontSize(UILabelViewController* self) {
...
 
             label.numberOfLines = 7;
-            ASSERT_TRUE(textBlock.maxLines == 7);
+            ASSERT_TRUE(textBlock.MaxLines() == 7);
 
         });
     }
 
-    TEST_METHOD(UILabel_VerifyLinBreakMode) {
+#if 0  // Test temporarily broken
```
bviglietta: This test has always been broken. These changes have improved the test so that it is broken and failing instead of broken and passing. I'll open a separate issue to fix the test, and then we can re-enable it.

tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
```diff
@@ -17,11 +17,56 @@
...
+#import <winrt/Windows.UI.Xaml.Media.Imaging.h>
+#include "COMIncludes_End.h"
+
 using namespace UXTestAPI;
+using namespace winrt::Windows::UI::Xaml;
+
+// Callback object for the CurrentBackgroundImage test.
+// Normally a closure can be created with a lambda, but a compiler issue prevents that in
+// this case. Until it's fixed, use this object as a workaround.
+class LayoutUpdatedCallback
```
bviglietta: There's a compiler bug tracking this problem. Once the compiler is fixed, we can remove this workaround.

include/UIKit/UIView.h
```diff
@@ -379,11 +379,10 @@ UIKIT_EXPORT_CLASS
 @end
 
 // WinObjC Xaml Extensions
-@class WXFrameworkElement;
 @interface UIView (UIKitXamlExtensions)
-+ (WXFrameworkElement*)createXamlElement;
-@property (nonatomic, readonly, strong) WXFrameworkElement* xamlElement;
-- (instancetype)initWithFrame:(CGRect)frame xamlElement:(WXFrameworkElement*)xamlElement;
++ (RTObject*)createXamlElement;
```
bviglietta: We can't reference WXFrameworkElement directly if we don't want UIKit to have a dependency on the projection libraries. Instead we'll return a generic RTObject*.

Notice that this is a breaking change for clients. Not only is the return type different, but you cannot downcast from RTObject* to WXFrameworkElement*. The returned object is genuinely not a WXFrameworkElement. To get from the RTObject* to WXFrameworkElement*, clients will have to do a rt_dynamic_cast.

bviglietta: Code inside the frameworks should not call xamlElement to get the underlying XAML element. Instead, it should call _xamlElementInternal, declared in UIViewInternal.h.

The xamlElement property is intended for use only by application code. It will work inside the frameworks, but it's less efficient and harder to use.

bbowman:
:shipit:

DHowett-MSFT:
I am interested in seeing actual performance and disk space usage deltas for this change.

msvc/sdk-build.props
```diff
@@ -90,7 +92,7 @@
       <RuntimeLibrary>MultiThreadedDLL</RuntimeLibrary>
       <RuntimeLibrary Condition="'$(Configuration)'=='Debug'">MultiThreadedDebugDLL</RuntimeLibrary>
       <SystemHeaderDeps>true</SystemHeaderDeps>
-      <AdditionalOptions>-Werror -Wno-microsoft %(AdditionalOptions)</AdditionalOptions>
+      <AdditionalOptions>-Werror -Wno-microsoft -Wno-c++1z-extensions -Wno-ignored-attributes -Wno-deprecated-declarations %(AdditionalOptions)</AdditionalOptions>
```
DHowett-MSFT: I'm worried that adding `-Wno-deprecated-declarations` will come back to bite us, since we use deprecation internally to signify unimplementedness.

If it doesn't warn, we will lose compiler warnings (and therefore errors) when we try to use something we have ourselves not implemented.

bviglietta: The problem is that the C++/WinRT headers generate these warnings. In a near-future release those headers will be updated not to produce those warnings, but that fix is only going to work for VC++ and not for clang. They do support clang, but it's a second-class citizen.

One possible solution is to disable these warnings only around the C++/WinRT header inclusions. That would be cumbersome, except they're all already wrapped in COMIncludes.h/COMIncludes_End.h, so maybe an easy fix is to disable and re-enable warnings in those headers.

---
In reply to: [109239222](https://github.com/Microsoft/WinObjC/pull/2382#discussion_r109239222) [](ancestors = 109239222)

Frameworks/AVFoundation/AVAudioPlayer.mm
```diff
@@ -65,25 +70,25 @@ @implementation AVAudioPlayer {
...
+        _hiddenView = [[_UIHiddenMediaView alloc] initWithFrame:{ 0, 0, 0, 0 } xamlElement:objcwinrt::to_rtobj(_mediaElement)];
 
-        _lastState = _mediaElement.currentState;
+        _lastState = _mediaElement.CurrentState();
 
         // Set up events
         __weak AVAudioPlayer* weakSelf = self;
 
-        [_mediaElement addCurrentStateChangedEvent:^(RTObject* sender, WXRoutedEventArgs* e) {
+        _mediaElement.CurrentStateChanged([weakSelf] (auto&& sender, auto&& e) {
```
jaredhms: >auto&& sender, auto&& e [](start = 54, length = 23)

would prefer types on these...across a lot of these files.  our CCs say that we scope the use of auto to cases where declaring the type would be redundant.

Frameworks/include/CppWinRTHelpers.h
```diff
@@ -0,0 +1,139 @@
...
+
+//
+// Generate IReference<T> instances
+//
+
+template <typename T>
+winrt::Windows::Foundation::IReference<T> optional(const T& t) {
+    extern int not_implemented;
+    return not_implemented;
```
jaredhms: does this surface as a compile-time error, or as a linker error?

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -281,12 +290,13 @@ static void DoDisplayList(CALayer* layer) {
     _name = nil;
 
     // Create our backing layer proxy
-    _layerProxy = GetCACompositor()->CreateLayerProxy(xamlElement ? [xamlElement comObj] : nullptr);
+    auto rawElement = reinterpret_cast<ABI::Windows::UI::Xaml::IFrameworkElement*>(winrt::get_abi(xamlElement));
```
jaredhms: >auto rawElement = reinterpret_cast<ABI::Windows::UI::Xaml::IFrameworkElement*>(winrt::get_abi(xamlElement)); [](start = 4, length = 108)

why this vs. to_insp?

Frameworks/include/CppWinRTHelpers.h
```diff
@@ -0,0 +1,139 @@
...
+}
+
+template <typename T>
+T from_insp(IInspectable* insp) {
+    T retval = nullptr;
+    winrt::ABI::default_interface<winrt::abi<T>>* retvalAbi;
+
+    if (SUCCEEDED(insp->QueryInterface(IID_PPV_ARGS(&retvalAbi)))) {
+        winrt::attach_abi(retval, retvalAbi);
+    }
```
jaredhms: Does silently failing align with the old pattern?  For example, did '_xamlElement = [WXFrameworkElement createWith:inspectable.Get()]; ' throw on failure, or return nullptr?

Frameworks/include/UIViewInternal.h
```diff
@@ -146,7 +150,7 @@ class UIViewPrivateState : public LLTreeNode<UIViewPrivateState, UIView> {
 - (void)__setContentsImage:(id)image;
 - (UIWindow*)_getWindowInternal;
 - (BOOL)_isEnabled;
-
+- (winrt::Windows::UI::Xaml::FrameworkElement)_xamlElementInternal;
```
jaredhms: >_xamlElementInternal [](start = 46, length = 20)

maybe _winrtXamlElement for clarity?

Frameworks/UIKit/UIButton.mm
```diff
@@ -187,7 +189,7 @@ - (instancetype)initWithCoder:(NSCoder*)coder {
 
 - (void)_initUIButton {
     // Store a strongly-typed backing button
-    _xamlButton = rt_dynamic_cast<WXCButton>([self xamlElement]);
+    _xamlButton = [self _xamlElementInternal].try_as<Controls::Button>();
```
jaredhms: >try_as [](start = 46, length = 6)

would .as<> throw on failure?  if so, we can probably just do that and remove the FAIL_FAST below

Frameworks/UIKit/UIButton.mm
```diff
@@ -212,7 +214,9 @@ - (void)_initUIButton {
     _titleLabel.userInteractionEnabled = NO;
 
     // Create our child UIImageView; its frame will be updated in layoutSubviews
-    WXCImage* templateImage = rt_dynamic_cast([WXCImage class], [_xamlButton getTemplateChild:@"buttonImage"]);
+    auto control = _xamlButton.as<Controls::IControlProtected>();
+    auto templateImage = control.GetTemplateChild(L"buttonImage").as<Controls::Image>();
```
jaredhms: no longer needed as of yesterday

Frameworks/include/CppWinRTHelpers.h
bviglietta: Linker error.

---
In reply to: [109241603](https://github.com/Microsoft/WinObjC/pull/2382#discussion_r109241603) [](ancestors = 109241603)

Frameworks/QuartzCore/CALayer.mm
bviglietta: Because I wrote this before I invented to_insp. Will fix...

---
In reply to: [109241662](https://github.com/Microsoft/WinObjC/pull/2382#discussion_r109241662) [](ancestors = 109241662)

Frameworks/UIKit/UIButton.mm
bviglietta: Yes, as() throws on failure.

---
In reply to: [109241845](https://github.com/Microsoft/WinObjC/pull/2382#discussion_r109241845) [](ancestors = 109241845)

Frameworks/include/CppWinRTHelpers.h
jaredhms: a static_assert would fail at compile time; i think that would be preferable?

---
In reply to: [109242330](https://github.com/Microsoft/WinObjC/pull/2382#discussion_r109242330) [](ancestors = 109242330,109241603)

Frameworks/include/Starboard.h
```diff
@@ -205,4 +205,19 @@ inline CAPoint3D operator*(float f, const CAPoint3D& v) {
     return CAPoint3D(v.x * f, v.y * f, v.z * f);
 }
 
+// Wrapper for C++/WinRT types when you need a trivial default constructor
+template <typename T>
+class TrivialDefaultConstructor : public T {
+public:
+    TrivialDefaultConstructor() : T(nullptr) { }
+
+    auto operator=(const T& other) {
```
DHowett-MSFT: >) [](start = 33, length = 1)

should these be marked const/

DHowett-MSFT: sorry, I misread this as `operator==`. Ignore.

---
In reply to: [109243318](https://github.com/Microsoft/WinObjC/pull/2382#discussion_r109243318) [](ancestors = 109243318)

Frameworks/include/CppWinRTHelpers.h
bviglietta: createWith returns nil on failure.

I would argue that this provides a less silent failure than before. If this fails, the next thing that happens will be an access violation. It's not like Objective-C, where a nil pointer can go unnoticed for a long time.

---
In reply to: [109241704](https://github.com/Microsoft/WinObjC/pull/2382#discussion_r109241704) [](ancestors = 109241704)

Frameworks/include/CppWinRTHelpers.h
bviglietta: Sure, will do.

---
In reply to: [109242868](https://github.com/Microsoft/WinObjC/pull/2382#discussion_r109242868) [](ancestors = 109242868,109242330,109241603)

bviglietta: Actually, now that I think about it, this would be a compiler error, because it's trying to return an int instead of an IReference<T>. But regardless, a static_assert would be more explicit.

---
In reply to: [109244077](https://github.com/Microsoft/WinObjC/pull/2382#discussion_r109244077) [](ancestors = 109244077,109242868,109242330,109241603)

Frameworks/UIKit/UITextField.mm
jaredhms: can you file a tracking bug and reference it here?

---
In reply to: [109038960](https://github.com/Microsoft/WinObjC/pull/2382#discussion_r109038960) [](ancestors = 109038960)

Frameworks/UIKit/UITextField.mm
```diff
@@ -307,15 +330,15 @@ - (void)setTextColor:(UIColor*)color {
  @Status Interoperable
 */
 - (UIColor*)textColor {
-    WUXMSolidColorBrush* colorBrush = nil;
+    Media::SolidColorBrush colorBrush = nullptr;
     if (_secureTextMode) {
-        colorBrush = rt_dynamic_cast<WUXMSolidColorBrush>(_passwordBox.foreground);
+        colorBrush = _passwordBox.Foreground().try_as<Media::SolidColorBrush>();
```
jaredhms: >try_as [](start = 47, length = 6)

should these be .as?

Frameworks/UIKit/UITextField.mm
```diff
@@ -307,15 +330,15 @@ - (void)setTextColor:(UIColor*)color {
  @Status Interoperable
 */
 - (UIColor*)textColor {
-    WUXMSolidColorBrush* colorBrush = nil;
+    Media::SolidColorBrush colorBrush = nullptr;
```
jaredhms: > = nullptr [](start = 37, length = 10)

nit: is this necessary?

Frameworks/UIKit/UITextField.mm
```diff
@@ -1083,9 +1106,9 @@ - (void)setEnabled:(BOOL)enabled {
 */
 - (BOOL)isEnabled {
     if (self.secureTextEntry) {
-        return self->_passwordBox.isEnabled;
+        return self->_passwordBox.IsEnabled() ? YES : NO;
```
jaredhms: >self->_passwordBox.IsEnabled() [](start = 15, length = 30)

nit: static_cast<BOOL>(self->_passwordBox.IsEnabled())?

Frameworks/UIKit/UIView.mm
```diff
@@ -257,10 +265,15 @@ - (UITouchPhase)_processPointerEvent:(WUXIPointerRoutedEventArgs*)pointerEventAr
     }
 
     // Get the point value in this view's parent UIWindow's coordinates
-    WUIPointerPoint* pointerPoint = [pointerEventArgs getCurrentPoint:(owningWindow ? owningWindow.layer._xamlElement : nil)];
+    UIElement owningWindowElement = nullptr;
```
jaredhms: > = nullptr [](start = 33, length = 10)

nit: is this necessary? or does the default ctor init an object?

Frameworks/UIKit/UIView.mm
```diff
@@ -530,17 +543,18 @@ - (void)_initPrivWithFrame:(CGRect)frame xamlElement:(WXFrameworkElement*)xamlEl
...
 
     // If we don't have a backing xamlElement, create one for this UIView type
-    if (!xamlElement) {
+    if (xamlElement == nil) {
         xamlElement = [[self class] createXamlElement];
     }
 
     // Create and initialize our backing presentation layer
-    if (!xamlElement) {
+    if (xamlElement == nil) {
```
jaredhms: >xamlElement == nil [](start = 8, length = 18)

nit: why change these?

Frameworks/UIKit/UITextField.mm
bviglietta: Maybe. Using try_as just preserves the existing logic.

---
In reply to: [109244995](https://github.com/Microsoft/WinObjC/pull/2382#discussion_r109244995) [](ancestors = 109244995)

Frameworks/UIKit/UIView.mm
bviglietta: It is extremely necessary because yes, the default constructor actually creates an object.

... usually. I'm not sure that UIElement is actually activatable, so I'm not sure what happens there.

In any case, you usually want to say "winrt::SomeType myObj = nullptr;" rather than "winrt::SomeType myObj;"

---
In reply to: [109245698](https://github.com/Microsoft/WinObjC/pull/2382#discussion_r109245698) [](ancestors = 109245698)

Frameworks/UIKit/UIView.mm
bviglietta: No good reason. At one point I had changed the xamlElement to a C++/WinRT type, which is why I did this, but then I changed it again to RTObject*.

---
In reply to: [109245834](https://github.com/Microsoft/WinObjC/pull/2382#discussion_r109245834) [](ancestors = 109245834)

Frameworks/UIKit/XamlControls.h
```diff
@@ -26,118 +26,97 @@
...
-        _delegate(_createBareRTObj(arg0), _createRtProxy<WUXIPointerRoutedEventArgs>(arg1));
+        winrt::Windows::Foundation::IInspectable sender = nullptr;
+        winrt::copy_from_abi(sender, reinterpret_cast<winrt::ABI::Windows::Foundation::IInspectable*>(arg0));
 
+        winrt::Windows::UI::Xaml::Input::PointerRoutedEventArgs args = nullptr;
+        winrt::copy_from_abi(args, reinterpret_cast<winrt::ABI::Windows::UI::Xaml::Input::IPointerRoutedEventArgs*>(arg1));
+
+        NSAutoreleasePool* p = [NSAutoreleasePool new];
+        _handler(sender, args);
         [p release];
```
jaredhms: yikes; i hadn't thought about the implications of this.  we'll need to do this in every one of our event handlers; maybe create a special class that takes a lambda, implements the same API as said lambda, and wraps it in an AutoReleasePool push/pop?

bviglietta: That's a good catch. Yeah, we'll need to do that.

---
In reply to: [109247283](https://github.com/Microsoft/WinObjC/pull/2382#discussion_r109247283) [](ancestors = 109247283)

jaredhms: This may also be a good time to move to @autoreleasepool or lower-level push/pop calls vs. allocating an NSAutoReleasePool every time.

---
In reply to: [109247283](https://github.com/Microsoft/WinObjC/pull/2382#discussion_r109247283) [](ancestors = 109247283)

Frameworks/include/CppWinRTHelpers.h
aballway: Is there a reason we _wouldn't_ want to make these easily available?

include/UIKit/UIView.h
```diff
@@ -379,11 +379,10 @@ UIKIT_EXPORT_CLASS
...
+@property (nonatomic, readonly, strong) RTObject* xamlElement;
```
bviglietta: >RTObject* [](start = 40, length = 9)

This is a breaking change for our clients. Not only has the type of this property changed, but you can't downcast it to WXFrameworkElement* or WXCTextField* or whatever, because the object returned is genuinely not one of those things. You also can't just go ahead and invoke WXFrameworkElement methods on the object. You now have to use rt_dynamic_cast to do a downcast first.

jaredhms: that's a great point; can you please add a small comment that calls this out, and demonstrates how to rt_dynamic_cast to the desired projected type?

---
In reply to: [109249693](https://github.com/Microsoft/WinObjC/pull/2382#discussion_r109249693) [](ancestors = 109249693)

jaredhms:
:clock1:

Frameworks/include/CppWinRTHelpers.h
bviglietta: We've been presenting projections as the way for our clients to interact with WinRT. I wouldn't want to muddy the waters and start talking about C++/WinRT as an alternative without a very deliberate decision from team leads.

---
In reply to: [109248642](https://github.com/Microsoft/WinObjC/pull/2382#discussion_r109248642) [](ancestors = 109248642)

Frameworks/include/Starboard.h
```diff
@@ -205,4 +205,19 @@ inline CAPoint3D operator*(float f, const CAPoint3D& v) {
...
+        return T::operator=(other);
+    }
+
+    auto operator=(T&& other) {
+        return T::operator=(other);
```
bviglietta: >other [](start = 28, length = 5)

This should be std::forward<T>(other)

bviglietta:
Here are some preliminary numbers taken from the HelloUI sample app:

* AppX directory size reduced by 8.9 MB

* Peak working set reduced from 66.7 MB to 60.5 MB, a reduction of 9.2%

* App startup time (process launch to window focus) reduced from 286ms to 250ms, a reduction of 12.5%

Bear in mind that there's more work to be done here. I've mostly just removed the use of XAML projections, but there are others we can eliminate as well.

Also note that there's currently a bug in our packaging that causes all of the projection DLLs to be copied into the AppX package, regardless of whether they're used. That means that, until that bug is fixed, you won't see a reduction in AppX package size from this change. You will see the runtime effects, though.

---
In reply to: [290803157](https://github.com/Microsoft/WinObjC/pull/2382#issuecomment-290803157) [](ancestors = 290803157)

msvc/sdk-build.props
bviglietta: Aha, a better solution: clang will not generate warnings from headers that it considers "system" headers. So if we add --system-header-prefix=winrt/ to the command line, we get no more warnings from the C++/WinRT headers.

---
In reply to: [109240292](https://github.com/Microsoft/WinObjC/pull/2382#discussion_r109240292) [](ancestors = 109240292,109239222)

bbowman:
@bviglietta sounds like a feature to me ;-) I think after this change is applied across the board we will be putting Frameworks.UWP at a layer above/apart from Frameworks so that apps won't need to include projections at all unless they opt-in (Add a dependency themselves to it). I also have a task to be smarter about how we copy in but that problem is much harder with packaging and multiple middleware /languages potentially in the picture.

Frameworks/include/Starboard.h
```diff
@@ -205,4 +205,19 @@ inline CAPoint3D operator*(float f, const CAPoint3D& v) {
     return CAPoint3D(v.x * f, v.y * f, v.z * f);
 }
 
+// Wrapper for C++/WinRT types when you need a trivial default constructor
```
pradipd: >C++/WinRT types when you need a trivial default constructo [](start = 15, length = 58)

Out of curiosity, do you know why c++/winrt doesn't provide this?  Seems odd that clients would have to provide this.

bviglietta: C++/WinRT does provide default constructors, but they have behavior that can be surprising. The default constructor will instantiate the underlying object, unlike other smart pointer types like ComPtr or unique_ptr. This wrapper is provided for those cases where you need to have a default constructor that doesn't  instantiate the underlying object.

---
In reply to: [109485708](https://github.com/Microsoft/WinObjC/pull/2382#discussion_r109485708) [](ancestors = 109485708)

pradipd: [https://github.com/Microsoft/cppwinrt/blob/master/Docs/Migrating%20C%2B%2B%20CX%20source%20code%20to%20C%2B%2B%20WinRT.md](https://github.com/Microsoft/cppwinrt/blob/master/Docs/Migrating%20C%2B%2B%20CX%20source%20code%20to%20C%2B%2B%20WinRT.md)

I assume that is the topic covered in "Instantiating a C++/WinRT class".


---
In reply to: [109488370](https://github.com/Microsoft/WinObjC/pull/2382#discussion_r109488370) [](ancestors = 109488370,109485708)

Frameworks/UIKit/UITextField.mm
bviglietta: Actually it looks like the C++/WinRT headers from NuGet have this fix, so I've removed the try/catch entirely.

---
In reply to: [109244741](https://github.com/Microsoft/WinObjC/pull/2382#discussion_r109244741) [](ancestors = 109244741,109038960)

Frameworks/UIKit/XamlControls.h
bviglietta: I've added objcwinrt::callback, which wraps every callback in an autorelease pool. Try not to forget to use it!

---
In reply to: [109247664](https://github.com/Microsoft/WinObjC/pull/2382#discussion_r109247664) [](ancestors = 109247664,109247283)

Frameworks/AVFoundation/AVAudioPlayer.mm
```diff
@@ -382,7 +381,9 @@ - (NSTimeInterval)deviceCurrentTime {
  @Status Interoperable
 */
 - (NSTimeInterval)duration {
-    NSTimeInterval interval = (NSTimeInterval)((double)_mediaElement.naturalDuration.timeSpan.duration / c_durationCoef);
```
pradipd: >c_durationCoef [](start = 105, length = 14)

I think we can remove this constant.

Frameworks/AVFoundation/AVAudioPlayer.mm
```diff
@@ -160,27 +165,21 @@ - (instancetype)initWithData:(NSData*)data fileTypeHint:(NSString*)utiString err
...
+        // Hook the stream up to the media control when the load is complete
+        async.Completed([self] (auto&& operation, auto&& status) {
+            if (status == winrt::Windows::Foundation::AsyncStatus::Completed) {
+                // Access XAML control on main thread
+                dispatch_sync(dispatch_get_main_queue(), ^{
+                    self->_mediaElement.SetSource(self->_stream, L"");
+                    self->_stream = nullptr;
+                });
+            }
+        });
```
pradipd: How did the old way ever work?
Did we just not call it?
Either way this looks correct.

pradipd:
:shipit:

pradipd:
I mainly looked at the changes in avfoudnation.

Frameworks/AVFoundation/AVAudioPlayer.mm
```diff
@@ -160,27 +165,21 @@ - (instancetype)initWithData:(NSData*)data fileTypeHint:(NSString*)utiString err
...
 
-        writer->WriteBuffer(buffer.Get());
-        writer->StoreAsync(&comp);
-
-        [_mediaElement setSource:stream mimeType:@""];
-
-        // TODO: Is this needed?
-        [rw detachStream];
+        // Hook the stream up to the media control when the load is complete
+        async.Completed([self] (auto&& operation, auto&& status) {
```
pradipd: >auto [](start = 32, length = 4)

I suppose Jared's comment about auto applies to these typesas well.

pradipd: >self [](start = 25, length = 4)

Should this be weakself?

Frameworks/include/CppWinRTHelpers.h
```diff
@@ -155,4 +156,32 @@ winrt::Windows::Foundation::IReference<float> optional<float>(const float& f) {
...
+        // Make sure all callback operations are wrapped in an autorelease pool
+        template <typename... TArgs>
+        auto operator()(TArgs&& ...args) {
+            @autoreleasepool {
+                return _callback(std::forward<TArgs>(args)...);
+            }
+        }
+
+    private:
+        const TCallable& _callback;
```
DHowett-MSFT: Will this safely extend the lifetime of _callback? My lambda-fu is relatively weak, but I recall that lambdas are intended to be copied or moved, not captured by const reference. Especially when they close over locals.

DHowett-MSFT: (since the scope they close over will disappear shortly after they're registered via `objcwinrt::callback`.

bviglietta: Good catch. Capturing by reference is a mistake here.

Frameworks/UIKit/XamlControls.h
```diff
@@ -26,120 +26,99 @@
...
 };
+
 #endif // __has_feature(objc_arc)
 
 ////////////////////////////////////////////////////////////////////////////////////
 // Button
 ////////////////////////////////////////////////////////////////////////////////////
-WXCButton* CreateButton();
+winrt::Windows::UI::Xaml::Controls::Button CreateButton();
 WXFrameworkElement* GetButtonLabel(WXCButton* button);
```
bviglietta: This use of projections is a merge error.

PR #2399, opened 2017-03-31
Make NSConstantString hashes match up with NSString hashes

### LINKED ISSUE #2395: Non-ASCII NSConstantStrings hash differently than their NSString counterparts.
This is a regression from the merge of #2323.

Dictionary and set insertions lookups with constant strings and constructed strings with the same contents will act erratically.

## Code

```objc
TEST(NSString, ComparingDifferentTypes) {
    NSString* stringTwo = @"\u0CA0_\u0CA0";
    const unsigned char stringTwoUTF8[] = { 0xE0, 0xB2, 0xA0, '_', 0xE0, 0xB2, 0xA0, 0 };
    const unsigned char stringTwoUTF16[] = { 0xA0, 0x0C, '_', 0, 0xA0, 0x0C, 0, 0 };
    NSString* stringTwoUTF8String = [NSMutableString stringWithUTF8String:reinterpret_cast<const char*>(stringTwoUTF8)];
    NSString* stringTwoUTF16String = [NSMutableString stringWithCharacters:reinterpret_cast<const unichar*>(stringTwoUTF16) length:3];

    EXPECT_OBJCEQ(stringTwo, stringTwoUTF8String);
    EXPECT_OBJCEQ(stringTwo, stringTwoUTF16String);
    EXPECT_OBJCEQ(stringTwoUTF8String, stringTwoUTF16String);

    EXPECT_EQ(stringTwo.hash, stringTwoUTF8String.hash); // <- FAILING
    EXPECT_EQ(stringTwo.hash, stringTwoUTF16String.hash); // <- FAILING
    EXPECT_EQ(stringTwoUTF8String.hash, stringTwoUTF16String.hash);
}
```

## Test Result

```
StartGroup: NSString_ComparingDifferentTypes::Test
Error: Value of: stringTwoUTF8String.hash
  Actual: 2379796146
Expected: stringTwo.hash
Which is: 2616353738 [File: E:\src\WinObjC\tests\unittests\Foundation\NSStringTests.m, Function: void NSString_ComparingDifferentTypes::Test(), Line: 754]
Error: Value of: stringTwoUTF16String.hash
  Actual: 2379796146
Expected: stringTwo.hash
Which is: 2616353738 [File: E:\src\WinObjC\tests\unittests\Foundation\NSStringTests.m, Function: void NSString_ComparingDifferentTypes::Test(), Line: 755]
EndGroup: NSString_ComparingDifferentTypes::Test [Failed]
```

ms-jihua:
Fixes #2395

### DIFF
--- a/Frameworks/CoreFoundation/String.subproj/CFString.c
+++ b/Frameworks/CoreFoundation/String.subproj/CFString.c
@@ -1192,6 +1192,11 @@ CFHashCode __CFStringHash(CFTypeRef cf) {
     }
 }
 
+// WINOBJC: _CFStringHash is for Foundation's benefit. It should not be consumed by anyone else.
+CF_PRIVATE CF_EXPORT CFHashCode _CFStringHash(CFStringRef str) {
+    return __CFStringHash(str);
+}
+
 
 static CFStringRef __CFStringCopyDescription(CFTypeRef cf) {
     return CFStringCreateWithFormat(kCFAllocatorSystemDefault, NULL, CFSTR("<CFString %p [%p]>{contents = \"%@\"}"), cf, __CFGetAllocator(cf), cf);
--- a/Frameworks/Foundation/NSConstantString.mm
+++ b/Frameworks/Foundation/NSConstantString.mm
@@ -1,6 +1,6 @@
 //******************************************************************************
 //
-// Copyright (c) 2015 Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
 //
 // This code is licensed under the MIT License (MIT).
 //
@@ -14,6 +14,7 @@
 //
 //******************************************************************************
 #import "Starboard.h"
+#import "CFFoundationInternal.h"
 #import "ForFoundationOnly.h"
 #import "NSConstantString.h"
 #import "BridgeHelpers.h"
@@ -47,7 +48,7 @@
 
 // Override [NSString hash] and [NSString isEqualToString:] for faster perf
 - (NSUInteger)hash {
-    return CFStringHashCString(reinterpret_cast<uint8_t*>(c_string), len);
+    return _CFStringHash(__CFStringMakeConstantString(c_string));
 }
 
 - (BOOL)isEqualToString:(NSString*)str {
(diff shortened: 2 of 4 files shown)


ms-jihua:
image

Unfortunately regresses the perf to about 4x slower than develop :( On a minor bright side, still about 2x faster than before #2323. @DHowett-MSFT is looking at a larger-scale, longer-term fix, but we ought to fix the regression first.

Frameworks/CoreFoundation/String.subproj/CFString.c
```diff
@@ -1192,6 +1192,11 @@ CFHashCode __CFStringHash(CFTypeRef cf) {
     }
 }
 
+// WINOBJC: _CFStringHash is for Foundation's benefit. It should not be consumed by anyone else.
+CF_PRIVATE CF_EXPORT CFHashCode _CFStringHash(CFStringRef str) {
```
msft-Jeyaram: ForFoundationOnly.h? maybe?

ms-jihua: It's in the same file as, with the same type of comment as _CFStringIsMutable

ms-jihua:
ping!

rajsesh-msft:
What is the "longer term, larger scale fix?" Is there an issue tracking it?

Frameworks/CoreFoundation/String.subproj/CFString.c
msft-Jeyaram: nit: but does that mean it's the right place?

---
In reply to: [109267433](https://github.com/Microsoft/WinObjC/pull/2399#discussion_r109267433) [](ancestors = 109267433)

ms-jihua: ForFoundationOnly.h is a file from the swift open-source, we probably ought to mess with it as little as possible. CFFoundationInternal.h has been where we've been putting functions of this nature.

ms-jihua:
I don't remember the exact details from a brief discussion we had on Friday, but it had to do with messing with how we static init constant strings, so we don't have to do the indirection with CFMakeConstantString. I don't believe an issue has been filed yet.

PR #2402, opened 2017-04-01
Update projection binaries: Fixed bug with nsStrToHStr.

mukhole:
Binaries are generated using the nugetized projections.
This changes the names of libraries.

The problem is WindowsCreateString method returns a null pointer if the length of the string to create is 0. So an empty string will return a null object.


DHowett-MSFT:
Why did the export constants lose their `_`s?

pradipd:
+1 on Dustin's comment.  It is not clear why _ is missing.

mukhole:
I removed "_" because I was hitting command line string limit when I tried to restore packages.
A bug is filed on Nuget team.

The comment of getting a null string refers to nsStrToHstr method.

PR #2410, opened 2017-04-03
Update API status for CGColorGetConstantColor

msft-Jeyaram:


Frameworks/CoreGraphics/CGColor.mm
```diff
@@ -27,21 +27,17 @@
...
         ret = static_cast<CGColorRef>([[__LazyUIColor blackColor] CGColor]);
-        colorspace = woc::MakeStrongCF<CGColorSpaceRef>(CGColorSpaceCreateDeviceGray());
     } else if (CFEqual(kCGColorWhite, name)) {
         ret = static_cast<CGColorRef>([[__LazyUIColor whiteColor] CGColor]);
-        colorspace = woc::MakeStrongCF<CGColorSpaceRef>(CGColorSpaceCreateDeviceGray());
     } else if (CFEqual(kCGColorClear, name)) {
         ret = static_cast<CGColorRef>([[__LazyUIColor clearColor] CGColor]);
-        colorspace = woc::MakeStrongCF<CGColorSpaceRef>(CGColorSpaceCreateDeviceGray());
     } else {
         UNIMPLEMENTED_WITH_MSG("CGColorGetConstantColor does not support color %s", CFStringGetCStringPtr(name, kCFStringEncodingUTF8));
```
rajsesh-msft: Nit: Unknown constant color name should be what we trace here, also not through UNIMPLEMENTED macros, since we support all documented names.

PR #2418, opened 2017-04-04
[vsimporter] Don't add extra dependency to libs from Frameworks & Frameworks.Core

### LINKED ISSUE #2037: VSimporter probably should remove "known" frameworks from the AdditionalDependencies line
Actual:

`<AdditionalDependencies>%(AdditionalDependencies);AVFoundation.lib;AudioToolbox.lib;CoreGraphics.lib;Foundation.lib;GLKit.lib;OpenAL.lib;OpenGLES.lib;QuartzCore.lib;UIKit.lib</AdditionalDependencies>`

Expected:

Nothing.

memontic-ms:
#### Currently:
vsimporter adds frameworks (such as UIKit and Foundation) by including them under AdditionalDependencies (as UIKit.lib and Foundation.lib).

#### Proposed:
vsimporter skips adding AdditionalDependencies for any "known" frameworks (as listed in frameworks.txt).

_Why_: The consumption targets for WinObjC packages are adding the dependency instead (see [WinObjC.Frameworks.targets](https://github.com/Microsoft/WinObjC/blob/develop/build/WinObjC.Frameworks/WinObjC.Frameworks.targets) and [WinObjC.Frameworks.Core.targets](https://github.com/Microsoft/WinObjC/blob/develop/build/WinObjC.Frameworks.Core/WinObjC.Frameworks.Core.targets)). Therefore, imported projects only need the dependency on the WinObjC.Frameworks and WinObjC.Frameworks.Core packages themselves (which is handled in the project.json).

Yes, it's probably not optimal to have a static list of the frameworks in frameworks.txt. However, it would be over-engineering it a bit to do it another way and we might want to rethink this as a whole in the near future.

Please note: Leaving it as it currently is does not break anything (for now). However, it's a bit weird to have the libs included twice.

fixes #2037


memontic-ms:
@pradipd @bbowman

tools/WinObjC.Tools/vsimporter-templates/frameworks.txt
```diff
@@ -0,0 +1,61 @@
...
+QuartzCore.lib
+QuickLook.lib
+SafariServices.lib
+Security.lib
+Social.lib
+StoreKit.lib
+SystemConfiguration.lib
+Twitter.lib
+UIKit.lib
+UIKit.Xaml.lib
```
pradipd: It seems like this list should only contain iOS libraries as they are not likely to change.
uikit.xaml and starboard are libraries we created and can change.

memontic-ms: I agree but I was a bit unsure which ones to take out. Would it just be UIKit.Xaml.lib and Starboard.lib?

pradipd:
The change itself is fine.
However, I don't fully understand the motivation for it.

"Very low pri as we can add the linker search paths ourselves but since this in the vcxproj proper it makes things a little harder if we change our minds later."

Do we currently add the linker search paths somewhere?
If so, can we remove it now?

memontic-ms:
@pradipd see [WinObjC.Frameworks.targets](https://github.com/Microsoft/WinObjC/blob/develop/build/WinObjC.Frameworks/WinObjC.Frameworks.targets) and [WinObjC.Frameworks.Core.targets](https://github.com/Microsoft/WinObjC/blob/develop/build/WinObjC.Frameworks.Core/WinObjC.Frameworks.Core.targets)

```
<Link Condition="'$(ConfigurationType)' != 'StaticLibrary'">
    <AdditionalLibraryDirectories>$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\;$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\$(Configuration)\;$(MSBuildThisFileDirectory)\deps\prebuilt\$(TargetOsAndVersion)\$(PlatformTarget)\;$(MSBuildThisFileDirectory)\deps\prebuilt\$(TargetOsAndVersion)\$(PlatformTarget)\$(Configuration)\;%(AdditionalLibraryDirectories);</AdditionalLibraryDirectories>
</Link>
```
```
<_ImportLibsFromPackage Include="$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\*.lib"/>    
<_ImportLibsFromPackage Include="$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\$(Configuration)\*.lib"/>
...
<Link Include="@(_ImportLibsFromPackage)"/>
```
@bbowman, input?

pradipd:
So, is the issue that vsimporter does not  need to add these lib's since they are already included in the package? If so, what's the harm in keeping them? Extra link time?


---
In reply to: [291965047](https://github.com/Microsoft/WinObjC/pull/2418#issuecomment-291965047) [](ancestors = 291965047)

tools/WinObjC.Tools/vsimporter-templates/frameworks.txt
pradipd: Yes.

---
In reply to: [110000107](https://github.com/Microsoft/WinObjC/pull/2418#discussion_r110000107) [](ancestors = 110000107)

bbowman:
@memontic-ms I think its fine if we remove the search paths since we add it to Link directly and we are now removing duplicates from the link line when importing. The one concern would be if someone adds one back themselves and expects it to work. I don't see a need to remove the search path so i'd say leave it for now?

memontic-ms:
1. Correct.   
2. If they are left in: If someone goes and removes the lib from AdditionalDependencies in their vcxproj expecting the lib to not be used anymore, it still gets used.

---
In reply to: [291970098](https://github.com/Microsoft/WinObjC/pull/2418#issuecomment-291970098) [](ancestors = 291970098,291965047)

pradipd:
I think this is fine.  I don't really have the foresight to see any short comings.  

---
In reply to: [291970362](https://github.com/Microsoft/WinObjC/pull/2418#issuecomment-291970362) [](ancestors = 291970362)

pradipd:
:shipit:

memontic-ms:
@bbowman Leaving in the search paths for now.

PR #2419, opened 2017-04-04
Make SmartTypes and NSException (WinObjC) public for middleware

### LINKED ISSUE #2317: The generated code for middleware projects should not depend on internal headers
The generated code for middleware projects depends on starboard.h and included headers under:
Frameworks\include
include\xplat

pradipd:
Fixes #2317.

Changes:
1. SmartTypes.h and TypeTraits.h are public in the language package.    These were public in the old zip file but, are missing in the new packages.  I made them available in the language package.   I'm not sure this is the correct place.    I feel  CF* should be in foundation.core.  And, while AutoId could be in language, it does have dependency on Foundation (for nil....).  I spent some time trying to make this happen, but, it just wasn't worth the effort.
In the end, I took the easiest approach and just made SmartTypes.h public in language.

2. The NSException category in FoundationErrorHandling has been moved to NSException.  wil\result.h, which is a public header references FoundationErrorHandling.h.  So, we have to make FoundationErrorHandling.h public.  I decided to move all the category methods into NSException(winobjc).  Again, not sure if this is the best place.


DHowett-MSFT:
[updated pull request title to avoid truncation/improve clarity]

DHowett-MSFT:
I don't like exposing the `_` things, but it's more ABI-breaking to fix that.

pradipd:
The original PR title was supposed to be a teaser to make people click on it.  Just kidding.
thanks.



---
In reply to: [291669860](https://github.com/Microsoft/WinObjC/pull/2419#issuecomment-291669860) [](ancestors = 291669860)

pradipd:
Isn't the  only other thing we'd have to update is libobjc2?
I'm fine updating libobjc2 to remove the `_`.



---
In reply to: [291669943](https://github.com/Microsoft/WinObjC/pull/2419#issuecomment-291669943) [](ancestors = 291669943)

DHowett-MSFT:
@pradipd We'd need to rev the projections libraries again, as well as all modules using ErrorHandling (since they contain static copies of the error handling functions.) 😦 It would probably temporarily damage incremental build for anyone who pulled the changes down.

DHowett-MSFT:
@pradipd And thanks. I'm piloting talking to people about commit message guidelines, and I wanted to passive-aggressively try them on for size before I did. ;)

PR #2422, opened 2017-04-05
Redo Any CPU build multiplexing

bbowman:
The problem with the existing implementation is two fold:

1. MSBuild isn't as smart as I thought.
2. I'm not as smart as I thought.

more seriously the two issues were:

1. GetPackageContents was not correctly invoking ResolveReferences
2. Global property differences between .nuproj build multiplexing and test multiplexing
   was triggering incorrect rebuilds.

In the new mechanism, the build of all frameworks will be complete before GetPackageContents is invoked on that framework. This is because GetPackageContents *for .nuproj projects* now depends on ResolveReferences. This means that inside of GetPackageContents, no build resolution is needed so *for .vcxproj projects*, it does NOT depend on ResolveReference. This makes GetPackageContents a little more lightweight. Additionally, when multiplexing any "multiplex only" global properties are removed before invoking project reference builds to uniformly invoke them.

### DIFF
--- a/common/winobjc.nuproj.common.targets
+++ b/common/winobjc.nuproj.common.targets
@@ -37,7 +37,9 @@
   </Target>
 
   <!-- 
-    In the above AddSingleConfigBuildOutput target, several properties are injected into this project so that this project knows if its being multiplexed or not. Don't pass on this extra information to project references as it can issues with MSBuild knowing if a project is already built.
+    In the above AddSingleConfigBuildOutput target, several properties are injected into this project so that this project knows 
+    if its being multiplexed or not. Don't pass on this extra information to project references as it can cause issues with MSBuild 
+    knowing if a project is already built.
   -->
   <Target Name="RemoveGlobalPropertiesFromMultiplexing" BeforeTargets="BeforeResolveReferences">
     <PropertyGroup>
@@ -46,7 +48,7 @@
     </PropertyGroup>
     <ItemGroup>
       <ProjectReference>
-        <GlobalPropertiesToRemove>%(ProjectReference.GlobalPropertiesToRemove);$(_GlobalPropertiesToRemove);;</GlobalPropertiesToRemove>
+        <GlobalPropertiesToRemove>%(ProjectReference.GlobalPropertiesToRemove);$(_GlobalPropertiesToRemove);</GlobalPropertiesToRemove>
       </ProjectReference>
     </ItemGroup>
   </Target>
--- a/common/winobjc.nuproj.common.targets
+++ b/common/winobjc.nuproj.common.targets
@@ -38,20 +38,16 @@
 
   <!-- 
     In the above AddSingleConfigBuildOutput target, several properties are injected into this project so that this project knows if its being multiplexed or not. Don't pass on this extra information to project references as it can issues with MSBuild knowing if a project is already built.
-
-    Additionally, remove any .nuproj to .nuproj project references when going around the second time. This is because GetPackageContents will pass
-    through a BuildingPackage global property. This property will trigger a rebuild of the dependent nuproj's dependencies because of GetPackageContents
-    depending on ResolveReferences. This is all very confusing.
   -->
-  <Target Name="RemoveGlobalPropertiesFromMultiplexing" BeforeTargets="BeforeResolveReferences" Condition="'$(SkipGetPackageContentsForOtherPlatforms)' == 'true'">
+  <Target Name="RemoveGlobalPropertiesFromMultiplexing" BeforeTargets="BeforeResolveReferences">
+    <PropertyGroup>
+      <_GlobalPropertiesToRemove>BuildingPackage;</_GlobalPropertiesToRemove>
+      <_GlobalPropertiesToRemove Condition="'$(SkipGetPackageContentsForOtherPlatforms)' == 'true'">$(_GlobalPropertiesToRemove);SkipGetPackageContentsForOtherPlatforms;</_GlobalPropertiesToRemove>
+    </PropertyGroup>
     <ItemGroup>
       <ProjectReference>
-        <GlobalPropertiesToRemove>%(ProjectReference.GlobalPropertiesToRemove);SkipGetPackageContentsForOtherPlatforms;</GlobalPropertiesToRemove>
+        <GlobalPropertiesToRemove>%(ProjectReference.GlobalPropertiesToRemove);$(_GlobalPropertiesToRemove);;</GlobalPropertiesToRemove>
       </ProjectReference>
-
-      <!-- Don't evaluate nuproj to nuproj project references again in the multiplex'ed case. They were already evaluated before and nuprojs dependencies 
-           don't need multiplexed since its just ends up as a nuget dep. -->
-      <ProjectReference Remove="*.nuproj" />
     </ItemGroup>
   </Target>
 
--- a/common/winobjc.nuproj.common.targets
+++ b/common/winobjc.nuproj.common.targets
@@ -38,12 +38,20 @@
 
   <!-- 
     In the above AddSingleConfigBuildOutput target, several properties are injected into this project so that this project knows if its being multiplexed or not. Don't pass on this extra information to project references as it can issues with MSBuild knowing if a project is already built.
+
+    Additionally, remove any .nuproj to .nuproj project references when going around the second time. This is because GetPackageContents will pass
+    through a BuildingPackage global property. This property will trigger a rebuild of the dependent nuproj's dependencies because of GetPackageContents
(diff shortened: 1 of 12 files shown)


msvc/sdk-build.props
```diff
@@ -35,6 +35,7 @@
     <WindowsTargetPlatformMinVersion Condition="'$(WindowsTargetPlatformMinVersion)'==''">10.0.10586.0</WindowsTargetPlatformMinVersion>
     <EnableDotNetNativeCompatibleProfile Condition="'$(EnableDotNetNativeCompatibleProfile)'==''">true</EnableDotNetNativeCompatibleProfile>
     <IncludeOutputsInPackage>false</IncludeOutputsInPackage>
+    <IncludeFrameworkReferencesInPackage>false</IncludeFrameworkReferencesInPackage>
```
bbowman: This causes GetPackageContents to not depend on resolve references. See [here](https://github.com/NuGet/NuGet.Build.Packaging/blob/2fa6c67fd099ba08855c930de727f195d4ed1704/src/Build/NuGet.Build.Packaging.Tasks/NuGet.Build.Packaging.targets#L199-L201)

tools/WinObjC.Packaging/WinObjC.Packageable.targets
```diff
@@ -29,13 +29,8 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <GetPackageContentsDependsOn>$(GetPackageContentsDependsOn);$(ComputeLinkInputsTargets)</GetPackageContentsDependsOn>
```
bbowman: This needs to stay to make sure that import libs are included. I need to check what this guy depends on to make sure its not trying to resolve references to get a static lib link input or updated import lib.

bbowman:
Looks like parallel builds are still busted?

msvc/sdk-build.targets
```diff
@@ -34,19 +34,16 @@
...
-
   <Target Name="AddBuiltOutput" BeforeTargets="GetPackageContents" DependsOnTargets="$(GetPackageContentsDependsOn)" Condition="'$(ConfigurationType)' == 'DynamicLibrary' or '$(ConfigurationType)' == 'Application'">
     <ItemGroup>
       <!-- Public include folder for project  -->
       <_IncludeFiles Include="$(MSBuildThisFileDirectory)..\include\$(ProjectName)\**\*.*"/>
 
+      <!-- Since GetPackageContents runs in a second pass outside of the normal build, add a tmp Link item -->
+      <Link Include="tmp" Condition="'@(Link)'==''">
+       <DeleteSoon>true</DeleteSoon>
+      </Link>
```
pradipd: What is this linking?
I would have thought the link task is run in some other target.

bbowman: @pradipd I want to very precise when are talking about MSBuild stuff cause we all know how confusing the syntax and stuff is. This is adding a temp item to the Link itemgroup. Link, the task, is run in the Link target against items in the Link item group. This is used here to extract the ImportLibrary metadata off of the Link item group without needing to run the targets that could normally precede the Link target as those include resolving references / compiling the project itself which is already done at this point.

This can all mostly be avoided though (I think) if I figure out how to clear the BuildingPackage global property that .nuprojs add when calling GetPackageContents. I believe that was what was triggering a re-execution of already completed targets and still is for some cases.

tools/WinObjC.Packaging/WinObjC.Packager.targets
```diff
@@ -45,6 +40,17 @@
     </ItemGroup>
   </Target>
 
+  <!-- 
+    In the above AddSingleConfigBuildOutput target, several properties are injected into this project so that this project knows if its being multiplexed or not. Don't pass on this extra information to project references as it can issues with MSBuild knowing if a project is already built.
+  -->
```
memontic-ms: nit: make this comment multiple lines. I think a couple other places have overrun comments as well. 
also add a word: "[...] project references as it can _have_ issues with MSBuild [...]"

tools/common/common-build.props
```diff
@@ -11,6 +11,7 @@
     <WindowsTargetPlatformMinVersion Condition="'$(WindowsTargetPlatformMinVersion)'==''">10.0.10586.0</WindowsTargetPlatformMinVersion>
     <EnableDotNetNativeCompatibleProfile Condition="'$(EnableDotNetNativeCompatibleProfile)'==''">true</EnableDotNetNativeCompatibleProfile>
     <IncludeOutputsInPackage>false</IncludeOutputsInPackage>
+    <IncludeFrameworkReferencesInPackage>false</IncludeFrameworkReferencesInPackage>
```
memontic-ms: ><IncludeFrameworkReferencesInPackage>false</IncludeFrameworkReferencesInPackage> [](start = 4, length = 80)

Projects in tools.sln can importer common-build.props/targets, tools-build.props/targets, or neither. Does this needed to be added elsewhere as well?

bbowman: It looked like all the other ones in Tools already had this. common-build in tools.sln is for vcxproj

bbowman:
@memontic-ms @pradipd @rajsesh-msft and other reviewers, I think it is actually good to review now. I am spinning another CI build to confirm. Sorry for jumping the gun here.

tools/common/common-build.props
bbowman: (Also tools is a lot easier basically only Logging.dll would be a real issue here since most of the problem is coming from nuproj->nuproj->vcxproj project reference chains.)

@kzu also for FYI. Not sure if the BuildingPackage invoking a unique MSBuild which causes rebuilds would be a bug for Nugetizer in general.

common/winobjc.nuproj.common.targets
```diff
@@ -38,6 +36,21 @@
...
+    In the above AddSingleConfigBuildOutput target, several properties are injected into this project so that this project knows if its being multiplexed or not. Don't pass on this extra information to project references as it can issues with MSBuild knowing if a project is already built.
+  -->
+  <Target Name="RemoveGlobalPropertiesFromMultiplexing" BeforeTargets="BeforeResolveReferences">
+    <PropertyGroup>
+      <_GlobalPropertiesToRemove>BuildingPackage;</_GlobalPropertiesToRemove>
+      <_GlobalPropertiesToRemove Condition="'$(SkipGetPackageContentsForOtherPlatforms)' == 'true'">$(_GlobalPropertiesToRemove);SkipGetPackageContentsForOtherPlatforms;</_GlobalPropertiesToRemove>
+    </PropertyGroup>
+    <ItemGroup>
+      <ProjectReference>
+        <GlobalPropertiesToRemove>%(ProjectReference.GlobalPropertiesToRemove);$(_GlobalPropertiesToRemove);;</GlobalPropertiesToRemove>
```
memontic-ms: >; [](start = 108, length = 1)

nit: extra ;

tools/WinObjC.Packaging/WinObjC.Packager.targets
```diff
@@ -45,6 +40,19 @@
...
+  <Target Name="RemoveGlobalPropertiesFromMultiplexing" BeforeTargets="BeforeResolveReferences">
+    <_GlobalPropertiesToRemove>BuildingPackage;</_GlobalPropertiesToRemove>
+    <_GlobalPropertiesToRemove Condition="'$(SkipGetPackageContentsForOtherPlatforms)' == 'true'">$(_GlobalPropertiesToRemove);SkipGetPackageContentsForOtherPlatforms;IsPackable;</_GlobalPropertiesToRemove>
```
memontic-ms: >IsPackable; [](start = 167, length = 11)

Why does WinObjC.Packager get IsPackable here but winobjc.nuproj.common doesn't?

msvc/sdk-build.targets
pradipd: Thanks for the explanation.
I still prefer tmp to be renamed "Zelda", but....

---
In reply to: [109974322](https://github.com/Microsoft/WinObjC/pull/2422#discussion_r109974322) [](ancestors = 109974322)

pradipd:
:shipit:

tools/WinObjC.Packaging/WinObjC.Packager.targets
bbowman: Because you wrote Packager and I wrote common? You pass through IsPackable on line #22 whereas I keyed off of the SkipGetPackageContentsForOtherPlatforms to set it in a props. Its not a big deal either way we just need to account for it.

---
In reply to: [110029872](https://github.com/Microsoft/WinObjC/pull/2422#discussion_r110029872) [](ancestors = 110029872)

PR #2426, opened 2017-04-05
Complete implementation of NSSet

### LINKED ISSUE #2080: Implement NSSet KVO methods
addObserver:forKeyPath:options:context:
removeObserver:forKeyPath:
removeObserver:forKeyPath:context:

### LINKED ISSUE #2081: finish NSSet stubs
makeObjectsPerformSelector:withObject:
objectsWithOptions:passingTest:
descriptionWithLocale:

aballway:
Adds implementations for several stubbed methods, exceptions for KVO methods, as well as some small cleanup.

Fixes #2080
Fixes #2081

### DIFF
--- a/Frameworks/Foundation/NSSet.mm
+++ b/Frameworks/Foundation/NSSet.mm
@@ -519,7 +519,7 @@ BASE_CLASS_REQUIRED_IMPLS(NSSet, NSSetPrototype, CFSetGetTypeID);
  @Status Interoperable
 */
 - (NSSet*)objectsWithOptions:(NSEnumerationOptions)opts passingTest:(BOOL (^)(id, BOOL*))predicate {
-    __block NSMutableSet* ret = [NSMutableSet setWithCapacity:0];
+    __block NSMutableSet* ret = [NSMutableSet setWithCapacity:[self count]];
 
     [self enumerateObjectsWithOptions:opts
                            usingBlock:^void(id obj, BOOL* stop) {
@@ -589,7 +589,7 @@ BASE_CLASS_REQUIRED_IMPLS(NSSet, NSSetPrototype, CFSetGetTypeID);
 - (NSString*)descriptionWithLocale:(id)locale indent:(NSUInteger)level {
     NSMutableString* s = [NSMutableString string];
     NSString* indentStr = @"    ";
-    for (unsigned int i = 0; i < level; ++i) {
+    for (NSUInteger i = 0; i < level; ++i) {
         [s appendString:indentStr];
     }
 
--- a/Frameworks/Foundation/NSSet.mm
+++ b/Frameworks/Foundation/NSSet.mm
@@ -1,6 +1,6 @@
 //******************************************************************************
 //
-// Copyright (c) 2015 Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
 //
 // This code is licensed under the MIT License (MIT).
 //
@@ -27,7 +27,13 @@
 #include "Starboard.h"
 #include "StubReturn.h"
 #include "VAListHelper.h"
+#import "NSKeyValueObserving-Internal.h"
 #import <_NSKeyValueCodingAggregateFunctions.h>
+
+@interface NSSet ()
+// Public interface for NSSet does not have this method, but is necessary for proper formatting with nested collections
+- (NSString*)descriptionWithLocale:(id)locale indent:(NSUInteger)level;
+@end
 
 @implementation NSSet
 
@@ -98,7 +104,7 @@ BASE_CLASS_REQUIRED_IMPLS(NSSet, NSSetPrototype, CFSetGetTypeID);
 /**
  @Status Interoperable
 */
-- (instancetype)initWithObjects:(id*)objects count:(unsigned)count {
+- (instancetype)initWithObjects:(id _Nonnull const*)objects count:(NSUInteger)count {
     // Derived classes are required to implement this initializer.
     return NSInvalidAbstractInvocationReturn();
 }
@@ -276,7 +282,7 @@ BASE_CLASS_REQUIRED_IMPLS(NSSet, NSSetPrototype, CFSetGetTypeID);
 /**
  @Status Interoperable
 */
-- (unsigned)countByEnumeratingWithState:(NSFastEnumerationState*)state objects:(id*)stackBuf count:(unsigned)maxCount {
+- (NSUInteger)countByEnumeratingWithState:(NSFastEnumerationState*)state objects:(id _Nullable*)stackBuf count:(NSUInteger)maxCount {
     if (state->state == 0) {
         state->mutationsPtr = (unsigned long*)&state->extra[1];
         state->extra[0] = (unsigned long)[self objectEnumerator];
@@ -388,23 +394,38 @@ BASE_CLASS_REQUIRED_IMPLS(NSSet, NSSetPrototype, CFSetGetTypeID);
 }
 
 /**
- @Status Caveat
- @Notes options parameter not supported
+ @Status Interoperable
+ @Notes NSEnumerationReverse is undefined on reference platform so we will ignore it
 */
 - (void)enumerateObjectsWithOptions:(NSEnumerationOptions)options usingBlock:(void (^)(id, BOOL*))block {
-    if (options & NSEnumerationReverse) {
-        assert(0);
-    } else {
-        BOOL stop = FALSE;
-        int i = 0;
+    dispatch_queue_t queue;
+    dispatch_group_t group;
 
-        for (id curObj in self) {
+    // Initialize dispatch queue for concurrent enumeration
+    if (options & NSEnumerationConcurrent) {
+        queue = dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0);
+        group = dispatch_group_create();
+    }
+
+    __block BOOL stop = NO;
+    for (id curObj in self) {
+        if (options & NSEnumerationConcurrent) {
+            dispatch_group_async(group, queue, ^() {
+                block(curObj, &stop);
+            });
+        } else {
             block(curObj, &stop);
-            i++;
-            if (stop) {
-                break;
-            }
         }
+
+        if (stop) {
+            break;
+        }
+    }
+
+    // Wait for dispatch queue to execute
+    if (options & NSEnumerationConcurrent) {
+        dispatch_group_wait(group, DISPATCH_TIME_FOREVER);
+        dispatch_release(group);
     }
 }
 
@@ -450,12 +471,10 @@ BASE_CLASS_REQUIRED_IMPLS(NSSet, NSSetPrototype, CFSetGetTypeID);
 }
 
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 + (BOOL)supportsSecureCoding {
-    UNIMPLEMENTED();
-    return StubReturn();
+    return YES;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


Frameworks/Foundation/NSSet.mm
```diff
@@ -554,12 +573,93 @@ - (void)setValue:(id)value forKey:(NSString*)key {
...
+}
+
+/**
+ @Status Interoperable
+*/
+- (NSString*)description {
+    return [self descriptionWithLocale:nil indent:0];
+}
+
+- (NSString*)descriptionWithLocale:(id)locale indent:(NSUInteger)level {
```
aballway: Unfortunately this can't be implemented by wrapping NSArray's descriptionWithLocale:indent: in braces because levels will cause spaces to appear between the first brace "{" and the array's parenthesis "("

Frameworks/Foundation/NSSet.mm
```diff
@@ -27,8 +27,14 @@
 #include "Starboard.h"
 #include "StubReturn.h"
 #include "VAListHelper.h"
+#import "NSKeyValueObserving-Internal.h"
 #import <_NSKeyValueCodingAggregateFunctions.h>
 
+@interface NSSet ()
+// Public interface for NSSet does not have this method, but is necessary for proper formatting with nested collections
+- (NSString*)descriptionWithLocale:(id)locale indent:(NSUInteger)level;
```
DHowett-MSFT: You don't really need to declare this; internal consumers can use it because it is in the `@implementation`.

Frameworks/Foundation/NSSet.mm
```diff
@@ -388,24 +394,39 @@ - (void)enumerateObjectsUsingBlock:(void (^)(id, BOOL*))block {
...
+    // Initialize dispatch queue for concurrent enumeration
+    if (options & NSEnumerationConcurrent) {
+        queue = dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0);
+        group = dispatch_group_create();
+    }
+
+    __block BOOL stop = NO;
+    for (id curObj in self) {
+        if (options & NSEnumerationConcurrent) {
+            dispatch_group_async(group, queue, ^() {
```
DHowett-MSFT: how does stop work for concurrent enumeration? o_O

Frameworks/Foundation/NSSet.mm
```diff
@@ -481,35 +500,35 @@ - (NSSet*)filteredSetUsingPredicate:(NSPredicate*)predicate {
...
 
 /**
- @Status Stub
- @Notes
+ @Status Interoperable
 */
 - (NSSet*)objectsWithOptions:(NSEnumerationOptions)opts passingTest:(BOOL (^)(id, BOOL*))predicate {
-    UNIMPLEMENTED();
-    return StubReturn();
+    __block NSMutableSet* ret = [NSMutableSet setWithCapacity:0];
```
DHowett-MSFT: nit: A more pessimistic capacity hint would be "all of our objects"; consider?

tests/unittests/Foundation/NSSetTests.mm
```diff
@@ -148,4 +148,98 @@
     NSMutableSet* set = [NSMutableSet set];
     EXPECT_ANY_THROW([set addObject:nil]);
     EXPECT_EQ(0, [set count]);
-}
\ No newline at end of file
+}
+
+// Cannot be certain of order of elements so cannot compare directly with reference platform
```
DHowett-MSFT: we cannot be certain of the order of elements on our platform either.

DHowett-MSFT: this test _will_ spontaneously fail at some point in the future.

tests/unittests/Foundation/NSSetTests.mm
```diff
@@ -148,4 +148,98 @@
...
+
+TEST(NSSet, ObjectsWithOptionsPassingTest) {
+    NSSet* set = [NSSet setWithObjects:@1, @2, @3, @4, @5, nil];
+
+    // Verify that the NSEnumerationConcurrent option executes the blocks concurrently
+    __block NSCondition* condition = [[NSCondition new] autorelease];
+    __block unsigned waitingCount = 0;
+    NSSet* matchingForward = [set objectsWithOptions:NSEnumerationConcurrent
+                                         passingTest:^BOOL(id obj, BOOL* stop) {
+                                             [condition lock];
```
DHowett-MSFT: (can you explain the logic in here?)

tests/unittests/Foundation/NSSetTests.mm
aballway: Would it be reasonable then to leave this method untested?

Frameworks/Foundation/NSSet.mm
aballway: I'm unsure, logic was heavily borrowed from NSArray.  Given ordering & concurrency aren't determinable, I imagine it must be undefined

Frameworks/Foundation/NSSet.mm
aballway: Is there any downside to it? I rather like it 😺

tests/unittests/Foundation/NSSetTests.mm
```diff
@@ -148,4 +148,98 @@
...
+                                             [condition wait];
+                                             waitingCount--;
+                                             [condition signal];
+                                             [condition unlock];
+                                             return [obj unsignedIntegerValue] % 2 == 0 ? YES : NO;
+                                         }];
+
+    ASSERT_EQ(0, waitingCount);
+
+    NSSet* matchingReverse = [set objectsWithOptions:(NSEnumerationConcurrent | NSEnumerationReverse)
```
MSFTFox: Personally, I don't see value in testing the behavior of undefined options. This test could fail in the future and still be an acceptable implementation, right?

tests/unittests/Foundation/NSSetTests.mm
```diff
@@ -148,4 +148,98 @@
...
+                          ")}";
+    EXPECT_OBJCEQ(expected, nestedSet.description);
+}
+
+TEST(NSSet, MakeObjectsPerformSelector) {
+    NSSet* set = [NSSet setWithObjects:@[ @1, @2, @3 ], [NSSet setWithObjects:@4, @5, nil], @[ @6 ], nil];
+    __block NSMutableSet* otherSet = [NSMutableSet set];
+    [set makeObjectsPerformSelector:@selector(enumerateObjectsUsingBlock:)
+                         withObject:^(id obj, BOOL* stop) {
+                             [otherSet addObject:obj];
```
MSFTFox: I'm not sure this is actually testing the intended effect of makeObjectsPerformSelector. This should be testing that the objects in the set respond appropriately to the message in the block, right? Instead, this is just enumerateWithBlock where the block variable, otherSet, is adding each enumerated object.

Frameworks/Foundation/NSSet.mm
MSFTFox: Would it make sense to go the other way then and modify NSArray's implementation? Or possibly break this out into a partially shared implementation since so much of the code is similar? Just a couple thoughts.

tests/unittests/Foundation/NSSetTests.mm
aballway: No this is making the elements of the set, which are sets and arrays, perform `enumerateObjectsUsingBlock` with the block as the `withObject` object.

tests/unittests/Foundation/NSSetTests.mm
aballway: It's not testing the undefined behavior since it's inserting everything into a set, just that it enumerates over the entire set.

Frameworks/Foundation/NSSet.mm
aballway: The reference platform will stop enumerating with NSEnumerationConcurrent when stop is set to `YES` but because of the asynchronicity we can't be sure if it stops any others that are currently running (if there's even a way to do that).  Given that, I think we should go with the implementation that's consistent w/ NSArray

Frameworks/Foundation/NSSet.mm
aballway: There's no real way to keep NSArray's description interoperable while supporting this, and I don't think the gains of cutting down code duplication would be worth tying these together.  I'm flexible on this if we think it's the right move.

tests/unittests/Foundation/NSSetTests.mm
MSFTFox: Yup, you're right.

tests/unittests/Foundation/NSSetTests.mm
MSFTFox: I see what you're saying. You're just using the reverse enumeration as another set to match against. I still think it shouldn't be used though, since it is still technically undefined.

Frameworks/Foundation/NSSet.mm
MSFTFox: Well it's worth thinking about that there's a handful of classes that each have a similar API. Like NSOrderedSet, NSDictionary, etc... They're unimplemented at the moment but if each of these produces a similar result then it might be useful to consider. I'm not saying to do it right now, but as we fill out the implementations of Foundation we may have to revisit this.

Frameworks/Foundation/NSSet.mm
MSFTFox: I agree, it's probably the right thing. I was just curious since the documentation mentions this in NSArray but not NSSet.

aballway:
Ping!

Frameworks/Foundation/NSSet.mm
ms-jihua: seconded: doesn't have to be right now, but moving some of this logic into a shared helper is starting to look very tempting

tests/unittests/Foundation/NSSetTests.mm
ms-jihua: Isn't this comment no longer the case, now that the test no longer depends on ordering?

tests/unittests/Foundation/NSSetTests.mm
```diff
@@ -148,4 +148,81 @@
     NSMutableSet* set = [NSMutableSet set];
     EXPECT_ANY_THROW([set addObject:nil]);
     EXPECT_EQ(0, [set count]);
-}
\ No newline at end of file
+}
+
+// Cannot be certain of order of elements so cannot compare directly with reference platform
+// The reference platform also has some unexpected behaviours when nesting collections, so we have opted for consistency
```
ms-jihua: What's the remaining unexpected behavior?

aballway: NSSets nested within other containers will not return expected descriptions, it will contain escaped escape sequences (like literal \n)

tests/unittests/Foundation/NSSetTests.mm
aballway: It's still an explanation of _why_ we're not doing ordering-dependent tests

ms-jihua: Sure, but maybe edit the last part of the comment then.

tests/unittests/Foundation/NSSetTests.mm
ms-jihua: Eesh. Probably call this out specifically in a comment?

Frameworks/Foundation/NSSet.mm
aballway: I'm working on this w/ a few other changes in NSDictionary so I'll clean these all up in a separate PR once I've got them working as expected w/ tests n everything

PR #2430, opened 2017-04-05
Unify all WinObjC Versioning

bbowman:
This changes simplifies and unifies all versioning to be based on Git Tags.
This removes the need for the version file and no longer requires master packages
to be built with a version override (though they still can).

vcxproj, csproj, and VSIX projects all get automatically versioned in a concsistent
way and have more metadata like the Copyright and description correctly filled out.


common/winobjc.versioning.common.targets
```diff
@@ -0,0 +1,208 @@
...
+      <PackageVersion>$(PackageVersion_Major).$(PackageVersion_Minor).$(PackageVersion_Patch)$(PackageVersion_PreRelease)</PackageVersion>
+    </PropertyGroup>
+
+    <PropertyGroup Condition="'$(PackageVersionOverride)' != ''">
+
+      <!-- 
+      Regex for package version override string in escaped .NET syntax:
+      ^v?(?<MAJOR>\d+)\.(?<MINOR>\d+)\.(?<PATCH>\d+)(\-(?<LABEL&>[\dA-Za-z\-\.]+))?$
+       -->
+      <_PackageVersionRegex>^v?(?&lt;MAJOR&gt;\d+)\.(?&lt;MINOR&gt;\d+)\.(?&lt;PATCH&gt;\d+)(\-(?&lt;LABEL&gt;[\dA-Za-z\-\.]+))?$</_PackageVersionRegex>
```
memontic-ms: Yep. That's a fun regex string alright. 👏

common/winobjc.versioning.common.targets
```diff
@@ -0,0 +1,245 @@
...
+          Condition="'$(CreateVsixContainer)' == 'true'">
+
+    <XmlPoke XmlInputPath="$(IntermediateVsixManifest)"
+             Namespaces="&lt;Namespace Prefix='vsix' Uri='http://schemas.microsoft.com/developer/vsx-schema/2011' /&gt;"
+             Query="/vsix:PackageManifest/vsix:Metadata/vsix:Identity/@Version"
+             Value="$(PackageResourceVersion)"/>
+  </Target>
+
+
+  <!--
```
memontic-ms: Will this overwrite whatever AssemblyInfo.cs file is already included in a csproj? Or just add on to it?

common/winobjc.versioning.common.targets
```diff
@@ -0,0 +1,245 @@
...
+      <PackageResourceVersion_Build>$(PackageVersion_Patch.Substring(0,4))</PackageResourceVersion_Build>
+      <PackageResourceVersion_Revision>$(PackageVersion_Patch.Substring(4))</PackageResourceVersion_Revision>
+    </PropertyGroup>
+
+    <PropertyGroup>
+      <PackageResourceVersion_Build Condition="'$(PackageResourceVersion_Build)' == ''">$(PackageVersion_Patch)</PackageResourceVersion_Build>
+      <PackageResourceVersion_Revision Condition="'$(PackageResourceVersion_Revision)' == ''">0</PackageResourceVersion_Revision>
+
+      <PackageResourceVersion>$(PackageVersion_Major).$(PackageVersion_Minor).$(PackageResourceVersion_Build).$(PackageResourceVersion_Revision)</PackageResourceVersion>
+    </PropertyGroup>
```
memontic-ms: Should these two targets be more.. nested? Are we guaranteed that the first PropertyGroup would get run first? As in, could PackageResourceVersion get set with PackageResourceVersion_Build and PackageResourceVersion_Revision before the first group sets them?

common/winobjc.versioning.common.targets
```diff
@@ -0,0 +1,245 @@
...
+    <PropertyGroup Condition="$([System.UInt64]::Parse($(PackageVersion_Patch))) &gt; $([MSBuild]::Subtract($([System.UInt16]::MaxValue), '1'))">
+      <PackageResourceVersion_Build>$(PackageVersion_Patch.Substring(0,4))</PackageResourceVersion_Build>
+      <PackageResourceVersion_Revision>$(PackageVersion_Patch.Substring(4))</PackageResourceVersion_Revision>
+    </PropertyGroup>
+
+    <PropertyGroup>
+      <PackageResourceVersion_Build Condition="'$(PackageResourceVersion_Build)' == ''">$(PackageVersion_Patch)</PackageResourceVersion_Build>
+      <PackageResourceVersion_Revision Condition="'$(PackageResourceVersion_Revision)' == ''">0</PackageResourceVersion_Revision>
+
+      <PackageResourceVersion>$(PackageVersion_Major).$(PackageVersion_Minor).$(PackageResourceVersion_Build).$(PackageResourceVersion_Revision)</PackageResourceVersion>
```
memontic-ms: Can we get a bit more explanation of the intended uses for PackageResourceVersion_Build, PackageResourceVersion_Revision, and PackageResourceVersion?

common/winobjc.versioning.common.targets
```diff
@@ -0,0 +1,245 @@
...
+
+                if (!File.Exists(AssemblyInfoFileFullPath) || 
+                    !tempAssemblyInfo.Equals(File.ReadAllText(AssemblyInfoFileFullPath).Trim()))
+                {
+                  File.WriteAllText(AssemblyInfoFileFullPath, tempAssemblyInfo);
+                }
+            }
+            catch (Exception)
+            {
+            }
```
memontic-ms: nit: standardize on spacing/indentation
Also, nothing in the catch?

common/winobjc.versioning.common.targets
```diff
@@ -0,0 +1,245 @@
...
+    This target will inject an AssemblyInfo.cs file into any csproj project that gets built.
+    This is a little cumbersome so an inline task to do the file management (to help avoid rebuilds)
+    is used. All metadata in the AssemblyInfo is still controllable with MSBuild properties, however.
+  -->
+  <UsingTask  
+    TaskName="WriteAssemblyInfoFile"
+    TaskFactory="CodeTaskFactory"  
+    AssemblyFile="$(MSBuildToolsPath)\Microsoft.Build.Tasks.v12.0.dll" >
+    <ParameterGroup>
+      <AssemblyInfoFileFullPath  ParameterType="System.String" Required="true" />
```
memontic-ms: >  [](start = 32, length = 1)

nitty nit nit since already looking at this section: extra space here

tools/legacy/ObjectiveC/Properties/AssemblyInfo.cs
```diff
@@ -21,29 +21,11 @@
...
-//      Minor Version 
-//      Build Number
-//      Revision
-//
-// You can specify all the values or you can default the Build and Revision Numbers 
-// by using the '*' as shown below:
-// [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyVersion("1.0.0.0")]
-[assembly: AssemblyFileVersion("1.0.0.0")]
+[assembly: ComVisible(false)]
```
kellymor: Is this change necessary?  I don't think it will negatively impact the legacy VSIX but, we might want to be sure this doesn't prevent install/uninstall and functionality of the VSIX.

bbowman:
@kellymor what is the concern exactly? I don't follow.

common/winobjc.versioning.common.targets
```diff
@@ -0,0 +1,245 @@
...
+
+    <WriteAssemblyInfoFile 
+      AssemblyInfoFileFullPath="$(TempAssemblyInfoFileFullPath)"
+      AssemblyVersion="$(PackageResourceVersion)"
+      AssemblyFileVersion="$(PackageResourceVersion)"
+      AssemblyInformationalVersion="$(PackageVersion)+$(GitVersion_FullBuildMetaData)"
+      AssemblyTitle="$(TargetFileName)"
+      AssemblyDescription="https://github.com/Microsoft/WinObjC"
+      AssemblyCompany="Microsoft"
+      AssemblyCopyright="Copyright (C) Microsoft Corporation. All rights reserved."
```
rajsesh-msft: >Corporation [](start = 49, length = 11)

Copyright (C) Microsoft. All rights reserved. (drop Corporation).

memontic-ms:
>             VALUE "ProductName", "Objective-C for Windows v" STRINGIZE(PACKAGE_VERSION)

Is the version usually appended to the product name considering there is also a ProductVersion field?    
Here it doesn't say you should _only_ have the name but it also doesn't say you _should_ have the version: [https://msdn.microsoft.com/en-us/library/windows/desktop/aa381058(v=vs.85).aspx](https://msdn.microsoft.com/en-us/library/windows/desktop/aa381058(v=vs.85).aspx)

---
Refers to: common/common.rc:171 in f4bbf31. [](commit_id = f4bbf31cb90a64512a01760d3c22bc259181321c, deletion_comment = False)

common/winobjc.versioning.common.targets
```diff
@@ -0,0 +1,245 @@
...
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup>
+    <CalculatePackageVersionDependsOn>GetVersion;$(CalculatePackageVersionDependsOn)</CalculatePackageVersionDependsOn>
+    <UpdateAssemblyInfo>false</UpdateAssemblyInfo>
+    <WriteVersionInfoToBuildLog>false</WriteVersionInfoToBuildLog>
+  </PropertyGroup>
+
+  <!--
+    Common target to caluclate the version for anything that needs it.
```
DHowett-MSFT: nit: calculate

PR #2431, opened 2017-04-06
Update compiler with fixes for C++/WinRT

bviglietta:
These new compiler binaries have fixes to allow building C++/WinRT.


bviglietta:

bviglietta:
The changes in UIButtonTests.mm are included because it appears that the new compiler exposed a test bug that wasn't previously causing the test to fail.

tests/functionaltests/Tests/UIKitTests/UIButtonTests.mm
```diff
@@ -1411,21 +1413,23 @@ - (void)dealloc {
...
+        EventRegistrationToken ert{};
+        EventRegistrationToken* const ertAddr = &ert;
 
         WXFrameworkElement* xamlElement = [buttonToTest xamlElement];
         ASSERT_OBJCNE(xamlElement, nil);
 
         // Wait for the layerContent to be part of the visual tree
         dispatch_sync(dispatch_get_main_queue(), ^{
-            EventRegistrationToken ert = [xamlElement addLayoutUpdatedEvent:^(RTObject* sender, RTObject* args) {
+            *ertAddr = [xamlElement addLayoutUpdatedEvent:^(RTObject* sender, RTObject* args) {
```
DHowett-MSFT: this seems like an unusual workaround to using `__block`

bviglietta: Does __block tunnel through multiple levels of blocks? I tried that first and the test was still crashing, so I assumed that it didn't, but I don't know for sure how it's supposed to work.

DHowett-MSFT: huh.. it certainly should

bviglietta: I've verified that things _should_ work as you describe. I believe the reason it works with a pointer and not with a __block variable is because of a compiler bug, which I'll follow up on.

Issue #352, opened 2016-02-18
Apps crash on ARM when trying to catch exceptions


PooyaZv:
Current stack unwind technology does not support the ARM EH model. As a result, catching exceptions currently causes a crash.

Symptoms: Your stack frame will look like the following with objc_exception_throw (roughly) on top of your stack.

```
KernelBase.dll!7735e1f2()
vcruntime140d_app.dll!_CxxThrowException(void * p ...
libobjc2.dll!objc_exception_throw(void * object)
Davayi.exe!_i_ViewController__exceptionGuy(ViewCo ...
Davayi.exe!_i_ViewController__switchChanged_(View ...
Foundation.dll!_i_NSObject__performSelector_withO ...
UIKit.dll!_i_UIControl__sendAction_to_forEvent_(U ...
UIKit.dll!_i_UIControl__sendActionsForControlEven ...
UIKit.dll!_i_UISwitch___didTap_(UISwitch * self,  ...
Foundation.dll!_i_NSObject__performSelector_withO ...
UIKit.dll!_i_UIGestureRecognizer___fire(UIGesture ...
UIKit.dll!_c_UIGestureRecognizer___fireGestures_( ...
UIKit.dll!_i_UIApplication__sendEvent_(UIApplicat ...
UIKit.dll!_i_UIApplication__newMouseEvent(UIAppli ...
Foundation.dll!_i_NSObject__performSelector_(NSOb ...
Foundation.dll!_i_NSRunLoopSource__fire(NSRunLoop ...
Foundation.dll!_i_NSRunLoopState__acceptInputForM ...
Foundation.dll!_i_NSRunLoop__acceptInputForMode_b ...
Foundation.dll!_i_NSRunLoop__runMode_beforeDate_( ...
Foundation.dll!_i_NSRunLoop__run(NSRunLoop * self ...
UIKit.dll!UIApplicationMainLoop() Line 286
StarboardWS.dll!ApplicationMainStart(int argc, ch ...
StarboardWS.dll!WinObjcMainLoop(void * param) Lin ...
KernelBase.dll!77390f3e()
StarboardWS.dll!StartCompositedRunLoop() Line 131 ...
```

PR #2432, opened 2017-04-06
Add support for NSUncaughtExceptionHandler

### LINKED ISSUE #2373: implement NSExceptionHandler and related classes/methods
Also NSGetUncaughtExceptionHandler and NSSetUncaughtExceptionHandler

DHowett-MSFT:
This commit introduces support for Foundation's last line of defense against uncaught exceptions. The application is being torn down, and there's nothing a consumer can do to stop it.

They can now at least log something.

This change cannot be automatically tested, and it cannot be tested with an attached debugger.
This was tested manually on x86.

### x86

WOCCatalog was augmented to contain the following snippet:

```patch
diff --git a/samples/WOCCatalog/WOCCatalog/main.m b/samples/WOCCatalog/WOCCatalog/main.m
index 28f4a68fc..136b8748e 100644
--- a/samples/WOCCatalog/WOCCatalog/main.m
+++ b/samples/WOCCatalog/WOCCatalog/main.m
@@ -17,8 +17,15 @@
 #import <UIKit/UIKit.h>
 #import "AppDelegate.h"

+extern __declspec(dllimport) __stdcall int MessageBoxA(void*,char*,char*,unsigned int);
+#pragma comment(lib, "user32")
+void eh(NSException* ex) {
+    MessageBoxA(0, [[ex reason] UTF8String], "Caught ourselves an NSException!", 0);
+}
+
 int main(int argc, char * argv[]) {
     @autoreleasepool {
+        NSSetUncaughtExceptionHandler(&eh);
```

#### Result
image

### ARM

There's no user32 or MessageBox on ARM, and since we're in an exception handler it would be dangerous to try to create any UI or interact too much with COM and WinRT.

Since named exception catch is already broken on ARM (#352), this likely does not work. No tests (including file writing tests) succeeded.

Fixes #2373.

### DIFF
--- a/Frameworks/Foundation/NSException.cl.cpp
+++ b/Frameworks/Foundation/NSException.cl.cpp
@@ -1,0 +1,65 @@
+#include <windows.h>
+#include <type_traits>
+#include "NSExceptionInternal.h"
+
+// From ucrt's throw.cpp.
+#define EH_EXCEPTION_NUMBER ('msc' | 0xE0000000)
+#define EH_MAGIC_NUMBER1 0x19930520
+
+template <typename T, typename U>
+static std::add_const_t<std::decay_t<T>> rebase_and_cast(uintptr_t base, U value) {
+    return reinterpret_cast<std::add_const_t<std::decay_t<T>>>(base + (uintptr_t)(value));
+}
+
+// _ThrowInfo, _CatchableTypeArray, _CatchableType and _TypeDescriptor (nee std::type_info) are defined by CL at compile time.
+extern "C" LONG WINAPI _NSWindowsUnhandledExceptionFilter(struct _EXCEPTION_POINTERS* ExceptionInfo) {
+    const EXCEPTION_RECORD* ex = ExceptionInfo->ExceptionRecord;
+
+    // An exception thrown by libobjc2 goes through _CxxThrowException, which populates the EXCEPTION_RECORD with the magic C++
+    // exception number (0xE0000000 ORed with the bytes 'msc' (in platform endianness)) and the exception parameters with
+    // a _ThrowInfo*, another magic number, and the object being thrown.
+    if (ex->ExceptionCode == EH_EXCEPTION_NUMBER && ex->ExceptionInformation[0] == EH_MAGIC_NUMBER1 && ex->NumberParameters >= 3) {
+        // On some platforms, thrown exception catch data are relative virtual addresses off the module base.
+        uintptr_t moduleBase = ex->NumberParameters >= 4 ? ex->ExceptionInformation[3] : 0;
+
+        auto throwInfo = reinterpret_cast<_ThrowInfo*>(ex->ExceptionInformation[2]);
+        if (throwInfo && throwInfo->pCatchableTypeArray) {
+            auto catchableTypes = rebase_and_cast<_CatchableTypeArray*>(moduleBase, throwInfo->pCatchableTypeArray);
+            // _ThrowInfo contains a handful of useful things that we completely ignore as we bumble towards the catchable type array.
+            // libobjc2 populates the catchable type array with a list of mangled Objective-C class names that aid the exception catcher
+            // in determining which exception type was thrown. Here we walk that list of types to find one containing NSException.
+            //
+            // This _will_ cause an issue if a clever developer C++-throws a value of a type with NSException in its name that derives from
+            // a type with objc_object in its name.
+
+            bool foundNSException = false;
+            bool foundobjc_object = false;
+            for (int i = 0; i < catchableTypes->nCatchableTypes; ++i) {
+                const _CatchableType* catchableType =
+                    rebase_and_cast<_CatchableType*>(moduleBase, catchableTypes->arrayOfCatchableTypes[i]);
+                const _TypeDescriptor* typeDescriptor = rebase_and_cast<_TypeDescriptor*>(moduleBase, catchableType->pType);
+                if (strstr(typeDescriptor->name, "NSException") != nullptr) {
+                    // We're using strstr because:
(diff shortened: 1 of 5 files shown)


Frameworks/Foundation/NSException.cl.cpp
```diff
@@ -0,0 +1,65 @@
...
+// _ThrowInfo, _CatchableTypeArray, _CatchableType and _TypeDescriptor (nee std::type_info) are defined by CL at compile time.
+extern "C" LONG WINAPI _NSWindowsUnhandledExceptionFilter(struct _EXCEPTION_POINTERS* ExceptionInfo) {
+    const EXCEPTION_RECORD* ex = ExceptionInfo->ExceptionRecord;
+
+    // An exception thrown by libobjc2 goes through _CxxThrowException, which populates the EXCEPTION_RECORD with the magic C++
+    // exception number (0xE0000000 ORed with the bytes 'msc' (in platform endianness)) and the exception parameters with
+    // a _ThrowInfo*, another magic number, and the object being thrown.
+    if (ex->ExceptionCode == EH_EXCEPTION_NUMBER && ex->ExceptionInformation[0] == EH_MAGIC_NUMBER1 && ex->NumberParameters >= 3) {
+        // On some platforms, thrown exception catch data are relative virtual addresses off the module base.
+        uintptr_t moduleBase = ex->NumberParameters >= 4 ? ex->ExceptionInformation[3] : 0;
```
aballway: nit: nullptr, unless this should be a ptrdiff_t, in which case ptrdiff_t

Frameworks/Foundation/NSException.cl.cpp
```diff
@@ -0,0 +1,65 @@
...
+#define EH_EXCEPTION_NUMBER ('msc' | 0xE0000000)
+#define EH_MAGIC_NUMBER1 0x19930520
+
+template <typename T, typename U>
+static std::add_const_t<std::decay_t<T>> rebase_and_cast(uintptr_t base, U value) {
+    return reinterpret_cast<std::add_const_t<std::decay_t<T>>>(base + (uintptr_t)(value));
+}
+
+// _ThrowInfo, _CatchableTypeArray, _CatchableType and _TypeDescriptor (nee std::type_info) are defined by CL at compile time.
+extern "C" LONG WINAPI _NSWindowsUnhandledExceptionFilter(struct _EXCEPTION_POINTERS* ExceptionInfo) {
```
aballway: nit: exceptionInfo

Frameworks/Foundation/NSException.mm
```diff
@@ -38,20 +40,31 @@
 NSString* const NSMallocException = @"NSMallocException";
 NSString* const WinObjCException = @"WinObjC Exception"; // not exported
 
+// The uncaught exception handler is not thread-specific.
+static NSUncaughtExceptionHandler* s_uncaughtExceptionHandler = nullptr;
+
+// Entry point used by NSException_Win32.cpp after C++ ThrowInfo unboxing.
+void _NSExceptionCallUnhandledExceptionHandler(void* untypedException) {
+    auto handler = NSGetUncaughtExceptionHandler();
```
aballway: nit: no auto here

Frameworks/Foundation/NSException.cl.cpp
```diff
@@ -0,0 +1,65 @@
+#include <windows.h>
+#include <type_traits>
+#include "NSExceptionInternal.h"
+
+// From ucrt's throw.cpp.
+#define EH_EXCEPTION_NUMBER ('msc' | 0xE0000000)
+#define EH_MAGIC_NUMBER1 0x19930520
+
+template <typename T, typename U>
+static std::add_const_t<std::decay_t<T>> rebase_and_cast(uintptr_t base, U value) {
```
aballway: comment on what this does and why it needs so many casts 🛩

Frameworks/Foundation/NSException.cl.cpp
```diff
@@ -0,0 +1,65 @@
...
+        if (throwInfo && throwInfo->pCatchableTypeArray) {
+            auto catchableTypes = rebase_and_cast<_CatchableTypeArray*>(moduleBase, throwInfo->pCatchableTypeArray);
+            // _ThrowInfo contains a handful of useful things that we completely ignore as we bumble towards the catchable type array.
+            // libobjc2 populates the catchable type array with a list of mangled Objective-C class names that aid the exception catcher
+            // in determining which exception type was thrown. Here we walk that list of types to find one containing NSException.
+            //
+            // This _will_ cause an issue if a clever developer C++-throws a value of a type with NSException in its name that derives from
+            // a type with objc_object in its name.
+
+            bool foundNSException = false;
```
bviglietta: >foundNSException [](start = 17, length = 16)

Does this work if someone @throws something that doesn't derive from NSException?

bviglietta: Upon closer inspection, it looks like NSUncaughtExceptionHandler is only defined to work for thrown NSExceptions, so I guess we don't want to handle the other cases.

---
In reply to: [110259226](https://github.com/Microsoft/WinObjC/pull/2432#discussion_r110259226) [](ancestors = 110259226)

bviglietta:
:shipit:

rajsesh-msft:
Can we add a test that sets and verifies this?  Granted this was tested with WOCCatalog, but we need an automated test to ensure that we don't regress this in some manner.

DHowett-MSFT:
@rajsesh-msft We can't automate unhandled exception testing. The unhandled exception filter aborts the application, and TAEF/GTest may install their own unhandled exception filters, and any effort to ensure that the uncaught exception filter gets called will destroy the stack scope of the test.

We could implement something based on setjmp and longjmp, but I think that it would be more fragile than we'd like. It may also not work everywhere.

PR #2448, opened 2017-04-07
Fix AutoCF's reference-stealing value move constructor

### LINKED ISSUE #2447: woc::AutoCF<T>(T&&) is preferred over ...const T&..., breaking ref counting
Since function return values are more readily promoted to rvalue refs (here `T&&`) than const lvalue refs, the following code is incorrect.

```cpp
CFTypeRef ThingThatReturnsAnUnownedCFValue();
// Returns a +0 value

woc::StrongCF<CFTypeRef> local{ThingThatReturnsAnUnownedCFValue()};
// Consumes the +0 value, releases it later.
// Boom.
```

DHowett-MSFT:
This commit replaces AutoCF<T, L>(T&&) with
AutoCF<T, L>(TakeOwnershipT, const T&) to clear up the confusion between
the const ref constructor (+1 refcount) and the value move constructor
(+0 refcount; reference-stealing).

Fixes #2447.

**NOTE:** This commit introduces changes in `WinObjC.Language`, and proper consumption will require a new package to be built. _However_, the framework-side changes do not depend on the new SmartTypes changes and will function properly without them.

### DIFF
--- a/Frameworks/CoreGraphics/CGColorSpace.mm
+++ b/Frameworks/CoreGraphics/CGColorSpace.mm
@@ -86,8 +86,8 @@ CFTypeID CGColorSpaceGetTypeID() {
  @Status Interoperable
 */
 CGColorSpaceRef CGColorSpaceCreateDeviceRGB() {
-    static const woc::StrongCF<CGColorSpaceRef> sc_rgbColorSpace{ __CGColorSpace::CreateInstance(kCFAllocatorDefault,
-                                                                                                 kCGColorSpaceModelRGB) };
+    static const auto sc_rgbColorSpace =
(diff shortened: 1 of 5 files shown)


DHowett-MSFT:
I'm going to dismiss some of your reviews to re-review some new changes. Since it's comparatively difficult to revise `WinObjC.Language` files, I'd like to get the AutoId `TakeOwnershipT` stuff in today.

DHowett-MSFT:
@rajsesh-msft @bbowman I dismissed your reviews because I specialized AutoId for TakeOwnershipT as well.

DHowett-MSFT:
With these changes, we no longer need to do this:

```objc
AutoId<NSObject> thing;
thing.attach([NSObject new]);
```

We can instead do this:

```objc
AutoId<NSObject> thing{woc::TakeOwnership, [NSObject new]);
```

PR #2458, opened 2017-04-07
One more try. Missed an errant ItemGroup opening tag.

bbowman:


msft-Jeyaram:
hmm would building catch this? maybe?

PR #2464, opened 2017-04-07
Refactor parts of NSURLConnection

### LINKED ISSUE #2298: NSData dataWithContentsOfURL hangs
Trying to fetch the contents of a URL hangs. For instance, this call will never return, even though the HTTP request completes successfully:

    NSURL* url = [NSURL URLWithString:@"https://www.wikipedia.org/portal/wikipedia.org/assets/img/Wikipedia-logo-v2.png"];
    NSData* data = [NSData dataWithContentsOfURL:url];

ms-jihua:
- Implement [NSURLConnection sendAsynchronousRequest:]
        - Re-implement [NSURLConnection sendSynchronousRequest:] on top of it, as stated in docs
    - Implement a not-yet fully-working setDelegateQueue: (callbacks are not yet called on it)
    - Remove NSURLConnectionState, replace with a lighter-weight, more flexible inner class
    - Changed the NSURLConnection functional tests to no longer run on the main run loop.
      This no longer works, which is correct (doesn't work on reference platform either)
        - Instead, run on another thread's run loop, and an operation queue.
        - Also added tests for sendSynchronousRequest:
    - Change parts of NSURLProtocol, NSRunLoop to use stricter typing
        - Change instances of NSRunLoopMode to actually say NSRunLoopMode instead of NSString*

Fixes #2298


tests/functionaltests/Tests/NSURLConnection.mm
```diff
@@ -87,6 +94,151 @@ - (void)connection:(NSURLConnection*)connection didFailWithError:(NSError*)e {
...
+
+- (NSURLConnection*)createAndStartConnectionWithRequest:(NSString*)URLString {
+    NSURLConnection* connection = [self _createConnectionWithRequest:URLString];
+    [connection performSelector:@selector(start) onThread:_thread withObject:nil waitUntilDone:NO];
+    return connection;
+}
+@end
+
+// Test to verify a request can be successfully made and a valid response and data is received.
+static void _testRequestWithURL(NSURLConnectionTestHelper* connectionTestHelper) {
```
ms-jihua: I considered doing parameterized tests, but since TAEF only parameterizes on Wex Strings, it felt like I was trying too hard to make it fit, so I opted for this instead.

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -19,168 +20,289 @@
...
+@end
 
-    id result = state.receivedData;
+// Actual NSURLConnection implementation
+@implementation NSURLConnection {
+@private
+    StrongId<NSObject> _delegate;
 
-    [connection cancel];
+    std::vector<std::pair<StrongId<NSRunLoop>, StrongId<NSString>>> _scheduledRunLoops; // NSRunLoop, NSRunLoopMode
```
aballway: nit: use NSRunLoopMode

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -19,168 +20,289 @@
...
 }
 
-/**
- @Status Caveat
- @Notes NSError returned is not detailed
-*/
-+ (NSData*)sendSynchronousRequest:(id)request returningResponse:(NSURLResponse**)responsep error:(NSError**)errorp {
-    NSURLConnectionState* state = [[[NSURLConnectionState alloc] init] autorelease];
-    NSURLConnection* connection = [[self alloc] initWithRequest:request delegate:state startImmediately:FALSE];
++ (_NSURLConnection_DataDelegate*)delegateWithQueue:(NSOperationQueue*)queue handler:(void (^)(NSURLResponse*, NSData*, NSError*))handler {
```
DHowett-MSFT: nit: prefer all factory functions X to go through an appropriate initializer `initWithX:`

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -19,168 +20,289 @@
...
 
-    id mode = @"NSURLConnectionRequestMode";
+- (void)connection:(NSURLConnection*)connection didReceiveData:(NSData*)data {
+    [_data appendData:data];
+}
 
-    [connection scheduleInRunLoop:[NSRunLoop currentRunLoop] forMode:mode];
-    [connection start];
+- (void)connectionDidFinishLoading:(NSURLConnection*)connection {
+    [_operationQueue addOperation:[NSBlockOperation blockOperationWithBlock:^void() {
```
DHowett-MSFT: Do we know whether the queue is fire-and-forget or fire-and-wait?
I assumed NSNotificationCenter+OperationQueue was fire-and-forget, but it actually waits for the notification block to complete.

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -19,168 +20,289 @@
...
 
-    [state receiveAllDataInMode:mode];
-    [connection unscheduleFromRunLoop:[NSRunLoop currentRunLoop] forMode:mode];
+@end
 
-    id result = state.receivedData;
+// Actual NSURLConnection implementation
+@implementation NSURLConnection {
+@private
+    StrongId<NSObject> _delegate;
```
DHowett-MSFT: This delegate should be protocol-qualified; `StrongId<NSObject<NSURLConnectionDelegate>>`

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -19,168 +20,289 @@
...
+        __block NSData* data = nil;
+        __block bool finished = false;
+        __block NSCondition* condition = [[NSCondition new] autorelease];
+        NSOperationQueue* queue = [[NSOperationQueue new] autorelease];
+
+        [self sendAsynchronousRequest:request
+                                queue:queue
+                    completionHandler:^void(NSURLResponse* innerResponse, NSData* innerData, NSError* innerError) {
+                        if (response) {
+                            *response = innerResponse ? [[innerResponse retain] autorelease] : nil;
```
aballway: How does autorelease work within an async block?  If that block is executed within an autoreleasepool, will these be released then?

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -19,168 +20,289 @@
...
+    NSURLCacheStoragePolicy _storagePolicy;
+    StrongId<NSURLProtocol> _protocol;
+}
 
-    [connection release];
+/**
+ @Status Interoperable
+*/
++ (BOOL)canHandleRequest:(id)request {
+    return ([NSURLProtocol _URLProtocolClassForRequest:request] != nil) ? YES : NO;
```
DHowett-MSFT: This also requires a `-[protocolClass canInitWithRequest:]`

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -19,168 +20,289 @@
...
+                    completionHandler:^void(NSURLResponse* innerResponse, NSData* innerData, NSError* innerError) {
+                        if (response) {
+                            *response = innerResponse ? [[innerResponse retain] autorelease] : nil;
+                        }
+
+                        if (error) {
+                            *error = innerError ? [[innerError retain] autorelease] : nil;
+                        }
+
+                        data = innerData ? [[innerData retain] autorelease] : nil;
```
DHowett-MSFT: It would be safer to only `retain` data in here and only `autorelease` it outside the block. Otherwise, this +1/-1 pair could resolve when the Operation Queue loops around again and drains its pool.

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -19,168 +20,289 @@
...
+                                                  order:0
+                                                  modes:@[ NSDefaultRunLoopMode ]];
+        }
     }
+}
 
-    [_protocol startLoading];
+/**
+ @Status Caveat
+ @Notes  TODO #: Cannot currently schedule on more than one runloop
```
aballway: nit: add issue number here

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -19,168 +20,289 @@
...
 
-    return result;
+    return nil;
 }
 
 /**
  @Status Interoperable
 */
-+ (id)connectionWithRequest:(id)request delegate:(id)delegate {
++ (NSURLConnection*)connectionWithRequest:(NSURLRequest*)request delegate:(id)delegate {
```
DHowett-MSFT: nit: this could have `instancetype` as return type (even though the docs don't specify it)

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -19,168 +20,289 @@
...
+                 Callbacks do not currently occur on scheduled runloop
+*/
+- (void)scheduleInRunLoop:(NSRunLoop*)aRunLoop forMode:(NSRunLoopMode)mode {
+    @synchronized(self) {
+        if (_delegateQueue.get()) {
+            [NSException raise:NSInternalInconsistencyException
+                        format:@"A connection cannot be scheduled with both an operation queue and a run loop."];
+        }
+
+        if (_scheduledRunLoops.size() >= 1) {
```
aballway: should this just be > 1?

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -19,168 +20,289 @@
...
-*/
-+ (NSData*)sendSynchronousRequest:(id)request returningResponse:(NSURLResponse**)responsep error:(NSError**)errorp {
-    NSURLConnectionState* state = [[[NSURLConnectionState alloc] init] autorelease];
-    NSURLConnection* connection = [[self alloc] initWithRequest:request delegate:state startImmediately:FALSE];
++ (_NSURLConnection_DataDelegate*)delegateWithQueue:(NSOperationQueue*)queue handler:(void (^)(NSURLResponse*, NSData*, NSError*))handler {
+    _NSURLConnection_DataDelegate* ret = [[_NSURLConnection_DataDelegate new] autorelease];
+    if (ret) {
+        ret->_operationQueue = queue;
+        ret->_completionHandler = handler;
+    }
```
msft-Jeyaram: nit: should really be in an initializer then called by the +method

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -19,168 +20,289 @@
...
+        }
+
+        if (_scheduledRunLoops.size() >= 1) {
+            TraceError(TAG, L"NSURLConnection: Scheduling on more than one run loop is currently not supported.");
+            return;
+        }
+
+        StrongId<NSRunLoop> retainedLoop = aRunLoop;
+        StrongId<NSString> retainedMode = mode;
+        _scheduledRunLoops.push_back(std::make_pair(retainedLoop, retainedMode));
```
aballway: nit: emplace_back, might not need the std::make_pair then but I'm not certain

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -19,168 +20,289 @@
...
+    }
+}
+
+/**
+ @Status Caveat
+ @Notes  TODO #: Callbacks do not currently occur on delegate queue
```
aballway: nit: same as above

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -19,168 +20,289 @@
...
 
-    id result = state.receivedData;
+// Actual NSURLConnection implementation
+@implementation NSURLConnection {
+@private
+    StrongId<NSObject> _delegate;
 
-    [connection cancel];
+    std::vector<std::pair<StrongId<NSRunLoop>, StrongId<NSString>>> _scheduledRunLoops; // NSRunLoop, NSRunLoopMode
+    // TODO #: Cannot currently schedule on more than one runloop
```
msft-Jeyaram: >#: [](start = 12, length = 2)

Task ID

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -19,168 +20,289 @@
...
-        _didRetain = TRUE;
+    @synchronized(self) {
+        NSURLProtocol* protocol = _protocolForRequest(_originalRequest, self);
+        if (!protocol) {
+            TraceError(TAG, L"[NSURLConnection start] could not create a NSURLProtocol for its NSURLRequest.");
+            return;
+        }
+        _protocol = protocol;
+
+        if (_delegateQueue.get()) {
```
DHowett-MSFT: nit: prefer `if(_delegateQueue)`; StrongId/AutoId have an `operator bool` that will not require the pointer to be returned to the caller.

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -19,168 +20,289 @@
...
+}
+
+/**
+ @Status Caveat
+ @Notes NSError returned is not detailed
+*/
++ (NSData*)sendSynchronousRequest:(NSURLRequest*)request
+                returningResponse:(NSURLResponse* _Nullable*)response
+                            error:(NSError* _Nullable*)error {
+    if (request) {
```
msft-Jeyaram: > if (request) { [](start = 3, length = 15)

nit: early return :p

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -19,168 +20,289 @@
...
+    @synchronized(self) {
+        NSURLProtocol* protocol = _protocolForRequest(_originalRequest, self);
+        if (!protocol) {
+            TraceError(TAG, L"[NSURLConnection start] could not create a NSURLProtocol for its NSURLRequest.");
+            return;
+        }
+        _protocol = protocol;
+
+        if (_delegateQueue.get()) {
+            [_delegateQueue addOperation:[NSBlockOperation blockOperationWithBlock:^void() {
```
DHowett-MSFT: nit: why `addOperation:` and not `addOperationWithBlock:`?

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -19,168 +20,289 @@
...
 
-    [_protocol startLoading];
+/**
+ @Status Caveat
+ @Notes  TODO #: Cannot currently schedule on more than one runloop
+                 Callbacks do not currently occur on scheduled runloop
+*/
+- (void)scheduleInRunLoop:(NSRunLoop*)aRunLoop forMode:(NSRunLoopMode)mode {
+    @synchronized(self) {
+        if (_delegateQueue.get()) {
```
DHowett-MSFT: nit: as above

Frameworks/Foundation/NSURLConnection.mm
DHowett-MSFT: nit: prefer emplace_back; you get to pass the constructor arguments directly, and these three lines flatten to 

```
xxx.emplace_back(aRunLoop, mode);
```

This also avoids the +1/-1/+1/-1 on both runloop and mode.

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -19,168 +20,289 @@
...
 }
 
 /**
  @Status Interoperable
 */
-- (void)scheduleInRunLoop:(id)runLoop forMode:(id)mode {
-    _scheduled = YES;
+- (void)unscheduleFromRunLoop:(NSRunLoop*)aRunLoop forMode:(NSRunLoopMode)mode {
+    @synchronized(self) {
+        for (auto it = _scheduledRunLoops.begin(); it != _scheduledRunLoops.end(); ++it) {
```
aballway: nit: shouldn't modify a collection while iterating over it, even if it's breaking out of it at the end.  Would prefer std::find() with an erase if != end()

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -19,168 +20,289 @@
...
+    }
+}
+
+/**
+ @Status Caveat
+ @Notes  TODO #: Callbacks do not currently occur on delegate queue
+*/
+- (void)setDelegateQueue:(NSOperationQueue*)queue {
+    @synchronized(self) {
+        if (_scheduledRunLoops.size() > 0) {
```
DHowett-MSFT: do we need a `&& queue`? What if you set the delegate queue to nil?

What if you set the delegate queue when it's already running?

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -19,168 +20,289 @@
...
 
 /**
  @Status Interoperable
 */
 - (void)dealloc {
-    [_request release];
-    [_protocol release];
-    [_delegate release];
-    [_response release];
+    [_originalRequest release];
```
msft-Jeyaram: >   [_originalRequest release]; [](start = 1, length = 30)

good candidate for StrongId

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -19,168 +20,289 @@
...
  @Status Interoperable
 */
-- (void)scheduleInRunLoop:(id)runLoop forMode:(id)mode {
-    _scheduled = YES;
+- (void)unscheduleFromRunLoop:(NSRunLoop*)aRunLoop forMode:(NSRunLoopMode)mode {
+    @synchronized(self) {
+        for (auto it = _scheduledRunLoops.begin(); it != _scheduledRunLoops.end(); ++it) {
+            if ([it->first isEqual:aRunLoop] && [it->second isEqual:mode]) {
+                _scheduledRunLoops.erase(it);
+                return;
```
DHowett-MSFT: may want to caveat this for "only one run loop" so we can find it later

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -194,29 +316,25 @@ - (void)URLProtocol:(id)urlProtocol didFailWithError:(id)error {
...
     }
     */
 
-    _response = [response retain];
+    _response = response;
     _storagePolicy = policy;
 
     if ([_delegate respondsToSelector:@selector(connection:willCacheResponse:)]) {
-        [_delegate connection:self willCacheResponse:response];
+        [_delegate connection:self willCacheResponse:static_cast<NSCachedURLResponse*>(response)];
```
DHowett-MSFT: whoa, this is scary. I realize it is what was already happening, but those two things are very much not the same thing

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -296,30 +398,22 @@ - (void)URLProtocol:(NSURLProtocol*)urlProtocol wasRedirectedToRequest:(NSURLReq
...
 
     if (!newRequest) {
         [self cancel];
         return;
     }
-    [self _setRequest:newRequest]; // regenerates _protocol
 
+    NSURLProtocol* protocol = _protocolForRequest(request, self);
+    if (!protocol) {
+        TraceError(TAG, L"[NSURLConnection start] could not create a NSURLProtocol for its NSURLRequest.");
```
DHowett-MSFT: not `start` anymore!

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -296,30 +398,22 @@ - (void)URLProtocol:(NSURLProtocol*)urlProtocol wasRedirectedToRequest:(NSURLReq
...
+        return;
+    }
+    _protocol = protocol;
```
DHowett-MSFT: do we need to update `_request` like setRequest was doing?

Frameworks/Foundation/NSURLProtocol.mm
```diff
@@ -46,20 +46,20 @@ + (BOOL)registerClass:(id)cls {
...
-+ (id)_URLProtocolClassForRequest:(id)request {
++ (Class)_URLProtocolClassForRequest:(NSURLRequest*)request {
     @synchronized(self) {
-        NSArray* classes = [self _registeredClasses];
+        NSArray<Class>* classes = [self _registeredClasses];
         NSInteger count = [classes count];
 
         while (--count >= 0) {
-            id check = [classes objectAtIndex:count];
+            Class check = [classes objectAtIndex:count];
```
aballway: nit: update this to be for in loop

Frameworks/Foundation/NSURLProtocol.mm
```diff
@@ -74,34 +74,13 @@ - (instancetype)initWithRequest:(NSURLRequest*)request
                  cachedResponse:(NSCachedURLResponse*)cachedResponse
                          client:(id<NSURLProtocolClient>)client {
     if (self = [super init]) {
-        _request.attach([request mutableCopy]);
-        _cachedResponse = cachedResponse;
-        _client = client;
+        _request = [request mutableCopy];
```
DHowett-MSFT: hmm.. why move away from StrongId?

aballway: won't these overretain?

Frameworks/Foundation/NSURLProtocol_WinHTTP.mm
```diff
@@ -251,7 +251,7 @@ - (void)startLoading {
 
         NSError* error = nil;
         try {
-            auto httpRequestMessage = _requestMessageForNS(_request);
+            auto httpRequestMessage = _requestMessageForNS(self.request);
```
DHowett-MSFT: these getter calls seem unnecessary; NSURLProtocol for HTTP is a leaf class and cannot be subclassed.

Frameworks/Foundation/NSURLConnection.mm
DHowett-MSFT: it's a property, which makes it a _less good_ candidate for StrongId

Frameworks/Foundation/NSURLConnection.mm
aballway: I disagree 😛

msft-Jeyaram:

nit: remove

---
Refers to: Frameworks/include/NSURLProtocolInternal.h:3 in 120e3ec. [](commit_id = 120e3ec903c8e40be38978eafa2a2c440154840d, deletion_comment = False)

tests/functionaltests/Tests/NSURLConnection.mm
```diff
@@ -23,36 +23,43 @@ typedef NS_ENUM(NSInteger, NSURLConnectionDelegateType) {
...
+- (NSURLConnection*)_createConnectionWithRequest:(NSString*)URLString;
+- (NSURLConnection*)createAndStartConnectionWithRequest:(NSString*)URLString;
 
 @end
 
 @implementation NSURLConnectionTestHelper
-
 - (instancetype)init {
     if (self = [super init]) {
-        self->_condition = [[NSCondition alloc] init];
```
msft-Jeyaram: >[NSCondition alloc] init]; [](start = 28, length = 26)

:)

tests/functionaltests/Tests/NSURLConnection.mm
```diff
@@ -23,36 +23,43 @@ typedef NS_ENUM(NSInteger, NSURLConnectionDelegateType) {
...
-- (NSURLConnection*)createConnectionWithRequest:(NSString*)URLString {
+- (NSURLConnection*)_createConnectionWithRequest:(NSString*)URLString {
     NSMutableURLRequest* request = [NSMutableURLRequest requestWithURL:[NSURL URLWithString:URLString]];
     NSURLConnection* connection = [[NSURLConnection alloc] initWithRequest:request delegate:self startImmediately:NO];
     return connection;
 }
 
+- (NSURLConnection*)createAndStartConnectionWithRequest:(NSString*)URLString {
+    ASSERT_TRUE_MSG(false, "This superclass version should never be called.");
+    return nil;
```
msft-Jeyaram: why assert? throw/fail?

tests/functionaltests/Tests/NSURLConnection.mm
```diff
@@ -87,6 +94,151 @@ - (void)connection:(NSURLConnection*)connection didFailWithError:(NSError*)e {
...
+@property (retain) NSThread* thread;
+- (void)doNothing;
+- (void)spinRunLoop;
+@end
+
+@implementation NSURLConnectionTestHelper_RunLoop
+
+- (instancetype)init {
+    if (self = [super init]) {
+        self->_thread = [[NSThread alloc] initWithTarget:self selector:@selector(spinRunLoop) object:nil];
```
msft-Jeyaram: >self->_thread = [[NSThread alloc] initWithTarget:self selector:@selector(spinRunLoop) object:nil]; [](start = 8, length = 98)

+1 on the init and +1 on the retain?

include/Foundation/NSURLConnection.h
```diff
@@ -18,32 +18,23 @@
...
-    id _response;
-    unsigned _storagePolicy;
-    BOOL _didRetain, _didRelease, _scheduled;
-}
-
+@interface NSURLConnection : NSObject <NSURLProtocolClient>
 + (BOOL)canHandleRequest:(NSURLRequest*)request;
 @property (readonly, copy) NSURLRequest* originalRequest;
-@property (readonly, copy) NSURLRequest* currentRequest;
+@property (readonly, copy) NSURLRequest* currentRequest STUB_PROPERTY;
```
rajsesh-msft: >STUB_PROPERTY [](start = 56, length = 13)

while we are here, can we implement this?  this should be easy...

Frameworks/Foundation/NSURLConnection.mm
ms-jihua: sigh okkkk

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -19,168 +20,289 @@
...
+    [_operationQueue addOperation:[NSBlockOperation blockOperationWithBlock:^void() {
+                         _completionHandler(_response.get(), nil, error);
+                     }]];
+}
 
-        return nil;
+- (void)connection:(NSURLConnection*)connection didReceiveResponse:(NSURLResponse*)response {
+    // We get some of these for each redirect. Only capture the final one.
+    _response = response;
+    if (!_data.get() || [_data length] > 0) {
```
ms-jihua: didnt know that worked, til

Frameworks/Foundation/NSURLConnection.mm
ms-jihua: it's fire-and-forget

Frameworks/Foundation/NSURLConnection.mm
ms-jihua: can't, NSRunLoopMode is unfortunately typedef'd as NSString* and not NSString :(

Frameworks/Foundation/NSURLConnection.mm
ms-jihua: i know, i knowww - i dont like filing these tasks until the PR is actually out because they often lack context without the PR,

Frameworks/Foundation/NSURLConnection.mm
DHowett-MSFT: huh.. AutoId should be able to remove_pointer that properly. I wanted for `AutoId<SNString>` and `AutoId<NSString*>` to be functionally equivalent, so it strips or adds the pointer as necessary.

Frameworks/Foundation/NSURLProtocol.mm
DHowett-MSFT: @aballway This is iterating in reverse. Cheaper than a reverseObjectEnumerator,.

aballway: ⭕️ that they are

Frameworks/Foundation/NSURLConnection.mm
ms-jihua: #2469

Frameworks/Foundation/NSURLConnection.mm
ms-jihua: pretty much

Frameworks/Foundation/NSURLConnection.mm
ms-jihua: good point.

Frameworks/Foundation/NSURLConnection.mm
ms-jihua: did not know that

Frameworks/Foundation/NSURLConnection.mm
ms-jihua: oh. i didn't actually try it i just figured it wouldn't work. ;;;;

Frameworks/Foundation/NSURLConnection.mm
ms-jihua: i completely forgot that existed despite writing it ;;;;;;;;;;;

Frameworks/Foundation/NSURLConnection.mm
ms-jihua: ? no, as is, it only allows something in if the vector is empty, which is what i want.

Frameworks/Foundation/NSURLConnection.mm
ms-jihua: it works. i'm silly.

Frameworks/Foundation/NSURLConnection.mm
ms-jihua: Only the first queue seems to take.

Documented to be bad, but seems to not throw or anything on the reference platform. I'll make our version log an error.

Frameworks/Foundation/NSURLConnection.mm
ms-jihua: yuuuppp, lemme leave a TODO here to fix up later.

Frameworks/Foundation/NSURLConnection.mm
ms-jihua: oops

Frameworks/Foundation/NSURLProtocol.mm
ms-jihua: they were properties, and one was working incorrectly and not taking a copy. seemed like a good idea to just pivot to autosynthesis

ms-jihua: _foo = doesn't call the setter, so everything is +1 as it should be

Frameworks/Foundation/NSURLProtocol_WinHTTP.mm
ms-jihua: NSURLProtocol_WinHTTP doesn't have access to the superclass member now that it's synthesized

tests/functionaltests/Tests/NSURLConnection.mm
ms-jihua: this is an arc-enabled file

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -54,27 +55,32 @@ @implementation _NSURLConnection_DataDelegate {
 
     StrongId<NSURLResponse> _response;
     StrongId<NSMutableData> _data;
+
+    bool _started;
```
DHowett-MSFT: Wait, this doesn't even seem like it gets used in the body of this data delegate... but _started gets used in NSURLConnection. Did some wires get crossed?

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -109,8 +115,13 @@ @implementation NSURLConnection {
...
 */
-+ (BOOL)canHandleRequest:(id)request {
-    return ([NSURLProtocol _URLProtocolClassForRequest:request] != nil) ? YES : NO;
++ (BOOL)canHandleRequest:(NSURLRequest*)request {
+    Class protocolClass = [NSURLProtocol _URLProtocolClassForRequest:request];
+    if (!protocolClass) {
+        return NO;
+    }
+
+    return [protocolClass canInitWithRequest:request];
```
DHowett-MSFT: you can keep this folded, actually. `[nil canInitWithRequest:request]` returns NO

DHowett-MSFT: (no need for the if check that the compiler is going to emit anyway)

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -260,37 +294,45 @@ - (void)scheduleInRunLoop:(NSRunLoop*)aRunLoop forMode:(NSRunLoopMode)mode {
...
+ @Notes  TODO #2469: Cannot currently schedule on more than one runloop
 */
 - (void)unscheduleFromRunLoop:(NSRunLoop*)aRunLoop forMode:(NSRunLoopMode)mode {
     @synchronized(self) {
-        for (auto it = _scheduledRunLoops.begin(); it != _scheduledRunLoops.end(); ++it) {
-            if ([it->first isEqual:aRunLoop] && [it->second isEqual:mode]) {
-                _scheduledRunLoops.erase(it);
-                return;
-            }
+        auto it = std::find(_scheduledRunLoops.begin(), _scheduledRunLoops.end());
```
DHowett-MSFT: this appears to be finding "the run loop"? It doesn't care about its arguments. If that's the case, why not just _scheduledRunLoops.clear()?

tests/unittests/Foundation/NSURLTests.m
```diff
@@ -291,4 +291,234 @@ void testNSURLMethod(SEL selector, NSURL* input, id argument, NSURL* expected) {
...
+    if (self = [super init]) {
+        self->_operationQueue = [NSOperationQueue new];
+    }
+    return self;
+}
+
+- (NSURLConnection*)createAndStartConnectionWithRequest:(NSString*)URLString {
+    NSURLConnection* connection = [self _createConnectionWithRequest:URLString];
+    [connection setDelegateQueue:_operationQueue];
+    // [connection setDelegateQueue:[NSOperationQueue new]];
```
DHowett-MSFT: commented code

tests/unittests/Foundation/NSURLTests.m
```diff
@@ -291,4 +291,234 @@ void testNSURLMethod(SEL selector, NSURL* input, id argument, NSURL* expected) {
     NSString* pathAsString = @"/temp/hello/world/中文/你好.txt";
     NSURL* pathAsURL = [NSURL fileURLWithPath:pathAsString];
     ASSERT_OBJCEQ(pathAsString, pathAsURL.path);
-}
\ No newline at end of file
+}
```
DHowett-MSFT: nit: prefer not cluttering the URL tests with URL Download tests

tests/unittests/Foundation/NSURLTests.m
```diff
@@ -291,4 +291,234 @@ void testNSURLMethod(SEL selector, NSURL* input, id argument, NSURL* expected) {
...
+    NSString* urlString = @"https://httpbin.org/cookies/set?winobjc=awesome";
+    LOG_INFO("Establishing download task with url %@", urlString);
+    NSURLConnection* connection = [connectionTestHelper createAndStartConnectionWithRequest:urlString];
+
+    // Wait for data.
+    [connectionTestHelper.condition lock];
+    for (int i = 0; (i < 5) && ([connectionTestHelper.delegateCallOrder count] != 2); i++) {
+        [connectionTestHelper.condition waitUntilDate:[NSDate dateWithTimeIntervalSinceNow:3]];
+    }
+    ASSERT_EQ_MSG(2, [connectionTestHelper.delegateCallOrder count], "FAILED: We should have received two delegates call by now!");
```
DHowett-MSFT: nit: delegate calls

tests/unittests/Foundation/NSURLTests.m
```diff
@@ -291,4 +291,234 @@ void testNSURLMethod(SEL selector, NSURL* input, id argument, NSURL* expected) {
...
+
+    // Make sure we received data.
+    ASSERT_EQ_MSG(NSURLConnectionDelegateDidReceiveData,
+                  [(NSNumber*)[connectionTestHelper.delegateCallOrder objectAtIndex:1] integerValue],
+                  "FAILED: didReceiveData should be the first delegate to be called!");
+    ASSERT_TRUE_MSG((connectionTestHelper.data != nil), "FAILED: We should have received some data!");
+    LOG_INFO("Received data: %@", [[NSString alloc] initWithData:connectionTestHelper.data encoding:NSUTF8StringEncoding]);
+
+    // Make sure there was no error.
+    ASSERT_TRUE_MSG((connectionTestHelper.error == nil), "FAILED: Connection returned error %@!", connectionTestHelper.error);
```
DHowett-MSFT: nit: assert_eq?

tests/unittests/Foundation/NSURLTests.m
```diff
@@ -291,4 +291,234 @@ void testNSURLMethod(SEL selector, NSURL* input, id argument, NSURL* expected) {
...
+}
+
+// Test to verify a request can be successfully made but no data was received and a valid response error code was received.
+static void _testRequestWithURL_Failure(NSURLConnectionTestHelper* connectionTestHelper) {
```
DHowett-MSFT: this appears unused

Frameworks/Foundation/NSURLProtocol_WinHTTP.mm
DHowett-MSFT: If that's the case, we should at least lift these out to locals then. Otherwise, each hit is a `[[_request retain] autorelease]`

DHowett-MSFT: (the compiler can't perform common subexpression elimination as it can't prove that `self.request` is idempotent)

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -19,168 +20,333 @@
...
+ @Status Caveat
+ @Notes  TODO #2469: Cannot currently schedule on more than one runloop
 */
-- (void)scheduleInRunLoop:(id)runLoop forMode:(id)mode {
-    _scheduled = YES;
+- (void)unscheduleFromRunLoop:(NSRunLoop*)aRunLoop forMode:(NSRunLoopMode)mode {
+    @synchronized(self) {
+        auto it = std::find(_scheduledRunLoops.begin(), _scheduledRunLoops.end());
+        if (it != _scheduledRunLoops.end()) {
+            _scheduledRunLoops.erase(it);
```
DHowett-MSFT: also, shouldn't this call something _on_ the runloop to unschedule ourselves?

ms-jihua: Nope. https://developer.apple.com/reference/foundation/nsurlconnection/1409722-unschedulefromrunloop?language=objc

"Use this method to unschedule the connection from an undesired run loop and mode before starting the connection.
You cannot reschedule a connection after it has started."

tests/unittests/Foundation/NSURLTests.m
ms-jihua: oops, ty for catching

tests/unittests/Foundation/NSURLTests.m
ms-jihua: oops, I was testing something on the reference platform and apparently forgot to delete this?

Frameworks/Foundation/NSURLConnection.mm
ms-jihua: ugh this is what i get for rushing on friday...

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -110,18 +109,15 @@ @implementation NSURLConnection {
     StrongId<NSURLResponse> _response;
     NSURLCacheStoragePolicy _storagePolicy;
     StrongId<NSURLProtocol> _protocol;
+
+    bool _started;
```
DHowett-MSFT: (is this superior to property autosynthesis? it looks like there's already a property for this ivar, and the ivar will be automatically generated just like it was before this change)

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -329,7 +325,11 @@ - (void)setDelegateQueue:(NSOperationQueue*)queue {
 */
 - (void)unscheduleFromRunLoop:(NSRunLoop*)aRunLoop forMode:(NSRunLoopMode)mode {
     @synchronized(self) {
-        auto it = std::find(_scheduledRunLoops.begin(), _scheduledRunLoops.end());
+        auto it = std::find_if(_scheduledRunLoops.begin(),
+                               _scheduledRunLoops.end(),
+                               [aRunLoop](std::pair<StrongId<NSRunLoop>, StrongId<NSRunLoopMode>>& pair) {
```
DHowett-MSFT: should this ref be const?

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -329,7 +325,11 @@ - (void)setDelegateQueue:(NSOperationQueue*)queue {
...
+                                   return aRunLoop == pair.first.get();
```
DHowett-MSFT: nit: AutoId should have an operator== for T (but I wouldn't be surprised if it didn't)

Frameworks/Foundation/NSURLConnection.mm
ms-jihua: ? There's no property for this ivar.

ms-jihua:
ping!

ms-jihua:
ping! last call @msft-Jeyaram @aballway @rajsesh-msft

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -19,168 +20,333 @@
...
 
+#import <algorithm>
+#import <vector>
+
 static const wchar_t* TAG = L"NSURLConnection";
 
-@implementation NSURLConnection
+// Private helper function that returns a NSURLProtocol object capable of handling a NSURLRequest
+// Returns nil if one cannot be created
+static NSURLProtocol* _protocolForRequest(NSURLRequest* request, id<NSURLProtocolClient> client) {
```
ms-jihua: Hmm, originally i only proposed this convention for within CoreText/Graphics, given that we had a Hhigh incidence of both file-private and framework-private functions. Did we begin to use this more broadly?

Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -19,168 +20,333 @@
...
+
+    [self sendAsynchronousRequest:request
+                            queue:queue
+                completionHandler:^void(NSURLResponse* innerResponse, NSData* innerData, NSError* innerError) {
+                    if (response) {
+                        *response = innerResponse ? [innerResponse retain] : nil;
+                    }
+
+                    if (error) {
+                        *error = innerError ? [innerError retain] : nil;
```
msft-Jeyaram: >innerError ? [innerError retain] : nil; [](start = 33, length = 39)

does this matter, if the innerError is nil, retain would return nil?

msft-Jeyaram:

still needs update 

---
In reply to: [292643560](https://github.com/Microsoft/WinObjC/pull/2464#issuecomment-292643560) [](ancestors = 292643560)

---
Refers to: Frameworks/include/NSURLProtocolInternal.h:3 in 120e3ec. [](commit_id = 120e3ec903c8e40be38978eafa2a2c440154840d, deletion_comment = False)

tests/functionaltests/Tests/NSURLConnection.mm
```diff
@@ -23,36 +23,43 @@ typedef NS_ENUM(NSInteger, NSURLConnectionDelegateType) {
...
+- (NSURLConnection*)createAndStartConnectionWithRequest:(NSString*)URLString;
 
 @end
 
 @implementation NSURLConnectionTestHelper
-
 - (instancetype)init {
     if (self = [super init]) {
-        self->_condition = [[NSCondition alloc] init];
+        self->_condition = [NSCondition new];
```
msft-Jeyaram: >self->_condition [](start = 8, length = 16)

just use _condition, I almost confused myself thinking this was +2.

Frameworks/Foundation/NSURLConnection.mm
ms-jihua: mm, i guess that's probably fine, yeah

tests/functionaltests/Tests/NSURLConnection.mm
ms-jihua: is _condition really more clear than self->_condition?

msft-Jeyaram: It would be, given self.* isn't a syntax access of the instance variable directly. It would be confusing to use self-> which is a direct instance accessing (by passing the property) 

---
In reply to: [111228615](https://github.com/Microsoft/WinObjC/pull/2464#discussion_r111228615) [](ancestors = 111228615)

msft-Jeyaram: it wouldn't be a good choice to use self->, unless you have a specific reason to.

---
In reply to: [111229286](https://github.com/Microsoft/WinObjC/pull/2464#discussion_r111229286) [](ancestors = 111229286,111228615)

tests/functionaltests/Tests/NSURLConnection.mm
```diff
@@ -87,6 +94,151 @@ - (void)connection:(NSURLConnection*)connection didFailWithError:(NSError*)e {
...
 
+// Subclass of NSURLConnectionHelper that sets its connection on a delegate queue before starting it.
+@interface NSURLConnectionTestHelper_OperationQueue : NSURLConnectionTestHelper
+@property (retain) NSOperationQueue* operationQueue;
+@end
+
+@implementation NSURLConnectionTestHelper_OperationQueue
+- (instancetype)init {
+    if (self = [super init]) {
+        self->_operationQueue = [NSOperationQueue new];
```
msft-Jeyaram: > self-> [](start = 7, length = 7)

here too

tests/functionaltests/Tests/NSURLConnection.mm
ms-jihua: I generally default to using self-> in inits, to guard against [super init] doing a replacement. That's not really possible here since the superclass is NSObject, so I'll change it, but I disagree that there needs to be a specific reason to use self->, especially since inheritance interactions aren't always this clear...

---
In reply to: [111229643](https://github.com/Microsoft/WinObjC/pull/2464#discussion_r111229643) [](ancestors = 111229643,111229286,111228615)

ms-jihua: oops im just wrong

---
In reply to: [111231093](https://github.com/Microsoft/WinObjC/pull/2464#discussion_r111231093) [](ancestors = 111231093,111229643,111229286,111228615)

PR #2465, opened 2017-04-07
Fix incorrect vsimporter logic

memontic-ms:
When I rejiggered my logic before committing I seem to have dropped a !
Oops.
It is concerning that this error did not break my CI build. I'll look into it.


DHowett-MSFT:
Nit; title not in imperative mood

PR #2470, opened 2017-04-07
remove some things in the nuget.config that were messing things up (#…

bbowman:
…2368)


bbowman:
I swore I already cherry picked this

PR #1515, opened 2016-12-06
Added a dummy modulemap for include/objc

mukhole:
This is a fix for issue#1397.
This module excludes all headers so its just a dummy to prevent other modules from adding symbols from objc to their own module.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/1515)

PR #2473, opened 2017-04-10
Move file that lost its way during packaging merge

memontic-ms:
The fix from PR #1515 was left behind in an old location during the packaging merge.

This change moves
include/objc/module.modulemap → tools/include/objc/module.modulemap


pradipd:
:shipit:

PR #2474, opened 2017-04-10
Convert AddressBook framework to C++/WinRT

bviglietta:
This change removes uses of projections from the AddressBook framework and replaces them with C++/WinRT.

### DIFF
--- a/Frameworks/AddressBook/ABAddressBookManagerInternal.mm
+++ b/Frameworks/AddressBook/ABAddressBookManagerInternal.mm
@@ -14,19 +14,38 @@
 //
 //******************************************************************************
 
+#import <unordered_set>
+#import <unordered_map>
+#import <utility>
 #import "ABAddressBookManagerInternal.h"
 #import "ABContactInternal.h"
-#import "UWP/WindowsApplicationModelContacts.h"
+#import "CppWinRTHelpers.h"
 
-@implementation __ABContactOperation
+using namespace winrt::Windows::ApplicationModel::Contacts;
+namespace WF = winrt::Windows::Foundation;
+namespace WFC = winrt::Windows::Foundation::Collections;
 
-- (id)initWithContact:(WACContact*)contact shouldDelete:(BOOL)shouldDelete {
+@implementation __ABContactOperation {
+    TrivialDefaultConstructor<Contact> _contact;
+}
+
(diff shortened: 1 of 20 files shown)


build/Tests/Tests.Shared/Tests.Shared.vcxitems
```diff
@@ -41,6 +41,10 @@
   </ItemGroup>
   <ItemGroup>
     <ClInclude Include="$(MSBuildThisFileDirectory)..\..\..\Frameworks\include\CppUtils.h" />
+    <ClInclude Include="$(MSBuildThisFileDirectory)..\..\..\tests\unittests\Tests.Shared\ByteUtils.h" />
```
bviglietta: ><ClInclude Include="$(MSBuildThisFileDirectory)..\..\..\tests\unittests\Tests.Shared\ByteUtils.h" /> [](start = 4, length = 100)

This change is accidental. I'll remove it.

bviglietta:

Frameworks/AddressBook/ABAddressBookManagerInternal.mm
```diff
@@ -172,22 +200,36 @@ - (id)init {
     return self;
 }
 
+@dynamic contactStore;
```
DHowett-MSFT: No need for this if you're inmplementing it directly under it. Also, `@dynamic` isn't for manually-implemented properties; it's more for "these will exist at runtime, compiler, I promise"

bviglietta: I believe I need @dynamic to suppress auto-generation of the field backing the property, don't I?

DHowett-MSFT: Huh. It looks like [the goal posts keep moving](http://stackoverflow.com/a/9368743) on what/when autosynthesis is engaged. Github does not have an emoticon that is glowering quite enough for me to express myself.

bviglietta: I would be happy enough to let autosynthesis do its thing, except that in this case I want the property and the synthesized field to have different types.

DHowett-MSFT: Nah, you're right to use `@dynamic` here. :)

PR #2479, opened 2017-04-10
Re-Implementation of reverse string via CFString

### LINKED ISSUE #2478: NSString _reverseString should be implemented via CFString

msft-Jeyaram:
- Implement _reverseString via CoreFoundation and _wcsrev
- Remove _reverseString from Foundation (only CoreText uses it)
- CTline _getTruncatedStringFromSourceLine should use CF*based reverse string
- Remove redundant tests

fixes #2478


DHowett-MSFT:
nit: reword to be imperative mood

msft-Jeyaram:
Getting some bench mark for old vs new

Frameworks/CoreFoundationAdditions/CFStringHelpers.mm
```diff
@@ -0,0 +1,42 @@
...
+
+CFStringRef _CFStringCreateReversedString(CFStringRef string) {
+    if (string == nullptr) {
+        return nullptr;
+    }
+
+    CFIndex index = CFStringGetLength(string);
+
+    if (index < 2) {
+        return CFStringCreateCopy(kCFAllocatorDefault, string);
```
DHowett-MSFT: this could also be CFRetain

Frameworks/CoreFoundationAdditions/CFStringHelpers.mm
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    CFMutableStringRef result = CFStringCreateMutable(kCFAllocatorDefault, 0);
```
DHowett-MSFT: If you create the new string with a capacity hint from the original string, you can avoid the inevitable reallocations

Frameworks/CoreFoundationAdditions/CFStringHelpers.mm
```diff
@@ -0,0 +1,42 @@
...
+
+    while (index > 0) {
```
DHowett-MSFT: it may be cheaper to create a unichar buffer and pass that off to CFStringCreateWithCharacters -- you get the added benefit of not creating a _mutable_ string

It should be somewhat "safe" to use utf-16 codepoints or unichars; it's definitely safer than utf-8

rajsesh-msft: This is going to suck when it comes to perf.

DHowett-MSFT:
Also, I wonder how it works with Composing Diacritics! It might be reasonable to normalize the string first...

Frameworks/CoreFoundationAdditions/CFStringHelpers.mm
```diff
@@ -0,0 +1,42 @@
...
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#import <CoreFoundation/CFString.h>
+#import <Starboard/SmartTypes.h>
+#import <CFInternal.h>
+#import "CFFoundationInternal.h"
+
+CFStringRef _CFStringCreateReversedString(CFStringRef string) {
```
rajsesh-msft: Given that there is exactly one place where this is used (and was used in the past), I don't like that we are doing this.  Why not have the helper in CT and just use it there?

Frameworks/CoreFoundationAdditions/CFStringHelpers.mm
msft-Jeyaram: If the input is a mutable string, retain might not be a good idea/

DHowett-MSFT: Good point!

Frameworks/CoreFoundationAdditions/CFStringHelpers.mm
msft-Jeyaram: that's what i was going with first, it might be more sensible given that only CT uses it.
Hmm i might as well move it in there, but this kind of makes it more util based.

Frameworks/CoreFoundationAdditions/CFStringHelpers.mm
msft-Jeyaram: That's what my initial plan was, but since this seems very common and could be utilized by others, it would make a good addition. Also we already have some tests that can be tested independently.

Frameworks/CoreFoundationAdditions/CFStringHelpers.mm
msft-Jeyaram: YEP utf-16 is safer than utf-8 :)

---
In reply to: [110780284](https://github.com/Microsoft/WinObjC/pull/2479#discussion_r110780284) [](ancestors = 110780284)

msft-Jeyaram: might as well use buffer indexing, given it's somewhat "safe"

---
In reply to: [110780304](https://github.com/Microsoft/WinObjC/pull/2479#discussion_r110780304) [](ancestors = 110780304)

Frameworks/CoreFoundationAdditions/CFStringHelpers.mm
rajsesh-msft: Lets not do this.  the _reverseString was added because some other code was using it.  Given that there is only one instance where _reverseString was used, there is no need to go overboard here.  

---
In reply to: [110781327](https://github.com/Microsoft/WinObjC/pull/2479#discussion_r110781327) [](ancestors = 110781327)

rajsesh-msft: also, what is wrong with good old fashioned strrev variants?

---
In reply to: [110782273](https://github.com/Microsoft/WinObjC/pull/2479#discussion_r110782273) [](ancestors = 110782273,110781327)

msft-Jeyaram: _strrev + UTF-16 encoding should do the trick then.

---
In reply to: [110782324](https://github.com/Microsoft/WinObjC/pull/2479#discussion_r110782324) [](ancestors = 110782324,110782273,110781327)

msft-Jeyaram:
@rajsesh-msft  still need reverse in CFString backed by strrev, that is contained in CoreText.
Updating the CR

Frameworks/CoreText/CTUtilities.mm
```diff
@@ -15,7 +15,10 @@
 //******************************************************************************
 
 #import <CoreText/CTUtilities.h>
+#import <Starboard/SmartTypes.h>
```
msft-Jeyaram: removed

Frameworks/CoreText/CTLine.mm
```diff
@@ -188,8 +187,9 @@ CTLineRef CTLineCreateTruncatedLine(CTLineRef sourceLine, double width, CTLineTr
         CFDictionaryRef attribs = CTRunGetAttributes(currentRun);
 
         if (truncationType == kCTLineTruncationStart) {
-            NSString* reverse = [runString _reverseString];
-            NSAttributedString* string = [[NSAttributedString alloc] initWithString:reverse attributes:(NSDictionary*)attribs];
+            auto reverse = woc::MakeStrongCF<CFStringRef>(_CTCreateReversedString(static_cast<CFStringRef>(runString)));
```
ms-jihua: if there's just the one usage of it, can it be a file-scope function instead?

Frameworks/CoreText/CTUtilities.mm
```diff
@@ -24,3 +27,24 @@
...
 }
+
+// Create a reversed version of the string
+CFStringRef _CTCreateReversedString(CFStringRef string) {
+    if (string == nullptr) {
+        return nullptr;
+    }
+
+    // Returns the number (in terms of UTF-16 code pairs) of Unicode characters in a string.
+    CFIndex length = CFStringGetLength(string);
```
msft-Jeyaram: using CFStringGetBytes to be safer

msft-Jeyaram:
@DHowett-MSFT  @ms-jihua  please take a look

Frameworks/CoreText/CTLine.mm
```diff
@@ -16,16 +16,38 @@
...
 #import <StubReturn.h>
-#import "NSStringInternal.h"
 #import "CoreTextInternal.h"
 #import "CGContextInternal.h"
 #import "DWriteWrapper_CoreText.h"
 #import <CoreText/CTTypesetter.h>
 
 #import <algorithm>
 #import <numeric>
+#import <wchar.h>
```
DHowett-MSFT: nit: prefer `<cwchar>`

Frameworks/CoreText/CTLine.mm
```diff
@@ -16,16 +16,38 @@
...
+
+    CFIndex length = CFStringGetLength(string);
+    CFIndex usedBufLen;
+    CFStringGetBytes(string, CFRangeMake(0, length), kCFStringEncodingUTF16, 0, false, nullptr, length, &usedBufLen);
+
+    if (length < 2) {
+        return CFStringCreateCopy(kCFAllocatorDefault, string);
+    }
+
+    std::vector<UniChar> characters((usedBufLen / (sizeof(UniChar))) + 1, '\0');
```
DHowett-MSFT: If you use unique_iw instead of std::vector, you can use `CFStringCreateWithBytesNoCopy` and detach it instead of copying the buffer.

Frameworks/CoreText/CTLine.mm
```diff
@@ -188,8 +210,9 @@ CTLineRef CTLineCreateTruncatedLine(CTLineRef sourceLine, double width, CTLineTr
         CFDictionaryRef attribs = CTRunGetAttributes(currentRun);
 
         if (truncationType == kCTLineTruncationStart) {
-            NSString* reverse = [runString _reverseString];
-            NSAttributedString* string = [[NSAttributedString alloc] initWithString:reverse attributes:(NSDictionary*)attribs];
+            auto reverse = woc::MakeStrongCF<CFStringRef>(_CTCreateReversedString(static_cast<CFStringRef>(runString)));
+            NSAttributedString* string =
+                [[NSAttributedString alloc] initWithString:static_cast<NSString*>(reverse.get()) attributes:(NSDictionary*)attribs];
```
DHowett-MSFT: ??? can't we use CFAttributedStringCreate?

msft-Jeyaram: I already have changes for all NSAttributed, this is strictly for the reverse.

Frameworks/CoreText/CTLine.mm
```diff
@@ -16,16 +16,38 @@
...
 #import <vector>
 
+static CFStringRef _CTCreateReversedString(CFStringRef string) {
+    if (string == nullptr) {
+        return nullptr;
+    }
+
+    CFIndex length = CFStringGetLength(string);
+    CFIndex usedBufLen;
+    CFStringGetBytes(string, CFRangeMake(0, length), kCFStringEncodingUTF16, 0, false, nullptr, length, &usedBufLen);
```
ms-jihua: Seems kind of expensive for this purpose (unless CFStringGetBytes has a fastpath here?)

I wonder if CFStringGetMaximumSizeForEncoding() would be faster/would still work?

msft-Jeyaram: Yes it does take a faster way than if you provide a buffer.
Further improvements can be addressed by a bug to improve core text. The CR is mainly to remove Foundation dependency and what we have now is better than the earlier Foundation/copy based implementation.

Frameworks/CoreText/CTLine.mm
```diff
@@ -16,16 +16,48 @@
...
 #import "CGContextInternal.h"
 #import "DWriteWrapper_CoreText.h"
 #import <CoreText/CTTypesetter.h>
 
 #import <algorithm>
 #import <numeric>
+#import <cwchar>
 #import <vector>
 
+static CFStringRef __CTCreateReversedString(CFStringRef string) {
```
aballway: how is this change being tested?

msft-Jeyaram: UnitTests

---
In reply to: [111007038](https://github.com/Microsoft/WinObjC/pull/2479#discussion_r111007038) [](ancestors = 111007038)

Frameworks/CoreText/CTLine.mm
```diff
@@ -16,16 +16,48 @@
...
+    CFIndex usedBufLen;
+    CFStringGetBytes(string, CFRangeMake(0, length), kCFStringEncodingUTF16, 0, false, nullptr, length, &usedBufLen);
+
+    if (length < 2) {
+        return CFStringCreateCopy(kCFAllocatorDefault, string);
+    }
+
+    CFIndex bufLen = usedBufLen + 1 * sizeof(UniChar);
+    woc::unique_iw<UniChar> characters(static_cast<UniChar*>(IwMalloc(bufLen)));
+    memset(characters.get(), '\0', bufLen);
```
DHowett-MSFT: c'mon, you only need to set the last character in the buffer

Frameworks/CoreText/CTLine.mm
```diff
@@ -16,16 +16,48 @@
...
+    if (result == nullptr) {
+        return nullptr;
+    }
+
+    return CFStringCreateWithBytesNoCopy(kCFAllocatorDefault,
+                                         reinterpret_cast<UInt8*>(characters.release()),
+                                         bufLen,
+                                         kCFStringEncodingUTF16,
+                                         false,
+                                         nullptr);
```
DHowett-MSFT: you will need to provide an allocator to delete this string; look at the implementation of `initWithBytesNoCopy:freeWhenDone:` for NSString, expecting freeWhenDone=YES

Frameworks/CoreText/CTLine.mm
msft-Jeyaram: 🤣  that was coming 😄

msft-Jeyaram:
@DHowett-MSFT please take a look

Frameworks/CoreText/CTLine.mm
msft-Jeyaram: Already passing in the nullptr for deallocation.
Switched to use characters, which does it for you.

Frameworks/CoreText/CTLine.mm
```diff
@@ -16,16 +16,46 @@
...
+static CFStringRef __CTCreateReversedString(CFStringRef string) {
+    if (string == nullptr) {
+        return nullptr;
+    }
+
+    CFIndex length = CFStringGetLength(string);
+    CFIndex usedBufLen;
+    CFStringGetBytes(string, CFRangeMake(0, length), kCFStringEncodingUTF16, 0, false, nullptr, length, &usedBufLen);
+
+    if (length < 2) {
```
aballway: nit: move this above CFStringGetBytes call

Frameworks/include/CFFoundationInternal.h
```diff
@@ -1,6 +1,6 @@
 //******************************************************************************
 //
-// Copyright (c) 2015 Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
```
aballway: These clean-ups, though nice, seem unrelated to this PR.  Just revert these files?

PR #2480, opened 2017-04-10
Change Drawing Tests to use Perceptibility Threshold

### LINKED ISSUE #2107: Root-cause antialiasing differences in text rendering on build machines

aballway:
Differences between versions of Windows are generally imperceptibly small, so tests have been changed to ignore these differences.  Some tests have a few pixels where the differences are slightly beyond our threshold, so they still have the failure threshold as well.

Some test images have been updated to reflect updated output which fell behind slightly due to thresholds.  

Fixes #2107


ms-jihua:
@DHowett-MSFT I neglected to mention this to you after you came back, but I have a proposal.

https://en.wikipedia.org/wiki/Color_difference I didn't read through all of it, but basically the 1976 primitive measurement of minimal perceptible color difference by the human eye is something like an RGBA difference of 2.3*256/100 = ~5.8.

Rather than shifting everything to mask, I think we should add a comparator that treats pixels as the same if they fall within an RGBA distance of... let's say half that, so 3. Thoughts?

aballway:
Updated because they were failing on dev box with blue failures.  I think a few were still failing w/ perceptual comparison but don't want out of date images ykwim

tests/unittests/CoreGraphics.drawing/ImageComparison.cpp
```diff
@@ -219,6 +219,30 @@ Pixel PixelComparisonModeMask<FailureThreshold>::ComparePixels(const LP& backgro
     return gp;
 }
 
+template <size_t FailureThreshold>
+template <typename LP, typename RP>
+Pixel PixelComparisonModeVisual<FailureThreshold>::ComparePixels(const LP& background, const LP& bp, const RP& cp, size_t& npxchg) {
```
ms-jihua: I feel that this name does not effectively convey what it actually checks for. 'Perceptible', 'HumanPerceptible', 'NoticeableDifference', or even 'Limen'/'DifferenceLimen' (if you want to be extra nerdy) would be more descriptive.

tests/unittests/CoreGraphics.drawing/ImageComparison.h
```diff
@@ -24,11 +24,14 @@ enum class ImageComparisonResult : unsigned int { Unknown = 0, Incomparable, Dif
...
     uint8_t r, g, b, a;
     bool operator==(const Pixel& o) const {
-        return r == o.r && g == o.g && g == o.g && a == o.a;
+        return r == o.r && g == o.g && g == o.g && b == o.b && a == o.a;
     }
     bool operator!=(const Pixel& o) const {
         return !(*this == o);
     }
+    bool near(const Pixel& o) const {
+        return (abs(r - o.r) + abs(g - o.g) + abs(b - o.b) + abs(a - o.a)) < 4;
```
ms-jihua: should add a comment about that threshold

tests/unittests/CoreGraphics.drawing/ImageComparison.cpp
ms-jihua: or 'CloseEnough' if you wanna be casual about it

tests/unittests/CoreGraphics.drawing/CGContextBlendModeTests.cpp
```diff
@@ -131,7 +131,8 @@ static void __drawRightStackPink(CGContextRef context, CGRect bounds) {
     CGContextRestoreGState(context);
 }
 
-class CGContextBlendMode : public WhiteBackgroundTest<>, public ::testing::WithParamInterface<::testing::tuple<bool, CGNamedBlendMode>> {
+class CGContextBlendMode : public WhiteBackgroundTest<PixelByPixelImageComparator<PixelComparisonModeDifferenceLimen<100>>>,
```
DHowett-MSFT: So blending still requires a threshold even with perceptual?

DHowett-MSFT:
Nit: reword everything in title and body to better reflect what we are now doing

tests/unittests/CoreGraphics.drawing/CGContextBlendModeTests.cpp
aballway: Mmhmm, there were between 1-75 pixels where the differences were beyond our threshold.

tests/unittests/CoreGraphics.drawing/ImageComparison.cpp
```diff
@@ -171,53 +171,79 @@ struct RGBAImageBuffer {
...
+        gp.r = gp.g = gp.b = 0;
+        gp.a = 255;
+    }
+
+    return gp;
+}
+
+template <size_t FailureThreshold>
+template <typename LP, typename RP>
+Pixel PixelComparisonModeDifferenceLimen<FailureThreshold>::ComparePixels(const LP& background,
```
ms-jihua: oh darnit you actually went with it

tests/unittests/CoreGraphics.drawing/ImageComparison.h
```diff
@@ -22,6 +22,8 @@
 enum class ImageComparisonResult : unsigned int { Unknown = 0, Incomparable, Different, Same };
 
 struct Pixel {
+    // Value at which pixel differences become perceptible
```
ms-jihua: I'm pretty sure that this isn't the actual value for just-noticeable-difference, just most likely under it. Could you change the comment to reflect that?

tests/unittests/CoreGraphics.drawing/CGContextBlendModeTests.cpp
ms-jihua: _how far_ beyond our threshold? related: do we want to be even more nice and change our difference images to reflect strength of difference between the two images? (possibly just change color coding to be abs(expected.r - actual.r) etc etc

tests/unittests/CoreGraphics.drawing/ImageComparison.cpp
ms-jihua: you totally should've used "EhCloseEnough" instead

tests/unittests/CoreGraphics.drawing/CGContextBlendModeTests.cpp
aballway: It was something like 4-5 off.  I don't think we need to get that granular with our difference images.

tests/unittests/CoreGraphics.drawing/ImageComparison.cpp
DHowett-MSFT: `PixelComparisonModeDifferenceIt_sFridayWhoCares`

PR #2493, opened 2017-04-11
Move tools/include/Starboard/ to a new xplat directory

### LINKED ISSUE #2444: OSX tests no longer build
Sorry for not realizing the implications earlier, but as a result of moving SmartTypes.h in #2419, OSX tests no longer build. I'm not certain what the correct fix here is, as the majority of headers in tools/include cannot be included on OSX.

ms-jihua:
Fixes #2444


pradipd:
Any particular reason for the change?  Did I break something by not having the xplat/startboard?

ms-jihua:
@pradipd see #2444

pradipd:
Bah humbug.  I knew I should have mirrored the whole directory structure.  
My fault.
What does the inlcude directory of the language package look like?

---
In reply to: [293397480](https://github.com/Microsoft/WinObjC/pull/2493#issuecomment-293397480) [](ancestors = 293397480)

pradipd:
Also, thanks for fixing this!

---
In reply to: [293399501](https://github.com/Microsoft/WinObjC/pull/2493#issuecomment-293399501) [](ancestors = 293399501,293397480)

ms-jihua:
Oh, hm. There is a Starboard _and_ a xplat/Starboard in there. Oops!

ms-jihua:
Language package dir structure is now the same as before.

tools/WinObjC.Language/WinObjC.Language.Packaging.targets
```diff
@@ -3,11 +3,12 @@
 
   <Target Name="AddBuiltOutput" BeforeTargets="GetPackageContents" Condition="'$(SkipGetPackageContentsForOtherPlatforms)' != 'true'">
       <ItemGroup>
-          <!-- Inlcude *all* of /include/  and /include_next/ folder.  -->
-          <_IncludeFiles Include="$(MSBuildThisFileDirectory)..\include\**\*.*"/>
+          <!-- Include *all* of /include/, /include/xplat/ and /include_next/ folder.  -->
+          <_IncludeFiles Include="$(MSBuildThisFileDirectory)..\include\**\*.*" Exclude="$(MSBuildThisFileDirectory)..\include\xplat\**\*.*"/>
+          <_IncludeFiles Include="$(MSBuildThisFileDirectory)..\include\xplat\**\*.*"/>
```
pradipd: Would it be cleaner to not to do this and leave the headers in the package as include\xplat\starboard\*.h?

this way there is consistency in how we include the headers.
So, consumers *in* the tool package use include\xplat to use the headers.
And, consumers *of( the tool package also use include\xplat.

Or something like that.

pradipd: Interesting apparently putting asteriks around words italicizes or bolds them.

---
In reply to: [111030228](https://github.com/Microsoft/WinObjC/pull/2493#discussion_r111030228) [](ancestors = 111030228)

DHowett-MSFT: @pradipd 
image


😉

ms-jihua: I'm conflicted because that seems like the more correct thing, but also would involve touching like 80 different files. At that point, the "add a SmartTypes.h that's just \#include \\<..\..\..\tools\include\Starboard\SmartTypes.h\> to where it used to be" solution starts seeming preferable.

pradipd: Most projects include Starboard.h and that includes Starboard\SmartTypes.h.
So, we should just have to change that to xplat\Starboard\smarttypes.h

And for the files that directly include <Starboard/SmartTypes.h>we change them to xplat/Starboard/SmartTypes.h



---
In reply to: [111036207](https://github.com/Microsoft/WinObjC/pull/2493#discussion_r111036207) [](ancestors = 111036207)

pradipd: I'm fine with either:
1. Leaving as is.
2. Modifying 80 files to use xplat/starboard/smarttypes.h

Please don't do the ..\..\..\tools\include\starboard.h because that would be including headers outside the package.


---
In reply to: [111221091](https://github.com/Microsoft/WinObjC/pull/2493#discussion_r111221091) [](ancestors = 111221091,111036207)

pradipd:
:shipit:

ms-jihua:
@pradipd I discussed with @DHowett-MSFT and decided on a different, smaller change. Let me know if you're ok with it.

pradipd:
Looks good.

---
In reply to: [293728227](https://github.com/Microsoft/WinObjC/pull/2493#issuecomment-293728227) [](ancestors = 293728227)

PR #2494, opened 2017-04-11
Fix CGCatalog Filters

MSFTFox:
Fix CGCatalog's project filters for SBResourceCopy on reference images.


DHowett-MSFT:
nit: reword to be in imperative mood

PR #2496, opened 2017-04-11
Remove xcassets from CTCatalog

### LINKED ISSUE #2270: CTCatalog is perpetually out of date
Attempting to launch CTCatalog from VS always causes VS to warn about CTCatalog being out of date and needing to be built. The following line from build logs looks suspect:

1>Project not up to date because build input 'C:\WinObjC\tests\testapps\CTCatalog\CTCatalog\Assets.xcassets' is missing. 

this line prints even though that dir exists:
image

aballway:
Prevents Visual Studio from prompting to build every time the app is run

Fixes #2270

### DIFF
--- a/tests/testapps/CTCatalog/CTCatalog.vsimporter/CTCatalog-WinStore10/CTCatalog.vcxproj
+++ b/tests/testapps/CTCatalog/CTCatalog.vsimporter/CTCatalog-WinStore10/CTCatalog.vcxproj
@@ -1,4 +1,4 @@
-<?xml version="1.0"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemGroup Label="ProjectConfigurations">
     <ProjectConfiguration Include="Debug|ARM">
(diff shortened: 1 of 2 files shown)


ms-jihua:
This'll probably fix the problem, but... didn't we get <AssetCatalogCompile Include="..\..\CTCatalog\Assets.xcassets" /> from vsimporter? Should there be a vsimporter change?

Additionally, what is the context on the filters change?

aballway:
Unsure why filters changed beyond remove Assets.xcassets, Visual Studio did that on its own...
XamlCatalog also has Assets.xcassets but doesn't face the same problem so I'm not sure from where the problem originates.

---
In reply to: [293412412](https://github.com/Microsoft/WinObjC/pull/2496#issuecomment-293412412) [](ancestors = 293412412)

ms-jihua:
Would you be willing to dig a little deeper? I believe that this is not the only app where I've seen this issue, and would like to know the root cause.

aballway:
Now XamlCatalog is having it every time I hit the play button, so idk why it wasn't happening before.
I'm not sure if Assets.xcassets is something that we should have in the .vcxproj.  If not, vsimporter will need to be fixed.  If it is, then it may be a problem with Visual Studio either modifying files or not recognizing it as a folder properly.  Or some combination of both.  @bbowman @memontic-ms 

---
In reply to: [293418223](https://github.com/Microsoft/WinObjC/pull/2496#issuecomment-293418223) [](ancestors = 293418223)

PR #2499, opened 2017-04-12
Implement NSProxy

### LINKED ISSUE #2348: Add NSProxy support
All of this class is entirely stubbed out, its time we had them for real.

ms-jihua:
Fixes #2348

### DIFF
--- a/Frameworks/Foundation/NSObject.mm
+++ b/Frameworks/Foundation/NSObject.mm
@@ -196,6 +196,9 @@ static id _NSWeakLoad(id obj) {
     return obj;
 }
 
+// clang format doesn't play well with 'class' here, so turn it off
+// clang-format off
+
 /**
  @Status Interoperable
 */
@@ -203,12 +206,14 @@ static id _NSWeakLoad(id obj) {
     return object_getClass(self);
 }
 
-    /**
-     @Status Interoperable
-    */
-    - (Class)superclass {
-    return class_getSuperclass(object_getClass(self));
+/**
+ @Status Interoperable
+*/
+- (Class)superclass {
+    return class_getSuperclass([self class]);
 }
+
+// clang-format on
 
 /**
  @Status Interoperable
@@ -557,7 +562,7 @@ static IMP _NSIMPForward(id object, SEL selector) {
  @Status Interoperable
 */
 + (id)self {
-    return (id)self;
+    return self;
 }
 
 /**
@@ -699,6 +704,7 @@ void WinObjC_SetMissingSelectorFatal(BOOL fatal) {
     _NSSelectorNotFoundIsNonFatal = !fatal;
 }
 
+// clang-format off
 #pragma region NSZombie Support
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wdeprecated-objc-isa-usage" // We are knowingly accessing ->isa directly.
@@ -707,8 +713,8 @@ __attribute__((objc_root_class)) @interface NSZombie {
 }
 @end
 
-    static void
-    _dealloc_zombify(id self, SEL _cmd); // forward declaration
+static void _dealloc_zombify(id self, SEL _cmd); // forward declaration
+// clang-format on
 
 static void _NSCFZombifyHook(id object) {
     _dealloc_zombify(object, nullptr);
--- a/Frameworks/Foundation/NSObject.mm
+++ b/Frameworks/Foundation/NSObject.mm
@@ -42,7 +42,8 @@
 #import "NSInvocationInternal.h"
 
 static void _NSObjCEnumerationMutation(id object) {
-    [NSException raise:NSInternalInconsistencyException format:@"Collection <%s %p> mutated while being enumerated!", object_getClassName(object), object];
+    [NSException raise:NSInternalInconsistencyException
+                format:@"Collection <%s %p> mutated while being enumerated!", object_getClassName(object), object];
 }
 
 static BOOL _NSSelectorNotFoundIsNonFatal;
@@ -258,28 +259,28 @@ static id _NSWeakLoad(id obj) {
  @Status Interoperable
 */
 - (id)performSelector:(SEL)selector {
-    return ((id (*)(id, SEL))objc_msgSend)(self, selector);
+    return ((id(*)(id, SEL))objc_msgSend)(self, selector);
 }
 
 /**
  @Status Interoperable
 */
 - (id)performSelector:(SEL)selector withObject:(id)obj1 {
-    return ((id (*)(id, SEL, id))objc_msgSend)(self, selector, obj1);
+    return ((id(*)(id, SEL, id))objc_msgSend)(self, selector, obj1);
 }
 
 /**
  @Status Interoperable
 */
 - (id)performSelector:(SEL)selector withObject:(id)obj1 withObject:(id)obj2 {
-    return ((id (*)(id, SEL, id, id))objc_msgSend)(self, selector, obj1, obj2);
+    return ((id(*)(id, SEL, id, id))objc_msgSend)(self, selector, obj1, obj2);
 }
 
 /**
  @Status Interoperable
 */
 - (id)performSelector:(SEL)selector withObject:(id)obj1 withObject:(id)obj2 withObject:(id)obj3 {
-    return ((id (*)(id, SEL, id, id, id))objc_msgSend)(self, selector, obj1, obj2, obj3);
+    return ((id(*)(id, SEL, id, id, id))objc_msgSend)(self, selector, obj1, obj2, obj3);
 }
 
 /**
@@ -504,7 +505,6 @@ static IMP _NSIMPForward(id object, SEL selector) {
     return _NSSlotForward(object, selector)->method;
 }
 
-
 /**
  @Status Interoperable
 */
@@ -677,12 +677,11 @@ static IMP _NSIMPForward(id object, SEL selector) {
 }
 
 /**
- @Status Stub
+ @Status Interoperable
  @Notes
 */
 - (BOOL)isProxy {
-    UNIMPLEMENTED();
-    return StubReturn();
+    return NO;
 }
 
 - (id)autoContentAccessingProxy {
@@ -708,7 +707,8 @@ __attribute__((objc_root_class)) @interface NSZombie {
 }
 @end
 
-static void _dealloc_zombify(id self, SEL _cmd); // forward declaration
+    static void
+    _dealloc_zombify(id self, SEL _cmd); // forward declaration
 
 static void _NSCFZombifyHook(id object) {
     _dealloc_zombify(object, nullptr);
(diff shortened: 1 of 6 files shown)


Frameworks/Foundation/NSProxy.mm
```diff
@@ -14,220 +14,317 @@
...
 }
 
 /**
- @Status Stub
+ @Status Interoperable
  @Notes
 */
 - (void)forwardInvocation:(NSInvocation*)anInvocation {
-    UNIMPLEMENTED();
+    [NSException raise:NSInvalidArgumentException
```
aballway: nit: should these just be NSInvalidAbstractInvocationReturn() ? I know it's not a class cluster per se, but it's still "Ya gotta override this"

Frameworks/Foundation/NSProxy.mm
```diff
@@ -14,220 +14,317 @@
...
- @Status Stub
+ @Status Interoperable
  @Notes
 */
 - (oneway void)release {
-    UNIMPLEMENTED();
+    intptr_t* refCount = ((intptr_t*)self) - 1;
+    // We allow refcounts to run into the negative, but should only
+    // deallocate once.
+    if (__sync_sub_and_fetch(refCount, 1) == -1) {
```
aballway: should this be <= ? Ideally we'd never go below -1 but in case we ever should due to async nonsense we wouldn't want to leak

Frameworks/Foundation/NSProxy.mm
ms-jihua: nope, documented NSInvalidArgumentException: https://developer.apple.com/reference/foundation/nsproxy/1416417-forwardinvocation?language=objc

Frameworks/Foundation/NSProxy.mm
ms-jihua: - this is just lifted from NSObject.mm
- async nonsense wouldn't be able to cause that since __sync_sub_and_fetch is atomic

Frameworks/Foundation/NSObject.mm
```diff
@@ -708,7 +707,8 @@ void WinObjC_SetMissingSelectorFatal(BOOL fatal) {
 }
 @end
 
-static void _dealloc_zombify(id self, SEL _cmd); // forward declaration
+    static void
```
DHowett-MSFT: nit: I'd prefer not having the format changes in this PR

Frameworks/Foundation/NSObject.mm
```diff
@@ -737,7 +737,7 @@ + (BOOL)isSubclassOfClass:(Class)classRef {
 /**
  @Status Interoperable
 */
-- (void)doesNotRecognizeSelector: (SEL)selector {
+- (void)doesNotRecognizeSelector:(SEL)selector {
```
DHowett-MSFT: then again, some of them are good ..

Frameworks/Foundation/NSProxy.mm
```diff
@@ -14,220 +14,317 @@
...
 
 /**
- @Status Stub
+ @Status Interoperable
  @Notes
 */
 - (NSMethodSignature*)methodSignatureForSelector:(SEL)aSelector {
-    UNIMPLEMENTED();
-    return StubReturn();
+    [NSException raise:NSInvalidArgumentException
```
DHowett-MSFT: nit: macro these, and use `%hs` + `__PRETTY_FUNCTION__` to get the method name in the macro

Frameworks/Foundation/NSProxy.mm
```diff
@@ -14,220 +14,317 @@
...
-     @Notes
-    */
-    - (Class) class {
-    UNIMPLEMENTED();
-    return StubReturn();
+/**
+ @Status Interoperable
+ @Notes
+*/
+- (Class)superclass {
```
DHowett-MSFT: Not sure on whether this should be calling `[self class]`; thoughts?

Frameworks/Foundation/NSProxy.mm
```diff
@@ -14,220 +14,317 @@
...
+}
+
+/**
+ @Status Interoperable
  @Notes
 */
 - (instancetype)self {
-    UNIMPLEMENTED();
-    return StubReturn();
+    return (id)self;
```
DHowett-MSFT: nit: cast necessary?

Frameworks/Foundation/NSProxy.mm
```diff
@@ -14,220 +14,317 @@
...
 
 /**
- @Status Stub
+ @Status Interoperable
  @Notes
 */
 - (BOOL)isKindOfClass:(Class)aClass {
-    UNIMPLEMENTED();
-    return StubReturn();
+    Class forwardClass = static_cast<Class>([self performSelector:@selector(class)]);
```
DHowett-MSFT: aaa is it really guaranteed that it calls `forwardInvocation:` for `class`?

DHowett-MSFT: i'm not sure I like that this is different from calling `class`

Frameworks/Foundation/NSProxy.mm
```diff
@@ -14,220 +14,317 @@
...
 
 /**
- @Status Stub
+ @Status Interoperable
  @Notes
 */
 - (id)performSelector:(SEL)aSelector withObject:(id)anObject {
-    UNIMPLEMENTED();
-    return StubReturn();
+    NSInvocation* invocation = [NSInvocation invocationWithMethodSignature:[self methodSignatureForSelector:aSelector]];
```
DHowett-MSFT: it's strange to me that these go through invocation instead of hitting the invocation-forwarding path

Frameworks/Foundation/NSProxy.mm
```diff
@@ -14,220 +14,317 @@
...
+ *  opt out of the fast path, and must implement both retain and release.
+ *
+ *  To actually be ARC compliant, we need to store our refcount sizeof(intptr_t) bytes before
+ *  self->isa, and mimic the runtime's behaviour for direct invocations of retain/release.
+ */
+- (void)_ARCCompliantRetainRelease {
 }
 
-@end
+@end
```
DHowett-MSFT: you may have to implement doesNotRecognizeSelector:, and call it from the default forwardInvocation: implementation just like NSObject -- that's how people will get missing selector exceptions

include/Foundation/NSProxy.h
```diff
@@ -22,19 +22,25 @@
...
 + (id)alloc;
 + (id)allocWithZone:(NSZone*)zone;
 - (void)dealloc;
 - (void)finalize;
 - (void)forwardInvocation:(NSInvocation*)anInvocation;
 - (NSMethodSignature*)methodSignatureForSelector:(SEL)aSelector;
 + (BOOL)respondsToSelector:(SEL)aSelector;
-+ (Class) class;
++ (Class)class;
++ (Class)superclass; // Not in doc, but is supported
```
DHowett-MSFT: Doesn't the NSObject protocol contain this?

tests/unittests/Foundation/NSProxyTests.mm
```diff
@@ -0,0 +1,161 @@
...
+- (NSUInteger)count;
+- (id)objectAtIndex:(NSUInteger)index;
+- (void)addObject:(id)anObject;
+- (void)insertObject:(id)anObject atIndex:(NSUInteger)index;
+@end
+
+@implementation _NSTestMutableArrayProxy
+
+- (instancetype)initWithMutableArray:(NSMutableArray*)array {
+    self->_forwardArray = array;
```
DHowett-MSFT: nit: document not calling super init

Frameworks/Foundation/NSProxy.mm
ms-jihua: it's the contract unfortunately :( [NSProxy class] does not return a forwarded class

include/Foundation/NSProxy.h
ms-jihua: it does not!

tests/unittests/Foundation/NSProxyTests.mm
ms-jihua: can't/shouldn't call super init on NSProxy - it's NS_UNAVAILABLE (this matches the reference platform)

Frameworks/Foundation/NSProxy.mm
ms-jihua: what do you mean?

DHowett-MSFT: It seems like `performSelector:` and friends _could_ just call objc_msgSend; that would go through the invocation forwarding and capture codepath

It's less verbose, but perhaps there is an advantage to creating the invocation directly!

tests/unittests/Foundation/NSProxyTests.mm
DHowett-MSFT: Indeed! It'd be good to have a comment in here so somebody doesn't think it's broken.
Our CC (or that one really didactic talk I gave) states that we should _almost never_ not call init. Since this is one of the cases where we mustn't, it deserves callout!

Frameworks/Foundation/NSObject.mm
ms-jihua: i'll fix this one and leave the rest, if that's ok

Frameworks/Foundation/NSProxy.mm
ms-jihua: This is also the same as NSObject.mm... going to propose we leave both for the moment and let sleeping beasts lie.

ms-jihua: I changed my mind and got curious how the reference platform behaved and it appears superclass is polymorphic so I will make this change in both places,

aballway: May want to do the fix for NSObject in a separate PR.  Doubt it would cause any regressions but it's foundational enough that it might and should be thoroughly tested.

Frameworks/Foundation/NSProxy.mm
ms-jihua: This is already tested for in the unit test - I believe it forwards to the remote doesNotRecognizeSelector.

Frameworks/Foundation/NSProxy.mm
ms-jihua: discussed offline, leaving as is

Frameworks/Foundation/NSProxy.mm
ms-jihua: seems not

Frameworks/Foundation/NSProxy.mm
ms-jihua: Mm, my thought is that if a change like that breaks something, we'll either be able to catch it in CI build, or it's so incredibly subtle despite being in such an exercised class that doing it in a separate change won't really gain us anything, so I'll elect to risk it. (I will probably regret saying this a week from now)

ms-jihua:
Could not repro CI build failure locally, queuing again.

PR #2505, opened 2017-04-12
Convert CoreMotion to C++/WinRT

bviglietta:
Convert CoreMotion to C++/WinRT

### DIFF
--- a/Frameworks/CoreMotion/CMMotionManager.mm
+++ b/Frameworks/CoreMotion/CMMotionManager.mm
@@ -18,15 +18,20 @@
 #import <CoreMotion/CoreMotion.h>
 #import <StubReturn.h>
 #import <NSLogging.h>
+#import <chrono>
 #import "CMMotionManagerInternal.h"
 #import "UWP/WindowsGraphicsDisplay.h"
 #import "windows.h"
+#import "CppWinRTHelpers.h"
 
-// Timestamps for readings are in seconds for iOS and in 100-nanoseconds for WinRT
-static const double c_secondToHundredNanoseconds = 10000000.0;
+#include "COMIncludes.h"
+#import <winrt/Windows.Foundation.h>
+#import <winrt/Windows.Graphics.Display.h>
+#include "COMIncludes_End.h"
 
-// GetTickCount64() unit is millisecond and SystemTime unit is 100-nanosecond
-static const double c_millisecondToHundredNanoseconds = 10000.0;
+using namespace winrt::Windows::Devices::Sensors;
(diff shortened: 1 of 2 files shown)


bviglietta:

Frameworks/CoreMotion/CMMotionManager.mm
```diff
@@ -37,123 +42,108 @@
...
-        ULARGE_INTEGER nowTime;
-        nowTime.LowPart = fileTime.dwLowDateTime;
-        nowTime.HighPart = fileTime.dwHighDateTime;
-
-        // On iOS the timestamps represent the number of seconds passed since the device booted while
-        // WinRT returns the number of 100-nanosecond intervals passed after midnight on January 1, 1601.
-        // Hence we calculate the bootTime as number of 100-nanosecond intervals after midnight on January 1, 1601
-        // till device boot and then we can simply subtract this value from each readings' timestamp.
-        _bootTime = static_cast<double>(nowTime.QuadPart) - (_bootTime * c_millisecondToHundredNanoseconds);
+        ULONGLONG ticks = GetTickCount64();
```
aballway: nit: unsigned long long int vs. shouty

bviglietta: Normally yes, but GetTickCount64 is declared as returning ULONGLONG, so I think it's best to conform.

---
In reply to: [111271794](https://github.com/Microsoft/WinObjC/pull/2505#discussion_r111271794) [](ancestors = 111271794)

Frameworks/CoreMotion/CMMotionManager.mm
```diff
@@ -190,18 +180,18 @@ - (void)startAccelerometerUpdatesToQueue:(NSOperationQueue*)queue withHandler:(C
         self.accelerometerQueue = queue;
 
         self.accelerometerToken =
-            [self.accelerometer addReadingChangedEvent:^void(WDSAccelerometer* sender, WDSAccelerometerReadingChangedEventArgs* e) {
```
DHowett-MSFT: Will the C++ lambda here extend the lifetime of queue if it gets released before the reading callback fires? I know blocks will explicitly do so.

This could be a concern for all async objective-c things, and necessitate the use of AutoId/StrongId with its copy constructor to ensure object lifetimes.

bviglietta: When ARC is enabled (as it is here), objects captured by lambdas should be retained.

---
In reply to: [111273056](https://github.com/Microsoft/WinObjC/pull/2505#discussion_r111273056) [](ancestors = 111273056)

PR #2510, opened 2017-04-13
Convert GLKit to C++/WinRT

bviglietta:
Convert GLKit to C++/WinRT

### DIFF
--- a/Frameworks/GLKit/GLKViewController.mm
+++ b/Frameworks/GLKit/GLKViewController.mm
@@ -34,7 +34,7 @@ namespace WF = winrt::Windows::Foundation;
 static NSTimeInterval timeSpanToInterval(const WF::TimeSpan& timeSpan) {
     // NSTimeInterval is in seconds
     auto millis = std::chrono::duration_cast<std::chrono::milliseconds>(timeSpan);
-    return millis / 1000.0;
+    return millis.count() / 1000.0;
 }
 
 @protocol _GLKViewControllerInformal <NSObject>
--- a/Frameworks/GLKit/GLKViewController.mm
+++ b/Frameworks/GLKit/GLKViewController.mm
@@ -18,11 +18,24 @@
 #import <GLKit/GLKitExport.h>
 #import <GLKit/GLKView.h>
 #import <GLKit/GLKViewController.h>
-
-#import <UWP/WindowsFoundation.h>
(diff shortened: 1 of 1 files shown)


Frameworks/GLKit/GLKViewController.mm
```diff
@@ -91,13 +101,12 @@ - (void)viewWillDisappear:(BOOL)animated {
...
-    int64_t frameTime = dt.universalTime;
+    _calendar.SetToNow();
+    WF::DateTime frameTime = _calendar.GetDateTime();
 
-    _timeSinceFirstResume = static_cast<NSTimeInterval>(frameTime - _firstStart) * 0.0000001;
-    _timeSinceLastResume = static_cast<NSTimeInterval>(frameTime - _lastStart) * 0.0000001;
-    _timeSinceLastUpdate = static_cast<NSTimeInterval>(frameTime - _lastFrame) * 0.0000001;
+    _timeSinceFirstResume = timeSpanToInterval(frameTime - _firstStart);
+    _timeSinceLastResume = timeSpanToInterval(frameTime - _lastStart);
+    _timeSinceLastUpdate = timeSpanToInterval(frameTime - _lastFrame);
```
DHowett-MSFT: I don't understand why this needs WinRT at all; can't it use the `std::high_resolution_clock`? It seems like using the system's localized clock will introduce interesting rendering issues as the user travels across spacetime.

bviglietta: I agree that use of WinRT is not really required here.

I'm not sure I understand the concern about the system clock, though. Are you suggesting that the interval calculation would be wrong if the user moves, e.g., into a different time zone while the system is running?

DHowett-MSFT: Indeed. Since we're using the calendar from `Windows.Globalization`, it should be primed with the user's timezone and location; we're not, as far as I can tell, compensating for distance from UTC or anything.

bviglietta: The timestamps are offsets from the Windows epoch, which is fixed, so I'd be surprised if localization affected them at all.

I haven't done any serious tests on this, though. I just translated the code as-is.

DHowett-MSFT: Ah, that's an interesting feature of the API. Let's file a bug to reevaluate its use here!

PR #2512, opened 2017-04-13
Fix failing tests on reference platform

### LINKED ISSUE #1725: NSBundle.Paths test fails on OSX
[ RUN      ] NSBundle.Paths
ReferenceFoundation/TestNSBundle.mm:48: Failure
Value of: nullptr
  Actual: NULL
Expected: bundle.appStoreReceiptURL
Which is: _MASReceipt/receipt -- file:///Volumes/WinObjC/build/Tests/UnitTests/Foundation/OSX/
[  FAILED  ] NSBundle.Paths (10 ms)

aballway:
Removes a couple of tests from ReferenceFoundation that are no longer passing.
NSSet enumerateWithOptions will not always respect NSEnumerateConcurrent so can cause a deadlock 🐙 
NSURLSessionConfiguration TLSMaximumSupportedProtocol has been updated on the reference platform to reflect documented value
NSURL URLByDeletingPathExtension has a bug on the reference platform where it will improperly delete non-pathextensions.
Another test failed because the file it was looking for is in a different directory on OSX.
NSFileManagerTests was using a Lstring literal as a w_char* to create an NSString which failed on the reference platform.

Fixes #1725

### DIFF
--- a/tests/unittests/Foundation/NSFileHandleTestFile.txt
+++ b/tests/unittests/Foundation/NSFileHandleTestFile.txt
@@ -1,1 +1,0 @@
-The Quick Brown Fox.
--- a/tests/unittests/Foundation/NSFileManagerTests.mm
+++ b/tests/unittests/Foundation/NSFileManagerTests.mm
@@ -210,8 +210,7 @@ TEST(NSFileManager, MoveFileViaURL) {
 }
 
 TEST(NSFileManager, DirectoryWithUTF16Chars) {
-    wchar_t* specialFolder = L"oÖo winobjc";
-    NSString* directoryName = [NSString stringWithCharacters:(const unichar*)specialFolder length:wcslen(specialFolder) * sizeof(wchar_t)];
+    NSString* directoryName = @"oÖo winobjc";
     NSString* testPath = getPathToFile(directoryName);
 
     EXPECT_TRUE([[NSFileManager defaultManager] createDirectoryAtPath:testPath attributes:nil]);
--- a/tests/unittests/Foundation/NSSetTests.mm
+++ b/tests/unittests/Foundation/NSSetTests.mm
@@ -192,7 +192,9 @@ TEST(NSSet, MakeObjectsPerformSelector) {
     EXPECT_TRUE([otherSet containsObject:@6]);
 }
 
-TEST(NSSet, ObjectsWithOptionsPassingTest) {
+// Reference platform is not guaranteed to run concurrently so this can fail
+// Still keep to guarantee properly support concurrency
+OSX_DISABLED_TEST(NSSet, ObjectsWithOptionsPassingTest) {
     NSSet* set = [NSSet setWithObjects:@1, @2, @3, @4, @5, nil];
 
     // Verify that the NSEnumerationConcurrent option executes the blocks concurrently
--- a/tests/unittests/Foundation/NSURLTests.m
+++ b/tests/unittests/Foundation/NSURLTests.m
@@ -206,7 +206,8 @@ TEST(NSURL, URLByAppendingPathExtension) {
     ASSERT_OBJCEQ_MSG(fileURLString, newFileURLString, "File URLs do not match!");
 }
 
-TEST(NSURL, CheckResourceIsReachable) {
+// Reference platform has bug on some machines where currentDirectoryPath returns the incorrect path
+OSX_DISABLED_TEST(NSURL, CheckResourceIsReachable) {
     NSFileManager* manager = [NSFileManager defaultManager];
     NSString* originalPath = [manager currentDirectoryPath];
     // construct target URL using current directory and relative URL
--- a/tests/unittests/Foundation/NSURLTests.m
+++ b/tests/unittests/Foundation/NSURLTests.m
@@ -213,9 +213,7 @@ TEST(NSURL, CheckResourceIsReachable) {
     EXPECT_TRUE_MSG([targetURL checkResourceIsReachableAndReturnError:nullptr], "The target URL %@ exists", targetURL);
 
     NSURL* targetURLNonExist = [NSURL URLWithString:@"data/foo.txt" relativeToURL:baseURL];
-    EXPECT_FALSE_MSG([targetURLNonExist checkResourceIsReachableAndReturnError:nullptr],
-                     "The target %@URL does not exist",
-                     targetURLNonExist);
+    EXPECT_FALSE_MSG([targetURLNonExist checkResourceIsReachableAndReturnError:nullptr], "The target %@ does not exist", targetURLNonExist);
 }
 
 TEST(NSURL, GetFileSystemRepresentation) {
--- a/tests/unittests/Foundation/NSURLTests.m
+++ b/tests/unittests/Foundation/NSURLTests.m
@@ -1,6 +1,6 @@
 //******************************************************************************
 //
-// Copyright (c) 2015 Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
 //
 // This code is licensed under the MIT License (MIT).
 //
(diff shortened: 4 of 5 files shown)


tests/unittests/Foundation/NSFileManagerTests.mm
```diff
@@ -210,8 +210,7 @@
 }
 
 TEST(NSFileManager, DirectoryWithUTF16Chars) {
-    wchar_t* specialFolder = L"oÖo winobjc";
-    NSString* directoryName = [NSString stringWithCharacters:(const unichar*)specialFolder length:wcslen(specialFolder) * sizeof(wchar_t)];
+    NSString* directoryName = @"oÖo winobjc";
```
DHowett-MSFT: l o l

tests/unittests/Foundation/NSSetTests.mm
```diff
@@ -195,27 +195,12 @@
 TEST(NSSet, ObjectsWithOptionsPassingTest) {
     NSSet* set = [NSSet setWithObjects:@1, @2, @3, @4, @5, nil];
 
-    // Verify that the NSEnumerationConcurrent option executes the blocks concurrently
-    // The blocks will increment waitingCount and wait for the signal unless it is the last element,
-    // in which case it will signal the other blocks to continue and they all decrement waitingCount
-    // so all blocks should be run concurrently, otherwise waitingCount will not return to 0
-    __block NSCondition* condition = [[NSCondition new] autorelease];
-    __block unsigned waitingCount = 0;
+    // Reference platform is not guaranteed to respect NSEnumerationConcurrent, so tests will sometimes fail
```
DHowett-MSFT: you could mark this OSX_DISABLED_TEST?

tests/unittests/Foundation/ReferenceFoundation/TestNSBundle.mm
```diff
@@ -44,7 +44,6 @@
     // #endif
 
     ASSERT_OBJCEQ([bundle pathForAuxiliaryExecutable:@"no_such_file"], nil);
-    ASSERT_OBJCEQ(bundle.appStoreReceiptURL, nil);
```
DHowett-MSFT: please note when you've changed reference tests.

tests/unittests/Foundation/ReferenceFoundation/TestNSURL.mm
```diff
@@ -331,15 +331,6 @@ static boolean setup_test_paths() {
                 modifiedExpectedNSResult = [static_cast<NSDictionary<NSString*, id>*>(expectedNSResult) mutableCopy];
                 modifiedExpectedNSResult[@"standardizedURL"] = nil;
             }
-            if ([title isEqualToString:@"NSURLWithString-parse-ambiguous-url-001"]) {
```
DHowett-MSFT: _ut supra_

tests/unittests/Foundation/NSSetTests.mm
aballway: Possibly but we don't need to force concurrency if it's undefined, so it's not worth testing IMO

tests/unittests/Foundation/NSURLTests.m
```diff
@@ -23,154 +23,169 @@
...
 
 TEST(NSURL, NSURLTests) {
-    NSURL* testURL = [[NSURL alloc] initWithString:@"http://www.test.com/home/index.html?Foo#24"];
-    NSURL* testURL2 = [[NSURL alloc] initWithString:@"http://www.test.com/"];
-    NSURL* testURL3 = [[NSURL alloc] initWithString:@"http://www.test.com/home/asdf/./index.html?Foo#24"];
-    NSURL* testURL4 = [[NSURL alloc] initWithString:@"file://www.test.com/home/../home/./index.html?Foo#24"];
-    NSURL* testURL5 = [[NSURL alloc] initWithString:@"http://www.test.com?Foo#24"];
-    NSURL* testURL6 = [[NSURL alloc] initWithString:@"http://www.test.com/home.page/asdf"];
-    NSURL* testURL7 = [[NSURL alloc] initWithString:@"http://www.test.com/.foo"];
+    NSURL* testURL = [[[NSURL alloc] initWithString:@"http://www.test.com/home/index.html?Foo#24"] autorelease];
```
ms-jihua: nit: URLWithString

tests/unittests/Foundation/NSURLTests.m
```diff
@@ -23,154 +23,169 @@
...
+
+    TEST_NSURL_METHOD(@selector(URLByAppendingPathExtension:),
+                      testURL6,
+                      extension,
+                      [[[NSURL alloc] initWithString:@"http://www.test.com/home.page/asdf.txt"] autorelease]);
+
+    TEST_NSURL_METHOD(@selector(URLByAppendingPathExtension:),
+                      testURL7,
+                      extension,
+                      [[[NSURL alloc] initWithString:@"http://www.test.com/.foo.txt"] autorelease]);
```
ms-jihua: was the memory management the only thing that changed in this test?

aballway: No there was a URLByDeletingPathExtension that changed on the reference platform.  I only fixed the memory because I noticed it while removing that one.

tests/unittests/Foundation/NSURLTests.m
```diff
@@ -225,16 +240,20 @@ void testNSURLMethod(SEL selector, NSURL* input, id argument, NSURL* expected) {
...
     ASSERT_TRUE(SetCurrentDirectoryW(startUpPath) != 0);
+    NSURL* baseURL = [NSURL fileURLWithPath:[manager currentDirectoryPath]];
+    NSURL* targetURL = [NSURL URLWithString:@"data/NSFileManagerUT.txt" relativeToURL:baseURL];
 #else
     char tempBuffer[_MAX_PATH];
     wcstombs(tempBuffer, startUpPath, _MAX_PATH);
 
     ASSERT_TRUE(chdir(dirname(tempBuffer)) == 0);
+    NSURL* baseURL = [NSURL fileURLWithPath:[manager currentDirectoryPath]];
+    NSURL* targetURL = [NSURL URLWithString:@"OSX/data/NSFileManagerUT.txt" relativeToURL:baseURL];
```
ms-jihua: Concerned about this actually... currentDirectoryPath should be returning build/Tests/UnitTests/Foundation/OSX, as that's the directory the executable (Foundation.UnitTests.exe) is in. Why isn't it?

ms-jihua: worth noting also that this is currently not failing on my machine... unless for some reason in a future release of Foundation they made currentDirectoryPath return the dir one up? that seems really unlikely though...

tests/unittests/Foundation/ReferenceFoundation/TestNSBundle.mm
```diff
@@ -44,7 +44,8 @@
...
+
+    // WINOBJC: Remove test that no longer passes on reference platform
```
ms-jihua: could you leave the original code here, commented out? that way it's easier to compare with the reference

tests/unittests/Foundation/ReferenceFoundation/TestNSURL.mm
```diff
@@ -331,15 +331,9 @@ static boolean setup_test_paths() {
...
-                // TODO: Fix this test
-            } else {
-                auto results = generateResults(url, inPathComponent, inPathExtension);
-                std::pair<BOOL, NSArray*> compare = compareResults(url, modifiedExpectedNSResult, results);
-                BOOL isEqual = compare.first;
-                NSArray* differences = compare.second;
-                ASSERT_TRUE_MSG(isEqual, @"%@: %@", title, differences);
-            }
+
+            // WINOBJC: Remove test that no longer passes on reference platform
```
ms-jihua: same

tests/unittests/Foundation/NSURLTests.m
aballway: Apparently my machine needs an update.  This could be a bug in the version I'm running, in which case I'll revert this.

tests/unittests/Foundation/NSSetTests.mm
DHowett-MSFT: I do not understand what this means. Wouldn't disabling the test on OS X ensure that it is not forced when undefined?

aballway:
Updated machine to Sierra, now fewer tests fail.  Still have currentDirectoryPath returning the incorrect path so I've disabled that test on OSX.

tests/unittests/Foundation/NSURLTests.m
```diff
@@ -206,7 +206,8 @@ void testNSURLMethod(SEL selector, NSURL* input, id argument, NSURL* expected) {
     ASSERT_OBJCEQ_MSG(fileURLString, newFileURLString, "File URLs do not match!");
 }
 
-TEST(NSURL, CheckResourceIsReachable) {
+// Reference platform has bug on some machines where currentDirectoryPath returns the incorrect path
```
DHowett-MSFT: it literally is _the path to the current directory_; if this test is based on the folder from which you're running it, the test is incorrect.

aballway: It's returning neither the directory from which the test is being run nor the directory the executable is in

DHowett-MSFT: This entire test is awful and should be shot.

It manipulates global state (current directory) to do something it could simply do from the base path returned by `GetCurrentTestDirectory`.

DHowett-MSFT: Can you just rewrite it to not be bad?

tests/unittests/Foundation/ReferenceFoundation/TestNSBundle.mm
```diff
@@ -44,7 +44,9 @@
     // #endif
 
     ASSERT_OBJCEQ([bundle pathForAuxiliaryExecutable:@"no_such_file"], nil);
-    ASSERT_OBJCEQ(bundle.appStoreReceiptURL, nil);
+
+    // WINOBJC: Disable because no longer correct on reference platform
+    //ASSERT_OBJCEQ(bundle.appStoreReceiptURL, nil);
```
msft-Jeyaram: nit: hmm did this test get an update from the original authors?

tests/unittests/Foundation/NSURLTests.m
```diff
@@ -206,42 +206,16 @@ void testNSURLMethod(SEL selector, NSURL* input, id argument, NSURL* expected) {
...
     NSURL* targetURL = [NSURL URLWithString:@"data/NSFileManagerUT.txt" relativeToURL:baseURL];
-    ASSERT_TRUE_MSG([targetURL checkResourceIsReachableAndReturnError:nullptr], "The target URL %@ exists", targetURL);
+    EXPECT_TRUE_MSG([targetURL checkResourceIsReachableAndReturnError:nullptr], "The target URL %@ exists", targetURL);
 
     NSURL* targetURLNonExist = [NSURL URLWithString:@"data/foo.txt" relativeToURL:baseURL];
-    ASSERT_FALSE_MSG([targetURLNonExist checkResourceIsReachableAndReturnError:nullptr],
+    EXPECT_FALSE_MSG([targetURLNonExist checkResourceIsReachableAndReturnError:nullptr],
                      "The target %@URL does not exist",
                      targetURLNonExist);
-    ASSERT_TRUE([manager changeCurrentDirectoryPath:originalPath]);
```
DHowett-MSFT: we were literally manually setting the cwd 10 lines above this, and then using the API for it down here? 😿

tests/unittests/Foundation/NSURLTests.m
```diff
@@ -206,42 +206,16 @@ void testNSURLMethod(SEL selector, NSURL* input, id argument, NSURL* expected) {
...
-    NSURL* baseURL = [NSURL fileURLWithPath:[manager currentDirectoryPath]];
+    NSURL* baseURL = [NSURL fileURLWithPath:[NSString stringWithUTF8String:fullPath.c_str()]];
     NSURL* targetURL = [NSURL URLWithString:@"data/NSFileManagerUT.txt" relativeToURL:baseURL];
-    ASSERT_TRUE_MSG([targetURL checkResourceIsReachableAndReturnError:nullptr], "The target URL %@ exists", targetURL);
+    EXPECT_TRUE_MSG([targetURL checkResourceIsReachableAndReturnError:nullptr], "The target URL %@ exists", targetURL);
 
     NSURL* targetURLNonExist = [NSURL URLWithString:@"data/foo.txt" relativeToURL:baseURL];
-    ASSERT_FALSE_MSG([targetURLNonExist checkResourceIsReachableAndReturnError:nullptr],
+    EXPECT_FALSE_MSG([targetURLNonExist checkResourceIsReachableAndReturnError:nullptr],
                      "The target %@URL does not exist",
```
DHowett-MSFT: nit: `%@URL` will format hilariously.

tests/unittests/Foundation/NSFileHandleTestFile.txt
```diff
@@ -1 +0,0 @@
-The Quick Brown Fox.
```
ms-jihua: ? why is this gone?

msft-Jeyaram: hmm i don't think this is being used :0

---
In reply to: [111624327](https://github.com/Microsoft/WinObjC/pull/2512#discussion_r111624327) [](ancestors = 111624327)

aballway: The file was being generated and then deleted by a few tests.  It probably was accidentally left at some point, but it doesn't seem to be needed.

PR #2518, opened 2017-04-13
Switch member initialization from MakeStrongCF to TakeOwnership

msft-Jeyaram:


msft-Jeyaram:
@DHowett-MSFT  but you must TakeOwnership of thy work ;P

DHowett-MSFT:
:clap:

PR #2519, opened 2017-04-13
Change NSOperation to dealloc safely without init

### LINKED ISSUE #2490: NSOperations throw on dealloc if -init is not called
This damages the ability for NSOperation subclasses to release themselves on the init path (such as in NSInvocationOperation, where the operation must be deleted and nil returned if `initWithTarget:...` receives bad parameters.)

ms-jihua:
- Only removeObserver: if addObserver: was actually called
- Remove an unnecessary mutex locking in dealloc

Fixes #2490

### DIFF
--- a/Frameworks/Foundation/NSOperation.mm
+++ b/Frameworks/Foundation/NSOperation.mm
@@ -57,6 +57,8 @@ static wchar_t TAG[] = L"NSOperation";
 
     std::atomic<NSOperationQueuePriority> _queuePriority;
     std::atomic<BOOL> _inQueue;
+
+    bool _observingSelfIsFinished;
 }
 
 @synthesize cancelled = _cancelled;
@@ -143,6 +145,7 @@ static const NSString* NSOperationContext = @"context";
         _dependencies.attach([NSMutableSet new]);
         _ready = YES;
         [self addObserver:self forKeyPath:@"isFinished" options:0 context:(void*)NSOperationContext];
+        _observingSelfIsFinished = true;
     }
 
     return self;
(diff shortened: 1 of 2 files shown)


Frameworks/Foundation/NSOperation.mm
```diff
@@ -57,6 +57,8 @@ @implementation NSOperation {
 
     std::atomic<NSOperationQueuePriority> _queuePriority;
     std::atomic<BOOL> _inQueue;
+
+    bool _observingSelfIsFinished;
```
jaredhms: >_observingSelfIsFinished [](start = 9, length = 24)

nit: maybe just call this _isInitialized?

Frameworks/Foundation/NSOperation.mm
```diff
@@ -344,14 +347,14 @@ - (NSArray*)dependencies {
  @Status Interoperable
 */
 - (void)dealloc {
-    { // _dependenciesLock scope
-        std::lock_guard<std::recursive_mutex> lock(_dependenciesLock);
```
jaredhms: >std::lock_guard<std::recursive_mutex> lock(_dependenciesLock); [](start = 8, length = 62)

why isn't this lock necessary anymore?

Frameworks/Foundation/NSOperation.mm
ms-jihua: I had considered this, but I prefer being more specific here. Leaving as-is.

Frameworks/Foundation/NSOperation.mm
ms-jihua: It wasn't ever. Shouldn't need to lock anything in dealloc, by which time no one should be referencing the object anymore.

Frameworks/Foundation/NSOperation.mm
```diff
@@ -344,14 +347,14 @@ - (NSArray*)dependencies {
...
-        for (NSOperation* op in (NSSet*)_dependencies) {
-            [op removeObserver:self forKeyPath:@"isFinished" context:(void*)NSOperationContext];
-        }
+    for (NSOperation* op in (NSSet*)_dependencies) {
+        [op removeObserver:self forKeyPath:@"isFinished" context:(void*)NSOperationContext];
+    }
```
jaredhms: nit: don't need to do this either if you haven't initialized.  also, a nil _dependencies *could* signify that you're not initialized.

jaredhms:
:shipit:

Frameworks/Foundation/NSOperation.mm
ms-jihua: this ends up being a no-op if _dependencies is nil, so it's redundant to check for it. additionally, I would prefer to keep the check strongly tied to the self-observation, as it's the only thing that's actually causing problems

PR #2520, opened 2017-04-14
Add missing import to NSFileManager.h to fix ApiAnalyzer

asimonov-msft:


aballway:
> @class NSDictionary<KeyType, ObjectType>;

Now that this is being included there's no need for this forward declaration

---
Refers to: include/Foundation/NSFileManager.h:26 in 350786f. [](commit_id = 350786fa78d460d032b6b46247c4b0e5d27287ac, deletion_comment = False)

include/Foundation/NSFileManager.h
```diff
@@ -18,6 +18,7 @@
 #import <Foundation/FoundationExport.h>
 #import <Foundation/NSObject.h>
 #import <Foundation/NSPathUtilities.h>
+#import <Foundation/NSDictionary.h>
```
pradipd: Why is libclang choking when this isn't here?
Is this new?
Presumably this file builds in build.sln. 
So, it is going to be very hard to know when apianalyzer is going to break #import stuff vs. @class

DHowett-MSFT: @pradipd libclang is used to parse these headers in isolation. We don't have any apps that consume each header individually, so the problem gets hidden by the include order.

It is usually _important_ that each header stand alone, but it is only a _requirement_ because of libclang

pradipd: Ok. Thanks.  I guess I find it amazing that this is the first apianalyzer failure of this kind that I can remember.

---
In reply to: [111599749](https://github.com/Microsoft/WinObjC/pull/2520#discussion_r111599749) [](ancestors = 111599749)

pradipd:
:shipit:

aballway:
:shipit:

PR #2522, opened 2017-04-14
Fix #1983 and fix in RTHelpers.h

### LINKED ISSUE #1983: Projections expose protected interfaces
The UWP projections expose protected interfaces, which are not intended to be used by client code.

For example, the projected XAML Control class and all its subclasses expose the IControlProtected interface, which has the [protected] attribute and therefore shouldn't be accessible. I should not be able to write this code:

WXCButton* button = [WXCButton make];
[button getTemplateChild:@"name"];    // Bad! This method should not be visible

pradipd:


bviglietta:
The title refers to a fix in RTHelpers.h, but there's no change to RTHelpers.h in this pull request?

bviglietta:
Oh wait, I get it. These are projection binaries that have been rebuilt with a previous fix to RTHelpers.h?

---
In reply to: [294205085](https://github.com/Microsoft/WinObjC/pull/2522#issuecomment-294205085) [](ancestors = 294205085)

pradipd:
There is a change to rthelpers.h.  You need to use the github because codeflow can't handle such large reviews.

---
In reply to: [294205679](https://github.com/Microsoft/WinObjC/pull/2522#issuecomment-294205679) [](ancestors = 294205679,294205085)

include/Platform/Universal Windows/UWP/RTHelpers.h
```diff
@@ -893,22 +893,19 @@ static HString nsStrToHstr(NSString* str) {
         // We increment this just to make sure that we return an empty HString.
         dataBytes++;
 
-        // Include size of the additional terminating '\0' character.
-        data.reserve(dataBytes + 1);
+        // Allocate a vector containing only the null character.
+        data.resize(dataBytes, L'\0');
```
bviglietta: "dataBytes" is an exceedingly bad name for a count of characters.

include/Platform/Universal Windows/UWP/RTHelpers.h
```diff
@@ -893,22 +893,19 @@ static HString nsStrToHstr(NSString* str) {
...
+        [str getCharacters : data.data()];
+        
+        // Add terminating NULL character.
+        data[dataBytes] = L'\0';
     }
 
-    // Add terminating NULL character.
-    data[dataBytes] = L'\0';
-
     HRESULT hr = WindowsCreateString((LPWSTR)&data[0], dataBytes, ret.GetAddressOf());
```
bviglietta: It looks like, when the source string is empty, we're creating a length-1 HSTRING containing a single embedded NUL character. That doesn't seem like the right behavior, since that's not the same as an empty string. In particular, WindowsIsStringEmpty will return FALSE for such a string, and WindowsGetStringLen will return 1.

I see the comment that says we're trying to avoid having WindowsCreateString return a null pointer for empty strings. But why are we trying to avoid that?

bviglietta:
Thanks, I see it now.

---
In reply to: [294206399](https://github.com/Microsoft/WinObjC/pull/2522#issuecomment-294206399) [](ancestors = 294206399,294205679,294205085)

include/Platform/Universal Windows/UWP/RTHelpers.h
```diff
@@ -893,22 +893,19 @@ static HString nsStrToHstr(NSString* str) {
         // We increment this just to make sure that we return an empty HString.
```
aballway: Could this be replaced with the NSString*->HSTRING in StringHelpers.h?

pradipd: I would need to modify StringHelpers.h to not include Starboard.h.  
I'll look into it, but, I'd prefer not to change more than I need to do for this bug fix.

aballway: It's not necessary for this PR but something to consider in the future

PR #2535, opened 2017-04-17
Fix #2533 by making cppwinrt a private dependency

### LINKED ISSUE #2533: Nuget dependency on cppwinrt should be marked as PrivateAssets All

bviglietta:
Fixes #2533
cppwinrt is a build-time dependency for WinObjC. The dependency should
not propagate to other projects that depend on WinObjC.


bviglietta:

PR #2536, opened 2017-04-17
Update README to account for bug in Nugetizer VSIX install

bbowman:
@andrejo-msft please comment back with how the install goes with this.

You can see the new look at https://github.com/bbowman/WinObjC/tree/bbowman/updateReadmeXamarin#download-and-installation


andrejo-msft:
:shipit:

PR #2539, opened 2017-04-17
Convert Security framework to C++/WinRT

bviglietta:
Convert Security framework to C++/WinRT

### DIFF
--- a/Frameworks/Security/GenericPasswordItemHandler.mm
+++ b/Frameworks/Security/GenericPasswordItemHandler.mm
@@ -25,6 +25,7 @@
 #import <CoreFoundation/CoreFoundation.h>
 #import <Security/SecItem.h>
 #import <ErrorHandling.h>
+#import "CppWinRTHelpers.h"
 
 #import <Foundation/NSData.h>
 #import <Foundation/NSString.h>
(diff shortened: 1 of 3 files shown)


Frameworks/Security/GenericPasswordItemHandler.mm
```diff
@@ -589,16 +558,14 @@ - (OSStatus)remove:(NSDictionary*)queryDictionary {
...
-    NSArray* credArray = [_vault retrieveAll];
+    WFC::IVectorView<PasswordCredential> credArray = _vault.RetrieveAll();
 
     // Documentation for remove states that it by default removes all matching rows while documentation
     // for query defaults to a single item. Go figure.
-    std::vector<StrongId<WSCPasswordCredentialWrapper>> credentialsToRemove = FindMatchingCredentials(queryDictionary, credArray, true);
+    std::vector<StrongId<PasswordCredentialWrapper>> credentialsToRemove = FindMatchingCredentials(queryDictionary, credArray, true);
 
     for (const auto& credential : credentialsToRemove) {
-        // TODO: Error handling when projections support it.
```
jaredhms: >// TODO: Error handling when projections support it. [](start = 8, length = 52)

Is crashing on failure considered 'error handling'?  do we need a separate issue tracking properly handling errors?

Frameworks/Security/GenericPasswordItemHandler.mm
```diff
@@ -518,11 +490,11 @@ - (OSStatus)add:(NSDictionary*)attributes withResult:(id*)result {
...
     try {
-        [_vault add:[credential credential]];
-    } catch (NSException* e) {
-        if ([e _hresult] == E_INVALIDARG) {
+        _vault.Add([credential credential]);
+    } catch (const winrt::hresult_error& e) {
+        // Add is guaranteed to throw if you try to store more than ten credentials per app
+        if (e.code() == E_INVALIDARG) {
             return errSecParam;
         }
```
jaredhms: should we trace here since this is a lossy transform?

jaredhms:
:shipit:

Frameworks/Security/GenericPasswordItemHandler.mm
bviglietta: The comment dated from a time when the projections swallowed errors instead of throwing them. I removed it because it was out of date.

C++/WinRT handles errors in pretty much the same way that our projections do: by throwing them. Most WinRT APIs should not throw under normal circumstances, so I wouldn't advocate putting catch blocks everywhere. Maybe we should at least translate the C++ exceptions to Objective-C exceptions?

jaredhms: Probably not a bad idea; we do have CATCH_THROW_NSEXCEPTION in result.h.

bviglietta: I put in a feature request to the C++/WinRT team to provide a facility for custom error handling.

bbowman: Yeah I think we should transform to NSException where reasonable. WinRT is getting better at only throwing where exceptional but we probably want to try to do the nice thing where possible.

DHowett-MSFT: Security is nominally below the Foundation line and shouldn't take another dependency. It can't throw exceptions. :frowning:

Doing so will complicate NSURLSession, NSURLConnection, and NSURLProtocol's ability to do authentication exchanges, as those live in Foundation.

PR #2543, opened 2017-04-18
Finish NSNull

DHowett-MSFT:
There were a few stubs left in NSNull. Now there aren't.

### DIFF
--- a/Frameworks/Foundation/NSNull.mm
+++ b/Frameworks/Foundation/NSNull.mm
@@ -14,9 +14,7 @@
 //
 //******************************************************************************
 
-#include "Starboard.h"
-#include "StubReturn.h"
-#include "Foundation/NSNull.h"
+#import <Foundation/NSNull.h>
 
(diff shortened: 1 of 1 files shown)


DHowett-MSFT:
Probably should update the copyright header.

Frameworks/Foundation/NSNull.mm
```diff
@@ -14,9 +14,7 @@
 //
 //******************************************************************************
 
-#include "Starboard.h"
-#include "StubReturn.h"
-#include "Foundation/NSNull.h"
+#import <Foundation/NSNull.h>
 
 static NSNull* nullSingleton;
```
DHowett-MSFT: This won't be switched to StrongId because `NSNull` cannot be retained or released, and making that switch would confer no benefit.

DHowett-MSFT:
Huh, actually: this is supposed to be bridged to CFNull.

PR #2545, opened 2017-04-18
Put back in dependency walking to get a minset of binaries in the fin…

bbowman:
…al package.

This works by removing extra items from ReferenceCopyLocalPaths by walking the target
binary's dependency (and any implicit dependencies from References) and keying off of
a special piece of metadata added only by WinObjC binaries. NOTE: this requires that
dependencies and references all be transitive. In other words by depending on a package,
and producing a package, a consumer of the second package also depends on the first. This
creates an end result where all binaries are trying to be copied into the final output and
the final output has extra references added to it since the package can't distinguish between
a "true consumer" and a "transitive consumer".

Additionally clean up sbclang.targets and other packaging items to do with linking.


tools/msvc/sbclang.targets
```diff
@@ -631,14 +649,14 @@
   </ItemDefinitionGroup>
 
   <PropertyGroup>
-    <ComputeLinkInputsTargets>$(ComputeLinkInputsTargets);ComputeClangLinkInputs;ComputeRuntimeInputs</ComputeLinkInputsTargets>
+    <ComputeLinkInputsTargets>$(ComputeLinkInputsTargets);ComputeClangLinkInputs;</ComputeLinkInputsTargets>
     <ComputeLibInputsTargets>$(ComputeLibInputsTargets);ComputeClangLibInputs;</ComputeLibInputsTargets>
     <ComputeImpLibInputsTargets>$(ComputeImpLibInputsTargets);ComputeClangImpLibInputs;</ComputeImpLibInputsTargets>
   </PropertyGroup>
 
   <!-- Fix up AdditionalLibraryDirectories before the CreateWinMD target runs. -->
```
bviglietta: >Fix up AdditionalLibraryDirectories [](start = 7, length = 35)

Comment is out of date.

tools/WinObjC.Language/WinObjC.Language.targets
```diff
@@ -6,4 +6,65 @@
...
+    <_ReferencesFromPackage Include="$(MSBuildThisFileDirectory)\deps\prebuilt\$(TargetOsAndVersion)\$(PlatformTarget)\*.winmd"/>
+    <_ReferencesFromPackage Include="$(MSBuildThisFileDirectory)\deps\prebuilt\$(TargetOsAndVersion)\$(PlatformTarget)\$(Configuration)\*.winmd"/>
+
+    <_LibsFromPackage Include="$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\*.dll"/>    
+    <_LibsFromPackage Include="$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\$(Configuration)\*.dll"/>
+
+    <_LibsFromPackage Include="$(MSBuildThisFileDirectory)\deps\prebuilt\$(TargetOsAndVersion)\$(PlatformTarget)\*.dll"/>
+    <_LibsFromPackage Include="$(MSBuildThisFileDirectory)\deps\prebuilt\$(TargetOsAndVersion)\$(PlatformTarget)\$(Configuration)\*.dll"/>
+
+    <_ImportLibsFromPackage Include="$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\*.lib"/>
```
bviglietta: >_ImportLibsFromPackage [](start = 5, length = 22)

Seems like a bit of a misnomer. This will pick up vcclang.lib, among other things, which isn't an import library.

tools/WinObjC.Language/WinObjC.Language.targets
```diff
@@ -6,4 +6,65 @@
   <ImportGroup>
     <Import Project="$(MsBuildThisFileDirectory)msvc\starboard-sdk.targets" Condition="Exists('$(MsBuildThisFileDirectory)msvc\starboard-sdk.targets')" />
   </ImportGroup>
+
+  <!-- Make sure that the output artifacts are included in the build. This means adding include paths and link libraries for it.
+
+     NOTE: It is important that this information is added at the end in the .targets instead of in the .props so that consumers don't
+     accidentally stomp it and get really confused. This is contrary to a configurable type property that a consumer may want / need to adjust. 
+     To depend on this package means the consumer gets these these things.  -->
```
bviglietta: >these these [](start = 55, length = 11)

these

bviglietta:
:shipit:

tools/msvc/sbclang.targets
```diff
@@ -649,27 +667,26 @@
...
       </Link>
     </ItemGroup>
 
   </Target>
 
 
   <PropertyGroup>
-    <FixAdditionalLibraryDirectoriesLinkDependsOn>FixAdditionalLibraryDirectoriesWinMD;</FixAdditionalLibraryDirectoriesLinkDependsOn>
-    <FixAdditionalLibraryDirectoriesLinkDependsOn Condition="'$(WindowsAppContainer)' == 'true'">$(FixAdditionalLibraryDirectoriesLinkDependsOn);XamlPreLink</FixAdditionalLibraryDirectoriesLinkDependsOn>
+    <FixAdditionalOptionsLinkDependsOn>FixAdditionalOptionsWinMD;</FixAdditionalOptionsLinkDependsOn>
```
memontic-ms: Add $(FixAdditionalOptionsLinkDependsOn) to safeguard against squashing this property during future changes.

DHowett-MSFT:
nit: reword <64 char when merging so it doesn't wrap and insert an ellipsis.

build/WinObjC.Frameworks/WinObjC.Frameworks.targets
```diff
@@ -61,7 +63,7 @@
   </ItemGroup>
 
   <!-- TODO: Remove when ANGLE package is correct -->
```
DHowett-MSFT: can we file an issue on this, or document why ANGLE is not "correct"?

tools/msvc/sbclang.targets
```diff
@@ -294,38 +287,78 @@
...
+        <Command>"$(MSBuildThisFileDirectory)..\bin\depends.exe" /c /f:1 /d:"$(DepWalkerSearchPathFile)" /oc:"%(OutputFile)" "%(FullPath)"</Command>
+      </DepWalkerInputs>
+    </ItemGroup>
+
+    <!-- 2. Collect dep walker search paths. This is the closure of all WinObjC ReferenceCopyLocalPaths dlls. -->
+    <ItemGroup>
+      <DepWalkerSearchPaths Include="@(ReferenceCopyLocalPaths -> 'UserDir %(RootDir)%(Directory)')" Condition="'%(ReferenceCopyLocalPaths.Extension)' == '.dll' And '%(ReferenceCopyLocalPaths.IsWinObjC)' == 'true'" />
+    </ItemGroup>
+
+        <!-- Write a Dependency Walker Path file -->
```
memontic-ms: nit: spacing
Also, should this be number 3?

tools/msvc/sbclang.targets
```diff
@@ -711,13 +727,6 @@
     </PropertyGroup>
   </Target>
 
-  <Target Name="ComputeRuntimeInputs" DependsOnTargets="ComputeClangOptions;ComputeStarboardRequiresObjectiveCRuntime">
-    <ItemGroup>
-      <Link Condition="'$(StarboardLinkObjCRuntime)' == 'true' and '$(StarboardRequiresObjectiveCRuntime)' == 'true'" Include="$(StarboardObjCRuntimeLib)"/>
```
DHowett-MSFT: what here ensures that libobjc2 is linked?

also, i'm worried about the libobjc2 build itself perhaps trying to links libobjc2 since we got rid of StarboadLinkObjcRuntime

tools/WinObjC.Language/WinObjC.Language.targets
```diff
@@ -6,4 +6,65 @@
...
+
+    <_ImportLibsFromPackage Include="$(MSBuildThisFileDirectory)\deps\prebuilt\$(TargetOsAndVersion)\$(PlatformTarget)\*.lib"/>    
+    <_ImportLibsFromPackage Include="$(MSBuildThisFileDirectory)\deps\prebuilt\$(TargetOsAndVersion)\$(PlatformTarget)\$(Configuration)\*.lib"/>
+
+    <_PriFilesFromPackage Include="$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\*.pri"/>    
+    <_PriFilesFromPackage Include="$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\$(Configuration)\*.pri"/>
+  </ItemGroup>
+
+  <ItemDefinitionGroup>
+    <ClangCompile>
```
memontic-ms: Shouldn't this InternalSystemIncludePaths be in sbclang.props with the others?

tools/WinObjC.Language/WinObjC.Language.targets
```diff
@@ -6,4 +6,65 @@
...
+      <InternalSystemIncludePaths>$(MSBuildThisFileDirectory)\include\;%(InternalSystemIncludePaths);</InternalSystemIncludePaths>
+    </ClangCompile>
+
+    <!-- Add in the link bits only if its a "final" linked product like a dll or exe -->
```
memontic-ms: This comment implies that Link should have a condition on it, which I believe it did before. Is that condition no longer needed?

tools/WinObjC.Language/WinObjC.Language.targets
```diff
@@ -6,4 +6,65 @@
...
+    <Link>
+      <AdditionalLibraryDirectories>$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\;$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\$(Configuration)\;$(MSBuildThisFileDirectory)\deps\prebuilt\$(TargetOsAndVersion)\$(PlatformTarget)\;$(MSBuildThisFileDirectory)\deps\prebuilt\$(TargetOsAndVersion)\$(PlatformTarget)\$(Configuration)\;%(AdditionalLibraryDirectories);</AdditionalLibraryDirectories>
+    </Link>
+  </ItemDefinitionGroup>
+
+  <!-- TODO: This should use XMLPoke or similar to dynamically figure out where the important bits are packaged to.-->
```
memontic-ms: Is this TODO still needed?

tools/WinObjC.Logging/WinObjC.Logging.targets
```diff
@@ -8,25 +8,57 @@
          NOTE: It is important that this information is added at the end in the .targets instead of in the .props so that consumers don't
          accidentally stomp it. This is contrary to a configurable type property that a consumer may want / need to adjust. To depend on
          this package means the consumer gets these these things.  -->
+    <ItemGroup Condition="'$(ConfigurationType)' != 'StaticLibrary'">
+      <_ReferencesFromPackage Include="$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\*.winmd"/>    
+      <_ReferencesFromPackage Include="$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\$(Configuration)\*.winmd"/>
+
+      <_ReferencesFromPackage Include="$(MSBuildThisFileDirectory)\deps\prebuilt\$(TargetOsAndVersion)\$(PlatformTarget)\*.winmd"/>
```
memontic-ms: Logging doesn't have a deps\prebuilt\ directory. If this is supposed to be a generic "include the stuff from the package" thing, can you pull this out into another targets files for Logging and Language to include? Else get rid of Logging trying to include things from a location that it doesn't have.

tools/WinObjC.Language/WinObjC.Language.targets
```diff
@@ -6,4 +6,65 @@
...
+  <ItemGroup Condition="'$(ConfigurationType)' != 'StaticLibrary'">
+    
+    <Reference Include="@(_ReferencesFromPackage)">
+      <Implementation>%(Filename).dll</Implementation>
+    </Reference>
+
+    <ReferenceCopyLocalPaths Include="@(_LibsFromPackage)">
+      <IsWinObjC>true</IsWinObjC>
```
memontic-ms: I'm a bit confused/concerned how IsWinObjC is being used. It looks like it can only ever be set under this (and Logging's) '$(ConfigurationType)' != 'StaticLibrary' conditional. A name that better identifies when or why it should be set would be helpful.

tools/WinObjC.Language/WinObjC.Language.targets
bbowman: Fair.

tools/msvc/sbclang.targets
bbowman: fixed.

tools/msvc/sbclang.targets
bbowman: not sure I want people messing with that one but sure.

build/WinObjC.Frameworks/WinObjC.Frameworks.targets
bbowman: https://github.com/Microsoft/angle/issues/107

tools/msvc/sbclang.targets
bbowman: Again we should split it up correctly and possibly have libobjc2 build as part of the normal build instead of as a checked in binary. 

Nothing in this file ensures that its linked. I am making as much of this as simple as possible / not special. It is added to the linker in WinObjC.Language.targets like in all the other packages.

tools/WinObjC.Language/WinObjC.Language.targets
bbowman: I am trying to standardize our packages as much as possible (to make it easier to use the packaging package). sbclang.props is definitely not standardized and should probably reduced down to the strict core of what is needed to call clang and compile things (imagine a WinObjC.Compiler package that libobjc2 could depend on).

Perhaps I should remove more things from the default InternalSystemIncludePaths in sbclang.props.

tools/WinObjC.Language/WinObjC.Language.targets
bbowman: Yeah I decided that we don't need that condition since all that is happening is adding additional search directories. If someone want to manually add libdispatch.lib as an AdditionalLibrary to a static lib, that should probably link successfully. We want to be more conservative with what we add to the linker ourselves but we should let devs do what they want.

tools/WinObjC.Language/WinObjC.Language.targets
bbowman: I mean it would be nicer to do it better but probably not ever going to happen realistically.

tools/WinObjC.Logging/WinObjC.Logging.targets
bbowman: Yes its basically supposed to be a generic one. This is all supposed to be handled by the packaging package. Didn't we have a task to use it in the main build?

tools/WinObjC.Language/WinObjC.Language.targets
bbowman: lets chat. I don't get your concern.

tools/WinObjC.Language/WinObjC.Language.targets
memontic-ms: I agree. I'd prefer all the InternalSystemIncludePaths standardized for this PR. And/or create another task to standardize sbclang.props/targets.

build/build.sln
```diff
@@ -2690,7 +2690,7 @@ Global
 		{B309943F-E9E9-482E-93C7-56FA7D58F898} = {1C64EC86-5799-4311-BED7-0286F61A4D50}
 		{3DC04C10-245A-4EF7-ACB6-12BE7D4ECF76} = {B309943F-E9E9-482E-93C7-56FA7D58F898}
 		{BC8C5D49-060C-4375-904C-FE472C757E58} = {3DC04C10-245A-4EF7-ACB6-12BE7D4ECF76}
-		{A44DA03C-DDA0-495A-9168-30CA6BF87CA5} = {83BDC65C-1E0A-482B-886C-52B78B8B9C97}
+		{A44DA03C-DDA0-495A-9168-30CA6BF87CA5} = {E1BA5F74-6863-437D-9AF5-72666F6D25B5}
```
pradipd: >E1BA5F74-6863-437D-9AF5-72666F6D25B5 [](start = 44, length = 36)

What is the change here?

pradipd: Just want to make sure we don't do something like [https://github.com/Microsoft/WinObjC/pull/2503](https://github.com/Microsoft/WinObjC/pull/2503)


---
In reply to: [112072314](https://github.com/Microsoft/WinObjC/pull/2545#discussion_r112072314) [](ancestors = 112072314)

tools/WinObjC.Logging/WinObjC.Logging.targets
```diff
@@ -8,25 +8,57 @@
...
+      <_ReferencesFromPackage Include="$(MSBuildThisFileDirectory)\deps\prebuilt\$(TargetOsAndVersion)\$(PlatformTarget)\$(Configuration)\*.winmd"/>
+
+      <_LibsFromPackage Include="$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\*.dll"/>    
+      <_LibsFromPackage Include="$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\$(Configuration)\*.dll"/>
+
+      <_LibsFromPackage Include="$(MSBuildThisFileDirectory)\deps\prebuilt\$(TargetOsAndVersion)\$(PlatformTarget)\*.dll"/>
```
pradipd: >_LibsFromPackage [](start = 7, length = 16)

Nit: LibsFromPackage seems misleading since these are dll's.
Not that big a deal though.

build/build.sln
bbowman: Its moving the subfolder for SafariServices. Somehow it got moved under security.

PR #2549, opened 2017-04-18
fix syntax error in versioning for master

bbowman:


DHowett-MSFT:
You could have just changed the base branch in the original PR 😜

PR #2550, opened 2017-04-18
Add explicit doesNotRecognizeSelector: to NSProxy

ms-jihua:
- Fixes a unit test failure on ARM
- Also cleaned up some header includes in NSProxy.mm

### DIFF
--- a/Frameworks/Foundation/NSObjectInternal.h
+++ b/Frameworks/Foundation/NSObjectInternal.h
@@ -14,6 +14,8 @@
 //
 //******************************************************************************
 
+#pragma once
+
 #import <Foundation/NSObject.h>
 
 long _throwUnrecognizedSelectorException(id self, Class isa, SEL sel);
--- a/Frameworks/Foundation/NSObjectInternal.h
+++ b/Frameworks/Foundation/NSObjectInternal.h
@@ -1,0 +1,19 @@
+//******************************************************************************
+//
+// Copyright (c) Microsoft. All rights reserved.
+//
+// This code is licensed under the MIT License (MIT).
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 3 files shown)


Frameworks/Foundation/NSProxy.mm
```diff
@@ -202,12 +210,40 @@ - (BOOL)conformsToProtocol:(Protocol*)aProtocol {
     return class_conformsToProtocol(forwardClass, aProtocol);
 }
 
+// NOTE: long return value to allow nonfatal continuation to get a "valid" result (for non-fpret/non-stret calls)
+static long _throwUnrecognizedSelectorException(id self, Class isa, SEL sel) {
```
aballway: why is this helper needed if it's only being used in -[doesNotRecognizeSelector] ?

ms-jihua: it's more or less just lifted from NSObject.mm - I can consolidate it if you want, but I don't think it makes much of a difference either way

DHowett-MSFT: I'd prefer we expose it in NSObject as an internal function and use it from there, but I'm won't be horribly hurt if we don't.

msft-Jeyaram: nit: __throwUn*

rajsesh-msft:
:shipit:

Frameworks/Foundation/NSObjectInternal.h
```diff
@@ -0,0 +1,19 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
```
msft-Jeyaram: #pragma once

PR #2565, opened 2017-04-19
Updated UIScrollview to allow programmatic scrolling when user scrolling is disabled

### LINKED ISSUE #2471: UIScrollView programmatic scroll methods do not work if .scrollEnabled = NO
I am unable to use -scrollRectToVisible: or change the content offset of a UIScrollView that has scrollEnabled set to NO.

This differs from how iOS works: scrollEnabled should only control whether the user can scroll with their finger.

Programmatically, these methods should still let you animate a scroll view change regardless of whether scrollEnabled is set.

tadam-msft:
Fixes #2471 UIScrollView programmatic scroll methods do not work if .scrollEnabled = NO 

Updated UIScrollView to use ScrollMode to disable scrolling as opposed to ScrollBarVisibility, which allows programmatic scrolling to still work while user scrolling is disabled.  I also added tests to validate that both contentOffset and scrollRectToVisible APIs continued to function with scrollEnabled = NO.


Frameworks/UIKit/UIScrollView.mm
```diff
@@ -65,7 +65,7 @@
 static const wchar_t* TAG = L"UIScrollView";
 
 static const bool DEBUG_ALL = false;
-static const bool DEBUG_VERBOSE = DEBUG_ALL || false;
+static const bool DEBUG_VERBOSE = true; // DEBUG_ALL || false;
```
jaredhms: >// DEBUG_ALL || false; [](start = 40, length = 22)

need to switch back

Frameworks/UIKit/UIScrollView.mm
```diff
@@ -217,7 +217,8 @@ - (BOOL)_isAnimating {
 
 - (void)_setupViewChangingHandler {
     __weak UIScrollView* weakself = self;
-    _viewChangingEventToken = _scrollViewer.ViewChanging(objcwinrt::callback([weakself] (const WF::IInspectable&, const Controls::ScrollViewerViewChangingEventArgs& arg) {
+    _viewChangingEventToken = _scrollViewer.ViewChanging(objcwinrt::callback([weakself](
+        const WF::IInspectable&, const Controls::ScrollViewerViewChangingEventArgs& arg) {
```
jaredhms: nit: did clang-format move this?

jaredhms: so many formatting changes - makes it ~tough to find the fix.

---
In reply to: [112261921](https://github.com/Microsoft/WinObjC/pull/2565#discussion_r112261921) [](ancestors = 112261921)

Frameworks/UIKit/UIScrollView.mm
```diff
@@ -602,16 +607,19 @@ - (void)setScrollEnabled:(BOOL)enable {
 }
 
 - (void)_setScrollEnabled:(BOOL)enable {
-    _scrollViewer.HorizontalScrollBarVisibility(enable ? _previousHorizontalScrollBarVisibility : Controls::ScrollBarVisibility::Disabled);
-    _scrollViewer.VerticalScrollBarVisibility(enable ? _previousVerticalScrollBarVisibility : Controls::ScrollBarVisibility::Disabled);
+    _scrollViewer.HorizontalScrollBarVisibility(enable ? _previousHorizontalScrollBarVisibility : Controls::ScrollBarVisibility::Hidden);
+    _scrollViewer.HorizontalScrollMode(enable ? Controls::ScrollMode::Enabled : Controls::ScrollMode::Disabled);
+
+    _scrollViewer.VerticalScrollBarVisibility(enable ? _previousVerticalScrollBarVisibility : Controls::ScrollBarVisibility::Hidden);
+    _scrollViewer.VerticalScrollMode(enable ? Controls::ScrollMode::Enabled : Controls::ScrollMode::Disabled);
```
jaredhms: worth adding some comments?

jaredhms:
>     }

do we need the same fix here?

---
Refers to: Frameworks/UIKit/UIScrollView.mm:971 in fb79b1a. [](commit_id = fb79b1a11afe1be10f18bd397fd0830e40911a5f, deletion_comment = False)

Frameworks/UIKit/UIScrollView.mm
```diff
@@ -602,16 +607,19 @@ - (void)setScrollEnabled:(BOOL)enable {
 }
 
 - (void)_setScrollEnabled:(BOOL)enable {
-    _scrollViewer.HorizontalScrollBarVisibility(enable ? _previousHorizontalScrollBarVisibility : Controls::ScrollBarVisibility::Disabled);
-    _scrollViewer.VerticalScrollBarVisibility(enable ? _previousVerticalScrollBarVisibility : Controls::ScrollBarVisibility::Disabled);
+    _scrollViewer.HorizontalScrollBarVisibility(enable ? _previousHorizontalScrollBarVisibility : Controls::ScrollBarVisibility::Hidden);
```
jaredhms: So if I setShowsHorizontalScrollIndicator:YES, and then setScrollEnabled:NO, the scrollbars disappear.  Is that expected/intentional?

tadam-msft:
>     }

In this case ScrollBarVisibility is already being set to hidden so it conforms to the pattern.  It would be an issue if it set it to disabled.

---
In reply to: [295364641](https://github.com/Microsoft/WinObjC/pull/2565#issuecomment-295364641) [](ancestors = 295364641)

---
Refers to: Frameworks/UIKit/UIScrollView.mm:971 in fb79b1a. [](commit_id = fb79b1a11afe1be10f18bd397fd0830e40911a5f, deletion_comment = False)

Frameworks/UIKit/UIScrollView.mm
tadam-msft: Sorry, this was done when I "git commit"ed the change.

---
In reply to: [112262947](https://github.com/Microsoft/WinObjC/pull/2565#discussion_r112262947) [](ancestors = 112262947,112261921)

Frameworks/UIKit/UIScrollView.mm
tadam-msft: Yes, iOS doesn't show the ScrollIndicators (ScrollBars) when scrollEnabled = NO.

---
In reply to: [112270292](https://github.com/Microsoft/WinObjC/pull/2565#discussion_r112270292) [](ancestors = 112270292)

Frameworks/UIKit/UIScrollView.mm~RF9233065.TMP
```diff
@@ -0,0 +1,1520 @@
+//******************************************************************************
+//
+// Copyright (c) 2015 Microsoft Corporation. All rights reserved.
+//
+// This code is licensed under the MIT License (MIT).
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
```
jaredhms: oops; definitely don't want to push this file

jaredhms:
>     }

No, I mean -- why did you have to check for 'if (self.isScrollEnabled)' in setShowsHorizontalScrollIndicator, but not here?

---
In reply to: [295403526](https://github.com/Microsoft/WinObjC/pull/2565#issuecomment-295403526) [](ancestors = 295403526,295364641)

---
Refers to: Frameworks/UIKit/UIScrollView.mm:971 in fb79b1a. [](commit_id = fb79b1a11afe1be10f18bd397fd0830e40911a5f, deletion_comment = False)

tadam-msft:
>     }

Good catch.

---
In reply to: [295431348](https://github.com/Microsoft/WinObjC/pull/2565#issuecomment-295431348) [](ancestors = 295431348,295403526,295364641)

---
Refers to: Frameworks/UIKit/UIScrollView.mm:971 in fb79b1a. [](commit_id = fb79b1a11afe1be10f18bd397fd0830e40911a5f, deletion_comment = False)

jaredhms:
:shipit:

yiyang-msft:
:shipit:

PR #2566, opened 2017-04-19
Disable drawing tests using Segoe on ARM

### LINKED ISSUE #2564: CoreGraphics Drawing Tests fail for tests rendering Segoe UI on ARM
There seems to be a slight difference between the version of Segoe UI on x86 (Version 5.55) and on ARM (Version 5.55; sf) which causes the tests to fail.  This is a regression by 22f536ab3cbe509104658037a2b9cd85953f5b69 which removed the necessary threshold from these tests.

aballway:
Fixes #2564

### DIFF
--- a/tests/unittests/CoreGraphics.drawing/CTDrawingTests.cpp
+++ b/tests/unittests/CoreGraphics.drawing/CTDrawingTests.cpp
@@ -500,18 +500,13 @@ TEXT_DRAW_TEST_P(Fonts, TestFonts) {
 }
 
 // Segoe UI is noticeably different on ARM, but we still want to test other fonts so only remove Segoe
-#ifdef WINOBJC_DISABLE_TESTS
 static CFStringRef c_fontNames[] = { CFSTR("Arial"), CFSTR("Times New Roman"), CFSTR("Wingdings") };
-#else
-static CFStringRef c_fontNames[] = { CFSTR("Arial"), CFSTR("Times New Roman"), CFSTR("Wingdings"), CFSTR("Segoe UI") };
-#endif
 INSTANTIATE_TEST_CASE_P(TestDrawingTextInFonts, Fonts, ::testing::ValuesIn(c_fontNames));
 
 #ifdef WINOBJC
 
 // The fall-through font for WinObjC is Segoe, so we can't run this on ARM
-#ifndef WINOBJC_DISABLE_TESTS
-TEXT_DRAW_TEST_F(CTFontManager, DrawWithCustomFont, WhiteBackgroundTest<>) {
+ARM_DISABLED_TEXT_DRAW_TEST_F(CTFontManager, DrawWithCustomFont, WhiteBackgroundTest<>) {
     CGContextRef context = GetDrawingContext();
     CGRect bounds = GetDrawingBounds();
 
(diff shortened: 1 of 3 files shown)


tests/unittests/CoreGraphics.drawing/CTDrawingTests.cpp
```diff
@@ -499,10 +499,18 @@ TEXT_DRAW_TEST_P(Fonts, TestFonts) {
     __DrawLoremIpsum(context, path.get(), keys, values);
 }
 
+// Segoe UI is noticeably different on ARM, but we still want to test other fonts so only remove Segoe
+#ifdef WINOBJC_DISABLE_TESTS
```
msft-Jeyaram: what is the long term plan for this?
are we going to figure out a way to still validate it on ARM (fix the issue?)?
If so could you open a task and put the number here.

tests/unittests/CoreGraphics.drawing/CTDrawingTests.cpp
```diff
@@ -499,10 +499,18 @@ TEXT_DRAW_TEST_P(Fonts, TestFonts) {
...
+static CFStringRef c_fontNames[] = { CFSTR("Arial"), CFSTR("Times New Roman"), CFSTR("Wingdings") };
+#else
 static CFStringRef c_fontNames[] = { CFSTR("Arial"), CFSTR("Times New Roman"), CFSTR("Wingdings"), CFSTR("Segoe UI") };
+#endif
 INSTANTIATE_TEST_CASE_P(TestDrawingTextInFonts, Fonts, ::testing::ValuesIn(c_fontNames));
 
 #ifdef WINOBJC
+
+// The fall-through font for WinObjC is Segoe, so we can't run this on ARM
+#ifndef WINOBJC_DISABLE_TESTS
```
msft-Jeyaram: Why can't you just disable it only on arm?

aballway: This is only disabled on ARM, this is the ARM flag for tests

msft-Jeyaram: hmm it's also in the TestFramework.h
try using ARM_DISABLED_TEST?

tests/unittests/CoreGraphics.drawing/CTDrawingTests.cpp
aballway: For this one we could simply get rid of the Segoe test completely since we already test other fonts.

This might be the only place CTFontDrawGlyphs is validated, so that one shouldn't be removed, 
maybe just make that one a different font and hope x86 and ARM don't diverge on that one.

As for drawing with WinObjC-Regular I don't believe we have implemented setting different fall-back fonts so we're at the mercy of the platform.

I'm not sure if we should expect the fonts to converge at any point or if this is just a difference we're going to have to live with.  Thoughts @DHowett-MSFT @rajsesh-msft?

rajsesh-msft: We don't need to test every font with every API.  We use DWrite under the covers, and have to trust it.  So, just remove Segoe UI from this list.  The fonts could sometimes be different on the phone either to adjust for the form factor or size or both.  

---
In reply to: [112273877](https://github.com/Microsoft/WinObjC/pull/2566#discussion_r112273877) [](ancestors = 112273877)

rajsesh-msft: missed the comment from code flow. Just remove Segoe UI tests.  We don't do anything fancy with the fonts themselves - just use the DWrite apis.

PR #2569, opened 2017-04-19
Remove projections from app activation

bviglietta:
There are a couple of places where we expose projection classes in our public interface. One of them is in the app activation path: if the app is activated via some protocol, then a projection class will be delivered to the app delegate. This change removes those uses of projection classes. This is a **breaking change** for clients.

There are several different kinds of activation that we support.

1. Toast notification: This never used projection classes, so there's no change here.

2. Protocol (i.e., URL) activation: This previously delivered a WFUri* to the app and will now deliver a NSURL*.

3. Voice command activation
4. File activation
These two delivered a WMSSpeechRecognitionResult* and a WAAFileActivatedEventArgs*, respectively. Now they will both deliver a RTObject*, the base class of all projection classes. In order to get to the derived type, client code will have to rt_dynamic_cast the object to the appropriate type.

### DIFF
--- a/Frameworks/UIKit/UIApplicationMain.mm
+++ b/Frameworks/UIKit/UIApplicationMain.mm
@@ -57,7 +57,6 @@ static const wchar_t* TAG = L"UIApplicationMain";
 
 using namespace Microsoft::WRL;
 using namespace winrt::Windows::ApplicationModel;
-namespace WF = winrt::Windows::Foundation;
 
 @interface NSAutoreleasePoolWarn : NSAutoreleasePool
 @end
@@ -356,15 +355,19 @@ extern "C" void UIApplicationMainHandleWindowVisibilityChangeEvent(bool isVisibl
 }
 
 extern "C" void UIApplicationMainHandleVoiceCommandEvent(IInspectable* voiceCommandResult) {
-    WF::IInspectable inspectable = objcwinrt::from_insp<WF::IInspectable>(voiceCommandResult);
-    RTObject* result = objcwinrt::to_rtobj(inspectable);
+    id result = _createProjectionObject("WMSSpeechRecognitionResult", voiceCommandResult, "voice commands");
 
-    [[UIApplication sharedApplication] _sendVoiceCommandReceivedEvent:result];
+    if (result) {
+        [[UIApplication sharedApplication] _sendVoiceCommandReceivedEvent:result];
+    }
 }
 
-extern "C" void UIApplicationMainHandleFileEvent(IInspectable* result) {
-    WF::IInspectable inspectable = objcwinrt::from_insp<WF::IInspectable>(result);
-    [[UIApplication sharedApplication] _sendFileReceivedEvent:objcwinrt::to_rtobj(inspectable)];
+extern "C" void UIApplicationMainHandleFileEvent(IInspectable* fileResult) {
+    id result = _createProjectionObject("WAAFileActivatedEventArgs", fileResult, "file activation");
+
+    if (result) {
+        [[UIApplication sharedApplication] _sendFileReceivedEvent:result];
+    }
 }
 
 static NSString* _bundleIdFromPackageFamilyName(const wchar_t* packageFamily) {
@@ -391,10 +394,27 @@ static NSString* _bundleIdFromPackageFamilyName(const wchar_t* packageFamily) {
 }
 
 extern "C" void UIApplicationMainHandleProtocolEvent(IInspectable* protocolUri, const wchar_t* sourceApplication) {
-    WF::Uri uri = objcwinrt::from_insp<WF::Uri>(protocolUri);
-    NSURL* url = [NSURL URLWithString:objcwinrt::string(uri.AbsoluteUri())];
+    id result = _createProjectionObject("WFUri", protocolUri, "protocol activation");
 
-    NSString* source = _bundleIdFromPackageFamilyName(sourceApplication);
+    if (result) {
+        NSString* source = _bundleIdFromPackageFamilyName(sourceApplication);
 
-    [[UIApplication sharedApplication] _sendProtocolReceivedEvent:url source:source];
+        [[UIApplication sharedApplication] _sendProtocolReceivedEvent:result source:source];
+    }
+}
+
+id _createProjectionObject(const char* className, IInspectable* source, const char* description) {
+    // We can't refer to projection classes directly, because we don't want UIKit to have any link-time
+    // dependencies on the projections. However, we can try to load the classes dynamically.
+
+    id result = nil;
+    id cls = objc_getClass(className);
+
+    if (cls) {
+        result = objc_msgSend(cls, @selector(createWith:), source);
+    } else {
+        TraceWarning(TAG, L"%hs class not registered - link with WinObjC UWP library to receive %hs", className, description);
+    }
+
+    return result;
 }
--- a/Frameworks/UIKit/UIApplicationMain.mm
+++ b/Frameworks/UIKit/UIApplicationMain.mm
@@ -241,6 +241,13 @@ void _UIApplicationMainInit(NSString* principalClassName, NSString* delegateClas
 
     // Populate launch options
     NSMutableDictionary* launchOption = [NSMutableDictionary dictionary];
+
+    if (!activationArg) {
+        // If no activation argument provided, possibly because projections aren't loaded,
+        // populate the launch options with null just to let the app know they're missing something
+        activationArg = [NSNull null];
+    }
+
     switch (activationType) {
         case ActivationTypeToast:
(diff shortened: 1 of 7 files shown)


DHowett-MSFT:
Due to the nature of Objective-C, it's very easy for us to do this. We can just pass along an `NSObject<SomeProtocolHere>*` and hide a proxy class inside UIKit.

jaredhms:
@DHowett-MSFT - this change lets them either use projections OR CppWinRT, but I agree it's a bit messy.  I don't understand the 'internal crap' comment though - these are actual WinRT types that the app needs to know about and work with - definitely not internal to us.  

My main concern is that RTObject* loses all type info - maybe we just add placeholder classes that are documented to be 'cast'-able to either their Projections OR CppWinRT counterparts.

bviglietta:
Dustin, I'm not sure I understand all of your points here. We already do implement these activations as application launch with options: they show up as entries inside the options dictionary that the app delegate receives in application:didFinishLaunchingWithOptions:.

Also, why is it burdensome to use the projection classes? I don't see how WMSSpeechRecognitionResult is more difficult to use than NSObject<SpeechRecognitionResultProtocolThatIJustInvented>.

I'll grant that the rt_dynamic_cast requirement is a new thing that this change adds, but I think it's pretty minor.

bviglietta:
If people really don't like the typeless RTObjects, I'm entirely open to introducing some new classes -- SpeechActivation and FileActivation, let's say -- that wrap the underlying WinRT objects.

DHowett-MSFT:
> this change lets them either use projections OR CppWinRT

Not exactly; we're not exposing CppWinRT across the boundary, so they'd have to unbox RTObject and that is not supported externally. We have `comObj` internally for unboxing.

> "internal crap"

Yes and no. They're projected objects, and their contents are important, but we are forcing them to take a projections dependency to use them. It leaks an implementation detail across the API boundary. If we go for a protocol (specifically: that gives us full agency as to internal implementation, where a publically exposed class name does not), we can change things as we wish without betraying internal details.

> We already do implement these activations as application launch with options

We're not, though. We augmented the UIApplicationDelegate protocol with these three callbacks.

```objc
- (void)application:(UIApplication*)application didReceiveVoiceCommand:(WMSSpeechRecognitionResult*)result;
- (void)application:(UIApplication*)application didReceiveFile:(WAAFileActivatedEventArgs*)result;
- (void)application:(UIApplication*)application didReceiveProtocol:(WFUri*)uri;
```

I don't think it's bad that we did that -- after all, we've greatly expanded its capabilities -- but it doesn't fit the application activation model from the reference platform.

As an iOS developer, I hated having to bifurcate my app initialization across the iOS 4 and iOS 5 and iOS 6 initializers. It's no different when _we_ add new ones.

> more burdensome

You're right. I think it's more mentally taxing than truly taxing, as on both sides of the implementation we/they have to care about precisely what variety of object it is.

> guaranteed to be "cast"-able

protocols, mang! it'll rock our world.

bviglietta:
> We're not, though [implementing these as application launch options].

We are, really! The methods you pointed out that we added to the application delegate are useful if an activation comes in after the app is already running.

Frameworks/UIKit/UIApplicationMain.mm
```diff
@@ -336,7 +340,7 @@ void _UIApplicationShutdown() {
     THROW_NS_IF_FAILED(comPtr.As(&map));
 
     NSMutableDictionary* userInput = nil;
-    THROW_NS_IF_FAILED(Collections::WRLToNSCollection(map, &userInput));
+    THROW_NS_IF_FAILED(::Collections::WRLToNSCollection(map, &userInput));
```
aballway: Is there a need for this to be changed?

DHowett-MSFT:
> The methods you pointed out that we added to the application delegate are useful if an activation comes in after the app is already running.

Cool! I didn't consider that use case.

I don't want to be too bullish on the compatibility front. I know that we're still a pre-release and at 0.2... but I think it's important for us to come up with a way forward that maintains our agency while exposing an Objective-C or Objective-C++ facade that doesn't require the consumer to worry too deeply about _how_ we've done something.

If we're going to make a breaking change, I guess, let's do it once and let's do it right? 😸

tests/functionaltests/Tests/CortanaTests.mm
```diff
@@ -277,6 +277,13 @@ - (void)application:(UIApplication*)application didReceiveProtocol:(WFUri*)uri {
                 return S_OK;
             });
```
aballway: nit: can ToString be safely removed?

Frameworks/UIKit/UIApplicationMain.mm
bviglietta: Yes, it's to disambiguate between global Collections and Windows::Foundation::Collections.

---
In reply to: [112285489](https://github.com/Microsoft/WinObjC/pull/2569#discussion_r112285489) [](ancestors = 112285489)

tests/functionaltests/Tests/CortanaTests.mm
bviglietta: No, ToString is also used.

---
In reply to: [112285907](https://github.com/Microsoft/WinObjC/pull/2569#discussion_r112285907) [](ancestors = 112285907)

bviglietta:
In the latest commit, I've switched back to using the projection types, so this is no longer a breaking change. The only difference between this and the original code is that we now load the projection classes dynamically, so there's no link-time dependency between UIKit and the projections.

(I even switched protocol activation back to WFUri* from NSURL*, even though NSURL was arguably a strict improvement.)

Frameworks/UIKit/UIApplicationMain.mm
```diff
@@ -352,12 +355,19 @@ void _UIApplicationShutdown() {
 }
 
 extern "C" void UIApplicationMainHandleVoiceCommandEvent(IInspectable* voiceCommandResult) {
-    WMSSpeechRecognitionResult* result = [WMSSpeechRecognitionResult createWith:voiceCommandResult];
-    [[UIApplication sharedApplication] _sendVoiceCommandReceivedEvent:result];
+    id result = _createProjectionObject("WMSSpeechRecognitionResult", voiceCommandResult, "voice commands");
+
+    if (result) {
```
jaredhms: perhaps we should fire the event even if 'result' is nil, so it's more clear to the app dev that they need to pull in projections?

Frameworks/UIKit/UIApplicationMain.mm
```diff
@@ -384,8 +394,27 @@ void _UIApplicationShutdown() {
...
+    id result = _createProjectionObject("WFUri", protocolUri, "protocol activation");
+
+    if (result) {
+        NSString* source = _bundleIdFromPackageFamilyName(sourceApplication);
+
+        [[UIApplication sharedApplication] _sendProtocolReceivedEvent:result source:source];
+    }
+}
+
+id _createProjectionObject(const char* className, IInspectable* source, const char* description) {
```
jaredhms: >_createProjectionObject [](start = 3, length = 23)

should we switch to this approach for UIView's public xamlElement property, and for initWith:xamlElement?

bviglietta: Seems like a good idea.

---
In reply to: [112307470](https://github.com/Microsoft/WinObjC/pull/2569#discussion_r112307470) [](ancestors = 112307470)

Frameworks/UIKit/UIApplicationMain.mm
bviglietta: Maybe so. If we do that, then we should also, for consistency, put an NSNull into the launchOptions dictionary.

---
In reply to: [112307083](https://github.com/Microsoft/WinObjC/pull/2569#discussion_r112307083) [](ancestors = 112307083)

Frameworks/UIKit/UIApplicationMain.mm
```diff
@@ -384,8 +394,27 @@ void _UIApplicationShutdown() {
...
+    // We can't refer to projection classes directly, because we don't want UIKit to have any link-time
+    // dependencies on the projections. However, we can try to load the classes dynamically.
+
+    id result = nil;
+    id cls = objc_getClass(className);
+
+    if (cls) {
+        result = objc_msgSend(cls, @selector(createWith:), source);
```
DHowett-MSFT: nit: `id objc_msgSend(id, SEL, ...)` does not guarantee that vararg calling convention matches that of fixed arguments. This requires a cast to be completely safe.

Alternatively,
```objc
@protocol _ProjectionClassInformal
+ (id)createWith:(IInspectable*)i;
@end
// ...
result = [static_cast<Class<_ProjectionClassInformal>>(cls) createWith:source]
```

jaredhms:
:shipit:

bbowman:
:shipit:

PR #2580, opened 2017-04-20
Enchance the WinObjCRT shim to redirect memory management calls

bbowman:
Forward all heap call to one dll so that we can ensure the correct heap is used.
Additionally clean up CRT use in tests to be more app like.


tests/unittests/CoreGraphics.drawing/TAEFEntryPoint.cpp
```diff
@@ -128,6 +130,7 @@ std::shared_ptr<DrawingTestConfigImpl> _configImpl;
 
 BEGIN_MODULE()
 MODULE_PROPERTY(L"RunAs", L"UAP")
+MODULE_PROPERTY(L"UAP:AppXManifest", L"Default.AppxManifest.xml")
```
DHowett-MSFT: If it is default, why must it be specified?

bbowman: its still not the one provided by TAEF. Its our default custom one :-p

bbowman:
Not super tested. It was validated by looking at the imports of WOCCatalog and doing some light testing with it. After the change we get:

    WinObjCRT.dll
                46C484 Import Address Table
                524CD8 Import Name Table
                     0 time date stamp
                     0 Index of first forwarder reference

                   1A free
                   22 malloc

So I assume it works.

build/Tests/Tests.Shared/Tests.Shared.vcxitems
```diff
@@ -35,14 +35,61 @@
...
+    <ReferenceCopyLocalPaths Include="$(MSBuildThisFileDirectory)\..\..\..\tests\functionaltests\Tests\$(PlatformTarget)\Default.AppxManifest.xml" />
   </ItemGroup>
   <ItemGroup>
     <ClInclude Include="$(MSBuildThisFileDirectory)..\..\..\Frameworks\include\CppUtils.h" />
     <ClInclude Include="$(MSBuildThisFileDirectory)..\..\..\Frameworks\include\MockClass.h" />
     <ClInclude Include="$(MSBuildThisFileDirectory)..\..\..\tests\unittests\Tests.Shared\ByteUtils.h" />
   </ItemGroup>
   <Target Name="_GenerateRestoreGraphProjectEntry" />
+
+  <!-- This target adds reference to the Microsoft.UniversalCRT.Debug Extension SDK. This is that same thing that the normal Appx targets do but only for
```
bviglietta: >This target adds reference to the Microsoft.UniversalCRT.Debug Extension SDK [](start = 7, length = 76)

I see that this will find ucrtbased.dll, but what about all the other CRT DLLs? Previously they were checked in to ContainerCRTDeps, and now where do they come from?

bviglietta: Make sure to try the tests on ARM. Devices typically don't have the debug CRT installed to system32.

---
In reply to: [112344528](https://github.com/Microsoft/WinObjC/pull/2580#discussion_r112344528) [](ancestors = 112344528)

DHowett-MSFT: When you do so, please endeavour to clean out the test directory. It might have left-over runtime libraries.

bbowman: @Bviglietta look at the appxmanifest. The vclibs come from an appx dependency. I do need to test on arm though because I'm not sure how that install process looks.

PR #2582, opened 2017-04-20
Complete implementation of NSScanner

### LINKED ISSUE #2083: complete NSScanner stubs
locale
copyWithZone:
localizedScannerWithString:
scanDecimal:
scanHexDouble:
scanHexFloat:
localizedScannerWithString:

ms-jihua:
- Implemented the following functions/properties:
	- locale
	- copyWithZone:
	- localizedScannerWithString:
	- scanDecimal:
	- scanHexDouble:
	- scanHexFloat:
- Added unit tests for NSScanner (there were none, including for previously implemented features)
- Fixed a number of edge cases that were incorrect

Fixes #2083

### DIFF
--- a/Frameworks/Foundation/NSScanner.mm
+++ b/Frameworks/Foundation/NSScanner.mm
@@ -1,5 +1,5 @@
 /* Copyright (c) 2006-2007 Christopher J. W. Lloyd
-   Copyright (c) 2016 Microsoft Corporation. All rights reserved.
+   Copyright (c) Microsoft. All rights reserved.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
 documentation files (the "Software"), to deal in the Software without restriction, including without limitation the
@@ -25,39 +25,45 @@ OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 // See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 
-#include "Starboard.h"
-#include "StubReturn.h"
-#include "Foundation/NSString.h"
-#include "Foundation/NSCharacterSet.h"
-#include "Foundation/NSScanner.h"
-#include <limits.h>
+#import <Foundation/NSScanner.h>
+#import <Foundation/NSCharacterSet.h>
+#import <Foundation/NSDecimalNumber.h>
+#import <Foundation/NSString.h>
 
-#include "unicode/uniset.h"
+#import <Starboard.h>
 
-typedef unsigned short unichar;
+#import <cctype>
+#import <limits>
+#import <vector>
 
 @implementation NSScanner
 
 /**
  @Status Interoperable
 */
-+ (instancetype)scannerWithString:(id)str {
-    NSScanner* ret = [self alloc];
-
-    return [[ret initWithString:str] autorelease];
++ (instancetype)scannerWithString:(NSString* _Nonnull)string {
+    return [[[self alloc] initWithString:string] autorelease];
 }
 
 /**
  @Status Interoperable
 */
-- (instancetype)initWithString:(id)string {
-    self = [self init];
-    _string = [string copy];
-    _location = 0;
-    _len = [_string length];
-    _skipSet = [[NSCharacterSet whitespaceAndNewlineCharacterSet] retain];
-    _isCaseSensitive = NO;
-    _locale = nil;
++ (instancetype)localizedScannerWithString:(NSString* _Nonnull)string {
+    NSScanner* ret = [self scannerWithString:string];
+    ret.locale = [NSLocale currentLocale];
+    return ret;
+}
+
+/**
+ @Status Interoperable
+*/
+- (instancetype)initWithString:(NSString* _Nonnull)string {
+    if (self = [super init]) {
+        _string = [string copy];
+
+        // Don't need to copy here as whitespaceAndNewlineCharacterSet is a copy property
+        _charactersToBeSkipped = [[NSCharacterSet whitespaceAndNewlineCharacterSet] retain];
+    }
 
     return self;
 }
@@ -65,83 +71,42 @@ typedef unsigned short unichar;
 /**
  @Status Interoperable
 */
-- (NSString*)string {
-    return _string;
-}
-
-/**
- @Status Interoperable
-*/
 - (void)dealloc {
     [_string release];
-    [_skipSet release];
+    [_charactersToBeSkipped release];
     [_locale release];
+
     [super dealloc];
 }
 
 /**
  @Status Interoperable
 */
-- (NSCharacterSet*)charactersToBeSkipped {
-    return _skipSet;
+- (id)copyWithZone:(NSZone*)zone {
+    NSScanner* ret = [[[self class] alloc] initWithString:_string];
+    ret->_scanLocation = _scanLocation;
+    ret->_caseSensitive = _caseSensitive;
+    ret->_charactersToBeSkipped = [_charactersToBeSkipped copy];
+    ret->_locale = [_locale copy];
+
+    return ret;
 }
 
-/**
- @Status Interoperable
-*/
-- (BOOL)caseSensitive {
-    return _isCaseSensitive;
-}
+// Helper function that returns the index of the next character not in charactersToBeSkipped
+// Returns string length if no unskippable characters are found
+- (NSUInteger)_indexOfNextUnskippedCharacter {
+    const NSUInteger length = _string.length;
+    NSUInteger ret = _scanLocation;
 
-/**
- @Status Interoperable
-*/
-- (void)setCharactersToBeSkipped:(id)set {
-    if (set != _skipSet) {
-        set = [set retain];
-        [_skipSet release];
-        _skipSet = set;
-    }
-}
-
-/**
- @Status Interoperable
-*/
-- (void)setCaseSensitive:(BOOL)flag {
-    _isCaseSensitive = flag;
-}
-
-/**
- @Status Interoperable
-*/
-- (BOOL)isAtEnd {
-    NSUInteger length = [_string length];
-    NSUInteger currentLocation = _location;
-
-    for (; currentLocation < length; currentLocation++) {
-        if ([_skipSet characterIsMember:[_string characterAtIndex:currentLocation]] == TRUE) {
-            continue;
-        } else {
-            return FALSE;
+    if (_charactersToBeSkipped) {
+        for (; ret < length; ++ret) {
+            if (![_charactersToBeSkipped characterIsMember:(unichar)[_string characterAtIndex:ret]]) {
+                return ret;
+            }
         }
     }
 
-    return TRUE;
-}
-
-/**
- @Status Interoperable
-*/
-- (NSUInteger)scanLocation {
-    // return _location < _len ? _location : _len - 1;
-    return _location;
-}
-
-/**
- @Status Interoperable
-*/
-- (void)setScanLocation:(NSUInteger)pos {
-    _location = pos;
+    return length;
 }
 
 /**
@@ -152,7 +117,7 @@ typedef unsigned short unichar;
 
     // This assumes sizeof(long long) >= sizeof(int).
     if (![self scanLongLong:&scanValue]) {
-        return FALSE;
+        return NO;
     } else if (valuep) {
         if (scanValue > INT_MAX) {
             *valuep = INT_MAX;
@@ -163,7 +128,7 @@ typedef unsigned short unichar;
         }
     }
 
-    return TRUE;
+    return YES;
 }
 
 /**
@@ -174,7 +139,7 @@ typedef unsigned short unichar;
 
     // This assumes sizeof(long long) >= sizeof(int).
     if (![self scanLongLong:&scanValue]) {
-        return FALSE;
+        return NO;
     } else if (valuep) {
         if (scanValue > LONG_MAX) {
             *valuep = LONG_MAX;
@@ -185,45 +150,43 @@ typedef unsigned short unichar;
         }
     }
 
-    return TRUE;
+    return YES;
 }
 
 /**
  @Status Interoperable
 */
 - (BOOL)scanLongLong:(long long*)valuep {
-#define long_long_MAX 0x7fffffffffffffffLL
-#define long_long_MIN (-0x7fffffffffffffffLL - 1)
-
-    NSUInteger length;
     int sign = 1;
-    BOOL hasSign = FALSE;
+    BOOL hasSign = NO;
     long long value = 0;
-    BOOL hasValue = FALSE;
-    BOOL hasOverflow = FALSE;
+    BOOL hasValue = NO;
+    BOOL hasOverflow = NO;
 
-    for (length = (NSUInteger)[_string length]; _location < length; _location++) {
-        unichar unicode = (unichar)[_string characterAtIndex:_location];
+    const long long longLongMax = std::numeric_limits<long long>::max();
 
-        if (!hasValue && [_skipSet characterIsMember:unicode])
-            continue;
-        else if (!hasSign && unicode == '-') {
+    const NSUInteger length = _string.length;
+    NSUInteger i = [self _indexOfNextUnskippedCharacter];
+    for (; i < length; ++i) {
+        unichar unicode = (unichar)[_string characterAtIndex:i];
+
+        if (!hasSign && unicode == '-') {
             sign = -1;
-            hasSign = TRUE;
+            hasSign = YES;
         } else if (!hasSign && unicode == '+') {
             sign = 1;
-            hasSign = TRUE;
+            hasSign = YES;
         } else if (unicode >= '0' && unicode <= '9') {
             if (!hasOverflow) {
                 int c = unicode - '0';
 
                 // Inspired by http://www.math.utoledo.edu/~dbastos/overflow.html
-                if ((long_long_MAX - c) / 10 < value)
-                    hasOverflow = TRUE;
+                if ((longLongMax - c) / 10 < value)
+                    hasOverflow = YES;
                 else
                     value = (value * 10) + c;
-                hasSign = TRUE;
-                hasValue = TRUE;
+                hasSign = YES;
+                hasValue = YES;
             }
         } else
             break;
@@ -232,21 +195,24 @@ typedef unsigned short unichar;
     if (hasOverflow) {
         if (valuep) {
             if (sign > 0) {
-                *valuep = long_long_MAX;
+                *valuep = longLongMax;
             } else {
-                *valuep = long_long_MIN;
+                *valuep = std::numeric_limits<long long>::min();
             }
         }
 
-        return TRUE;
+        _scanLocation = i;
+        return YES;
     } else if (hasValue) {
         if (valuep) {
             *valuep = sign * value;
         }
-        return TRUE;
+
+        _scanLocation = i;
+        return YES;
     }
 
-    return FALSE;
+    return NO;
 }
 
 /**
@@ -268,25 +234,56 @@ typedef unsigned short unichar;
  @Status Interoperable
 */
 - (BOOL)scanDouble:(double*)valuep {
-    char* pScanStart = (char*)[_string UTF8String];
-    char* pScanEnd = NULL;
+    NSUInteger startLocation = [self _indexOfNextUnskippedCharacter];
+    const NSUInteger length = _string.length;
+    if (startLocation >= length) {
+        return NO;
+    }
 
-    pScanStart += _location;
+    char* pScan = (char*)[_string UTF8String] + startLocation;
 
-    double val = strtod(pScanStart, &pScanEnd);
-    assert(pScanEnd != NULL);
+    // strtod automatically takes care of spaces (needs to be explicitly specified in charactersToBeSkipped),
+    // and only expects '.' as a decimal separator
+    // copy a scannable range of the string to a new buffer, adjusting for these constraints
+    char decimalSeparator;
+    if ((!_locale) ||
+        (![[_locale objectForKey:NSLocaleDecimalSeparator] getCString:&decimalSeparator maxLength:1 encoding:NSASCIIStringEncoding])) {
+        decimalSeparator = '.';
+    }
 
-    _location += pScanEnd - pScanStart;
+    std::vector<char> charsToScan;
+    charsToScan.reserve(_string.length - startLocation);
+    for (NSUInteger i = startLocation; i < length; ++i, ++pScan) {
+        if ((*pScan >= '0') && (*pScan <= '9')) {
+            charsToScan.push_back(*pScan);
+
+        } else if (*pScan == decimalSeparator) {
+            // Add a decimal separator that strtod will accept
+            charsToScan.push_back('.');
+
+        } else {
+            break;
+        }
+    }
+
+    if (charsToScan.size() == 0) {
+        return NO;
+    }
+
+    char* pScanEnd = nullptr;
+    double val = strtod(charsToScan.data(), &pScanEnd);
+
+    if ((!pScanEnd) || (pScanEnd == charsToScan.data())) {
+        return NO;
+    }
+
+    _scanLocation = startLocation + (pScanEnd - charsToScan.data());
 
     if (valuep) {
         *valuep = val;
     }
 
-    if ((pScanEnd - pScanStart) == 0) {
-        return FALSE;
-    }
-
-    return TRUE;
+    return YES;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


Frameworks/Foundation/NSScanner.mm
```diff
@@ -575,118 +557,283 @@ - (BOOL)scanUpToString:(NSString*)string intoString:(NSString**)intoString {
...
+                    return NO;
+                }
+                break;
+
+            case STATE_MANTISSA:
+                if (unicode == '.') {
+                    if (!scannedPeriod) {
+                        scannedPeriod = true;
+                    } else {
+                        goto done_scanning_hex_double; // Stop scanning once a second period has been scanned
```
ms-jihua: Since break inside a switch in a for loop breaks the switch and not the outer loop, this unfortunately seemed to be the best out of a group of bad options:
- Add a done bool that is checked on every iteration: this causes ++i to occur once more than necessary unless it's the end of the string, making the set of _scanLocation = 1 at the end more complicated than necessary
- Move the switch to a helper function: would need to manipulate both mantissa and exponent through out-pointers, seems unnecessarily messy
- Add a continue to every 'good' case, then add a break outside the switch that the 'bad' cases can fall too: wwwwaayyy too messy

Frameworks/Foundation/NSScanner.mm
```diff
@@ -185,45 +150,43 @@ - (BOOL)scanInteger:(int*)valuep {
...
+
+        if (!hasSign && unicode == '-') {
             sign = -1;
-            hasSign = TRUE;
+            hasSign = YES;
         } else if (!hasSign && unicode == '+') {
             sign = 1;
-            hasSign = TRUE;
+            hasSign = YES;
         } else if (unicode >= '0' && unicode <= '9') {
```
DHowett-MSFT: `iswdigit`?

Frameworks/Foundation/NSScanner.mm
```diff
@@ -185,45 +150,43 @@ - (BOOL)scanInteger:(int*)valuep {
...
             if (!hasOverflow) {
                 int c = unicode - '0';
 
                 // Inspired by http://www.math.utoledo.edu/~dbastos/overflow.html
-                if ((long_long_MAX - c) / 10 < value)
-                    hasOverflow = TRUE;
+                if ((longLongMax - c) / 10 < value)
```
DHowett-MSFT: aaaa my eyes they are burning

nit: please fix these weirdos :smile:

Frameworks/Foundation/NSScanner.mm
```diff
@@ -268,62 +234,93 @@ - (BOOL)scanFloat:(float*)valuep {
...
+    NSUInteger i = [self _indexOfNextUnskippedCharacter];
+    for (; i < length; ++i) {
+        unichar unicode = (unichar)[_string characterAtIndex:i];
 
         switch (state) {
-            case STATE_SPACE:
-                if ([_skipSet characterIsMember:unicode])
-                    state = STATE_SPACE;
-                else if (unicode == '0') {
+            case STATE_START:
```
DHowett-MSFT: can these states be unified?

DHowett-MSFT: (the first two)

Frameworks/Foundation/NSScanner.mm
```diff
@@ -268,62 +234,93 @@ - (BOOL)scanFloat:(float*)valuep {
...
     }
 
-    return TRUE;
+    return YES;
 }
 
 /**
  @Status Interoperable
 */
 - (BOOL)scanHexInt:(unsigned*)valuep {
```
DHowett-MSFT: can we implement scanHexInt in terms of scanHexLongLong? we'd have to accoutn for overflow behaviour, but it would kill half our hex reading code

Frameworks/Foundation/NSScanner.mm
```diff
@@ -436,40 +431,37 @@ - (BOOL)scanHexLongLong:(unsigned long long*)valuep {
...
                     }
                 } else if (unicode >= 'A' && unicode <= 'F') {
                     if (!overflow) {
-                        unsigned check = (unsigned)(value * 16 + (unicode - 'A') + 10);
-
-                        if (check >= value)
-                            value = check;
-                        else {
-                            value = -1;
+                        if (0xF000000000000000 & value) {
```
DHowett-MSFT: doesn't this need ULL at the end? hmm

Frameworks/Foundation/NSScanner.mm
```diff
@@ -575,118 +557,283 @@ - (BOOL)scanUpToString:(NSString*)string intoString:(NSString**)intoString {
...
-    for (; _location < length; _location++) {
-        unichar unicode = (unichar)[_string characterAtIndex:_location];
+    std::vector<unichar> result;
+    result.reserve(length - i);
+    for (; i < length; ++i) {
+        unichar unicode = (unichar)[_string characterAtIndex:i];
 
-        if ([_skipSet characterIsMember:unicode] && (scanStarted == NO)) {
-            // do nothing
+        if ([charset characterIsMember:unicode]) {
```
DHowett-MSFT: these two seem close enough to be mergeable, if you abstract out the terminal condition

DHowett-MSFT: (scanFrom, scanUpTo)

Frameworks/Foundation/NSScanner.mm
```diff
@@ -575,118 +557,283 @@ - (BOOL)scanUpToString:(NSString*)string intoString:(NSString**)intoString {
...
+        STATE_ZERO,
+        STATE_MANTISSA,
+        STATE_P,
+        STATE_EXPONENT,
+    } state = STATE_START;
+
+    bool mantissaIsNegative = false;
+    bool exponentIsNegative = false;
+    bool scannedPeriod = false;
+    bool hasValue = false;
```
DHowett-MSFT: nit: your new code has `hasValue` as a `bool` but the old code has it as `BOOL`; standardize?

Frameworks/Foundation/NSScanner.mm
```diff
@@ -575,118 +557,283 @@ - (BOOL)scanUpToString:(NSString*)string intoString:(NSString**)intoString {
...
+                } else {
+                    goto done_scanning_hex_double;
+                }
+                break;
+        }
+    }
+
+done_scanning_hex_double:; // Allows breaking out of the loop from inside the switch
+    if (hasValue) {
+        if (result) {
```
DHowett-MSFT: is there an overflow case to be had for decimals?

Frameworks/Foundation/NSScanner.mm
```diff
@@ -701,21 +848,23 @@ - (BOOL)scanUnsignedLongLong:(unsigned long long*)pValue {
...
- @Status Stub
- @Notes
+ @Status Interoperable
 */
-+ (id)localizedScannerWithString:(NSString*)aString {
-    UNIMPLEMENTED();
-    return StubReturn();
-}
+- (BOOL)isAtEnd {
+    @synchronized(self) {
```
DHowett-MSFT: none of the other APIs are `@synchronized`... should they be?

tests/unittests/Foundation/NSScannerTests.mm
```diff
@@ -0,0 +1,602 @@
...
+#import <Foundation/Foundation.h>
+#import <TestFramework.h>
+
+#import <objc/message.h>
+#import <cmath>
+#import <limits>
+
+static const double c_floatingPtTolerance = 0.00001;
+
+// Helper template function that calls a scan_______ function specified by the selector and does the nececssary casts
```
DHowett-MSFT: nit: nececssary

tests/unittests/Foundation/NSScannerTests.mm
```diff
@@ -0,0 +1,602 @@
...
+    EXPECT_EQ(expectedScanLocation += 40, scanner.scanLocation);
+    EXPECT_TRUE(scanner.atEnd);
+}
+
+TEST(NSScanner, ScanHexInt) {
+    testScanHexIntegral<unsigned int>(@selector(scanHexInt:));
+}
+
+TEST(NSScanner, ScanHexLongLong) {
+    testScanHexIntegral<unsigned long long>(@selector(scanHexLongLong:));
```
DHowett-MSFT: I'd like to see an additional test for something that fits in 64 but not 32; this doesn't have that

Frameworks/Foundation/NSScanner.mm
```diff
@@ -575,118 +557,283 @@ - (BOOL)scanUpToString:(NSString*)string intoString:(NSString**)intoString {
...
+                    if (scannedPeriod) {
+                        mantissa += (mantissaDigitPlace /= 16) * (unicode - '0');
+                    } else {
+                        mantissa = (mantissa * 16) + (unicode - '0');
+                    }
+                    hasValue = true;
+
+                } else if (unicode >= 'a' && unicode <= 'f') {
+                    if (scannedPeriod) {
+                        mantissa += (mantissaDigitPlace /= 16) * (unicode - 'a' + 10);
```
msft-Jeyaram: you can use the public struct NSDecimal to keep track of this.

Frameworks/Foundation/NSScanner.mm
aballway: nit: I'm not sure if our coding standards _handle_ gotos, but I think if this is the best option then we should make it follow the macro-style naming convention of ALL_CAPS so it really stands out.

Frameworks/Foundation/NSScanner.mm
```diff
@@ -575,118 +557,283 @@ - (BOOL)scanUpToString:(NSString*)string intoString:(NSString**)intoString {
...
+        *result = static_cast<float>(doubleValue);
+    }
+    return YES;
 }
 
 /**
  @Status Interoperable
  @Notes
 */
 - (BOOL)scanUnsignedLongLong:(unsigned long long*)pValue {
```
aballway: super nit: no implicit int

Frameworks/Foundation/NSScanner.mm
```diff
@@ -232,21 +195,24 @@ - (BOOL)scanLongLong:(long long*)valuep {
     if (hasOverflow) {
         if (valuep) {
             if (sign > 0) {
-                *valuep = long_long_MAX;
+                *valuep = longLongMax;
             } else {
-                *valuep = long_long_MIN;
+                *valuep = std::numeric_limits<long long>::min();
```
aballway: should this be `lowest()`?

tests/unittests/Foundation/NSScannerTests.mm
```diff
@@ -0,0 +1,602 @@
...
+TEST(NSScanner, ScanInt) {
+    testScanSignedIntegral<int>(@selector(scanInt:));
+}
+
+TEST(NSScanner, ScanInteger) {
+    testScanSignedIntegral<NSInteger>(@selector(scanInteger:));
+}
+
+TEST(NSScanner, ScanLongLong) {
+    testScanSignedIntegral<long long>(@selector(scanLongLong:));
```
aballway: nit: they're everywhere 😿

include/Foundation/NSScanner.h
```diff
@@ -25,37 +25,29 @@
...
+- (BOOL)scanHexDouble:(double*)result;
+- (BOOL)scanHexFloat:(float*)result;
 - (BOOL)scanHexInt:(unsigned int*)intValue;
 - (BOOL)scanHexLongLong:(unsigned long long*)result;
 - (BOOL)scanInteger:(NSInteger*)value;
 - (BOOL)scanInt:(int*)intValue;
 - (BOOL)scanLongLong:(long long*)longLongValue;
-- (BOOL)scanString:(NSString*)string intoString:(NSString* _Nullable*)stringValue;
+- (BOOL)scanString:(NSString* _Nonnull)string intoString:(NSString* _Nullable*)stringValue;
 - (BOOL)scanUnsignedLongLong:(unsigned long long*)unsignedLongLongValue;
```
aballway: nit: they're everywhere 😿

Frameworks/Foundation/NSScanner.mm
```diff
@@ -185,45 +150,43 @@ - (BOOL)scanInteger:(int*)valuep {
...
     }
 
-    return TRUE;
+    return YES;
 }
 
 /**
  @Status Interoperable
 */
 - (BOOL)scanLongLong:(long long*)valuep {
```
rajsesh-msft: ugh.  I am pretty sure this works, but why not use strtoll here?

Frameworks/Foundation/NSScanner.mm
```diff
@@ -268,62 +234,93 @@ - (BOOL)scanFloat:(float*)valuep {
...
+    if ((!_locale) ||
+        (![[_locale objectForKey:NSLocaleDecimalSeparator] getCString:&decimalSeparator maxLength:1 encoding:NSASCIIStringEncoding])) {
+        decimalSeparator = '.';
+    }
 
-    _location += pScanEnd - pScanStart;
+    std::vector<char> charsToScan;
+    charsToScan.reserve(_string.length - startLocation);
+    for (NSUInteger i = startLocation; i < length; ++i, ++pScan) {
+        if ((*pScan >= '0') && (*pScan <= '9')) {
```
rajsesh-msft: >(*pScan >= '0') && (*pScan <= '9' [](start = 12, length = 33)

isdigit?

Frameworks/Foundation/NSScanner.mm
```diff
@@ -25,123 +25,88 @@
...
+- (NSUInteger)_indexOfNextUnskippedCharacter {
+    const NSUInteger length = _string.length;
+    NSUInteger ret = _scanLocation;
 
-    for (; currentLocation < length; currentLocation++) {
-        if ([_skipSet characterIsMember:[_string characterAtIndex:currentLocation]] == TRUE) {
-            continue;
-        } else {
-            return FALSE;
+    if (_charactersToBeSkipped) {
```
rajsesh-msft: >_charactersToBeSkipped [](start = 8, length = 22)

This method is going to be particularly slow.

Frameworks/Foundation/NSScanner.mm
```diff
@@ -575,118 +557,283 @@ - (BOOL)scanUpToString:(NSString*)string intoString:(NSString**)intoString {
...
+    }
 
     // Scan the string for a base ten positive integer starting at the internally stored position
-    pScanStart += _location;
-    unsigned long long val = strtoull(pScanStart, &pScanEnd, 10);
+    const char* pScanStart = (char*)[_string UTF8String] + startLocation;
+    char* pScanEnd = nullptr;
+
+    // Reject any negative numbers
+    long long signedVal = strtoll(pScanStart, &pScanEnd, 10);
```
rajsesh-msft: >strtoll [](start = 26, length = 7)

could sscanf_s be used to do the heavy lifting here?

Frameworks/Foundation/NSScanner.mm
```diff
@@ -575,118 +557,283 @@ - (BOOL)scanUpToString:(NSString*)string intoString:(NSString**)intoString {
...
+
+    return YES;
 }
 
 /**
- @Status Stub
+ @Status Interoperable
  @Notes
 */
 - (BOOL)scanHexDouble:(double*)result {
```
rajsesh-msft: >scanHexDouble [](start = 8, length = 13)

could sscanf_s be used for the heavylifting here?

Frameworks/Foundation/NSScanner.mm
ms-jihua: was not sure how we felt about swapping out large chunks of the previous code. there are also unichar -> char conversion implications I haven't thought about yet, but it should be fine I think...

Frameworks/Foundation/NSScanner.mm
ms-jihua: sscanf doesn't return where it stopped scanning. you can sort of work around that by specifying "%a%s" as the format, but that scans the whole rest of the string unnecessarily... you could also do a bunch of preprocessing to give it a prepped state, but then it just ends up being the same, but with sscanf instead of ldexp

bviglietta:
> - (BOOL)scanInteger:(int*)valuep {

NSInteger

---
Refers to: Frameworks/Foundation/NSScanner.mm:137 in 658f544. [](commit_id = 658f544ba5a4cf29c9ac0b41baca20a0ad7ea12b, deletion_comment = False)

Frameworks/Foundation/NSScanner.mm
```diff
@@ -174,7 +139,7 @@ - (BOOL)scanInteger:(int*)valuep {
 
     // This assumes sizeof(long long) >= sizeof(int).
     if (![self scanLongLong:&scanValue]) {
-        return FALSE;
+        return NO;
     } else if (valuep) {
         if (scanValue > LONG_MAX) {
```
bviglietta: >LONG_MAX [](start = 24, length = 8)

NSIntegerMax, etc.

Frameworks/Foundation/NSScanner.mm
```diff
@@ -268,62 +234,93 @@ - (BOOL)scanFloat:(float*)valuep {
...
-    char* pScanStart = (char*)[_string UTF8String];
-    char* pScanEnd = NULL;
+    NSUInteger startLocation = [self _indexOfNextUnskippedCharacter];
+    const NSUInteger length = _string.length;
+    if (startLocation >= length) {
+        return NO;
+    }
 
-    pScanStart += _location;
+    char* pScan = (char*)[_string UTF8String] + startLocation;
```
bviglietta: >(char*)[_string UTF8String] + startLocation [](start = 18, length = 43)

This math won't work out well if there are non-ASCII characters in the string.

Frameworks/Foundation/NSScanner.mm
```diff
@@ -268,62 +234,93 @@ - (BOOL)scanFloat:(float*)valuep {
...
 
-    double val = strtod(pScanStart, &pScanEnd);
-    assert(pScanEnd != NULL);
+    // strtod automatically takes care of spaces (needs to be explicitly specified in charactersToBeSkipped),
+    // and only expects '.' as a decimal separator
+    // copy a scannable range of the string to a new buffer, adjusting for these constraints
+    char decimalSeparator;
+    if ((!_locale) ||
+        (![[_locale objectForKey:NSLocaleDecimalSeparator] getCString:&decimalSeparator maxLength:1 encoding:NSASCIIStringEncoding])) {
```
bviglietta: >NSASCIIStringEncoding [](start = 109, length = 21)

In some locales the decimal separator won't be representable as an ASCII character.

Frameworks/Foundation/NSScanner.mm
ms-jihua: Would be longer code, and we don't need NSDecimal's greater precision, since we only need exactly as much precision as would fit in a double's mantissa.

Frameworks/Foundation/NSScanner.mm
ms-jihua: ?

Frameworks/Foundation/NSScanner.mm
ms-jihua: you mean besides the exponent getting too large and causing ldexp to return HUGE_VAL or 0?

if the mantissa becomes too precise to represent correctly, the latter (least significant) bits get chopped off during the double operations in STATE_MANTISSA

include/Foundation/NSScanner.h
ms-jihua: ?

tests/unittests/Foundation/NSScannerTests.mm
ms-jihua: ???

Frameworks/Foundation/NSScanner.mm
aballway: `unsigned` and `long` are just modifiers, not actual types, so saying `unsigned long long` makes the compiler infer the type which by default is `int`.  It's common to elide the int, but I'm not a fan 🐋

ms-jihua: ah, that. we as well as the docs forgo the int, so I'm leaving as is.

Frameworks/Foundation/NSScanner.mm
ms-jihua: NSScanner is in the list of thread-unsafe Foundation classes.

Frameworks/Foundation/NSScanner.mm
```diff
@@ -575,147 +339,249 @@ - (BOOL)scanUpToString:(NSString*)string intoString:(NSString**)intoString {
...
+    int exponent = 0;
+
+    // Once a '.' has been scanned, this /= 16 every character, so that each base-16 decimal digit is appropriately placed
+    // ie: 1.1ae = 1 + (1 * (mantissaDigitPlace /= 16)) + (a * (mantissaDigitPlace /= 16)) + (e * (mantissaDigitPlace /= 16))
+    double mantissaDigitPlace = 1;
+
+    const NSUInteger length = _string.length;
+    NSUInteger i = [self _indexOfNextUnskippedCharacter];
+    for (; i < length; ++i) {
+        unichar unicode = _stringChars[i];
```
rajsesh-msft: nit: ch or even c is more appropriate than unicode.

Frameworks/Foundation/NSScanner.mm
```diff
@@ -575,147 +339,249 @@ - (BOOL)scanUpToString:(NSString*)string intoString:(NSString**)intoString {
...
+
+            case STATE_MANTISSA:
+                if (unicode == '.') {
+                    if (!scannedPeriod) {
+                        scannedPeriod = true;
+                    } else {
+                        goto DONE_SCANNING_HEXDOUBLE; // Stop scanning once a second period has been scanned
+                    }
+
+                } else if (unicode >= '0' && unicode <= '9') {
```
rajsesh-msft: you could extract the value of the current character first for digits and hex first and then do the scannedPeriod check.  Better yet, you could have a multiplier value which is 16 before you scan period and 1/16 after you scannedPeriod, then you don't need the if checks at all.  also, isdigit instead of actual comparison here?

Frameworks/Foundation/NSScanner.mm
```diff
@@ -575,147 +339,249 @@ - (BOOL)scanUpToString:(NSString*)string intoString:(NSString**)intoString {
...
+    double mantissaDigitPlace = 1;
+
+    const NSUInteger length = _string.length;
+    NSUInteger i = [self _indexOfNextUnskippedCharacter];
+    for (; i < length; ++i) {
+        unichar unicode = _stringChars[i];
+
+        switch (state) {
+            case STATE_START:
+                if (unicode == '-') {
```
rajsesh-msft: If these are meant to be UTF16 unichars, then need L'-' etc in front of the chars.  Not sure if that is automatic.

Frameworks/Foundation/NSScanner.mm
```diff
@@ -25,547 +25,311 @@
...
     }
 
-    for (; _location < length; _location++) {
-        unichar unicode = (unichar)[_string characterAtIndex:_location];
-        range.location = _location;
-        if ((_len - _location) < [string length]) {
-            result = NO;
-            break;
+    const NSUInteger length = _string.length;
+    NSUInteger i = [self _indexOfNextUnskippedCharacter];
```
ms-jihua: it could in this instance, but there are other instances in the file where i needs to be available outside the for loop and I'd rather keep them in line

rajsesh-msft: Understood.  You can ignore any and all the nit comments.

Frameworks/Foundation/NSScanner.mm
```diff
@@ -25,547 +25,311 @@
...
 }
 
 /**
  @Status Interoperable
 */
-- (BOOL)caseSensitive {
-    return _isCaseSensitive;
++ (instancetype)localizedScannerWithString:(NSString* _Nonnull)string {
+    NSScanner* ret = [self scannerWithString:string];
+    ret.locale = [NSLocale currentLocale];
```
msft-Jeyaram: nit: do it via an init,

Frameworks/Foundation/NSScanner.mm
```diff
@@ -575,147 +339,249 @@ - (BOOL)scanUpToString:(NSString*)string intoString:(NSString**)intoString {
...
+    for (; i < length; ++i) {
+        unichar unicode = _stringChars[i];
+
+        switch (state) {
+            case STATE_START:
+                if (unicode == '-') {
+                    state = STATE_SIGN;
+                    mantissaIsNegative = true;
+                } else if (unicode == '+') {
+                    state = STATE_SIGN;
```
msft-Jeyaram: >        state = STATE_SIGN; [](start = 12, length = 27)

you could just use the NSDecimal struct and if the precision is not too high, it would only take up the first slot of the mantissa array.
Also it would make the code much simpler, with tracking all the states.

msft-Jeyaram: hmm maybe not, given this is being looped over.

---
In reply to: [113068748](https://github.com/Microsoft/WinObjC/pull/2582#discussion_r113068748) [](ancestors = 113068748)

Frameworks/Foundation/NSScanner.mm
```diff
@@ -575,147 +339,249 @@ - (BOOL)scanUpToString:(NSString*)string intoString:(NSString**)intoString {
...
+                    state = STATE_ZERO;
+                } else {
+                    return NO;
+                }
+                break;
+
+            case STATE_ZERO:
+                if (unicode == 'x' || unicode == 'X') {
+                    state = STATE_MANTISSA;
+                } else {
```
msft-Jeyaram: >else [](start = 18, length = 4)

nit: can just return NO

Frameworks/Foundation/NSScanner.mm
```diff
@@ -575,147 +339,249 @@ - (BOOL)scanUpToString:(NSString*)string intoString:(NSString**)intoString {
...
+                    state = STATE_P;
+
+                } else {
+                    goto DONE_SCANNING_HEXDOUBLE;
+                }
+                break;
+
+            case STATE_P:
+                if (!hasValue) {
+                    goto DONE_SCANNING_HEXDOUBLE;
```
msft-Jeyaram: >goto [](start = 20, length = 4)

goto :(

Frameworks/Foundation/NSScanner.mm
ms-jihua: ? the top if case would fall through...

Frameworks/Foundation/NSScanner.mm
ms-jihua: leaving the comparison since it better contextualizes the arithmetic and is more in line with the other checks

ms-jihua: if checks would still be necessary even using that multiplier value, since the mantissa advances in different directions before/after the period ie:

0x12, scan 3 = 0x12 * 16 + 0x3 = 0x123 (multiplier applies to previous number, is constant)
0x1.2, scan 3 = 0x1.2 + 0x3 / 16^2 = 0x1.23 (multiplier applies to new digit, changes)

I could combine scanned period and the multiplier into one var, but I think that would hurt readability.

Frameworks/Foundation/NSScanner.mm
ms-jihua: there is no equivalent public init, seems like unnecessary frills to make one just for this

ms-jihua:
Ping

Frameworks/Foundation/NSScanner.mm
```diff
@@ -25,547 +25,295 @@
...
+#import <Starboard.h>
+
+#import <cctype>
+#import <functional>
+#import <limits>
+#import <vector>
+
+// Helper function for implementing a scan function for a numeric type using a CRT function such as wcstoll
+template <typename TNumeric>
+static BOOL __scanNumeric(NSScanner* scanner,
```
aballway: nit: make templated on lambda so we don't have to use std::function, and then just inline the lambdas into the __scanNumeric calls

ms-jihua: would rather have it be strongly-typed - if it turns out we are bottlenecked by std::function overhead, we can make this change then

PR #2585, opened 2017-04-20
Push UIActivityIndicatorView stopAnimating/startAnimating calls to the UIThread

### LINKED ISSUE #2577: UIActivityIndicatorView stopAnimating needs to push its Xaml interaction to the UIThread
IslandwoodNews is crashing on recent builds, _likely_ surfaced by 71042590d.  The fix is easy; we need to push the contents of UIActivityIndicatorView stopAnimating to the UIThread (just as we do for UIView initWithFrame and UIView delloc.

jaredhms:
This regression was likely exposed by the recent NSOperationQueue changes (71042590); we were previously getting called back on the UIThread accidentally.

Whether or not this is a proper long term fix will be re-evaluated with the broader backlog task around UIView threading concerns.

Fixes #2577.

### DIFF
--- a/Frameworks/UIKit/UIActivityIndicatorView.mm
+++ b/Frameworks/UIKit/UIActivityIndicatorView.mm
@@ -22,8 +22,10 @@
 #import <UIKit/UIActivityIndicatorView.h>
 #import <UIKit/UIColor.h>
 
-#import "XamlUtilities.h"
+
 #import "CppWinRTHelpers.h"
+#import "UIViewInternal.h"
+#import "XamlUtilities.h"
 
 #include "COMIncludes.h"
 #import <winrt/Windows.UI.Xaml.Controls.h>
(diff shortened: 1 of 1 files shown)


Frameworks/UIKit/UIActivityIndicatorView.mm
```diff
@@ -226,28 +228,32 @@ - (UIActivityIndicatorViewStyle)activityIndicatorViewStyle {
  @Status Interoperable
 */
 - (void)startAnimating {
-    if (_isAnimating) {
-        return;
-    }
+    RunSynchronouslyOnMainThread(^{
```
yiyang-msft: >RunSynchronouslyOnMainThread [](start = 4, length = 28)

should not the caller make sure invoking this from UIThread?

jaredhms: Many apps don't bother to check, and appear to work fine on the reference platform.  We already do this for UIView init/dealloc and in a few other spots (for sone middleware scenarios), but as called out in the commit message, we may revisit this in a future task that addresses the issue as a whole.

yiyang-msft: ok. 

---
In reply to: [112565027](https://github.com/Microsoft/WinObjC/pull/2585#discussion_r112565027) [](ancestors = 112565027)

yiyang-msft:
:shipit:

PR #2587, opened 2017-04-21
Implementation of NSMutableOrderedSet:

RG-J:
- unionOrderedSet
- removeObject
- minusOrderedSet
- tests for the above functions


RG-J:
@rajesh-msft yeah I can do them

PR #2588, opened 2017-04-21
Include versioning information on the sample apps.

### LINKED ISSUE #2513: Islandwood apps should have the build version in the title
When trying to track down a regression, I need to install several different builds to find out where the regression is appearing.  However, there is no easy way to determine what version of the application is installed.  Each application built internally should have the build version displayed windows chrome for the app.

bbowman:
Fixes #2513

### DIFF
--- a/common/winobjc.versioning.common.targets
+++ b/common/winobjc.versioning.common.targets
@@ -159,7 +159,7 @@
 
   <Target Name="FixUpAppxManifest"
           BeforeTargets="_ValidateAppxManifest"
-          DependsOnTargets="_GenerateCurrentprojectAppxManifest"
+          DependsOnTargets="_GenerateCurrentProjectAppxManifest"
           Condition="'$(AppxPackage)' == 'true' And Exists('$(FinalAppxManifestName)')">
 
     <!-- Take care of package version. Use Trimmed Version since appxmanifests hilariously don't like leading 0s  -->
--- a/common/winobjc.versioning.common.targets
+++ b/common/winobjc.versioning.common.targets
@@ -5,6 +5,7 @@
     <CalculatePackageVersionDependsOn>GetVersion;$(CalculatePackageVersionDependsOn)</CalculatePackageVersionDependsOn>
     <UpdateAssemblyInfo>false</UpdateAssemblyInfo>
     <WriteVersionInfoToBuildLog>false</WriteVersionInfoToBuildLog>
+    <_WinObjCVersioningTargetsImported>true</_WinObjCVersioningTargetsImported>
   </PropertyGroup>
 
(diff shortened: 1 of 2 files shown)


common/winobjc.versioning.common.targets
```diff
@@ -147,6 +150,40 @@
...
+    <!--
+    APPX Versioning target
+
+    This target will modify the Appx manifest that is generated during the compilation of an app.
+    The version of the app is modified as well as the display name to make it easy to identify sample app versions.
+  -->
+
+  <Target Name="FixUpAppxManifest"
+          BeforeTargets="_ValidateAppxManifest"
+          DependsOnTargets="_GenerateCurrentprojectAppxManifest"
```
memontic-ms: Where is _GenerateCurrentprojectAppxManifest defined? Also it's weird that project is lowercase.

common/winobjc.versioning.common.targets
```diff
@@ -147,6 +150,40 @@
...
+          Condition="'$(AppxPackage)' == 'true' And Exists('$(FinalAppxManifestName)')">
+
+    <!-- Take care of package version. Use Trimmed Version since appxmanifests hilariously don't like leading 0s  -->
```
memontic-ms: It's unclear exactly which xml element is being modified here.

memontic-ms:
Does this change the behavior of how the app loads/installs? Does it overwrite a previous version of the app or add another version without deleting the old one? I currently have 4 WOCCatalogs sitting around that seem to differ by the PublisherDisplayName. I'm expecting that I'll now have 1 sitting around for each version I use.

pradipd:
Change itself is fine.  
Does this cover all our apps or just the samples apps in winobjc?  If not all, probably want to confirm with @tadam-msft  that it covers the majority of the apps.

common/winobjc.versioning.common.targets
bbowman: from the appx targets and I messed up the caps but that doesn't matter apparently. Fixed.

common/winobjc.versioning.common.targets
bbowman: The Version and uap:VisualElements/DisplayName elements in the appxManifest

bbowman:
@memontic-ms no. It only changes the version and what is shown on the window chrome as the app display name. It seems that the appx install is a little wonky and overwrites the appx if its the same publisher but a different version (I guess its sort of like an update but it doesn't prompt you) and allows multiple installs of the same app with different publishers.

bbowman:
@pradipd it affects all apps that use the override targets which will be all our apps except for the samples repo. Seems good enough to me.

PR #2600, opened 2017-04-22
Fix CRT Shim bugs

bbowman:


bbowman:
looks like one test app has an AV in libGLESv2. Its possible this is due to me and possible its something real that got uncovered by re-jiggering this. Need to investigate more to be sure.

tools/msvc/sbclang.targets
```diff
@@ -828,6 +814,7 @@
     <Link Condition="'$(UseWinObjCCRT)' == 'true'">
       <!-- Ensure we pick up operator new from our project rather than the C runtime -->
```
DHowett-MSFT: since we got rid of our operator new and friends, does this warrant removal?

rajsesh-msft:
Would be good to understand this crash before merging.

---
In reply to: [296339161](https://github.com/Microsoft/WinObjC/pull/2600#issuecomment-296339161) [](ancestors = 296339161)

tools/IwNew/lib/IwNew.cpp
```diff
@@ -1,58 +0,0 @@
...
-// THE SOFTWARE.
-//
-//******************************************************************************
-
-#include <new>
-#include "IwMalloc.h"
-
-// Override new and delete to go through IwMalloc and IwFree
-
-void* operator new(std::size_t size, const std::nothrow_t&) noexcept {
```
pradipd: Does this mean new will no longer call IwMalloc()?

tools/msvc/sbclang.targets
```diff
@@ -828,6 +814,7 @@
...
       <ForceSymbolReferences>??2@YAPAXI@Z</ForceSymbolReferences>
+      <AdditionalOptions>/DISALLOWLIB:libucrt.lib /DISALLOWLIB:libucrtd.lib /nodefaultlib:ucrt /nodefaultlib:ucrtd /nodefaultlib:vccorlib /nodefaultlib:msvcrt /nodefaultlib:vccorlibd /nodefaultlib:msvcrtd /defaultlib:winobjcrt winobjcrt.lib %(AdditionalOptions)</AdditionalOptions>
```
DHowett-MSFT: putting a .lib in the additionaloptions is scary; it can't be an Input? I worry that a change in link option order would break this.

bbowman: no it can be I just didn't feel like doing it as input and wanted to make sure it was there. the important part is making sure no one else is a default lib. I'm going to leave it as is for now.

tools/IwNew/lib/IwNew.cpp
bbowman: correct. I will also be removing Iw* soon-ish

tools/msvc/sbclang.targets
bbowman: probably a good idea. Don't know about need.

bbowman:
@rajsesh-msft look like this crash repros in the debug-debug case before (tested with 0.2.170407 release bits) so my CRT changes aren't the issue here. I think I've narrowed it down to an app problem so I'm going to go ahead with this change.

PR #2606, opened 2017-04-24
Adjust versioning for sample apps not in the same repo

bbowman:


common/winobjc.versioning.common.targets
```diff
@@ -92,12 +94,15 @@
...
       <PackageAssemblyVersionTrimmed>$([System.UInt64]::Parse($(PackageVersion_Major)).ToString()).$([System.UInt64]::Parse($(PackageVersion_Minor)).ToString()).$([System.UInt64]::Parse($(PackageAssemblyVersion_Build)).ToString()).$([System.UInt64]::Parse($(PackageAssemblyVersion_Revision)).ToString())</PackageAssemblyVersionTrimmed>
-      <PackageAssemblyInformationalVersion>$(PackageVersion)+$(GitVersion_FullBuildMetaData)</PackageAssemblyInformationalVersion>
       <PackageAssemblyDescription>https://github.com/Microsoft/WinObjC</PackageAssemblyDescription>
       <PackageAssemblyCompany>Microsoft</PackageAssemblyCompany>
       <PackageAssemblyCopyright>Copyright (C) Microsoft. All rights reserved</PackageAssemblyCopyright>
       <PackageAssemblyProduct>Objective-C for Windows v$(PackageVersion)</PackageAssemblyProduct>
     </PropertyGroup>
+
+    <Message Text="##vso[build.updatebuildnumber]$(PackageVersion)"/>
+    <Message Text="##vso[build.addbuildtag]$(PackageVersion)" Condition="'$(GitVersion_BranchName)' == ''"/>
```
DHowett-MSFT: do you intend to condition this on the presence of BranchName?

bbowman: oops just the branch one and with the opposite condition. I'm smart.

common/winobjc.versioning.common.targets
```diff
@@ -30,8 +30,8 @@
       <PackageVersion_PreReleaseFormat Condition="'$(PackageVersion_PreReleaseFormat)' == ''">{PackageVersion_PreReleaseLabel}-{PackageVersion_Timestamp}</PackageVersion_PreReleaseFormat>
     </PropertyGroup>
 
-    <Error Condition="'$(GitVersion_BranchName)' == ''"
-           Text="Cannot determine git branch. Please make sure GitVersionTask is listed as a nuget dependency for this project and that git.exe is part of the PATH environment variable." />
+    <Error Condition="'$(GitVersion_BranchName)' == '' And '$(PackageVersionOverride)' == ''"
+           Text="Cannot calculate version. Please make sure either PackageVersionOverride is det or GitVersionTask is listed as a nuget dependency for this project and that git.exe is part of the PATH environment variable." />
```
pradipd: set (not det).

bbowman:
@pradipd  yes partially.

common/winobjc.versioning.common.targets
```diff
@@ -78,9 +66,47 @@
...
+      <PackageVersion_Patch Condition="'$(PackageVersion_Patch)' == '' And '$(GitVersion_PreReleaseLabel)' != ''">$(GitVersion_Patch)</PackageVersion_Patch>
+
+      <PackageVersion_PreReleaseLabel>$(GitVersion_PreReleaseLabel)</PackageVersion_PreReleaseLabel>
+    </PropertyGroup>
+
+    <!-- Last up, use calculated values. -->
+    <PropertyGroup Condition="'$(PackageVersionOverride)' == ''">
+      <!-- 
+         Three places to get a build timestamp: 
+         1. Off of the build number for a build server build (Before its been modified to be the package version)
```
pradipd: Doesn't the build produce different version/timestamps for different packages.  I've always noticed the tools packages (language, logging, tools, packaging) have a different timestamp than the packages build in build.sln.

Will we get one timestamp for all packages and have be consistent with the build number?

common/winobjc.versioning.common.targets
```diff
@@ -78,9 +66,47 @@
...
+         2. From a timestamp file created at the beginning of a local build (see NuprojInitializer.msbuildproj)
```
pradipd: Why can't the build server publish the  timestamp file and subsequent release defs use that timestamp file?
I'd would be nice if the build were consistent for local builds as well as builds on the build server.

common/winobjc.versioning.common.targets
bbowman: Yes it will rename the build. It is currently being called multiple times. 

---
In reply to: [113063928](https://github.com/Microsoft/WinObjC/pull/2606#discussion_r113063928) [](ancestors = 113063928)

common/winobjc.versioning.common.targets
bbowman: That wasn't a specific goal of this change but yes. After this it will choose to prefer the build number over the timestamp file so it will use a uniform version across solutions

---
In reply to: [113305792](https://github.com/Microsoft/WinObjC/pull/2606#discussion_r113305792) [](ancestors = 113305792)

common/winobjc.versioning.common.targets
bbowman: the timestamp file is per solution not per ? whatever it would be for cross-repo, cross-solution builds. If you want it to be smart about how to regenerate the timestamp file on local rebuilds, you can't really do a truly global file.

---
In reply to: [113306502](https://github.com/Microsoft/WinObjC/pull/2606#discussion_r113306502) [](ancestors = 113306502)

PR #2608, opened 2017-04-24
Update UIActivityIndicatorView design pattern and add tests

### LINKED ISSUE #2385: Update UIActivityIndicatorView to follow control design pattern

### LINKED ISSUE #2386: Add unit tests for UIActivityIndicatorView

nathpete-msft:
This updates UIActivityIndicatorView to follow our current control design patterns, adds tests, and resolves a few existing bugs I came across:
1. The size of the activity indicator did not shrink when applying a
normal style after the large style.
2. The color for the gray indicator style was black, not gray.
3. It was possible to resize the activity indicator frame to be smaller
than the internal ProgressRing, obscuring it.

Fixes #2385
Fixes #2386


nathpete-msft:

tadam-msft:
> <Grid

We should update the names of these files to be ActivityIndicatorView.xxx.

---
Refers to: Frameworks/UIKit.Xaml/ProgressRing.xaml:1 in ea5027a. [](commit_id = ea5027ad0f02ac09e0944b25c38c08b8f6c78d7d, deletion_comment = False)

Frameworks/UIKit/UIActivityIndicatorView.mm
```diff
@@ -22,17 +22,24 @@
...
 #import "XamlUtilities.h"
+#import "CppWinRTHelpers.h"
 
 #include "COMIncludes.h"
 #import <winrt/Windows.UI.Xaml.Controls.h>
 #include "COMIncludes_End.h"
 
+// This color matches what iOS uses internally for the gray style.
+// NOTE: UIColor.grayColor differs slightly from this. Its components are W:0.5, A:1.0
+#define activityIndicatorStyleGrayColor = [UIColor colorWithWhite:0.0 alpha:0.45]
```
jaredhms: >define [](start = 1, length = 6)

#defines are usually frowned upon; static variables are better.  However, in this case since this is used in a single spot, you can just move it (and the comment) down to where it's used.

Frameworks/UIKit/UIActivityIndicatorView.mm
```diff
@@ -129,57 +133,49 @@ - (instancetype)initWithFrame:(CGRect)frame {
...
+    if (self = [super initWithFrame:frame xamlElement:xamlElement]) {
+        [self _initUIActivityIndicatorView];
     }
 
     return self;
 }
 
 /**
  @Status Interoperable
+ @Note Frame will not be resized smaller than the contained ProgressRing, as this would obscure it.
```
jaredhms: Hmmm...do we really need to cap this? What happens on iOS?

tests/functionaltests/Tests/UIKitTests/UIActivityIndicatorTests.mm
```diff
@@ -63,4 +72,479 @@
...
+            ASSERT_FALSE(progressRing.IsActive());
+            // Element should be still "Visible" after stopping animation, to ensure space is preserved in the layout.
+            EXPECT_EQ(winrt::Windows::UI::Xaml::Visibility::Visible, xamlElement.Visibility());
+            ASSERT_FALSE(indicator.hidden);
+            EXPECT_EQ(0.0, xamlElement.Opacity());
+            EXPECT_EQ(0.0, indicator.alpha);
+        });
+    }
+
+    TEST_METHOD(UIActivityIndicatorView_HidesWhenStopped_BeforeStartAnimation) {
```
tadam-msft: >UIActivityIndicatorView_HidesWhenStopped_BeforeStartAnimation [](start = 16, length = 61)

Some comments about expected behavior on this test and others (where not obvious) about expected behavior would be good.

Frameworks/UIKit/UIActivityIndicatorView.mm
```diff
@@ -129,57 +133,49 @@ - (instancetype)initWithFrame:(CGRect)frame {
...
-    //       and we would no longer have a _subview.
-    _subView = [[UIView alloc] initWithFrame:CGRectZero
-                                 xamlElement:(xamlElement ? xamlElement : objcwinrt::to_rtobj(_progressRing))];
+    [super setFrame:frame];
+}
 
-    [self addSubview:_subView];
+- (void)_initUIActivityIndicatorView {
     _isAnimating = NO;
+    [self setHidden:NO]; // Ensure space is reserved in the layout. See notes for setHidesWhenStopped.
```
jaredhms: isn't NO the default?  if we're never setting to YES, then we don't ever need to set it to NO, right?

tests/functionaltests/Tests/UIKitTests/UIActivityIndicatorTests.mm
```diff
@@ -63,4 +72,479 @@
...
+
+            auto solidForegroundBrush = progressRing.Foreground().as<Media::SolidColorBrush>();
+            UIColor* expectedForegroundColor = [UIColor colorWithWhite:0.0 alpha:0.45];
+            EXPECT_TRUE(UXTestAPI::IsRGBAEqual(solidForegroundBrush, expectedForegroundColor));
+
+            EXPECT_EQ(expectedNormalSize, progressRing.Height());
+            EXPECT_EQ(expectedNormalSize, progressRing.Width());
+
+            // Frame size should not change
+            EXPECT_EQ(50, xamlElement.Height());
```
tadam-msft: >50 [](start = 22, length = 2)

Might be good to have a constant for this value as well--consistent with expectedNormalSize

Frameworks/UIKit/UIActivityIndicatorView.mm
nathpete-msft: iOS allows it to be resized smaller than the indicator, but still displays the indicator and the frame. For example, if you were to set the background color to black, and set the indicator frame to be 5x5. You would see the spinning indicator (normal size is 20x20) with a small 5x5 black square inside of it. In XAML, if the dimensions of the containing frame are smaller than the ProgressRing inside, it will be partially obscured.

Frameworks/UIKit/UIActivityIndicatorView.mm
nathpete-msft: Okay, so kill the define and just inline that where it's used? Yi had commented on the original review I submitted suggesting I add a constant for the color components to make it more clear, so I was trying to think of a way to make the entire color definition more clear since this is a special case.

Frameworks/UIKit/UIActivityIndicatorView.mm
```diff
@@ -229,13 +242,12 @@ - (UIActivityIndicatorViewStyle)activityIndicatorViewStyle {
 */
 - (void)startAnimating {
     RunSynchronouslyOnMainThread(^{
-        if (_isAnimating) {
-            return;
+        if (!_isAnimating) {
+            _isAnimating = YES;
+            [self setHidden:NO];
```
jaredhms: >   [self setHidden:NO]; [](start = 9, length = 23)

Does startAnimating set the control back to visible if the app explicitly set isHidden:YES?

nathpete-msft: Yes, because by default iOS hides the indicator when it is not animating.

Frameworks/UIKit/UIActivityIndicatorView.mm
jaredhms: Interesting; that sounds like the spinner is in a separate, non-clipped layer on iOS.  We could do something similar, but it seems like a corner case that we probably don't need to worry about.  However, I'm worried that silently capping the min frame size will break layout - cause layout cycles, etc.  The correct way to do this would be to implement intrinsicContentSize, but perhaps we just remove the capping and see how that works out for most apps (unless you already know of an app that requires this min frame size?)

---
In reply to: [113078524](https://github.com/Microsoft/WinObjC/pull/2608#discussion_r113078524) [](ancestors = 113078524)

Frameworks/UIKit/UIActivityIndicatorView.mm
```diff
@@ -247,11 +259,13 @@ - (void)stopAnimating {
...
         if (_isAnimating) {
             _isAnimating = NO;
-        }
 
-        _progressRing.IsActive(false);
-        if (_hidesWhenStopped) {
-            [self setHidden:YES];
+            if (_hidesWhenStopped) {
+                // Make our view transparent, rather than hidden, to ensure space is reserved in the layout.
+                [self setAlpha:0.0];
```
jaredhms: >0.0 [](start = 31, length = 3)

hmm - this assumes the alpha was 0.0 to begin with

Frameworks/UIKit/UIActivityIndicatorView.mm
nathpete-msft: So, the more likely case is that someone initializes a UIActivityIndicatorView with a CGRect and either sets the size to the default indicator size, 20x20, or omits the size by just putting 0x0 (since that Just Works™ on iOS). Obviously if it's the second case, it would be immediately obvious. However, the first case would appear fine until they were to set it to the White Large style, which would then resize the indicator to 37x37, obscuring it in XAML but not in iOS.

---
In reply to: [113079902](https://github.com/Microsoft/WinObjC/pull/2608#discussion_r113079902) [](ancestors = 113079902,113078524)

Frameworks/UIKit.Xaml/ProgressRing.xaml.cpp
```diff
@@ -18,22 +18,121 @@
...
+}
+
+UIKIT_XAML_EXPORT double XamlGetProgressRingWidthValue(const Microsoft::WRL::ComPtr<IInspectable>& progressRingInspectable) {
+	auto progressRing = safe_cast<UIKit::Xaml::ProgressRing^>(reinterpret_cast<Platform::Object^>(progressRingInspectable.Get()));
+	return progressRing->ProgressRing_Width;
+}
+
+UIKIT_XAML_EXPORT void XamlSetProgressRingWidthValue(const Microsoft::WRL::ComPtr<IInspectable>& progressRingInspectable, double width) {
+	auto progressRing = safe_cast<UIKit::Xaml::ProgressRing^>(reinterpret_cast<Platform::Object^>(progressRingInspectable.Get()));
+	progressRing->ProgressRing_Width = width;
```
jaredhms: >	 [](start = 0, length = 1)

we should get rid of all the tabs (in favor of 4 spaces)

Frameworks/UIKit.Xaml/ProgressRing.xaml.h
```diff
@@ -21,17 +21,65 @@
...
+		[Windows::Foundation::Metadata::WebHostHidden]
+		public ref class ProgressRing sealed {
+		public:
+			ProgressRing();
 
-private:
-    // TODO: Add support for padding and other missing properties, and to expose them as needed to UIKit.
-};
+			// TODO: This should be renamed to eliminate the naming prefix once the DependencyProperty is fixed
+			property Platform::Boolean ProgressRing_IsActive {
```
jaredhms: we should get rid of all the tabs (in favor of 4 spaces)

Frameworks/UIKit/UIActivityIndicatorView.mm
nathpete-msft: Okay, so kill the define and just inline that where it's used? Yi had commented on the original review I submitted suggesting I add a constant for the color components to make it more clear, so I was trying to think of a way to make the entire color definition more clear since this is a special case.

---
In reply to: [113078096](https://github.com/Microsoft/WinObjC/pull/2608#discussion_r113078096) [](ancestors = 113078096)

Frameworks/UIKit/UIActivityIndicatorView.mm
nathpete-msft: Yes, because by default iOS hides the indicator when it is not animating.

---
In reply to: [113078951](https://github.com/Microsoft/WinObjC/pull/2608#discussion_r113078951) [](ancestors = 113078951)

Frameworks/UIKit/UIActivityIndicatorView.mm
nathpete-msft: Ah, yeah. We could handle for that, but since this is also another edge case (user explicitly setting alpha on the entire view, not just the indicator itself) it may not be worth it. Since the ProgressRing already does not display anything when it is not active, this is technically handling for another edge case: the user explicitly setting a background on the indicator view. If a background is set, it would not be hidden when the animation is stopped since we don't hide the view in order to preserve space in the layout.
Now the question is, do we try to handle for either case, or simply leave it as a caveat that if the user wants to use a background in the indicator then they are responsible for hiding it when animation stops as well?

---
In reply to: [113080688](https://github.com/Microsoft/WinObjC/pull/2608#discussion_r113080688) [](ancestors = 113080688)

jaredhms: How about leaving the Grid as-is, and simply toggling the visibility of the internal ProgressRing to Visible/Collapsed?  That would let us occupy space and show the background even when the ProgressRing is hidden.  Isn't that closer to iOS?

---
In reply to: [113288024](https://github.com/Microsoft/WinObjC/pull/2608#discussion_r113288024) [](ancestors = 113288024,113080688)

jaredhms: ...you'd just need to plumb _hidesWhenStopped through so the backing Xaml control could do the right thing - it would simplify this code, too.

---
In reply to: [113292895](https://github.com/Microsoft/WinObjC/pull/2608#discussion_r113292895) [](ancestors = 113292895,113288024,113080688)

nathpete-msft: In iOS the background is set on the entire UIActivityIndicatorView, and the entire view hides when stopped (not just the indicator itself). So we actually don't want to show the background when the ProgressRing is hidden. Currently the background and frame size are both modifying the Grid that contains the ProgressRing, which is correct from an iOS standpoint. 

---
In reply to: [113292895](https://github.com/Microsoft/WinObjC/pull/2608#discussion_r113292895) [](ancestors = 113292895,113288024,113080688)

yiyang-msft:
> <Grid

yeah, so the naming conventions that we established here for UIkit.xaml is that "UIContorl" in UIkit is mapped to "Control" in UIkit.xaml

---
In reply to: [296847032](https://github.com/Microsoft/WinObjC/pull/2608#issuecomment-296847032) [](ancestors = 296847032)

---
Refers to: Frameworks/UIKit.Xaml/ProgressRing.xaml:1 in ea5027a. [](commit_id = ea5027ad0f02ac09e0944b25c38c08b8f6c78d7d, deletion_comment = False)

Frameworks/UIKit.Xaml/ProgressRing.xaml
```diff
@@ -1,11 +1,18 @@
...
-    <ProgressRing x:Name="progressRing" Background="Green" IsActive="True" ></ProgressRing>
-</UserControl>
+    mc:Ignorable="d"
+    x:Name="Root">
+    <!-- TODO: Update property names in path for binding once they have been updated in the header -->
+    <ProgressRing x:Name="progressRing"
+                  IsActive="{Binding ElementName=Root, Path=ProgressRing_IsActive, Mode=OneWay, UpdateSourceTrigger=PropertyChanged}"
+                  Foreground="{Binding ElementName=Root, Path=ProgresRing_Foreground, Mode=OneWay, UpdateSourceTrigger=PropertyChanged}"
+                  Height="{Binding ElementName=Root, Path=ProgressRing_Height, Mode=OneWay, UpdateSourceTrigger=PropertyChanged}"
+                  Width="{Binding ElementName=Root, Path=ProgressRing_Width, Mode=OneWay, UpdateSourceTrigger=PropertyChanged}"
```
yiyang-msft: >UpdateSourceTrigger=PropertyChanged [](start = 90, length = 35)

given all these are one way binding, you don't need to update the source then.

Frameworks/UIKit.Xaml/ProgressRing.xaml.cpp
```diff
@@ -18,22 +18,121 @@
...
+						because type information for custom controls is not currently being exposed to
+						the testing framework.
+						See Issue #2607 for current status.
+				*/
+				// TODO: This DependencyProperty should be attached to UIKit::Xaml::ProgressRing
+				s_isActiveProperty = DependencyProperty::RegisterAttached("ProgressRing_IsActive",
+					Platform::Boolean::typeid,
+					FrameworkElement::typeid,
+					nullptr);
+				// TODO: This DependencyProperty should be attached to UIKit::Xaml::ProgressRing
```
yiyang-msft: >// TODO: This DependencyProperty should be attached to UIKit::Xaml::ProgressRing [](start = 4, length = 80)

nit: keep one comment line with slightly rewording for clarity?  also, can use one newline as separator before the comment for clarity.

[https://github.com/Microsoft/WinObjC/wiki/Coding-Standards](https://github.com/Microsoft/WinObjC/wiki/Coding-Standards) is the general code conventions for islandwood.

Frameworks/UIKit/UIActivityIndicatorView.mm
```diff
@@ -129,57 +133,49 @@ - (instancetype)initWithFrame:(CGRect)frame {
...
 - (void)setFrame:(CGRect)frame {
-    [super setFrame:frame];
-
-    _subView.center = { self.center.x - self.frame.origin.x, self.center.y - self.frame.origin.y };
-}
-
-- (void)_initUIActivityIndicatorView:(RTObject*)xamlElement {
-    if (xamlElement != nil) {
-        _progressRing = objcwinrt::from_rtobj<Controls::ProgressRing>(xamlElement);
+    if (frame.size.width < [self _width] || frame.size.height < [self _height]) {
```
tadam-msft: >(frame.size.width < [self _width] || frame.size.height < [self _height]) [](start = 7, length = 72)

Parenthesis to group the expression on each side of the || would be good.

Frameworks/UIKit.Xaml/ActivityIndicatorView.xaml
```diff
@@ -0,0 +1,18 @@
+﻿<Grid
+    x:Class="UIKit.Xaml.ActivityIndicatorView"
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:UIKit.Xaml"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    mc:Ignorable="d"
+    x:Name="Root">
+    <!-- TODO: Update property names in path for binding once they have been updated in the header -->
```
jaredhms: >TODO: Update property names in path for binding once they have been updated in the header [](start = 9, length = 89)

github issue tracking this?

Frameworks/UIKit.Xaml/ActivityIndicatorView.xaml
```diff
@@ -0,0 +1,18 @@
...
+    <ProgressRing x:Name="progressRing"
+                  IsActive="{Binding ElementName=Root, Path=ActivityIndicatorView_IsActive, Mode=OneWay}"
+                  Foreground="{Binding ElementName=Root, Path=ActivityIndicatorView_Foreground, Mode=OneWay}"
+                  Height="{Binding ElementName=Root, Path=ActivityIndicatorView_Height, Mode=OneWay}"
+                  Width="{Binding ElementName=Root, Path=ActivityIndicatorView_Width, Mode=OneWay}"
```
jaredhms: I'm confused by these - why not just bind to Width/Height of the containing Grid?

Frameworks/UIKit/UIActivityIndicatorView.mm
jaredhms: If we switch to "White Large Style", shouldn't we adjust our frame accordingly?

---
In reply to: [113080757](https://github.com/Microsoft/WinObjC/pull/2608#discussion_r113080757) [](ancestors = 113080757,113079902,113078524)

Frameworks/UIKit.Xaml/ActivityIndicatorView.xaml
nathpete-msft: The frame and indicator have independent sizes. The user is able to resize the frame, but the indicator inside can only be set via the provided styles.

---
In reply to: [113538678](https://github.com/Microsoft/WinObjC/pull/2608#discussion_r113538678) [](ancestors = 113538678)

Frameworks/UIKit.Xaml/ActivityIndicatorView.xaml
```diff
@@ -0,0 +1,23 @@
...
+    <Canvas x:Name="indicatorCanvas"
+            Height="{Binding ElementName=Root, Path=ActivityIndicatorView_Height, Mode=OneWay}"
+            Width="{Binding ElementName=Root, Path=ActivityIndicatorView_Width, Mode=OneWay}"
+            HorizontalAlignment="Center"
+            VerticalAlignment="Center">
+        <ProgressRing x:Name="progressRing"
+                      IsActive="{Binding ElementName=Root, Path=ActivityIndicatorView_IsActive, Mode=OneWay}"
+                      Foreground="{Binding ElementName=Root, Path=ActivityIndicatorView_Foreground, Mode=OneWay}"
+                      Height="{Binding ElementName=indicatorCanvas, Path=Height, Mode=TwoWay}"
+                      Width="{Binding ElementName=indicatorCanvas, Path=Width, Mode=TwoWay}" />
```
jaredhms: OneWay?

Frameworks/UIKit.Xaml/ActivityIndicatorView.xaml
```diff
@@ -0,0 +1,23 @@
...
+    mc:Ignorable="d"
+    x:Name="Root">
+
+    <!-- TODO: Update property names in paths for binding once they have been updated in the header. See #2607 -->
+    <Canvas x:Name="indicatorCanvas"
+            Height="{Binding ElementName=Root, Path=ActivityIndicatorView_Height, Mode=OneWay}"
+            Width="{Binding ElementName=Root, Path=ActivityIndicatorView_Width, Mode=OneWay}"
+            HorizontalAlignment="Center"
+            VerticalAlignment="Center">
+        <ProgressRing x:Name="progressRing"
```
jaredhms: Probably worth a large comment as to why we have an extra Canvas (so we can opt out of Grid's clipping).  Too bad Grid doesn't have an option to explicitly opt out of its clipping behavior. :'(

Frameworks/UIKit.Xaml/ActivityIndicatorView.xaml
```diff
@@ -0,0 +1,23 @@
...
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:UIKit.Xaml"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    mc:Ignorable="d"
+    x:Name="Root">
+
+    <!-- TODO: Update property names in paths for binding once they have been updated in the header. See #2607 -->
+    <Canvas x:Name="indicatorCanvas"
+            Height="{Binding ElementName=Root, Path=ActivityIndicatorView_Height, Mode=OneWay}"
```
jaredhms: >Binding [](start = 21, length = 7)

did you try x:Bind for any of these?

Frameworks/UIKit/UIActivityIndicatorView.mm
```diff
@@ -271,14 +270,54 @@ - (void)setColor:(UIColor*)color {
...
++ (RTObject*)createXamlElement {
+    return objcwinrt::to_rtobj(XamlControls::CreateActivityIndicatorView());
+}
+
+/**
+ Gets the height underlying ProgressRing
+*/
+- (double)_height {
+    return XamlControls::GetActivityIndicatorViewHeightValue([self _winrtXamlElement]);
+}
```
jaredhms: no longer necessary?

Frameworks/UIKit/UIActivityIndicatorView.mm
```diff
@@ -271,14 +270,54 @@ - (void)setColor:(UIColor*)color {
...
+    XamlControls::SetActivityIndicatorViewHeightValue([self _winrtXamlElement], height);
+}
+
+/**
+ Gets the width underlying ProgressRing
+*/
+- (double)_width {
+    return XamlControls::GetActivityIndicatorViewWidthValue([self _winrtXamlElement]);
+}
+
```
jaredhms: no longer necessary? - same with the XamlControls getter?

Frameworks/UIKit/UIActivityIndicatorView.mm
```diff
@@ -271,14 +270,54 @@ - (void)setColor:(UIColor*)color {
...
+- (double)_height {
+    return XamlControls::GetActivityIndicatorViewHeightValue([self _winrtXamlElement]);
+}
+
+/**
+ Sets the height underlying ProgressRing
+*/
+- (void)_setHeight:(double)height {
+    XamlControls::SetActivityIndicatorViewHeightValue([self _winrtXamlElement], height);
+}
```
jaredhms: I'd probably just call these directly, since they're only called in one spot

Frameworks/UIKit/UIActivityIndicatorView.mm
```diff
@@ -271,14 +270,54 @@ - (void)setColor:(UIColor*)color {
...
+- (double)_width {
+    return XamlControls::GetActivityIndicatorViewWidthValue([self _winrtXamlElement]);
+}
+
+/**
+ Sets the width underlying ProgressRing
+*/
+- (void)_setWidth:(double)width {
+    XamlControls::SetActivityIndicatorViewWidthValue([self _winrtXamlElement], width);
+}
```
jaredhms: I'd probably just call these directly, since they're only called in one spot

Frameworks/UIKit.Xaml/ActivityIndicatorView.xaml
nathpete-msft: Yeah, I also tried manually clearing the Clip property on Grid, but that didn't work either.

---
In reply to: [113579888](https://github.com/Microsoft/WinObjC/pull/2608#discussion_r113579888) [](ancestors = 113579888)

Frameworks/UIKit.Xaml/ActivityIndicatorView.xaml
nathpete-msft: So I tried to x:Bind again, and still couldn't manage to get it working for our own properties. It does seem to work fine when binding to an existing property on a control though.

---
In reply to: [113579961](https://github.com/Microsoft/WinObjC/pull/2608#discussion_r113579961) [](ancestors = 113579961)

jaredhms:
:shipit:

yiyang-msft:
:shipit:

PR #2616, opened 2017-04-25
Adapt UITextField to use UIKit.Xaml.TextField as backing control

yiyang-msft:
This change 
1. mainly move the core implementation of UITextField to UIkit.TextField to follow the new UIkit paradigm. As a result. We have a lot less cached object in UITextField. 
2. Refactor code to use shared functions for AdjustFontSizeToFit which is used on UILabel as well.


Frameworks/UIKit.Xaml/TextField.xaml
```diff
@@ -0,0 +1,16 @@
...
+    x:Class="UIKit.Xaml.TextField"
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:UIKit.Xaml"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    mc:Ignorable="d"
+    x:Name="Root">
+    <!-- 
+        Current the backing control is dynamically created in the code
```
yiyang-msft: >Current [](start = 8, length = 7)

typos

Frameworks/UIKit/XamlControls.h
```diff
@@ -100,10 +139,75 @@ winrt::Windows::UI::Xaml::Controls::Grid CreateLabel();
...
+void SetTextFieldVerticalTextAlignment(const winrt::Windows::UI::Xaml::FrameworkElement& inspectableTextField, int verticalTextAlignment);
+
+int GetTextFieldVerticalTextAlignment(const winrt::Windows::UI::Xaml::FrameworkElement& inspectableTextField);
+
+void SetTextFieldBorderStyle(const winrt::Windows::UI::Xaml::FrameworkElement& inspectableTextField, UITextBorderStyle borderStyle);
+
+UITextBorderStyle GetTextFieldBorderStyle(const winrt::Windows::UI::Xaml::FrameworkElement& inspectableTextField);
+
+void TextFieldApplyFont(const winrt::Windows::UI::Xaml::FrameworkElement& inspectableTextField,
+                        NSString* fontFmailyname,
```
jaredhms: >fontFmailyname [](start = 34, length = 14)

nit: typo

Frameworks/UIKit.Xaml/TextField.xaml.h
```diff
@@ -0,0 +1,268 @@
...
+    void _SwitchToSecureMode();
+    void _SwitchToNonSecureMode();
+    void _SetTextVerticalAlignment();
+    void _SetBorderStyle();
+
+    // Layer elements; created on demand
+    Windows::UI::Xaml::Controls::Image^ _content;
+
+    // events handler for internal controls
+    void OnTextChanged(Platform::Object ^sender, Windows::UI::Xaml::Controls::TextChangedEventArgs ^e);
```
jaredhms: > ^ [](start = 39, length = 2)

nit: all of the ^ are on the wrong side

Frameworks/UIKit/UITextField.mm
```diff
@@ -640,26 +384,8 @@ - (UITextBorderStyle)borderStyle {
...
-        imageBrush.ImageSource(bitmapImageSource);
-
-        [_secureModeLock lock];
-        if (_secureTextMode) {
-            _passwordBox.Background(imageBrush);
-        } else {
-            _textBox.Background(imageBrush);
-        }
-        [_secureModeLock unlock];
+    if (self.borderStyle != UITextBorderStyleRoundedRect) {
```
jaredhms: comment as to why we need this check? or maybe we just don't do the check and let it look bad until Xaml fixes clipping.

Frameworks/UIKit/UITextField.mm
```diff
@@ -950,14 +680,7 @@ - (UIKeyboardAppearance)keyboardAppearance {
...
     _keyboardType = type;
-
-    [_secureModeLock lock];
-    if (_secureTextMode) {
-        _passwordBox.InputScope(XamlUtilities::ConvertKeyboardTypeToInputScope(_keyboardType, YES));
-    } else {
-        _textBox.InputScope(XamlUtilities::ConvertKeyboardTypeToInputScope(_keyboardType, NO));
-    }
-    [_secureModeLock unlock];
+    XamlControls::SetTextFieldInputScope(_textField, _keyboardType, self.isSecureTextEntry);
```
jaredhms: >self.isSecureTextEntry [](start = 68, length = 22)

probably don't need to pass this down, do we?

Frameworks/UIKit/UITextField.mm
```diff
@@ -1183,348 +891,198 @@ - (void)setContentVerticalAlignment:(UIControlContentVerticalAlignment)alignment
...
-            control.UpdateLayout();
-            [strongSelf _adjustFontSizeToFitWidthOrApplyCurrentFont];
-        }
-    }));
-}
-
-// Handler when KeyDown
-- (void)_setupControlKeyDownHandler:(const Controls::Control&)control {
-    __weak UITextField* weakSelf = self;
+- (void)_ProcessGotFocusEvent {
```
jaredhms: >_ProcessGotFocusEvent [](start = 8, length = 21)

super nit: should be _camelCase

Frameworks/UIKit/UITextField.mm
```diff
@@ -1183,348 +891,198 @@ - (void)setContentVerticalAlignment:(UIControlContentVerticalAlignment)alignment
...
+        [self sendActionsForControlEvents:UIControlEventEditingDidBegin];
+        [[NSNotificationCenter defaultCenter] postNotificationName:UITextFieldTextDidBeginEditingNotification object:self];
+    });
 
-            e.Handled(true);
-        } else {
-            e.Handled(false);
-        }
-    }));
+    [self _adjustFontSizeToFitWidthOrApplyCurrentFont];
```
jaredhms: >   [self _adjustFontSizeToFitWidthOrApplyCurrentFont]; [](start = 1, length = 54)

why is this needed when we get focus?

Frameworks/UIKit/UITextField.mm
```diff
@@ -1183,348 +891,198 @@ - (void)setContentVerticalAlignment:(UIControlContentVerticalAlignment)alignment
...
-                [[NSNotificationCenter defaultCenter] postNotificationName:UITextFieldTextDidChangeNotification object:strongSelf];
-            });
-        }
-    });
-
-    // set up focus and keydown handlers
-    [self _setupControlGotFocusHandler:_textBox];
-    [self _setupControlLostFocusHandler:_textBox];
-    [self _setupControlKeyDownHandler:_textBox];
+    [self _adjustFontSizeToFitWidthOrApplyCurrentFont];
```
jaredhms: same here - do we truly need to adjust font sizes based on focus events?

Frameworks/UIKit/UITextField.mm
```diff
@@ -1183,348 +891,198 @@ - (void)setContentVerticalAlignment:(UIControlContentVerticalAlignment)alignment
...
+                                                             [strongSelf sendActionsForControlEvents:UIControlEventEditingChanged];
+                                                             [[NSNotificationCenter defaultCenter]
+                                                                 postNotificationName:UITextFieldTextDidChangeNotification
+                                                                               object:strongSelf];
+                                                         });
+                                                     }
+                                                 }),
+
+                                                 // Got focus event
+                                                 objcwinrt::callback([weakSelf](const WF::IInspectable& sender, const RoutedEventArgs&) {
```
jaredhms: >objcwinrt::callback( [](start = 49, length = 20)

I don't think you need these, because the event callback already pushes/pops an autoreleasepool

Frameworks/UIKit.Xaml/TextField.xaml.h
```diff
@@ -0,0 +1,268 @@
...
+private:
+    static Windows::UI::Xaml::DependencyProperty^ s_textProperty;
+    static Windows::UI::Xaml::DependencyProperty^ s_secureEntryProperty;
+    static Windows::UI::Xaml::DependencyProperty^ s_placeholderProperty;
+    static Windows::UI::Xaml::DependencyProperty^ s_inputScopeProperty;
+    static Windows::UI::Xaml::DependencyProperty^ s_enabledProperty;
+    static Windows::UI::Xaml::DependencyProperty^ s_foregroundProperty;
+    static Windows::UI::Xaml::DependencyProperty^ s_textAlignmentProperty;
+    static Windows::UI::Xaml::DependencyProperty^ s_textVerticalAlignmentProperty;
+    static Windows::UI::Xaml::DependencyProperty^ s_borderStyleProperty;
```
jaredhms: how many of these do you actually need now that you don't have any controls in the default markup?

Frameworks/UIKit.Xaml/TextField.xaml.cpp
```diff
@@ -0,0 +1,759 @@
...
+}
+
+void TextField::KillFocus() {
+    if (SecureEntry) {
+        PasswordBox->IsEnabled = false;
+        PasswordBox->IsEnabled = true;
+    } else {
+        TextBox->IsEnabled = false;
+        TextBox->IsEnabled = true;
+    }
```
jaredhms: should add a comment about how this is a hack and we should switch to Windows APIs if/when available

Frameworks/UIKit.Xaml/TextField.xaml.cpp
```diff
@@ -0,0 +1,759 @@
...
+}
+
+void TextField::SetFocus() {
+    if (SecureEntry) {
+        PasswordBox->Focus(FocusState::Programmatic);
+    } else {
+        TextBox->Focus(FocusState::Programmatic);
+    }
```
jaredhms: many (most?) of these can call into _backingControl->Focus, etc.

Frameworks/UIKit.Xaml/TextField.xaml.cpp
```diff
@@ -0,0 +1,759 @@
...
+}
+
+void TextField::OnKeyDown(Platform::Object ^sender, Windows::UI::Xaml::Input::KeyRoutedEventArgs ^e)
+{
+    if (_enterKeyDownHandler) {
+        if (e->Key == Windows::System::VirtualKey::Enter) {
+            _enterKeyDownHandler->Invoke(
+                InspectableFromObject(this).Get(),
+                ObjectToType<ABI::Windows::UI::Xaml::IRoutedEventArgs>(e).Get());
+            e->Handled = true;
```
jaredhms: we don't want to always handle it, do we?  What if the user actually wants to type enter and the delegate doesn't kill focus when enter is clicked?

jaredhms: you can probably set to handled (if needed) in the  actual _enterKeyDownHandler

---
In reply to: [113563242](https://github.com/Microsoft/WinObjC/pull/2616#discussion_r113563242) [](ancestors = 113563242)

Frameworks/UIKit.Xaml/TextField.xaml.cpp
```diff
@@ -0,0 +1,759 @@
...
+        } else {
+            e->Handled = false;
+        }
```
jaredhms: nit: unnecessary

Frameworks/UIKit.Xaml/TextField.xaml.cpp
```diff
@@ -0,0 +1,759 @@
...
+        PasswordBox->FontSize = fontSize;
+        PasswordBox->FontWeight = fontWeight;
+    }
+    else {
+        TextBox->FontFamily = ref new FontFamily(fontFamily);
+        TextBox->FontStretch = strech;
+        TextBox->FontStyle = fontStyle;
+        TextBox->FontSize = fontSize;
+        TextBox->FontWeight = fontWeight;
+    }
```
jaredhms: same here - could just use _backingControl

Frameworks/UIKit.Xaml/TextField.xaml.cpp
```diff
@@ -0,0 +1,759 @@
...
+}
+
+void TextField::_SetTextVerticalAlignment() {
+    // Set vertical aligment of both content and placeholder to the same value
+    FrameworkElement^ target = FindTemplateChild<FrameworkElement^>(_backingControl, "ContentElement");
+    if (target) {
+        target->VerticalAlignment = VerticalTextAlignment;
+    }
+
+    target = FindTemplateChild<FrameworkElement^>(_backingControl, "PlaceholderTextContentPresenter");;
```
jaredhms: >; [](start = 102, length = 1)

nit: extra ';"

Frameworks/UIKit.Xaml/TextField.xaml.cpp
```diff
@@ -0,0 +1,759 @@
...
+    if (target) {
+        target->VerticalAlignment = VerticalTextAlignment;
+    }
```
jaredhms: does this work for both TextBox and PasswordBox?  Might want to add a few more comments about how you know where to find these, since the Xaml markup is practically empty.

Frameworks/UIKit.Xaml/TextField.xaml.cpp
```diff
@@ -0,0 +1,759 @@
...
+    return textField->SecureEntry;
+}
+
+// Set the UIKit::Xaml::TextField's Text property
+UIKIT_XAML_EXPORT void XamlSetTextFieldText(const Microsoft::WRL::ComPtr<IInspectable>&  inspectableTextField, const std::wstring& text) {
+    auto textField = safe_cast<UIKit::Xaml::TextField^>(reinterpret_cast<Platform::Object^>(inspectableTextField.Get()));
+    textField->Text = Platform::StringReference(text.c_str());
+    if (!textField->SecureEntry) {
+        textField->TextBox->SelectionStart = textField->Text->Length();
+        textField->TextBox->SelectionLength = 0;
```
jaredhms: nit: might want to comment about why you're doing this, and only doing it for TextBoxes

Frameworks/UIKit.Xaml/TextField.xaml.cpp
```diff
@@ -0,0 +1,759 @@
...
+        s_borderStyleProperty = DependencyProperty::RegisterAttached("BorderStyle",
+            UIKit::Xaml::TextBorderStyle::typeid,
+            TextField::typeid,
+            nullptr);
+    }
+}
+
+// Accessor for our Layer content
+Image^ TextField::LayerContent::get() {
+    return nullptr;
```
jaredhms: we should either delete _content, or allocate/use it here.

Frameworks/UIKit.Xaml/TextField.xaml
```diff
@@ -0,0 +1,16 @@
...
+        depeneding on the SecureEntry value, if NO, a TextBox is created
+        otherwise, a passwordBox is created. In the meantime, properties
+        binding will be set up in code when controls is loaded.
```
tadam-msft: >controls [](start = 44, length = 8)

control

Frameworks/UIKit.Xaml/TextField.xaml
```diff
@@ -0,0 +1,16 @@
...
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:UIKit.Xaml"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    mc:Ignorable="d"
+    x:Name="Root">
+    <!-- 
+        Current the backing control is dynamically created in the code
+        depeneding on the SecureEntry value, if NO, a TextBox is created
```
tadam-msft: >depeneding [](start = 8, length = 10)

depending

Frameworks/UIKit.Xaml/TextField.xaml.cpp
yiyang-msft: Yes, both have the same elements.

---
In reply to: [113563703](https://github.com/Microsoft/WinObjC/pull/2616#discussion_r113563703) [](ancestors = 113563703)

Frameworks/UIKit.Xaml/TextField.xaml.h
yiyang-msft: only s_textVerticalAlignmentProperty s_borderStyleProperty are not used for bindings to internal controls. But i kind of want to keep them as dependence properties as well.  We might find ways to use binding for them as well. e.g. through Converter etc.

---
In reply to: [113562562](https://github.com/Microsoft/WinObjC/pull/2616#discussion_r113562562) [](ancestors = 113562562)

Frameworks/UIKit.Xaml/TextField.xaml.cpp
yiyang-msft: nice, didn't realize this in on control.

---
In reply to: [113563428](https://github.com/Microsoft/WinObjC/pull/2616#discussion_r113563428) [](ancestors = 113563428)

Frameworks/UIKit.Xaml/TextField.xaml.h
jaredhms: Why do you need data binding for any of them?  Why not just set the properties directly?  DataBinding only really makes sense if you need to use it in markup.

---
In reply to: [113595069](https://github.com/Microsoft/WinObjC/pull/2616#discussion_r113595069) [](ancestors = 113595069,113562562)

Frameworks/UIKit/UITextField.mm
yiyang-msft: good point. the behavior on reference platform is the font change while typing, so this needs to moved to textchanged handler.

---
In reply to: [113561500](https://github.com/Microsoft/WinObjC/pull/2616#discussion_r113561500) [](ancestors = 113561500)

Frameworks/UIKit.Xaml/TextField.xaml.cpp
yiyang-msft: make sure.

---
In reply to: [113563339](https://github.com/Microsoft/WinObjC/pull/2616#discussion_r113563339) [](ancestors = 113563339,113563242)

Frameworks/UIKit.Xaml/TextField.xaml.h
tadam-msft: I think it also makes sense for the two way scenario; but we don't have that here.  

---
In reply to: [113753740](https://github.com/Microsoft/WinObjC/pull/2616#discussion_r113753740) [](ancestors = 113753740,113595069,113562562)

Frameworks/UIKit.Xaml/TextField.xaml.cpp
```diff
@@ -0,0 +1,759 @@
...
+        _keydownHandlerRegistrationToken = {0};
+        
+        unsigned int index = 0;
+        if (Children->IndexOf(_backingControl, &index)) {
+            Children->RemoveAt(index);
+        }
+
+        _backingControl = nullptr;
+    }
+
```
tadam-msft: The selected code could easily pulled out into a separate function and reused--all the APIs are on control, or it could be combined with the next function since much of the code is similar.

Frameworks/UIKit.Xaml/TextField.xaml.h
yiyang-msft: text is two way binding, which i am keeping. for the rest, i changed them to properties.

---
In reply to: [113789238](https://github.com/Microsoft/WinObjC/pull/2616#discussion_r113789238) [](ancestors = 113789238,113753740,113595069,113562562)

Frameworks/UIKit/XamlControls.mm
```diff
@@ -119,6 +118,194 @@ void XamlContentDialogSetDestructiveButtonIndex(const Controls::ContentDialog& c
...
+    XamlSetTextFieldPlaceholder(objcwinrt::to_insp(textField), placeholderText);
+}
+
+NSString* GetTextFieldPlaceholder(const winrt::Windows::UI::Xaml::FrameworkElement& textField) {
+    ComPtr<IInspectable> inspPropVal(XamlGetTextFieldPlaceholder(objcwinrt::to_insp(textField)));
+    return XamlUtilities::NSStringFromPropertyValue(inspPropVal);
+}
+
+void SetTextFieldInputScope(const winrt::Windows::UI::Xaml::FrameworkElement& textField, UIKeyboardType keyboardType) {
+    auto inputSCopeNameValue =
```
jaredhms: >C [](start = 15, length = 1)

nit: inputScopeNameValue

Frameworks/UIKit.Xaml/TextField.xaml.h
```diff
@@ -0,0 +1,282 @@
...
+
+    property Platform::String^ Placeholder
+    {
+        Platform::String^ get()
+        {
+            return _placeholder;
+        }
+        void set(Platform::String^ value)
+        {
+            if (_placeholder != value) {
```
jaredhms: > if (_placeholder != value) { [](start = 11, length = 29)

it seems wasteful to do this check; why not just set whatever is passed in?

Frameworks/UIKit.Xaml/TextField.xaml.cpp
```diff
@@ -0,0 +1,692 @@
...
+// Get the UIKit::Xaml::TextField's PlaceHolder property
+UIKIT_XAML_EXPORT IInspectable* XamlGetTextFieldPlaceholder(const Microsoft::WRL::ComPtr<IInspectable>& inspectableTextField) {
+    auto textField = safe_cast<UIKit::Xaml::TextField^>(reinterpret_cast<Platform::Object^>(inspectableTextField.Get()));
+    Platform::Object^ propVal = Windows::Foundation::PropertyValue::CreateString(textField->Placeholder);
+    return InspectableFromObject(propVal).Detach();
+}
+
+// Set the UIKit::Xaml::TextField's InputScope property
+UIKIT_XAML_EXPORT void XamlSetTextFieldInputScope(const Microsoft::WRL::ComPtr<IInspectable>&  inspectableTextField, int inputSCopeNameValue) {
```
jaredhms: nit: inputScopeNameValue

Frameworks/UIKit.Xaml/TextField.xaml.cpp
```diff
@@ -0,0 +1,692 @@
...
+    return textField->SecureEntry;
+}
+
+// Set the UIKit::Xaml::TextField's Text property
+UIKIT_XAML_EXPORT void XamlSetTextFieldText(const Microsoft::WRL::ComPtr<IInspectable>&  inspectableTextField, const std::wstring& text) {
+    auto textField = safe_cast<UIKit::Xaml::TextField^>(reinterpret_cast<Platform::Object^>(inspectableTextField.Get()));
+    textField->Text = Platform::StringReference(text.c_str());
+    if (!textField->SecureEntry) {
+        // this is to set the selection to the end of of text so that
+        // to ensure when gaining focus programmatically, the caret is at
```
jaredhms: >gaining focus programmatically [](start = 26, length = 30)

should we do this on programmatic focus then, vs. on programmatic text updates?

Frameworks/UIKit.Xaml/TextField.xaml.cpp
```diff
@@ -0,0 +1,692 @@
...
+    BecomeFirstResponder();
+}
+
+bool TextField::BecomeFirstResponder() {
+    if (_backingControl) {
+        return _backingControl->Focus(FocusState::Programmatic);
+    }
+
+    return false;
+}
```
jaredhms: nit: could be folded into a single SetFocus method that returns true/false

jaredhms:
:shipit:

Frameworks/UIKit.Xaml/TextField.xaml.cpp
yiyang-msft: sorry, wrong comments. This is when user tap into the text control. I will correct the comments.

---
In reply to: [113971401](https://github.com/Microsoft/WinObjC/pull/2616#discussion_r113971401) [](ancestors = 113971401)

PR #2617, opened 2017-04-25
Adjust override targets to work better for cross repo scenarios

bbowman:


common/winobjc.versioning.common.targets
```diff
@@ -6,6 +6,7 @@
     <UpdateAssemblyInfo>false</UpdateAssemblyInfo>
     <WriteVersionInfoToBuildLog>false</WriteVersionInfoToBuildLog>
     <_WinObjCVersioningTargetsImported>true</_WinObjCVersioningTargetsImported>
+    <GitVersionPath Condition="Exists('$(MSBuildThisFileDirectory)\..\.git')">$(MSBuildThisFileDirectory)\..\</GitVersionPath>
```
memontic-ms: Where is GitVersionPath used?

bbowman: https://github.com/GitTools/GitVersion/blob/1094a95117488a62e46a3c97e0f2b5d3449acb11/src/GitVersionTask/NugetAssets/buildMultiTargeting/GitVersionTask.targets#L51

bbowman: It tells GitVersion where to look for the tags and other information. So when this guy is imported in a sample in another repo it would previously use that repo's tags which probably don't line up with ours. This roots it to the main repo as long as the .git directory is there (just being overly cautious).

PR #2622, opened 2017-04-26
Remove appx versioning target for WACK tests

### LINKED ISSUE #2621: WACK tests fail due to versioning
```xml
      <TEST INDEX="31" NAME="App manifest" DESCRIPTION="App manifest must include valid entries for all required fields." EXECUTIONTIME="00h:00m:01s.50ms">
        <MESSAGES>
          <MESSAGE TEXT="The revision number in the app package version and bundle version must be set to 0, for example Version=&quot;10.1.2.0&quot;. This is restricted for Windows Store use only." />
        </MESSAGES>
        <RESULT><![CDATA[FAIL]]></RESULT>
      </TEST>
```
This is r a really, really silly restriction but I guess that means I just won't adjust the version only the display name.

bbowman:
Fixes #2621


pradipd:
If you prefer, you can just disable this test in wack.ps1

bbowman:
I think it does more appxmanifest validationso I'd like to keep it

PR #2628, opened 2017-04-26
Fix all of our invalid format strings

DHowett-MSFT:

PR #2629, opened 2017-04-26
Light up NS_FORMAT_FUNCTION and annotate every formatter

DHowett-MSFT:
This pull request must merge _after_ #2628. Otherwise, the build will be broken.


include/Foundation/NSException.h
```diff
@@ -46,8 +46,8 @@ FOUNDATION_EXPORT NSString* const NSOverflowException;
 FOUNDATION_EXPORT_CLASS
 @interface NSException : NSObject <NSCoding, NSCopying>
 + (NSException*)exceptionWithName:(NSString*)name reason:(NSString*)reason userInfo:(NSDictionary*)userInfo;
-+ (void)raise:(NSString*)name format:(NSString*)format, ...;
-+ (void)raise:(NSString*)name format:(NSString*)format arguments:(va_list)argList;
++ (void)raise:(NSString*)name format:(NSString*)format, ... NS_FORMAT_FUNCTION(2, 3);
++ (void)raise:(NSString*)name format:(NSString*)format arguments:(va_list)argList NS_FORMAT_FUNCTION(2, 0);
```
aballway: why is this one (2, 0) ?

DHowett-MSFT: since it takes a `va_arg`, the compiler cannot tell us anything about the arguments -- we still want the format string security check though!

DHowett-MSFT: the reference macro is not expressive enough to capture a single arg

DHowett-MSFT:
@bviglietta FYI, this is the other half of the formatting change -- it introduces build breaks for the things in #2628!

bbowman:
:shipit:

DHowett-MSFT:
It appears that there were formats that changed since my previous format branch forked.

DHowett-MSFT:
This is in build again.

I fully expect that this will cause our warnings to spike; there may even be build failures in the sample app collateral. This is expected and required for us to move forward withformatter safety.

Issue #2619, opened 2017-04-26
HelloUI fails to launch after C++/winrt switch


bbowman:
KERNELBASE!RaiseException+0x62
VCRUNTIME140_APP!_CxxThrowException+0x65
UIKit!winrt::impl::throw_hresult+0x2de
UIKit!winrt::Windows::Foundation::IUnknown::as<winrt::Windows::UI::Xaml::Controls::TextBlock>+0x56
UIKit!-[UILabel _initUILabel]+0x112
UIKit!-[UILabel initWithFrame:]+0x93
HelloUI+0x10f5
UIKit!_UIApplicationMainInit+0x5e6
UIKit!RunApplicationMain+0x4bb
UIKit!DoApplicationLaunch+0x29a
UIKit!_ApplicationLaunch+0x1fd
UIKit!UIApplicationLaunched+0x16a
HelloUI+0x4453
Windows_UI_Xaml!DirectUI::FrameworkApplicationGenerated::OnLaunchedProtected+0x44
Windows_UI_Xaml!DirectUI::FrameworkView::OnActivated+0xbc
Windows_UI_Xaml!Microsoft::WRL::Callback::__l2::<lambda_c07226b0ad2d361bff900a8596192414>::operator()+0x1d
Windows_UI_Xaml!Microsoft::WRL::Details::InvokeHelper<Windows::Foundation::Collections::VectorChangedEventHandler<IInspectable *>,<lambda_c07226b0ad2d361bff900a8596192414>,2,-1>::Invoke+0x28
twinapi_appcore!Microsoft::WRL::EventSource<Windows::Foundation::ITypedEventHandler<Windows::ApplicationModel::Core::CoreApplicationView *,Windows::ApplicationModel::Activation::IActivatedEventArgs *>,Microsoft::WRL::InvokeModeOptions<-2> >::InvokeAll::__l2::<lambda_1ed4baef827929e5ad54274a6ed92695>::operator()+0x18
twinapi_appcore!Microsoft::WRL::InvokeTraits<-2>::InvokeDelegates<<lambda_1ed4baef827929e5ad54274a6ed92695>,Windows::Foundation::ITypedEventHandler<Windows::ApplicationModel::Core::CoreApplicationView *,Windows::ApplicationModel::Activation::IActivatedEventArgs *> >+0x4f
[...]

PR #2634, opened 2017-04-26
Fix 2618 by forcing RTObjCInterop library to be included

### LINKED ISSUE #2618: Sample Apps Fail to launch after C++/winrt switchover
KERNELBASE!RaiseException+0x62
VCRUNTIME140_APP!_CxxThrowException+0x65
UIKit!winrt::impl::throw_hresult+0x2de
UIKit!winrt::Windows::Foundation::IUnknown::as<winrt::Windows::UI::Xaml::Controls::Button>+0x56
UIKit!-[UIButton _initUIButton]+0x92
UIKit!-[UIButton initWithFrame:]+0x93
UIKit!initControls+0x67
UIKit!-[UIBarButtonItem initWithTitle:style:target:action:]+0xa2
UIKit!-[UINavigationBar initWithCoder:]+0x181
UIKit!constructObject+0x1aa
UIKit!idForItem+0x49
UIKit!-[UINibUnarchiver decodeObjectForKey:]+0x2d
UIKit!-[UINavigationController initWithCoder:]+0x28
UIKit!constructObject+0x1aa
UIKit!idForItem+0x49
UIKit!-[UINibUnarchiver decodeObjectForKey:]+0x2d
UIKit!-[UIRuntimeOutletConnection initWithCoder:]+0x36

Seen in AddressBookSample

bviglietta:
Fixes #2618, #2619
The RTObjCInterop library is needed even if projections aren't being used, because UI elements sometimes get manipulated as RTObjects. This change ensures that RTObjCInterop.dll is included in the app package.

### DIFF
--- a/Frameworks/UIKit/ForceInclusion.mm
+++ b/Frameworks/UIKit/ForceInclusion.mm
@@ -25,6 +25,7 @@
 #include "UIClassSwapper.h"
 #include "../QuartzCore/CATransaction.h"
 #include "../Foundation/NSColor.h"
+#import <UWP/InteropBase.h>
 
 extern "C" void NSObjForceinclude();
 void NSStringForceinclude();
@@ -47,6 +48,7 @@ void ForceInclusion() {
     [UISwitch class];
     [UISlider class];
     [UIProxyObject class];
+    [RTObject class];
     NSStringForceinclude();
     NSIndexPathForceInclude();
     NSValueForceinclude();


Frameworks/UIKit/ForceInclusion.mm
```diff
@@ -47,6 +48,7 @@ void ForceInclusion() {
     [UISwitch class];
     [UISlider class];
     [UIProxyObject class];
+    [RTObject class];
```
bbowman: @bviglietta  Should they be treated as something else? It doesn't make sense for them to be used as RTObject if they aren't actually one right?

bviglietta: Well, they really are RTObjects, after we convert them to RTObjects. But you're right that there isn't a lot of reason to do this. One thing that I'd like to do in the near future is eliminate any continued use of RTObject (primarily in the xamlElement property on UIViews) and instead dynamically load the appropriately-typed projection classes as they're needed.

PR #2636, opened 2017-04-26
Remove all instances of IwMalloc and friends

bbowman:


Frameworks/CoreGraphics/CGImage.mm
```diff
@@ -360,13 +360,13 @@ CGDataProviderRef CGImageGetDataProvider(CGImageRef img) {
 
     const unsigned int stride = CGImageGetBytesPerRow(img);
     const unsigned int size = CGImageGetHeight(img) * stride;
-    woc::unique_iw<unsigned char> buffer(static_cast<unsigned char*>(IwMalloc(size)));
+    std::unique_ptr<unsigned char> buffer(static_cast<unsigned char*>(malloc(size)));
```
bviglietta: >std::unique_ptr [](start = 4, length = 15)

This isn't exactly equivalent. unique_iw used IwFree to deallocate memory, while unique_ptr uses operator delete.

We should make sure the alloc and dealloc methods match.

(Also, the right type would be unique_ptr<unsigned char[]>.)

bbowman: @bviglietta good point, I should switch to use new to allocate the buffer instead of malloc.

tests/unittests/Foundation/NSInvocationTests.mm
```diff
@@ -313,21 +327,21 @@ - (void)largeUniformVFPs:(UniformAggregateD4)one
...
-                                    :(char)c4
-                                    :(std::max_align_t)m4
-                                    :(char)c5
-                                    :(std::max_align_t)m5
-                                    :(char)c6
-                                    :(std::max_align_t)m6
-                                    :(char)c7
-                                    :(std::max_align_t)m7
-                                    :(unsigned char)c8 {
+- (void)interleavedAlignmentsCharMax:(char)
```
DHowett-MSFT: lel clang is actually completely wrong here

DHowett-MSFT: it thinks `c1` is the selector name, but it's the argument name. the method is truly called `interleavedAlignmentsCharMax::::::::::::`

Frameworks/Foundation/NSInvocation.mm
```diff
@@ -143,17 +143,17 @@ - (void)setArgument:(void*)buf atIndex:(NSInteger)index {
             // Release old value:
             id oldValue = nil;
             [self getArgument:&oldValue atIndex:index];
-            [(*(id*)buf) retain];
+            [(*(id*)buf)retain];
```
bviglietta: >[(*(id*)buf)retain]; [](start = 12, length = 20)

I guess clang-format thinks we're casting a variable called retain here?

bbowman: yeah I guess so.

---
In reply to: [113569073](https://github.com/Microsoft/WinObjC/pull/2636#discussion_r113569073) [](ancestors = 113569073)

Frameworks/UIKit/UIViewController.mm
```diff
@@ -715,7 +715,7 @@ - (NSString*)nibName {
         NSString* ret = nil;
         NSString* nibPath;
 
-        char* ourClass = IwStrDup(object_getClassName(self));
+        char* ourClass = _strdup(object_getClassName(self));
```
bviglietta: >_strdup [](start = 25, length = 7)

Unrelated to change, but it's unclear why it's necessary to duplicate this string at all.

bbowman: agreed.

---
In reply to: [113573646](https://github.com/Microsoft/WinObjC/pull/2636#discussion_r113573646) [](ancestors = 113573646)

Frameworks/CoreGraphics/DWriteWrapper.mm
```diff
@@ -142,7 +142,7 @@ CFStringRef _CFStringFromLocalizedString(IDWriteLocalizedStrings* localizedStrin
     RETURN_NULL_IF_FAILED(localizedString->GetStringLength(index, &length));
 
     // Get the string. Use length + 1 here as GetString requires room for the null terminator.
-    woc::unique_iw<wchar_t> wcharString(static_cast<wchar_t*>(IwMalloc(sizeof(wchar_t) * (length + 1))));
+    std::unique_ptr<wchar_t[]> wcharString(new wchar_t[sizeof(wchar_t) * (length + 1)]);
```
bviglietta: >sizeof(wchar_t) * [](start = 55, length = 17)

Don't need this anymore.

bbowman: actually this should be malloc still since we hand it off to CF

---
In reply to: [113577598](https://github.com/Microsoft/WinObjC/pull/2636#discussion_r113577598) [](ancestors = 113577598)

Frameworks/UIKit/StarboardXaml/LayerTransaction.mm
```diff
@@ -84,7 +86,7 @@
 class QueuedProperty : public ICompositorTransaction {
 public:
     std::shared_ptr<LayerProxy> _layer;
-    char* _propertyName;
+    std::String _propertyName;
```
bviglietta: >std::String [](start = 4, length = 11)

Is there really a class called std::String?

bbowman: nope! oops!

---
In reply to: [113578228](https://github.com/Microsoft/WinObjC/pull/2636#discussion_r113578228) [](ancestors = 113578228)

PR #2638, opened 2017-04-26
Fix more format strings

DHowett-MSFT:
Casting a `wchar_t*` (win32: `short`) to `unichar*` (always: `short`) is safe only on our platform.

This is the format change delta from #2628 to develop that caused #2629 to break.


bviglietta:
:shipit:

Issue #2624, opened 2017-04-26
[x86] GLKitComplex/WoCCatalog freezes while transitioning between window states in GL scenarios (snapped, free-floating, etc)


andrejo-msft:
Master commit: https://github.com/Microsoft/WinObjC/commit/9d41b1649075a328f9569d811ba66410c4d34364 
Workstation: Z440 (multi-monitor)

I am testing the commit above, and am using a build from [debug|release]/x86 of GLKitComplex.
*also repro'd on WoCCatalog in GL screens.*

I sideload the app.
It goes through its normal set of transformations. 
I use the windows + arrows keys to move it through different states (new-snap, free-floating, maximize).

Expected:
The app moves from one location to another with the tea kettle still turning.

Actual: 
After a few transitions of the app, the tea kettle freezes and the frame no longer renders. glkit eventually closes.

![image](https://cloud.githubusercontent.com/assets/5649461/25448070/318a1294-2a6c-11e7-970e-f01a86db8bdd.png)

This issue started to appear sometime between:

4/7: https://github.com/Microsoft/WinObjC/commit/43caf1228f111227a06863afb3f71f312e658aac
4/18: https://github.com/Microsoft/WinObjC/commit/9d41b1649075a328f9569d811ba66410c4d34364

andrejo-msft:
The same problem can be seen from mousing the app around the screen into corners, the side, etc.

I have not seen the problem occur with non-GL frame as of yet.

DHowett-MSFT:
This sounds like it could be a memory leak.

andrejo-msft:
I don't see a jump in memory use via task manager. 

One note: it's not limited to GLKitComplex; I can repro on WocCatalog's OpenGLES 1.1 screen as well.

rajsesh-msft:
@DHowett-MSFT could you please dig into this?

ehren:
Noticed earlier that some files need ARC enabled in GLKit. I can open a pull unless anyone has a patch in progress.

DHowett-MSFT:
@bviglietta this is happening in some code that was switched to C++/WinRT:

### CAEAGLLayer.mm
```objc
/**
 @Status Interoperable
*/
- (void)setContentsScale:(float)factor {
    [super setContentsScale:factor];

    Media::ScaleTransform scaleTransform;
    scaleTransform.ScaleX(1.0 / factor);
    scaleTransform.ScaleY(1.0 / factor);

[...]

PR #2639, opened 2017-04-26
Fix leaks in various GLKit classes by enabling ARC

ehren:
Noticed this earlier (should've filed an issue!). This is likely partially responsible for #2624 though may not completely solve the issue.

I'm just completing the build / verifying, will add a comment once ready for review.

### DIFF
--- a/Frameworks/GLKit/GLKEffect.mm
+++ b/Frameworks/GLKit/GLKEffect.mm
@@ -25,6 +25,7 @@
 #import "ShaderGen.h"
 #import "ShaderProg.h"
 #import "NSLogging.h"
+#import "AssertARCEnabled.h"
 
 #include <algorithm>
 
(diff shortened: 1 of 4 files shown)


Frameworks/GLKit/GLKEffect.mm
```diff
@@ -528,7 +529,6 @@ - (BOOL)prepareShaders {
         self.shader = [[GLKShaderCache get] addShaderNamed:self.shaderName source:p];
         if (self.shader == nil) {
             NSTraceError(TAG, @"There was a problem generating a shader for material %@", self.shaderName);
-            [p release];
```
ehren: ```addShaderNamed:source:``` doesn't retain ```p``` so this is missing a release in the non-error case were ARC to be left disabled

DHowett-MSFT:
It scares me that turning on ARC was so easy because _nobody used `release` in GLKit` :wink:

ehren:
no issues running WOCCatalog

DHowett-MSFT:
Thank you, @ehren!

PR #2642, opened 2017-04-27
Fix path to custom target/propts in packager

pradipd:


tools/WinObjC.Packaging/WinObjC.Packager.targets
```diff
@@ -71,10 +71,10 @@
       <PackageFile Include="$(MSBuildThisFileDirectory)Package.Default.targets" Condition="Exists('$(MSBuildThisFileDirectory)Package.Default.targets')">
         <PackagePath>build\$(PackageId).Default.targets</PackagePath>
       </PackageFile>
-      <PackageFile Include="$(MSBuildProjectDirectory)\$(PackageId).Custom.targets" Condition="Exists('$(MSBuildProjectDirectory)\$(PackageId).Custom.targets')">
```
pradipd: PackageId == winobjc.facebooksdk
MSBuildProjectName == facebooksdkpackage
So we were using the wrong file name when trying to package the facebooksdkpackage.custom.targets/props.

memontic-ms: It has to be named the PackageId for the targets & props to automatically be imported when the package is included. See [Including MSBuild props and targets in a package](https://docs.microsoft.com/en-us/nuget/create-packages/creating-a-package#including-msbuild-props-and-targets-in-a-package).

pradipd:
I see. I changed the default name of the package.

Issue #2440, opened 2017-04-06
Clean up UIWindow/UIView point/rect conversion - get rid of [UIWindow _convertPoint]


jaredhms:
Many classes were calling [UIWindow convertPoint:(CGPoint) fromView:(UIView*) toView(UIView*)] to do point conversion, but this isn't a public API.  We moved it to UIWindowInternal.h and renamed it to _convertPoint, but we should get rid of it altogether.  The fact that there are 3 views used in the API makes it confusing.

jaredhms:
These are _closer_ to what UIView's APIs should look like, but are not yet validated:


/**
 @Status Interoperable
*/
- (CGRect)convertRect:(CGRect)rect toView:(UIView*)toView {
    UIWindow* ourWindow = [self _getWindowInternal];
    if (!ourWindow ) {
        return rect;
    }

    if (!toView) {
        toView = ourWindow;
    }

    CGPoint pt1 = rect.origin;
    CGPoint pt2 = rect.origin;
    pt2.x += rect.size.width;
    pt2.y += rect.size.height;

    pt1 = [ourWindow convertPoint:pt1 toView:toView];
    pt2 = [ourWindow convertPoint:pt2 toView:toView];

    CGRect ret;
    ret.origin.x = pt1.x < pt2.x ? pt1.x : pt2.x;
    ret.origin.y = pt1.y < pt2.y ? pt1.y : pt2.y;
    ret.size.width = fabsf(pt1.x - pt2.x);
    ret.size.height = fabsf(pt1.y - pt2.y);

    return ret;
}

/**
 @Status Interoperable
*/
- (CGRect)convertRect:(CGRect)rect fromView:(UIView*)fromView {
    UIWindow* ourWindow = [self _getWindowInternal];
    if (!ourWindow) {
        return rect;
    }

    if (!fromView) {
        fromView = ourWindow;
    }

    CGPoint pt1 = rect.origin;
    CGPoint pt2 = rect.origin;
    pt2.x += rect.size.width;
    pt2.y += rect.size.height;

    pt1 = [ourWindow convertPoint:pt1 fromView:fromView];
    pt2 = [ourWindow convertPoint:pt2 fromView:fromView];

    CGRect ret;
    ret.origin.x = pt1.x < pt2.x ? pt1.x : pt2.x;
    ret.origin.y = pt1.y < pt2.y ? pt1.y : pt2.y;
    ret.size.width = fabsf(pt1.x - pt2.x);
[...]

PR #2650, opened 2017-04-28
Fix pointer calculation when running in framework mode

bviglietta:
Our pointer calculation for hit testing and the like doesn't work when WinObjC is running as a framework (i.e., in middleware mode). We often try to convert coordinates into the coordinate space of the top-level UIWindow, but in framework mode there's no meaningful top-level UIWindow. There actually is a top-level UIWindow, but it's just a facade used to present popups; its underlying XAML element is not meaningful and is not part of the visual tree.

This change makes it so that when in framework mode we use the coordinate space of the top-level XAML element in the visual tree, not the top-level UIWindow. In theory this would also work when not running in framework mode, but I was reluctant to make such a sweeping change without a good reason.


Frameworks/UIKit/UITouch.mm
```diff
@@ -225,7 +226,14 @@ - (CGPoint)locationInView:(UIView*)viewAddr {
...
         if (window != nil) {
-            ret = [window _convertPoint:location fromView:window toView:viewAddr];
+            UIWindow* fromWindow = window;
+
+            if (GetCACompositor()->IsRunningAsFramework()) {
+                // In framework mode the parent UIWindow is not visible
+                fromWindow = nil;
+            }
+
+            ret = [window _convertPoint:location fromView:fromWindow toView:viewAddr];
```
bviglietta: >window _convertPoint [](start = 19, length = 20)

This code may look strange. Why are we calling a method on a window that we just declared is not relevant? Well, you see, the _convertPoint method actually makes no sense.

jaredhms:
> + (CGPoint)convertPoint:(CGPoint)point fromLayer:(CALayer*)fromLayer toLayer:(CALayer*)toLayer {

we should also update convertRect: (just below this method) for consistency 

---
Refers to: Frameworks/QuartzCore/CALayer.mm:2417 in 5bbe237. [](commit_id = 5bbe2373a4dd52cc5d70625afc5f54f9ca77cde1, deletion_comment = False)

Frameworks/UIKit/UITouch.mm
jaredhms: yeah, those convert methods on UIWindow don't even belong there - tracked by [https://github.com/Microsoft/WinObjC/issues/2440](https://github.com/Microsoft/WinObjC/issues/2440).  maybe we should just fix instance this to call UIView's point conversion method instead?

---
In reply to: [113979397](https://github.com/Microsoft/WinObjC/pull/2650#discussion_r113979397) [](ancestors = 113979397)

bviglietta: UIView's point conversion method doesn't really work here because there's no UIView that we're converting _from_.

---
In reply to: [113980192](https://github.com/Microsoft/WinObjC/pull/2650#discussion_r113980192) [](ancestors = 113980192,113979397)

jaredhms:
:shipit:

Frameworks/QuartzCore/CALayer.mm
```diff
@@ -2477,24 +2484,31 @@ + (CGRect)convertRect:(CGRect)pos fromLayer:(CALayer*)fromLayer toLayer:(CALayer
...
+    UIElement fromLayerElement = nullptr;
+    UIElement toLayerElement = nullptr;
+
+    if (fromLayer) {
+        // Get the backing sublayer xaml UIElement for fromLayer
+        // We use the sublayer element to support proper point conversion within a scrolled view.
+        fromLayerElement = [fromLayer _getSublayerXamlElement];
+    } else if (GetCACompositor()->IsRunningAsFramework()) {
+        // In framework mode we don't own the topmost layer
+        fromLayerElement = Window::Current().Content();
```
yiyang-msft: >fromLayerElement = Window::Current().Content(); [](start = 8, length = 47)

in your description ; "In theory this would also work when not running in framework mode, but I was reluctant to make such a sweeping change without a good reason.". So in theory, this line should work for either cases?

bviglietta: I think so, yes.

---
In reply to: [113999848](https://github.com/Microsoft/WinObjC/pull/2650#discussion_r113999848) [](ancestors = 113999848)

yiyang-msft:
:shipit:

PR #2652, opened 2017-04-28
Make sure UIAlertViews are always rendered vertically

### LINKED ISSUE #2611: Islandwood House Demo Completed popup is rotated 90 degrees.

jaredhms:
WoCDisplayMode uses a 'presentationTransform' to rotate the presentation surface for certan apps, but we don't want to use that transform to rotate UIAlertViews (else they render sideways if the app is rotated).  This will be handled automatically when we move UIAlertView over Xaml (as we've already done for UIActionSheet), but for now we need to undo the 'presentationTransform' rotation for UIAlertViews.

Fixes #2611.

### DIFF
--- a/Frameworks/UIKit/UIAlertView.mm
+++ b/Frameworks/UIKit/UIAlertView.mm
@@ -28,6 +28,7 @@
 #import <UIKit/UILabel.h>
 #import <UIKit/UIScreen.h>
 #import <UIKit/UIView.h>
+#import <UIKit/UIWindow.h>
 
 #import <Foundation/NSString.h>
 #import <Foundation/NSMutableArray.h>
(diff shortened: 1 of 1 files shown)


Frameworks/UIKit/UIAlertView.mm
```diff
@@ -1,641 +1,671 @@
...
+    alertPriv->_titleLabel = nil;
+    alertPriv->_message = nil;
+    alertPriv->_messageLabel = nil;
+    alertPriv->_cancelText = nil;
+    alertPriv->_cancelView = nil;
+
+    delete alertPriv;
+
+    [super dealloc];
+}
```
DHowett-MSFT: This file is impossible to review; did the line endings change?

yiyang-msft: ?? I don't have problem to view this

---
In reply to: [113998799](https://github.com/Microsoft/WinObjC/pull/2652#discussion_r113998799) [](ancestors = 113998799)

Frameworks/UIKit/UIAlertView.mm
```diff
@@ -1,641 +1,671 @@
...
+/**
+ @Status Interoperable
+*/
+- (BOOL)isVisible {
+    return alertPriv->_isVisible;
+}
+
+- (void) _handleRotation {
+    // We need to offset the 'presentationTransform' used in WOCDispalyMode, because we always want to render alerts vertically.
+    // TODO: We'll remove this when we switch UIAlertView over to a Xaml ContentDialog, since rotation will be handled for us.
```
yiyang-msft: nit: if we have WI#, we can put it here so that we don't forget to remove this.

yiyang-msft:
:shipit:

Frameworks/UIKit/UIAlertView.mm
jaredhms: Basically this whole file will be gutted at that time.  I don't think this particular backlog item has been filed on github though...

tadam-msft:
:shipit:

Frameworks/UIKit/UIAlertView.mm
jaredhms: Doh - nice catch.  Reinstall of git seems to have stomped my settings.

---
In reply to: [113998959](https://github.com/Microsoft/WinObjC/pull/2652#discussion_r113998959) [](ancestors = 113998959,113998799)

Frameworks/UIKit/UIAlertView.mm
```diff
@@ -396,6 +397,32 @@ - (BOOL)isVisible {
     return alertPriv->_isVisible;
 }
 
+- (void) _handleRotation {
+    // We need to offset the 'presentationTransform' used in WOCDispalyMode, because we always want to render alerts vertically.
```
DHowett-MSFT: nit: "dispaly"

PR #2658, opened 2017-05-02
Fix execution order of operations in NSOperationQueue

ms-jihua:
"An operation queue executes its queued operation objects based on their priority and readiness.
If all of the queued operation objects have the same priority and are ready to execute when they are put in the queue—
that is, their ready method returns YES—they are executed in the order in which they were submitted to the queue."

### DIFF
--- a/Frameworks/Foundation/NSOperation.mm
+++ b/Frameworks/Foundation/NSOperation.mm
@@ -56,7 +56,9 @@ static wchar_t TAG[] = L"NSOperation";
     std::recursive_mutex _completionBlockLock; // guards _completionBlock
 
     std::atomic<NSOperationQueuePriority> _queuePriority;
-    std::atomic<BOOL> _inQueue;
+
+    bool _inQueue;
+    unsigned int _queueIndex;
 
     bool _observingSelfIsFinished;
 }
@@ -146,6 +148,8 @@ static const NSString* NSOperationContext = @"context";
         _ready = YES;
         [self addObserver:self forKeyPath:@"isFinished" options:0 context:(void*)NSOperationContext];
         _observingSelfIsFinished = true;
+
+        _queuePriority = NSOperationQueuePriorityNormal;
     }
 
     return self;
@@ -387,8 +391,21 @@ static const NSString* NSOperationContext = @"context";
     }
 }
 
-- (BOOL)_markInQueue {
-    return !_inQueue.fetch_or(YES); // only returns true for the first caller
+- (unsigned int)_queueIndex {
+    @synchronized(self) {
+        return _queueIndex;
+    }
+}
+
+- (BOOL)_canAddToQueueAsIndex:(unsigned int)queueIndex {
+    @synchronized(self) {
+        if (!_inQueue) {
+            _inQueue = true;
+            _queueIndex = queueIndex;
+            return YES;
+        }
+        return NO;
+    }
 }
 
 @end
--- a/Frameworks/Foundation/NSOperationQueue.mm
+++ b/Frameworks/Foundation/NSOperationQueue.mm
@@ -23,6 +23,7 @@
 #import "Starboard.h"
 
 #import <atomic>
+#import <map>
 #import <mutex>
 #import <queue>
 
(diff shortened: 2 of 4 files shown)


Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -51,22 +52,16 @@ static inline long _QOSClassForNSQualityOfService(NSQualityOfService quality) {
...
-
 // STL compare for NSOperation*, sorting by queuePriority, putting the greatest priority first
-struct _NSOperation_ComparePrioritySTL {
+// If queue priorities are equal, put the lower queueIndex first
+struct _NSOperation_CompareSTL {
     bool operator()(NSOperation* lhs, NSOperation* rhs) {
-        return lhs.queuePriority < rhs.queuePriority;
+        const NSOperationQueuePriority priLhs = lhs.queuePriority;
+        const NSOperationQueuePriority priRhs = rhs.queuePriority;
+        if (priLhs != priRhs) {
```
aballway: nit: just `return (lhs.queuePriority < rhs.queuePriority) || ([lhs _queueIndex] > [rhs _queueIndex];`

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -362,18 +361,23 @@ - (void)addOperation:(NSOperation*)op {
  @Status Interoperable
 */
 - (void)addOperations:(NSArray<NSOperation*>*)ops waitUntilFinished:(BOOL)wait {
-    NSArray<NSOperation*>* sortedOps = [ops sortedArrayUsingFunction:_compareOperationPriority context:nullptr];
+    // Sort ops by queue priority and add order
+    std::multimap<NSOperationQueuePriority, NSOperation*, std::greater<NSOperationQueuePriority>> sortedOps;
```
aballway: should this use StrongId to retain the operations for the duration of this?

Frameworks/Foundation/NSOperationQueue.mm
ms-jihua: not equivalent when lhs.queuePriority > rhs.queuePriority and lhs _queueIndex > rhs _queueIndex

Frameworks/Foundation/NSOperationQueue.mm
ms-jihua: unnecessary since the nsarray already retains it

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -254,6 +251,8 @@ - (instancetype)init {
         } else {
             [self _setDispatchQueueUsingQualityOfService];
         }
+
+        _operationIndex = 0;
```
DHowett-MSFT: nit: this is autoinitialized to 0

Frameworks/Foundation/NSOperation.mm
```diff
@@ -56,7 +56,9 @@ @implementation NSOperation {
     std::recursive_mutex _completionBlockLock; // guards _completionBlock
 
     std::atomic<NSOperationQueuePriority> _queuePriority;
-    std::atomic<BOOL> _inQueue;
+
+    bool _inQueue;
+    unsigned int _queueIndex;
```
DHowett-MSFT: nit: I'd prefer `NSUInteger`; it lets us expand seamlessly if we ever widen the `NS*` family of types

ms-jihua: leaving this as is

Frameworks/Foundation/NSOperationQueue.mm
```diff
@@ -252,7 +252,7 @@ - (instancetype)init {
             [self _setDispatchQueueUsingQualityOfService];
         }
 
-        _operationIndex = 0;
+        _operationIndex;
```
DHowett-MSFT: `warning: statement result not used;`

ms-jihua: oh derp

Issue #2649, opened 2017-04-28
[x86] [WOCCatalog] Text Missing in Text Display Mode Page


aballway:
UITextViews will scrollable text and CTLineCreateTruncatedLine are missing text.

msft-Jeyaram:
Taking a look at this

msft-Jeyaram:
The root cause is that the sample was written wrong.
You should be passing CGColor to CoreText, but the sample was passing UIColor. Thus it was not able to render it.

rajsesh-msft:
https://www.youtube.com/watch?v=JYc05gZFly0

msft-Jeyaram:
I'm working on it :)

![image](https://cloud.githubusercontent.com/assets/18413092/25550307/f47ec594-2c2c-11e7-9aac-061685b1615a.png)

msft-Jeyaram:
Closed as per #2662

PR #2662, opened 2017-05-03
bridge UIColor to be treated as a CGColor.

### LINKED ISSUE #2656: CoreGraphics should be able to consume UIColor
Although UIColor and CGColor aren't bridged, in the reference platform a UIColor can take a place of a CGColor for some APIs

msft-Jeyaram:
This is to support cases in CoreText where an attribute can take a CGColor or UIColor.
Fixes #2656, #2649, #2671

### DIFF
--- a/Frameworks/UIKit/UICGColor.mm
+++ b/Frameworks/UIKit/UICGColor.mm
@@ -143,8 +143,12 @@ rgb hsv2rgb(hsv in) {
 
 #pragma region Inits
 
-@implementation _UIColorConcrete {
-    woc::StrongCF<CGColorRef> _cgColor;
+@implementation UICGColorPrototype
+
+PROTOTYPE_CLASS_REQUIRED_IMPLS(_UIColorConcrete)
+
+- (instancetype)init {
+    return [self initWithRed:0 green:0 blue:0 alpha:0];
 }
 
 - (instancetype)initWithHue:(CGFloat)h saturation:(CGFloat)s brightness:(CGFloat)v alpha:(CGFloat)a {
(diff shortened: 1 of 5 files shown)


Frameworks/UIKit/UICGColor.mm
```diff
@@ -161,56 +165,36 @@ - (instancetype)initWithHue:(CGFloat)h saturation:(CGFloat)s brightness:(CGFloat
...
-        return nil;
-    }
-
-    if (self = [super init]) {
-        auto pattern = woc::MakeStrongCF<CGPatternRef>(_CGPatternCreateFromImage(img));
-        _cgColor.attach(CGColorCreateWithPattern(CGImageGetColorSpace(img), pattern, nullptr));
-    }
-
-    return self;
+    auto pattern = woc::MakeStrongCF<CGPatternRef>(_CGPatternCreateFromImage(img));
```
aballway: nit: just inline [image CGImage] instead of... whatever the line above is doing

Frameworks/UIKit/UICGColor.mm
```diff
@@ -219,31 +203,35 @@ - (void)set {
...
+    return reinterpret_cast<CGColorRef>(self);
 }
 
 @end
 
 // Const colors are expected to be a singleton
 @implementation _UIColorConcreteConst
 
++ (id)alloc {
+    return static_cast<id>([UIColor allocWithZone:nil]);
```
DHowett-MSFT: Are we sure this const implementation works with the constant CGColors?

msft-Jeyaram: Yep they do

PR #2663, opened 2017-05-04
Fix double application of CTM on text stroke

### LINKED ISSUE #2613: TextDrawingMode + Context Transformations Seems Incorrect
Using Context translations in conjunction with different text drawing modes seems to produce incorrect values. The code below should produce solid text but instead seems to apply the transformations twice when using a combination of fill modes.

CGContextSaveGState(context);
CGContextScaleCTM(context, 1.0f, -1.0f);
CGContextTranslateCTM(context, bounds.size.width * .2, -bounds.size.height * heightLocation);
CGContextSetTextDrawingMode(context, kCGTextFillStroke);

image

aballway:
Fixes #2613


MSFTFox:
The changes on the images seem pretty drastic. Do we see something comparable from the reference platform? If so, you have my sign off.

aballway:
The reference platform performs a little differently than we do wrt drawing modes and frames but the scale is the same.

---
In reply to: [299242879](https://github.com/Microsoft/WinObjC/pull/2663#issuecomment-299242879) [](ancestors = 299242879)

msft-Jeyaram:
@aballway how does ref platform render differently?
sample?

aballway:
the reference platform will draw each run individually, so with clipping, etc. the regions do not overlap and causes nothing to show up.  

---
In reply to: [299271074](https://github.com/Microsoft/WinObjC/pull/2663#issuecomment-299271074) [](ancestors = 299271074)

PR #2612, opened 2017-04-25
Update obj2winmd to support exporting globals from middleware SDK.

pradipd:
Pushing the objc2winmd.exe binary to winobjc repo.

PR #2674, opened 2017-05-05
1704 blocking issues build breaks

rajsesh-msft:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2674)


pradipd:
I think you'll also need:
https://github.com/Microsoft/WinObjC/pull/2612

Issue #2675, opened 2017-05-05
iLabyrinth and iLabyrinthLite are crashing on app launch


pradipd:
Repro:
take the latest build and launch it.

Callstack:
08 079cdf5c 74ad12bb ntdll!RtlFreeHeap
09 079cdf70 74ad1288 ucrtbase!_free_base
0a 079cdf80 5d046b19 ucrtbase!free
0b 079ce034 5d035b7f AudioToolbox!CAFDecoder::InitForRead+0x599 [d:\build\agent7\_work\17\s\frameworks\audiotoolbox\cafdecoder.mm @ 347] 
0c 079ce870 5d031f7e AudioToolbox!AudioFileCAF::openURL+0xef [d:\build\agent7\_work\17\s\frameworks\audiotoolbox\audiofile.mm @ 402] 
0d 079ce8a0 5d032b4c AudioToolbox!AudioFileOpenURL+0x16e [d:\build\agent7\_work\17\s\frameworks\audiotoolbox\audiofile.mm @ 526] 
0e 079ce8c0 0124b345 AudioToolbox!_ExtAudioFileOpenURL+0x1c [d:\build\agent7\_work\17\s\frameworks\audiotoolbox\audiofile.mm @ 892]

yiyang-msft:
fixed.

PR #2680, opened 2017-05-08
Fix WinObjC Issue #2675 - iLabyrinth and iLabyrinthLite are crashing on app launch

yiyang-msft:
The issue here is ordering of linking. We need make sure we link WinObjCRT later than all instances of msvcrt.  otherwise, mismatched CRT could be linked for different dll for malloc/free which causes badness.

Ideally we should have a build target just execute right before linking phase which look through build items and dynamically move WinObjCRT item to the last. 

Also tested other failing apps like islandwood news/house and verified that they can be launched now  with Tools/Framework/App built with Release/Release/Debug configs.


bbowman:
:shipit:

yiyang-msft:
#2681 is used to track the work of long term solution.

PR #2693, opened 2017-05-11
Implement set mutation KVO methods

### LINKED ISSUE #2641: Implement NSObject (will/did)changeValueForKey:WithSetMutation:usingObjects

### LINKED ISSUE #2709: KVO NSKeyValueObservingOptionOld value invalid when observing multiple keypaths on an object.
If an observer is observing multiple keypaths on the same object, the value for NSKeyValueObservingOptionOld can be invalid for some of the change notifications for a keypath.

This can be reproduced using the following example:

```
@interface TestObject
@property (nonatomic, strong) NSString *v;
@end

@implementation TestObject

- (instancetype)init
{
	if (self = [super init]) {
		self.v = [[NSProcessInfo processInfo] globallyUniqueString];
	}

	return self;
}

@end

@interface Controller : NSObject

@property TestObject *testObject;

@end

@implementation

- (instancetype)init
{
  if (self = [super init]) {
    [self addObserver:self forKeyPath : @"testObject" options : NSKeyValueObservingOptionInitial | NSKeyValueObservingOptionNew | NSKeyValueObservingOptionOld context : NULL];
    [self addObserver:self forKeyPath : @"testObject.v" options : NSKeyValueObservingOptionInitial | NSKeyValueObservingOptionNew | NSKeyValueObservingOptionOld context : NULL];
  }

  return self;
}

- (void)doSomething
{
  self.testObject = [[TestObject alloc] init];
}

-(void)observeValueForKeyPath:(NSString *)keyPath ofObject : (id)object change : (NSDictionary *)change context : (void *)context
{
	id oldValue = [change objectForKey : NSKeyValueChangeOldKey];
	id newValue = [change objectForKey : NSKeyValueChangeNewKey];

	NSLog(@"%@ old %@ new %@", keyPath, oldValue, newValue);
}

@end
```

When run and `[Controller doSomething]` is called the change logging will be...
```
testObject.v old 49A2D6B0-0C0E-D54E-821E-47145BF2E0B7 new 5871790A-36D0-6748-B41C-A8BFB017ACA0
testObject old 49A2D6B0-0C0E-D54E-821E-47145BF2E0B7 new <TestObject: 0ED63AB4>
```

As you can see the old object value for the keyPath ```testObject``` is actually the old value for the ```testObject.v``` keyPath.

aballway:
Implements NSObject willChangeValueForKey:withSetMutation:usingObjects: and didChangeValueForKey:withSetMutation:usingObjects:

Also does a small amount of clean-up

Fixes #2641
Fixes #2709

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2693)

### DIFF
--- a/Frameworks/Foundation/NSKVOSupport.mm
+++ b/Frameworks/Foundation/NSKVOSupport.mm
@@ -530,12 +530,18 @@ static inline NSKeyValueChange _changeFromSetMutationKind(NSKeyValueSetMutationK
 }
 
 static inline id _valueForPendingChangeAtIndexes(
-    id notifyingObject, NSString* key, id rootObject, _NSKVOKeyObserver* keyObserver, NSIndexSet* indexes) {
-    id collection = [notifyingObject valueForKey:key];
-    NSString* restOfKeypath = keyObserver.restOfKeypath;
-    id value = restOfKeypath.length > 0 ? [collection valueForKeyPath:restOfKeypath] : collection;
-    if ([value respondsToSelector:@selector(objectsAtIndexes:)]) {
-        value = [value objectsAtIndexes:indexes];
+    id notifyingObject, NSString* key, NSString* keypath, id rootObject, _NSKVOKeyObserver* keyObserver, NSDictionary* pendingChange) {
+    id value = nil;
+    NSIndexSet* indexes = pendingChange[NSKeyValueChangeIndexesKey];
+    if (indexes) {
+        NSArray* collection = [notifyingObject valueForKey:key];
+        NSString* restOfKeypath = keyObserver.restOfKeypath;
+        value = restOfKeypath.length > 0 ? [collection valueForKeyPath:restOfKeypath] : collection;
+        if ([value respondsToSelector:@selector(objectsAtIndexes:)]) {
+            value = [value objectsAtIndexes:indexes];
+        }
+    } else {
+        value = [rootObject valueForKeyPath:keypath];
     }
 
     return value ?: [NSNull null];
@@ -650,25 +656,26 @@ static void _dispatchDidChange(id notifyingObject, NSString* key, TFunc&& func) 
 */
 - (void)willChange:(NSKeyValueChange)changeKind valuesAtIndexes:(NSIndexSet*)indexes forKey:(NSString*)key {
     if ([self observationInfo]) {
-        _dispatchWillChange(self, key, [ changeKind, indexes, key, notifyingObject = self ](_NSKVOKeyObserver * keyObserver) {
+        _dispatchWillChange(self, key, [ changeKind, indexes, key, notifyingObject = self ](_NSKVOKeyObserver * keyObserver) mutable {
             NSMutableDictionary* change = [NSMutableDictionary
                 dictionaryWithObjectsAndKeys:@(changeKind), NSKeyValueChangeKindKey, indexes, NSKeyValueChangeIndexesKey, nil];
-            NSKeyValueChange localChangeKind = changeKind;
             _NSKVOKeypathObserver* keypathObserver = keyObserver.keypathObserver;
             NSKeyValueObservingOptions options = keypathObserver.options;
             id rootObject = keypathObserver.object;
 
             // The reference platform does not support to-many mutations on nested keypaths.
             // We have to treat them as to-one mutations to support aggregate functions.
-            if (localChangeKind != NSKeyValueChangeSetting && keyObserver.restOfKeypathObserver) {
+            if (changeKind != NSKeyValueChangeSetting && keyObserver.restOfKeypathObserver) {
                 // This only needs to be done in willChange because didChange derives from the existing changeset.
-                change[NSKeyValueChangeKindKey] = @(localChangeKind = NSKeyValueChangeSetting);
+                change[NSKeyValueChangeKindKey] = @(changeKind = NSKeyValueChangeSetting);
                 [change removeObjectForKey:NSKeyValueChangeIndexesKey];
             }
 
-            if ((options & NSKeyValueObservingOptionOld) && localChangeKind != NSKeyValueChangeInsertion) {
+            if ((options & NSKeyValueObservingOptionOld) && changeKind != NSKeyValueChangeInsertion) {
                 // For to-many mutations, we can't get the old values at indexes that have not yet been inserted.
-                change[NSKeyValueChangeOldKey] = _valueForPendingChangeAtIndexes(notifyingObject, key, rootObject, keyObserver, indexes);
+                NSString* keypath = keypathObserver.keypath;
+                change[NSKeyValueChangeOldKey] =
+                    _valueForPendingChangeAtIndexes(notifyingObject, key, keypath, rootObject, keyObserver, change);
             }
 
             keypathObserver.pendingChange = change;
@@ -694,7 +701,8 @@ static void _dispatchDidChange(id notifyingObject, NSString* key, TFunc&& func) 
                 NSString* keypath = keypathObserver.keypath;
                 id newValue = keypathValueCache[keypath];
                 if (!newValue) {
-                    newValue = _valueForPendingChangeAtIndexes(notifyingObject, key, rootObject, keyObserver, indexes);
+                    NSString* keypath = keypathObserver.keypath;
+                    newValue = _valueForPendingChangeAtIndexes(notifyingObject, key, keypath, rootObject, keyObserver, change);
                     keypathValueCache[keypath] = newValue;
                 }
 
--- a/Frameworks/Foundation/NSKVOSupport.mm
+++ b/Frameworks/Foundation/NSKVOSupport.mm
@@ -530,7 +530,7 @@ static inline NSKeyValueChange _changeFromSetMutationKind(NSKeyValueSetMutationK
 }
 
 static inline id _valueForPendingChangeAtIndexes(
-    id notifyingObject, NSString* key, id rootObject, NSString* keypath, _NSKVOKeyObserver* keyObserver, NSIndexSet* indexes) {
+    id notifyingObject, NSString* key, id rootObject, _NSKVOKeyObserver* keyObserver, NSIndexSet* indexes) {
     id collection = [notifyingObject valueForKey:key];
     NSString* restOfKeypath = keyObserver.restOfKeypath;
     id value = restOfKeypath.length > 0 ? [collection valueForKeyPath:restOfKeypath] : collection;
@@ -550,8 +550,19 @@ inline static void _dispatchWillChange(id notifyingObject, NSString* key, TFunc&
 
         // Skip any keypaths that are in the process of changing.
         if ([keypathObserver pushWillChange]) {
-            // Call into the lambda function, which will do the actual set-up and notifying if necessary
+            // Call into the lambda function, which will do the actual set-up for pendingChanges
             func(keyObserver);
+
+            NSKeyValueObservingOptions options = keypathObserver.options;
+            if (options & NSKeyValueObservingOptionPrior) {
+                NSMutableDictionary* change = keypathObserver.pendingChange;
+                [change setObject:@(YES) forKey:NSKeyValueChangeNotificationIsPriorKey];
+                [keypathObserver.observer observeValueForKeyPath:keypathObserver.keypath
+                                                        ofObject:keypathObserver.object
+                                                          change:change
+                                                         context:keypathObserver.context];
+                [change removeObjectForKey:NSKeyValueChangeNotificationIsPriorKey];
+            }
 
             // This must happen regardless of whether we are currently notifying.
             _removeNestedObserversAndOptionallyDependents(keyObserver, false);
@@ -572,12 +583,12 @@ static void _dispatchDidChange(id notifyingObject, NSString* key, TFunc&& func) 
         // Skip any keypaths that are in the process of changing.
         _NSKVOKeypathObserver* keypathObserver = keyObserver.keypathObserver;
         if ([keypathObserver popDidChange]) {
+            // Call into lambda, which will do set-up for finalizing changes dictionary
             func(keyObserver, keypathValueCache);
 
             id observer = keypathObserver.observer;
             NSString* keypath = keypathObserver.keypath;
             id rootObject = keypathObserver.object;
-            NSKeyValueObservingOptions options = keypathObserver.options;
             NSMutableDictionary* change = keypathObserver.pendingChange;
             void* context = keypathObserver.context;
             [observer observeValueForKeyPath:keypath ofObject:rootObject change:change context:context];
@@ -591,25 +602,18 @@ static void _dispatchDidChange(id notifyingObject, NSString* key, TFunc&& func) 
 */
 - (void)willChangeValueForKey:(NSString*)key {
     if ([self observationInfo]) {
-        NSMutableDictionary* change = [NSMutableDictionary dictionaryWithObject:@(NSKeyValueChangeSetting) forKey:NSKeyValueChangeKindKey];
-        _dispatchWillChange(self, key, [change](_NSKVOKeyObserver* keyObserver) {
+        _dispatchWillChange(self, key, [](_NSKVOKeyObserver* keyObserver) {
+            NSMutableDictionary* change =
+                [NSMutableDictionary dictionaryWithObject:@(NSKeyValueChangeSetting) forKey:NSKeyValueChangeKindKey];
             _NSKVOKeypathObserver* keypathObserver = keyObserver.keypathObserver;
             NSKeyValueObservingOptions options = keypathObserver.options;
-            id rootObject = keypathObserver.object;
-            NSString* keypath = keypathObserver.keypath;
 
             if (options & NSKeyValueObservingOptionOld) {
                 // For to-many mutations, we can't get the old values at indexes that have not yet been inserted.
+                id rootObject = keypathObserver.object;
+                NSString* keypath = keypathObserver.keypath;
                 id oldValue = [rootObject valueForKeyPath:keypath] ?: [NSNull null];
                 change[NSKeyValueChangeOldKey] = oldValue;
-            }
-
-            if (options & NSKeyValueObservingOptionPrior) {
-                id observer = keypathObserver.observer;
-                void* context = keypathObserver.context;
-                [change setObject:@(YES) forKey:NSKeyValueChangeNotificationIsPriorKey];
-                [observer observeValueForKeyPath:keypath ofObject:rootObject change:change context:context];
-                [change removeObjectForKey:NSKeyValueChangeNotificationIsPriorKey];
             }
 
             keypathObserver.pendingChange = change;
@@ -625,10 +629,9 @@ static void _dispatchDidChange(id notifyingObject, NSString* key, TFunc&& func) 
         _dispatchDidChange(self, key, [](_NSKVOKeyObserver* keyObserver, NSMutableDictionary* keypathValueCache) {
             _NSKVOKeypathObserver* keypathObserver = keyObserver.keypathObserver;
             NSKeyValueObservingOptions options = keypathObserver.options;
-            NSString* keypath = keypathObserver.keypath;
             NSMutableDictionary* change = keypathObserver.pendingChange;
-
             if ((options & NSKeyValueObservingOptionNew) && [change[NSKeyValueChangeKindKey] integerValue] != NSKeyValueChangeRemoval) {
+                NSString* keypath = keypathObserver.keypath;
                 id rootObject = keypathObserver.object;
                 id newValue = keypathValueCache[keypath];
                 if (!newValue) {
@@ -647,15 +650,13 @@ static void _dispatchDidChange(id notifyingObject, NSString* key, TFunc&& func) 
 */
 - (void)willChange:(NSKeyValueChange)changeKind valuesAtIndexes:(NSIndexSet*)indexes forKey:(NSString*)key {
     if ([self observationInfo]) {
-        NSMutableDictionary* change = [NSMutableDictionary
-            dictionaryWithObjectsAndKeys:@(changeKind), NSKeyValueChangeKindKey, indexes, NSKeyValueChangeIndexesKey, nil];
-
-        _dispatchWillChange(self, key, [ changeKind, indexes, key, notifyingObject = self, change ](_NSKVOKeyObserver * keyObserver) {
+        _dispatchWillChange(self, key, [ changeKind, indexes, key, notifyingObject = self ](_NSKVOKeyObserver * keyObserver) {
+            NSMutableDictionary* change = [NSMutableDictionary
+                dictionaryWithObjectsAndKeys:@(changeKind), NSKeyValueChangeKindKey, indexes, NSKeyValueChangeIndexesKey, nil];
             NSKeyValueChange localChangeKind = changeKind;
             _NSKVOKeypathObserver* keypathObserver = keyObserver.keypathObserver;
             NSKeyValueObservingOptions options = keypathObserver.options;
             id rootObject = keypathObserver.object;
-            NSString* keypath = keypathObserver.keypath;
 
             // The reference platform does not support to-many mutations on nested keypaths.
             // We have to treat them as to-one mutations to support aggregate functions.
@@ -667,16 +668,7 @@ static void _dispatchDidChange(id notifyingObject, NSString* key, TFunc&& func) 
 
             if ((options & NSKeyValueObservingOptionOld) && localChangeKind != NSKeyValueChangeInsertion) {
                 // For to-many mutations, we can't get the old values at indexes that have not yet been inserted.
-                change[NSKeyValueChangeOldKey] =
-                    _valueForPendingChangeAtIndexes(notifyingObject, key, rootObject, keypath, keyObserver, indexes);
-            }
-
-            if (options & NSKeyValueObservingOptionPrior) {
-                id observer = keypathObserver.observer;
-                void* context = keypathObserver.context;
-                [change setObject:@(YES) forKey:NSKeyValueChangeNotificationIsPriorKey];
-                [observer observeValueForKeyPath:keypath ofObject:rootObject change:change context:context];
-                [change removeObjectForKey:NSKeyValueChangeNotificationIsPriorKey];
+                change[NSKeyValueChangeOldKey] = _valueForPendingChangeAtIndexes(notifyingObject, key, rootObject, keyObserver, indexes);
             }
 
             keypathObserver.pendingChange = change;
@@ -695,15 +687,14 @@ static void _dispatchDidChange(id notifyingObject, NSString* key, TFunc&& func) 
                                         notifyingObject = self ](_NSKVOKeyObserver * keyObserver, NSMutableDictionary * keypathValueCache) {
             _NSKVOKeypathObserver* keypathObserver = keyObserver.keypathObserver;
             NSKeyValueObservingOptions options = keypathObserver.options;
-            id rootObject = keypathObserver.object;
-            NSString* keypath = keypathObserver.keypath;
             NSMutableDictionary* change = keypathObserver.pendingChange;
-
             if ((options & NSKeyValueObservingOptionNew) && [change[NSKeyValueChangeKindKey] integerValue] != NSKeyValueChangeRemoval) {
                 // For to-many mutations, we can't get the new values at indexes that have been deleted.
+                id rootObject = keypathObserver.object;
+                NSString* keypath = keypathObserver.keypath;
                 id newValue = keypathValueCache[keypath];
                 if (!newValue) {
-                    newValue = _valueForPendingChangeAtIndexes(notifyingObject, key, rootObject, keypath, keyObserver, indexes);
+                    newValue = _valueForPendingChangeAtIndexes(notifyingObject, key, rootObject, keyObserver, indexes);
                     keypathValueCache[keypath] = newValue;
                 }
 
@@ -722,8 +713,8 @@ static const NSString* _NSKeyValueChangeOldSetValue = @"_NSKeyValueChangeOldSetV
 - (void)willChangeValueForKey:(NSString*)key withSetMutation:(NSKeyValueSetMutationKind)mutationKind usingObjects:(NSSet*)objects {
     if ([self observationInfo]) {
         NSKeyValueChange changeKind = _changeFromSetMutationKind(mutationKind);
-        NSMutableDictionary* change = [NSMutableDictionary dictionaryWithObject:@(changeKind) forKey:NSKeyValueChangeKindKey];
-        _dispatchWillChange(self, key, [mutationKind, objects, change, changeKind](_NSKVOKeyObserver* keyObserver) {
+        _dispatchWillChange(self, key, [mutationKind, objects, changeKind](_NSKVOKeyObserver* keyObserver) {
+            NSMutableDictionary* change = [NSMutableDictionary dictionaryWithObject:@(changeKind) forKey:NSKeyValueChangeKindKey];
             _NSKVOKeypathObserver* keypathObserver = keyObserver.keypathObserver;
             NSKeyValueObservingOptions options = keypathObserver.options;
             id rootObject = keypathObserver.object;
@@ -750,14 +741,6 @@ static const NSString* _NSKeyValueChangeOldSetValue = @"_NSKeyValueChangeOldSetV
                 }
             }
 
-            if (options & NSKeyValueObservingOptionPrior) {
-                id observer = keypathObserver.observer;
-                void* context = keypathObserver.context;
-                [change setObject:@(YES) forKey:NSKeyValueChangeNotificationIsPriorKey];
-                [observer observeValueForKeyPath:keypath ofObject:rootObject change:change context:context];
-                [change removeObjectForKey:NSKeyValueChangeNotificationIsPriorKey];
-            }
-
             if (options & NSKeyValueObservingOptionNew) {
                 // Save old value in change dictionary for didChangeValueForKey:withSetMutation:usingObjects:
                 // to use for determining added objects
@@ -782,7 +765,6 @@ static const NSString* _NSKeyValueChangeOldSetValue = @"_NSKeyValueChangeOldSetV
 
             if ((options & NSKeyValueObservingOptionNew) && changeKind != NSKeyValueChangeRemoval) {
                 // New values only exist for inserting or replacing, not removing
-                NSKeyValueObservingOptions options = keypathObserver.options;
                 NSString* keypath = keypathObserver.keypath;
                 NSMutableDictionary* change = keypathObserver.pendingChange;
                 NSSet* newValue = keypathValueCache[keypath];
--- a/Frameworks/Foundation/NSKVOSupport.mm
+++ b/Frameworks/Foundation/NSKVOSupport.mm
@@ -1,6 +1,6 @@
 //******************************************************************************
 //
-// Copyright (c) 2015 Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
 // Copyright (c) 2006-2009 Johannes Fortmann, Cocotron Contributors et al.
 //
 // This code is licensed under the MIT License (MIT).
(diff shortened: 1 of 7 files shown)


Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -472,8 +472,7 @@ + (BOOL)automaticallyNotifiesObserversForKey:(NSString*)key {
 @Status Interoperable
 */
 + (NSSet*)keyPathsForValuesAffectingValueForKey:(NSString*)key {
-    NSSet* ret = _keyPathsForValuesAffectingValueForKey(self, key);
-    return ret ? ret : [NSSet set];
+    return _keyPathsForValuesAffectingValueForKey(self, key) ?: [NSSet set];
```
DHowett-MSFT: nit: got a problem?

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -537,17 +536,15 @@ static id _valueForPendingChange(
...
+static void _dispatchWillChange(id notifyingObject, NSString* key, NSMutableDictionary* change, bool setMutation = false) {
     _NSKVOObservationInfo* observationInfo = (__bridge _NSKVOObservationInfo*)[notifyingObject observationInfo];
 
     NSUInteger changeKind = [change[NSKeyValueChangeKindKey] unsignedIntegerValue];
     for (_NSKVOKeyObserver* keyObserver in [observationInfo observersForKey:key]) {
         _NSKVOKeypathObserver* keypathObserver = keyObserver.keypathObserver;
 
-        if (![keypathObserver pushWillChange]) {
-            // Skip any keypaths that are in the process of changing.
-            continue;
```
DHowett-MSFT: got a problem?
this is more specific.

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -604,81 +600,82 @@ static void _dispatchDidChange(id notifyingObject, NSString* key) {
         // This must happen regardless of whether we are currently notifying.
         _addNestedObserversAndOptionallyDependents(keyObserver, false);
 
-        if (![keypathObserver popDidChange]) {
-            // Skip any keypaths that are in the process of changing.
-            continue;
-        }
+        // Skip any keypaths that are in the process of changing.
+        if ([keypathObserver popDidChange]) {
```
DHowett-MSFT: i wish this didn't have such a large diff

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -604,81 +600,82 @@ static void _dispatchDidChange(id notifyingObject, NSString* key) {
...
     }
 }
 
 /**
 @Status Interoperable
 */
 - (void)didChangeValueForKey:(NSString*)key {
-    if (![self observationInfo]) {
-        return;
+    if ([self observationInfo]) {
```
DHowett-MSFT: why?

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -604,81 +600,82 @@ static void _dispatchDidChange(id notifyingObject, NSString* key) {
...
-    _dispatchDidChange(self, key);
 }
 
 /**
 @Status Interoperable
 */
 - (void)willChange:(NSKeyValueChange)change valuesAtIndexes:(NSIndexSet*)indexes forKey:(NSString*)key {
-    if (![self observationInfo]) {
-        return;
+    if ([self observationInfo]) {
```
DHowett-MSFT: why?

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -604,81 +600,82 @@ static void _dispatchDidChange(id notifyingObject, NSString* key) {
...
-                            dictionaryWithObjectsAndKeys:@(change), NSKeyValueChangeKindKey, indexes, NSKeyValueChangeIndexesKey, nil]);
 }
 
 /**
 @Status Interoperable
 */
 - (void)didChange:(NSKeyValueChange)change valuesAtIndexes:(NSIndexSet*)indexes forKey:(NSString*)key {
-    if (![self observationInfo]) {
-        return;
+    if ([self observationInfo]) {
```
DHowett-MSFT: why?

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -604,81 +600,82 @@ static void _dispatchDidChange(id notifyingObject, NSString* key) {
...
 }
 
 /**
- @Status Stub
+ @Status Interoperable
 */
-- (void)willChangeValueForKey:(NSString*)key withSetMutation:(NSKeyValueSetMutationKind)mutationKind usingObjects:(NSSet*)options {
-    UNIMPLEMENTED();
+- (void)willChangeValueForKey:(NSString*)key withSetMutation:(NSKeyValueSetMutationKind)mutationKind usingObjects:(NSSet*)objects {
+    if ([self observationInfo]) {
```
DHowett-MSFT: why?

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -604,81 +600,82 @@ static void _dispatchDidChange(id notifyingObject, NSString* key) {
...
 }
 
 /**
- @Status Stub
+ @Status Interoperable
 */
-- (void)didChangeValueForKey:(NSString*)key withSetMutation:(NSKeyValueSetMutationKind)mutationKind usingObjects:(NSSet*)options {
-    UNIMPLEMENTED();
+- (void)didChangeValueForKey:(NSString*)key withSetMutation:(NSKeyValueSetMutationKind)mutationKind usingObjects:(NSSet*)objects {
+    if ([self observationInfo]) {
```
DHowett-MSFT: why?

Frameworks/Foundation/NSKVOSwizzling.mm
```diff
@@ -128,12 +128,12 @@ static void notifyingVariadicSetImpl(id self, SEL _cmd, ...) {
     // in the same way: on the stack. For our two supported platforms, this seems to hold true.
     NSMethodSignature* sig = [self methodSignatureForSelector:_cmd];
     auto argSz = [sig getArgumentSizeAtIndex:2];
-    auto nStackArgs = argSz / sizeof(unsigned);
-    unsigned* raw = static_cast<unsigned*>(calloc(sizeof(unsigned), nStackArgs));
+    auto nStackArgs = argSz / sizeof(unsigned int);
+    unsigned int* raw = static_cast<unsigned int*>(calloc(sizeof(unsigned int), nStackArgs));
```
DHowett-MSFT: cut it out.

Frameworks/Foundation/NSKVOSwizzling.mm
```diff
@@ -166,7 +166,7 @@ static void notifyingVariadicSetImpl(id self, SEL _cmd, ...) {
             imp(self, _cmd, raw[0], raw[1], raw[2], raw[3], raw[4], raw[5]);
             break;
         default:
-            FAIL_FAST_MSG("Can't override setter with more than 6 sizeof(long) stack arguments.");
+            FAIL_FAST_MSG("Can't override setter with more than 6 sizeof(long int) stack arguments.");
```
DHowett-MSFT: seriously?

Frameworks/Foundation/NSKVOSwizzling.mm
```diff
@@ -385,6 +406,10 @@ static void _NSKVOEnsureCollectionWillNotify(id object, NSString* key, const cha
                                formatSelector(@"replace%sAtIndexes:with%s:", rawKey, rawKey),
                                key,
                                (IMP)NSKVONotifying$replaceXxxAtIndexes$withXxx$);
+    replaceAndAssociateWithKey(object, formatSelector(@"add%s:", rawKey), key, (IMP)NSKVONotifying$addXxx$);
```
DHowett-MSFT: is swizzling every method appropriate here?

we're only supposed to swizzle in the array case if it has an insertion method. this changes that logic.

tests/unittests/Foundation/NSObject_KeyValueObservationTests.mm
```diff
@@ -54,6 +56,7 @@ - (void)dealloc {
     [_manualNotificationArray release];
     [_kvcMediatedArray release];
```
DHowett-MSFT: add a kvcMediatedSet test.

tests/unittests/Foundation/NSObject_KeyValueObservationTests.mm
```diff
@@ -54,6 +56,7 @@ - (void)dealloc {
     [_manualNotificationArray release];
```
DHowett-MSFT: add a manualNotificationSet test

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -577,24 +572,25 @@ static void _dispatchWillChange(id notifyingObject, NSString* key, NSMutableDict
...
 }
 
 /**
 @Status Interoperable
 */
 - (void)willChangeValueForKey:(NSString*)key {
-    if (![self observationInfo]) {
-        return;
+    if ([self observationInfo]) {
+        _dispatchWillChange(self,
```
msft-Jeyaram: lol, is this for real? :P

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -604,81 +600,82 @@ static void _dispatchDidChange(id notifyingObject, NSString* key) {
...
 
 /**
 @Status Interoperable
 */
 - (void)didChange:(NSKeyValueChange)change valuesAtIndexes:(NSIndexSet*)indexes forKey:(NSString*)key {
-    if (![self observationInfo]) {
-        return;
+    if ([self observationInfo]) {
+        _dispatchDidChange(self, key);
     }
```
msft-Jeyaram: unnecessary changes and deviating from convention.

Frameworks/Foundation/NSKVOSwizzling.mm
```diff
@@ -253,21 +243,65 @@ static void notifyingVariadicSetImpl(id self, SEL _cmd, ...) {
...
+    [self willChangeValueForKey:key withSetMutation:NSKeyValueUnionSetMutation usingObjects:set];
+
+    objc_super super{ self, ABI_SUPER(self) };
+    auto imp = (void (*)(id, SEL, NSSet*))objc_msg_lookup_super(&super, _cmd);
+    imp(self, _cmd, set);
+
+    [self didChangeValueForKey:key withSetMutation:NSKeyValueUnionSetMutation usingObjects:set];
+}
+
+static void NSKVONotifying$removeXxx$(id self, SEL _cmd, NSSet* set) {
```
msft-Jeyaram: These are all the same with one parameter variant, make this into a utility function and pass the variant parameter.

Frameworks/Foundation/NSKVOSwizzling.mm
ms-jihua: +1. unnngghhhhh

Frameworks/Foundation/NSKVOSwizzling.mm
```diff
@@ -128,12 +128,12 @@ static void notifyingVariadicSetImpl(id self, SEL _cmd, ...) {
...
     auto argSz = [sig getArgumentSizeAtIndex:2];
-    auto nStackArgs = argSz / sizeof(unsigned);
-    unsigned* raw = static_cast<unsigned*>(calloc(sizeof(unsigned), nStackArgs));
+    auto nStackArgs = argSz / sizeof(unsigned int);
+    unsigned int* raw = static_cast<unsigned int*>(calloc(sizeof(unsigned int), nStackArgs));
     va_list ap;
     va_start(ap, _cmd);
     for (unsigned int i = 0; i < nStackArgs; ++i) {
-        raw[i] = va_arg(ap, unsigned);
+        raw[i] = va_arg(ap, unsigned int);
```
DHowett-MSFT: since you are changing this code, you should bring it up to par with NSInvocation's perfect forwarding implementation

..
..


..
..

..
..

😀

aballway: http://i.imgur.com/Xn4hKRC.gif

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -516,169 +515,301 @@ - (void)removeObserver:(id)observer forKeyPath:(NSString*)keyPath {
...
+    if ([value respondsToSelector:@selector(objectsAtIndexes:)]) {
+        value = [value objectsAtIndexes:indexes];
     }
-    return value;
+
+    return value ?: [NSNull null];
 }
 
-static void _dispatchWillChange(id notifyingObject, NSString* key, NSMutableDictionary* change) {
+// void TFunc(_NSKVOKeyObserver* keybserver);
```
DHowett-MSFT: `keyb?`

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -516,169 +515,301 @@ - (void)removeObserver:(id)observer forKeyPath:(NSString*)keyPath {
...
-    if (![self observationInfo]) {
-        return;
+- (void)willChange:(NSKeyValueChange)changeKind valuesAtIndexes:(NSIndexSet*)indexes forKey:(NSString*)key {
+    if ([self observationInfo]) {
+        _dispatchWillChange(self, key, [&changeKind, indexes, key, notifyingObject = self ](_NSKVOKeyObserver * keyObserver) {
+            _NSKVOKeypathObserver* keypathObserver = keyObserver.keypathObserver;
+            NSKeyValueObservingOptions options = keypathObserver.options;
+            id rootObject = keypathObserver.object;
+            NSString* keypath = keypathObserver.keypath;
+            NSMutableDictionary* change = [NSMutableDictionary dictionaryWithDictionary:@{
```
DHowett-MSFT: perf: creating two dictionaries _per observer_

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -516,169 +515,301 @@ - (void)removeObserver:(id)observer forKeyPath:(NSString*)keyPath {
...
 
 /**
 @Status Interoperable
 */
-- (void)didChangeValueForKey:(NSString*)key {
-    if (![self observationInfo]) {
-        return;
+- (void)willChange:(NSKeyValueChange)changeKind valuesAtIndexes:(NSIndexSet*)indexes forKey:(NSString*)key {
+    if ([self observationInfo]) {
+        _dispatchWillChange(self, key, [&changeKind, indexes, key, notifyingObject = self ](_NSKVOKeyObserver * keyObserver) {
```
DHowett-MSFT: format: ` = self ]`

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -516,169 +515,301 @@ - (void)removeObserver:(id)observer forKeyPath:(NSString*)keyPath {
...
+            NSMutableDictionary* change = [NSMutableDictionary dictionaryWithDictionary:@{
+                NSKeyValueChangeKindKey : @(changeKind),
+                NSKeyValueChangeIndexesKey : indexes
+            }];
+
+            // The reference platform does not support to-many mutations on nested keypaths.
+            // We have to treat them as to-one mutations to support aggregate functions.
+            if (changeKind != NSKeyValueChangeSetting && keyObserver.restOfKeypathObserver) {
+                // This only needs to be done in willChange because didChange derives from the existing changeset.
+                change[NSKeyValueChangeKindKey] = @(changeKind = NSKeyValueChangeSetting);
```
DHowett-MSFT: nit: using side effect from `=`

Frameworks/Foundation/NSKVOSupport.mm
DHowett-MSFT: capturing by reference here is not likely the correct thing to do.

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -516,169 +515,301 @@ - (void)removeObserver:(id)observer forKeyPath:(NSString*)keyPath {
...
+            NSString* keypath = keypathObserver.keypath;
+            NSKeyValueChange changeKind = _changeFromSetMutationKind(mutationKind);
+            NSMutableDictionary* change = [NSMutableDictionary dictionaryWithObject:@(changeKind) forKey:NSKeyValueChangeKindKey];
+
+            NSSet* oldValues = [rootObject valueForKeyPath:keypath];
+            if ((options & NSKeyValueObservingOptionOld) && changeKind != NSKeyValueChangeInsertion) {
+                // The old value should only contain values which are removed from the original dictionary
+                switch (mutationKind) {
+                    case NSKeyValueMinusSetMutation:
+                        // Objects removed are in both oldValues and objects
```
DHowett-MSFT: expand these a little more; what do they mean?

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -516,169 +515,301 @@ - (void)removeObserver:(id)observer forKeyPath:(NSString*)keyPath {
...
+                void* context = keypathObserver.context;
+                [change setObject:@(YES) forKey:NSKeyValueChangeNotificationIsPriorKey];
+                [observer observeValueForKeyPath:keypath ofObject:rootObject change:change context:context];
+                [change removeObjectForKey:NSKeyValueChangeNotificationIsPriorKey];
+            }
+
+            if (options & NSKeyValueObservingOptionNew) {
+                // Save old value in change dictionary for didChangeValueForKey:withSetMutation:usingObjects: to use for determining added
+                // objects
+                // Only needed if observer wants New value
```
DHowett-MSFT: you can short-circuit out if the observer does _not_ want the new value

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -516,169 +515,301 @@ - (void)removeObserver:(id)observer forKeyPath:(NSString*)keyPath {
...
+                change[_NSKeyValueChangeOldSetValue] = [[oldValues copy] autorelease];
```
DHowett-MSFT: so like, this will be "the entire set"?

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -516,169 +515,301 @@ - (void)removeObserver:(id)observer forKeyPath:(NSString*)keyPath {
...
+                    keypathValueCache[keypath] = newValue;
+                }
 
-/**
- @Status Stub
-*/
-- (void)didChangeValueForKey:(NSString*)key withSetMutation:(NSKeyValueSetMutationKind)mutationKind usingObjects:(NSSet*)options {
-    UNIMPLEMENTED();
+                change[NSKeyValueChangeNewKey] = newValue;
+                change[_NSKeyValueChangeOldSetValue] = nil;
```
DHowett-MSFT: please prefer `removeObjectForKey:` -- remember how painful it was to figure out that subscript assignment to nil deleted?

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -516,169 +515,301 @@ - (void)removeObserver:(id)observer forKeyPath:(NSString*)keyPath {
...
+            NSMutableDictionary* change = keypathObserver.pendingChange;
+            void* context = keypathObserver.context;
+            NSKeyValueChange changeKind = _changeFromSetMutationKind(mutationKind);
+
+            if ((options & NSKeyValueObservingOptionNew) && changeKind != NSKeyValueChangeRemoval) {
+                // For to-many mutations, we can't get the new values at indexes that have been deleted.
+                NSSet* newValue = keypathValueCache[keypath];
+                if (!newValue) {
+                    NSSet* oldValues = change[_NSKeyValueChangeOldSetValue];
+                    // The new value should only contain values which are added to the original dictionary
```
DHowett-MSFT: to the original dictionary? or to the original set?

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -516,169 +515,301 @@ - (void)removeObserver:(id)observer forKeyPath:(NSString*)keyPath {
...
+            NSKeyValueObservingOptions options = keypathObserver.options;
+            id rootObject = keypathObserver.object;
+            id observer = keypathObserver.observer;
+            NSString* keypath = keypathObserver.keypath;
+            NSMutableDictionary* change = keypathObserver.pendingChange;
+            void* context = keypathObserver.context;
+            NSKeyValueChange changeKind = _changeFromSetMutationKind(mutationKind);
+
+            if ((options & NSKeyValueObservingOptionNew) && changeKind != NSKeyValueChangeRemoval) {
+                // For to-many mutations, we can't get the new values at indexes that have been deleted.
```
DHowett-MSFT: this comment was clearly copied, since indexes don't exist

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -516,169 +515,301 @@ - (void)removeObserver:(id)observer forKeyPath:(NSString*)keyPath {
...
- @Status Stub
-*/
-- (void)didChangeValueForKey:(NSString*)key withSetMutation:(NSKeyValueSetMutationKind)mutationKind usingObjects:(NSSet*)options {
-    UNIMPLEMENTED();
+                change[NSKeyValueChangeNewKey] = newValue;
+                change[_NSKeyValueChangeOldSetValue] = nil;
+            }
+
+            [observer observeValueForKeyPath:keypath ofObject:rootObject change:change context:context];
+            keypathObserver.pendingChange = nil;
```
DHowett-MSFT: this is common to all three `DidChange`s, right?

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -516,169 +515,301 @@ - (void)removeObserver:(id)observer forKeyPath:(NSString*)keyPath {
...
+    if ([self observationInfo]) {
+        _dispatchDidChange(self, key, [mutationKind, objects](_NSKVOKeyObserver* keyObserver, NSMutableDictionary* keypathValueCache) {
+            _NSKVOKeypathObserver* keypathObserver = keyObserver.keypathObserver;
+            NSKeyValueObservingOptions options = keypathObserver.options;
+            id rootObject = keypathObserver.object;
+            id observer = keypathObserver.observer;
+            NSString* keypath = keypathObserver.keypath;
+            NSMutableDictionary* change = keypathObserver.pendingChange;
+            void* context = keypathObserver.context;
+            NSKeyValueChange changeKind = _changeFromSetMutationKind(mutationKind);
```
DHowett-MSFT: so much code duplication.

can you fold this into a common function?

DHowett-MSFT: like `_dispatchDidChange`?

Frameworks/Foundation/NSKVOSwizzling.mm
DHowett-MSFT: not even kidding.

you've taken ownership of this code, it is now your responsibility ;)

Frameworks/Foundation/NSKVOSwizzling.mm
```diff
@@ -253,21 +243,44 @@ static void notifyingVariadicSetImpl(id self, SEL _cmd, ...) {
...
     imp(self, _cmd, indexes, objects);
 
     [self didChange:NSKeyValueChangeReplacement valuesAtIndexes:indexes forKey:key];
 }
 
+template <NSKeyValueSetMutationKind Kind>
+static inline void _NSKVOSetDispatch(id self, SEL _cmd, NSSet* set) {
+    NSString* key = _keyForSelector(self, _cmd);
+
+    [self willChangeValueForKey:key withSetMutation:Kind usingObjects:set];
```
DHowett-MSFT: nit: `TKind`

Frameworks/Foundation/NSKVOSwizzling.mm
```diff
@@ -355,36 +367,45 @@ static SEL formatSelector(NSString* format, ...) {
...
-static void _NSKVOEnsureCollectionWillNotify(id object, NSString* key, const char* rawKey) {
-    bool found = false;
-    auto insertOneSel = formatSelector(@"insertObject:in%sAtIndex:", rawKey);
-
-    if ([object respondsToSelector:insertOneSel]) {
-        found = true;
+static inline void _NSKVOEnsureOrderedCollectionWillNotify(id object, NSString* key, const char* rawKey) {
+    SEL insertOne = formatSelector(@"insertObject:in%sAtIndex:", rawKey);
+    SEL insertMany = formatSelector(@"insert%s:atIndexes:", rawKey);
+    if ([object respondsToSelector:insertOne] || [object respondsToSelector:insertMany]) {
```
DHowett-MSFT: the point of the previous code was to _avoid even formatting_ a single selector that we wouldn't end up using

Frameworks/Foundation/NSKVOSupport.mm
DHowett-MSFT: this has the added effect of changing `changeKind` for all future observers; is that what we want?

I don't think it is; keyObserver is local to a single observer dispatch, but changeKind if global to all observer dispatches for this key!

aballway: This was there before ;D so I assumed it was because of something with our implementation of aggregates

Frameworks/Foundation/NSKVOSwizzling.mm
aballway: I thought the NSInvocation wasn't able to support this yet?  If it can I can just swap it out, but fixing NSInvocation is well out of the scope of this PR.

Frameworks/Foundation/NSKVOSupport.mm
aballway: This is clangformat, probably thinks... well idk what it thinks but I don't want to disable it for this one line #WontFix

Frameworks/Foundation/NSKVOSwizzling.mm
DHowett-MSFT: NSInvocation is the thing that supports an infinite amount of args; this no longer matches NSInvocation!~

Frameworks/Foundation/NSKVOSupport.mm
aballway: This was to be able to modify it, which is apparently not the correct thing so will fix

Frameworks/Foundation/NSKVOSupport.mm
aballway: ^ good point

Frameworks/Foundation/NSKVOSwizzling.mm
aballway: I don't want to imply it's a Type since it it's just a value #ByDesign

Frameworks/Foundation/NSKVOSwizzling.mm
aballway: This is doing the same amount of formatting if they're not being used?

Frameworks/Foundation/NSKVOSupport.mm
aballway: I could but I prefer this style

Frameworks/Foundation/NSKVOSupport.mm
aballway: Mmhmm, just need an unmolested copy to know which values were just added

Frameworks/Foundation/NSKVOSupport.mm
aballway: set whoops

Frameworks/Foundation/NSKVOSupport.mm
aballway: Looks like I can reduce duplication a little more, will do!

Frameworks/Foundation/NSKVOSwizzling.mm
aballway: I see.  Either way still outside the scope of this PR.

DHowett-MSFT: Since you're making changes here, please update these types to ensure register width; perhaps `uintptr_t`?

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -516,169 +515,292 @@ - (void)removeObserver:(id)observer forKeyPath:(NSString*)keyPath {
...
+            return NSKeyValueChangeRemoval;
+        default:
+            return NSKeyValueChangeReplacement;
     }
+}
 
-    if (!value) {
-        value = [NSNull null];
+static inline id _valueForPendingChangeAtIndexes(
+    id notifyingObject, NSString* key, id rootObject, NSString* keypath, _NSKVOKeyObserver* keyObserver, NSIndexSet* indexes) {
```
ms-jihua: keypath seems to not be used at all here?

aballway:
Ran KVOPerf, difference is negligible (~5μs per notification, about 4% more)

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -516,169 +515,292 @@ - (void)removeObserver:(id)observer forKeyPath:(NSString*)keyPath {
...
+
+        // Skip any keypaths that are in the process of changing.
+        _NSKVOKeypathObserver* keypathObserver = keyObserver.keypathObserver;
+        if ([keypathObserver popDidChange]) {
+            func(keyObserver, keypathValueCache);
+
             id observer = keypathObserver.observer;
             NSString* keypath = keypathObserver.keypath;
+            id rootObject = keypathObserver.object;
+            NSKeyValueObservingOptions options = keypathObserver.options;
```
ms-jihua: unused

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -516,169 +515,292 @@ - (void)removeObserver:(id)observer forKeyPath:(NSString*)keyPath {
...
 
-/**
- @Status Stub
-*/
-- (void)didChangeValueForKey:(NSString*)key withSetMutation:(NSKeyValueSetMutationKind)mutationKind usingObjects:(NSSet*)options {
-    UNIMPLEMENTED();
+            if ((options & NSKeyValueObservingOptionNew) && changeKind != NSKeyValueChangeRemoval) {
+                // New values only exist for inserting or replacing, not removing
+                NSKeyValueObservingOptions options = keypathObserver.options;
```
ms-jihua: unused, redeclaration

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -516,169 +515,292 @@ - (void)removeObserver:(id)observer forKeyPath:(NSString*)keyPath {
...
-                // This only needs to be done in willChange because didChange derives from the existing changeset.
-                change[NSKeyValueChangeKindKey] = @(changeKind = NSKeyValueChangeSetting);
-                [change removeObjectForKey:NSKeyValueChangeIndexesKey];
-            }
+/**
+@Status Interoperable
+*/
+- (void)willChangeValueForKey:(NSString*)key {
+    if ([self observationInfo]) {
+        NSMutableDictionary* change = [NSMutableDictionary dictionaryWithObject:@(NSKeyValueChangeSetting) forKey:NSKeyValueChangeKindKey];
```
ms-jihua: any reason that this isn't created inside the lambda? (do we want to minimize time inside the lambda or something?)

aballway: Lambda will be called in a loop, moved out to prevent recreation

DHowett-MSFT: You've introduced an observer information leak.

```objc
[object setValue:@"x" forKeyPath:@"key"]
...
[object addObserver:observer1 forKeyPath:@"key" options:/* prior, old, new */ context:0];
[object addObserver:observer2 forKeyPath:@"key" options:/* prior */ context:0];
...
[object setValue:@"y" forKeyPath:@"key"]
...
// observer1 will receive old="x"
// observer2 will receive old="x" as well, despite not requesting it.
```

DHowett-MSFT:
(this is why it was called `changeSeed`...)

DHowett-MSFT:
By doing this level of lambda recapture and deduplication, have you contributed meaningfully to the readability or understandability of the code? Or have you damaged it?

aballway:
?

---
In reply to: [301566522](https://github.com/Microsoft/WinObjC/pull/2693#issuecomment-301566522) [](ancestors = 301566522)

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -591,27 +602,20 @@ static void _dispatchDidChange(id notifyingObject, NSString* key, TFunc&& func)
...
+        _dispatchWillChange(self, key, [](_NSKVOKeyObserver* keyObserver) {
+            NSMutableDictionary* change =
+                [NSMutableDictionary dictionaryWithObject:@(NSKeyValueChangeSetting) forKey:NSKeyValueChangeKindKey];
             _NSKVOKeypathObserver* keypathObserver = keyObserver.keypathObserver;
             NSKeyValueObservingOptions options = keypathObserver.options;
-            id rootObject = keypathObserver.object;
-            NSString* keypath = keypathObserver.keypath;
 
             if (options & NSKeyValueObservingOptionOld) {
                 // For to-many mutations, we can't get the old values at indexes that have not yet been inserted.
```
DHowett-MSFT: outdated comment :)

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -647,15 +650,13 @@ - (void)didChangeValueForKey:(NSString*)key {
...
 - (void)willChange:(NSKeyValueChange)changeKind valuesAtIndexes:(NSIndexSet*)indexes forKey:(NSString*)key {
     if ([self observationInfo]) {
-        NSMutableDictionary* change = [NSMutableDictionary
-            dictionaryWithObjectsAndKeys:@(changeKind), NSKeyValueChangeKindKey, indexes, NSKeyValueChangeIndexesKey, nil];
-
-        _dispatchWillChange(self, key, [ changeKind, indexes, key, notifyingObject = self, change ](_NSKVOKeyObserver * keyObserver) {
+        _dispatchWillChange(self, key, [ changeKind, indexes, key, notifyingObject = self ](_NSKVOKeyObserver * keyObserver) {
+            NSMutableDictionary* change = [NSMutableDictionary
+                dictionaryWithObjectsAndKeys:@(changeKind), NSKeyValueChangeKindKey, indexes, NSKeyValueChangeIndexesKey, nil];
             NSKeyValueChange localChangeKind = changeKind;
```
DHowett-MSFT: since we're not mutating this, do we need to store it locally instead of in the capture?

DHowett-MSFT: er sorry, since we're not using this after we mutate it

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -667,16 +668,7 @@ - (void)willChange:(NSKeyValueChange)changeKind valuesAtIndexes:(NSIndexSet*)ind
...
-                    _valueForPendingChangeAtIndexes(notifyingObject, key, rootObject, keypath, keyObserver, indexes);
-            }
-
-            if (options & NSKeyValueObservingOptionPrior) {
-                id observer = keypathObserver.observer;
-                void* context = keypathObserver.context;
-                [change setObject:@(YES) forKey:NSKeyValueChangeNotificationIsPriorKey];
-                [observer observeValueForKeyPath:keypath ofObject:rootObject change:change context:context];
-                [change removeObjectForKey:NSKeyValueChangeNotificationIsPriorKey];
+                change[NSKeyValueChangeOldKey] = _valueForPendingChangeAtIndexes(notifyingObject, key, rootObject, keyObserver, indexes);
```
DHowett-MSFT: oh, i understand what was happening. do you know if this holds up with the comment on 662? We're degrading a collection mutation event to a single setter mutation event. `valueForPendingChange...` was supposed to _explicitly not_ do `objectsAtIndexes:` in that case.

It was intended to fall back to the `!indexes` case (original code, line 524)

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -695,15 +687,14 @@ - (void)didChange:(NSKeyValueChange)changeKind valuesAtIndexes:(NSIndexSet*)inde
...
             NSMutableDictionary* change = keypathObserver.pendingChange;
-
             if ((options & NSKeyValueObservingOptionNew) && [change[NSKeyValueChangeKindKey] integerValue] != NSKeyValueChangeRemoval) {
                 // For to-many mutations, we can't get the new values at indexes that have been deleted.
+                id rootObject = keypathObserver.object;
+                NSString* keypath = keypathObserver.keypath;
                 id newValue = keypathValueCache[keypath];
                 if (!newValue) {
-                    newValue = _valueForPendingChangeAtIndexes(notifyingObject, key, rootObject, keypath, keyObserver, indexes);
+                    newValue = _valueForPendingChangeAtIndexes(notifyingObject, key, rootObject, keyObserver, indexes);
```
DHowett-MSFT: I believe this is subject to the same caveats as above.

I swear we had tests testing this.

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -722,8 +713,8 @@ - (void)didChange:(NSKeyValueChange)changeKind valuesAtIndexes:(NSIndexSet*)inde
 - (void)willChangeValueForKey:(NSString*)key withSetMutation:(NSKeyValueSetMutationKind)mutationKind usingObjects:(NSSet*)objects {
     if ([self observationInfo]) {
         NSKeyValueChange changeKind = _changeFromSetMutationKind(mutationKind);
-        NSMutableDictionary* change = [NSMutableDictionary dictionaryWithObject:@(changeKind) forKey:NSKeyValueChangeKindKey];
-        _dispatchWillChange(self, key, [mutationKind, objects, change, changeKind](_NSKVOKeyObserver* keyObserver) {
+        _dispatchWillChange(self, key, [mutationKind, objects, changeKind](_NSKVOKeyObserver* keyObserver) {
```
DHowett-MSFT: it will be worth testing this to see what happens for the `key.collection.key` case -- I think you're supposed to get a `set` notification since `.key` is a calculated subcollection.

Frameworks/Foundation/NSKVOSupport.mm
aballway: compiler won't let me modify a capture value :(

bviglietta: You can mark a lambda as mutable.

`[x] () mutable { x = 3; }`

---
In reply to: [116584355](https://github.com/Microsoft/WinObjC/pull/2693#discussion_r116584355) [](ancestors = 116584355)

Frameworks/Foundation/NSKVOSupport.mm
DHowett-MSFT: Incidentally, this is why that function pulled `Indexes` out of the change dictionary -- it was the ideal way to communicate that we had degraded a collection mutation to a single setter mutation across the `will`->`did` divice.

aballway: I see, since that was used by regular `willChangeValueForKey` I thought it was just being used as a flag

Frameworks/Foundation/NSKVOSupport.mm
```diff
@@ -650,25 +656,26 @@ - (void)didChangeValueForKey:(NSString*)key {
...
             id rootObject = keypathObserver.object;
 
             // The reference platform does not support to-many mutations on nested keypaths.
             // We have to treat them as to-one mutations to support aggregate functions.
-            if (localChangeKind != NSKeyValueChangeSetting && keyObserver.restOfKeypathObserver) {
+            if (changeKind != NSKeyValueChangeSetting && keyObserver.restOfKeypathObserver) {
                 // This only needs to be done in willChange because didChange derives from the existing changeset.
-                change[NSKeyValueChangeKindKey] = @(localChangeKind = NSKeyValueChangeSetting);
+                change[NSKeyValueChangeKindKey] = @(changeKind = NSKeyValueChangeSetting);
                 [change removeObjectForKey:NSKeyValueChangeIndexesKey];
```
DHowett-MSFT: it will be worth documenting what removing this does to the later function call

aballway:
That fails during `addObserver:` because one can't KVO an NSSet*

aballway:
Re-ran KVOPerf, difference is still negligible.

DHowett-MSFT:
I'd find some bees and get you their knees if you'd write a test for the to-many to to-one array notification downgrade case.

Frameworks/Foundation/NSKVOSupport.mm
DHowett-MSFT: still says `keybserver`

aballway:
ping @msft-Jeyaram

PR #2697, opened 2017-05-12
Fix delegate dispatch order/threading in NSURLProtocol

### LINKED ISSUE #2521: NSURLProtocol and NSURLSession do not guarantee delegate dispatch order and thread

ms-jihua:
- Client callbacks now occur on client thread
- Client callbacks now have guaranteed order
- NSURLConnection/Session, which are built on top of Protocol, now also have guaranteed ordering/correct callback context
- Re-enabled NSURLSession tests
- Removed schedule/unscheduleInRunLoop: from NSURLProtocol - this is not a part of the contract and is unused even internally

Fixes #2521

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2697)

### DIFF
--- a/Frameworks/Foundation/NSURLConnection.mm
+++ b/Frameworks/Foundation/NSURLConnection.mm
@@ -126,18 +126,9 @@ static void __dispatchDelegateCallback(NSURLConnection* connection, void (^callb
     } else if (connection->_scheduledRunLoops.size() == 1) {
         // TODO #2469: Cannot currently schedule on more than one runloop
         for (auto pair : connection->_scheduledRunLoops) {
-            [pair.first performSelector:@selector(_invokeBlock:)
-                                 target:connection
-                               argument:callbackBlock
-                                  order:0
-                                  modes:@[ pair.second.get() ]];
+            [pair.first performSelector:@selector(invoke) target:callbackBlock argument:nil order:0 modes:@[ pair.second.get() ]];
         }
     }
-}
-
-// Helper function that just invokes a block
-- (void)_invokeBlock:(void (^)())block {
-    block();
 }
 
 // Body of the thread on which the NSURLProtocol loads
--- a/Frameworks/Foundation/NSURLConnection.mm
+++ b/Frameworks/Foundation/NSURLConnection.mm
@@ -382,19 +382,17 @@ static void __dispatchDelegateCallback(NSURLConnection* connection, void (^callb
  @Status Interoperable
 */
 - (void)URLProtocol:(NSURLProtocol*)urlProtocol didFailWithError:(NSError*)error {
-    __block StrongId<NSURLConnection> strongSelf = self;
-    __block StrongId<NSError> strongError = error;
     __dispatchDelegateCallback(self, ^void() {
         TraceError(TAG, L"URL protocol did fail");
         // if ( [_delegate respondsToSelector:@selector(connection:willSendRequest:redirectResponse:)] ) [_delegate
         // connection:self willSendRequest:_request redirectResponse:nil];
-        if ([strongSelf->_delegate respondsToSelector:@selector(connection:didFailWithError:)]) {
-            [strongSelf->_delegate connection:strongSelf didFailWithError:strongError];
+        if ([_delegate respondsToSelector:@selector(connection:didFailWithError:)]) {
+            [_delegate connection:self didFailWithError:error];
         }
 
-        strongSelf->_protocol = nil;
-        strongSelf->_delegate = nil;
-        strongSelf->_done = true;
+        _protocol = nil;
+        _delegate = nil;
+        _done = true;
     });
 }
 
@@ -404,8 +402,6 @@ static void __dispatchDelegateCallback(NSURLConnection* connection, void (^callb
 - (void)URLProtocol:(NSURLProtocol*)urlProtocol
     didReceiveResponse:(NSURLResponse*)response
     cacheStoragePolicy:(NSURLCacheStoragePolicy)policy {
-    __block StrongId<NSURLConnection> strongSelf = self;
-    __block StrongId<NSURLResponse> strongResponse = response;
     __dispatchDelegateCallback(self, ^void() {
         /*
         if ( [response respondsToSelector:@selector(statusCode)] && [response statusCode] != 200 ) {
@@ -413,15 +409,15 @@ static void __dispatchDelegateCallback(NSURLConnection* connection, void (^callb
         }
         */
 
-        strongSelf->_response = response;
-        strongSelf->_storagePolicy = policy;
+        _response = response;
+        _storagePolicy = policy;
 
-        if ([strongSelf->_delegate respondsToSelector:@selector(connection:willCacheResponse:)]) {
+        if ([_delegate respondsToSelector:@selector(connection:willCacheResponse:)]) {
             // TODO #2469: This is currently incorrect, these are not the same thing
-            [strongSelf->_delegate connection:strongSelf willCacheResponse:static_cast<NSCachedURLResponse*>(response)];
+            [_delegate connection:self willCacheResponse:static_cast<NSCachedURLResponse*>(response)];
         }
-        if ([strongSelf->_delegate respondsToSelector:@selector(connection:didReceiveResponse:)]) {
-            [strongSelf->_delegate connection:strongSelf didReceiveResponse:strongResponse];
+        if ([_delegate respondsToSelector:@selector(connection:didReceiveResponse:)]) {
+            [_delegate connection:self didReceiveResponse:response];
         }
     });
 }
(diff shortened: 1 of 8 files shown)


Frameworks/Foundation/NSURLConnection.mm
```diff
@@ -346,118 +373,152 @@ - (void)cancel {
...
  @Status Interoperable
 */
-- (void)URLProtocol:(id)urlProtocol didReceiveAuthenticationChallenge:(id)challenge {
-    if ([_delegate respondsToSelector:@selector(connection:willSendRequestForAuthenticationChallenge:)]) {
-        [_delegate connection:self willSendRequestForAuthenticationChallenge:challenge];
-    } else {
-        [_delegate connection:self didReceiveAuthenticationChallenge:challenge];
-    }
+- (void)URLProtocol:(NSURLProtocol*)urlProtocol didReceiveAuthenticationChallenge:(NSURLAuthenticationChallenge*)challenge {
+    __block StrongId<NSURLConnection> strongSelf = self;
```
DHowett-MSFT: I believe using __block on these _avoids_ copying the `StrongId` -- self is therefore _not_ strongly held by the block.

Frameworks/Foundation/NSURLProtocol_WinHTTP.mm
```diff
@@ -243,20 +244,35 @@ - (instancetype)initWithRequest:(NSURLRequest*)request
...
+    block();
+}
+
 - (void)startLoading {
     @synchronized(self) {
         if (_httpRequestOperation) {
             return;
         }
 
+        _clientThread = [NSThread currentThread];
```
DHowett-MSFT: Is it possible for the thread to be dismantled before we finish here?

Frameworks/Foundation/NSURLProtocol_WinHTTP.mm
```diff
@@ -412,19 +428,24 @@ - (HRESULT)_asyncHttpOperation:(AsyncHttpOperation*)pOperation completedWithStat
 
                 // The location header can specify a relative or absolute URL: attempt to join it with the request URL.
                 NSURL* targetUrl = [NSURL URLWithString:Strings::WideToNSString(uriString.Get()) relativeToURL:[request URL]];
-                NSURLRequest* req = [NSURLRequest requestWithURL:targetUrl];
-                [self.client URLProtocol:self wasRedirectedToRequest:req redirectResponse:nsResponse];
+                __block StrongId<NSURLRequest> req = [NSURLRequest requestWithURL:targetUrl];
```
DHowett-MSFT: It's safe to just capture things in blocks normally; Strong semantics are preserved since blocks referring to `id` types will retain/release.

DHowett-MSFT: (it's clearer, in fact, and avoids the StrongId machinery)

tests/functionaltests/Tests/NSURLSession.mm
```diff
@@ -281,7 +281,6 @@ - (void)URLSession:(NSURLSession*)session downloadTask:(NSURLSessionDownloadTask
 public:
     BEGIN_TEST_CLASS(NSURLSessionTests)
     TEST_CLASS_PROPERTY(L"UAP:AppXManifest", L"NSURL.AppxManifest.xml")
-    TEST_CLASS_PROPERTY(L"ignore", L"true")
```
DHowett-MSFT: there ma ybe other ignores on these tests; would you mind checking?

Frameworks/Foundation/NSURLProtocol_WinHTTP.mm
ms-jihua: i completely failed to think this through oops

tests/functionaltests/Tests/NSURLSession.mm
ms-jihua: just the arm ones

Frameworks/Foundation/NSURLProtocol_WinHTTP.mm
ms-jihua: er, i'm assuming by 'before we finish here' you mean the whole set of callbacks and not just this call. pretty sure it's possible, though not in our internal use-cases. with our current impl of performSelector:, this raises an exception. I think this is acceptable since letting the client thread die seems like programmer error.

Frameworks/Foundation/NSURLProtocol_WinHTTP.mm
DHowett-MSFT: this file still has `__block StrongId`s!

Frameworks/Foundation/NSURLProtocol_WinHTTP.mm
DHowett-MSFT: makes perfect sense. thanks!

Frameworks/Foundation/NSURLProtocol_WinHTTP.mm
```diff
@@ -243,20 +244,35 @@ - (instancetype)initWithRequest:(NSURLRequest*)request
     };
 }
 
+// Helper function that executes a block on the client thread - used for client callbacks
+static void __dispatchClientCallback(NSURLProtocol_WinHTTP* protocol, void (^callbackBlock)()) {
+    if (protocol->_clientThread) {
+        [protocol performSelector:@selector(_invokeBlock:) onThread:protocol->_clientThread withObject:callbackBlock waitUntilDone:NO];
```
DHowett-MSFT: it's valid/specified for block objects to have an `invoke` method, so you could use `callbackBlock performSelector:@selector(invoke) ...`

DHowett-MSFT: you know, I take that back. It's valid on the reference platform and on winobjc, but it isn't _specified in the [Language Specification for Blocks](https://clang.llvm.org/docs/BlockLanguageSpec.html)_

ms-jihua: 1) don't our NSBlocks not inherit from NSObject and thus not have that function?
2) our NSBlock has a comment on it that we ought to be reducing internal usage of the function. I think that's silly, but...

ms-jihua: mhm, it's considered hidden functionality

DHowett-MSFT: 1. They do inherit NSObject! It's awesome/terrifying all at the same time.
2. Meh, I left that comment a long time ago because I didn't think `-invoke` was valid.

I did just test it, though:

```
(duhowett-mac) ~/woc/block % cat main.m
#include <stdio.h>

int main() {
        void (^x)() = ^{ fprintf(stderr, "lol?\n"); };
        [x invoke];
        return 0;
}
(duhowett-mac) ~/woc/block % ./main
lol?
```

DHowett-MSFT: so if you want to do it, be my guest!

Frameworks/Foundation/NSURLProtocol_WinHTTP.mm
```diff
@@ -395,10 +411,10 @@ - (HRESULT)_asyncHttpOperation:(AsyncHttpOperation*)pOperation completedWithStat
                 default:
                     httpVersionStr = nil;
             }
-            NSHTTPURLResponse* nsResponse = [[[NSHTTPURLResponse alloc] initWithURL:[request URL]
-                                                                         statusCode:statusCode
-                                                                        HTTPVersion:httpVersionStr
-                                                                       headerFields:nsHeaders] autorelease];
+            __block StrongId<NSHTTPURLResponse> nsResponse = [[[NSHTTPURLResponse alloc] initWithURL:[request URL]
```
DHowett-MSFT: I still don't think this one's necessary either

Frameworks/Foundation/NSURLProtocol_WinHTTP.mm
DHowett-MSFT: (or this one, since `req` will be captured)

Frameworks/Foundation/NSURLProtocol_WinHTTP.mm
```diff
@@ -495,10 +515,15 @@ - (void)_consumeDataStreamForIHttpContent:(IHttpContent*)pHttpContent {
             // We're opting to take a copy of the buffer here, as NSURLProtocol's consumer can legally expect the data
             // to remain valid long after the connection is gone. Eventually, it would be very nice to have an IBuffer-backed
             // NSData.
-            [self.client URLProtocol:strongSelf didLoadData:[NSData dataWithBytes:pOutputBuffer length:length]];
+            __block StrongId<NSData> data = [NSData dataWithBytes:pOutputBuffer length:length];
```
DHowett-MSFT: another one here :)

Issue #2607, opened 2017-04-24
Dependency Properties should be attached directly to custom XAML controls


nathpete-msft:
Currently we must attach Dependency Properties for our UIKit XAML controls to `FrameworkElement` due to type information for custom XAML controls not being exposed to the testing framework.

An unfortunate side effect of this is that `FrameworkElement` will be polluted with our Dependency Properties, and potential conflicts could arise due to duplicate property names from multiple controls.
As a temporary workaround to prevent naming conflicts with properties on `FrameworkElement`, each property will be prefixed with the owning control name: `ControlName_PropertyName`. For example, a `DependencyProperty` for the `IsActive` property on `ProgressRing` would be: `ProgressRing_IsActive`.

Once it is possible to expose the type information of our custom XAML controls to the testing framework, these Dependency Properties should be cleaned up to attach to controls directly and remove the naming prefixes.

bbowman:
@nathpete-msft what do you mean by: 

> Once it is possible to expose the type information of our custom XAML controls to the testing framework

Perhaps we can chat about what we need in testing?

nathpete-msft:
@jaredhms looked into this issue before, and I think he might have better understanding on what needs to be done to resolve this.

bbowman:
@nathpete-msft if you are owning this task / bug, I'd like you to figure that out. I did most of the investigation with @jaredms and can answer questions.

jaredhms:
@bbowman  - we need TAEF's appx test execution.

PR #2706, opened 2017-05-16
This change Implement UITextField tests following new functional test paradigm

yiyang-msft:
1. Implement UITextField functional tests using new functional test paradigm
2. Fixed a couple of issues found in the process
   1.) Background color was not binding correctly earlier
   2.) adding a const padding 16.0 represents textbox and its container control, 16 is the actual value through experiements.
   3.)In order for test to know about the dependency properties (#2607), I have to register the DP on frameworkElement, instead of TextField, similar to others. To void potential conflict, also rename it from Text to TextField_Text.
   4.)Per request, make UISlider tooltips disabled by default
   5.) Keep legacy UIText Field Test because it has some test for interactions which is still valuable.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2706)

### DIFF
--- a/Frameworks/UIKit.Xaml/TextField.xaml.cpp
+++ b/Frameworks/UIKit.Xaml/TextField.xaml.cpp
@@ -118,7 +118,6 @@ void TextField::_SwitchToMode(bool secure) {
         // switch to Non-secure mode, create a textbox as backingControl
         _backingControl = ref new Controls::TextBox();
     }
-
     _backingControl->Loaded += ref new Windows::UI::Xaml::RoutedEventHandler(this, &UIKit::Xaml::TextField::OnBackingControlLoaded); 
     _gotFocusHandlerRegistrationToken = _backingControl->GotFocus += ref new Windows::UI::Xaml::RoutedEventHandler(this, &TextField::OnGotFocus);
     _lostFocusHandlerRegistrationToken = _backingControl->LostFocus += ref new Windows::UI::Xaml::RoutedEventHandler(this, &UIKit::Xaml::TextField::OnLostFocus);
@@ -167,9 +166,16 @@ void TextField::OnBackingControlLoaded(Platform::Object ^sender, Windows::UI::Xa
         Windows::UI::Xaml::Data::Binding^ binding = ref new Windows::UI::Xaml::Data::Binding();
         binding->Mode = Windows::UI::Xaml::Data::BindingMode::TwoWay;
         binding->ElementName = "Root";
-        binding->Path = ref new Windows::UI::Xaml::PropertyPath("Text");
+        binding->Path = ref new Windows::UI::Xaml::PropertyPath("TextField_Text");
         binding->UpdateSourceTrigger = Windows::UI::Xaml::Data::UpdateSourceTrigger::PropertyChanged;
         passwordBox->SetBinding(Controls::PasswordBox::PasswordProperty, binding);
+
+        // one way binding background to inner passwordbox background
+        binding = ref new Windows::UI::Xaml::Data::Binding();
+        binding->Mode = Windows::UI::Xaml::Data::BindingMode::OneWay;
+        binding->ElementName = "Root";
+        binding->Path = ref new Windows::UI::Xaml::PropertyPath("Background");
+        passwordBox->SetBinding(Controls::PasswordBox::BackgroundProperty, binding);
 
         passwordBox->PlaceholderText = _placeholder;
         passwordBox->InputScope = _FindBestFitInputScope();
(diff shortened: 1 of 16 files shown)


nathpete-msft:
Review status: 0 of 16 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[Frameworks/UIKit.Xaml/TextField.xaml.cpp, line 428 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/2706#-KkHnjsn2ssKseR2RhhZ:-KkHnjsn2ssKseR2Rhh_:bb752ax) ([raw file](https://github.com/microsoft/winobjc/blob/64657e5df19fd875b02c8a2323571c847d37f384/Frameworks/UIKit.Xaml/TextField.xaml.cpp#L428)):*
> ```cpp
> }
> 
> void TextField::_SetBackgournd() {
> ```

Typo

---

*[Frameworks/UIKit.Xaml/TextField.xaml.h, line 246 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/2706#-KkHo9otMDRFGpR-ixFk:-KkHo9otMDRFGpR-ixFl:bb752ax) ([raw file](https://github.com/microsoft/winobjc/blob/64657e5df19fd875b02c8a2323571c847d37f384/Frameworks/UIKit.Xaml/TextField.xaml.h#L246)):*
> ```objc
>     void _SetEnabled();
>     void _SetForeground();
>     void _SetBackgournd();
> ```

Typo

---

*[tests/functionaltests/UXTestHelpers.mm, line 230 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/2706#-KkHqFPU1lGdHZp1Mmvq:-KkHqFPU1lGdHZp1Mmvr:b-6p43pj) ([raw file](https://github.com/microsoft/winobjc/blob/64657e5df19fd875b02c8a2323571c847d37f384/tests/functionaltests/UXTestHelpers.mm#L230)):*
> ```XML
> 
>     CGFloat r1, g1, b1, a1;
>     [col1 getRed:&r1 green:&g1 blue:&b1 alpha:&a1];
> ```

I don't think this works if supplied a UIColor that isn't in the RBG color space (ie. [UIColor grayColor]).

I handled this by extracting the CG color components and mapping them based on color space, here: https://github.com/Microsoft/WinObjC/pull/2608/files#diff-1ffe8a84e54a7ef5d70e87484bf501a7R163

---

*[tests/functionaltests/Tests/UIKitTests/UILabelTests.mm, line 471 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/2706#-KkHsGDEGJXtk796cOLm:-KkHsGDEGJXtk796cOLn:b52dgm0) ([raw file](https://github.com/microsoft/winobjc/blob/64657e5df19fd875b02c8a2323571c847d37f384/tests/functionaltests/Tests/UIKitTests/UILabelTests.mm#L471)):*
> ```XML
>             auto colorBrush = textBlock.Foreground().as<Media::SolidColorBrush>();
>             ASSERT_TRUE(colorBrush);
>             EXPECT_TRUE(UXTestAPI::compareRGBAValues(UXTestAPI::ConvertWUColorToUIColor(colorBrush.Color()), label.textColor));
> ```

Any reason not to use the existing `UXTestAPI::IsRGBAEqual` for all of these checks? It already takes a `SolidColorBrush` and a `UIColor` as parameters to compare for you.

---

*[tests/functionaltests/Tests/UIKitTests/UITextFieldTests.mm, line 145 at r1](https://reviewable.io:443/reviews/microsoft/winobjc/2706#-KkHrL1Aepumn7LOB6KP:-KkHrL1Aepumn7LOB6KQ:bb752ax) ([raw file](https://github.com/microsoft/winobjc/blob/64657e5df19fd875b02c8a2323571c847d37f384/tests/functionaltests/Tests/UIKitTests/UITextFieldTests.mm#L145)):*
> ```XML
>             EXPECT_TRUE(textBox);
> 
>             textField.placeholder = @"placerholder";
> ```

Typo

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/winobjc/2706)*

DHowett-MSFT:
@nathpete-msft I believe that `getRed:green:blue:alpha:` will do the colorspace conversion if necessary.

nathpete-msft:
@DHowett-MSFT Yeah looks like you're right. I just tried it with `grayColor` and `whiteColor` and it does. I wasn't sure, since the docs say that it only handles conversion if it's in a compatible color space. I'm not sure what an incompatible color space would be as far as this method is concerned, or if this comment in the docs just happens to be outdated.

nathpete-msft:
:shipit:

nathpete-msft:
Reviewed 5 of 16 files at r1, 11 of 11 files at r2.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/winobjc/2706)*

Issue #561, opened 2016-06-15
NSUserDefaults synchronize blocks main thread


triplef:
Synchronizing NSUserDefaults is always [done on the main thread](https://github.com/Microsoft/WinObjC/blob/master/Frameworks/Foundation/NSUserDefaults.mm#L379-L384), and can easily take several hundred milliseconds or seconds depending of the size of the user defaults. This can cause the app to noticeably hang after setting a user defaults value.

Following is an example synchronizing 500 user defaults entries, which takes roughly 1 second on our development machines.

```objc
for (int i = 0; i < 500; i++) {
	[[NSUserDefaults standardUserDefaults] setValue:@(i) forKey:[NSString stringWithFormat:@"Test%d", i]];
}

NSTimeInterval start = [NSDate timeIntervalSinceReferenceDate];
[[NSUserDefaults standardUserDefaults] synchronize];
NSTimeInterval end = [NSDate timeIntervalSinceReferenceDate];

NSLog(@"Synchronize took %lf s", end-start);
```

jofre-ms:
Reworked this to use asynchronous file writes, which reduced synchronize time for your example on my machine from approximately 0.45 seconds to 0.04 seconds.

PR #2707, opened 2017-05-16
Change [NSUserDefaults synchronize] to actually be synchronous

### LINKED ISSUE #2705: NSUserDefaults does not sync
If you save data to NSUserDefaults at app launch, NSUserDefaults sync and seems to write data (if I write to log straight after sync, I can see the data). But if I re-open app, NSUserDefaults do not contain that data anymore. On the third launch, NSUserdefaults synced data properly.

You can check this by the following code:

```
- (void)viewDidLoad {
    [super viewDidLoad];
    // Do any additional setup after loading the view, typically from a nib.
    
   UIFont* font = [UIFont fontWithName : @"HelveticaNeue" size : 11.0];
   NSNumber *shouldShowButton = [self userDefaultsExample];

   if (shouldShowButton) {
         UIButton *callbackButton = [UIButton buttonWithType : UIButtonTypeRoundedRect];
         [callbackButton setTitle : @"Watch tutorial" forState : UIControlStateNormal];
         [self.view addSubview : callbackButton];
         callbackButton.frame = CGRectMake(0, 400, 100, 35);
         callbackButton.titleLabel.font = font;
    }
}

- (NSNumber *) userDefaultsExample {

    NSUserDefaults *userDefaults = [NSUserDefaults standardUserDefaults];
    NSMutableDictionary *exampleInfo;
    
    if (![userDefaults objectForKey : @"ExampleInfo"]) {
         exampleInfo = [NSMutableDictionary dictionary];
         [userDefaults setObject : exampleInfo forKey : @"ExampleInfo"];
    }
    else {
         exampleInfo = [NSMutableDictionary dictionaryWithDictionary : [userDefaults objectForKey : @"ExampleInfo"]];
    }

    NSString *exampleKey = @"ExampleKey";
    BOOL value = [exampleInfo[exampleKey] boolValue];
    if (!value) {
         // look for information in user defaults root
         value = [userDefaults boolForKey : exampleKey];
         // cleanup
         [userDefaults removeObjectForKey : exampleKey];
    }

    // save info with updated value
    exampleInfo[exampleKey] = @(YES);
    [userDefaults setObject : exampleInfo forKey : @"ExampleInfo"];
    [userDefaults synchronize];
    
    NSLog(@"ExampleInfo: %@",[userDefaults objectForKey : @"ExampleInfo"]);
    return (value) ? nil : @(YES);
}

```

The button will be shown at the 1st and 2nd launch of the app, it won't be shown at the third launch. Expected: the button should be shown only at 1st launch of the app.

Issue takes place on Windows 10 OS Build 15063 and later versions, issue does not happen on OS Build 14393

ms-jihua:
Fixes #2705

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2707)

### DIFF
--- a/Frameworks/CoreFoundation/Preferences.subproj/CFPreferences.c
+++ b/Frameworks/CoreFoundation/Preferences.subproj/CFPreferences.c
@@ -190,10 +190,10 @@ if (completePath) {
     CFStringAppend(completePath, CFSTR("Preferences\\"));
     url = CFURLCreateWithFileSystemPath(alloc, completePath, kCFURLWindowsPathStyle, true);
 
-	char cPath[CFMaxPathSize];
-	if (CFURLGetFileSystemRepresentation(url, true, (unsigned char *)cPath, CFMaxPathSize)) {
-		success = _CFCreateDirectory(cPath);
-	}
+    char cPath[CFMaxPathSize];
+    if (CFURLGetFileSystemRepresentation(url, true, (unsigned char *)cPath, CFMaxPathSize)) {
+        success = _CFCreateDirectory(cPath);
+    }
 
     CFRelease(completePath);
 }
@@ -882,11 +882,7 @@ static Boolean synchronizeApplicationDomain(CFTypeRef context, void *domain) {
     CFDataRef data = CFPropertyListCreateData(__CFPreferencesAllocator(), (CFPropertyListRef)domain, kCFPropertyListXMLFormat_v1_0, 0, NULL);
     Boolean success = false;
     if (data) {
-#if DEPLOYMENT_TARGET_WINDOWS
-        success = _CFWriteBytesToFileAsync((CFURLRef)context, CFDataGetBytePtr(data), CFDataGetLength(data));
-#else
         success = _CFWriteBytesToFile((CFURLRef)context, CFDataGetBytePtr(data), CFDataGetLength(data));
-#endif
         CFRelease(data);
     }
 
--- a/Frameworks/CoreFoundation/Preferences.subproj/CFPreferences.c
+++ b/Frameworks/CoreFoundation/Preferences.subproj/CFPreferences.c
@@ -190,10 +190,10 @@ if (completePath) {
     CFStringAppend(completePath, CFSTR("Preferences\\"));
     url = CFURLCreateWithFileSystemPath(alloc, completePath, kCFURLWindowsPathStyle, true);
 
-    char cPath[CFMaxPathSize];
-    if (CFURLGetFileSystemRepresentation(url, true, (unsigned char *)cPath, CFMaxPathSize)) {
-        success = _CFCreateDirectory(cPath);
-    }
+	char cPath[CFMaxPathSize];
+	if (CFURLGetFileSystemRepresentation(url, true, (unsigned char *)cPath, CFMaxPathSize)) {
+		success = _CFCreateDirectory(cPath);
+	}
 
     CFRelease(completePath);
 }


Frameworks/CoreFoundation/Preferences.subproj/CFPreferences.c
```diff
@@ -882,11 +882,7 @@ static Boolean synchronizeApplicationDomain(CFTypeRef context, void *domain) {
     CFDataRef data = CFPropertyListCreateData(__CFPreferencesAllocator(), (CFPropertyListRef)domain, kCFPropertyListXMLFormat_v1_0, 0, NULL);
     Boolean success = false;
     if (data) {
-#if DEPLOYMENT_TARGET_WINDOWS
-        success = _CFWriteBytesToFileAsync((CFURLRef)context, CFDataGetBytePtr(data), CFDataGetLength(data));
```
rajsesh-msft: Please, no! We deliberately added async because we cannot block the UI thread.  What we may want to do though is to wait for the IO operation to complete in case the app is exiting.

Frameworks/CoreFoundation/Preferences.subproj/CFPreferences.c
```diff
@@ -190,10 +190,10 @@ if (completePath) {
     CFStringAppend(completePath, CFSTR("Preferences\\"));
     url = CFURLCreateWithFileSystemPath(alloc, completePath, kCFURLWindowsPathStyle, true);
 
-	char cPath[CFMaxPathSize];
-	if (CFURLGetFileSystemRepresentation(url, true, (unsigned char *)cPath, CFMaxPathSize)) {
-		success = _CFCreateDirectory(cPath);
-	}
+    char cPath[CFMaxPathSize];
```
DHowett-MSFT: Not 100% sure we should change this indentation unless it's in code WinObjC added.

ms-jihua:
This function was originally made async due to #561. At the time, [NSUserDefaults synchronize] had an erroneous requirement that it be called from the main thread. Since this is no longer the case, switching the function back to synchronous to guarantee write completion.

Frameworks/CoreFoundation/Preferences.subproj/CFPreferences.c
```diff
@@ -882,11 +882,7 @@ static Boolean synchronizeApplicationDomain(CFTypeRef context, void *domain) {
     CFDataRef data = CFPropertyListCreateData(__CFPreferencesAllocator(), (CFPropertyListRef)domain, kCFPropertyListXMLFormat_v1_0, 0, NULL);
     Boolean success = false;
     if (data) {
-#if DEPLOYMENT_TARGET_WINDOWS
```
DHowett-MSFT: should we document this removal?

Frameworks/CoreFoundation/Preferences.subproj/CFPreferences.c
ms-jihua: I could not repro the reported issue when exiting out of the app normally (through the x button), and could only repro through debugger-stopping the app. This would not help the latter case. Additionally, judging by https://medium.com/@hanru.yeh/nsuserdefaults-is-planned-to-deprecated-cc185e19e6f8, it seems that [NSUserDefaults synchronize] is blocking on the reference platform.

Frameworks/CoreFoundation/Preferences.subproj/CFPreferences.c
ms-jihua: seems unnecessary since this is a WinObjC function

Frameworks/CoreFoundation/Preferences.subproj/CFPreferences.c
ms-jihua: not sure how this happened given this file is // clang-format off ;;;;

Frameworks/CoreFoundation/Preferences.subproj/CFPreferences.c
rajsesh-msft: If we do this, we would still regress #561, perhaps there is some perf improvement to be gained there?

ms-jihua: That seems pretty unlikely? I would guess that the majority of time spent is in the actual file write. If a user finds that time to be too long, they can always offload to a background thread themselves, now that synchronize no longer has a requirement that it be called on the main thread.

Issue #2704, opened 2017-05-15
UIWebView: Access violation exception when completion handler is nil for evaluatejavascript 


jboich:
Invoking method evaluatejavascript with completionHandler nil causes access violation exception, that leads to app crash. Please, fix this. 

Kind regards,
Julia

Issue #2695, opened 2017-05-12
UIWebView Release 47 Regression


edvv:
For release 0.2.170510:

`- [UIWebView loadHTMLString:baseURL:]`

throws an exception if the delegate is set (is non-nil).

That did not happen in previous releases.

rajsesh-msft:
@edvv I am unable to repro this.  I changed woccatalog page with webview to use loadHTMLString: and it shows the string fine.  Can you provide a short repro?

jboich:
Hi, @rajsesh-msft . I also have Access violation exception when try to use 
`[UIWebView loadHTMLString:baseURL:]` 

The issue takes place if you try to load page from a file:

```
NSString* htmlPath = [[NSBundle mainBundle] pathForResource:@"ExampleApp" ofType:@"html"];
NSString* appHtml = [NSString stringWithContentsOfFile:htmlPath encoding:NSUTF8StringEncoding error:nil];
NSURL *baseURL = [NSURL fileURLWithPath:htmlPath];
 [webView loadHTMLString:appHtml baseURL:baseURL];
```

edvv:
@rajsesh-msft Please see the attached file/project.

[webviewbug.zip](https://github.com/Microsoft/WinObjC/files/1005261/webviewbug.zip)

rajsesh-msft:
This is probably because this change https://github.com/Microsoft/WinObjC/commit/9c96cf25d3dad76fe60381c785a3977ea7d493ae didnt get cherry picked.

rajsesh-msft:
Closed with #2710

PR #2710, opened 2017-05-18
Fix github issue #2695 and #2704: UIWebView Release 47 Regression

yiyang-msft:
The issue here is another instance of side effect of switching to Cpp/winRT from projection. In this case, e.Uri() could be nullptr when webControl load HTML string block directly, instead of from a URL.  

Previously using projection, the following Object-C usage is totally fine even if uri is nil.
     e.uri.absoluteUri

But in cpp/Winrt,  e.Uri().AbsoluteUri() will cause AV if e.Uri() is nullptr 

It does not repro in our WocCatalog because the reproing code path is inside of delegate check code path and our test app WocCatalog  does not provide a delegate.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2710)

### DIFF
--- a/Frameworks/UIKit/UIWebView.mm
+++ b/Frameworks/UIKit/UIWebView.mm
@@ -160,17 +160,20 @@ static void _initUIWebView(UIWebView* self) {
     self->_xamlUnsupportedUriSchemeEventCookie =
         self->_xamlWebControl.UnsupportedUriSchemeIdentified(objcwinrt::callback([self] (const Controls::WebView&, const Controls::WebViewUnsupportedUriSchemeIdentifiedEventArgs& e) {
             if ([self.delegate respondsToSelector:@selector(webView:shouldStartLoadWithRequest:navigationType:)]) {
-                NSString* urlStr = objcwinrt::string(e.Uri().AbsoluteUri());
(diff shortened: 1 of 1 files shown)


Frameworks/UIKit/UIWebView.mm
```diff
@@ -136,16 +136,18 @@ static void _initUIWebView(UIWebView* self) {
...
-                    return;
+                if (e.Uri()) {
+                    NSString* urlStr = objcwinrt::string(e.Uri().AbsoluteUri());
+                    NSURL* url = [NSURL URLWithString : urlStr];
+                    NSURLRequest* request = [NSURLRequest requestWithURL : url];
+
+                    // ???? XAML doesn't expose this information to us
+                    UIWebViewNavigationType navigationType = UIWebViewNavigationTypeOther;
+
+                    if (![self.delegate webView:self shouldStartLoadWithRequest : request navigationType : navigationType]) {
```
rajsesh-msft: I think we may have to call the delegate even if the URI is nil - at least the old code did.

yiyang-msft:
also adding the fix for https://github.com/Microsoft/WinObjC/issues/2704

PR #2712, opened 2017-05-18
Remove NSGeometry and NSAffineTransform from Foundation

### LINKED ISSUE #2711: Remove NSGeometry and NSAffineTransform from Foundation
These are mac only APIs and cause circular dependencies to foundation

msft-Jeyaram:
NSAffineTransform and NSGeomerty are Mac only Apis and are not part of ios.

Inclusion of NSGeometry.h & NSAffineTransform.h prevents frameworks.core package from being used independently. 
NSGeometry.h defines NSPoint,NSRect,etc in terms of CGPoint,CGRect, etc. Thus making foundation depend on CoreGraphics. I need to break this dependency for packaging work, given these are not even part of ios and are not used, I've removed it.

Fixes #2711

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2712)


rajsesh-msft:
Why? What exactly is this hurting?

msft-Jeyaram:
@rajsesh-msft Inclusion of NSGeometry.h & NSAffineTransform.h prevents frameworks.core package from being used independently. 
NSGeometry.h defines NSPoint,NSRect,etc in terms of CGPoint,CGRect, etc. Thus making foundation depend on CoreGraphics. I need to break this dependency for packaging work, given these are not even part of ios and are not used, I've removed it.

Issue #2229, opened 2017-03-14
NSException details should be available in crash dumps


triplef:
Crash dump files (CABs) of crashes caused by a raised NSException should contain details about the NSException such as the exception name, reason, and userInfo properties.

When running in the debugger this information should be printed to the console (#666).

DHowett-MSFT:
The incoming fixes will make sure that the exception reason is preserved in heap dumps. We can open a new issue to track future work, such as including the userinfo and the like.

The Windows Error Reporting APIs don't offer the richness we're seeking here, so this might be a platform ask.

PR #2714, opened 2017-05-22
Use our unhandled exception filter to propagate COM error info

DHowett-MSFT:
This change will leverage our unhandled exception filter (which is now fixed 😛) to call `RoReportUnhandledError`. `RoReportUnhandledError` will take the error object created by `RoOriginateLanguageException`, box it, and stow it with its callstack for further future analysis.

There is a platform bug that prevents the exception text from being raised to the error reporter, but when it is resolved, this change will provide for rich crash report exception info.

**Note**: This only works for exceptions off the main thread; language boundaries are hard to get exceptions across, and the XAML exception handler installed by CoreApplication simply can't deal with our C++ exception model.

This is a partial fix for #2229.

### Exception Chain

```
NSException
 -> RoOriginateLanguageException
    -> First Chance (40080201)
    -> ::SetRestrictedErrorInfo(...boxed name and HRESULT from NSException...)
 -> _CxxThrowException
    -> Second Chance (0xE0 'msc')
NSUnhandledExceptionFilter
 -> RoReportUnhandledError(... the errorInfo object from above ...)
```

When `NSUnhandledExceptionFilter` is missing (or replaced by XAML), the second-chance exception bubbles up and aborts the application. Since it's a C++ exception, and no information can be forwarded about C++ exceptions, the exception text is lost.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2714)


DHowett-MSFT:
Another downside: The COM error reporting APIs we're using do not allow us to slice any parts off the stack; it'll always look like Foundation's _processException is the culprit.

Whether this is bad remains to be seen.

Frameworks/Foundation/NSException.mm
```diff
@@ -266,15 +268,9 @@ - (HRESULT)_hresult {
...
-    assert(hr != S_OK);
-
-    if(self.reason != nil) {
-        LPCWSTR reason = (LPCWSTR)([[self reason] cStringUsingEncoding:NSUnicodeStringEncoding]);
-        Microsoft::WRL::Wrappers::HStringReference hstrRef(reason);
-        RoOriginateError(hr, hstrRef.Get());
-    } else {
-        RoOriginateError(hr, nullptr);
-    }
+    auto reason = Strings::NarrowToWide<HSTRING>(self.reason);
```
aballway: nit: I'm not sure that `auto` is good here since the `<HSTRING>` suggests it's an HSTRING rather than an HString

Frameworks/Foundation/NSException.mm
```diff
@@ -225,11 +227,11 @@ case __HRESULT_FROM_WIN32(ERROR_NOT_READY):
...
 
 // Returns a HRESULT for an exception name
-// Note: The table is not the exact inverse of _exceptionNameForHRESULT. 
+// Note: The table is not the exact inverse of _exceptionNameForHRESULT.
 // E.g. E_ACCESSDENIED and RPC_E_WRONG_THREAD both map to NSObjectInaccessibleException.
 // However, NSObjectInaccessibleException only maps to E_ACCESSDENIED
-+ (HRESULT)_HRESULTForExceptionName:(NSString *)exceptionName {
-    if([exceptionName isEqualToString:NSInvalidArgumentException]) {
++ (HRESULT)_HRESULTForExceptionName:(NSString*)exceptionName {
+    if ([exceptionName isEqualToString:NSInvalidArgumentException]) {
```
aballway: nit: unnecessary elses

DHowett-MSFT: this change is due to formatting alone

DHowett-MSFT:
0a8281c fixes the other half issue #2229.

By using **C++/WinRT**'s `hresult_error`, we can taxi exception information across the language barrier. `hresult_error` lets us communicate that we have already created the `IRestrictedErrorInfo` object and that it should pass it along but translate it into an `HRESULT` for downstream consumers.

The only downside now is that in the crash bucket named `STOWED_EXCEPTION_XAML_TEXT_-[NSException_krazy]:_unrecognized_selector_sent_to_instance_16678D44_80070490` we encode an instance pointer (oops!) :wink:

Frameworks/Foundation/NSException.cl.cpp
```diff
@@ -54,9 +56,18 @@ extern "C" LONG WINAPI _NSWindowsUnhandledExceptionFilter(struct _EXCEPTION_POIN
             }
 
             if (foundNSException && foundobjc_object) {
+                Microsoft::WRL::ComPtr<IRestrictedErrorInfo> comErrorInfo;
+                ::GetRestrictedErrorInfo(&comErrorInfo); // save this here since the unhandled exception handler might stomp it.
```
yiyang-msft: >GetRestrictedErrorInfo [](start = 18, length = 22)

nit:check S_OK vs S_S_FALSE?

yiyang-msft: nit: probably move it closer to if (comErrorInfo) check

---
In reply to: [118055298](https://github.com/Microsoft/WinObjC/pull/2714#discussion_r118055298) [](ancestors = 118055298)

yiyang-msft:
:shipit:

Frameworks/Foundation/NSException.cl.cpp
yiyang-msft: nvm for the closer comment cause you mention _NSExceptionCallUnhandledExceptionHandler might stomp it.

---
In reply to: [118055620](https://github.com/Microsoft/WinObjC/pull/2714#discussion_r118055620) [](ancestors = 118055620,118055298)

DHowett-MSFT: S_FALSE would be fine here; it'll not return a `comErrorInfo`

PR #2716, opened 2017-05-23
[APIAnalyzer] Fix the headers to import and annotate properly 

msft-Jeyaram:
This fixes the following issues:
1. CGBase.h is no longer imported via Foundation.h, so some of the headers need CGBase.h to be imported.
2. Fix some Annotation mismatches

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2716)


yiyang-msft:
:shipit:

Issue #2687, opened 2017-05-09
Open source projections generation code


rajsesh-msft:
We may need to clear up the winmd generation part.  This should be a separate solution which produces the nuget packages.

PR #2722, opened 2017-05-24
update winmd2objc to use proper version of nuget and gitversiontask

msft-Jeyaram:
update winmd2objc to use proper versions of nuget and gitversiontask

- Nuget build 0.1.186 is the build that has been well tested with our repo, there are breaking changes in newer versions.
- GitVersionTask should use the same version as the common target, this is to make sure we don't have version mismatches

#2687,#2688

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2722)

### DIFF
--- a/tools/winmd2objc/lib/CodeGen.cpp
+++ b/tools/winmd2objc/lib/CodeGen.cpp
@@ -2326,8 +2326,7 @@ wstring generateNugetProject(const map<wstring, pair<wstring, vector<shared_ptr<
     }
 
     referenceBlock += LR"~(
-    <PackageReference Include="GitVersionTask" Version="*" />
-    <PackageReference Include="Nuget.Build.Packaging" Version="*" />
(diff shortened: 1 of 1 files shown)


tools/winmd2objc/lib/CodeGen.cpp
```diff
@@ -2326,8 +2326,7 @@ wstring generateNugetProject(const map<wstring, pair<wstring, vector<shared_ptr<
     }
 
     referenceBlock += LR"~(
-    <PackageReference Include="GitVersionTask" Version="*" />
-    <PackageReference Include="Nuget.Build.Packaging" Version="*" />
+    <PackageReference Include="Nuget.Build.Packaging" Version="0.1.186" />
```
DHowett-MSFT: no version for GitVersionTask?

msft-Jeyaram: The common target defines a version, so it will use whatever that version is.

Issue #2466, opened 2017-04-07
Investigate design/architecture surrounding NSURL____ classes


ms-jihua:
It seems moderately likely that our current object model surrounding NSURLConnection, Protocol, Session, Request and Response, are incorrect. Connection callbacks are currently built on top of Protocol callbacks, which do not appear to support uploading, making it unclear how Connection should support uploading. Additionally, the mapping of Projections objects to NS objects seems intuitively incorrect, as Protocol wraps nearly all of them, and the rest wrap basically none.

Spend some time figuring out what the correct approach here is.

rajsesh-msft:
Please be sure to add support for 
NSURLProtocol	propertyForKey:inRequest:
NSURLProtocol	setProperty:forKey:inRequest:
NSURLSessionTask	suspend

PR #2727, opened 2017-05-25
Implement NSURLProtocol propertyForKey:inRequest: methods

ms-jihua:
- Add unit test
- Also mark NSURLSessionTask suspend as not in plan

Related: #2466

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2727)


Frameworks/Foundation/NSURLSessionTask.mm
```diff
@@ -219,23 +219,18 @@ - (void)_stopLoading {
 }
 
 /**
- @Status Stub
+ @Status NotInPlan
```
ms-jihua: @rajsesh-msft there do not appear to be IHttpClient et al subclasses that can offer this to us. I was able to find the test page we talked about in the Foundation-Dev test app, but it only has Cancel/Resume and not suspend.

Frameworks/Foundation/NSMutableURLRequest.mm
```diff
@@ -76,4 +84,18 @@ - (instancetype)copyWithZone:(NSZone*)zone {
     return [self mutableCopyWithZone:zone];
 }
 
+- (void)_setProperty:(id)value forKey:(NSString*)key {
+    @synchronized(self) {
+        [self _lazyInitProperties];
+        [_properties setValue:value forKey:key];
```
DHowett-MSFT: how should this method handle `value==nil`?

include/Foundation/NSURLSession.h
```diff
@@ -181,27 +181,27 @@ FOUNDATION_EXPORT_CLASS
...
 - (void)URLSession:(NSURLSession*)session downloadTask:(NSURLSessionDownloadTask*)downloadTask didFinishDownloadingToURL:(NSURL*)location;
 
 @optional
 - (void)URLSession:(NSURLSession*)session
-      downloadTask:(NSURLSessionDownloadTask*)downloadTask
- didResumeAtOffset:(int64_t)fileOffset
-expectedTotalBytes:(int64_t)expectedTotalBytes;
+          downloadTask:(NSURLSessionDownloadTask*)downloadTask
+     didResumeAtOffset:(int64_t)fileOffset
+    expectedTotalBytes:(int64_t)expectedTotalBytes;
```
DHowett-MSFT: this formatting is objectively worse than the one that existed before :(

objective-c often lines up on colons; I noticed that clang-format wasn't doing that in a prior PR, but didn't think much of it.

Frameworks/Foundation/NSURLProtocol.mm
```diff
@@ -113,28 +117,27 @@ + (BOOL)canInitWithRequest:(NSURLRequest*)request {
...
 }
 
 /**
- @Status Stub
+ @Status Interoperable
  @Notes
 */
 + (void)removePropertyForKey:(NSString*)key inRequest:(NSMutableURLRequest*)request {
-    UNIMPLEMENTED();
+    [request _removePropertyForKey:key];
```
DHowett-MSFT: It might be worth checking whether this works on immutable requests... even though this is specified for `NSMutableURLRequest*`, we'll want to figure out what the error case looks like

Frameworks/Foundation/NSMutableURLRequest.mm
ms-jihua: Method has nonnull declaration on ref plat, but acts the same as remove if forced, which is already handled by the NSMutableDictionary side of things.

Frameworks/Foundation/NSURLProtocol.mm
ms-jihua: It just works anyway on the reference platform, actually, but do we really care to emulate that?

Frameworks/Foundation/NSMutableURLRequest.mm
DHowett-MSFT: NSMutableDictionary should not be doing that.

Additionally, actually, this should be using `setObject:forKey:` -- `setValue:...` is a KVC method that dictionary conforms to with additional logic. As well, since it's inherited, NSDictionary can't override its type to support lightweight genericity.

`setObject:forKey:`, however, does not remove on nil set.

Frameworks/Foundation/NSURLProtocol.mm
DHowett-MSFT: If it's not burdensome to emulate, why not do it?

Also, do we know how those properties hold up when the request is copied/mutableCopied?

include/Foundation/NSURLSession.h
ms-jihua: I agree, but this is pretty out of my hands :(

Frameworks/Foundation/NSMutableURLRequest.mm
ms-jihua: Oh yeah, oops.

Frameworks/Foundation/NSURLProtocol.mm
DHowett-MSFT: Oh crap. I just realized our `.clang-format` is configured for `Cpp`, not `ObjC` -- that could explain some of the really bizarre behaviours we're seeing.

DHowett-MSFT: I replied to the wrong comment, but ... 😛

ms-jihua: It gets copied. Implemented and added to test.

Frameworks/Foundation/NSURLRequest.mm
```diff
@@ -208,4 +210,37 @@ - (void)dealloc {
     [super dealloc];
 }
 
+- (void)_lazyInitProperties {
```
DHowett-MSFT: Here's a curious one: do they get encoded/decoded via archival?

ms-jihua: oh goshdarnit

Frameworks/Foundation/NSURLRequest.mm
```diff
@@ -138,6 +140,8 @@ - (void)encodeWithCoder:(NSCoder*)coder {
     [coder encodeBool:_HTTPShouldUsePipelining forKey:@"HTTPShouldUsePipelining"];
     [coder encodeBool:_HTTPShouldHandleCookies forKey:@"HTTPShouldHandleCookies"];
     [coder encodeBool:_allowsCellularAccess forKey:@"allowsCellularAccess"];
+
+    [coder encodeObject:_properties forKey:@"_properties"];
```
aballway: do we need to _lazyInitProperties for this?

Frameworks/Foundation/NSURLRequest.mm
```diff
@@ -208,4 +214,37 @@ - (void)dealloc {
     [super dealloc];
 }
 
+- (void)_lazyInitProperties {
+    @synchronized(self) {
```
aballway: nit: can you use a call_once for this?

Frameworks/Foundation/NSURLRequest.mm
ms-jihua: oh yeah probably should

DHowett-MSFT: It is fine to encode nil; if there is nothing in properties there's no point in storing an empty dictionary

DHowett-MSFT: In fact, it's actively bad to do that: it makes it so that all requests that get decoded get a properties dictionary even if they didn't have properties on the other end.

Use objective-c's nil propagation to your advantage!

Frameworks/Foundation/NSURLRequest.mm
DHowett-MSFT: That requires you to store an additional once token... and then make sure that it doesn't lie when you set _properties manually (through copying or decoding)...

PR #2729, opened 2017-05-25
Projections should not use custom props or targets.

### LINKED ISSUE #2728: [Projections][Winmd2objc] Projections should not use custom props or targets

msft-Jeyaram:
It should use the default props and targets from packaging.
The current packaging does not play nice with multiple props and targets.

Fixes #2728

#2687,#2688

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2729)


msft-Jeyaram:
blocked via #2730

DHowett-MSFT:
nit: reword title to be imperative mood

PR #2733, opened 2017-05-26
Implement steps to consume projections via nupkg vs pre-checked in binaries

### LINKED ISSUE #2688: Establish projections build + nuget publishing steps

msft-Jeyaram:
This will establish the WinObjC.Frameworks.UWP/MSAds packages and will be consumed by projects via the nupkg

- Removing the pre-checked in binaries, headers
- Removing the pre-checked in binary package project and removing it from the main build

Fixes #2688, #2687


msft-Jeyaram:
There is two instance of WINOBJC_SDK\ usage, that is being cleaned up now :)

msft-Jeyaram:
@rajsesh-msft  the projection code is generated after applying the fixes to winmd2objc  (my other CRs)

msvc/sdk-build.props
```diff
@@ -24,6 +24,10 @@
     </ProjectReference>
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Condition="'$(BuildingFrameworksCore)' != 'true'" Include="WinObjC.Frameworks.UWP" Version="*" />
```
DHowett-MSFT: This is an interesting approach!

Do you think it should go down with the other PackageReferences?

build/Tests/UnitTests/Starboard/Starboard.UnitTests.vcxproj
```diff
@@ -223,6 +223,9 @@
   <ItemGroup>
     <ClInclude Include="..\..\..\..\tests\unittests\Starboard\LifetimeCounting.h" />
   </ItemGroup>
+    <ItemGroup>
+    <PackageReference Include="WinObjC.Frameworks.UWP" Version="*" />
```
DHowett-MSFT: Indentation is odd here!

msvc/sdk-build.props
msft-Jeyaram: nope, we want to avoid Frameworks.Core from consuming this package.

DHowett-MSFT: That doesn't mean it can't go in the same block of package references... that just means we have to keep the Condition

msft-Jeyaram: ah, I assumed you meant to just include it without the condition.
Consider it moved.

msft-Jeyaram:
This saw random failures with NotificationQueue and filemanager tests.
Now it second run and local run seem to be fine

msft-Jeyaram:
ping! @rajsesh-msft @DHowett

PR #2750, opened 2017-05-31
Foundation the NSNotificationQueueTests to Move.OnlyNSNotificationQueue

DHowett-MSFT:
Two things:

1. THRunLoopSpinner does need an input source on OS X (@ms-jihua you were totally right.) That isn't fixed here.
2. NSNotificationQueue is far simpler in WinObjC; the runloop spinner forces the run loop to never be "idle" on OS X, which makes the "three kinds of notification" test fail.'

This fixes the OS X test break.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2750)

### DIFF
--- a/tests/unittests/Foundation/NSNotificationQueueTests.mm
+++ b/tests/unittests/Foundation/NSNotificationQueueTests.mm
@@ -18,8 +18,10 @@
 #import <TestFramework.h>
 #import <Starboard/SmartTypes.h>
 #import <Helpers/TestHelpers.h>
+#import <Windows.h>
 
-TEST(NSNotificationQueue, PostNow) {
+// NSNotificationQueue has different semantics on Windows.
+OSX_DISABLED_TEST(NSNotificationQueue, PostNow) {
     static NSString* s_TestNotificationName = @(GetTestFullName().c_str());
     NSNotificationCenter* notificationCenter = [[NSNotificationCenter new] autorelease];
     StrongId<THRunLoopSpinner> spinner{ woc::TakeOwnership, [THRunLoopSpinner new] };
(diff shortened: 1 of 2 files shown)


DHowett-MSFT:
Due to a class initialization order issue with WindowsOnly tests, NSNotificationQueue cannot be moved there. I've switched to `OSX_DISABLED_TEST` and made sure it compiled on OS X.

tests/unittests/Foundation/NSNotificationQueueTests.mm
```diff
@@ -117,7 +122,7 @@
     }];
 
     __block THBooleanCondition* condition = [[THBooleanCondition new] autorelease];
-    __block uint32_t counter = 0;
+    __block long counter = 0;
```
aballway: excuse me

tests/unittests/Foundation/NSNotificationQueueTests.mm
```diff
@@ -18,8 +18,10 @@
 #import <TestFramework.h>
 #import <Starboard/SmartTypes.h>
 #import <Helpers/TestHelpers.h>
+#import <Windows.h>
```
aballway: is this needed?

DHowett:
fwiw, the name of this pull request will be ignored when I choose _rebase and merge_

tests/unittests/Foundation/NSNotificationQueueTests.mm
DHowett: yes. `Interlocked*` are from Windows.h, and none of the OS X platform headers include Windows.h for us

tests/unittests/Foundation/NSNotificationQueueTests.mm
DHowett: sorry man, that's how InterlockedExchange is specified

tests/unittests/Foundation/NSNotificationQueueTests.mm
aballway: Interlocked? AFAICT this just changed TEST->OSX_DISABLED_TEST

DHowett: There're three issues here:

* The tests don't compile on OS X
* The tests don't _work_ on OS X
* The tests can't be in WindowsOnly because of ... something? Something fun and compiler-related due to the fact that WO tests consume Foundation*Lib*

aballway: OIC

DHowett:
(title fixed)

tests/unittests/Foundation/NSNotificationQueueTests.mm
aballway: https://static1.squarespace.com/static/55aa9727e4b09b833beef19c/t/56d9e42562cd940b4bdd4413/1457120340701/

PR #2753, opened 2017-06-01
Implement missing UIBezierPath APIs and add tests

### LINKED ISSUE #2700: UIBezierPath currentPoint never used
Hi

UIBezierPath currentPoint property never used and is not marked as stub. Any chance to implement currentPoint functionality?

Kind regards,
Julia

### LINKED ISSUE #2701: UIBezierPath containsPoint not implemented
Hi,
Could you please tell if there are any plans to implement containsPoint method of UIBezierPath ?

### LINKED ISSUE #2702: UIBezierPath applyTransform does not apply to the path
Hi,
When we use applyTransform after draw of UIBezierPath, transformation does not apply to the path:

```
UIBezierPath *path = [UIBezierPath bezierPath];
[path moveToPoint:CGPointMake(50, 50)];
[path moveToPoint:CGPointMake(300, 300)];
[path closePath];
CGAffineTransform transform = CGAffineTransformMakeScale(2, 2);
[path applyTransform:transform];
```

As per documentation for [applyTransform](https://developer.apple.com/reference/uikit/uibezierpath/1624340-applytransform) 

> This method applies the specified transform to the path’s points immediately. The modifications made to the path object are permanent. If you do not want to permanently modify a path object, you should consider applying the transform to a copy.

As a workaround for winobjc, I apply transform before draw of path, but this is not the best approach.

```
UIBezierPath *path = [UIBezierPath bezierPath];
CGAffineTransform transform = CGAffineTransformMakeScale(2, 2);
[path applyTransform:transform];
[path moveToPoint:CGPointMake(50, 50)];
[path moveToPoint:CGPointMake(300, 300)];
[path closePath];
```

MSFTFox:
Implement some missing APIs in UIBezierPath in terms of CGPath. Some general clean up was also done. Automated pixel comparison tests were added as well as simple API tests for the newly enabled APIs.

Fixes #2700 
Fixes #2701 
Fixes #2702

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2753)

### DIFF
--- a/Frameworks/CoreGraphics/CGPath.mm
+++ b/Frameworks/CoreGraphics/CGPath.mm
@@ -46,10 +46,10 @@ public:
         : m_geometrySink(sink),
           m_lastPoint{ 0, 0 },
           m_isFigureOpen(false),
-          m_allowsFigureCalls(false),
+          m_allowsFigureCalls(0),
           m_hasGeometryStarted(false),
           m_lastClose(D2D1_FIGURE_END_CLOSED),
-          m_allowsFigureEndExceptFinal(false),
+          m_allowsFigureEndExceptFinal(0),
           m_delayedEndFigure(false) {
     }
 
@@ -115,7 +115,7 @@ public:
     }
 
     STDMETHOD_(void, BeginFigure)(D2D1_POINT_2F startPoint, D2D1_FIGURE_BEGIN figureBegin) {
-        if (m_allowsFigureCalls) {
+        if (m_allowsFigureCalls > 0) {
             _BeginFigure(startPoint, figureBegin);
         }
     }
@@ -135,10 +135,10 @@ public:
 
     // Only end the figure if we're requesting to actually close a figure, or if we have allowed the backing sink to dictate figure calls.
     STDMETHOD_(void, EndFigure)(D2D1_FIGURE_END figureEnd) {
-        if (m_allowsFigureEndExceptFinal && figureEnd != D2D1_FIGURE_END_CLOSED) {
+        if (m_allowsFigureEndExceptFinal > 0 && figureEnd != D2D1_FIGURE_END_CLOSED) {
             m_delayedEndFigure = true;
             m_delayedEndFigureClosure = figureEnd;
-        } else if (figureEnd == D2D1_FIGURE_END_CLOSED || m_allowsFigureCalls) {
+        } else if (figureEnd == D2D1_FIGURE_END_CLOSED || m_allowsFigureCalls > 0) {
             _EndFigure(figureEnd);
         }
     }
@@ -159,7 +159,7 @@ public:
     }
 
     void SetAllowsFigureCalls(bool allows) {
-        m_allowsFigureCalls = allows;
+        m_allowsFigureCalls += (allows ? 1 : -1);
     }
 
     bool IsFigureEndClosed() {
@@ -199,7 +199,7 @@ public:
     }
 
     void AllowAllExceptFinalEndFigure(bool allow) {
-        m_allowsFigureEndExceptFinal = allow;
+        m_allowsFigureEndExceptFinal += (allow ? 1 : -1);
         m_delayedEndFigure = false;
     }
 
@@ -222,9 +222,9 @@ private:
     // closed.
     bool m_hasGeometryStarted;
     // True if this geometry can end its own figures.
-    bool m_allowsFigureCalls;
+    int m_allowsFigureCalls;
     // True if this geometry will leave it's last EndFigure call ignored.
-    bool m_allowsFigureEndExceptFinal;
+    int m_allowsFigureEndExceptFinal;
     // True when the next figure created should end the previous figure first.
     bool m_delayedEndFigure;
     // The figure closure type to end the last figure with.
@@ -969,12 +969,7 @@ bool CGPathIsEmpty(CGPathRef path) {
         return true;
     }
 
-    UINT32 count;
-
-    RETURN_FALSE_IF_FAILED(path->ClosePath());
-
-    RETURN_FALSE_IF_FAILED(path->GetPathGeometry()->GetFigureCount(&count));
-    return count == 0;
+    return !path->IsGeometryStarted();
 }
 
 /**
--- a/Frameworks/CoreGraphics/CGPath.mm
+++ b/Frameworks/CoreGraphics/CGPath.mm
@@ -218,7 +218,7 @@ private:
     D2D1_POINT_2F m_lastPoint;
     // The starting point this geometry will move to if it is closed.
     D2D1_POINT_2F m_startPoint;
-    // Whether or not the figure is open and ready to recieve new path information
+    // Whether or not the figure is open and ready to receive new path information
     bool m_isFigureOpen;
     // Whether any geometry has been started. This can be inferred from other APIs, but those would require the geometry to be closed.
     bool m_hasGeometryStarted;
--- a/Frameworks/CoreGraphics/CGPath.mm
+++ b/Frameworks/CoreGraphics/CGPath.mm
@@ -158,8 +158,9 @@ public:
         return m_geometrySink.Get();
     }
 
-    void SetAllowsFigureCalls(bool allows) {
+    HRESULT SetAllowsFigureCalls(bool allows) {
         m_allowsFigureCalls += (allows ? 1 : -1);
+        return m_allowsFigureCalls >= 0 ? S_OK : E_INVALIDARG;
     }
 
     bool IsFigureEndClosed() {
@@ -198,9 +199,10 @@ public:
         m_geometrySink = sink;
     }
 
-    void AllowAllExceptFinalEndFigure(bool allow) {
+    HRESULT AllowAllExceptFinalEndFigure(bool allow) {
         m_allowsFigureEndExceptFinal += (allow ? 1 : -1);
         m_delayedEndFigure = false;
+        return m_allowsFigureEndExceptFinal >= 0 ? S_OK : E_INVALIDARG;
     }
 
 private:
@@ -218,12 +220,11 @@ private:
     D2D1_POINT_2F m_startPoint;
     // Whether or not the figure is open and ready to recieve new path information
     bool m_isFigureOpen;
-    // Whether or not any geometry has been started. This can be inferred from other APIs, but those would require the geometry to be
-    // closed.
+    // Whether any geometry has been started. This can be inferred from other APIs, but those would require the geometry to be closed.
     bool m_hasGeometryStarted;
-    // True if this geometry can end its own figures.
+    // Greater than 0 if this geometry can end its own figures.
     int m_allowsFigureCalls;
-    // True if this geometry will leave it's last EndFigure call ignored.
+    // Greater than 0 if this geometry will leave it's last EndFigure call ignored.
     int m_allowsFigureEndExceptFinal;
     // True when the next figure created should end the previous figure first.
     bool m_delayedEndFigure;
@@ -270,8 +271,8 @@ struct __CGPath : CoreFoundation::CppBase<__CGPath> {
         return geometrySink && geometrySink->IsGeometryStarted();
     }
 
-    void SetAllowsFigureCalls(bool allows) {
-        geometrySink->SetAllowsFigureCalls(allows);
+    HRESULT SetAllowsFigureCalls(bool allows) {
+        return geometrySink->SetAllowsFigureCalls(allows);
     }
 
     CGPoint GetCurrentPoint() const {
@@ -303,9 +304,12 @@ struct __CGPath : CoreFoundation::CppBase<__CGPath> {
         return &lastTransform;
     }
 
-    void AllowAllExceptFinalEndFigure(bool allow) {
-        SetAllowsFigureCalls(allow);
-        geometrySink->AllowAllExceptFinalEndFigure(allow);
+    HRESULT AllowAllExceptFinalEndFigure(bool allow) {
+        HRESULT hr = SetAllowsFigureCalls(allow);
+        if (SUCCEEDED(hr)) {
+            hr = geometrySink->AllowAllExceptFinalEndFigure(allow);
+        }
+        return hr;
     }
 
     // A private helper function for re-opening a path geometry. CGPath does not
@@ -335,9 +339,9 @@ struct __CGPath : CoreFoundation::CppBase<__CGPath> {
 
             geometrySink->SetBackingSink(newBackingSink.Get());
             // Allow all calls except final endfigure
-            AllowAllExceptFinalEndFigure(true);
+            RETURN_IF_FAILED(AllowAllExceptFinalEndFigure(true));
             RETURN_IF_FAILED(pathGeometry->Stream(geometrySink.Get()));
-            AllowAllExceptFinalEndFigure(false);
+            RETURN_IF_FAILED(AllowAllExceptFinalEndFigure(false));
             geometrySink->SetFillMode(D2D1_FILL_MODE_WINDING);
 
             pathGeometry = newPath;
@@ -430,9 +434,9 @@ struct __CGPath : CoreFoundation::CppBase<__CGPath> {
             d2dTransform = __CGAffineTransformToD2D_F(*transform);
         }
 
-        AllowAllExceptFinalEndFigure(true);
+        RETURN_IF_FAILED(AllowAllExceptFinalEndFigure(true));
         RETURN_IF_FAILED(path->GetPathGeometry()->Widen(lineWidth, newStrokeStyle.Get(), d2dTransform, geometrySink.Get()));
-        AllowAllExceptFinalEndFigure(false);
+        RETURN_IF_FAILED(AllowAllExceptFinalEndFigure(false));
 
         pathGeometry = newPath;
 
@@ -597,9 +601,9 @@ CGMutablePathRef CGPathCreateMutableCopy(CGPathRef path) {
     // Otherwise the D2D calls will return that a bad state has been entered.
     FAIL_FAST_IF_FAILED(path->ClosePath());
 
-    mutableRet->AllowAllExceptFinalEndFigure(true);
+    FAIL_FAST_IF_FAILED(mutableRet->AllowAllExceptFinalEndFigure(true));
     FAIL_FAST_IF_FAILED(path->GetPathGeometry()->Stream(mutableRet->GetGeometrySink()));
-    mutableRet->AllowAllExceptFinalEndFigure(false);
+    FAIL_FAST_IF_FAILED(mutableRet->AllowAllExceptFinalEndFigure(false));
 
     mutableRet->SetCurrentPoint(path->GetCurrentPoint());
     mutableRet->SetLastTransform(path->GetLastTransform());
@@ -872,9 +876,9 @@ void CGPathAddPath(CGMutablePathRef path, const CGAffineTransform* transform, CG
     FAIL_FAST_IF_FAILED(toAdd->ClosePath());
     CGPoint newStart = toAdd->GetStartingPoint();
     //    CGPathMoveToPoint(path, transform, newStart.x, newStart.y);
-    path->AllowAllExceptFinalEndFigure(true);
+    FAIL_FAST_IF_FAILED(path->AllowAllExceptFinalEndFigure(true));
     FAIL_FAST_IF_FAILED(path->AddGeometryToPathWithTransformation(toAdd->GetPathGeometry(), transform));
-    path->AllowAllExceptFinalEndFigure(false);
+    FAIL_FAST_IF_FAILED(path->AllowAllExceptFinalEndFigure(false));
 }
 
 /**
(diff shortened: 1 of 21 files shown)


Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -135,10 +135,10 @@
 
     // Only end the figure if we're requesting to actually close a figure, or if we have allowed the backing sink to dictate figure calls.
     STDMETHOD_(void, EndFigure)(D2D1_FIGURE_END figureEnd) {
-        if (m_allowsFigureEndExceptFinal && figureEnd != D2D1_FIGURE_END_CLOSED) {
+        if (m_allowsFigureEndExceptFinal > 0 && figureEnd != D2D1_FIGURE_END_CLOSED) {
             m_delayedEndFigure = true;
             m_delayedEndFigureClosure = figureEnd;
-        } else if (figureEnd == D2D1_FIGURE_END_CLOSED || m_allowsFigureCalls) {
+        } else if (figureEnd == D2D1_FIGURE_END_CLOSED || m_allowsFigureCalls > 0) {
```
DHowett-MSFT: technically, evaluating `integer` as a logical expression folds to `integer != 0`

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -159,7 +159,7 @@
     }
 
     void SetAllowsFigureCalls(bool allows) {
-        m_allowsFigureCalls = allows;
+        m_allowsFigureCalls += (allows ? 1 : -1);
```
DHowett-MSFT: Negative values can be considered failures, right? Should we update this method to return an HRESULT?

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -969,12 +969,7 @@ bool CGPathIsEmpty(CGPathRef path) {
         return true;
     }
 
-    UINT32 count;
-
-    RETURN_FALSE_IF_FAILED(path->ClosePath());
-
-    RETURN_FALSE_IF_FAILED(path->GetPathGeometry()->GetFigureCount(&count));
-    return count == 0;
+    return !path->IsGeometryStarted();
```
DHowett-MSFT: Interesting; did this exist the whole time? Does it have the same semantics?

Frameworks/UIKit/UIBezierPath.mm
```diff
@@ -64,57 +50,31 @@ + (instancetype)bezierPathWithRoundedRect:(CGRect)rect byRoundingCorners:(UIRect
     CGFloat outside_top = rect.origin.y;
     CGFloat outside_left = rect.origin.x;
 
-    CGPathMoveToPoint(ret->_path, getTransform(ret), innerRect.origin.x, outside_top);
+    CGPathMoveToPoint(ret->_workingPath, NULL, innerRect.origin.x, outside_top);
```
DHowett-MSFT: Interesting. This doesn't respect the corner flags?

Frameworks/UIKit/UIBezierPath.mm
```diff
@@ -190,75 +150,76 @@ - (void)addArcWithCenter:(CGPoint)center
...
-    CGPathCloseSubpath(_path);
+    CGPathCloseSubpath(_workingPath);
 }
 
 /**
  @Status Interoperable
 */
 - (CGPathRef)CGPath {
-    return _path;
+    return CGPathCreateCopy(_workingPath);
```
DHowett-MSFT: You will want to cast this to id and autorelease it. Right now it's returning a +1.

I guess _that_ is why it uses `NS_RETURNS_INNER_POINTER`!

Frameworks/UIKit/UIBezierPath.mm
```diff
@@ -190,75 +150,76 @@ - (void)addArcWithCenter:(CGPoint)center
...
 }
 
 /**
  @Status Interoperable
 */
 - (void)setCGPath:(CGPathRef)path {
-    CGPathRetain(path);
-    CGPathRelease(_path);
-    _path = path;
+    CGPathRelease(_workingPath);
```
DHowett-MSFT: We should be using StrongCF for this.

Frameworks/UIKit/UIBezierPath.mm
```diff
@@ -190,75 +150,76 @@ - (void)addArcWithCenter:(CGPoint)center
...
+    _workingPath = CGPathCreateMutableCopy(path);
 }
 
 /**
  @Status Interoperable
 */
 - (void)addQuadCurveToPoint:(CGPoint)endPoint controlPoint:(CGPoint)controlPoint {
-    CGPathAddQuadCurveToPoint(_path, getTransform(self), controlPoint.x, controlPoint.y, endPoint.x, endPoint.y);
+    CGPathAddQuadCurveToPoint(_workingPath, NULL, controlPoint.x, controlPoint.y, endPoint.x, endPoint.y);
```
DHowett-MSFT: replace NULL with nullptr everywhere

Frameworks/UIKit/UIBezierPath.mm
```diff
@@ -366,25 +299,34 @@ - (void)setLineDash:(const CGFloat*)pattern count:(NSInteger)count phase:(CGFloa
     _patternCount = count;
     _dashPhase = phase;
 
-    assert(sizeof(float) == sizeof(CGFloat));
     if (count > 0) {
-        _pattern = (float*)malloc(sizeof(float) * count);
-        memcpy(_pattern, pattern, sizeof(float) * count);
+        _pattern = (CGFloat*)malloc(sizeof(CGFloat) * count);
```
DHowett-MSFT: `!DANGER! !DANGER! !DANGER!`

Please use a smart pointer type for this.

DHowett-MSFT: Or a vector!

build/build.sln
```diff
@@ -743,6 +743,9 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "CoreGraphics", "CoreGraphics", "{0A79AFE5-2685-433C-BC78-A4AD09CD7BF8}"
 EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "CoreGraphics.UnitTests", "Tests\UnitTests\CoreGraphics\CoreGraphics.UnitTests.vcxproj", "{DA6C01EA-A22E-4807-BACE-63C5C6ABAE77}"
+	ProjectSection(ProjectDependencies) = postProject
```
DHowett-MSFT: Interesting; I didn't think other dependencies added this block.

Can you remove it and make sure everything else works? We're trying to avoid expressing dependencies in SLN

include/UIKit/UIBezierPath.h
```diff
@@ -68,12 +68,12 @@ UIKIT_EXPORT_CLASS
...
 - (void)closePath;
-- (void)removeAllPoints STUB_METHOD;
+- (void)removeAllPoints;
 - (void)appendPath:(UIBezierPath*)bezierPath;
 
-@property (nonatomic) CGPathRef CGPath NS_RETURNS_INNER_POINTER CF_RETURNS_NOT_RETAINED;
-@property (nonatomic, readonly) CGPoint currentPoint;
+@property (readonly) CGPoint currentPoint;
 
+@property (nonatomic) CGPathRef CGPath NS_RETURNS_INNER_POINTER CF_RETURNS_NOT_RETAINED;
```
DHowett-MSFT: both of these annotations are now incorrect.

tests/unittests/UIKit/UIBezierPathTests.mm
```diff
@@ -0,0 +1,56 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//******************************************************************************
+
+#import <TestFramework.h>
+#import <UIKit/UIBezierPath.h>
+#import <Starboard/SmartTypes.h>
+
+TEST(UIBezierPath, containsPoint) {
```
DHowett-MSFT: There is a file called SimpleTest that does not have a test backing it. Please make sure all test images are accounted for!

Frameworks/CoreGraphics/CGPath.mm
MSFTFox: I've had this in for a while used in other locations. It's just more simple than querying the amount of figures or segments.

Frameworks/UIKit/UIBezierPath.mm
MSFTFox: Yup, this implementation ignores them entirely.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -222,9 +222,9 @@ void _delayedEndFigureCall() {
     // closed.
     bool m_hasGeometryStarted;
     // True if this geometry can end its own figures.
```
rajsesh-msft: >/ True if this geometry can end its own f [](start = 5, length = 41)

comments out of date.

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -222,9 +222,9 @@ void _delayedEndFigureCall() {
...
-    bool m_allowsFigureCalls;
+    int m_allowsFigureCalls;
```
rajsesh-msft: >int m_allowsFigureCalls [](start = 4, length = 23)

also, what exactly is this cleaning up?  you are going from bool to tri-state, why?

DHowett-MSFT: @rajsesh-msft I think he's going from bool to "stack" -- to allow for nested calls to enable/disable figure state without trampling eachother

rajsesh-msft: Yes, saw that later on.  but this is not technically clean-up, it is a design change, would be useful to call that out in the comments.

---
In reply to: [119761248](https://github.com/Microsoft/WinObjC/pull/2753#discussion_r119761248) [](ancestors = 119761248)

MSFTFox: Exactly what Dustin said. There are times in the code where allowing figure calls will call into other functions which will then disable it. Having a stack of allow/disallow let's the top level call be the only one to disable this permission.

Frameworks/UIKit/UIBezierPath.mm
```diff
@@ -178,46 +180,45 @@ - (void)closePath {
...
-    CGPathAddCurveToPoint(_workingPath, NULL, controlPoint1.x, controlPoint1.y, controlPoint2.x, controlPoint2.y, endPoint.x, endPoint.y);
+    CGPathAddCurveToPoint(
+        _workingPath, nullptr, controlPoint1.x, controlPoint1.y, controlPoint2.x, controlPoint2.y, endPoint.x, endPoint.y);
 }
 
 /**
  @Status Interoperable
 */
 - (void)applyTransform:(CGAffineTransform)transform {
     CGPathRef transformedPath = CGPathCreateMutableCopyByTransformingPath(_workingPath, &transform);
```
ms-jihua: why not just oneline this?

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -218,12 +220,11 @@ void _delayedEndFigureCall() {
     D2D1_POINT_2F m_startPoint;
     // Whether or not the figure is open and ready to recieve new path information
```
ms-jihua: ultra nit: receive

MSFTFox: i before e except after c

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -216,15 +218,14 @@ void _delayedEndFigureCall() {
...
-    // Whether or not any geometry has been started. This can be inferred from other APIs, but those would require the geometry to be
-    // closed.
+    // Whether any geometry has been started. This can be inferred from other APIs, but those would require the geometry to be closed.
     bool m_hasGeometryStarted;
-    // True if this geometry can end its own figures.
-    bool m_allowsFigureCalls;
-    // True if this geometry will leave it's last EndFigure call ignored.
-    bool m_allowsFigureEndExceptFinal;
+    // Greater than 0 if this geometry can end its own figures.
+    int m_allowsFigureCalls;
```
aballway: nit: you may want to rename these, since "allows" implies it's a bool

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -303,9 +304,12 @@ void SetLastTransform(const CGAffineTransform* transform) {
         return &lastTransform;
     }
 
-    void AllowAllExceptFinalEndFigure(bool allow) {
-        SetAllowsFigureCalls(allow);
-        geometrySink->AllowAllExceptFinalEndFigure(allow);
+    HRESULT AllowAllExceptFinalEndFigure(bool allow) {
+        HRESULT hr = SetAllowsFigureCalls(allow);
```
aballway: nit: might just want to `RETURN_IF_FAILED` the first one and return the second to keep w/ the pattern we used in CG

Frameworks/UIKit/UIBezierPath.mm
```diff
@@ -190,75 +152,74 @@ - (void)addArcWithCenter:(CGPoint)center
...
-    CGPathCloseSubpath(_path);
+    CGPathCloseSubpath(_workingPath);
 }
 
 /**
  @Status Interoperable
 */
 - (CGPathRef)CGPath {
-    return _path;
+    return (CGPathRef)[(id)CGPathCreateCopy(_workingPath) autorelease];
```
aballway: nit: if you cast to `CGPathRef` instead of `id` do you still need to double cast?

Frameworks/UIKit/UIBezierPath.mm
```diff
@@ -316,97 +277,72 @@ - (void)stroke {
...
+    _patternCount = count;
+    _dashPhase = phase;
 
-/**
- @Status Interoperable
-*/
-- (void)setLineJoinStyle:(CGLineJoin)join {
-    _lineJoin = join;
+    if (count > 0) {
+        std::vector<CGFloat> newPattern(pattern, pattern + count);
```
aballway: nit: how should this handle a nil pattern?

Frameworks/CoreGraphics/CGPath.mm
MSFTFox: Oh... yeah. I don't know why I did it that way when I've used the macros in every other conceivable instance.

Frameworks/UIKit/UIBezierPath.mm
MSFTFox: I believe you do. It must be of type CGPath at return but requires an id for the autorelease.

Frameworks/UIKit/UIBezierPath.mm
MSFTFox: It should simply clear the pattern. I'll add a case for that.

Frameworks/UIKit/UIBezierPath.mm
aballway: Oh I see now.  Should this be a copy or a shared CGPathRef?  Like if a consumer modified path.CGPath would the original path be changed?

include/UIKit/UIBezierPath.h
DHowett-MSFT: both of these annotations are still incorrect.

tests/unittests/UIKit/UIBezierPathTests.mm
```diff
@@ -0,0 +1,72 @@
...
```
DHowett-MSFT: Please standardize on test names being in `PascalCase`

Frameworks/UIKit/UIBezierPath.mm
```diff
@@ -190,75 +152,74 @@ - (void)addArcWithCenter:(CGPoint)center
...
 }
 
 /**
  @Status Interoperable
 */
 - (void)setCGPath:(CGPathRef)path {
-    CGPathRetain(path);
-    CGPathRelease(_path);
-    _path = path;
+    _workingPath = woc::MakeStrongCF(CGPathCreateMutableCopy(path));
```
DHowett-MSFT: y'could just `.attach(..CreateCopy...)` instead of doing this.

DHowett-MSFT: The way it is now, it creates a temporary StrongCF, copies it (+1) to _workingPath, then destructs it (-1)... If you use `attach` you get a true `+0`

Frameworks/UIKit/UIBezierPath.mm
```diff
@@ -190,75 +152,74 @@ - (void)addArcWithCenter:(CGPoint)center
...
 }
 
 /**
  @Status Interoperable
 */
 - (instancetype)init {
-    _path = CGPathCreateMutable();
-    _lineWidth = 1.0f;
+    if (self = [super init]) {
+        _workingPath = woc::MakeStrongCF(CGPathCreateMutable());
```
DHowett-MSFT: attach _ut supra_

Frameworks/UIKit/UIBezierPath.mm
```diff
@@ -316,97 +277,72 @@ - (void)stroke {
...
+    }
 }
 
 /**
  @Status Interoperable
 */
-- (void)setUsesEvenOddFillRule:(BOOL)usesEvenOdd {
-    _usesEvenOddFillRule = usesEvenOdd;
+- (void)removeAllPoints {
+    _workingPath = woc::MakeStrongCF(CGPathCreateMutable());
```
DHowett-MSFT: attach _ut supra_

Frameworks/UIKit/UIBezierPath.mm
```diff
@@ -316,97 +277,72 @@ - (void)stroke {
...
+        std::copy(_pattern.begin(), _pattern.end(), pattern);
     }
 }
 
 /**
  @Status Stub
 */
-- (void)removeAllPoints {
+- (void)fillWithBlendMode:(CGBlendMode)blendMode alpha:(CGFloat)alpha {
     UNIMPLEMENTED();
```
DHowett-MSFT: technically stupidly easy to implement... and the others should be implemented _in terms of these ones_ by passing kCGBlendNormal or kCGBlendSourceOver

Frameworks/UIKit/UIBezierPath.mm
```diff
@@ -261,27 +251,23 @@ - (void)addClip {
...
-    CGContextRestoreGState(ctx);
+    [self strokeWithBlendMode:kCGBlendModeNormal alpha:1];
 }
 
 /**
  @Status Interoperable
 */
 - (void)setLineDash:(const CGFloat*)pattern count:(NSInteger)count phase:(CGFloat)phase {
+    if (!pattern || count == 0) {
+        _patternCount = 0;
```
DHowett-MSFT: saaaaayyyy...

DHowett-MSFT: doesn't `_pattern.size()` make a good stand-in for keeping a separate _patternCount?

Frameworks/UIKit/UIBezierPath.mm
```diff
@@ -261,27 +251,23 @@ - (void)addClip {
...
+        _dashPhase = 0;
+        _pattern.clear();
+        return;
```
DHowett-MSFT: :+1:

Frameworks/UIKit/UIBezierPath.mm
```diff
@@ -314,14 +300,38 @@ - (void)getLineDash:(CGFloat*)pattern count:(NSInteger*)count phase:(CGFloat*)ph
  @Status Stub
```
DHowett-MSFT: @Status nonyabiznis

include/UIKit/UIBezierPath.h
```diff
@@ -73,7 +73,7 @@ UIKIT_EXPORT_CLASS
 
 @property (readonly) CGPoint currentPoint;
 
-@property (nonatomic) CGPathRef CGPath NS_RETURNS_INNER_POINTER CF_RETURNS_NOT_RETAINED;
+@property (nonatomic) CGPathRef CGPath;
 @property (nonatomic) CGFloat lineWidth;
 @property (nonatomic) CGLineCap lineCapStyle;
 @property (nonatomic) CGLineJoin lineJoinStyle;
```
DHowett-MSFT: I can't comment on the STUB_METHODs on `blendmode`

MSFTFox: Woops, I forgot that. Good catch. Thanks!

Frameworks/CoreGraphics/CGPath.mm
```diff
@@ -303,9 +304,10 @@ void SetLastTransform(const CGAffineTransform* transform) {
         return &lastTransform;
     }
 
-    void AllowAllExceptFinalEndFigure(bool allow) {
-        SetAllowsFigureCalls(allow);
-        geometrySink->AllowAllExceptFinalEndFigure(allow);
+    HRESULT AllowAllExceptFinalEndFigure(bool allow) {
+        RETURN_IF_FAILED(SetAllowsFigureCalls(allow));
+        RETURN_IF_FAILED(geometrySink->AllowAllExceptFinalEndFigure(allow));
```
aballway: nit: you can just return this

Frameworks/UIKit/UIBezierPath.mm
```diff
@@ -300,113 +250,106 @@ - (void)addClip {
...
 */
-- (void)dealloc {
-    if (_pattern) {
-        free(_pattern);
-        _pattern = NULL;
+- (void)setLineDash:(const CGFloat*)pattern count:(NSInteger)count phase:(CGFloat)phase {
+    if (!pattern || count == 0) {
+        _dashPhase = 0;
+        _pattern.clear();
+        return;
```
aballway: http://i.imgur.com/YkoCx.gif

rajsesh-msft:
The CI build passed but GitHub was down when it passed, so rather than re-run the build, synced with @MSFTFox offline and force merging this.

PR #2755, opened 2017-06-02
Fix thread leak in NSURLConnection

### LINKED ISSUE #2754: NSURLConnection is leaking protocol handler threads
NSURLConnection is never signalling _runLoopCancelSource. If the run loop doesn't get a spurious wakeup, _done is never checked, and the thread is leaked.

ms-jihua:
Fixes #2754

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2755)


ms-jihua:
uhh what the heck? i have no idea why this diff is showing as this big. one sec

DHowett-MSFT:
@ms-jihua Your `core.autocrlf` setting has been toggled.

ms-jihua:
ok it's fixed

Issue #2756, opened 2017-06-02
[INVESTIGATION] CFData occasionally fails to realloc


DHowett-MSFT:
This is triggering a deadlock in our test fixture. The reallocation fails holding the heap lock, and then the test fixture attempts to allocate more memory. To do so, it must acquire the lock exclusively . . .

The realloc fails in `__CFGrowData` because the old address points to an invalid heap block.

### Observations
The CFData instance is in complete working order.

```
0:029> dx -r1 ((CoreFoundation!__CFData *)0xaafb0c8)
((CoreFoundation!__CFData *)0xaafb0c8)                 : 0xaafb0c8 : { binary data, 32464 bytes }
    [<Raw View>]     [Type: __CFData]
    [size]           : 32464 [Type: long]
    [mutable]        : yes
    [Raw Bytes]      : 0x7b <-- Valid first byte for this data
```

If we look at its raw view, `data->_bytes` is:

```
    [+0x014] _bytes           : 0xb184050 : 0x7b [Type: unsigned char *]
```

The code in `__CFGrowData`:

```c
    void *oldBytes = data->_bytes;
...
            bytes = realloc(oldBytes, numBytes * sizeof(uint8_t));
```

**oldBytes** is never reassigned.

_However_: We are seeing this:

```
0ae7f6b4          void * oldBytes = 0x0aace028
```

oldBytes no longer matches `data->_bytes`!

DHowett-MSFT:
### Running theory

CFData isn't safe for cross-thread manipulation. This could explain why the captured `_bytes` (`oldBytes`) and the current `_bytes` differ.

rajsesh-msft:
closing this issue since investigation is complete - fixed through #2757

PR #2757, opened 2017-06-03
Force the NSURLConnection delegate queue to be serial

DHowett-MSFT:
This is the root cause of #2756 and the test fixture hang.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2757)


DHowett-MSFT:
Ah, this is in the _tests_

PR #2761, opened 2017-06-07
Make the NSURLSession tests a bit more reliable

DHowett-MSFT:
This pull request:

* Switches NSURLSessionTests to use THBooleanCondition, _which can only trigger once_
* Makes the tests await only a single broadcast from THBooleanCondition for all delegate callbacks
* Removes the delegate callback triggers that do not signal completion
* Rewrites the download/cancel/resume test to not rely on nebulous magic
  * Previously, the test was waiting to receive one callback (out-of-band), then cancelling the task. There
    was a window where the task was still running. The new implementation cancels it immediately in the
    "received data" callback.
  * The test was leaving a 500MB download spinning in the background instead of also cancelling that one
  * The test now makes certain that it has received SOME data before it cancels the task so there is
    definitely something to resume.
* Synchronizes all access to the test helper properties
* Logs which delegate callbacks we've received for easier debugging
* Doesn't try to log an NSDictionary with the `%d` format (:frowning_face:)
* Implements `-[THBooleanCondition signal]`

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2761)

### DIFF
--- a/tests/frameworks/include/Helpers/TestHelpers.h
+++ b/tests/frameworks/include/Helpers/TestHelpers.h
@@ -29,6 +29,7 @@
 @interface THBooleanCondition : NSObject
 - (BOOL)waitUntilDate:(NSDate*)limit;
 - (void)broadcast;
+- (void)signal;
 @property (readonly) NSCondition* condition;
 @property (readonly) bool isOpen;
 @end
--- a/tests/functionaltests/Tests/NSURLSession.mm
+++ b/tests/functionaltests/Tests/NSURLSession.mm
@@ -14,9 +14,10 @@
 //
 //******************************************************************************
 
-#include <TestFramework.h>
-#include <Foundation/Foundation.h>
-#include <cmath>
+#import <TestFramework.h>
+#import <Helpers/TestHelpers.h>
+#import <Foundation/Foundation.h>
+#import <cmath>
 
 static const NSTimeInterval c_testDownloadTimeoutInSec = 30;
 
@@ -39,12 +40,12 @@ typedef NS_ENUM(NSInteger, NSURLSessionAllDelegateType) {
     NSOperationQueue* _delegateQueue;
 }
 
-@property NSCondition* condition;
-@property NSMutableArray* delegateCallOrder;
-@property (copy) NSURLRequest* redirectionRequest;
-@property (copy) NSURLResponse* response;
-@property (copy) NSData* data;
-@property (copy) NSError* completedWithError;
+@property (nonatomic) THBooleanCondition* condition;
+@property (nonatomic) NSMutableArray* delegateCallOrder;
+@property (nonatomic, copy) NSURLRequest* redirectionRequest;
+@property (nonatomic, copy) NSURLResponse* response;
+@property (nonatomic, copy) NSData* data;
+@property (nonatomic, copy) NSError* completedWithError;
 
 - (instancetype)init;
 - (NSURLSession*)createSession;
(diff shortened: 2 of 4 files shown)


DHowett-MSFT:
I have to compose the message in Notepad because the browser is generating spurious form submit events.

DHowett-MSFT:
This pull request:

* Switches NSURLSessionTests to use THBooleanCondition, _which can only trigger once_
* Makes the tests await only a single broadcast from THBooleanCondition for all delegate callbacks
* Removes the delegate callback triggers that do not signal completion
* Rewrites the download/cancel/resume test to not rely on nebulous magic
  * Previously, the test was waiting to receive one callback (out-of-band), then cancelling the task. There
    was a window where the task was still running. The new implementation cancels it immediately in the
    "received data" callback.
  * The test was leaving a 500MB download spinning in the background instead of also cancelling that one
  * The test now makes certain that it has received SOME data before it cancels the task so there is
    definitely something to resume.
* Synchronizes all access to the test helper properties
* Logs which delegate callbacks we've received for easier debugging
* Doesn't try to log an NSDictionary with the `%d` format (:frowning_face:)
* Implements `-[THBooleanCondition signal]`

tests/functionaltests/Tests/NSURLSession.mm
```diff
@@ -68,6 +69,48 @@ - (NSURLSession*)createSession {
 }
 
 //
+// Property getters
+//
+// Note: Since we're using @synchronized(self), the atomic annotation isn't strong enough for
```
ms-jihua: come again?

tests/frameworks/include/Helpers/TestHelpers.h
```diff
@@ -29,6 +29,7 @@
 @interface THBooleanCondition : NSObject
 - (BOOL)waitUntilDate:(NSDate*)limit;
 - (void)broadcast;
+- (void)signal;
```
ms-jihua: any particular reason to prefer signal over broadcast here?

tests/functionaltests/Tests/NSURLSession.mm
```diff
@@ -68,6 +69,48 @@ - (NSURLSession*)createSession {
...
+// guaranteeing nonconcurrent access.
```
rajsesh-msft: nit: this guarantees that concurrent access works (through locking).

tests/functionaltests/Tests/NSURLSession.mm
```diff
@@ -178,55 +218,96 @@ - (NSURLSession*)createSession {
...
-    [self.condition lock];
-    dispatch_sync(_queue, ^{
-        [self.delegateCallOrder addObject:[NSNumber numberWithInteger:NSURLSessionDelegateWillPerformHTTPRedirection]];
-    });
-    self.redirectionRequest = request;
-    completionHandler(request);
+    didReceiveChallenge:(NSURLAuthenticationChallenge*)challenge
+      completionHandler:(void (^)(NSURLSessionAuthChallengeDisposition disposition, NSURLCredential* credential))completionHandler {
+    // TODO::
+    // todo-nithishm-03312016 - Have tests for these.
```
rajsesh-msft: just kill these todo's.

tests/functionaltests/Tests/NSURLSession.mm
DHowett-MSFT: @ms-jihua: `@synchronized(self)` locks around self; an atomic property locks only around `&_ivar`. We'd like to guard the mutation of the members _as well as_ their access.

DHowett-MSFT: Incidentally, a mutable array property returned through synchronized locking and then queried is not itself concurrency-safe, even if the mutation is performed synchronized as well... however, since we're guaranteeing that mutation _should be_ done (with the boolean condition), it's "safer" than what we had before.

tests/frameworks/include/Helpers/TestHelpers.h
DHowett-MSFT: This conforms to the original implementation of NSURLSession tests... and it more completely fills out `THBooleanCondition`'s implementation of `NSCondition`

DHowett-MSFT:
When I was just changing comments, I was going to say "this doesn't need a rebuild."... however, now that I've gone and changed some code, a-rebuilding it'll go

PR #2763, opened 2017-06-07
Do not leave around immutable file

aballway:
Will cause GetSetAttributes to fail the second time it is run

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2763)

### DIFF
--- a/tests/unittests/Foundation/NSFileManagerTests.mm
+++ b/tests/unittests/Foundation/NSFileManagerTests.mm
@@ -210,17 +210,14 @@ TEST(NSFileManager, CopyFileViaURL) {
 }
 
 TEST(NSFileManager, GetSetAttributes) {
-    NSString* srcName = @"GETsetATTRIBUTES.txt";
+    NSString* srcName = @"attributestestfile.txt";
     SCOPE_DELETE_FILE(srcName);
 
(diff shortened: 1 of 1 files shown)


tests/unittests/Foundation/NSFileManagerTests.mm
```diff
@@ -219,7 +219,7 @@
 
     NSDictionary* attributes = @{
         NSFileType : NSFileTypeRegular,
-        NSFileImmutable : @YES,
```
DHowett-MSFT: This test becomes a no-op if you remove this. Specifically, the default state of a "regular file" is `regular file` + `!immutable`.

DHowett-MSFT: You should set it to immutable, and then set it to mutable... or test a different flag entirely.

DHowett-MSFT: Also, setting the "file type" of a file after its creation simply shouldn't be possible o_O

DHowett-MSFT:
nit: name does not fit into the sentence `when merged, this commit will ___`

DHowett-MSFT:
nit: please make sure you rename before merging ;)

PR #2766, opened 2017-06-08
Allow nuget to look at the locally built packages in build\outputpackages

msft-Jeyaram:
This is already part of my CodeReview for projections, to get the projections build with the flags in our lookup url, I need this change in.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2766)


msft-Jeyaram:
@DHowett-MSFT this is needed since build.sln never used the packages it produced. 
Given projections binaries were already checked in, it never had to refer to a nuget package.
Now that we won't have those binaries in place anymore, we need build.sln to look at the build\outputpackages to see if there are any nuget packages that it can use

PR #2773, opened 2017-07-07
Fix some NSArray/NSMutableArray issues

### LINKED ISSUE #2771: [NSMutableArray removeObject:] does not remove all occurrences of an object in the array.
`[NSMutableArray removeObject:]` currently only removes the first occurrence of an object instance in the array. As per the reference platform, this method should remove all occurrences of an object instance in the array.

DHowett-MSFT:
* `removeObject:inRange:` was originally finding the first instance of the
  object in the range, and then doing an array-wide search-and-remove
  every time it found one. This was very clearly wrong.
  This method was also augmented to do batch removals where possible.
* `removeObject:` is supposed to remove every instance of the found object.
  It wasn't doing so. Fixes #2771.
* `removeObjectIdenticalTo:` is supposed to remove every instance of the
  found object. It also wasn't doing so.
* `removeObjectIdenticalTo:inRange:` wasn't implemented.
* `exchangeObjectAtIndex:withObjectAtIndex:` was not using its CF
  optimized implementation.
* Methods that mutated ranges were not bounds-checking those ranges.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2773)

### DIFF
--- a/Frameworks/Foundation/NSCFArray.mm
+++ b/Frameworks/Foundation/NSCFArray.mm
@@ -17,6 +17,7 @@
 #include "Starboard.h"
 #include "CFHelpers.h"
 #include "CFFoundationInternal.h"
+#include "ForFoundationOnly.h"
 #include <CoreFoundation/CFArray.h>
 #include "NSCFArray.h"
 #include "BridgeHelpers.h"
@@ -84,7 +85,9 @@ BRIDGED_MUTABLE_CLASS_FOR_CODER(CFArrayRef, _CFArrayIsMutable, NSArray, NSMutabl
 - (id)objectAtIndex:(NSUInteger)index {
     if (index >= CFArrayGetCount((CFArrayRef)self)) {
         [NSException raise:@"Array out of bounds"
-                    format:@"objectAtIndex: index > count (%lu > %lu), throwing exception\n", (unsigned long)index, (unsigned long)CFArrayGetCount((CFArrayRef)self)];
+                    format:@"objectAtIndex: index > count (%lu > %lu), throwing exception\n",
+                           (unsigned long)index,
+                           (unsigned long)CFArrayGetCount((CFArrayRef)self)];
         return nil;
     }
     return (id)CFArrayGetValueAtIndex((CFArrayRef)self, index);
@@ -111,13 +114,18 @@ BRIDGED_MUTABLE_CLASS_FOR_CODER(CFArrayRef, _CFArrayIsMutable, NSArray, NSMutabl
     CFRange range;
     range.location = index;
     range.length = 1;
-    CFArrayReplaceValues(static_cast<CFMutableArrayRef>(self), range, (const void**)(&obj), 1);
+    _CFArrayReplaceValues(static_cast<CFMutableArrayRef>(self), range, (const void**)(&obj), 1);
 }
 
 - (void)insertObject:(NSObject*)objAddr atIndex:(NSUInteger)index {
     BRIDGED_THROW_IF_IMMUTABLE(_CFArrayIsMutable, CFArrayRef);
     NS_COLLECTION_THROW_IF_NULL_REASON(objAddr, [NSString stringWithFormat:@"*** %@ object cannot be nil", NSStringFromSelector(_cmd)]);
     CFArrayInsertValueAtIndex(static_cast<CFMutableArrayRef>(self), index, reinterpret_cast<const void*>(objAddr));
+}
+
+- (void)exchangeObjectAtIndex:(NSUInteger)atIndex withObjectAtIndex:(NSUInteger)withIndex {
+    BRIDGED_THROW_IF_IMMUTABLE(_CFArrayIsMutable, CFArrayRef);
+    CFArrayExchangeValuesAtIndices(static_cast<CFMutableArrayRef>(self), atIndex, withIndex);
 }
 
 - (void)removeAllObjects {
--- a/Frameworks/Foundation/NSCFCollectionSupport.h
+++ b/Frameworks/Foundation/NSCFCollectionSupport.h
@@ -38,4 +38,16 @@ CFHashCode _NSCFCallbackHash(const void* value);
         }                                                                                                        \
     } while (false)
 
+#define NS_COLLECTION_VALIDATE_RANGE(range, count)                                                 \
+    do {                                                                                           \
+        if (range.location + range.length > (count)) {                                             \
+            [NSException raise:NSInvalidArgumentException                                          \
+                        format:@"*** %s: Range {%lu, %lu} out of bounds; legal range is {0, %lu}", \
+                               __PRETTY_FUNCTION__,                                                \
+                               (unsigned long)range.location,                                      \
+                               (unsigned long)range.length,                                        \
+                               (unsigned long)(count)];                                            \
+        }                                                                                          \
+    } while (false)
+
 #endif // __OBJC__
--- a/Frameworks/Foundation/NSMutableArray.mm
+++ b/Frameworks/Foundation/NSMutableArray.mm
@@ -22,6 +22,7 @@
 #import "CFHelpers.h"
 #import "NSRaise.h"
 #import "NSCFArray.h"
+#import "NSCFCollectionSupport.h"
 #import "BridgeHelpers.h"
 
 #include <algorithm>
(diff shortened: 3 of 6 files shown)


Frameworks/Foundation/NSMutableArray.mm
```diff
@@ -157,46 +158,59 @@ - (void)exchangeObjectAtIndex:(NSUInteger)atIndex withObjectAtIndex:(NSUInteger)
...
  @Status Interoperable
 */
-- (void)removeObjectIdenticalTo:(NSObject*)objAddr {
-    int idx = [self indexOfObjectIdenticalTo:objAddr];
-    if (idx != NSNotFound) {
-        [self removeObjectAtIndex:idx];
+- (void)removeObjectsInRange:(NSRange)range {
+    NS_COLLECTION_VALIDATE_RANGE(range, self.count);
+
+    for (NSUInteger i = range.location + range.length; i > range.location; i--) {
```
ms-jihua: nit: --i

Frameworks/Foundation/NSCFCollectionSupport.h
```diff
@@ -38,4 +38,16 @@ CFHashCode _NSCFCallbackHash(const void* value);
         }                                                                                                        \
     } while (false)
 
+#define NS_COLLECTION_VALIDATE_RANGE(range, count)                                                 \
+    do {                                                                                           \
+        if (range.location + range.length > (count)) {                                             \
```
aballway: nit: save `count` to a local so it won't be expanded twice

Frameworks/Foundation/NSCFCollectionSupport.h
```diff
@@ -38,4 +38,16 @@ CFHashCode _NSCFCallbackHash(const void* value);
...
+            [NSException raise:NSInvalidArgumentException                                          \
+                        format:@"*** %s: Range {%lu, %lu} out of bounds; legal range is {0, %lu}", \
+                               __PRETTY_FUNCTION__,                                                \
+                               (unsigned long)range.location,                                      \
```
aballway: nit: no

Frameworks/Foundation/NSCFCollectionSupport.h
```diff
@@ -38,4 +38,16 @@ CFHashCode _NSCFCallbackHash(const void* value);
...
+                               (unsigned long)range.length,                                        \
```
aballway: nit: no

Frameworks/Foundation/NSCFCollectionSupport.h
```diff
@@ -38,4 +38,16 @@ CFHashCode _NSCFCallbackHash(const void* value);
...
+                               (unsigned long)(count)];                                            \
```
aballway: nit: no

Frameworks/Foundation/NSCFArray.mm
```diff
@@ -84,7 +85,9 @@ - (NSUInteger)count {
 - (id)objectAtIndex:(NSUInteger)index {
     if (index >= CFArrayGetCount((CFArrayRef)self)) {
         [NSException raise:@"Array out of bounds"
-                    format:@"objectAtIndex: index > count (%lu > %lu), throwing exception\n", (unsigned long)index, (unsigned long)CFArrayGetCount((CFArrayRef)self)];
+                    format:@"objectAtIndex: index > count (%lu > %lu), throwing exception\n",
+                           (unsigned long)index,
```
aballway: nit: no

Frameworks/Foundation/NSCFArray.mm
```diff
@@ -84,7 +85,9 @@ - (NSUInteger)count {
...
+                           (unsigned long)CFArrayGetCount((CFArrayRef)self)];
```
aballway: nit: no

Frameworks/Foundation/NSMutableArray.mm
```diff
@@ -157,46 +158,59 @@ - (void)exchangeObjectAtIndex:(NSUInteger)atIndex withObjectAtIndex:(NSUInteger)
...
-        id curObj = [self objectAtIndex:i];
-
-        if ([curObj isEqual:objAddr]) {
-            [self removeObject:curObj];
+- (void)removeObject:(NSObject*)object inRange:(NSRange)range {
+    NS_COLLECTION_VALIDATE_RANGE(range, self.count);
+    // NSIndexSet can coalesce adjacent ranges. Combined with an efficient removeObjectsInRange:
+    // this can remove large swaths of objects more efficiently than individual mutation.
+    NSMutableIndexSet* indexes = [NSMutableIndexSet indexSet];
+    NSUInteger max = std::min(self.count, range.location + range.length);
```
aballway: nit: range is already validated, so this is unnecessary

Frameworks/Foundation/NSMutableArray.mm
```diff
@@ -157,46 +158,59 @@ - (void)exchangeObjectAtIndex:(NSUInteger)atIndex withObjectAtIndex:(NSUInteger)
...
 */
-- (void)removeObject:(NSObject*)objAddr inRange:(NSRange)range {
-    for (int i = range.location + range.length - 1; i >= (int)range.location; i--) {
-        id curObj = [self objectAtIndex:i];
-
-        if ([curObj isEqual:objAddr]) {
-            [self removeObject:curObj];
+- (void)removeObject:(NSObject*)object inRange:(NSRange)range {
+    NS_COLLECTION_VALIDATE_RANGE(range, self.count);
+    // NSIndexSet can coalesce adjacent ranges. Combined with an efficient removeObjectsInRange:
```
aballway: removeObjectsInRange doesn't seem especially optimized. Would it be better to just remove them directly here rather than making an IndexSet, iterating and adding indexes to said index set, then iterating over the index set to remove the objects?

tests/unittests/Foundation/NSArrayTests.mm
```diff
@@ -726,3 +727,74 @@ static unsigned long int objectIndexInArray(NSArray* array, int value, int start
     indexes = [NSIndexSet indexSetWithIndexesInRange:NSMakeRange(11, 11)];
     EXPECT_ANY_THROW([arr enumerateObjectsAtIndexes:indexes options:0 usingBlock:block]);
 }
+
```
aballway: just for sanity sake, can you add a test with a range w/ 0 length?

Frameworks/Foundation/NSMutableArray.mm
```diff
@@ -157,46 +158,59 @@ - (void)exchangeObjectAtIndex:(NSUInteger)atIndex withObjectAtIndex:(NSUInteger)
...
+    [self removeObjectIdenticalTo:object inRange:NSRange{ 0, self.count }];
+}
+
+/**
+ @Status Interoperable
+*/
+- (void)removeObjectIdenticalTo:(id)object inRange:(NSRange)range {
+    NS_COLLECTION_VALIDATE_RANGE(range, self.count);
+    NSMutableIndexSet* indexes = [NSMutableIndexSet indexSet];
+    NSUInteger max = std::min(self.count, range.location + range.length);
```
aballway: same as above

DHowett-MSFT:
@aballway updated to address concerns

Frameworks/Foundation/NSMutableArray.mm
```diff
@@ -154,49 +155,53 @@ - (void)exchangeObjectAtIndex:(NSUInteger)atIndex withObjectAtIndex:(NSUInteger)
...
 
 /**
  @Status Interoperable
 */
-- (void)removeObjectsInRange:(NSRange)range {
-    for (int i = range.location + range.length - 1; i >= (int)range.location; i--) {
-        [self removeObjectAtIndex:i];
-    }
+- (void)removeObjectIdenticalTo:(id)object inRange:(NSRange)range {
+    [self _removeObject:object inRange:range withComparator:[](id left, id right) -> bool { return right == left; }];
```
aballway: nit: is the `-> bool` necessary?

Frameworks/Foundation/NSMutableArray.mm
```diff
@@ -154,49 +155,53 @@ - (void)exchangeObjectAtIndex:(NSUInteger)atIndex withObjectAtIndex:(NSUInteger)
...
 }
 
-/**
- @Status Interoperable
-*/
-- (void)removeObject:(NSObject*)objAddr {
-    if (objAddr == nil) {
-        TraceVerbose(TAG, L"objAddr = nil!");
-    }
+using _NSObjectComparator = bool (*)(id, id);
```
aballway: nit: this is only used once right? maybe just use directly.

Frameworks/Foundation/NSCFArray.mm
aballway: would still prefer these changed <3

Frameworks/Foundation/NSMutableArray.mm
DHowett-MSFT: This seems cleaner.

Frameworks/Foundation/NSMutableArray.mm
DHowett-MSFT: Alas, yes. `[x isEqual:y]` returns BOOL. Functions that return one/the other aren't implicitly convertible.

PR #2778, opened 2017-07-20
Fix endless loop if index set is nil

xFAR:
Raises an invalid arg exception per Apple implementation and docs.

### DIFF
--- a/Frameworks/Foundation/NSArray.mm
+++ b/Frameworks/Foundation/NSArray.mm
@@ -818,14 +818,21 @@ static CFComparisonResult _CFComparatorFunctionFromComparator(const void* val1, 
  @Status Interoperable
 */
 - (NSArray*)objectsAtIndexes:(NSIndexSet*)indexes {
-    unsigned idx = [indexes firstIndex];
-    id ret = [NSMutableArray array];
-    unsigned count = [self count];
+    NSUInteger idx = NSNotFound;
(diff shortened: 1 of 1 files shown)


xFAR:
Dustin,

My responses inline.



On Jul 21, 2017, at 1:08 AM, Dustin L. Howett (MSFT) <notifications@github.com> wrote:

> @DHowett-MSFT requested changes on this pull request.
> 
> Hi @xFAR! Thanks for the contribution.
> 
> I'd like to accept this pull request, but I've got a couple minor issues:
> 
> Our coding convention requires four spaces be used per indentation level than one tab
I will update the style.  I use tabs equal to four spaces so I did not notice.

> We have a macro (admittedly difficult to discover!) called NS_COLLECTION_THROW_IF_NULL_REASON; you can place it at the top of this function and keep the rest of the code the same.
> NS_COLLECTION_THROW_IF_NULL_REASON(indexes, @"indexes must not be nil”);

Ok.

> Would you mind writing tests (in NSArrayTests.mm) for this behaviour as well? You can use the other tests as a template.
> 
Hopefully, I can get to this later in the week.

> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.
> 



Regards,

Felipe A. Rodriguez


illumenos LLC
________________________

PR #2786, opened 2017-08-18
ObjC2Winmd Tool: Tool to convert ObjC middleware to WinRT component

mukhole:
This tool generates WinRT component for an ObjC middleware.
The tool auto-generates WinRT wrappers and glue/marshalling code for type marshalling to and from ObjC and WinRT.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2786)


mukhole:
@DHowett-MSFT, Yes the binary and the required DLL gets packaged to WinObjC-tools package and gets installed with it.

PR #2803, opened 2017-10-20
Update Expired App Certificates

MSFTFox:
Update the app certificates in test/testapps/ since they have expired.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2803)


MSFTFox:
I'll revert the project files affected by updating the certificates. These files shouldn't be changed.

PR #2816, opened 2017-11-15
Implement new libdispatch APIs

MSFTFox:
libdispatch has been updated to support the following APIs

dispatch_queue_get_specific
dispatch_get_specific
dispatch_queue_set_specific

Additionally, LibDispatch now properly exports various dispatch barrier functions.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2816)


deps/3rdparty/libdispatch/dispatch/dispatch.h
```diff
@@ -18,6 +18,7 @@
  * @APPLE_APACHE_LICENSE_HEADER_END@
  */
 
+
```
DHowett-MSFT: Nit: we should reduce unnecessary changes to vendored headers

deps/3rdparty/libdispatch/src/interop.c
```diff
@@ -42,6 +42,7 @@ struct dispatch_queue_s _dispatch_main_q = {
 	/*.dq_items_head           = */	0,
 	/*.dq_serialnum            = */	1,
 	/*.dq_finalizer_ctxt       = */	0,
+    /*.dq_specific_q           = */NULL,
```
DHowett-MSFT: Nit: it looks like dispatch uses tabs, not spaces. Please follow the existing format

DHowett-MSFT:
(ideally, it'd be three commits:

* **abcdefgh** [libdispatch] add support for queue specifics and export barrier
* **ghijklmn** Update libdispatch prebuilts for **abcdefgh**
* Add unit tests for the new dispatch APIs
)

deps/3rdparty/libdispatch/src/interop.c
MSFTFox: Yup, clang format certainly did a number on these files.

MSFTFox:
The commit has been broken out into 3 different commits as @DHowett-MSFT suggested. Additionally, I removed the clang formatting from each of the files affected.

DHowett-MSFT:
nit: _please_ change every instance of `LibDispatch` to `libdispatch` or `dispatch` -- it's never capitalized as such

DHowett-MSFT:
(I'd prefer the commit titles to be indicative of what they actually did.. like, `implement the libdispatch queue_specific APIs` and stuff)`

deps/3rdparty/libdispatch/dispatch/queue.h
```diff
@@ -180,6 +180,133 @@ dispatch_async_f(dispatch_queue_t queue,
...
+ * parameter passed to this function is the context provided to
+ * dispatch_barrier_async_f().
+ * The result of passing NULL in this parameter is undefined.
+ */
+
+__OSX_AVAILABLE_STARTING(__MAC_10_6, __IPHONE_4_0)
+DISPATCH_EXPORT DISPATCH_NONNULL1 DISPATCH_NONNULL3 DISPATCH_NOTHROW
+void
+dispatch_barrier_async_f(dispatch_queue_t queue,
+    void *_Nullable context,
```
DHowett-MSFT: If this is indented with spaces, it should be indented with tabs to match the libdispatch coding style.

deps/3rdparty/libdispatch/src/queue.c
```diff
@@ -675,6 +696,9 @@ _dispatch_queue_dispose(dispatch_queue_t dq)
 	if (slowpath(dq->dq_items_tail)) {
 		DISPATCH_CRASH("Release of a queue while items are enqueued");
 	}
+	if (slowpath(dq->dq_items_tail)) {
+		DISPATCH_CRASH("Release of a queue while items are enqueued");
```
DHowett-MSFT: it looks like this code's just been copied

deps/3rdparty/libdispatch/src/queue.c
```diff
@@ -1978,3 +2002,68 @@ dispatch_after_f(dispatch_time_t when, dispatch_queue_t queue, void *ctxt, void
...
+	dispatch_queue_specific_list_t dqsl = dq->dq_specific_q;
+	dispatch_queue_specific_t var;
+
+	TAILQ_FOREACH(var, &dqsl->contextList, specific) {
+		if (var->key == dqs->key) {
+			if(var->destructor) {
+				var->destructor(var->context);
+			}
+
+			// If a context exists for this key, simply update the context on that specific
```
DHowett-MSFT: nit: `if a specific exists for this key`

deps/3rdparty/libdispatch/src/queue.c
```diff
@@ -1978,3 +2002,68 @@ dispatch_after_f(dispatch_time_t when, dispatch_queue_t queue, void *ctxt, void
...
+			} else {
+				TAILQ_REMOVE(&dqsl->contextList, var, specific);
+				free(var);
+			}
+			free(dqs);
+			return;
+		}
+	}
+
+	TAILQ_INSERT_TAIL(&dqsl->contextList, dqs, specific);
```
DHowett-MSFT: Concern: On line 2027 you remove an *existing* specific if you set its value to null.

Here, you will insert a *new* specific _even if_ its value is null.

deps/3rdparty/libdispatch/src/queue.c
```diff
@@ -1978,3 +2002,68 @@ dispatch_after_f(dispatch_time_t when, dispatch_queue_t queue, void *ctxt, void
...
+}
+
+void dispatch_queue_set_specific(dispatch_queue_t queue, const void *key, void *_Nullable context, dispatch_function_t _Nullable destructor) {
+	if (slowpath(!key)) {
```
DHowett-MSFT: do we need to be wary of locking and threading here?

deps/3rdparty/libdispatch/src/queue.c
```diff
@@ -1978,3 +2002,68 @@ dispatch_after_f(dispatch_time_t when, dispatch_queue_t queue, void *ctxt, void
...
+			}
+			free(dqs);
+			return;
+		}
+	}
+
+	TAILQ_INSERT_TAIL(&dqsl->contextList, dqs, specific);
+}
+
+void dispatch_queue_set_specific(dispatch_queue_t queue, const void *key, void *_Nullable context, dispatch_function_t _Nullable destructor) {
```
DHowett-MSFT: nit: looks like dispatch uses `ctxt` for context; care to match?

deps/3rdparty/libdispatch/src/queue.c
```diff
@@ -1978,3 +2002,68 @@ dispatch_after_f(dispatch_time_t when, dispatch_queue_t queue, void *ctxt, void
 	dispatch_resume(as_do(ds));
 }
 
+static void _dispatch_queue_init_specific_list(dispatch_queue_t dq) {
+	dispatch_queue_specific_list_t dqsq;
```
DHowett-MSFT: nit: this seems more like a `dqsl`

deps/3rdparty/libdispatch/src/queue_kevent.c
```diff
@@ -512,6 +512,7 @@ struct dispatch_queue_s _dispatch_mgr_q = {
 	/*.dq_items_head           = */	0,
 	/*.dq_serialnum            = */	2,
 	/*.dq_finalizer_ctxt       = */	0,
+	/*.dq_specific_q           = */NULL,
```
DHowett-MSFT: nit: missing space before `NULL`

msvc/sdk-build.props
```diff
@@ -97,7 +97,7 @@
       <SystemHeaderDeps>true</SystemHeaderDeps>
       <AdditionalOptions>-Werror -Wno-microsoft --system-header-prefix=winrt/ %(AdditionalOptions)</AdditionalOptions>
       <PreprocessorDefinitions>_WINOBJC_DO_NOT_USE_NSLOG=;%(PreprocessorDefinitions)</PreprocessorDefinitions>
-      <InternalSystemIncludePaths>%(InternalSystemIncludePaths);$(IncludePath);$(MSBuildThisFileDirectory)..\include\;$(MSBuildThisFileDirectory)..\include\xplat;$(MSBuildThisFileDirectory)..\deps\prebuilt\include\;$(MSBuildThisFileDirectory)..\deps\prebuilt\include\icu;$([System.IO.Path]::Combine('$(MSBuildThisFileDirectory)..\include\Platform', '$(TargetOsAndVersion)'))</InternalSystemIncludePaths>
+      <InternalSystemIncludePaths>%(InternalSystemIncludePaths);$(IncludePath);$(MSBuildThisFileDirectory)..\include\;$(MSBuildThisFileDirectory)..\include\xplat;$(MSBuildThisFileDirectory)..\deps\prebuilt\include\;$(MSBuildThisFileDirectory)..\tools\deps\prebuilt\include\;$(MSBuildThisFileDirectory)..\deps\prebuilt\include\icu;$([System.IO.Path]::Combine('$(MSBuildThisFileDirectory)..\include\Platform', '$(TargetOsAndVersion)'))</InternalSystemIncludePaths>
```
DHowett-MSFT: 'scuse me?

tests/UnitTests/Foundation/DispatchTests.mm
```diff
@@ -0,0 +1,122 @@
...
+        dispatch_queue_set_specific(queue, key, context2, NULL);
+    });
+	dispatch_group_wait(group, DISPATCH_TIME_FOREVER);
+    
+	EXPECT_EQ(dispatch_queue_get_specific(queue, key), context2);
+    
+    dispatch_group_async(group, queue, ^{
+		EXPECT_EQ(dispatch_get_specific(key), context2);
+    });
+	dispatch_group_wait(group, DISPATCH_TIME_FOREVER);
```
DHowett-MSFT: `FOREVER` is not likely to be correct, just on principle

DHowett-MSFT: you need to release the group and the queue when you're done.

DHowett-MSFT: (otherwise you're not testing the queue -> specific_list dealloc codepath)

tests/UnitTests/Foundation/DispatchTests.mm
```diff
@@ -0,0 +1,122 @@
...
+    dispatch_group_async(group, queue, ^{
+		EXPECT_EQ(dispatch_get_specific(key), context);
+        dispatch_queue_set_specific(queue, key, NULL, NULL);
+    });
+	dispatch_group_wait(group, DISPATCH_TIME_FOREVER);
+
+	if(dispatch_queue_get_specific(queue, key)) {
+		EXPECT_EQ(1, 0);
+	}
+}
```
DHowett-MSFT: I'd like to see a test that involves multiple queues, just to make sure that we never get cross-contamination.

tools/deps/prebuilt/include/dispatch/continuation_cache.h
```diff
@@ -4,38 +4,44 @@
 // If dc_vtable is less than 127, then the object is a continuation.
 // Otherwise, the object has a private layout and memory management rules. The
 // first two words must align with normal objects.
-#define DISPATCH_CONTINUATION_HEADER(x) \
-    const void* do_vtable;              \
-    struct x* volatile do_next;         \
-    dispatch_function_t dc_func;        \
-    void* dc_ctxt
+#define DISPATCH_CONTINUATION_HEADER(x)	\
```
DHowett-MSFT: this file still got reformatted.

tools/deps/prebuilt/include/dispatch/debug.h
```diff
@@ -3,173 +3,160 @@
 
 #if __GNUC__
 void _dispatch_bug(size_t line, long val) __attribute__((__noinline__));
-void _dispatch_abort(size_t line, long val) __attribute__((__noinline__, __noreturn__));
-void _dispatch_log(const char* msg, ...) __attribute__((__noinline__, __format__(printf, 1, 2)));
-void _dispatch_logv(const char* msg, va_list) __attribute__((__noinline__, __format__(printf, 1, 0)));
+void _dispatch_abort(size_t line, long val) __attribute__((__noinline__,__noreturn__));
```
DHowett-MSFT: reformatted

msvc/sdk-build.props
MSFTFox: I removed the duplicate headers that were copied into deps\prebuilt\include\libdispatch. It doesn't make sense to have two additional copies of every header for libdispatch checked in. Tools uses them from the tools\deps\prebuilt\include folder.

deps/3rdparty/libdispatch/src/queue.c
MSFTFox: Good catch. There shouldn't be a specific with a null context.

deps/3rdparty/libdispatch/src/queue.c
MSFTFox: Woops... I was copying from a backup of the formatted file and copied the wrong lines.

tools/deps/prebuilt/include/dispatch/continuation_cache.h
MSFTFox: Believe it or not, the file was formatted incorrectly before. The file in deps\3rdparty\libdispatch\ is formatted to this. My assumption is that this file should be the most accurate since the one checked into tools\deps\prebuilt\include must be a copy. Copying over this file is why it changed and will happen each time libdispatch is built. This formatting most closely resembles the original fork of dr pizza's libdispatch repo.

Should I apply the (clearly better looking) formatting that was applied to the copy of this header to the original header?

MSFTFox: As is the same for each "formatted" file in the tools folder.

DHowett-MSFT: Ah, I see. Then this is a change we have to take.

Make sure it’s part of the “update prebuilts” commit :)

deps/3rdparty/libdispatch/src/queue.c
MSFTFox: Fixed the locking issues. A dispatch_queue_t has been added to a dispatch_queue_t with delayed initialization. The specific list is stored on this embedded queue such that this queue can be run/locked on to perform operations on the specific list. Using the barrier functions, it's guaranteed that a single operation finishes before another is performed on that particular queue.

DHowett-MSFT:
before merging, please split "address thread safety" into the dispatch, prebuilts and tests.

MSFTFox:
Mmk, all issues should have been addressed.

deps/3rdparty/libdispatch/src/queue.c
```diff
@@ -665,6 +679,27 @@ dispatch_queue_create(const char *label, dispatch_queue_attr_t attr)
 #endif
 }
 
+void _dispatch_queue_specific_release(dispatch_queue_t dq) {
+	dispatch_queue_specific_list_t dqsl = dq->dq_specific_q;
+	dispatch_queue_specific_t lastvar = NULL;
+	dispatch_queue_specific_t var;
+
+	TAILQ_FOREACH(var, &dqsl->contextList, specific) {
+		if (lastvar) {
```
DHowett-MSFT: You need to do this because var contains `var->next`, right?
There may be another TAILQ macro that handles this, but perhaps not.

deps/3rdparty/libdispatch/src/queue.c
```diff
@@ -682,6 +717,13 @@ _dispatch_queue_dispose(dispatch_queue_t dq)
 	}
 #endif
 
+	if (slowpath(dq->dq_specific_q)) {
+		_dispatch_queue_dispose(dq_specific_q);
```
DHowett-MSFT: I can't figure out how this compiles. `dq_specific_q` is a member of `dq`, but you are lacking a `->` here

deps/3rdparty/libdispatch/src/queue.c
```diff
@@ -682,6 +717,13 @@ _dispatch_queue_dispose(dispatch_queue_t dq)
...
+	}
+	if (slowpath(dq->dq_specific_list)) {
+		_dispatch_queue_specific_release(dq);
```
DHowett-MSFT: nit: this function should take a `dq_specific_list`, it's more of a `dispatch_queue_specific_list_release`

deps/3rdparty/libdispatch/src/queue.c
```diff
@@ -1978,3 +2020,88 @@ dispatch_after_f(dispatch_time_t when, dispatch_queue_t queue, void *ctxt, void
...
+			*context = var->context;
+			return;
+		}
+	}
+	*context = NULL;
+}
+
+void* _Nullable dispatch_queue_get_specific(dispatch_queue_t queue, const void *key) {
+	void* context = NULL;
+	if (dq->dq_specific_q) {
```
DHowett-MSFT: This won't compile, your `dq` is named `queue`..?

deps/3rdparty/libdispatch/src/queue.c
```diff
@@ -1978,3 +2019,92 @@ dispatch_after_f(dispatch_time_t when, dispatch_queue_t queue, void *ctxt, void
...
+		return;
+	}
+	dispatch_queue_specific_t specific = calloc(1, sizeof(struct dispatch_queue_specific_s));
+	specific->key = key;
+	specific->context = context;
+	specific->destructor = destructor;
+
+	// Delayed initialization
+	if (slowpath(!queue->dq_specific_q)) {
+		_dispatch_queue_init_specific_list(queue);
```
DHowett-MSFT: what happens if two callers get _here_ at the same time?

deps/3rdparty/libdispatch/src/queue.c
```diff
@@ -1978,3 +2019,92 @@ dispatch_after_f(dispatch_time_t when, dispatch_queue_t queue, void *ctxt, void
 	dispatch_resume(as_do(ds));
 }
 
+struct specificContainer {
```
DHowett-MSFT: this is unused

deps/3rdparty/libdispatch/src/queue.c
```diff
@@ -1978,3 +2019,92 @@ dispatch_after_f(dispatch_time_t when, dispatch_queue_t queue, void *ctxt, void
...
+			return;
+		}
+	}
+	*ctxt = NULL;
+}
+
+void* _Nullable dispatch_queue_get_specific(dispatch_queue_t queue, const void *key) {
+	void* context = NULL;
+	if (queue->dq_specific_q) {
+		context = key;
```
DHowett-MSFT: clever!

deps/3rdparty/libdispatch/src/queue.c
```diff
@@ -1978,3 +2019,92 @@ dispatch_after_f(dispatch_time_t when, dispatch_queue_t queue, void *ctxt, void
...
+	dispatch_barrier_sync_f(queue->dq_specific_q, specific, _dispatch_queue_insert_specific);
+}
+
+static void _dispatch_queue_get_specific(void* context) {
+	dispatch_queue_t dq = _dispatch_queue_get_current();
+	dispatch_queue_specific_list_t dqsl = dq->dq_specific_list;
+	dispatch_queue_specific_t var;
+
+	// The parameter passed in is reference to a void* key
+	void** ctxt = context;
```
DHowett-MSFT: you can save this temporary by just specifying that `context` is itself a `void**`

MSFTFox: We use it as a void* out pointer though for the return value of the context.

msvc/sdk-build.props
rajsesh-msft: Ideally, we would build this as part tools package and then we don't need to check in the header, but that is for another day.

tools/deps/prebuilt/include/dispatch/queue_internal.h
```diff
@@ -32,45 +32,63 @@
...
+	intptr_t dq_serialnum; \
+	void *dq_finalizer_ctxt; \
+	struct dispatch_queue_s *dq_specific_q; \
+	struct dispatch_queue_specific_list_s *dq_specific_list; \
+	void* dq_manually_drained; \
+	bool dq_is_manually_draining
 #endif
 
+struct dispatch_queue_specific_list_s {
+	TAILQ_HEAD(dispatch_queue_specific_head_s, dispatch_queue_specific_s) contextList;
```
rajsesh-msft: Please stick to the naming conventions in this file, so this would be context_list

tools/deps/prebuilt/include/dispatch/queue_internal.h
```diff
@@ -81,23 +99,25 @@ extern struct dispatch_queue_s _dispatch_root_queues[];
...
-    prev = fastpath(dispatch_atomic_xchg_pointer(&dq->dq_items_tail, tail));
-    if (prev) {
-        // if we crash here with a value less than 0x1000, then we are at a known bug in client code
-        // for example, see _dispatch_queue_dispose or _dispatch_atfork_child
-        prev->do_next = head;
-    } else {
-        _dispatch_queue_push_list_slow(dq, head);
-    }
+static void
+_dispatch_queue_push_list(dispatch_queue_t dq, dispatch_object_t _head, dispatch_object_t _tail)
```
rajsesh-msft: please revert these formatting changes.

tests/UnitTests/Foundation/DispatchTests.mm
```diff
@@ -0,0 +1,200 @@
...
+
+TEST(DispatchTests, SimpleSpecific) {
+    dispatch_queue_t queue = dispatch_queue_create("queue", nullptr);
+	dispatch_group_t group = dispatch_group_create();
+
+	dispatch_queue_set_specific(queue, key, context, nullptr);
+
+	dispatch_group_async(group, queue, ^{
+		EXPECT_EQ(dispatch_get_specific(key), context);
+        dispatch_queue_set_specific(queue, key, context2, nullptr);
```
DHowett-MSFT: The formatting  in this test file seems all sorts of strange. Is that just GitHub acting up?

tests/UnitTests/Foundation/DispatchTests.mm
```diff
@@ -0,0 +1,200 @@
...
+
+    dispatch_queue_set_specific(queue, key, context, nullptr);
+    
+    dispatch_group_async(group, queue, ^{
+		EXPECT_EQ(dispatch_get_specific(key), context);
+        dispatch_queue_set_specific(queue, key, nullptr, nullptr);
+    });
+	ASSERT_EQ(dispatch_group_wait(group, _createEndTestTime()), 0);
+
+	EXPECT_EQ(dispatch_queue_get_specific(queue, key), nullptr);
```
DHowett-MSFT: Minor nit: The EXPECT and ASSERT macros work better if the “expected” value is on the left.

msvc/sdk-build.props
bbowman: +1. I believe prior to this change, The Frameworks build didn't have any tendrils reaching over to tools and just depended on the tools package to keep the layers as separate and clean as possible. It worries me that we need to make this change especially when thinking about applications that need to depend on libdispatch that only use our packages. How will those work? 

I don't quite understand exactly what you are reaching over to include here?

---
In reply to: [151585379](https://github.com/Microsoft/WinObjC/pull/2816#discussion_r151585379) [](ancestors = 151585379)

deps/3rdparty/libdispatch/build/libdispatch.vcxproj
```diff
@@ -249,19 +246,13 @@
     <ClangCompile Include="..\src\shims\mach.c" />
   </ItemGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
-
   <PropertyGroup>
-    <PrebuiltPath Condition="'$(PrebuiltPath)' == ''">$(WINOBJC_SDK_ROOT)\deps\prebuilt\Universal Windows\$(PlatformTarget)\$(Configuration)\\</PrebuiltPath>
+    <PrebuiltRoot Condition="'$(PrebuiltRoot)' == ''">$(WINOBJC_SDK_ROOT)\tools\deps\prebuilt\\</PrebuiltRoot>
```
bbowman: >\\ [](start = 93, length = 2)

double backslash?

deps/3rdparty/libdispatch/build/libdispatch.vcxproj
```diff
@@ -249,19 +246,13 @@
...
+    <PrebuiltPath Condition="'$(PrebuiltPath)' == ''">$(PrebuiltRoot)\Universal Windows\$(PlatformTarget)\$(Configuration)\\</PrebuiltPath>
```
bbowman: >PrebuiltPath [](start = 5, length = 12)

I don't see this used. Whats the deal?

deps/3rdparty/libdispatch/build/libdispatch.vcxproj
bbowman: >tools [](start = 74, length = 5)

I don't understand why this is looking in tools?

deps/3rdparty/libdispatch/build/libdispatch.vcxproj
```diff
@@ -249,19 +246,13 @@
...
   </PropertyGroup>
-  <Target Name="CopyContent" AfterTargets="Build">
+  <Target Name="RemovePackageBinaries" BeforeTargets="Build;Deploy;_CopyFilesMarkedCopyLocal">
     <ItemGroup>
-      <LibraryOutput Include="$(OutDir)\*.lib" />
-      <LibraryOutput Include="$(OutDir)\*.dll" />
-      <LibraryOutput Include="$(OutDir)\*.pdb" />
-      <Headers Include="..\dispatch\*.h" />
-      <Headers Include="..\src\*.h" />
+      <ReferenceCopyLocalPaths Remove="@(ReferenceCopyLocalPaths)" />
```
bbowman: ><ReferenceCopyLocalPaths Remove="@(ReferenceCopyLocalPaths)" /> [](start = 6, length = 63)

What is this trying to do and why?

tools/deps/prebuilt/include/dispatch/queue_internal.h
DHowett-MSFT: This is actually _reverting_ formatting changes we made to prebuilt headers years ago :)

msvc/sdk-build.props
rajsesh-msft: The dispatch binaries and headers are published from tools package, so yes, we shouldn't have to change this props file at all.

deps/3rdparty/libdispatch/build/libdispatch.vcxproj
DHowett-MSFT: Brian, you and I discussed this a long time ago. The Language package includes libdispatch, and if not ignored it will copy it over the newly compiled version.

deps/3rdparty/libdispatch/build/libdispatch.vcxproj
DHowett-MSFT: Msbuild requirement. A dangling \ will eat an adjacent “ in the compiler command line or the command line for any other tool.

MSFTFox: I was merely copying the double backslash from the changed line. In retrospect, we don't need this anyway. The target that used it was removed.

deps/3rdparty/libdispatch/build/project.json
```diff
@@ -0,0 +1,16 @@
+{
+  "dependencies": {
+    "WinObjC.Language": "*",
```
bbowman: Why? I don't see you adding code that uses ObjC?

deps/3rdparty/libdispatch/build/libdispatch.vcxproj
MSFTFox: Yeah, the target that used them was removed. This target copies over checked in files in the main repo. In an ideological sense, this is bad. We don't want to have any 3rdparty projects copying over files checked into the main repo.

deps/3rdparty/libdispatch/build/project.json
MSFTFox: LibDispatch must be compiled with ClangCompile though. This exists in the WinObjC Language package.

tests/UnitTests/Foundation/DispatchTests.mm
MSFTFox: Yeah, it seems to just be github acting strangely. The format in the file itself seems fine to me.

tests/UnitTests/Foundation/DispatchTests.mm
MSFTFox: Fixed.

PR #2822, opened 2017-12-04
Update libdispatch to use asynchronous destructor calls

MSFTFox:
The destructor provided for libdispatch specific objects should be called asynchronously.

Also address formatting issues in dispatch tests file.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2822)


deps/3rdparty/libdispatch/src/queue.c
```diff
@@ -2039,7 +2039,7 @@ static void _dispatch_queue_insert_specific(dispatch_queue_specific_t dqs) {
 	TAILQ_FOREACH(var, &dqsl->context_list, specific) {
 		if (var->key == dqs->key) {
 			if(var->destructor) {
-				var->destructor(var->context);
+				dispatch_async_f(_dispatch_get_root_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, false), var->context, var->destructor);
```
DHowett-MSFT: Should this be default priority, or lower?

tests/UnitTests/Foundation/DispatchTests.mm
```diff
@@ -28,173 +28,180 @@
...
+        EXPECT_EQ(2, destructorCount);
     });
-	ASSERT_EQ(0, dispatch_group_wait(group, _createEndTestTime()));
+    ASSERT_EQ(0, dispatch_group_wait(group, _createEndTestTime()));
 
-	dispatch_release(queue);
-	dispatch_release(group);
+    dispatch_release(queue);
+    dispatch_release(group);
+    dispatch_release(destructorSemaphore);
```
DHowett-MSFT: Please make sure to set this to null. Otherwise, subsequent tests will signal on a free’d semaphore!

I guess also check if the semaphore exists in the dtor callback ;p

deps/3rdparty/libdispatch/src/queue.c
MSFTFox: I don't see why it should be lower. It's not a high priority item, but I can't imagine the user wants to wait longer than they have to.

DHowett-MSFT: Waiting on a destructor is not likely to be a very common case though :)

MSFTFox: Since it's their only way of freeing those resources though, we shouldn't be a barrier to getting it done.

PR #2824, opened 2017-12-14
Fix NSLocalizedString and Add Localization Tests

### LINKED ISSUE #2791: Localized bundle resources don’t respect user’s system language preference
Using the NSBundle / CFBundle methods for retrieving localized bundle resources currently always returns the resource from the `Base.lproj` or `en.lproj` localized bundle folder irrespective of the user’s system language setting, which seems to be caused by `NSBundle.mainBundle.preferredLocalizations` always returning the first available localization (Base/English).

Note that `NSBundle.mainBundle.localizations` correctly returns a list of all the localizations contained in the bundle.

#### Steps to reproduce:

1. Add two folders "Base.lproj" and "de.lproj" to a project.
2. Inside each folder create a `Localizable.strings` file and add it to the project with an item type of "Starboard Resource".
3. Edit the .vcxproj file to specify the containing .lproj directory for the `SBResourceCopy` item for both strings files by adding a `<VariantDir>` entry like this:
```xml
<SBResourceCopy Include="..\..\Project\Base.lproj\Localizable.strings">
  <FileType>Document</FileType>
  <VariantDir>Base.lproj</VariantDir>
</SBResourceCopy>
```
4. Edit the .strings files to contain a key-value pair with the same key and different values, e.g.:
* Base.lproj/Localizable.strings: `"Test" = "Base";`
* de.lproj/Localizable.strings: `"Test" = "German";`
5. Use `NSLocalizedString(@"Test", nil)` to retrieve the localized value in the app.
5. Change the system’s language preference to "Deutsch (Deutschland)" (i.e. German) as the first option.
6. Run the app.

#### Expected result:

The string "German" is returned by the `NSLocalizedString()` call.

#### Actual result:

The string "Base" is returned by the `NSLocalizedString()` call.

MSFTFox:
Add automated tests using NSLocalizedStringFromTableInBundle and a fabricated bundle. Also add a page in WoCCatalog for testing various system languages and NSLocalizedString from the main bundle.

Fixes #2791

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2824)


Frameworks/CoreFoundation/PlugIn.subproj/CFBundle_Locale.c
```diff
@@ -531,6 +536,10 @@ CF_PRIVATE CFArrayRef _CFBundleCopyUserLanguages() {
     static CFArrayRef _CFBundleUserLanguages = NULL;
     static dispatch_once_t once = 0;
     dispatch_once(&once, ^{
+// WINOBJC: __CFAppleLanguages does not exist on Windows
+#if DEPLOYMENT_TARGET_WINDOWS
+        _CFBundleUserLanguages = EnumerateUserPreferredLanguages();
```
DHowett-MSFT: Per offline suggestion, I'd prefer we use `CFLocaleGetPreferredUserLanguages` or whatever it's called. That way we don't need to use our WinRT `Enumerate` directly everywhere.

Frameworks/CoreFoundation/PlugIn.subproj/CFBundle_Locale.c
```diff
@@ -663,8 +673,66 @@ static CFStringRef _CFBundleCopyLanguageFoundInLocalizations(CFArrayRef localiza
     return NULL;
 }
 
+// WINOBJC: Helper functions for workaround in _CFBundleCreateMutableArrayOfFallbackLanguages
+static CFStringRef _copyStringTruncated(CFStringRef localization, CFRange cutoff) {
+    CFMutableStringRef truncatedString = CFStringCreateMutableCopy(NULL, 0, localization);
```
DHowett-MSFT: `CFStringCreateWithSubstring`

Frameworks/CoreFoundation/PlugIn.subproj/CFBundle_Locale.c
```diff
@@ -663,8 +673,66 @@ static CFStringRef _CFBundleCopyLanguageFoundInLocalizations(CFArrayRef localiza
...
+
+static CFStringRef _copyStringWithUnderscores(CFStringRef localization) {
+    CFMutableStringRef underscoredString = CFStringCreateMutableCopy(NULL, 0, localization);
+    CFStringFindAndReplace(underscoredString, CFSTR("-"), CFSTR("_"), CFRangeMake(0, CFStringGetLength(underscoredString)), 0);
+    return underscoredString;
+}
+
 // Given a list of localizations (e.g., provided as argument to API, or present as .lproj directories), return a mutable array of localizations in preferred order. Returns NULL if nothing is found.
 static CFMutableArrayRef _CFBundleCreateMutableArrayOfFallbackLanguages(CFArrayRef availableLocalizations, CFArrayRef preferredLocalizations) {
+// WINOBJC: The API that performs the work described below does not exist in our 3rd party libraries
```
DHowett-MSFT: nit: add note that `ualoc_` is an apple-specific ICU API suite

Frameworks/CoreFoundation/PlugIn.subproj/CFBundle_Locale.c
```diff
@@ -663,8 +673,66 @@ static CFStringRef _CFBundleCopyLanguageFoundInLocalizations(CFArrayRef localiza
...
+#if DEPLOYMENT_TARGET_WINDOWS
+    // Here we need to intersect the preferred languages with the available localizations
+    // We know the user languages are in preferred order, add to this list in this order
+    // Prefer the full language locale, attempt to convert any hyphens to underscores as
+    // Windows language settings are retrieved with hyphens while underscores are commonly used for localization.
+    // Finally, attempt to truncate any underscores from the language to find a base localization.
```
DHowett-MSFT: nit: give examples of these. "The user's preferred language list comes in as en-US and de-DE, ..."

Frameworks/CoreFoundation/PlugIn.subproj/CFBundle_Locale.c
DHowett-MSFT: minor: should we be transforming this list from `-` to `_` on creation? it could save us the step of converting the `-` to a `_` for every lookup.

DHowett-MSFT: we could do it inside Enumerate if CF prefers `_` locale names

Frameworks/CoreFoundation/PlugIn.subproj/CFBundle_Locale.c
```diff
@@ -663,8 +673,66 @@ static CFStringRef _CFBundleCopyLanguageFoundInLocalizations(CFArrayRef localiza
...
+
+    CFMutableArrayRef resultArray = CFArrayCreateMutable(kCFAllocatorSystemDefault, 0, &kCFTypeArrayCallBacks);
+    
+    for (CFIndex i = 0, count = CFArrayGetCount(preferredLocalizations); i < count; i++) {
+        CFStringRef preferredLocalization = (CFStringRef)CFArrayGetValueAtIndex(preferredLocalizations, i);
+        for (CFIndex j = 0, count = CFArrayGetCount(availableLocalizations); j < count; j++) {
```
DHowett-MSFT: these multiple scope-nested counts scare me

Frameworks/CoreFoundation/PlugIn.subproj/CFBundle_Locale.c
```diff
@@ -663,8 +673,66 @@ static CFStringRef _CFBundleCopyLanguageFoundInLocalizations(CFArrayRef localiza
...
+                    CFArrayAppendValue(resultArray, truncatedLocalization);
+                }
+                CFRelease(truncatedLocalization);
+            }
+        }
+    }
+    if (CFArrayGetCount(resultArray) > 0) {
+        return resultArray;
+    }
+    return NULL;
```
DHowett-MSFT: it looks like the codepath below has fallbacks for `en_US` and `devlang` (they've called them "backstops").

Frameworks/CoreFoundation/PlugIn.subproj/CFBundle_Locale.c
```diff
@@ -785,6 +853,7 @@ static CFMutableArrayRef _CFBundleCopyPreferredLanguagesInList(CFArrayRef search
         }
     }
     
+
```
DHowett-MSFT: nit: extra line

samples/WOCCatalog/WOCCatalog.vsimporter/WOCCatalog-WinStore10/WOCCatalog.vcxproj
```diff
@@ -274,6 +276,22 @@
...
+    </SBResourceCopy>
+    <SBResourceCopy Include="de.lproj\Localizable.strings">
+      <FileType>Document</FileType>
+      <VariantDir>de.lproj</VariantDir>
+    </SBResourceCopy>
+    <SBResourceCopy Include="zh.lproj\Localizable.strings">
+      <FileType>Document</FileType>
+      <VariantDir>zh.lproj</VariantDir>
+    </SBResourceCopy>
+    <SBResourceCopy Include="zh-Hant-TW.lproj\Localizable.strings">
```
DHowett-MSFT: is this the apple official name or the windows name for this locale?

samples/WOCCatalog/WOCCatalog/LocalizationViewController.m
```diff
@@ -0,0 +1,81 @@
...
+@private
+    UITableViewCell* _cell;
+    UITextField* _copyTextField;
+    UILabel* _pasteTextLabel;
+}
+- (void)viewDidLoad {
+}
+
+- (NSInteger)tableView:(UITableView*)tableView numberOfRowsInSection:(NSInteger)section {
+    [self tableView].allowsSelection = YES;
```
DHowett-MSFT: nit: prefer `self.tableView.allowsSelection`

DHowett-MSFT: NOT A NIT: don't do this in the getter for "how many rows are there" !!!

tests/UnitTests/Foundation/WindowsOnly/BundleInternalTests.mm
```diff
@@ -0,0 +1,88 @@
...
+
+    NSBundle* bundle = [NSBundle bundleWithPath:[_playground stringByAppendingString:_bundleName]];
+    ASSERT_OBJCNE(bundle, nil);
+
+    NSString* helloWorld = @"Hello World";
+    NSString* coding = @"Coding";
+    NSString* helloWorldInGerman = NSLocalizedStringFromTableInBundle(helloWorld, nil, bundle, nil);
+    NSString* codingInChinese = NSLocalizedStringFromTableInBundle(coding, nil, bundle, nil);
+    ASSERT_OBJCEQ(@"Hallo Welt", helloWorldInGerman);
+    // Cannot use OBJCEQ as these unicode characters will cause a crash. isEqualToString is a code point literal comparison.
```
DHowett-MSFT: i'm sorry? what? a crash?

tests/UnitTests/Foundation/WindowsOnly/BundleInternalTests.mm
```diff
@@ -0,0 +1,88 @@
...
+public:
+    explicit NSBundleLocalization() : ::testing::Test() {
+    }
+
+protected:
+    virtual void SetUp() {
+        NSString* _subDirectory = @"en.lproj";
+        NSString* localizationResourceName = @"Localizable.strings";
+        // Create a unique test directory
+        NSString* tempDir = [[@"./tmp_TestFoundation" stringByAppendingString:[NSUUID UUID].UUIDString] stringByAppendingString:@"/"];
```
DHowett-MSFT: agh `stringByAppendingPathComponent:`

tests/UnitTests/Foundation/WindowsOnly/BundleInternalTests.mm
```diff
@@ -0,0 +1,88 @@
...
+
+        NSError* error = nil;
+        [[NSFileManager defaultManager] createDirectoryAtPath:tempDir withIntermediateDirectories:false attributes:nil error:&error];
+        ASSERT_EQ(nil, error);
+
+        // Make a flat bundle in the playground
+        NSString* bundlePath = [tempDir stringByAppendingString:_bundleName];
```
DHowett-MSFT: same

tests/UnitTests/Foundation/WindowsOnly/BundleInternalTests.mm
```diff
@@ -0,0 +1,88 @@
...
+        [[NSFileManager defaultManager] createDirectoryAtPath:bundlePath withIntermediateDirectories:false attributes:nil error:&error];
+        ASSERT_EQ(nil, error);
+
+        NSString* localizationPath = [NSString stringWithFormat:@"%@/%@", bundlePath, _subDirectory];
+        [[NSFileManager defaultManager] createDirectoryAtPath:localizationPath
+                                  withIntermediateDirectories:false
+                                                   attributes:nil
+                                                        error:&error];
```
DHowett-MSFT: just flatten these all into one path, but set `withIntermediateDirectories:true`.

tests/UnitTests/Foundation/WindowsOnly/BundleInternalTests.mm
```diff
@@ -0,0 +1,88 @@
...
+static NSString* _bundleName = @"MyBundle.bundle";
+
+class NSBundleLocalization : public ::testing::Test {
+public:
+    explicit NSBundleLocalization() : ::testing::Test() {
+    }
+
+protected:
+    virtual void SetUp() {
+        NSString* _subDirectory = @"en.lproj";
```
DHowett-MSFT: ?? you're only testing looking strings up in the single lproj that exists, which usually matches the system locale.

Frameworks/CoreFoundation/PlugIn.subproj/CFBundle_Locale.c
MSFTFox: Windows locales are set up to use `-`. If somebody is setting up localization, they may think to use the natural windows syntax in their bundle names. I think it makes sense to support both `-` and `_` in that case.

Frameworks/CoreFoundation/PlugIn.subproj/CFBundle_Locale.c
MSFTFox: Yup. I'm not sure what you mean by this though.

samples/WOCCatalog/WOCCatalog.vsimporter/WOCCatalog-WinStore10/WOCCatalog.vcxproj
MSFTFox: Windows.

Frameworks/CoreFoundation/PlugIn.subproj/CFBundle_Locale.c
DHowett-MSFT: what I mean by this is that you're behoven to do the same ;)

samples/WOCCatalog/WOCCatalog.vsimporter/WOCCatalog-WinStore10/WOCCatalog.vcxproj
DHowett-MSFT: then you're not exercising the `_` variants

tests/UnitTests/Foundation/WindowsOnly/BundleInternalTests.mm
MSFTFox: Yup, unicode characters in ASSERT_OBJCEQ was just like... ¯\_(ツ)_/¯ crash.

DHowett-MSFT: Compile time? Runtime? Either way, please file it

tests/UnitTests/Foundation/WindowsOnly/BundleInternalTests.mm
MSFTFox: The point was to always find a localization we have. Since we know we can use English as a fallback the test doesn't care about what locale or language settings the machine has. This way we can ensure we're testing the localization changes.

tests/UnitTests/Foundation/WindowsOnly/BundleInternalTests.mm
MSFTFox: Run time. Will do!

tests/UnitTests/Foundation/WindowsOnly/BundleInternalTests.mm
DHowett-MSFT: by testing the base localization, we aren't testing anything related to 1. localization or 2. localization changes though.

Frameworks/CoreFoundation/PlugIn.subproj/CFBundle_Locale.c
MSFTFox: Mmm, no? We want to return NULL so that the rest of the code will hit it's fallbacks. Where it creates "devLangArray" and an "en_US" array to call _CFBundleCreateMutableArrayOfFallbackLanguages.

Sorry if I'm missing something.

DHowett-MSFT: Thanks to me viewing code in github's web UI, I didn't see that the backstop is in a completely different function (the one that calls this one), and that returning is the correct thing to do.

tests/UnitTests/Foundation/WindowsOnly/BundleInternalTests.mm
MSFTFox: I mean, we're certainly testing the NSLocalizedString path. Grabbing a particular locale is not the goal here, nor is that something I think we can rely on with a unit test. The system can only have one locale setting, so it doesn't seem to matter to me which locale that is. Since English is a known fallback is that not acceptable?

Frameworks/CoreFoundation/PlugIn.subproj/CFBundle_Locale.c
```diff
@@ -675,9 +675,10 @@ static CFStringRef _CFBundleCopyLanguageFoundInLocalizations(CFArrayRef localiza
 
 // WINOBJC: Helper functions for workaround in _CFBundleCreateMutableArrayOfFallbackLanguages
 static CFStringRef _copyStringTruncated(CFStringRef localization, CFRange cutoff) {
-    CFMutableStringRef truncatedString = CFStringCreateMutableCopy(NULL, 0, localization);
-    CFStringDelete(truncatedString, CFRangeMake(cutoff.location, CFStringGetLength(truncatedString) - cutoff.location));
-    return truncatedString;
+    return CFStringCreateWithSubstring(NULL, localization, CFRangeMake(0, cutoff.location));
+    //CFMutableStringRef truncatedString = CFStringCreateMutableCopy(NULL, 0, localization);
```
DHowett-MSFT: commented!

samples/WOCCatalog/WOCCatalog/LocalizationViewController.m
```diff
@@ -28,10 +28,10 @@ @implementation LocalizationViewController {
     UILabel* _pasteTextLabel;
 }
 - (void)viewDidLoad {
```
DHowett-MSFT: nit: you _may_ need `[super viewDidLoad]` here for completeness' sake

tests/UnitTests/Foundation/WindowsOnly/BundleInternalTests.mm
```diff
@@ -30,20 +30,13 @@ virtual void SetUp() {
         NSString* _subDirectory = @"en.lproj";
         NSString* localizationResourceName = @"Localizable.strings";
         // Create a unique test directory
-        NSString* tempDir = [[@"./tmp_TestFoundation" stringByAppendingString:[NSUUID UUID].UUIDString] stringByAppendingString:@"/"];
-
+        NSString* tempDir = [[@"./tmp_TestFoundation" stringByAppendingPathComponent:[NSUUID UUID].UUIDString] stringByAppendingPathComponent:@"/"];
```
DHowett-MSFT: `/` is not a path component, so you can leave this call out.

tests/UnitTests/Foundation/WindowsOnly/BundleInternalTests.mm
```diff
@@ -30,20 +30,13 @@ virtual void SetUp() {
...
+        NSString* bundlePath = [tempDir stringByAppendingPathComponent:_bundleName];
```
DHowett-MSFT: funny question: why not just expose bundlePath in the fixture and use that instead of re-combining `_playground` and `_bundleName` later?

Frameworks/CoreFoundation/PlugIn.subproj/CFBundle_Locale.c
MSFTFox: Oh woops. I wanted to keep that around in case I goofed anything in the substring call.

tests/UnitTests/Foundation/WindowsOnly/BundleInternalTests.mm
MSFTFox: Not funny at all. That seems like a good idea to avoid extra string work. 👍

tests/UnitTests/Foundation/WindowsOnly/BundleInternalTests.mm
```diff
@@ -61,14 +57,15 @@ virtual void TearDown() {
     }
 
     StrongId<NSString> _playground;
+    StrongId<NSString> bundlePath;
```
DHowett-MSFT: `_`

Frameworks/CoreFoundation/PlugIn.subproj/CFBundle_Locale.c
```diff
@@ -25,6 +25,11 @@
 #endif
 #include <ctype.h>
 
+// WINOBJC: Helper for retrieving list of preferred languages
+#if DEPLOYMENT_TARGET_WINDOWS
+#include "_CFLocaleInternal.h"
```
DHowett-MSFT: is this still necessary?

DHowett-MSFT:
you don't need to take all of my feedback; you are free to reject some.

triplef:
Thank you for this fix!

Just to clarify, does this only support getting localized strings or does it enable getting any kind of localized bundle resource?

MSFTFox:
It should have fixed a fundamental problem in lproj lookup. The problem was that an API for looking up localized resources did not have support for retrieving user preferred languages. This has been solved.

If you have some other particular APIs you would like us to take a look at, please let us know. This fix was really focused on ensuring localized strings were functional.

triplef:
Sounds great, thanks!

PR #2828, opened 2018-01-16
Add additional project.json files to frameworks to target net40

MSFTFox:
Nuget projects in each frameworks now require a project.json each specifying the net40 framework. An issue on nuget has been filed as the frameworks should not require a net40 dependency.

https://github.com/NuGet/Home/issues/6543

This fix is required for Visual Studio 2017 update 15.4.5.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2828)

### DIFF
--- a/build/WinObjC.Frameworks.Core/project.json
+++ b/build/WinObjC.Frameworks.Core/project.json
@@ -1,0 +1,25 @@
+﻿{
+  "dependencies": {
+    "NuGet.Build.Packaging": "0.1.186"
+  },
+  "frameworks": {
+    ".NETFramework,Version=v4.0": {
+      "imports": [
+        "dotnet",
+        "netstandard1.0",
+        "netstandard1.1",
+        "netstandard1.2",
+        "netstandard1.3",
+        "netstandard1.4",
+        "netstandard1.5",
+        "netstandard1.6",
+        "net20",
+        "net35",
+        "net40",
+        "native",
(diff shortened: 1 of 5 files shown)


build/WinObjC.Frameworks.Core/project.json
```diff
@@ -0,0 +1,25 @@
+﻿{
+  "dependencies": {
+    "NuGet.Build.Packaging": "0.2.0"
```
DHowett-MSFT: ! DANGER !

I don't think we can build with 0.2.0. Are we specifying conflicting versions in the project versus in the project.json?

common/winobjc.packagereference.override.targets
```diff
@@ -110,24 +110,6 @@
     </ItemGroup>
   </Target>
 
-  <Import Project="$(MSBuildThisFileDirectory)\workaround\Microsoft.Nuget.Workaround.targets"/>
```
DHowett-MSFT: I'm a bit worried about this, as well. We don't need `NugetRestore` anymore? For any packages consuming IW?

MSFTFox: This was a workaround to have each project depend on a single thing to manually nuget restore from. It appears that this workaround isn't needed anymore. When running a regular build with the workaround removed, Visual Studio seems able to restore the nuget packages itself.

PR #2833, opened 2018-02-07
Make sure objc_2stage_init is included during all linking steps

MSFTFox:
This change moves the generation of the objc_2stage_init.cpp ClCompile item to
ComputeCompileInputsTargets.  BeforeClCompileTargets targets are only triggered
for ClCompile, which doesn't happen on every Link.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2833)


DHowett-MSFT:
since you changed the title/body, make sure to *squash and merge*, and copy the wording into the squash body.

Reviewable may have left some gunk at the bottom.

PR #2835, opened 2018-02-08
Update projection binaries and headers for 10.0.16299

MSFTFox:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2835)


DHowett-MSFT:
As discussed:
* please separate out the objc2winmd source changes
* do not target `170xx` SDKs :smile:
* use `10.0.16299.0` since that's what the SDK is called

PR #2843, opened 2018-02-14
Fix a few C++ and compiler longevity issues in our code

DHowett-MSFT:
This pull request fixes a few issues in our code and toolchain support:
* NSLayoutConstraint was constructing `basic_regex` instances incorrectly
* ClangCompileTask was parsing the (fragile) output of `--version` to determine Clang's version
* We've had a directory called `LLVM-3.6.0` that has, for some time, contained Clang **3.7.0**.

This also fixes a _very minor_ performance issue in parsing layout constraints by way of caching compiled regular expressions.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2843)


DHowett-MSFT:
Hilariously, this failed due to a disk space issue and not because I moved the compiler all around.

Issue #2838, opened 2018-02-09
Ongoing engineering improvements to ensure future Visual Studio compatibility


DHowett-MSFT:
* [ ] #2837

PR #2864, opened 2018-04-11
Fix a number of warnings and compilation errors

DHowett-MSFT:
Related to #2838. Please review each commit individually, as the commit messages explain what's happening.

This will be merged via rebase to keep individual commit histories.

```
commit 8a03129dfb8891afbedf2912fd7d667a91e01722 (origin/cleanupc6)
Author: Dustin Howett <duhowett@microsoft.com>
Date:   Mon Apr 9 16:39:23 2018 -0700

    Remove the old VS2015 extension

commit 347d4547384700868744426669c9e897d0ce54c5
Author: Dustin Howett <duhowett@microsoft.com>
Date:   Wed Apr 11 11:20:38 2018 -0700

    Fix a few easy warnings about which Clang 6 is particularly loud

    * -Wnonportable-include-path
    * -Wincompatible-pointer-types
    * -Wobjc-macro-redefinition
    * -Wduplicate-decl-specifier

commit 77ee071737692fe85fa8dcf6d8ffc3d3880658ec
Author: Dustin Howett <duhowett@microsoft.com>
Date:   Mon Apr 9 16:37:52 2018 -0700

    wil: fix two missing noinline specifiers

commit bb1e660457272ead0e0b66f36076cfa2313ef1db
Author: Dustin Howett <duhowett@microsoft.com>
Date:   Mon Apr 9 16:28:33 2018 -0700

    MSVC: Disable TR1 and Identity deprecation warnings for gtest

commit 3e809896e28f07f7f0a267c1b80855f4fecf6830
Author: Dustin Howett <duhowett@microsoft.com>
Date:   Wed Apr 4 13:47:33 2018 -0700

    MSVC: Clean up layer property animator initialization to avoid new CL errors

    We were initializing a huge map from name to layer property animators in
    constructor-call syntax `X()`, which confused CL no end.

    This commit switches the initializer to C++11 uniform initialization
    syntax and replaces std::function with plain function pointers, since
    that's exactly what they are and std::function offers little to no
    benefit at a fairly high cost.

commit 18d299ce059d8cef589de885ffe6594d43c9f3bb
Author: Dustin Howett <duhowett@microsoft.com>
Date:   Wed Apr 11 11:21:22 2018 -0700

    Export NSCFType from CoreFoundation for CoreGraphics
```

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2864)


tools/Logging/include/wil/result.h
```diff
@@ -3155,7 +3155,7 @@ inline void ReportFailure_Msg(
     ReportFailure(__R_FN_CALL_FULL, type, hr, message);
 }
 
-inline void ReportFailure_ReplaceMsg(__R_FN_PARAMS_FULL, FailureType type, HRESULT hr, _Printf_format_string_ PCSTR formatString, ...) {
+__declspec(noinline) inline void ReportFailure_ReplaceMsg(__R_FN_PARAMS_FULL, FailureType type, HRESULT hr, _Printf_format_string_ PCSTR formatString, ...) {
```
MSFTFox, line 3158: What? "declspec(noinline) inline"???

build/CoreFoundation/dll/CoreFoundation.def
```diff
@@ -8,6 +8,9 @@ LIBRARY CoreFoundation
         __objc_class_name__NSCFNumber CONSTANT
         _OBJC_CLASS__NSCFBoolean DATA
         __objc_class_name__NSCFBoolean CONSTANT
+	; Exposed for other CoreFoundation bridging implementors
+        _OBJC_CLASS__NSCFType DATA
```
MSFTFox: This doesn't look like it's being used in CoreGraphics (unless I missed something). Is this in preparation for something else? If so, shouldn't it be checked in alongside that?

tools/legacy/BuildMonitor/Legacy_BuildMonitor.csproj
```diff
@@ -1,225 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="15.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
MSFTFox, line 2: This seems like a lot to delete without having a replacement. Are we sure this can be tossed so easily?

I understand it's all "legacy" but I'd simply like to be sure.

tools/Logging/include/wil/result.h
DHowett-MSFT: This is like `declspec(selectany)`, but for functions: It's not _inlined_, but it's also not emitted as a symbol in each translation unit. Info [here](https://blogs.msdn.microsoft.com/freik/2005/10/26/how-to-make-a-no-inline-inline-function-and-why-you-might-want-to/).

build/CoreFoundation/dll/CoreFoundation.def
DHowett-MSFT: (rebuilding to verify why I did this)

tools/legacy/BuildMonitor/Legacy_BuildMonitor.csproj
DHowett-MSFT: 1. All the code will be available for eternity in git history
2. This doesn't compile because VS2017 stopped shipping the VS2015 extensibility toolkit.
3. VS2017 and new rule sets largely obviated the need for these guys.

MSFTFox (approved):

build/CoreFoundation/dll/CoreFoundation.def
DHowett-MSFT: This is required in CoreText, actually, because CT uses `CD_OBJC_FUNCDISPATCHV`, which uses `CF_IS_OBJC`, which uses `NSCFType`

PR #2867, opened 2018-04-19
Prepare further tests and build things

DHowett-MSFT:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2867)


MSFTFox (approved):

DHowett-MSFT:
The merging of this pull request marks the point of no return past which the agents must be updated to 15.6 and the compiler changes merged.

Issue #2806, opened 2017-10-23
Properties with class attributes appear to not work.


calbot:
I get the following error message...
`Build error unknown property attribute 'class'`

For this line
`@property (class, nonatomic, strong, readonly) DDLog *sharedInstance;`

Does WinObjC support class properties?
https://useyourloaf.com/blog/objective-c-class-properties/

MSFTFox:
WinObjc ships with Clang version 3.7. This predates the availability of class properties.

DHowett-MSFT:
Fixed with #2869

Issue #2341, opened 2017-03-28
Superclasses are not anchored by symbol references and can be stripped by the linker


edvv:
In a big project, a certain call to a method like [super M1] works. However when the project is factored into a library and an app is linked to that library then [super M1] is not called by the code that is now in the library. If a method to super is added that is named M1_hack and the calling code is changed to [super M1_hack] then [super M1_hack] is called (executed). This has been verified by NSLog() logging statements. As of yet, the code has not been stepped over to figure what is going on. This happens on Release 44 and 45 and also VS2015 and VS2017.

DHowett-MSFT:
Sorry, I'm having trouble following what's going on.

Is the superclass in the library? Are both the derived class and superclass in the library?

Is it a static library or a dynamic one?

edvv:
@DHowett-MSFT This problem persists in release 0.2.170407.

It is a static library. The derived and superclass are both in the static library, along with the entire class hierarchy of the calling sequence.

The workaround at https://github.com/Microsoft/WinObjC/issues/236 is being used, i.e.: Each TU has a function hook that is called to pull in the object file during linking. Note: This workaround is also used on iOS and it works there.

DHowett-MSFT:
Oh! Great find. Yeah, this is another facet of #236 amplified by some legacy decisions made in the GNUstep ABI.

For legacy compatibility, the GNUstep Objective-C ABI does not fill superclass slots with class pointers. It instead fills them with class names, which are later populated by the runtime.

### Apple ABI
```objc
struct objc_class {
    Class isa;
    Class superclass; // <- populated by the linker and fully resolved
    // ...
}
```

### GNUstep ABI
```objc
struct objc_class {
    Class isa;
    union {
        char* name; // <- populated by the linker
        Class ptr; // <- resolved at load time
    } superclass;
    // ...
}
```

This allows you to inherit from classes that are not visible at link time, but introduces a hidden loading dependency that requires the equivalent of `-all_load`.

edvv:
That could be it, or at least related.

If M1 is implemented like this in the superclass: 

```
- (void)M1
{
NSLog(@"** got here 2");
}
```

and in the derived class:

```
- (void)M1
{
[super M1];
NSLog(@"** got here 1");
}
```

Then the logging is:

** got here 1

but if it is implemented like this in the superclass: 

```
- (void)M1_hack
{
NSLog(@"got here 2");
}
```

and in the derived class:

```
- (void)M1
{
[...]

Issue #1647, opened 2017-01-06
AV in LIBOBJC2.DLL!isEmptyProtocol() when loading a module that implements *only* a forward-declared protocol


jaredhms:
I thought we fixed this last march, but I am still able to reproduce the issue easily:

1. Once #1380 goes in, comment out the bottom line of UIApplication.h that #imports UIApplicationDelegate.h
2. Rebuild the functional tests
3. Run the functional tests under the debugger

You'll hit an AV in isEmptyProtocol() because the functional tests implement an undefined/forward-declared UIApplicationDelegate protocol.

DHowett-MSFT:
Fixed with #2869

Issue #1140, opened 2016-10-13
C++ instances as Objective-C++ method parameters - crashes in LIBOBJC2.DLL


gary17:
Passing instances of non-trivial C++ objects as Objective-C++ method parameters causes unexpected behavior, even if appropriate copy constructors and assignment operators have been provided, e.g. in case of smart pointer implementations - crashes.

CppParam.h (MSCVP)

``` c++
#pragma once

class CppParam
{
public:
    CppParam();
};
```

CppParam.cpp (MSCVP)

``` c++
#include "CppParam.h"

CppParam::CppParam()
{
}
```

partyCrasher.mm (Clang)

``` objective-c
#include "CppParam.h"

// a lightweight smart pointer library, http://yasper.sourceforge.net
#include "yasper.h"

@implementation MainViewController

- (void)doMe1:(yasper::ptr<CppParam>)param1
{
    // NOOP
}

- (void)doMe2:(yasper::ptr<CppParam>)param1 alongWith:(NSObject*)param2
{
    // NOOP
}

- (void)doMe3:(NSObject*)param1 alongWith:(yasper::ptr<CppParam>)param2
{
    // NOOP
}

- (void)partyCrasher
{
    // a non-trivial C++ smart pointer object (with additional counter class members, etc.)
    yasper::ptr<CppParam> pParam(new CppParam());

    // single-param: ok
    [self doMe1:pParam];

    // multi-param: crash!
    [self doMe2:pParam alongWith:[[NSObject alloc] init]];

    // multi-param, reversed: crash!
    [self doMe3:[[NSObject alloc] init] alongWith:pParam];
}

@end
```

The relevant error message:

```
Exception thrown at 0x0FB17FEC (LIBOBJC2.DLL) in PartyCrasher.exe: 0xC0000005: Access violation reading location 0x00000000.
```

Notes:

1.) The problem does not occur under Apple Xcode/iOS.

2.) The problem does not occur for single-parameter Objective-C++ method calls (Clang requires, thus uses, a copy constructor), hence it is probably not a problem with `yasper::ptr<>` itself, the sample smart pointer implementation. Some kind of Objective-C++ variadic `memcpy()` -ing stuff around instead of using a copy constructor?

As of: WinObjC version 0.2.160927 (September 27, 2016).

DHowett-MSFT:
This looks like a code generation issue.

[...]

PR #2869, opened 2018-04-21
Move to a custom build of Clang 6.0.1

### LINKED ISSUE #1769: ARM compiler generates wrong calling convention dispatch for uniform VFP aggregate return across COM interface type.
Issue placeholder. Reproduction case will follow.

### LINKED ISSUE #2837: Clang 3.7 is incompatible with the STL from VS 2017 Update 5+
A number of intrinsics were added to CL in the VS 2017 Update 5 timeframe, none of which are supported on Clang 3.7. Further compiler requirements will be exacted in the future.

DHowett-MSFT:
This pull request moves WinObjC to Clang 6.

* All commits labelled `C6Compat` were changes required to build the project.
* All commits labelled `NFC` are not final/not-for-checkin.
* The directory `LLVM/lib` contains upstream vendored files.

Fixes #2837, #2806, #1647, #405, #2341, #1140, #160.
Closes #1769.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2869)


DHowett-MSFT:
This pull request also breaks ARM support.

DHowett-MSFT:
Fixes #2837 #2806 #1647 #405 #2341 #1140 #160.
Closes #1769.

DHowett:
**TODO**: Include patches in `tools/msvc/patches`, but don't package them.

MSFTFox:
Why the custom build of clang 6? I thought the intention was to move away from having our own binaries here.
*Edit*
I realized this is probably an intermediate step as our proposed changes get merged with official binaries. There's no reason for us to halt changes blocked on outside dependencies.

msvc/package-build.props
```diff
@@ -2,7 +2,8 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <PropertyGroup Label="Globals">
-    <TargetFramework>uap10.0</TargetFramework>
+    <TargetFramework>NuGet,Version=v1.0</TargetFramework>
+    <PackageTargetFallback>net11;net20;net35;net40;net403;net45;net451;net452;net46;net461;net462;net47;net471;netcore;netcore45;netcore451;netcore50;win8;win81;win10;sl4;sl5;wp;wp7;wp75;wp8;wp81;wpa81;uap;uap10;netstandard1.0;netstandard1.1;netstandard1.2;netstandard1.3;netstandard1.4;netstandard1.5;netstandard1.6;netstandard2.0;netcoreapp1.0;netcoreapp2.0;monoandroid;monotouch;monomac;xamarinios;xamarinmac;xamarinpsthree;xamarinpsfour;xamarinpsvita;xamarinwatchos;xamarintvos;xamarinxboxthreesixty;xamarinxboxone</PackageTargetFallback>
```
MSFTFox, line 6: Whoah.... Quite the fallbacks here. What are the ramifications of this? Did we severely spike which packages we require to contribute to Islandwood by including all these now? I'm all for manipulating NuGet if it helps us unblock certain troublesome scenarios but a giant list of all these frameworks seems strange in the least.

MSFTFox (approved):

msvc/package-build.props
edvv: Exciting stuff!!

DHowett-MSFT: This is copied from Nugetizer-3000's default template, and is required so that package target lookup doesn't fail during restore. It doesn't add any additional dependencies!

DHowett-MSFT:
We have clang changes that may never be suitable for upstreaming; our compromise is to upstream what we can and provide patches for what we can't.

wjk:
I have a question about this PR. I noticed that the MSVC-derived binaries in the LLVM/bin folder (`mspdb*`, `c2.dll`, etc.) have been removed. Having watched Clang's upstream development, I know that their MSVC compatibility (and, in particular, their ability to generate and manipulate PDB files) is much improved over where it was just a few years ago. Furthermore, @DHowett-MSFT's comment about "clang-6.0.0+ms" implies that we are moving to upstream(-ish) Clang instead of a heavily modified, Microsoft-specific fork. Should this PR be merged, will it mean that the Clang compiler used will be open-source (or, at least, can be replaced with a self-built Clang and not cause too much breakage)?

Even though WinObjC was never really suitable for its intended purpose IMHO, I still think it has some interesting ideas under the hood that I might want to extract and use in my own projects. (In particular, I was thinking of adding support for the Mac frameworks — AppKit et al — as well as the iOS ones, and creating a macOS-to-Windows app bridge.) Should the Microsoft-proprietary compiler be finally removed, this would greatly aid the implementation of these ideas.

Thanks for answering my question, and keep up the good work! 😄

DHowett-MSFT:
Hi @wjk!

That's a very astute observation. We're planning on upstreaming any Clang changes that make sense and switching to an upstream build at our earliest convenience. For now, though, we have patches staged for upstreaming in issue #2860, and the ones that haven't made it up will be included in this pull request.

I'm very interested in your plan to support Mac frameworks. Would you mind e-mailing me at duhowett@microsoft.com to continue this conversation?

PR #2871, opened 2018-05-02
Split WinObjC.Compiler out of WinObjC.Language

DHowett-MSFT:
This commit introduces an independent WinObjC.Compiler package that can
be released separately from the rest of the WinObjC infrastructure. It
contains only clang, ClangCompileTask and the header map maker.

objc2winmd, SBResource, ConsumeRuntimeComponent, the Objective-C
dependency, WinObjCRT and /WHOLEARCHIVE support all remain in
WinObjC.Language. objc2winmd depends on the "public header"
infrastructure in cl-additions (now ClCompile.Additions), so that lives
in WinObjC.Language as well.

tools/msvc has been split and replaced with WinObjC.Compiler/
and WinObjC.Language/assets.

A few options have been migrated to WinObjC.Language as part of this
split. Due to limitations in the VS project system's property sheet
schema, this required the introduction of another settings group, "Clang
(WinObjC)", to house items including "Use WOCStdLib."

This change also introduces a new flag bookkeeping item,
ClangCompile.InternalCompilerFlags, which is patched into the
commandline template by the task instead of inserted into it during rule
evaluation. This is to allow WinObjC.Language to add items to the list
of compiler flags before any user options are evaluated.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2871)


tools/WinObjC.Language/assets/Islandwood.xml
```diff
@@ -19,12 +19,6 @@
         DisplayName="Linking" />
     </Rule.Categories>
 
-    <BoolProperty
```
DHowett-MSFT, line 22: this died long ago, we never read it in sbclang.targets/props

DHowett-MSFT:
We probably don't need `depends.exe` and `depends.dll` in this package: they're unused.

tools/WinObjC.Compiler/WinObjC.Compiler.targets
```diff
@@ -0,0 +1,44 @@
...
+    <_DllsFromPackage Include="$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\*.dll"/>    
+    <_DllsFromPackage Include="$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\$(Configuration)\*.dll"/>
+
+    <_LibsFromPackage Include="$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\*.lib"/>    
+    <_LibsFromPackage Include="$(MSBuildThisFileDirectory)\lib\$(TargetOsAndVersion)\$(PlatformTarget)\$(Configuration)\*.lib"/>
+  </ItemGroup>
+
+  <ItemDefinitionGroup>
+    <ClangCompile>
+      <InternalSystemIncludePaths>$(MSBuildThisFileDirectory)\include\;%(InternalSystemIncludePaths);</InternalSystemIncludePaths>
```
DHowett-MSFT: may not be necessary.

tools/WinObjC.Compiler/WinObjC.Compiler.Packaging.targets
```diff
@@ -0,0 +1,38 @@
...
+
+  <Target Name="AddBuiltOutput" BeforeTargets="GetPackageContents" Condition="'$(SkipGetPackageContentsForOtherPlatforms)' != 'true'">
+      <ItemGroup>
+          <_AssetFiles Include="$(MSBuildThisFileDirectory)assets\**\*.*"/>
+
+          <!-- Include vcclang -->
+          <_DepsFiles Include="$(MSBuildThisFileDirectory)..\deps\**\vcclang.lib"/>
+
+          <!-- Include /bin/ folder. -->
+          <_BinFiles Include="$(MSBuildThisFileDirectory)..\bin\depends.*"/>
```
DHowett-MSFT: yeah, don't need these. we do need hmapmaker though

tools/WinObjC.Compiler/assets/ClangCompile.props
```diff
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <StarboardSdkRoot>$([System.IO.Path]::Combine('$(MSBuildThisFileDirectory)', '..'))</StarboardSdkRoot>
```
DHowett-MSFT: another unused guy!

MSFTFox (approved):

DHowett-MSFT:
Depends may not be safely removeable here. Issues:
* We need to copy WinObjCRT.dll into the built product
* We only want to copy it if it was used in the build inputs
* DO we need WinObjCRT in .Compiler? If not, we can make it clang-only... but if we do, we need a complicated dance to depend on WinObjCRT.lib from libobjc2 and libdispatch.

DHowett-MSFT:
Downsides to keeping WinObjCRT in this package: It's not "just clang and clang support" if we do that.

DHowett-MSFT:
Decision: Move winobjcrt and the other IW-specific command line args to WinObjC.Language.

DHowett-MSFT:
Description updated to reflect latest changes.

DHowett-MSFT:
Merged externally as part of #2869

PR #2873, opened 2018-05-07
Hygiene: fix 1687 filename case collisions & renormalize CRLF & LFS

DHowett-MSFT:
This is the result of checking out on a case-sensitive filesystem and resolving all conflicts in case, then running `git add --renormalize .`. Binary files under `deps/3rdparty` have been reverted.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/winobjc/2873)


DHowett-MSFT:
It'll be easier to review this in whitespace-agnostic mode, [here](https://github.com/Microsoft/WinObjC/pull/2873/files?utf8=%E2%9C%93&diff=split&w=1).

MSFTFox:
I'm a bit confused. Why are the reference images being moved (read renamed) to the actual data folder? Isn't this where the test's generated images go?

DHowett-MSFT:
Good eye.

MSFTFox (approved):

DHowett-MSFT:
I've run the diff list through `uniq -ic` (which spits out a list of each line in a file prefixed by how many times it appears) and verified that each filename appears identical (case-insensitive) exactly twice -- that is to say, no files have moved to different folders this time.

PR #2913, opened 2018-12-10
tests: consolidate all properties into ut-build, make it match sdk-build

DHowett-MSFT:
The critical difference between `sdk-build` and `ut-build` is that `sdk-build` was intended to be included _instead of_ `Microsoft.Cpp.Common`, which allowed it to change a whole host of settings before the C++ infrastructure was initialized. This commit switches `ut-build` to the same pattern. It also removes an obscene number of identical properties and simplifies the test projects quite a bit.


MSFTFox:
"-3,230" Definitely an obscene number of duplicate properties. 👍

MSFTFox (approved):

PR #2938, opened 2019-12-21
Fixed generics declarations for collection subclasses.

triplef:
Generics declarations of NSMutableArray, NSMutableDictionary, NSMutableOrderedSet, and NSMutableSet were not being passed up to superclass.

This meant that calling generic methods declared in the superclass on a subclass object were not recognized by the compiler to return a declared generic object type.

All protocol declarations that are being replaced with generics declarations are already declared by the respective superclass, and thus the list of implemented protocols does not change. Also removed unecessary protocol declarations on NSMutableIndexSet to match the other collection subclasses.

Also added missing generics declaration to NSCountedSet (subclass of NSMutableSet).

### DIFF
--- a/include/Foundation/NSCountedSet.h
+++ b/include/Foundation/NSCountedSet.h
@@ -26,6 +26,6 @@
 @class NSEnumerator;
 
 FOUNDATION_EXPORT_CLASS
-@interface NSCountedSet : NSMutableSet <NSCopying, NSFastEnumeration, NSMutableCopying, NSSecureCoding>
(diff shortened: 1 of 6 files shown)


DHowett-MSFT (approved):
This is excellent. Thank you!

DHowett-MSFT:
I’ll run a CI build, but since it’s the holidays i can’t promise too much time spent getting this merged. :)

davidchisnall:
@DHowett-MSFT CI build seems not to have finished - what happened?

